Issue #261, opened 2011-06-10
Syntax for specifying image size


jgm:
Googlecode Issue #29

There is currently no way to specify the size of an image.
For some ideas, see http://code.google.com/p/pandoc/wiki/ImageSizes.

Jmuccigr:
Just bumping up against this problem today and wondering how far an implementation is?

Thanks.

jgm:
+++ John Muccigrosso [May 11 15 10:33 ]:
>   Just bumping up against this problem today and wondering how far an
>   implementation is?

I have an implementation on a branch in this repository, but
it's not planned for the next release (1.14).  We'll
probably have this in the release after; it's a big breaking
change.

Dashed:
@jgm Is it this branch? https://github.com/jgm/pandoc/commits/new-image-attributes

jgm:
Yes that's it.

+++ Alberto Leal [May 11 15 18:47 ]:
>@jgm Is it this branch? https://github.com/jgm/pandoc/commits/new-image-attributes

frumbert:
Also looking for this change; I'm using the windows binary. It's pretty similar to the format I already use with my custom parser (e.g. ![image caption](image.jpg {width: "500px"}) - this format just pushes the attrib map outside rather than inside the link grouping. Also in my implementation if the output format is HTML5, any custom "unknown" attributes are prefixed with "data-", so that it maintains as much of the input document as possible - such that {width: "500px", id: foo, cake: "lie"} becomes width="500px" id="foo" data-cake="lie".

I also had it [attrib map] implemented on standard hyperlinks, is that part of this change/branch?

RobTrew:
[pandoc.org/installing.html]() suggests [for Windows | OS X] that
> There is a package installer at pandoc’s [download](https://github.com/jgm/pandoc/releases/tag/1.14.0.3) page

I can't (2015-06-15) see installers at [https://github.com/jgm/pandoc/releases/tag/1.14.0.3]() ...

Should perhaps I be looking elsewhere in space, or in time ?

nkalvi:
https://github.com/jgm/pandoc/releases

A bit inconvenient - hopefully fixed soon :smile:

magnis:
[...]

PR #350, opened 2011-11-21
markdown additional Image width parameter for HTML/Epub, LaTeX/PDF output formats

### LINKED ISSUE #332: implement image width (percent) in pandoc markdown
To support image width specification in input languages like LaTeX,
a way to specify that value in pandoc markdown would be needed.
It will be sufficient to somehow give a percentage of the text width
as an integer between 1 and 100.

Proposed possibilities are (which would be best?):

```
![Caption](/path/to/file)@50
![Caption](/path/to/file @50)
![Caption]@50(/path/to/file)
```

No value assumes the default 100.

rwst:
Hello,
this resolves #332 and provides implementations in the HTML and LaTeX writers.
It is dependent on the two patches

* https://gist.github.com/1380600
* https://gist.github.com/1382799

What does it do? The markdown

    alttext

will produce markdown output

    alttext@100

or HTML output

    <div class="figure">
    alttext<p class="caption">alttext</p>
    </div>

or LaTeX output

    \begin{figure}[htbp]
    \centering
    \includegraphics[width=1.0\textwidth]{image.jpg}
     \caption{alttext}
     \end{figure}

Any digits after the @ character will be interpreted as width percentage, atm

### DIFF
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -350,7 +350,7 @@ pImage = do
   let url = fromAttrib "src" tag
   let title = fromAttrib "title" tag
   let alt = fromAttrib "alt" tag
-  return [Image (toList $ text alt) (escapeURI url, title)]
+  return [Image (toList $ text alt) (escapeURI url, title) "100"]
 
 pCode :: TagParser [Inline]
 pCode = try $ do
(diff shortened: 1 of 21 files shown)


strout:
Is this PR still relevant now that #261 is closed? Should it be closed?

jgm:
This can be closed.

PR #497, opened 2012-04-23
Text/Pandoc/Writers/PsuedoPod - implementation to merge, with tests

castaway:
Hi John,

I wrote this to use for my book source, the current version has coped with all the markdown-to-psuedopod I've thrown at it in the last month or so, so I would like to request that you merge this in to the main project.

The content I've been using it on can be found at: http://github.com/castaway/dbix-class-book .

Please let me know if there's anything I need to add before you can change it.. I'm not quite sure where docs for it should go, if anywhere.

Thanks!

Jess


danse:
do we want to keep this pull request open? is there enough community traction to justify the maintenance of this new writer?

PR #534, opened 2012-06-09
Linebreak handling control

favonia:
There are languages (Chinese, Japanese, etc) that do not use spaces to separate words. This is my hack to add a new option `--linebreak` to change the default newline conversion for `markdown` input. It can be either `space` (the default, original behavior), `skip`, or `preserve` now. I'd like to hear your opinions on this and work on this patch more. For example make this option work for other kinds of input.

Ideally we should be able to mark a particular region with a particular language, and the conversion should be context-aware. I only implemented the basic control here, and it's valid only for `markdown` input. For related information one can see [`text-space-collapse` in current CSS3 draft](http://www.w3.org/TR/css3-text/#white-space-collapsing) and [some question on stackoverflow](http://stackoverflow.com/q/8550112/985527).

### DIFF
--- a/README
+++ b/README
@@ -237,6 +237,11 @@ Reader options
 `-p`, `--preserve-tabs`
 :   Preserve tabs instead of converting them to spaces (the default).
 
+`--linebreak=`*METHOD*
+:   Specify the treatment of newline characters within paragraphs.
+    *METHOD* can be `space`, `skip`, or `preserve` (defaults to `space`).
+    This option currently has effect only on `markdown` input.
+
 `--tab-stop=`*NUMBER*
 :   Specify the number of spaces per tab (default is 4).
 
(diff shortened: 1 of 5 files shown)


episodeyang:
> Finally, if your text is mixed, with both Chinese and English, say, then
you can just avoid using hard line breaks in the English text.

or avoid using soft line breaks in Chinese text. This is the solution I'm going with atm.

jgm:
I believe this can be closed, now that we have the `ignore_line_breaks` and `east_asian_line_breaks` extensions.

jiucenglou:
> I believe this can be closed, now that we have the `ignore_line_breaks` and `east_asian_line_breaks` extensions.

 I have markdown files that actually use spaces to separate groups of Chinese words. Presumably, the spaces were used to make the content easier to understand. Now I would like to convert the markdown files to Word, ignore spaces between Chinese characters before converting to Word. Could you help to comment whether pandoc can handle this case via something such as east_asian_word_breaks ?

Issue #418, opened 2012-02-13
running parsers in API


noinia:
Would it be possible to extend the API so external programs using pandoc as a library can run the readers themselves? My reason for requesting this is because it allows for more flexibility in error handling. 

In my particular case I'm trying to run pandoc to convert a file in markdown format into html (as part as a webapplication in snap). In case pandoc will fail to parse the input file correctly some other part of my code should be executed. It seems that the API does not export the required functions to allow for this. I think there are two options to fix this: 

1) Provide an alternative function to `readWith' as found in Text.Pandoc.Parsing that simply returns an Either. e.g.

``` haskell
readWith' :: GenParser t ParserState a     -- ^ parser
         -> ParserState                                  -- ^ initial state
         -> [t]                                                  -- ^ input
         -> Either ParseError a
readWith' parser state input = runParser parser state "source" input 
```

or 

2) export the parse\* functions in Text.Pandoc.Readers.\* (e.g. parseMarkdown ) so a program can run a particular parser itself. 

In this case (1) would be sufficient. But (2) is a bit more flexible.

jgm:
We now do return an Either, so this was long ago fixed.

PR #567, opened 2012-07-14
Exposes parsers in the API (issue #418)

noinia:
I managed to get the changes required to fix issue 418 (https://github.com/jgm/pandoc/issues/418) in git. Hopefully this can cat committed and fixed.


jgm:
Closing because similar changes have already been made; the readers now return Either PandocError Pandoc.

PR #570, opened 2012-07-16
Add image caption support for ODT

ArloL:
I looked at how OpenOffice.org 3.4.0 renders image captions and tried to implement it for pandoc.

Please do check for code style issues or possible problems as I am still _very_ new to Haskell.

Currently the numbers are all set to 1 when opening the output file but "Tools -> Update -> Update All" updates all of the numbers properly.


blindmelon:
I'm very interested in having proper (i.e. using frames, autonumbering) captions for figures in .odt.
I think the implementation in this pull request might be slightly overcomplicated (see #2401 for an xml snippet) but I think it is much better to use the native method for captioning than the manually added captions odt currently receives from pandoc.

jgm:
Figure numbers have been dealt with now in commit ecd4d5b8d8cfda6a2cd8d8fb631e0d7c79bee363.
We don't have proper frames yet, though.

ArloL:
I will close this PR since it's been 7 years and the code can't be used anymore :)

PR #1163, opened 2014-02-15
Extended the rST parser's handling of roles.

merijn:
Extended the rST parser's handling of roles to more closely match rST reference.

rST parser now supports:
    \* All built-in rST roles
    \* New role definition
    \* Role inheritance

Issues/TODO:
    \* Silently ignores illegal fields on roles
    \* Silently drops class annotations for roles
    \* Only supports :format: fields with a single format for :raw: roles, requires a change to Text.Pandoc.Definition.Format to support multiple formats.
    \* Allows direct use of :raw: role, rST only allows indirect (i.e., inherited use of :raw:).


gwern:
Were warnings ever added for the rST parser? I was reading mickeynp's writeup of writing his [2015 book _Mastering Emacs_](https://www.masteringemacs.org/article/how-to-write-a-book-in-emacs), where he commented about Pandoc:

> Ultimately I had to delve into the source to figure out why some things worked and other things did not. As it turns out, reStructuredText is not *fully supported* like its markup cousin, Markdown; or the red-headed stepchild of Markdown, Pandoc Markdown, Pandoc’s own take on Markdown.
> 
> I found this part exceptionally frustrating. That one format or the other is not equally supported is totally fine, but I wish they’d spelled out what it could and could not do. I may have picked a different source format had I known — or maybe not, who knows?
> 
> For instance, in reStructuredText you can annotate text with _roles_. There’s a handful in the specification, but I wanted one for small caps, like so:
> 
> ```
> :small-caps:`GNU`
> ```
> 
> I discovered Pandoc converts everything into a kind of Abstract Syntax Tree in JSON format. So I was thinking: great, I’ll use Pandoc’s ability to call out to Python with each token from the AST and then hand-generate the LaTeX (and HTML for ePub) that I need. Nope. Not gonna happen.
> 
> It turns out that Pandoc “helpfully” strips roles and directives that it doesn’t recognize. So I figured: let’s hack the damn thing so it does what I need. But then I learn it’s written in Haskell. All right, I thought, the parser’s clean and easy to understand, let’s get a Haskell dev environment set up. Cue 900 MB of library cruft I had to download to even *run* the damn thing from source. I got flashbacks of bower, and npm, and the entire JS ecosystem and decided, in the end, to bard the reStructuredText source file with custom markers that I then map into LaTeX and HTML with sed.

jgm:
Yes, warnings have been added. Also:
```
% pandoc -f rst
:small-caps:`GNU`
<p><code class="interpreted-text" role="small-caps">GNU</code></p>
```

PR #1535, opened 2014-08-14
Markdown writer should obey Ext_backtick_code_blocks

liesen:
This changes the behavior of the Markdown writer so that it will always emit backticks around code blocks if `Ext_backtick_code_blocks` is enabled.


begriffs:
I recently converted a bunch of HTML files to markdown, and I wanted the `<pre>` blocks to turn into backtick-fenced rather than indented blocks in the markdown. What I ended up doing was adding a class attribute to each `<pre>` because it was easier to do that than to convert the blocks in the markdown result. My vote would be to always convert to the fenced style when the backtick_code_blocks extension is specified.

PR #1545, opened 2014-08-17
Overhaul line blocks.

neongreen:
Line blocks used to be parsed line-by-line, which made emphasis spanning across lines, using blockquotes inside of line blocks, etc. impossible. Now their content is parsed with standard block parser with `stateHardBreaks` set, which is a new flag in `ParserState` instructing end-of-line parser to emit hard line breaks when the flag is set. Additionally, the `code` parser (only in Markdown reader) handles code elements spanning across several lines correctly.


jgm:
Closing now that we have the first-class line blocks.

jgm:
If there's something your original PR achieved that pandoc still doesn't do, feel free to open an issue.

PR #1554, opened 2014-08-21
Add --parts command line option to LaTeX writer.

cwoac:
Add --parts command line argument.
This only effects LaTeX writer, and only for non-beamer output formats.
It changes the output levels so the top level is 'part', the next
'chapter' and then into sections.


ickc:
Since we now have `--top-level-division`, should this be closed?

PR #1571, opened 2014-08-27
LaTeX writer: Use a declaration for tight lists

jlduran:
Currently, pandoc has hard-coded the following in order to make tight lists in LaTeX:

```hs
text "\\itemsep1pt\\parskip0pt\\parsep0pt"
```

Which is fine, but does not allow customizations.  For example, the `memoir` class already has a `\tightlist` declaration for this purpose:

```tex
\newcommand{\tightlist}{%
  \setlength{\itemsep}{0pt}\setlength{\parskip}{0pt}}
```

I'm proposing to use a similar solution:

```diff
@@ In Writers/LaTeX.hs:
-then text "\\itemsep1pt\\parskip0pt\\parsep0pt"
+then text "\\tightlist"

@@ In templates/default.latex:
+\newcommand{\tightlist}{%
+  \setlength{\itemsep}{1pt}\setlength{\parskip}{0pt}\setlength{\parsep}{0pt}}
```

This allows us to customize the tightness to our needs.

### Backward Compatibility

If a person is using a custom LaTeX template (not based upon the `memoir` class), the `\tightlist` declaration must be added.

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -411,7 +411,7 @@ blockToLaTeX (BulletList lst) = do
   let inc = if incremental then "[<+->]" else ""
   items <- mapM listItemToLaTeX lst
   let spacing = if isTightList lst
-                   then text "\\itemsep1pt\\parskip0pt\\parsep0pt"
+                   then text "\\tightlist"
                    else empty
   return $ text ("\\begin{itemize}" ++ inc) $$ spacing $$ vcat items $$
              "\\end{itemize}"
@@ -446,7 +446,7 @@ blockToLaTeX (OrderedList (start, numstyle, numdelim) lst) = do
                         else "\\setcounter" <> braces enum <>
                               braces (text $ show $ start - 1)
   let spacing = if isTightList lst
-                   then text "\\itemsep1pt\\parskip0pt\\parsep0pt"
+                   then text "\\tightlist"
                    else empty
   return $ text ("\\begin{enumerate}" ++ inc)
          $$ stylecommand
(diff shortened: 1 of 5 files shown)


jiuks:
I'm getting a tightlist error when I try and convert documents now - is there some extra configuration required.

    ! Undefined control sequence.
    l.80 \tightlist

nkalvi:
@jiuks Are you getting this with Pandoc 1.14? It was released a few hours ago.
If the error is with 1.14, could you please provide sample source and the command line so that I can also test?

jlduran:
@jiuks You are probably using a custom template. `pandoc`now uses a `\tightlist` declaration that must be defined in your preamble. Try adding:

    \providecommand{\tightlist}{%
      \setlength{\itemsep}{0pt}\setlength{\parskip}{0pt}}

To the preamble of your custom template.

jiuks:
Thanks - I downloaded the default template and re-configured mine, so all now fixed.

LiamHuang0205:
This change could influent lots of customized templates by resulting in an LaTeX error:

```
! Undefined control sequence.
l.40 \tightlist
```

Since Pandoc doesn't show the specific `.tex` source file to be built, this kind of error comes hard to locate.

In addition, in the world of LaTeX, we don't recommend one to use this kind of tricky to define a tight list. The better way, and also the suggested way of making a tight list is to use the package `enumitem` as follows:

```TeX
\usepackage{enumitem}
\setlist{nosep, noitemsep}
```

jgm:
The advantage of the present method is that it doesn't add a
dependency on an external package, enumitem.  It also gives
users the ability to customize the amount of spacing.

If you use the default template, you'll be all right.
If you're using an old, customized template, then with the
change you propose there will still be an error.

I think this issue needs more discussion, but it should be
on pandoc-discuss rather than here (where only a few people will see it).

+++ Liam Huang [Jul 22 15 23:28 ]:

jlduran:
@LiamHuang0205 looking at the source of [enumitem](http://ctan.org/tex-archive/macros/latex/contrib/enumitem), the current declaration of `\tightlist` should do exactly the same thing, without the need to include an additional package.

Also remember that pandoc differentiates between [loose and compact lists](http://pandoc.org/README.html#bullet-lists). Wouldn't the use of your suggestion affect both types of lists?

PR #1689, opened 2014-10-14
Use '=' instead of '#' for atx-style headers in markdown+lhs.

kuribas:
Using atx-style headers in markdown+lhs doesn't give a legal literate haskell file, since the usage of '#' is reserved.  I suggest supporting '=' instead of '#' for headings of level 3 or more.  This should pose no compatibility problems, since '#' is already illegal, and the '=' syntax is not used for anything else.  Using html style subheaders works only when generating html, not for other backends.

### DIFF
--- a/README
+++ b/README
@@ -3080,7 +3080,8 @@ literate Haskell source. This means that
 
   - In markdown input, "bird track" sections will be parsed as Haskell
     code rather than block quotations.  Text between `\begin{code}`
-    and `\end{code}` will also be treated as Haskell code.
+    and `\end{code}` will also be treated as Haskell code.  For
+    atx-style headers the character '=' will be used instead of '#'.
 
   - In markdown output, code blocks with classes `haskell` and `literate`
     will be rendered using bird tracks, and block quotations will be
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -487,9 +487,15 @@ block = do
 header :: MarkdownParser (F Blocks)
 header = setextHeader <|> atxHeader <?> "header"
 
+atxChar :: MarkdownParser Char
+atxChar = do
+  exts <- getOption readerExtensions
+  return $ if Set.member Ext_literate_haskell exts
+    then '=' else '#'
+
 atxHeader :: MarkdownParser (F Blocks)
 atxHeader = try $ do
-  level <- many1 (char '#') >>= return . length
+  level <- atxChar >>= many1 . char >>= return . length
   notFollowedBy $ guardEnabled Ext_fancy_lists >>
                   (char '.' <|> char ')') -- this would be a list
   skipSpaces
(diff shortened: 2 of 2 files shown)


kuribas:
No feedback on this feature?

jgm:
+++ Kristof Bastiaensen [Jul 03 15 06:23 ]:
>   No feedback on this feature?

It just seemed a bit ad hoc to me.  I don't want to change
Markdown syntax in unprincipled ways.  So the change never
seemed right to me, though I do understand the problem.

kuribas:
What solution do you suggest?

kuribas:
If you have a better solution, I am willing to code it up.

jgm:
+++ Kristof Bastiaensen [Jul 04 15 04:01 ]:
>   What solution do you suggest?

You probably won't like any of these, but:

1.  Ask yourself whether you really need three levels of
heading in the document body.  (The top-level header can
be in the template.)

2.  Use paragraphs with boldface text for 3rd level
headings.  If that isn't good enough, you can use a filter
to convert these to Heading 3.

3.  Use LaTeX or reST.

4.  If you're just targeting HTML, you can use HTML tags.

kuribas:
I am currently using markdown, and running into this issue again.  Regarding your solutions:

1. Yes I need it.
2. Mixing markup with meaning is not an acceptable solution.  If I want to use a filter, I might as well use sed to preprocess my source before sending it to pandoc.  But this is not portable across systems.
3. Bypassing the issue.  Markdown is easier to read then LaTeX and more familiar than reST.
4. I need pdf too.

kuribas:
> I don't want to change Markdown syntax in unprincipled ways.

markdown+lhs is already unprincipled.  `>` is supposed to mean `blockquote`, not haskell code, but it is useful to do so, and so is my workaround.

jgm:
Your change would also break existing markdown+lhs documents.

jgm:
All right.  I'll merge this.

kuribas:
Thanks!  I hope it doesn't break existing documents, but it should be fine.  A header must be preceded by a blank line, only a line in the form of 

> === sometext

would be affected, which would be rather strange in a normal tekst

conal:
I'm very glad for this change, as I've been wishing for several-level headers in literate Haskell programs that don't choke CPP.

Meanwhile, however, I've been using ATX-style headers in my gitit-based journals for several years using Markdown+LHS. I recently upgraded gitit, and noticed that the ATX headers stopped working. Some digging around led me to this page.

So that I don't have to fix *all* previous journal pages, I made a gitit plugin that recognizes the old style and converts it. The new project is [gitit-atx-plugin](https://github.com/conal/gitit-atx-plugin). Share and enjoy. Comments and pull requests welcome.

conal:
To clarify, the [gitit-atx-plugin](https://github.com/conal/gitit-atx-plugin) recognizes and converts ATX-style headers, including '#'.

PR #1771, opened 2014-11-20
export more Html functions and use StateT monad transformer instead of S...

freiric:
...tate in order to implement Toc extending to subpage in gitit2:
jgm/gitit2/issues/11
Additional exported functions: `defaultWriterState`, `WriterState`, `inlineListToHtml`, `unordList`

### DIFF
--- a/src/Text/Pandoc/Readers/Muse.hs
+++ b/src/Text/Pandoc/Readers/Muse.hs
@@ -32,7 +32,6 @@ TODO:
 - {{{ }}} syntax for <example>
 - Page breaks (five "*")
 - Headings with anchors (make it round trip with Muse writer)
-- Verse markup (">")
 - Org tables
 - table.el tables
 - Images with attributes (floating and width)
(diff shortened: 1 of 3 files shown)


orezpraw:
I need this too (because I'm using @freiric 's code where he needed this.)

tarleb:
I don't have an opinion on whether this makes sense in general, so I'll just add some technical comments:

- Every function that's exported from a module should have a doc string.
- `Monad m` implies `Functor m`, it is sufficient to just specify the former in type definitions.
- Commits should be atomic and have a descriptive and short (≤ 50 chars) commit message.  A good split could be to have a commit for the `StateT` type change and one for the additional module exports.

I'm also wondering if this generalization would have a performance impact, although I guess that it doesn't (since `type State s = StateT s Identity`).

freiric:
Thanks for your comments, they are very welcome. I added  the doc string, and splitted the commit as you advised. I did not remove the Functor from the type definition, because I was not sure whether it is safe with GHC prior to 7.10.

PR #1791, opened 2014-12-07
Adding fenced div blocks to markdown reader

### LINKED ISSUE #168: Syntax for divs
<pre>
I've been missing a way to specify a div markdown
without using HTML tags.  The (relatively) new 
delimited code block syntax gave me an idea however.
Consider using a line of four or more periods to
start a div:

    Para before div

    ....
    First para inside div

    Second para inside div

    Third para inside div
    ....

Since one will often want and need to apply an ID
and/or class to a div one should be able specify
those.  I suggest a CSS-like syntax inside braces
after the opening line of periods

    .... {#id .class}
    div content
    ....

One problem would be nested divs.  I suggest marking
those by indenting.  A code/pre block containing div
syntax would need to be doubly indented, similar to 
such blocks inside lists:

    .... {#outer_div}
    
    Para of outer div

        .... {#nested_div}
        
        Para of nested div.

        A code block illustrating div syntax
        (a most unusual thing!):

                ....{#example_div}
                This is how you specify a div
                ....

        Another para in nested div

        ....

    Another para in outer div

    ....
markdown
Of course one could mark nested divs by more periods in the
line (and then at least two more!):

    .... {#outer_div}
    
    Para of outer div

    ...... {#nested_div}
    
    Para of nested div.

    A code block illustrating div syntax
    (a most unusual thing!):

        ....{#example_div}
        This is how you specify a div
        ....

    Another para in nested div

    ......

    Another para in outer div

    ....

But I think indenting makes things cleaner, making it
easier to remember closing the divs you have opened, and not 
to close more divs than you have opened.  If you have very 
many nested divs you should probably pause and consider why 
anyway!

Google Code Info:
Issue #: 195
Author: bpjonsson
Created On: 2010-01-07T13:54:47.000Z
Closed On: 
</pre>

mpickering:
This is speculative - I want to move discussion forward by having something concrete to talk about. 

Closes #168

### DIFF
--- a/README
+++ b/README
@@ -2303,6 +2303,16 @@ For the most part this should give the same output as
 `markdown_in_html_blocks`, but it makes it easier to write pandoc
 filters to manipulate groups of blocks.
 
+#### Extension: `fenced_divs` ####
+
+In addition to supporting using native HTML, it is often clearer to use a
+*fenced* div block. These begin with a row of three or more `^`, contain any
+kind of block level content and end with a row of an equal number of `^`.
+
+Attributes can be specified in exactly the same way as for
+[fenced code blocks]{#fenced-code-blocks}.
+
+
 #### Extension: `native_spans` ####
 
 Use native pandoc `Span` blocks for content inside `<span>` tags.
--- a/src/Text/Pandoc/Options.hs
+++ b/src/Text/Pandoc/Options.hs
@@ -109,6 +109,7 @@ data Extension =
     | Ext_implicit_header_references -- ^ Implicit reference links for headers
     | Ext_line_blocks         -- ^ RST style line blocks
     | Ext_epub_html_exts      -- ^ Recognise the EPUB extended version of HTML
+    | Ext_fenced_divs         -- ^ Enable div blocks surrounded by ~
     deriving (Show, Read, Enum, Eq, Ord, Bounded)
 
 pandocExtensions :: Set Extension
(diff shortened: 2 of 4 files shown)


ousia:
@mpickering, many thanks for this. It looks really awesome.

Contributing to the discussion (although I’m not sure whether this is the best place to do it): would it be possible to have also fenced spans (such as in `^span^{.class #identifier lang="grc"}`?

bpj:
@mpickering: please note what @jgm said at <https://github.com/jgm/pandoc/issues/168#issuecomment-65944322>.  I don't think 'uparrow' makes very much sense in this context.
Note that we already have `^superscript^` and I think `^^smallcaps^^` would be nice, the 'uparrows' being iconic for 'kind of uppercase'.
(Also it so happens that since they are 'accents' all of `` ~ ` ^ `` are shift/altgr deadkeys on my Swedish keyboard, so I must hold down shift/altgr and hit one of those keys six times to get three of those characters For some odd reason hitting the same key more than three or four times in a row is annoying. :-)  So I for one would be very grateful for `:::` instead of `^^^`!  I would be even more grateful if you implement `[span text]{ATTRIBUTES}` as well (and `^^smallcaps^^` if I dare ask! :-).

ousia:
@mpickering, I agree with both proposals from @bpj: `[span]{.class #identifier lang="grc"}` and `^^small caps^^`.

In that case, a more consistent syntax for fenced blocks would be:

```
[[[ {.warning #fenced-block-1}
This is a fenced div block.

With multiple pagragraphs.
]]]
```

Using different starting and closing characters makes nesting easier for the end user (so that indentation isn’t mandatory for nesting).

tarleb:
Superseded by fda0c0119f415c6df95b20730650388c0471241d.

Issue #1812, opened 2014-12-14
Org reader handles links poorly


tarleb:
The org format allows for many different types of [external link types](http://orgmode.org/manual/External-links.html). Many of these are not supported by the org reader.

See also #756.

tarleb:
The current link parsing code is a rather ugly piece of code. I'm rewriting it completely.  Progress is slow as free time is a bit scarce right now.

marcbowes:
In HTML, `/` means "relative from the server root". In mediawiki, `/` means "a subpage of the current path". What are you plans to handle this sort of conflict?

If you're only exporting for a single target (e.g. org to mediawiki) then it doesn't really matter because you can just author the source in a way that works in the target. But if you're exporting to both mediawiki and HTML then one of them is going to be broken.

Personally, I would prefer that the semantics of the source form were preserved. That is, the syntax for all of the relative links that org supports should be translated to the targets' syntax:

* subpage of the current path, e.g. if you are at `/foo` (HTML) then you need to be able to link to `/foo/bar`
* sibling of the current path, e.g. `/foo` -> `/bar`
* ancestor of the current path, e.g. `/foo/bar` -> `/foo` or `/baz`
* relative from the root, e.g. `/foo/bar/baz` -> `/peanuts/and/bananas`

I'm not exactly sure how you're meant to understand "root" in pandoc sites, so I'm happy to work around that with postprocessors if it's difficult..

jgm:
@tarleb is this still an issue?

tarleb:
It's not optimal yet, but link handling has improved considerably since my last comment.  I believe the remaining issues mentioned above are solvable via pandoc filters and not a reader problem, so this can be closed.

Issue #756, opened 2013-02-19
Internal Links in MediaWiki? (pandoc 1.10.1)


Milofax:
I try to create internal links in mediawiki that should point in the end to another wiki page. Therefore, I have different md-files that represent a wiki page each. 
For wiki, I convert each md-file separately to mediawiki. I also use the pdf conversion which works fine (and where I convert the content at once).
I would like to have the behaviour that internal links as [OtherWikiPage] would be converted to Mediak Wiki like [[OtherWikiPage]]. At the moment, pandoc doesn't interpret it at all.
Is that possible?

Mathias

PR #1813, opened 2014-12-14
Org reader: properly handle links to `file:target`

tarleb:
Org links like `[[file:target][title]]` were not handled correctly,
parsing the link target verbatim.  The org reader is changed such that
the leading `file:` is dropped from the link target.

This is related to issues #756 and #1812.


tarleb:
No, that's an unintended result.  Sorry if that causes inconveniences. I'm on it.

PR #1856, opened 2015-01-05
Allow compilation with ghc 7.10.1 RC1 and haddock-library 1.2

markwright:
Compiles with ghc 7.10.1 RC1 and ghc 7.8.4.


jgm:
Many thanks for doing this!

PR #1864, opened 2015-01-09
provide valid HTML in writer.html

ousia:
@jgm, there were some minor issues that made the contents from writer.html invalid XML. Here you have them fixed.

There seems to be also an issue with spaces, which I don’t know whether this is intended or not.


mpickering:
@ousia These changes will need to be made to the HTML writer, not the test file which it outputs. 

Could you please describe the invalid XML which we generate in a new ticket?

PR #1874, opened 2015-01-12
correct raw XHTML in tests/writer.markdown (closes #1869)

### LINKED ISSUE #1869: pandoc/tests/writer.html contains invalid XML
@mpickering, as warned in #1864 pandoc/tests/writer.html contains invalid XML.

There are two issues:

1. Identifiers must be unique (three elemens contain `id="bar"`).

2. Elements should be closed or self-closed (`<hr/>` instead of `<hr>`).

I hope it helps.

ousia:
@jgm, just in case it helps I corrected the raw XHTML (it would close #1869).


jgm:
I'm going to leave the writer test as is.  Since the writer does allow invalid HTML to be passed through unchanged, it is important not to suggest otherwise in the tests.

PR #1888, opened 2015-01-19
Better warning when trying to read multiple binary files

mpickering:


jgm:
By the way, can you figure out why the travis build is failing?  I see
```

src/Text/Pandoc/Writers/HTML.hs:52:1:
    Warning: The import of `Text.Blaze.Internal' is redundant
               except perhaps to import instances from `Text.Blaze.Internal'
             To import instances alone, use: import Text.Blaze.Internal()
<no location info>: 
Failing due to -Werror.
```
But we didn't used to get this warning.  I infer that `Text.Blaze.Html` now exposes `preEscapedString` (which is the function we import from `Text.Blaze.Internal`) but didn't used to.  However, looking on Hackage anyway, it doesn't seem that `Text.Blaze.Html` *does* export `preEscapedString`.  So why in the world do we get that warning?

I don't want to just remove the import, because it probably is needed for older versions of blaze-html, even if it isn't needed for the most current ones.

mpickering:
Not sure.. [this build](http://hydra.nixos.org/build/18780865/log/raw) succeeds without the warning as well.

mpickering:
[This](https://github.com/jaspervdj/blaze-markup/commit/14312c7c8d3ab585d459144e638069177ca4f105) is the commit causing the problems.

jgm:
Thanks for finding.  So, we just need to wrap the input in a
CPP macro.  I'll take care of that now.

+++ Matthew Pickering [Jan 19 15 02:07 ]:

PR #1901, opened 2015-01-26
Wrap code block HTML table in a div element

sapek:
Syntax highlighted code blocks are written out as HTML tables which are
impossible to format via styles for different screen widths. This means
that pandoc generated pages with code blocks must have a fixed minimal
width, making them hard to read on narrower screens.

Wrapping the table in a `<div>` allows formating a code block that will
have a horizontal scroll bar if the content doesn't fit the screen
width, by adding the following style:

    div.sourceCode {
        overflow-x: auto;
    }

This enables pandoc to generate responsive HTML pages.


sapek:
Here's an example of a page generated with the modified pandoc: https://microsoft.github.io/bond/manual/bond_cpp.html. When displayed on small screen (or narrow browser window) some of the code blocks will have horizontal scroll bar while the regular text will adjust to browser window width.

sapek:
I will send another PR with updated tests.

PR #1915, opened 2015-02-02
Update travis file (for GHC 7.10)

mpickering:


jgm:
But pandoc doesn't yet build on ghc 7.10, does it?
(I think we're still waiting for haddock-library, and possibly other
things.)

+++ Matthew Pickering [Feb 02 15 02:50 ]:

mpickering:
No not yet, this is just for when pandoc does build with ghc 7.10.

PR #1917, opened 2015-02-02
Remove F monad from Org reader and Markdown reader

mpickering:


mpickering:
Please don't merge this yet, there is a problem combining this approach and better error handling.

jgm:
Okay.  Maybe we should close this, then; if at some future point you see how to do it, you could open a new PR?

PR #1921, opened 2015-02-03
Update wikitext syntax

freephile:
for media in the "File" namespace.  
[[Image:]] is deprecated.  
Replaced with [[File:]]


freephile:
I do not believe there are any other changes necessary to support this update in syntax.  There are mentions of the word 'Image' in the source such as: 
`inlineToMediaWiki (Image alt (source, tit)) = do` 
but I believe those are fine.  I'll check on the reader side.

mpickering:
You need to update the tests. How long has it been deprecated for? I am a bit worried that a lot of people will still be needed old versions which require Image rather than File.

freephile:
ignore.  I created a new pull request with the test updated.

PR #1927, opened 2015-02-05
update syntax for Images/Media files in MediaWiki

freephile:
The preferred syntax for Images and other media is [[File:Foo.jpg]] in MediaWiki since v1.14 (6 years).  [[Image:Foo.jpg]] is deprecated but still works as an alias to the File namespace.  I don't think this would break any existing wikis since talk of switching the syntax/namespace for images started back in 2002 (https://phabricator.wikimedia.org/T2044).  NS_FILE became the new namespace for Files in v 1.14 in late 2008; released Feb 2009. (https://www.mediawiki.org/wiki/Release_notes/1.14)  There is still a namespace alias so '[[Image:]]' still works today.  It's just that MediaWiki supports other media as well, and so the name and syntax used in documentation (see https://www.mediawiki.org/wiki/Help:Images) has long been '[[File:foo.jpg]]'

This updates both the writer, and the test.  The reader already uses the "[[File: ]]" syntax.

### DIFF
--- a/src/Text/Pandoc/Writers/MediaWiki.hs
+++ b/src/Text/Pandoc/Writers/MediaWiki.hs
@@ -107,7 +107,7 @@ blockToMediaWiki (Para [Image txt (src,'f':'i':'g':':':tit)]) = do
   let opt = if null txt
                then ""
                else "|alt=" ++ if null tit then capt else tit ++ capt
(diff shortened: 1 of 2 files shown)


jgm:
Thanks!

PR #1947, opened 2015-02-16
Remove F Monad from Markdown and Org readers.

mpickering:
@tarleb 

Note that these two modules now contain circular definitions. Normal editing (not touching the footnote handling) should be fine but there are some points where care has to be exercised to avoid loops.

One particular point is the different semantics of

```
if foo 
  then
    return True
  else
    return False
```

and

```
return $ if foo then True else False
```

The second is lazier than the first as `foo` must be evaluated in the first instance but not in the second.


jgm:
+++ Matthew Pickering [Feb 16 15 15:25 ]:
>   Note that these two modules now contain circular definitions. Normal

Can you elaborate?

mpickering:
Here is an example which demonstrates how the reader now resolve footnotes. 

```
module Example where

import Text.Parsec
import Control.Monad.Reader

test1 :: ParsecT String [(Char, String)] (Reader [(Char, String)]) String
test1 = do
    a <- anyChar
    notes <- ask
    -- It is important that notes is not forced
    return $ case lookup a notes of
        Just v -> v
        Nothing -> ""

test1a :: (Char, String) -> ParsecT String [(Char, String)] (Reader [(Char, String)]) ()
test1a v = 
    updateState (v:)

prog :: ParsecT String [(Char, String)] (Reader [(Char, String)]) String
prog = do
    res <- test1
    res2 <- test1
    test1a ('a',"1")
    test1a ('b',"2")
    return (res ++ res2)

main :: IO ()
main = print runTest2

retState :: Monad m => ParsecT s st m a -> ParsecT s st m (a, st)
retState p = do
    res <- p
    st <- getState
    return (res, st)


runTest2 :: (String, [(Char, String)])
runTest2 = (res, n)
    where
        m  = runParserT (retState prog) [] "" "abc"
        -- Notice how this is circular
        (res, n) = either undefined id $ runReader m n
```

jgm:
Can you say a bit more about how that final, mind-bending line works?

Also: have you benchmarked the new code against the old (F Monad)
version?  I know that monad transformer stacks can sometimes be
a performance problem.

+++ Matthew Pickering [Feb 18 15 04:45 ]:

mpickering:
By default Haskell's evaluation strategy is lazy, that is, values are not computed before they are required. Notice that when we parse a footnote marker that we will definitely replace it with something (either a `Note` or `Str "[1^]"`, the only problem is that we don't know which until we reach the bottom of the file. This solution makes sure evaluation is delayed long enough so that by the time we force evaluation, we have all the necessary information to fill in the gaps.

The final line looks a bit mystical but there is no cycle as the state is not modified as a result of any reader actions. In fact, a very common pattern is the following.

```
st <- ask
return $ ...
```

Which ensures that the value of `st` is not forced. I'm not sure if this is a very good explanation, please ask more questions if anything is unclear.

Benchmarks

old
```
benchmarking markdown reader
mean: 55.39495 ms, lb 53.88569 ms, ub 57.09139 ms, ci 0.950
std dev: 3.738574 ms, lb 3.045484 ms, ub 4.593733 ms, ci 0.950
variance introduced by outliers: 28.456%
variance is moderately inflated by outliers
```

new
```
benchmarking markdown reader
mean: 70.94543 ms, lb 69.42924 ms, ub 73.30225 ms, ci 0.950
std dev: 4.368495 ms, lb 2.901871 ms, ub 6.838932 ms, ci 0.950
found 1 outliers among 20 samples (5.0%)
  1 (5.0%) high mild
variance introduced by outliers: 23.712%
variance is moderately inflated by outliers
```

jgm:
So there seem to be two disadvantages of this change:

- We have to be careful about laziness or things can blow up
- Performance is worse

What would you say are the advantages?  How should we weigh them against the disadvantages?

mpickering:
Here are a few reasons.

* Reduces cognitive overhead. I always found it difficult to work out how many times to `return` when making changes to the markdown reader. You often reach for functions such as `trimInlines` but then realise you need to use `trimInlinesF` which just disrupts the normal writing process. 
* Improves code reusability (for example [this](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Readers/Markdown.hs#L1931) can be reused from `Parsing`)
* Likely to reduce bugs (due to things like [this](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Readers/Markdown.hs#L494) and general clarity)
* Reduces code duplication with functions defined for `F` and normally.

As for the disadvantages.

1. You only have to be careful about laziness if you are making changes to the part of the program which relies on laziness. (Which is a small part!) This is in contrast to currently where you always have to think about `F` even in parts which don't concern it.
2. I'm not too concerned about performance implications. I can look at profiling the reader to find other places to improve if you feel the hit is unacceptable.

jgm:
OK, that's pretty convincing.

jgm:
@mpickering, this PR can't be automatically merged due to subsequent changes in the Org and Markdown readers.  Since you understand this code better than I, could you rebase your changes on top of current HEAD?

mpickering:
Rebased, just waiting for travis now.

PR #1948, opened 2015-02-16
Remove "error" landmines from readers

mpickering:
This is rebased on the f monad removal changes. I can rebase it onto HEAD if that is preferred and then submit another pr to get the changes to play nicely. 

You should definitely have a scan before merging. I am a bit unsure about 1a78a3f and a03e0f3.

TODO
- [x]  Add copyright notice
- [x] Fix interface to Except compat module
- [x]  Separate out commits


jgm:
+++ Matthew Pickering [Feb 16 15 15:34 ]:
>   This is rebased on the f monad removal changes. I can rebase it onto
>   HEAD if that is preferred and then submit another pr to get the changes
>   to play nicely.

If I recall, the F Monad removal proposal had some problems,
and you wanted to hold off on that.  Have you got that all
figured out now?  If not, rebasing onto HEAD would be
preferable.

mpickering:
I think I managed to resolve the problems.

jgm:
Thanks for all of this.  It will be a significant breaking API change, but I think it's worth it for the safety.

PR #1967, opened 2015-02-21
Fixes #661: --smart must also convert << to « and >> to »

### LINKED ISSUE #661: --smart must also convert << to « and >> to »
Subj.

sumitsahrawat:


mpickering:
Hi, thanks for the patch. Could you please add a small test as well [here](https://github.com/jgm/pandoc/blob/master/tests/Tests/Readers/Markdown.hs).

mpickering:
LGTM

sumitsahrawat:
@mpickering I've added a test :+1: 

While writing a test, I realized that it didn't work for markdown. Markdown, and some other readers were using `[apostrophe, dash, ellipses]` directly, so I changed them to use a list `smartPunctuationParsers` defined in `Text.Pandoc.Parsing`.

mpickering:
I think that we will be able to change them back to using `smartPunctuation` with #1947.

sumitsahrawat:
So, should I revert those changes, or is everything alright?

mpickering:
No this is fine -- more of a note to myself. I will let @jgm do the merge when he is happy.

sumitsahrawat:
Another issue I found is that converting to pdf doesn't convert LaTeX to \LaTeX\
I don't want to send another PR as that one will probably cause conflicts with this one.
So, should I stack those changes upon this PR?

timtylin:
> Another issue I found is that converting to pdf doesn't convert LaTeX to \LaTeX\
I don't want to send another PR as that one will probably cause conflicts with this one.
So, should I stack those changes upon this PR?

This probably can't be done through the `--smart` mechanism since it affect readers only, so there's no way to target just latex output. It's trivially done through a filter though.

sumitsahrawat:
@jgm I've fixed the merge conflicts.

tarleb:
I'm trying to revive the issue, so here are some thoughts:

`«` and `»` together are a kind of double quote, so it might make sense to treat them as such.  I would argue that it might be cleaner to alter the `singleQuoted` and `doubleQuoted` functions in `Text.Pandoc.Parsing`.  This way, the semantics of these characters are handled correctly.

It should be noted that these kind of quotes are used differently across languages. E.g. Germans would write »Quoted« while French would put it as «Quoted».  See [Wikipedia on quotation marks](https://en.wikipedia.org/wiki/Quotation_mark) for a list of different styles.

Issue #1933, opened 2015-02-10
Title/Date styles are lost when writing docx


nenono:
I tried to export Word docx with custom `reference.docx` modified by Microsoft Word 2013. I found that styles of Title and Date are lost in the output. It seems a bug.

### steps to reproduce
I created a markdown text like this. (test.md)
```markdown
% Some Document Title
% Some Document Author
% 2015-02-10

body
```

and I created reference.docx by the following steps.

1. execute command : `pandoc --print-default-data-file reference.docx > reference.docx`
2. open reference.docx by MSWord.
3. overwrite(by `save as` menu) and close the file.

then I exported docx by the command that : `pandoc test.md --reference-docx=reference.docx -o test.docx`.

### actual results
open the created test.docx file in MSWord, I found the following.
- `Some Document Title` is the Normal style.
- `Some Document Author` is the Author style.
- `2015-02-10` is the Normal style.

### expected results
using original version of reference.docx(created at the step 1), I verified the following.
- `Some Document Title` is the Title style
- `Some Document Author` is the Author style.
- `2015-02-10` is the Date style.

### environment
- Pandoc 1.13.2
- Microsoft Word 2013 (32bit, Microsoft Office Proffesional Plus 2013)
- Windows 7 (64bit)

nkalvi:
It works as expected when I tested it in a slightly different setting: MS Word 2011 on Mac (same Pandoc). The styles were present in both the ref and the test document, and you can see the change in title style (changed in the ref doc). Also I can see that Pandoc is writing the fields with expected styles in the code.

![pandoc-docx test](https://cloud.githubusercontent.com/assets/2199312/6123138/6a78a92e-b0cc-11e4-884b-83ac1e407aee.png)

nkalvi:
Works as expected under Win 7 (32bit), MS Word 2013 (as yours) with 1.13.2.

jgm:
Could you post or link to the `reference.docx` you're using?
Unfortunately I don't have MS Word 2013 to test with.

Also, can you clarify:  when you created the reference.docx, did
you edit it at all, or did you just open and save the default
reference.docx using MS Word 2013?

nkalvi:
I have tested it a couple of times (with modifying the ref.docx) under Win7/MS Office 2013 and I cannot replicate the problem.
I’ll be more than happy to take a look at the ref.docx used once we get it. I have to MS Office 2011, Office 2013 and Office 2014.

nenono:
[...]

Issue #1872, opened 2015-01-12
SourceCode style and *.docx output


dymanic:
The "Source Code" style in docx output does not work properly when edited and used as a reference style.

Input file `test.markdown`:

    ~~~{.haskell}
    putStrLn "hello"
    ~~~

Command:

    pandoc -f markdown -t docx -o test.docx test.markdown

Expected output:

The output file `test.docx` should contain a single "Source Code" style which can be edited, so that it can be used with `--reference-docx` and applied to subsequent docx output.

Actual output:

The output docx contains two copies of the "Source Code" style. 

This seems to cause further problems using the output as a reference docx. I.e.:

1. Create test.docx as detailed above.
2. Open test.docx in MS Word and select the whole line of source code.
3. Alt-Ctrl-Shift-S to open the Styles box. You will notice *two* styles named "Source Code", one which is selected and one which isn't.
4. Right click on the selected "Source Code" style (i.e. the one applied to the selected line).
5. Make some modifications to the style. [Note: you will notice if you try to make changes to the other "Source Code" style, you will receive the message "This style name already exists or is reserved for a built-in style" after clicking Modify -> Format]
6. Save the document.
7. Run the following to use test.docx as a style reference for another conversion:

        pandoc -f markdown -t docx -o test2.docx --reference-docx test.docx test.markdown

8. You will see that the changes made to the Source Code style in test.docx have *not* been applied in test2.docx

Versions:

pandoc 1.13.2 Windows 7
MS Word 2013

mpickering:
Related: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/pandoc-discuss/4YkX-B-3Umk/9ioO4TwnTHwJ

dymanic:
Thanks Matthew.

Yes, I can edit the reference test.docx so it only contains the Source Code style that I want (either by deleting the wrong Source Code style in Word or by editing styles.xml directly).

However, any further conversions using that as a `--reference-docx` *always* add a new Source Code style to the output, and then *use that new style*.

So, I end up with an extra "Select all source code blocks and change to the *other* Source Code style" step in my conversion process.

jgm:
The docx writer generates the SourceCode style based on the highlighting
style that was selected with `--highlight-style`.  The generated version
replaces whatever wa in the styles.xml file in the reference.docx. So,
I have removed the SourceCode style from the reference.docx.

This addresses a bug in pandoc, but it may not solve your problem,
[...]

PR #1968, opened 2015-02-21
Fixes for multiple docx writer style bugs.

### LINKED ISSUE #1607: Docx writer: common class names don't work in different languages.
"Header1" becomes "Overskrift 1", etc. We can solve this by parsing the styles file and applying styles with the correct "name" tags.

lierdakil:
:construction: :warning: **This  PR is work-in-progress. Not ready to merge** :warning: :construction: 

This is intended to fix #1607, #1933 and #1872 

This PR builds a style name → style id mapping based on styles.xml from reference doc, and changes pStyle and rStyle to accept style name as a parameter instead of styleId. There is a fallback mechanic that removes spaces from style name and returns it as style id, but it likely won't help much if there's no relevant style in styles.xml.

It also changes `styleToOpenXML` to skip generation of styles already present in styles.xml.

Style names are matched lower-case, since headings and `footnote text` have lowercase names, and Word doesn't seem to care.

Comments are welcome.

TODO:
- [x] Skip lookup for custom styles, since those are always well-defined
- [x] More efficient implementation? There is a little too much of IO monad for my taste
- [ ] Test cases (not sure on how to test docx writer, but oh well)
- [x] if --no-highlight is explicitly defined, remove *Tok styles.
- [x] Update reference.docx
- [x] Move StyleMap into separate module in Reader
- [ ] Update Reader to work with StyleMap


nkalvi:
Hi @lierdakil, just to understand (I don't know Haskell yet) what you're doing, are you 

1. Reading the styles.xml from the reference docx
2. Get the localized names for Pandoc's styles used
3. Use these localized style names in the output doc

#### Example illustrating the issue when using Norwegian Word 2013 

Part of source:
```
% Title
  spanning multiple lines
% Author One
  Author Two; Author Three;
  Author Four
 ```

Partial styles.xml (with *Tittel* vs *Title*):
```
  <w:style w:type="paragraph" w:styleId="Tittel">
    <w:name w:val="Title" />
    <w:basedOn w:val="Normal" />
    <w:next w:val="Normal" />
    <w:qFormat />
    <w:rsid w:val="00DE710F" />
    <w:pPr>
      <w:keepNext />
      <w:keepLines />
      <w:spacing w:before="480" w:after="240" />
      <w:jc w:val="center" />
    </w:pPr>
    <w:rPr>
      <w:rFonts w:asciiTheme="majorHAnsi" w:eastAsiaTheme="majorEastAsia" w:hAnsiTheme="majorHAnsi" w:cstheme="majorBidi" />
      <w:b />
      <w:bCs />
      <w:color w:val="C0504D" w:themeColor="accent2" />
      <w:sz w:val="36" />
      <w:szCs w:val="36" />
    </w:rPr>
  </w:style>
  <w:style w:type="paragraph" w:styleId="Undertittel">
    <w:name w:val="Subtitle" />
    <w:basedOn w:val="Tittel" />
    <w:next w:val="Normal" />
    <w:qFormat />
    <w:pPr>
      <w:spacing w:before="240" />
    </w:pPr>
    <w:rPr>
      <w:sz w:val="30" />
      <w:szCs w:val="30" />
    </w:rPr>
  </w:style>
  <w:style w:type="paragraph" w:customStyle="1" w:styleId="Author">
    <w:name w:val="Author" />
    <w:next w:val="Normal" />
    <w:qFormat />
    <w:rsid w:val="00DE710F" />
    <w:pPr>
      <w:keepNext />
      <w:keepLines />
      <w:jc w:val="center" />
    </w:pPr>
    <w:rPr>
      <w:color w:val="548DD4" w:themeColor="text2" w:themeTint="99" />
    </w:rPr>
  </w:style>
```

Portion from the produced docx using the above as reference docx:
```
 <w:body>
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Title" />
      </w:pPr>
      <w:r>
        <w:t xml:space="preserve">Title</w:t>
      </w:r>
      <w:r>
        <w:t xml:space="preserve"></w:t>
      </w:r>
      <w:r>
        <w:t xml:space="preserve">spanning</w:t>
      </w:r>
      <w:r>
        <w:t xml:space="preserve"></w:t>
      </w:r>
      <w:r>
        <w:t xml:space="preserve">multiple</w:t>
      </w:r>
      <w:r>
        <w:t xml:space="preserve"></w:t>
      </w:r>
      <w:r>
        <w:t xml:space="preserve">lines</w:t>
      </w:r>
    </w:p>
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Author" />
      </w:pPr>
      <w:r>
        <w:t xml:space="preserve">Author</w:t>
      </w:r>
```

**Comment by lierdakil** [comment]
@nkalvi 
Notice how there is child element `w:name` for each style with `w:val` set to english equivalent.
```xml
  <w:style w:type="paragraph" w:styleId="Tittel">
    <w:name w:val="Title" /> <!-- here -->
    ...
  </w:style>
  <w:style w:type="paragraph" w:styleId="Undertittel">
    <w:name w:val="Subtitle" /> <!-- and here -->
    ...
  </w:style>
  <w:style w:type="paragraph" w:customStyle="1" w:styleId="Author">
    <w:name w:val="Author" /> <!-- and even here -->
    ...
  </w:style>
```
This is quite universal. I parse all styles from reference doc's `styles.xml` and build a map from `name.val` to `style.styleId`. For a given example it would look something like this (in sudo js, to spare you haskellisms):
```javascript
styleMap = {
  "title": "Tittel",
  "subtitle": "Undertittel",
  "author": "Author"
}
```
Note that keys are lowercase. This is because Word doesn't care for `name` case, so neither do I.

Now I can reference styles by their english `name`. When creating, e.g. a paragraph with a `Title` style, I do something like this (again, in sudo js):
```javascript
paraNode = new ParaNode({"style": styleMap[lowercase("Title")], ...})
```
This will produce something like
```xml
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Tittel" />
      </w:pPr>
      ...
    </w:p>
```

In practice, there are two style `type`s: paragraph and character. They can be applied to different sets of elements, so I actually create *two* maps (which are stored as `stParaStyles` and `stCharStyles` in writer state). The rest is support code and changes from style ID to style name

Headings are a bit of special case. To make long story short, I construct heading `name` as 
```javascript
"Heading "+level
```
where `level` is integer heading level (`level=1,2,3...`)

That's what I'm doing in a nutshell.

nkalvi:
Thank you very much for kindly explaining this! This solution will be very useful for many international Pandoc users and others (including myself :)

lierdakil:
After some investigation about Word styles, here is what I found. This is a list of all styles, used by docx writer or defined in default reference.docx, except token styles.

| Name                   | Id                   | custom | word | type | alt                              |
|:-----------------------|:---------------------|:------:|:----:|:----:|:---------------------------------|
| Title                  | Title                |        |  +   |  p   |                                  |
| Subtitle               | Subtitle             |        |  +   |  p   |                                  |
| Author                 | Author               |   ?    |      |  p   | ?                                |
| Date                   | Date                 |        |  +   |  p   |                                  |
| Abstract               | Abstract             |   ?    |      |  p   | ?                                |
| Bibliography           | Bibliography         |        |  +   |  p   |                                  |
| Heading 1              | Heading1             |        |  +   |  p   |                                  |
| Heading 2              | Heading2             |        |  +   |  p   |                                  |
| Heading 3              | Heading3             |        |  +   |  p   |                                  |
| Heading 4              | Heading4             |        |  +   |  p   |                                  |
| Heading 5              | Heading5             |        |  +   |  p   |                                  |
| Compact                | Compact              |   ?    |      |  p   | ?                                |
| Image Caption          | ImageCaption         |        |      |  p   | caption                          |
| First Paragraph        | FirstParagraph       |        | ???  |  p   |                                  |
| Body Text              | BodyText             |        |  +   |  p   |                                  |
| Block Quote            | BlockQuote           |        |      |  p   | Intense Quote, Block Text, Quote |
| Source Code            | SourceCode           |   +    |      |  p   |                                  |
| Table Caption          | TableCaption         |        |      |  p   | caption                          |
| Definition Term        | DefinitionTerm       |   ?    |      |  p   | ?                                |
| Definition             | Definition           |   ?    |      |  p   | ?                                |
| Verbatim Char          | VerbatimChar         |   +    |      |  c   |                                  |
| Footnote Ref           | FootnoteRef          |   +    |      |  c   | footnote reference               |
| Footnote Text          | FootnoteText         |        |  +   |  p   |                                  |
| Link                   | Link                 |   +    |      |  c   | Hyperlink                        |
| Normal                 | Normal               |        |  d   |  p   |                                  |
| Default Paragraph Font | DefaultParagraphFont |        |  d   |  c   |                                  |
| Normal Table           | TableNormal          |        |  d   |  t   |                                  |
| Body Text Char         | BodyTextChar         |   +    |      |  c   |                                  |

Custom column is whether style is set as custom in default reference.docx. '?' means it probably should be, but isn't.

Word column is whether Word 'understands' a style. 'd' means it's default style, '???' means I couldn't test.

Type is style type, 'p' is for paragraph, 'c' is for character, 't' is for table.

Alt is Word's built-in styles that serve the same purpose as pandoc-defined ones. Maybe we should switch to those. '?' means that I feel like there probably should be one, but couldn't find it.

As you can see, there are a few ambiguous styles, that are neither defined as custom, nor recognized by Word. We need to decide what to do with those.

There are also 'Link' and 'Footnote Ref' styles that are essentially the same as Word-defined 'Hyperlink' and 'footnote reference', and could be possibly dropped.

Here is a list of styles I'm most concerned about:

| Name            | ambiguous | type | can be replaced by               |
|:----------------|:---------:|:----:|:---------------------------------|
| Author          |    yes    |  p   | ?                                |
| Abstract        |    yes    |  p   | ?                                |
| Compact         |    yes    |  p   | ?                                |
| Image Caption   |    yes    |  p   | caption?                          |
| Block Quote     |    yes    |  p   | Intense Quote, Block Text, Quote |
| Table Caption   |    yes    |  p   | caption?                         |
| Definition Term |    yes    |  p   | ?                                |
| Definition      |    yes    |  p   | ?                                |
| Link            |    no     |  c   | Hyperlink                        |
| Footnote Ref    |    no     |  c   | footnote reference               |

these are either ambiguous (again, meaning that they are neither custom, nor Word defaults), duplicate default Word styles, or both.

So the question is, what should be done with them? I've got the following options:

* Update reference.docx to define ambiguous styles as custom. Do nothing more. This is a quick-and-dirty solution.
* Replace styles in reference with word-recognized ones to the best of our ability. This is a longer path, but leading to hopefully better user experience.
* Do nothing at all, and hope for the best. This will likely lead to 'surprises' in the future, and not the pleasant kind, I fear.

First two options require updates to reference.docx though. @jgm, @jkr, @mpickering, what do you think?

nkalvi:
@lierdakil Good job! :)

Using 'Block Text' for 'Block Quote' may be better since the other styles are not available in Word 2003; that is if the style's name is replaced. 

I feel that taking care of the ones with '+' in the 'word' column would satisfy most users.

lierdakil:
@nkalvi, it will be backwards-compatible if we actually define said styles in reference doc (and if not, it won't work as expected anyway). It just means that word 2003 won't recognize these styles semantically, while later versions would.

I agree that 'Block Text' seems like a good choice though, since it's only one that's left-aligned by default. 'Intense Quote' and 'Quote' are center-aligned.

The question is not really about those styles that Word recognizes. They're pretty much handled fine already. I just feel that if I'm going to do this anyway, I should do it properly this time. Previous attempt addressed only part of the problem, so now I want to make sure I won't have to come back to this.

nkalvi:
@lierdakil @jkr @jgm @mpickering 

This is certainly not efficient and perhaps not as elegant, but I'm wondering about this:

1. Read localized/preferred style names in 'front-matter' either in the document or as an option (f.ex. --stylenames=styles-norsk.yaml). This is regardless of specifying reference.docx. Set a flag if mapping is done. 

		styleName:
			Title: Tittel
			Subtitle: Undertittel

2. If style names are not specified, but a reference.docx is specified, see whether mapping is needed and set a flag.
4. Then in (docx writer's) **renderXML** function, map the names based on the flag.

BTW, my preferred way is to use variables for style/base names and map them from default if needed.
I'll take the feedback as learning experience :)

lierdakil:
@nkalvi I think this actually should be a separate feature, not a substitute for a bugfix. This would allow users to use custom styles in reference docx, which could be nice, but it does not fix issues at hand (although provides a workaround). I could work on it after I'm done with this one. Would you care to create a topic for this proposal in pandoc-discuss?

mpickering:
I've not been following closely but it looks like you are working towards a more robust solution. 

Could you perhaps distil what's in this ticket into a new page on the wiki so that the behaviour is properly documented somewhere?

lierdakil:
@mpickering If you're talking to me, and not @nkalvi, then I'm a little lost as to what should I document. This PR keeps the same workflow, but fixes some issues (referenced in first post in this PR). @nkalvi's proposal, which got mixed in in discussion, is another matter, but I'm not implementing it right now.

And while we're at it, what are your thoughts on updating default reference.docx to correspond with Word more closely? I really can't move this forward until we decide on that...

nkalvi:
@lierdakil @mpickering 
Here's the feature request:
https://groups.google.com/forum/#!topic/pandoc-discuss/oLhedoPmMIs

mpickering:
I replied to both posts on the mailing list. To summarise I think that @lierdakil's solution will work well.

There is already some support for international handling in the docx reader, will this work surpass that?

mpickering:
PS. I think it would be good to move the reading of the `styles.xml` into a separate module. (Probably in the `src/Text/Pandoc/Readers/Docx` folder).

jgm:
@lierdakil - thanks for your detailed analysis. Marking styles as custom is certainly a good idea.  As for the more radical change of switching to standard Word styles, I'm positively inclined, but (as I noted on pandoc-discuss) I did have an earlier difficulty that led me to use "Footnote Ref" and "Link" instead of the Word styles.  I never did really understand what was going on with this, but I wanted to note that the issue had come up.

lierdakil:
@mpickering
As for Reader, kind of. Reader is slightly more complicated, because it needs to get information about base styles as well. I was thinking about merging this codes, but ultimately decided against it, since intent and handling is very different. I will, however, try to implement similar method in reader later.
Moving reading code into separate module is a good idea, but probably not Reader, for reasons explained above. I will think about this idea though, and may come up with a solution. Thank you.

@jgm
I already answered in pandoc-discuss, but for posterity's sake, answering here as well. I'll keep this in mind, and maybe try to fix underlying issues if I stumble upon them. Thank you for bringing this to my attention.

lierdakil:
So I've been thinking hard about tests for docx writer, and here is my best idea so far.

Ideally, docx reader should read writer's output. So, to avoid fumbling around with docx validation, I could utilise existing reader tests to verify that it is indeed the case.

Testing procedure would go something like this:
1. Generate docx from native format
2. Read docx obtained in (1) into native format
3. Compare initial data against one obtained from (2)

This is good in that it could potentially highlight problems in both Reader and Writer. It's bad in that it does not test Writer independently, so "even errors" would fly under the radar, if those are not caught by reader tests.

At the very least, this is better than no tests at all, and alternatives are significantly more involved (basically boiling down to writing custom XML validator)

@jgm, what do you think?

As a side note, I couldn't reproduce issues with `Footnote Reference` and `Hyperlink` styles with current pandoc and Word 2014. I will test with other Word versions when I can get my hands on those (it will require some going around asking for favors, I fear). There is, however, a slight chance that said issues were fixed by something already.

jgm:
+++ Nikolay Yakimov [Mar 01 15 12:24 ]:
>So I've been thinking hard about tests for docx writer, and here is my best idea so far.
>
>Ideally, docx reader should read writer's output. So, to avoid fumbling around with docx validation, I could utilise existing reader tests to verify that it is indeed the case.
>
>Testing procedure would go something like this:
>1. Generate docx from native format
>2. Read docx obtained in (1) into native format
>3. Compare initial data against one obtained from (2)
>
>This is good in that it could potentially highlight problems in both Reader and Writer. It's bad in that it does not test Writer independently, so "even errors" would fly under the radar, if those are not caught by reader tests.
>
>At the very least, this is better than no tests at all, and alternatives are significantly more involved (basically boiling down to writing custom XML validator)
>
>@jgm, what do you think?

Tests that don't separate reader and writer bugs are certainly better than no tests at all!

Note that, in general, pandoc readers and writers don't successfully round-trip, for a variety of reasons.  So you'd have to look into whether the kind of round-trip tests you have in mind do work for docx.

lierdakil:
I added some rudimentary tests. @jgm was right, not everything survives writer-reader roundtrips, and although I'd prefer it would, I worked around those mostly.

Please note that build fail has nothing to do with tests failing or anything, but with some problem on Travis CI side, namely
```
WARNING: The following packages cannot be authenticated!
  cabal-install-1.18 ghc-7.8.4
E: There are problems and -y was used without --force-yes
The command "sudo apt-get install cabal-install-1.18 ghc-7.8.4" failed.
```
see https://travis-ci.org/jgm/pandoc/jobs/53507864

I will need to add tests for mangled reference.docx and update reader code a bit, but for the most part I think it's done, although test coverage could probably be better.

Requesting comments on code style etc.

jgm:
Great, I'm going to go ahead and merge this, before it gets stale.  Thanks!

lierdakil:
Okay, I'll create a new PR for reader changes and test updates then, when I find some time for it. Sorry I didn't complete this earlier, but I suddenly became insanely busy at work.

PR #1974, opened 2015-02-25
Org reader: Allow image links with non-image targets

rootzlevel:
Org-Mode's own html exporter converts the following org link:

    [[http://example.com][https://www.haskell.org/static/img/logo.png]]

to

    <a href="http://example.com">
    
    </a>

but pandoc generates:

    <a href="http://example.com">
    <a href="https://www.haskell.org/static/img/logo.png" class="uri">
     https://www.haskell.org/static/img/logo.png
    </a>
    </a>

which is useless. With this patch, it generates:

    <a href="http://example.com">
    
    </a>

### DIFF
--- a/src/Text/Pandoc/Readers/Org.hs
+++ b/src/Text/Pandoc/Readers/Org.hs
@@ -1106,7 +1106,7 @@ explicitOrImageLink = try $ do
   char ']'
   return $ do
     src <- srcF
-    if isImageFilename src && isImageFilename title
+    if isImageFilename title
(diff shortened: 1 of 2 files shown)


jgm:
@tarleb, does this look okay to you?

tarleb:
Looks good, thanks @rootzlevel!  Would you add a test case for this bug to `tests/Tests/Readers/Org.hs`? The link parsing code is in dire need of some serious refactoring, and I'd like to be sure that it's not reintroduced at a later time.

rootzlevel:
done

PR #1980, opened 2015-02-28
Bump blaze-html and blaze-markup upper version bounds

RyanGlScott:
Allow `pandoc` to build with the latest verions of `blaze-html` (0.8) and `blaze-markup` (0.7).


mpickering:
Thanks.

PR #1984, opened 2015-03-02
Drawer capability for the Org Writer

gnusosa:
For the implementation of the Drawer element in the Org Writer, we
  make use of a Generic Block container with attributes. Where the first
  element of the attributes defines that the Div constructor is a
  Drawer. The first element of the list in the attributes, defines the
  Drawer name to use. The key-value list in the attributes defines the
  keys to add inside the Drawer. Lastly, the list of Block elements
  contains miscellaneous blocks elements to add inside of the Drawer.

  If the Div element does not contain a Drawer name, it will default
  to the standard :PROPERTIES: tag. If given a name, it's automatically
  decorated to be a valid Org tag.

### DIFF
--- a/src/Text/Pandoc/Writers/Org.hs
+++ b/src/Text/Pandoc/Writers/Org.hs
@@ -107,6 +107,19 @@ escapeString = escapeStringUsing $
 blockToOrg :: Block         -- ^ Block element
            -> State WriterState Doc
 blockToOrg Null = return empty
+blockToOrg (Div ("Drawer",drawer,kvs) bs) = do
+  contents <- blockListToOrg bs
+  let drawerNameTag = if null drawer
+                      then text ":PROPERTIES:"
+                      else ":" <> text (head drawer) <> ":"
+  let keys = vcat $ map (\(k,v) ->
+                       ":" <> text k <> ":"
+                       <> space <> text v) kvs
+  let drawerEndTag = text ":END:"
+  return $ drawerNameTag $$ cr $$ keys $$
+           blankline $$ contents $$
+           blankline $$ drawerEndTag $$
+           blankline
 blockToOrg (Div attrs bs) = do
   contents <- blockListToOrg bs
   let startTag = tagWithAttrs "div" attrs


mpickering:
@tarleb

tarleb:
Looks good to me, although I'm not sure how many readers can produce this special kind of div.  We should also ask @punchagan, who wrote the org writer.

mpickering:
It should also be noted that I'm generally against undocumented conventions such as this. It is always possible that a user will have a class called `drawer` and get wrong results when converting to markdown and it isn't clear how users would produce a drawer from their source material.

gnusosa:
I agree, but usually most other templates don't have other types that can translate to the drawer type. Not that I am aware of. I could use another token instead of Drawer, Something very unique like `DrawerOrgType` to identify this. If the problem is documentation missing, I will gladly add it, but is there somewhere else aside from the README document to add documentation?

gnusosa:
The use case I'm looking for, is from native to Org-mode. Which works perfect. Let's say another module or package makes use of Pandoc to render an unstructured datatype. If there are other readers, that can provide any information that's worth to put in Drawer, let me know. I really don't know.

tarleb:
What I think would be a really nice additions is to store header attributes in drawers: Pandoc's `Header` can take attributes like ID and additional key value pairs. Those would be the perfect candidate to be placed in property drawers.  I share @mpickering's concerns about overly-specific conventions.  Anyway, this shouldn't be a problem: The code looks clean enough to be easily adaptable; rewriting it for use in headers is probably not too difficult.  Apologies for not warning you earlier, I should have notices this beforehand.

fommil:
any movement on this? It would be great to get `PROPERTIES` support in org-mode!

tarleb:
@gnusosa: I've cherry-picked and slightly adapted your commit to match the way the Org reader handles drawers.  See #2941.

fommil:
wow, org-mode support is sounding like it's becoming a fully operational battlestation!

tarleb:
Sam Halliday <notifications@github.com> writes:

> wow, org-mode support is sounding like it's becoming a fully operational battlestation!
>
> ---
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/jgm/pandoc/pull/1984#issuecomment-220919080

I'm still trying to get those exhaust pipes photon-torpedo proofed.

gnusosa:
Thanks for merging this. Thanks for the adaptation @tarleb. Sorry for the late reply. :+1:  :smile:

Issue #969, opened 2013-09-05
Add an option for -shell-escape


protz:
Hi,

Would it be difficult to add an option to call LaTeX with the `-shell-escape` flag when compiling to PDF? My use-case is custom minted blocks that I'm adding directly, and this package requires latex to be called with the flag. Granted, I could generate .tex and then compile it myself, but it would definitely be more convenient to have pandoc add the right flag :).

Thanks,

jonathan

jgm:
We now have `--latex-engine-opt`, so this can be closed.

bsamadi:
I get an empty space instead of highlighted LaTeX code with the following code. I have add the minted package in a template. To get the right output, I have to compile the generated TeX file with -shell-escape option before the file name. I was wondering if someone has an example of using -shell-escape that works.
---
title: "Code Chunk with Highlight"
author: "Behzad Samadi"
output:
  beamer_presentation:
    pandoc_args: [
    "--latex-engine-opt", "-shell-escape"
    ]
    keep_tex: true
    template: src/beamertemplate.tex
    fonttheme: structurebold
---

## Basic - Interfaces {.fragile}

- LaTeX

\begin{minted}{latex}
\bein{itemize}
\item Here
\end{itemize}
\end{minted}

jgm:
I don't know what version of pandoc you're using, but
there was a problem with --latex-engine-opt that was fixed
(#1779) in pandoc 1.15.0.5.

+++ Behzad Samadi [Nov 05 15 08:11 ]:

bsamadi:
I was using 1.15.0.6. and then I updated to 1.15.1.1. The issue was the same.

jgm:
I assume you're running pandoc via some kind of wrapper
script?  Because `pandoc-options` in YAML metadata is not
support in pandoc.

+++ Behzad Samadi [Nov 05 15 11:54 ]:

bsamadi:
Yes. I am using RStudio. When I use:
    pandoc_args: [
    "--latex-engine-opt", "-shell-escape"
    ]
The -shell-escape option is passed to the latex engine because it generates the PDF file with no error. If I remove the -shell-escape option, I receive an error.

jgm:
@bsamadi I'm confused about what problem you're encountering, if any.
When using minted, you do need the -shell-escape option, that's true.
And you say it works when you provide this.  So, all is well, no?

bsamadi:
[...]

Issue #1779, opened 2014-11-27
Pass flags/options to latex-engine


Dashed:
Something like: `--latex-engine-opts='-shell-escape'`

Useful for https://github.com/jgm/pandoc/issues/969.

squisher:
Seems like this should have been closed with #969, BUT --latex-engine-opt adds the option AFTER the input filename. pdflatex does not accept -shell-escape that way, it has to be before the input file.

Current pdflatex invocation:

    pdflatex "-halt-on-error" "-interaction" "nonstopmode" foo.tex "-shell-escape"

It would work if called like this:

    pdflatex "-halt-on-error" "-interaction" "nonstopmode" -shell-escape foo.tex

PR #1988, opened 2015-03-04
Fix issue #969, #1779 by providing --latex-engine-opt

sumitsahrawat:
A simple demonstration :sparkles:

screenshot from 2015-03-04 10 39 21

### DIFF
--- a/README
+++ b/README
@@ -655,6 +655,11 @@ Options affecting specific writers
     The default is `pdflatex`.  If the engine is not in your PATH,
     the full path of the engine may be specified here.
 
+`--latex-engine-opt=`*STRING*
+:   Use the given string as a command-line argument to the `latex-engine`.
+    If used multiple times, the arguments are provided with spaces between
+    them. Note that no check for duplicate options is done.
+
 Citation rendering
 ------------------
 
--- a/pandoc.hs
+++ b/pandoc.hs
@@ -198,6 +198,7 @@ data Opt = Opt
     , optCiteMethod        :: CiteMethod -- ^ Method to output cites
     , optListings          :: Bool       -- ^ Use listings package for code blocks
     , optLaTeXEngine       :: String     -- ^ Program to use for latex -> pdf
+    , optLaTeXEngineArgs   :: [String]   -- ^ Flags to pass to the latex-engine
     , optSlideLevel        :: Maybe Int  -- ^ Header level that creates slides
     , optSetextHeaders     :: Bool       -- ^ Use atx headers for markdown level 1-2
     , optAscii             :: Bool       -- ^ Use ascii characters only in html
@@ -259,6 +260,7 @@ defaultOpts = Opt
     , optCiteMethod            = Citeproc
     , optListings              = False
     , optLaTeXEngine           = "pdflatex"
+    , optLaTeXEngineArgs       = []
     , optSlideLevel            = Nothing
     , optSetextHeaders         = True
     , optAscii                 = False
@@ -734,6 +736,14 @@ options =
                   "PROGRAM")
                  "" -- "Name of latex program to use in generating PDF"
 
+    , Option "" ["latex-engine-opt"]
+                 (ReqArg
+                  (\arg opt -> do
+                      let oldArgs = optLaTeXEngineArgs opt
+                      return opt { optLaTeXEngineArgs = arg : oldArgs })
+                  "STRING")
+                 "" -- "Flags to pass to the LaTeX engine, all instances of this option are accumulated and used"
+
     , Option "" ["bibliography"]
                  (ReqArg
                   (\arg opt -> return opt{ optMetadata = addMetadata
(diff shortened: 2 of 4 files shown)


sumitsahrawat:
The `pdf` is highly zoomed btw.

mpickering:
Please can you add a description of the option to the README.

sumitsahrawat:
Squashed additions to the README in a single commit.

sumitsahrawat:
The build is passing locally, it's a temporary bug on the Travis CI side.

mpickering:
I restarted the build.

sumitsahrawat:
Seems to be working. travis-ci was hitting no route to host while using apt-get.

mpickering:
LGTM -- @jgm ?

jgm:
Is this approach better than the alternative of allowing options in
`--latex-engine` itself?  (On this alternative, we'd split off the
first word, and check that it is `xelatex`, `pdflatex`, or `lualatex`,
and treat the rest as options.)

sumitsahrawat:
I personally like the way the `-V` flag behaves. You can use it multiple times, and the combined effect of all the options appears. It's easy to use such a flag in a `Makefile`, and I made `--latex-engine-opt` (not opts) similar to that one for this reason. But, we could implement both of them.

sumitsahrawat:
Should I implement the alternative too?

jgm:
+++ Sumit Sahrawat [Mar 04 15 22:20 ]:
>   Should I implement the alternative too?

I just wanted to throw this out there, and start a discussion of whether the alternative might be better.  I don't have a firm view on this.  I think your approach, with a separate `--latex-engine-opt` for each separate argument, may have some advantages.

arve0:
I can't get this to work. I'm using `--latex-engine-opt="-shell-escape"` with minted package and get the error `! Package minted Error: You must invoke LaTeX with the -shell-escape flag.`

```sh
$ pandoc --version
pandoc 1.14
Compiled with texmath 0.8.2, highlighting-kate 0.5.15.
Syntax highlighting is supported for the following languages:
    abc, actionscript, ada, agda, apache, asn1, asp, awk, bash, bibtex, boo, c,
    changelog, clojure, cmake, coffee, coldfusion, commonlisp, cpp, cs, css,
    curry, d, diff, djangotemplate, dockerfile, dot, doxygen, doxygenlua, dtd,
    eiffel, email, erlang, fasm, fortran, fsharp, gcc, glsl, gnuassembler, go,
    haskell, haxe, html, idris, ini, isocpp, java, javadoc, javascript, json,
    jsp, julia, latex, lex, lilypond, literatecurry, literatehaskell, lua, m4,
    makefile, mandoc, markdown, mathematica, matlab, maxima, mediawiki,
    metafont, mips, modelines, modula2, modula3, monobasic, nasm, noweb,
    objectivec, objectivecpp, ocaml, octave, opencl, pascal, perl, php, pike,
    postscript, prolog, pure, python, r, relaxng, relaxngcompact, rest, rhtml,
    roff, ruby, rust, scala, scheme, sci, sed, sgml, sql, sqlmysql,
    sqlpostgresql, tcl, tcsh, texinfo, verilog, vhdl, xml, xorg, xslt, xul,
    yacc, yaml, zsh
Default user data directory: /Users/arve/.pandoc
```
For more details, [Makefile is found here](https://github.com/arve0/master/blob/master/Makefile) and [template here](https://github.com/arve0/master/blob/master/latex_template/pandoc.tex).

Using pdflatex directly seems to work, though I get awfully a lot of error messages (not sure why).

Any ideas?

sumitsahrawat:
I'll look into it.

sumitsahrawat:
A temporary workaround is to first convert your markdown to tex using `--template` option.
And then convert that tex to pdf using pandoc without the `--template option`.

```bash
pandoc --template latex-template/pandoc.tex input.md -o output.tex
pandoc --latex-engine-opt="-shell-escape" output.tex -o output.pdf
```

Your template doesn't create any `\begin{minted}{...} ... \end{minted}` blocks, thus the output is not as one would expect.

It seems like the issue arises somewhere else, I think @jgm and @mpickering would know better.

arve0:
@sumitsahrawat You are correct that I do not use `\begin{minted}` anywhere (yet). Just had a local edit here trying it out, but did not get it to work from markdown->pdf. I've tried the intermediate markdown->latex->pdf, but then it fails with

```
! Undefined control sequence.
<argument> 24 \si 
                  {\milli \metre }
l.200 ...\si{\milli\metre}}{400 \si{\micro\metre}}
```

even if `\usepackage{siunitx}` is in the intermediate tex file.

arve0:
Nevermind, seems like the last error message is due to my `$$` pandoc-crossref equation. Using intermediate .tex works. Thanks!

arve0:
Sorry the noise, was to fast. Using intermediate `.tex` fixes the original error, but introduces errors with the filters.

Issue #1977, opened 2015-02-27
Request for markdown writer to support output of shortcut reference links


binaryphile:
There is an existing option for outputting reference links, --reference-links.  However, it only generates [collapsed reference links](http://spec.commonmark.org/0.17/#collapsed-reference-link).

I would like to be able to generate shortcut reference links which do not include the empty trailing square brackets [].  To my eyes the empty brackets are visual clutter and unnecessary.

Could there be a separate command-line switch which adds support for [shortcut reference links](http://spec.commonmark.org/0.17/#shortcut-reference-link)?  Such an option might be --shortcut-reference-links.

Thanks for any consideration.

jgm:
Instead of adding another option, I like the idea of adding
a "shortcut_reference_links" Markdown extension, and
enabling it in both the reader and the writer.  IF the
extension is enabled (which would be the default), the
writer could use a shortcut reference link unless doing so
would cause problems (e.g. because it was followed
immediately by another link).

Just about every Markdown processor since Markdown.pl 1.0.1
supports the shortcut format.

+++ Ted Lilley [Feb 27 15 13:58 ]:

wcaleb:
Would such a default mean that words in brackets (non-links) would always have to be backslash-escaped? Or would they only have to be escaped if a link reference for the bracketed words existed somewhere in the file?

binaryphile:
Words in brackets only qualify as shortcut reference links if there is a matching link definition at the bottom of the page.  If not, the words in brackets aren't intended to be a link and are then rendered normally.

wcaleb:
Thanks for the clarification.

zudov:
Just wanted to mention that markdown reader already supports shortcut references. 

Of course, as @jgm said the option also should be added to the reader. When the option is disabled reader shouldn't parse the example below as a reference, am I right? 

```bash
zudov@x200 ~/ $ cat | pandoc -t html
[foo]

[foo]: /foo "bar"

<p><a href="/foo" title="bar">foo</a></p>
zudov@x200 ~/ $ 
```

zudov:
By the way, is anyone working on this?

mpickering:
Yes and no.

wcaleb:
@mpickering If it's more "no" than "yes," I'd like to try my hand at this ...

zudov:
[...]

PR #1989, opened 2015-03-04
Support shortcut reference links in markdown writer

zudov:
Issue #1977

Most markdown processors support the [shortcut format] for reference links.
Pandoc's markdown reader parsed this shortcuts unoptionally.
Pandoc's markdown writer (with --reference-links option) never shortcutted links.

This commit adds an extension `shortcut_reference_links`. The extension is
enabled by default for those markdown flavors that support reading shortcut
reference links, namely:

    - pandoc
    - strict pandoc
    - github flavoured
    - PHPmarkdown

If extension is enabled, reader parses the shortcuts in the same way as
it preveously did. Otherwise it would parse them as normal text.

If extension is enabled, writer outputs shortcut reference links unless
doing so would cause problems (see test cases in `tests/Tests/Writers/Markdown.hs`).

### DIFF
--- a/README
+++ b/README
@@ -2785,6 +2785,11 @@ in several respects:
     we must either disallow lazy wrapping or require a blank line between
     list items.
 
+#### Extension: `shortcut_reference_links` ####
+
+Allows to use shortcut reference links: `[foo]` instead of `[foo][]`. Writer
+would shortcut links unless doing so might cause problems.
+
 Markdown variants
 -----------------
 
--- a/src/Text/Pandoc/Options.hs
+++ b/src/Text/Pandoc/Options.hs
@@ -109,6 +109,7 @@ data Extension =
     | Ext_implicit_header_references -- ^ Implicit reference links for headers
     | Ext_line_blocks         -- ^ RST style line blocks
     | Ext_epub_html_exts      -- ^ Recognise the EPUB extended version of HTML
+    | Ext_shortcut_reference_links -- ^ Shortcut reference links
     deriving (Show, Read, Enum, Eq, Ord, Bounded)
 
 pandocExtensions :: Set Extension
@@ -151,6 +152,7 @@ pandocExtensions = Set.fromList
   , Ext_header_attributes
   , Ext_implicit_header_references
   , Ext_line_blocks
+  , Ext_shortcut_reference_links
   ]
 
 phpMarkdownExtraExtensions :: Set Extension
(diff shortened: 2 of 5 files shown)


wcaleb:
One thought I had last night, is that shortcuts also need to be turned off if the link is followed by text in square brackets (or a space and then text in square brackets. Otherwise there is the possibility that the writer will produce improper Markdown that could lose a link when dealt with by the reader.

```
$ cat | pandoc --to native --reference-links
Here is a case of a [link] [followed by text].

  [link]: /url
[Para [Str "Here",Space,Str "is",Space,Str "a",Space,Str "case",Space,Str "of",Space,Str "a",Space,Str "[link]",Space,Str "[followed",Space,Str "by",Space,Str "text]."]]
```

The AST also discards the link if there is no space between the `[link]` and `[followed by text]`.

Thanks for all your work on this @zudov!

zudov:
@wcaleb Nice catch. I'll add a test case and will cover it. Thanks. I appreciate the help on that whole issue.

zudov:
Fixed, please notify me before merging this PR so that I would squash the commits together.

jgm:
Why don't you squash the commits now, so that the PR will be easier to
review?

+++ Konstantin Zudov [Mar 04 15 13:12 ]:

zudov:
@jgm Done

zudov:
Hmm. I also was thinking that it can cause problem if RawInline with opening brackets follows the shortcuted link, would check it.

wcaleb:
@zudov Could you solve both problems by creating a final pattern match that stringifies whatever inline follows the Link (or Link Space) and then checks for an opening bracket at beginning? Not sure if it's advisable, just thinking aloud.

zudov:
@wcaleb I think the most bulletproof approach would be to render the next element and check if it starts with "[" or " [". But I don't think it's really necessary.

zudov:
On the other hand it might be

zudov:
Fixed for `RawInline`. I guess that should be it. Can't come up with any other inlines that can break things.

timtylin:
Is this meant to target Pandoc markdown specifically? If so, you might have to worry about citations.

zudov:
@timtylin You are right. I've just fixed it. Thanks.

zudov:
@jgm I've updated updated a comment and added a note to README. Not sure is it good enough, feel free to change it.

zudov:
Squashed the commits.

jgm:
Looks good, thanks!

PR #1999, opened 2015-03-08
LaTeX reader: minor table parsing improvements

synsem:
This enables pandoc to parse `tabular` environments with optional arguments and non-empty column separators (which should be fairly common in real world LaTeX documents).

Basic test cases for simple tables are also included, modeled roughly after the test cases for the Org reader.


jgm:
Great, thanks!

PR #2013, opened 2015-03-19
Derive from Data and Typeable for libpandoc

ShabbyX:
See https://github.com/jgm/highlighting-kate/pull/64

### DIFF
--- a/src/Text/Pandoc/MediaBag.hs
+++ b/src/Text/Pandoc/MediaBag.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving, DeriveDataTypeable #-}
 {-
 Copyright (C) 2014 John MacFarlane <jgm@berkeley.edu>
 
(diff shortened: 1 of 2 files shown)


ShabbyX:
Thanks! This was quick!

jgm:
Easy fix.  By the way, in the old libpandoc repo
https://github.com/t0yv0/libpandoc
I note that libpandoc.cabal says the license is BSD3.
This is, at the very least, highly misleading.  Pandoc is GPL,
and anything that links with it needs to be GPL too.  Someone
might look at libpandoc, see it's BSD3, and not realize this.
Can this be fixed?

ShabbyX:
@jgm, a GPL library can be linked with any GPL-compatible library/application. If I understand correctly, BSD3 means _Modified BSD License_ otherwise known as _3-clause BSD license_, which according to [Wikipedia](https://en.wikipedia.org/wiki/BSD_licenses) is compatible with GPL.

Since I am not the copyright holder (that would be t0yv0) and my edits haven't been very extensive either, I don't think I have the rights to relicense the library. If it's legal issues you are worried about, there are none because the licenses are compatible.

If you just want people to know that Pandoc itself is licensed under GPL, I can make it clear in the README.

jgm:
I think @t0yv0 should relicense it.  I think it's legal for him to license his own code BSD3, but, as I said, it's incredibly misleading for others if the pandoc binding (which is only useful when linked with pandoc) is BSD3 licensed.  People who do not check the license of pandoc itself may think they're only linking with BSD3 licensed libraries, when in fact they aren't.  At the very least, there should be a very visible statement on the README, in the cabal file, and in the C header file that any library that links with libpandoc must comply with the terms of pandoc's license, the GPL.  But it would avoid much confusion simply to license the library as GPL.

+++ Shahbaz Youssefi [Mar 19 15 09:41 ]:

ShabbyX:
I will try to get a hold of him, but given his unresponsiveness regarding the PR on libpandoc itself, I wouldn't count on any response from him.

In the meantime, I would add the license information on the source files (in the README I already have). In the cabal file, I'm not sure how one can indicate that (Put under `Description:` maybe?).

ShabbyX:
@jgm t0yv0 was nice enough to transfer copyright of libpandoc to me. I relicensed it to GPLv2+ to clear the confusion.

jgm:
Great!

+++ Shahbaz Youssefi [Mar 21 15 10:56 ]:

PR #2023, opened 2015-03-24
Docx Writer: Fix StyleMap

lierdakil:
I've messed up badly with it, so it didn't work properly in many cases. Really sorry about that...

At the plus side, fallback mechanic is working wonderfully -- that's primary reason why it wasn't noticed until now.

I hope this is merged ASAP, since [some users are already affected][1]

[1]:https://groups.google.com/forum/#!topic/pandoc-discuss/iRml5jufqXg

### DIFF
--- a/src/Text/Pandoc/Readers/Docx/StyleMap.hs
+++ b/src/Text/Pandoc/Readers/Docx/StyleMap.hs
@@ -9,7 +9,6 @@ import           Text.XML.Light
 import           Text.Pandoc.Readers.Docx.Util
 import           Control.Monad.State
 import           Data.Char  (toLower)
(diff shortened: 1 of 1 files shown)


mpickering:
I'll wait for travis and then merge. Can you just describe what went wrong in a sentence?

lierdakil:
Basically, if it failed to get mapping for any single style, it failed altogether. And it fails to get style mappings for many style types by design (since we don't use those in writer).

In short, I didn't think this through and used a `StateT` transformer over `Maybe`, while single failure shouldn't halt whole algo.

lierdakil:
Travis is done.

Issue #2032, opened 2015-03-28
reference.docx corrupt


jgm:
Opening `data/reference.docx` gives the warning "One or more footnotes referenced in this document are missing or corrupt."  This must be due to a fairly recent change in `data/reference.docx` -- perhaps 1cb601d2885df09fe07533006c06d8a603c3020d or ba59e5447ffe8d15dd7fee69a7aa03706ce3c49b.

First step is to track down which commit introduced the problem.

nkalvi:
@jgm I'm wondering whether reference to footnotes in **word/settings.xml**

```
    <w:footnotePr>
        <w:footnote w:id="-1" />
        <w:footnote w:id="0" />
    </w:footnotePr>
```

is the cause. To remedy, something like the following can be added to **word/footnotes.xml** (which is now empty):

```
<?xml version="1.0" encoding="UTF-8"?>
<w:footnotes
    xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"
    xmlns:m="http://schemas.openxmlformats.org/officeDocument/2006/math"
    xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships"
    xmlns:o="urn:schemas-microsoft-com:office:office"
    xmlns:v="urn:schemas-microsoft-com:vml"
    xmlns:w10="urn:schemas-microsoft-com:office:word"
    xmlns:a="http://schemas.openxmlformats.org/drawingml/2006/main"
    xmlns:pic="http://schemas.openxmlformats.org/drawingml/2006/picture"
    xmlns:wp="http://schemas.openxmlformats.org/drawingml/2006/wordprocessingDrawing">
    <w:footnote w:type="separator" w:id="-1">
        <w:p w:rsidR="00AC05CC" w:rsidRDefault="00AC05CC" w:rsidP="0051672D">
            <w:r>
                <w:separator/>
            </w:r>
        </w:p>
    </w:footnote>
    <w:footnote w:type="continuationSeparator" w:id="0">
        <w:p w:rsidR="00AC05CC" w:rsidRDefault="00AC05CC" w:rsidP="0051672D">
            <w:r>
                <w:continuationSeparator/>
            </w:r>
        </w:p>
    </w:footnote>
</w:footnotes>
```

When I tested with this addition, the error was gone (of course separator etc. need to be set to the desired values.)

nkalvi:
[...]

PR #2034, opened 2015-03-29
Reference Docx: update word/footnotes.xml

lierdakil:
Quick fix for issue #2032. Docx Writer puts this contents into footnotes.xml by default, ignoring reference.docx/word/footnotes.xml. If user tries to edit reference.docx directly, however, Word will complain about corrupt footnotes. This PR fixes it.


lierdakil:
Oh, and by the way, two tests fail due to recent changes in fenced code blocks for html. I'll submit a PR updating said tests shortly.

jgm:
+++ Nikolay Yakimov [Mar 28 15 20:05 ]:
>Oh, and by the way, two tests fail due to recent changes in fenced code blocks for html. I'll submit a PR updating said tests shortly.

I changed a couple tests to conform to new highlighting-kate
behavior.

lierdakil:
@jgm, yes, I got it shortly after posting. I guess comment updates are not sent out via e-mail.

jgm:
Great, that worked.

nkalvi:
For future troubleshooting, would it be helpful to use revision numbers in reference.docx (f.ex. `<cp:revision>2034</cp:revision>` in doc properties when making changes?

lierdakil:
@nkalvi we do have git, and now it tracks changes in reference files, since those are unpacked, so I don't think this would be necessary.

nkalvi:
Thanks again @lierdakil. I agree that since the files are unpacked now one can easily see the history. With packed ('Binary file not shown') and I had to download and unpack each version to see the changes.

My thinking was, when someone reports and issue with reference.docx, it would be helpful have them specify the revision number along with Pandoc's version etc.

lierdakil:
@nkalvi, every pandoc release has a well-defined commit hash associated with it, and reference.docx is tracked in the same repository. And if someone is building from `master`, we can assume that HEAD revision is known. I think reference.docx revision would be redundant.

Issue #2031, opened 2015-03-28
DocX: figure and figure caption are separated at page boundaries


tolot27:
If a figure is placed at the end of a page, its caption is moved/placed at the next page. See https://groups.google.com/d/topic/pandoc-discuss/DKEuJl242sE/discussion for a more detailed explanation.

Proposol: either asign a new format style (i.e. 'image') which has the 'Keep with next' paragraph option set by default and is adjustable in any way (i. e. center align) or write the following XML node `w:pPr` to the image paragraph during docx creation:

```xml
		<w:p>
			<w:pPr>
				<w:keepNext />
				<w:jc w:val="center" />
			</w:pPr>
			<w:r>
				...
```

lierdakil:
#2035 should work, but it feels a little bit clunky, since figures with empty captions need special handling. If you have any suggestions, please leave a comment there.

jgm:
Closing this, as the patch has been merged.

PR #2035, opened 2015-03-29
Docx Writer/Reference: Add keepNext to objects w/ captions

lierdakil:
This changes `TableCaption` style, adding `w:keepNext` option, and adds `Figure` style to figure images and `FigureWithCaption` to figures with nonempty captions, which is based on `Figure`, and additionally has `w:keepNext` set as well

With this, caption and table/figure are not separated at page boundaries, as per #2031.

### DIFF
--- a/README
+++ b/README
@@ -625,10 +625,11 @@ Options affecting specific writers
     `--data-dir`). If this is not found either, sensible defaults will be
     used. The following styles are used by pandoc: [paragraph]
     Normal, Compact, Title, Subtitle, Authors, Date, Abstract, Heading 1,
-    Heading 2, Heading 3, Heading 4, Heading 5, Block Quote, Definition Term,
-    Definition, Bibliography, Body Text, Table Caption, Image Caption;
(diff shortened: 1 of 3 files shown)


lierdakil:
This feels a bit clunky with two separate styles for figures. If anyone has a better idea, I'm open to suggestions.

One option might be to hard-add `w:keepNext` to image paragraph itself, not depending on style, but I feel that would limit customisation options.

mpickering:
I think the section in the documentation needs to be updated as well.

lierdakil:
Ok, updated README with new docx style names. I've also updated those that were changed in 1cb601d2885df09fe07533006c06d8a603c3020d.
Docx style semantics probably should be documented somewhere on wiki as well, as it's not immediately intuitive.

tolot27:
Thanks for implementing this. I don't see any other option to implement this as long as Word has no `w:keepPrevious` (which is suggested to Microsoft for years, btw).
Word itself hardcodes the `w:keepNext` if you add a caption to a figure (see context menue of figure). I would follow the same procedure and do not "over style". Having a separate figure style is one step further already.

Issue #2021, opened 2015-03-23
preserve hyphenation settings in docx output


tolot27:
Some docx document settings (i. e. autoHyphenation) are stored in the settings.xml of an docx archive. Unfortunately, pandoc does not respect these settings (stored in a custom reference.docx), at least not the hyphenation settings.

It would be great if the docx writer would preserve/transfer these settings (see below for a more or less complete list) to the newly generated docx like it does for the margins, page size, page orientation, etc.

```xml
<w:autoHyphenation />
<w:consecutiveHyphenLimit w:val="2"/>
<w:hyphenationZone w:val="284"/>
<w:doNotHyphenateCaps/>
```

lierdakil:
We use dist archive for settings.xml, because Word sometimes adds references to footnotes or endnotes we don't have. Copying over settings that you mention should be relatively easy however. I'll see what I can do.

tolot27:
It works as expected. Great! Thanks!

PR #2038, opened 2015-03-29
Docx Writer: Copy hyphenation settings from reference.docx

lierdakil:
As requested in #2021.


lierdakil:
Actually, with ba59e54, it might be safe to copy `settings.xml` from reference.docx entirely. But I'd prefer to leave it like this, just to be on the safe side.

jgm:
Thanks for all the good patches!

Issue #1866, opened 2015-01-10
Block macro in latex causes parse failure


mpickering:
Reproduced from https://groups.google.com/d/msgid/pandoc-discuss/90cb9924-0ed1-441a-8ef2-f2e3928a165c%40googlegroups.com

-----

So this is what happens.. the macro parsing is all ok. The problem comes when \FIG is parsed. \FIG is replaced with the following..

```
{
  \begin{figure}[h!]
    \centering
    \includegraphics[width=0.5\columnwidth,angle=0]{test.jpg}
    \caption{Test caption}
    \label{fig:test.jpg}
  \end{figure}
}
```

The next line now fails

```
parseFromString inlines transformed
```

The next step is into `grouped inline` which matches the first `{`. Little does this function know, what is inside this block is actually a block level element.

Next `inline` tries to match `\begin{figure}` but all paths fail so the parse fails. Pre pandoc 1.13.3 (ie any released version of pandoc) the `parseFromString` function succeeded even if it didn't consume all the input (it just silently threw it away). Now in HEAD, it fails if not all the input is consumed.

mpickering:
Two things need doing.

- [ ] Fix this issue
- [ ] Improve error messages when code latex macros are expanded

jgm:
Related to #1390.

mpickering:
I've been thinking, would the following approach be sufficient?

1) Parse `\newcommand' as normal
2) Attempt to parse the body of the macro with the `inline` parser. If this succeeds then we mark the macro as `Inline`. 
3) If 2) fails, attempt to parse the body with the `block` parser. If this succeeds then we mark the macro as `Block. 
4) If both fail then the body would not produce valid LaTeX when expanded so we can fail with hopefully a better error message at that point.
5) When parsing, add a new clause to the `block` parser which checks for `Block` macros.

Thoughts?

lierdakil:
Soooo... I fiddled around with it a bit, and I've got myself lierdakil@b5e857c, which solves problems mentioned in this issue and issues mentioning this, without breaking anything (to my knowledge, that is, but apparently latex tests are non-exhaustive). I am, however, not entirely sure if it solves *all* associated problems. However, even cases mixing block- and inline-level macros are parsed reasonably well (notice the extra span here -- it's due to #1390):

```latex
\newcommand{\ch}[1]{\chapter{#1}}
\newcommand{\ita}[1]{\emph{#1}}

\ch{Test}\ita{hello}
```

```haskell
[...]

PR #2042, opened 2015-03-29
LaTeX Reader: check for block-level newcommand aliases in blockCommand

lierdakil:
This came to be in response to #1866. I'm creating a pull request to facilitate the process somewhat.

Comments are welcome.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -239,17 +239,28 @@ block = (mempty <$ comment)
 blocks :: LP Blocks
 blocks = mconcat <$> many block
 
+getRawCommand :: String -> LP String
+getRawCommand name' = do
+  rawargs <- withRaw (skipopts *> option "" dimenarg *> many braced)
+  return $ '\\' : name' ++ snd rawargs
+
 blockCommand :: LP Blocks
 blockCommand = try $ do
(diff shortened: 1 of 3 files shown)


jgm:
Question:  what happens if a paragraph begins with a macro that is meant to be inline-level? For example,
```
\newcommand{\wbal}{The Wikibook about \LaTeX}

\wbal is a good resource for learning \LaTeX.
```
Running your code in my head (which is not very reliable!), I worried that your code might parse this as a Plain or Para with "The Wikibook about LaTeX" followed by another Para with "is a good resource for learning LaTeX", rather than as a single paragraph.  What does it do with this input?  (Cf. @mpickering's suggestion in #1866 to treat a macro as block only if it doesn't parse as inlines.)

lierdakil:
That is a good point, thanks.

It is relatively simple to guard against (by following @mpickering's suggestion more closely), although it does introduce more backtracking. I'll push relevant changes shortly.

lierdakil:
A small style change. I believe it's easier to grasp what's going on this way.

As a side note, code style in LaTeX reader seems quite inconsistent, and could use some cleanup maybe.

jgm:
@lierdakil, code cleanup is always welcome!  (I note that some parts of the pandoc code date from when I was first learning Haskell. Gradually it improves, but it's a big code base.)

PR #2055, opened 2015-04-06
.md extension for README

raunaqrox:
Github was not showing the README in markdown


lierdakil:
It's not in gfm, so there is literally no point, it's malformatted anyway.

raunaqrox:
It looks better.

lierdakil:
It looks broken.

raunaqrox:
I understand.

jgm:
There's a correctly formatted version here: http://pandoc.org/README.html

lierdakil:
I think we can manage to use reST README for github. [pandoc/readme-rst branch](https://github.com/lierdakil/pandoc/tree/readme-rst) contains an example. However, there are a few "quirks" in reST writer, and docutils do not parse straightforward conversion of readme. I managed to work around most of those using a 10-line filter though.

jgm:
+++ Nikolay Yakimov [Apr 12 15 02:36 ]:
>I think we can manage to use reST README for github. [pandoc/readme-rst branch](https://github.com/lierdakil/pandoc/tree/readme-rst) contains an example. However, there are a few "quirks" in reST writer, and docutils do not parse straightforward conversion of readme. I managed to work around most of those using a 10-line filter though.

This would require either checking a generated README.rst into the repository,
or using rst rather than pandoc markdown for the master README, which
seems a bit odd for this project.  Is it worth it?

lierdakil:
Honestly, I think not. Personally, I'm fine with raw Markdown, since github page is not really meant for new users anyway. But this PR got me thinking, and I came up with this idea. If it's a good one -- well, I wasn't sure, so I shared it. We might as well bury it and never speak of it again :laughing: 

Another option to prevent this kind of PRs might be to include comment into README header with a link to rendered version (e.g. `<!-- Rendered version of this README can be found at http://pandoc.org/README.html -->`) -- it would be a much simpler solution.

jdickey:
Pandora itself seems to do a reasonable job of converting its own README to Markdown; is [this Gist](https://gist.github.com/jdickey/cd268f9751c0c54eb9a172f30360c030) a reasonable starting point, at least?

It's *hard* to get non-dyed-in-the-wool nerds (like quasi-technical stakeholders with veto powers) to read "jumbled text" as in the present README as viewed on GitHub.

jgm:
+++ Jeff Dickey [Mar 31 16 08:31 ]:
>   It's hard to get non-dyed-in-the-wool nerds (like quasi-technical
>   stakeholders with veto powers) to read "jumbled text" as in the present
>   README as viewed on GitHub.

Real dyed-in-the-wool nerds are happy to look at plain text,
like the pandoc README.

But the main point is this:  pandoc's README is written in
pandoc's Markdown extension, which github doesn't support.
So, things like definition lists and footnotes won't be
rendered correctly if we rename as .md.

jgm:
+++ John MacFarlane [Mar 31 16 09:34 ]:
>+++ Jeff Dickey [Mar 31 16 08:31 ]:
>>  It's hard to get non-dyed-in-the-wool nerds (like quasi-technical
>>  stakeholders with veto powers) to read "jumbled text" as in the present
>>  README as viewed on GitHub.
>
>Real dyed-in-the-wool nerds are happy to look at plain text,
>like the pandoc README.
>
>But the main point is this:  pandoc's README is written in
>pandoc's Markdown extension, which github doesn't support.
>So, things like definition lists and footnotes won't be
>rendered correctly if we rename as .md.

Two more notes on this:

1.  You can point quasi-technical stakeholders to the
official website, http://pandoc.org, which presents the
README/User's Guide in all its glory:
see http://pandoc.org/README.html, or, if you prefer,
http://pandoc.org/README.pdf.

2.  You might ask:  why not just stick to the extensions
github supports?  Mainly because footnotes and definition
lists are important for this.  The pandoc man page is
generated from the README, and definition lists are used
for lists of command-line options and other things.

jdickey:
@jgm Very true. As I've said in several Markdown-related discussions, "I'd cheerfully kill for reliable, reasonably-presented footnotes." Be careful what you wish for. This looks awesome, actually.

PR #2068, opened 2015-04-11
MediaWiki reader: Fix quotation mark parsing

### LINKED ISSUE #2012: broken mediawiki parsing of quotation marks
This markdown file:

``` markdown
"quoted *emphasized*"

*"emphasized quoted"*

"*emphasized* quoted"

"*quoted emphasized*"
```

renders with `pandoc -t native -S -o- test.md` as this:

``` pandoc
[Para [Quoted DoubleQuote [Str "quoted",Space,Emph [Str "emphasized"]]]
,Para [Emph [Quoted DoubleQuote [Str "emphasized",Space,Str "quoted"]]]
,Para [Quoted DoubleQuote [Emph [Str "emphasized"],Space,Str "quoted"]]
,Para [Quoted DoubleQuote [Emph [Str "quoted",Space,Str "emphasized"]]]]
```

...whereas this semantically identical mediawiki file:

``` mediawiki
"quoted ''emphasized''"

''"emphasized quoted"''

"''emphasized'' quoted"

"''quoted emphasized''"
```

renders with `pandoc -f mediawiki -t native -S -o- test.mediawiki` as this:

``` pandoc
[Para [Quoted DoubleQuote [Str "quoted",Space,Emph [Str "emphasized"]]]
,Para [Emph [Quoted DoubleQuote [Str "emphasized",Space,Str "quoted"]]]
,Para [Str "\"",Emph [Str "emphasized"],Space,Str "quoted\""]
,Para [Str "\"",Emph [Str "quoted",Space,Str "emphasized"],Str "\""]]
```

Seems mediawiki parsing is less smart in detecting quotation markers.

tgkokk:
Change MediaWiki reader's behaviour when the smart option is parsed to
match other readers' behaviour (inspired by the Org reader).

Fix #2012.

### DIFF
--- a/src/Text/Pandoc/Readers/MediaWiki.hs
+++ b/src/Text/Pandoc/Readers/MediaWiki.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE RelaxedPolyRec, FlexibleInstances, TypeSynonymInstances #-}
+{-# LANGUAGE RelaxedPolyRec, FlexibleInstances, TypeSynonymInstances, MultiParamTypeClasses #-}
 -- RelaxedPolyRec needed for inlinesBetween on GHC < 7
 {-
   Copyright (C) 2012-2014 John MacFarlane <jgm@berkeley.edu>
@@ -48,6 +48,8 @@ import Text.Pandoc.Shared ( stripTrailingNewlines, safeRead, stringify, trim )
 import Data.Monoid (mconcat, mempty)
 import Control.Applicative ((<$>), (<*), (*>), (<$))
 import Control.Monad
+import Control.Monad.Reader (Reader, runReader, asks, local)
+import Data.Default
 import Data.List (intersperse, intercalate, isPrefixOf )
 import Text.HTML.TagSoup
 import Data.Sequence (viewl, ViewL(..), (<|))
@@ -64,15 +66,9 @@ import Text.Pandoc.Error
 readMediaWiki :: ReaderOptions -- ^ Reader options
               -> String        -- ^ String to parse (assuming @'\n'@ line endings)
               -> Either PandocError Pandoc
-readMediaWiki opts s =
-  readWith parseMediaWiki MWState{ mwOptions = opts
-                                       , mwMaxNestingLevel = 4
-                                       , mwNextLinkNumber  = 1
-                                       , mwCategoryLinks = []
-                                       , mwHeaderMap = M.empty
-                                       , mwIdentifierList = []
-                                       }
-           (s ++ "\n")
+readMediaWiki opts s = fst <$> res
+  where imd = readWithM (returnState parseMediaWiki) def{ mwOptions = opts } (s ++ "\n")
+        res = runReader imd def
 
 data MWState = MWState { mwOptions         :: ReaderOptions
                        , mwMaxNestingLevel :: Int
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Readers/MediaWiki.hs
```diff
@@ -653,10 +671,15 @@ strong = B.strong <$> nested (inlinesBetween start end)
...
-doubleQuotes :: MWParser Inlines
-doubleQuotes = B.doubleQuoted . trimInlines . mconcat <$> try
- ((getState >>= guard . readerSmart . mwOptions) *>
-   openDoubleQuote *> manyTill inline closeDoubleQuote )
-    where openDoubleQuote = char '"' <* lookAhead alphaNum
-          closeDoubleQuote = char '"' <* notFollowedBy alphaNum
+smart :: MWParser Inlines
+smart = do
+  getOption readerSmart >>= guard
+  doubleQuotes <|> choice [apostrophe, dash, ellipses]
```
tgkokk: Heads up: For some reason, the `choice` part doesn't work (even if I remove the `doubleQuotes` part). I can open an issue once this PR gets merged.

lierdakil:
You can push into this branch, and new commits will be added to this PR.

lierdakil:
Sorry, misread your comment. Does it work in Org reader though? I assume it does.

tgkokk:
Yeah, it does.

lierdakil:
`choice` tries to apply parsers in order they are supplied. `...` will be parsed by `str` parser, since `.` is not a special character. Same goes for `--`. 

`'` *is* a special character, but it may be taken by other parser (or maybe it works?).

tgkokk:
Yeah, I forgot to mention that `'` is taken up by `smart

tgkokk:
OK, adding `.` and `-` to `specialChars` fixed it. Thanks!

jgm:
+++ tgkokk [Apr 11 15 03:59 ]:
>   Change MediaWiki reader's behaviour when the smart option is parsed to
>   match other readers' behaviour (inspired by the Org reader).

I wonder whether this is really desirable?  The aim of a
MediaWiki reader is to reproduce, as far as possible, the
way MediaWiki actually interprets text.  So, we should
have the reader parse smart punctuation the way the other
readers do if, and *only if*, MediaWiki works that way.

Does it?  If so, can you point to relevant documentation?

lierdakil:
@jgm
While MediaWiki doesn't work this way, at least to my knowledge, `--smart` is not enabled by default, and it may be a reasonable way of getting MW input to behave in a typographically correct way. If that's disagreeable, we could drop support for apostrophes, dashes and ellipses, but that would be somewhat counter-intuitive, IMO, and is likely to spawn some reports in pandoc-discuss and here. 

After all, MediaWiki doesn't support smart quotes either, and just includes quotation marks verbatim, and we do have support for those (although not as smart as other readers, as per #2012)

tgkokk:
@jgm

If we're only including smart punctuation if MediaWiki does, then no, MediaWiki doesn't seem to take it into account.

However, I did some research and `org` doesn't actually support "smart apostrophes" the way it's implemented in `pandoc` (when an option is passed, `'foo'` is converted to ``foo'` (that's backtick-foo-apostrophe)).

`TWiki` does not seem to support smart punctuation at all (well, it does have a [plugin](http://twiki.org/cgi-bin/view/Plugins/TypographyPlugin), but it's probably not enabled by default, considering the warning on the top of the page).

`Markdown` ([reference implementation](https://daringfireball.net/projects/markdown/)) does not support smart punctuation, and neither do any of the major implementations (including Github's). However, since `pandoc` uses its own version of `Markdown`, I guess we can do whatever we want.

`RST` supports only em-dashes, and not how it's done in `pandoc` (see [this](http://docutils.sourceforge.net/FAQ.html#how-can-i-represent-esoteric-characters-e-g-character-entities-in-a-document), bottom of the answer).

So, if we shouldn't support smart punctuation in the MediaWiki reader, I think we should fix the above inconsistencies.

However, I think I will agree with @lierdakil: It's not enabled by default and I guess it's nice to have, even if it is not supported in the respective markup format.

jgm:
This needs rebasing.  I tried manually but the patch uses `returnState` which is undefined?

tgkokk:
@jgm I must have accidentally deleted my fork, so I can't update this PR. I'm going to close it and create a new one.

PR #2070, opened 2015-04-11
ODT Writer: Figure captions

### LINKED ISSUE #376: Markdown to odt: image captions are missing
I am using pandoc 1.8.2.1 (from cabal) and LibreOffice 3.4.4 on Windows 7.

lierdakil:
Closes #376.

Works pretty much the same as Word writer.

Following styles are used for figures:

Figure -- for figure with empty caption
FigureWithCaption (based on Figure) -- for figure with caption
FigureCaption (based on Caption) -- for figure captions

Also, TableCaption (based on Caption) is used for table captions.

We need FigureWithCaption to set keepWithNext, in order to keep caption
with figure across pages.


lierdakil:
Forgot to update tests. Will push shortly.

PR #2072, opened 2015-04-12
LaTeX Reader: Code cleanup

lierdakil:
Something I wanted to do for a little while.

Removed redundant brackets and `$`s, consolidated imports, replaced some more obvious things with library functions, simplified nested `case` expressions, and changed `[Char]`s to `String`s.

I also accidentally stumbled on a little bug, that prevented `tildeEscape` parser from ever working. Fixed that as well (line 203 in new source). Should I also add a test, while I'm at it?

Behavior shouldn't be in any way changed, except for that bug thing, but wait for Travis just in case.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -42,20 +42,18 @@ import Text.Pandoc.Options
 import Text.Pandoc.Parsing hiding ((<|>), many, optional, space,
                                    mathDisplay, mathInline)
 import qualified Text.Pandoc.UTF8 as UTF8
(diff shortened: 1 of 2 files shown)


jgm:
+++ Nikolay Yakimov [Apr 12 15 00:00 ]:
>   I also accidentally stumbled on a little bug, that prevented
>   tildeEscape parser from ever working. Fixed that as well (line 203 in
>   new source). Should I also add a test, while I'm at it?

That's a good idea!

I'll wait for the test before merging.

lierdakil:
Done. Test added, Travis is happy.

jgm:
Thanks, this kind of cleanup is always welcome.

PR #2079, opened 2015-04-13
RST Writer: Normalize headings to sequential levels

lierdakil:
This is pretty much required by docutils.

First level will be interpreted as level 1 anyway, and subsequent levels must be sequential.

See [pandoc-discuss thread][1] for background.

I'm not entirely sure it's OK to handle this in such way, since heading level information is irreversibly lost during conversion, but I see no other option really.

[1]: https://groups.google.com/d/msg/pandoc-discuss/mTG4phv5roA/ljApKvFDrY4J

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -495,6 +495,7 @@ Test-Suite test-pandoc
                   Tests.Writers.AsciiDoc
                   Tests.Writers.LaTeX
                   Tests.Writers.Docx
+                  Tests.Writers.RST
   Ghc-Options:  -rtsopts -Wall -fno-warn-unused-do-bind -threaded
   Default-Language: Haskell98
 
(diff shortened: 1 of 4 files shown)


jgm:
Are headers allowed in block quotes and list items in reST?  If not, maybe we should remove them (convert them to regular paragraphs, perhaps in a Strong) as well.

lierdakil:
Well, technically there is a `rubric` directive, which is described as "informal section", i.e. a heading with no relation to document structure, which *can* be included in block quotes and lists. Its exact semantics are a little fuzzy though, f.ex. there is no obvious way to denote rubric level.

Proper sections (i.e. headings) are not allowed anywhere except at top level though.

jgm:
I'd be fine with making all non-outside headers into rubrics.
Or just making them paragraphs.  But we should do one or the other, to
avoid creating invalid rst.

lierdakil:
Please hold off on merging until I add some tests for this.

atorin:
FWIW, in my experience _rubrics_ tend to pollute the RST output with useless information. 
Would it not be easier to just use bold instead?

PR #2091, opened 2015-04-17
LaTeX Writer: Group biblatex citations

### LINKED ISSUE #2065: Problems getting numeric-comp citation style to work
Hi,

I have a set of MarkDown files in which I use references à la `[@REF1; @REF2; @REF3]`. When I run `pandoc` directly on the MarkDown files to produce a PDF output with the Nature CSL, I get them nicely formatted into `[REF1 - REF3]`.

In parallel, I am also creating a TeX output that I am running through `latexmk`. However, despite specifying

```
$if(biblatex)$
\usepackage[
    citestyle=numeric-comp,
    style=numeric,
    isbn=false,
    doi=false,
    sorting=none,
    autocite=inline
]{biblatex}
```

in a custom template file (derived from the original `default.latex`), the references are not compacted but are listed as `[REF1, REF2, REF3]`.

This seems to be related to `pandoc` creating `\autocites{REF1}{REF2}{REF3}` which apparently does not allow the references to be compacted. If I manually edit that to `\cite{REF1, REF2, REF3}` the result is the desired `[REF1 - REF3]` also when using `latexmk`.

It may appear somewhat counterintuitive to produce a PDF via `pandoc` **and in parallel** via `latexmk` but I am nevertheless wondering how to get the `latexmk` output to match the `pandoc` output.

Looking forward to your clarifications.

Best,

Cedric

lierdakil:
Closes #2065 

This is intended to group citations by prefix and suffix when using biblatex citations. Primary reason to want this is ability to collapse reference lists when using numeric-comp citation style. See #2065 for examples and discussion.

Grouping by suffix may not be optimal for all cases, but I think there is a very slight chance of unintended suffix collision. However, that's an arguable point.

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -39,7 +39,7 @@ import Text.Pandoc.Templates
 import Text.Printf ( printf )
 import Network.URI ( isURI, unEscapeString )
 import Data.List ( (\\), isSuffixOf, isInfixOf, stripPrefix,
-                   isPrefixOf, intercalate, intersperse )
+                   isPrefixOf, intercalate, intersperse, partition )
 import Data.Char ( toLower, isPunctuation, isAscii, isLetter, isDigit, ord )
 import Data.Maybe ( fromMaybe )
 import Data.Aeson.Types ( (.:), parseMaybe, withObject )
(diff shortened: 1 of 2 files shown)


jgm:
I'm worried about grouping by suffix (if I understand properly what you mean).  I'd be more comfortable with doing this kind of compacting only when there are no prefixes or suffixes, and leaving anything with a prefix or suffix by itself.

lierdakil:
Well, that's a reasonable opinion.

That said, I still feel like
```latex
\autocites[Church in][]{ref1,ref2,ref3}[Turing in]{ref4,ref5,ref6}
```
is a valid use-case. It would be unnecessarily verbose in Markdown with current implementation though, so maybe there could be another option for this -- I'm not entirely sure.

lierdakil:
Eh. This apparently needs more thought, since I mishandle `AuthorInText` to boot. I'll reopen when ready.

PR #2110, opened 2015-04-25
Use the new Travis sudo-less syntax for APT dependencies

timtylin:
Travis is planning to roll out a new [Container-based VM system](http://blog.travis-ci.com/2015-03-31-docker-default-on-the-way/) in the future. It promises shorter queue times and the possibility of [persisting/caching](http://docs.travis-ci.com/user/caching/) built dependencies. The latter could possibly be combined with LTS Haskell to enable <1 minute Travis build times, instead of ~15 minutes like now.

To do this safely they have to remove the ability to run as root via `sudo`. Since most uses of root access is to install APT packages, they came up with a [config syntax](http://docs.travis-ci.com/user/apt/) for specifying sources and packages.

@hvr suggested how the multi-version GHC build environments can use this new syntax [here](https://github.com/hvr/multi-ghc-travis/issues/17). This PR adapts that method to the current Pandoc Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,24 @@
 # NB: don't set `language: haskell` here
 
+# Ensures that sudo is disabled, so that containerized builds are allowed
+sudo: false
+
 # The following enables several GHC versions to be tested; often it's enough to test only against the last release in a major GHC version. Feel free to omit lines listings versions you don't need/want testing for.
-env:
-# - GHCVER=6.12.3
- - CABALVER=1.16 GHCVER=7.4.2 GHCOPTS="-Werror"
- - CABALVER=1.18 GHCVER=7.6.3 GHCOPTS="-Werror"
- - CABALVER=1.18 GHCVER=7.8.4 GHCOPTS="-Werror" # see note on Alex/Happy
- - CABALVER=1.22 GHCVER=7.10.1 GHCOPTS=""
-# - GHCVER=head  # see section about GHC HEAD snapshots
+matrix:
+ include:
+  - env: CABALVER=1.16 GHCVER=7.4.2 GHCOPTS="-Werror"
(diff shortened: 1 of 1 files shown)


jgm:
Thanks.  I'm assuming the failed ghc 7.10 build on travis was just a problem on their end.

timtylin:
Those GHC 7.10 builds always seem to take extra-long. Any idea why? Memory bottleneck? Travis has a soft time-out once the build takes longer than 50 minutes.

I did get the 7.10 build to pass on my repo, [but it took over 1 hour!](https://travis-ci.org/timtylin/scholdoc/builds/59995947).

Looking at the build history, the 7.10 build sometimes take up to 45min. Assuming that the container build environments are a little less powerful (presumably they are able to cram more builds onto the same infrastructure), that would be enough to push it over the time limit.

I will keep looking for ways to speed up the Travis builds. Perhaps multi-threading, the aforementioned persisting, etc...

FWIW, if we want to temporarily allow the 7.10.1 to fail (while still marking the overall build as "passed"), we can specify that with the following under the `matrix:` key

```
  allow_failures:
    - env: CABALVER=1.22 GHCVER=7.10.1 GHCOPTS=""
  fast_finish: true
```

jgm:
One possibility would be to disable optimizations on ghc 7.10 builds.

It's possible that the bottleneck is texmath -- see an issue in
jgm/texmath.

emwap:
I have seen issues where ghc 7.10 takes longer to build my code with more parallel threads. I.e building with 8 cores (native on my machine) takes longer than building with 4 and 2. 

I have seen somewhere that the travis docker instances report 32 cores while you only get 2. 

In my repos I build with -j2 and my build times went from an hour/timeout to a few minutes

PR #2141, opened 2015-05-08
Fix image URIs in ICML output

DigitalPublishingToolkit:
InDesign expects LinkResourceURI to start with "file:" for local filenames, and won't render/link the image without.


jgm:
Travis build fails.  Can you fix?  I think you just need to update the ICML test output (`tests/writer.icml`).

DigitalPublishingToolkit:
@jgm Updated the test output!

PR #2148, opened 2015-05-11
Note on non-working YAML export to docx for author

Jmuccigr:
Added note and examples of non-working export to docx of author info in YAML.

### DIFF
--- a/README
+++ b/README
@@ -2107,6 +2107,23 @@ custom template.  For example:
     $endif$
     $endfor$
 
+Note: Currently export to docx cannot handle structured data for multiple authors or even a single author in the YAML block. So, for example, the following single-author information will not appear in docx output:
+
+    ---
+    title:  'This is the title: it contains a colon'
+    author:
+    - name: Author One
+    ---
+
+Nor will any multiple authors:
+
+    ---
+    title:  'This is the title: it contains a colon'
+    author:
+    - name: Author One
+    - name: Author Two
+    ---
+
 
 Backslash escapes
 -----------------
--- a/README
+++ b/README
@@ -2107,7 +2107,9 @@ custom template.  For example:
     $endif$
     $endfor$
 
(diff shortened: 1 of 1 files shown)


Jmuccigr:
Any thoughts on this?

tarleb:
A brief comment unrelated to the contents: the first line in the added text is very long.  To ensure readability of the Markdown source even in small windows, a line should not be 80 characters or less.

I cannot say much on the contents, as I'm not very familiar with pandoc-templates.  My guess is that there may be other formats with similar restrictions. If so, the paragraph could be generalized, noting docx as a specific example.

Jmuccigr:
I've added line breaks to that long line.

I too don't know about the other formats, but docx seems like a popular enough one that a warning is warranted.

tarleb:
Hey @Jmuccigr, sorry for the long wait.  I had another look and noticed this paragraph right above your proposed addition:

> Note: The `author` variable in the default templates expects a simple list or
> string.  To use the structured authors in the example, you would need a
> custom template.

This seems to describe the same issue the proposed change is addressing, so I'm not sure the examples of something *not* possible are necessary.

Personally I frequently find negative examples confusing.  It's not uncommon to copy-paste sample code from the documentation without reading or understanding the surrounding text, so non-functional code could add to people's confusion.

Jmuccigr:
I found it confusing that that note is separate from the presentation of the example. Perhaps they could simply be closer together?

tarleb:
What do you think about a change like [this](https://github.com/tarleb/pandoc/commit/652f9475cc4e91617ac3487dec3da09afb7fcce0)?  Would that help to clear things up?  (Sidenote: criticism of my English is always welcome, I'm not a native speaker.)

Jmuccigr:
Looks good. Thanks for working on this.

PR #2211, opened 2015-06-04
Adding SVG support for PDF & DOCX

Twinside:
This pull request adds the following features:
- Convert SVG to PDF for PDF LaTeX output
- Convert SVG to PNG for Docx output
- Retrieve DPI information from Juicy.Pixels

Introduced dependencies:
- rasterific-svg (rendering module)
- svg-tree svg description types

By transitivity, Lens is required by svg-tree.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ install:
  - cabal-$CABALVER update
 # - git clone https://github.com/jgm/pandoc-types && cd pandoc-types && cabal-1.18 install && cd ..
  - cabal-$CABALVER install "Cabal >= 1.18" # -v2 provides useful information for debugging
+ - cabal-$CABALVER sandbox init --sandbox=tsandbox
  - cabal-$CABALVER install $JOPTS --only-dependencies --enable-tests
 
 # Here starts the actual work to be performed for the package under test; any command which exits with a non-zero exit code causes the build to fail.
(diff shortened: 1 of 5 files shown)


jgm:
This seems to have problems compiling with older versions of ghc.
I'm also a bit worried about the lens dependency, which is big.
But the feature is definitely a good one.  Can you see what is going wrong with older ghc?

Twinside:
I think it's easily fixable for GHC 7.6, do you need 7.4?

jgm:
We could probably stop supporting 7.4, now that 7.6 is in
debian stable.

+++ Vincent [Aug 17 15 00:28 ]:

Twinside:
This was tougher than expected, in order to compile:

  * GHC 7.4 is dropped from travis
  * the travis build now uses a sandbox, allowing for an updated `Cabal >= 1.18` (the library)

The use of a more up to date Cabal library is dictated by an more recent `unix` package stemming from a dependency and entering in a [known bug](http://stackoverflow.com/questions/30687926/should-ghc-be-linking-different-versions-of-the-same-libraries) of double linking for the `Setup.hs` file. The bug can be work-arounded by bumping the cabal version and ensure that a recent Cabal version is installed.

Don't hesitate to ask for clarification, as the build passing on GHC 7.6.3 is a work of painful trial and  error, I may have forgotten to clarify some points.

dhimmel:
Is this PR still alive? The ability for SVG images to render in DOCX of PDF exports would be huge!

atejeda:
Any update on this?

mb21:
@Twinside sorry for the hold-up, I guess @jgm couldn't make up his mind whether to merge in the `JuicyPixels` lib with its dependencies.

But seems like quite a bit of this code could be reused for https://github.com/jgm/pandoc/issues/4058#issuecomment-424613868 (i.e. using the `rsvg2pdf` commandline program instead of the `JuicyPixels` library). I can take a look, or do you want to?

Twinside:
Do what you want, I don't care

wilx:
This still seems like worthy addition...

PR #2216, opened 2015-06-09
replace old url with the new pandoc.org

ousia:
@jgm, I think it may be interesting to replace the old <http://johnmacfarlane.net/pandoc> references with the new <http://pandoc.org>.

I ignored both `relann1.6"  and `changelog`.

The reason not to replace the url in `changelog` was a historical one. At that time, the code was moved to <http://johnmacfarlane.net/pandoc>.

With `relann1.6`, I wonder whether it makes sense to have in the root directory and announcement for version 1.6.

I hope it helps.


jgm:
It's a good idea (and I just removed relann1.6), but this PR contains a bunch of stuff besides the URL change.  Can you do a new PR with just the URL change?

ousia:
Sorry, I didn’t know my pull request contained such a mess.

Closing this request and trying to provide a new one with that change.

PR #2217, opened 2015-06-09
replace old url with pandoc.org

ousia:
@jgm, here you have the new request.

BTW, could you check jgm/pandoc-templates#101?


jgm:
I will get to the other PR in time.

PR #2229, opened 2015-06-13
Fix #2225 - Plus(+) and Minus(-) are not encoded when writing textile

### LINKED ISSUE #2225: textile output does not properly escape + and - characters
Converting to textile does not properly escape + and - characters which means that they can get interpreted as formatting. + should probably be encoded to &amp;#<i></i>43; and - to &amp;#<i></i>45;

Testcase:

echo "&lt;p&gt;&amp;#<i></i>43;x+ &amp;#<i></i>45;y-&lt;/p&gt;" | pandoc --from=html --to=textile | pandoc --from=textile --to=html
&lt;p&gt;&lt;em&gt;x&lt;/em&gt; &lt;del&gt;x&lt;/del&gt;&lt;/p&gt;

daphnediane:
Fixes issue #2225


jgm:
Fixed in 7df4d86006c42db3e2cc68c09993f7f83cca7107
I guess I fixed it before noticing your PR, sorry!

daphnediane:
No worries.

PR #2250, opened 2015-06-29
Fix RST reference names with special characters

PromyLOPh:
The reference ``A-B_`` had the name ``ABB`` prior to this change, because the result of ``lookAhead`` was used.


jgm:
Thanks!

Issue #1924, opened 2015-02-05
Better output formatting of forced line breaks


timtylin:
This issues tracks the proposal to add a newline character `\n` after the associated `LineBreak` control character for certain writers, as per [the discussion on pandoc-discuss](https://groups.google.com/forum/#!topic/pandoc-discuss/qNJO_2R-t5s).

The motivation for this is better output formatting of things written in markdown line-blocks, such as poetry, addresses, and algorithms. Since line-blocks do not have its own Block-level element, right now all lines from a line-block are simply concatenated with `LineBreak`. This change would make some outputs with large amount of line-block content much easier to inspect.

After looking through all the writers we have, the following writers stand out as candidates for this change:

- Dokuwiki
- HTML
- EPUB (via HTML)
- LaTeX
- MediaWiki
- OpenDocument
- Texinfo

hftf:
I found what was breaking my program after installing it and Pandoc on a new computer. According to the [1.14 changelog](http://pandoc.org/releases.html#behavior-changes):
> Append newline to the LineBreak in Dokuwiki, HTML, EPUB, LaTeX, MediaWiki, OpenDocument, Texinfo writers (#1924, Tim Lin).

In 1.13:
```
$ echo "[Para [LineBreak]]" | pandoc -f native | tee foo
<p><br /></p>
$ xpath foo "p[not(text())]"
Found 1 nodes:
-- NODE --
<p><br /></p>
```

In 1.14:
```
$ echo "[Para [LineBreak]]" | pandoc -f native | tee foo
<p><br />
</p>
$ xpath foo "p[not(text())]"
No nodes found
```


I don't necessarily disagree with this change; I just found it very surprising.

PR #2255, opened 2015-07-01
Fix #2254 : OpenDocument writer adds space with hard line break

### LINKED ISSUE #2254: Hard line break adds blank space in odt
When converting from markdown to odt using the command `pandoc -f markdown -t odt -o foo.odt bar.md` in pandoc 1.14.04 (and 1.14.1), any hard line breaks in the markdown source are written in odt as a carriage return plus a space. This means there will be a space at the beginning of the second line when there shouldn't be.

Input used:
```markdown
Hard break using backslash escape:

This is line one.\
This is line two.

Hard break using two spaces at end of line:

This is line one.  
This is line two.
```
Output given:

    Hard break using backslash escape:
    
    This is line one.
     This is line two.
    
    Hard break using two spaces at end of line:
    
    This is line one.
     This is line two.

Output expected:

    Hard break using backslash escape:
    
    This is line one.
    This is line two.
    
    Hard break using two spaces at end of line:
    
    This is line one.
    This is line two.

I've tracked down the reason this happens. The OpenDocument writer generates the following from the input:

```xml
<text:p text:style-name="Text_20_body">Hard break using backslash
escape:</text:p>
<text:p text:style-name="Text_20_body">This is line
one.<text:line-break />
This is line two.</text:p>
<text:p text:style-name="Text_20_body">Hard break using two spaces at
end of line:</text:p>
<text:p text:style-name="Text_20_body">This is line
one.<text:line-break />
This is line two.</text:p>
```
The ODT writer then parses the carriage return after `<text:line-break />` as a space, so the ODT document generated from the input looks like this:

```xml
<text:p text:style-name="Text_20_body">Hard break using backslash escape:</text:p>
<text:p text:style-name="Text_20_body">This is line one.<text:line-break/> This is line two.</text:p>
<text:p text:style-name="Text_20_body">Hard break using two spaces at end of line:</text:p>
<text:p text:style-name="Text_20_body">This is line one.<text:line-break/> This is line two.</text:p>
```
The space between `<text:line-break/>` and `This is line two.` is the reason for the added space at the beginning of the line. In the `OpenDocument.hs` writer, the line break is written using:

```haskell
 | LineBreak <- ils = return $ selfClosingTag "text:line-break" [] <> cr
```
Removing `<> cr` fixes the issue. Didn't make a pull request because I'm not sure if this will break anything else the OpenDocument writer is used for or if it breaks the OpenDocument standard.

mchladek:
Fixes #2254 Removes the carriage return after a hard line break that causes the ODT writer to add a space at the beginning of the new line. Also changed the test, `tests/writer.opendocument`, to reflect this change.

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -378,7 +378,7 @@ inlineToOpenDocument :: WriterOptions -> Inline -> State WriterState Doc
 inlineToOpenDocument o ils
     | Space         <- ils = inTextStyle space
     | Span _ xs     <- ils = inlinesToOpenDocument o xs
-    | LineBreak     <- ils = return $ selfClosingTag "text:line-break" [] <> cr
+    | LineBreak     <- ils = return $ selfClosingTag "text:line-break" []
     | Str         s <- ils = inTextStyle $ handleSpaces $ escapeStringForXML s
     | Emph        l <- ils = withTextStyle Italic $ inlinesToOpenDocument o l
     | Strong      l <- ils = withTextStyle Bold   $ inlinesToOpenDocument o l
--- a/tests/writer.opendocument
+++ b/tests/writer.opendocument
@@ -896,8 +896,7 @@ of a paragraph looked like a list item.</text:p>
 <text:p text:style-name="Text_20_body">Here’s one with a bullet. *
 criminey.</text:p>
 <text:p text:style-name="Text_20_body">There should be a hard line
-break<text:line-break />
-here.</text:p>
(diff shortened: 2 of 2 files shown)


hftf:
This seems to be caused by #1924 (issue) / #1925 (PR). It might be worth looking if the recently appended newlines affected other writers in the same way.

mchladek:
HTML, EPUB, and LaTeX all seem fine. It looks like it's unique to the ODT format that a carriage return after a line break is interpreted as a space. I'm not sure about Dokuwiki, MediaWiki, or Texinfo, though, as I don't do any work in those formats.

PR #2266, opened 2015-07-03
RST: Support inline markup for field list names

PromyLOPh:
I’m not entirely sure if/how interpreted text (\`abc\`:emph:) is supposed to work in this context though.

### DIFF
--- a/src/Text/Pandoc/Readers/RST.hs
+++ b/src/Text/Pandoc/Readers/RST.hs
@@ -209,7 +209,7 @@ rawFieldListItem minIndent = try $ do
 fieldListItem :: Int -> RSTParser (Inlines, [Blocks])
 fieldListItem minIndent = try $ do
   (name, raw) <- rawFieldListItem minIndent
-  let term = B.str name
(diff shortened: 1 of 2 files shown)


jgm:
+++ PromyLOPh [Jul 03 15 07:59 ]:
>   I’m not entirely sure if/how interpreted text (`abc`:emph:) is supposed
>   to work in this context though.

What's your question, exactly?  If you're wondering about
how RST is supposed to behave, can you use docutils to test?
Or is it a pandoc question?

PromyLOPh:
I’m not sure which one is correct: Specification or reference
implementation. The first one says inline markup (and thus interpreted
text, I suppose) should work in field list names.  But the current
docutils HEAD is unable to render the following document as well.
Without `foo`:custom:, so just the inline emphasis, docutils processes
the document as expected and this pull requests adds that behavior to
pandoc.

--snip--
Title
=====

some paragraph

.. role:: custom

:Version: this
:*Date* `foo`:custom:: and that
--snap--

PR #2302, opened 2015-07-15
Candidate odt reader for review

### LINKED ISSUE #1768: Request: ODT reader
Since Libre/Open Office docbook export function si seriously broken, there is no easy way to convert ODT into md/asciidoc/name_a_format apart from rest since there is a odt2sphinx converter (that is far from perfect BTW). Actually I am converting ODT to Asciidoc in this way:

* odt->rest (odt2sphinx)
* rest->asciidoc (pandoc)

with many lost things in the process that I have to re-add manually... :-(

TIA

MarLinn:
A partial odt reader with the most basic features.
I feel the branch is now ready to be merged.
As it's only a partial implementation, this reader might or might not be enough to resolve #1768 (the problems there seem to require support for images, which is still missing for now)

# Fully implemented features

* Basic styling (emphasis, bold font, strikeout, subscript and superscript, but no smallcaps; underlined text is interpreted as being emphasised)
* Paragraphs
* Headers
* Unordered lists
* Ordered lists
* External links
* Internal links (with the help of injected anchors)
* Footnotes and Endnotes (as general *Note*s)
* Blockquotes (everything indented more than 5mm)

#  Partly implemented features

* Citations (only ids for linking are kept)
* Tables (without headers, without sizing, and with limited styling)


MarLinn:
I just updated everything and rebased again. No new code, but a lot of cleanup. Now ready to merge! Travis complains, but that seems to be a glitch. Something with different cabal versions.

jgm:
Great! I had no trouble compiling it, and it did fine on a couple of tests. I haven't had a chance to look much at the code (though I must say, I'm daunted by all those arrows -- and somewhat worried that, if I ever need to maintain this, I'll have trouble following the code).

Question:  Are you using the MediaBag interface, as we do with docx, to handle media in the ODT container?  That's a good thing to do for consistency.

MarLinn:
You can count on me for maintainance. The choice to use arrows was in fact in part a reaction to the docx reader code. I wanted something that would be more readable and easier to change by pushing as much as possible into the arrows. I may not have succeeded though. There are some major points I'm not satisfied with myself and would love to do some more refactoring with to increase understandability, but it will take some time before I can get to that. At least until then, maintaining it is my responsibility.

In the long term, I think a shared base for xml based readers might be a good idea to reduce the general maintainance burden, akin to `Pandoc.Text.Parsing` for plaintext based readers. I offer the generic parts of mine as a candidate for that, although I know they're far from perfect. And if some other such base is created, I will do my best to port my code.

As to your question: I'm currently not handling media content at all. Images are just ignored. I know this is probably a major feature for many people so it's on the agenda, it just wasn't a priority for me. So I'm not using MediaBag at all for now. But at the two interfaces where the reader is prepared for this feature (testing code and the main entry function), MediaBag is indeed in place to be used in the future.

jgm:
OK, I just merged this.  Still some trepidation about the
arrows and the long symbolic combinators.  It's not my
preferred style, but the code looks very clean and
well-organized.  If you can make it more
understandable/maintainable, that will be great.

I notice in tests on ODTs produced by pandoc that code
blocks don't come through correctly.  This could be due
to idiosyncracies in pandoc's ODT writer, which leaves a
lot to be desired.  But that might be something to look at.

Handling media would be another great feature when you have
a chance.  But this is a perfectly workable reader even
without that, and worth merging in now.


+++ MarLinn [Jul 23 15 12:46 ]:

MarLinn:
Thanks for the merge! :)

I learned a lot while writing this, some of it about the error of my ways. I can see the problems with the combinator style and have some quite precise ideas for changing that. Alas, the core changes are deep within and by now there is a large body of code to be changed on top. Although most of it is easy to adapt, the volume alone makes that a bigger task. But yes, I do plan to do that. If it helps anyone maintain the code, it helps me, too.

The reader should currently treat code blocks as block quotes because odt has no semantic markup for either and this was the easiest route for now. For the same reason, the possibility for inconsistent and unexpected behavior there is high and might never be fully resolved. I hadn't thought about consistency with the writer yet though, but that's a great approach. I will look into that. If you have any details about  your test cases, that might help pin down some of the issues.

jgm:
+++ MarLinn [Jul 23 15 17:53 ]:
>   The reader should currently treat code blocks as block quotes because
>   odt has no semantic markup for either and this was the easiest route
>   for now. For the same reason, the possibility for inconsistent and
>   unexpected behavior there is high and might never be fully resolved. I
>   hadn't thought about consistency with the writer yet though, but that's
>   a great approach. I will look into that. If you have any details about
>   your test cases, that might help pin down some of the issues.

You might try just using pandoc to convert the pandoc README
to ODT, and then using pandoc to convert that back to
Markdown.  See what changed.

jgm:
One other thought. Pandoc has an opendocument writer, which just produces a text file (opendocument xml), and an ODT writer, which produces the whole ODT container.

Would it make sense to have an opendocument reader (operating on an xml plain text input) and an ODT reader?

For docx, we only have docx readers and writers, because in docx the content of the document is split between a bunch of different files (e.g. the footnotes go in a separate file).  But ODT is saner in this respect; the content is pretty much all in one file (except for media).

hftf:
I would use a non-container docx reader/writer that either omits the other files or concatenates all of them, for ease of debugging. It can get annoying unzipping and xmllinting all the time.

PR #2303, opened 2015-07-16
Add missing Heading 6 style in reference Docx

hftf:
Based on [this mailing list discussion](https://groups.google.com/forum/#!topic/pandoc-discuss/WfNT_qDevF4).

This is a rough fix. The reference `styles.xml` was simply missing a `Heading 6` style. A better solution might consider any element with a `Heading #` style as a heading regardless of what styles are in `styles.xml`.

Note that this only really affects pandoc-created Docx files. Word-created Docx files will not omit the `Heading 6` style if it is used in the document.


jgm:
Great!

PR #2312, opened 2015-07-19
set font-style to normal in emphasized emphasis (epub.css)

ousia:
@jgm, I think it would be interesting to add the following line to `data/epub.css`:

``` css
em em { font-style: normal; }
```

So emphasis inside emphasis gets normal, as it is the standard practice ([sample](http://pandoc.org/try/?text=_emphasis+*emphasized+again*+should+be+upright%2C+not+in+italics_&from=markdown&to=html)).

### DIFF
--- a/data/epub.css
+++ b/data/epub.css
@@ -13,3 +13,4 @@ h3.date { }
 ol.toc { padding: 0; margin-left: 1em; }
 ol.toc li { list-style-type: none; margin: 0; padding: 0; }
 a.footnoteRef { vertical-align: super; }
+em em { font-style: normal; }


hftf:
With this change, there would be another reason to fix #2264.

Underlined emphasized text is sometimes [transformed](http://pandoc.org/try/?text=%2B_foo_%2B&from=textile&to=html) to emphasized emphasized text:
```
$ echo "+_foo_+ bar" | pandoc -f textile -t html
<p><em><em>foo</em></em> bar</p>
```

Users expecting to see a difference between `foo` and `bar` when using epub would be surprised that it was as if nothing had happened.

jgm:
What happens, with this change, if you have em-em-em?  Does it correctly show italics again?

hftf:
```
em em,
em em em em {
    font-style: normal;
}

em,
em em em,
em em em em em {
    font-style: italic;
}
```
is usually what I've [seen](https://github.com/search?utf8=%E2%9C%93&q=%22em+em+em%22+%22font-style%22+%22normal%22&type=Code&ref=searchresults).

ousia:
> What happens, with this change, if you have em-em-em? Does it correctly show italics again?

@jgm, it has to be enabled in the CSS file.

Do you want me to adapt the merge request to what @hftf has shown?

BTW, how many levels do you want? (I have only seen emphasis inside emphasis; that is, two levels.)

jgm:
Yes, let's fold in @hftf's proposal.

hftf:
I wonder if we should add similar styles to the default HTML template, or to other Writers.

hftf:
[FYI](https://help.github.com/articles/using-pull-requests/#sending-the-pull-request):

> After your pull request is sent, any new commits pushed to your branch will automatically be added to the pull request. This is especially useful if you need to make more changes.

ousia:
@hftf, after merging upstream/master, I would have also committed other changes from that merge.

I may be missing something, but I removed my fork and made a new one. From that fork, I made a clean merge request (#2352).

Issue #2048, opened 2015-04-01
Documenting usage of docx styles


jo-tud:
Even though paragraphs in a word docx document are formatted with a stlye named _Verbatim Char_ this is not reflected in the markdown document.
According to the documentation this should work, right?

lierdakil:
I'm confused. Are you talking about docx reader?

jo-tud:
Yes - is this the wrong place for docx reader?

Am 01.04.2015 21:37 schrieb Nikolay Yakimov <notifications@github.com>:

I'm confused. Are you talking about docx reader?

lierdakil:
No, it's the right place. Probably because I was working on docx *writer* last couple weeks, I became a little confused with your report is all.

There is a multitude of reasons why this might not work. Two most obvious are 

1. `Verbatim Char` should be a character style, and not paragraph style.
2. Word can sometimes mangle style identifiers, and docx reader does not handle all cases of that well (yet).

If you can't resolve this, consider sharing a minimal example docx so that we could look and see what went wrong. You can upload it somewhere, or e-mail me directly (see github profile).

jo-tud:
Thank you, it is working now. You were right, I had overlooked the fact that it must be a character style. While googling I stumbled upon the  *pandoc/tests/docx/codeblock.docx* file and saw that it is possible to have a style named *Source Code* as a paragraph style. This fit my purpose even more - but I think it is not yet reflected in the documentation.

jgm:
We should add something to the documentation about the SourceCode style,
and how it is added automatically.

lierdakil:
In all honesty, we should properly document use of docx styles by reader
*and* writer, because right now it's... not the best documentation I've
seen.
02 апр. 2015 г. 19:55 пользователь "John MacFarlane" <
notifications@github.com> написал:

mb21:
Fixed through https://github.com/jgm/pandoc/pull/2318

PR #2314, opened 2015-07-20
List all styles in manual for --reference-docx

infotroph:
I'm working on the assumption that styles.xml is the canonical list of styles known to the docx writer; if this is wrong, please advise. 

Possibly related to #2048, but only a minor subset of it.

### DIFF
--- a/README
+++ b/README
@@ -636,12 +636,13 @@ Options affecting specific writers
     for a file `reference.docx` in the user data directory (see
     `--data-dir`). If this is not found either, sensible defaults will be
     used. The following styles are used by pandoc: [paragraph]
-    Normal, Compact, Title, Subtitle, Authors, Date, Abstract, Heading 1,
-    Heading 2, Heading 3, Heading 4, Heading 5, Block Text, Definition Term,
-    Definition, Bibliography, Body Text, Table Caption, Image Caption,
-    Figure, FigureWithCaption;
+    Normal, Body Text, First Paragraph, Compact, Title, Subtitle, Author, Date,
+    Abstract, Bibliography, Heading 1, Heading 2, Heading 3, Heading 4,
+    Heading 5, Heading 6, Block Text, Footnote Text, Definition Term,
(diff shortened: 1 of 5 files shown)


jgm:
Thanks very much!  However, pandoc.1 shouldn't be modified by hand.  Instead, please modify README.  You can then generate pandoc.1 from README using `make man/pandoc.1`.  Note also that lines should be hard-wrapped as they are currently in README.

infotroph:
Sure thing, but let me make sure I'm following correctly: I should update the PR with the resulting changed versions of both README and pandoc.1, correct? Or do I edit README, test with `make`, and only commit the changes to README?

jgm:
It would be enough to edit README, since I'll update the man
page before the next release.

But if you like, you can regenerate man/pandoc.1 and commit it
too.  (We include this in the repository because adding it
to the build process in a way that works across all
platforms proved frustrating.)

+++ Chris Black [Jul 21 15 12:26 ]:

infotroph:
Here's the edited README. I took a short stab at rebuilding man/pandoc.1, but got errors that look to me like they come from issues with my (freshly-installed, untested) Haskell environment:

```
> make man/pandoc.1
pandoc README -t man -s --template man/pandoc.1.template \
		--filter man/capitalizeHeaders.hs \
		--filter man/removeNotes.hs \
		--filter man/removeLinks.hs \
		-o man/pandoc.1
pandoc: Error running filter man/capitalizeHeaders.hs
fd:4: hPutBuf: resource vanished (Broken pipe)
make: *** [man/pandoc.1] Error 83
```

I'm happy to keep trying on this and update when I have a working build, or if you prefer you can merge now. Sorry for the messy branch!

jgm:
Can you submit a PR with just the README changes, so I can
cleanly merge it?

+++ Chris Black [Jul 21 15 14:38 ]:

infotroph:
Clean version, with rebuilt man page, at #2318. Thanks for the tips and sorry for the extra noise!

PR #2318, opened 2015-07-21
List all styles in manual for --reference-docx

infotroph:
Cleaned-up version of #2314, with working regenerated man page.


jgm:
Great, thanks!

PR #2333, opened 2015-07-28
handle ikiwiki wikilinks

csrhodes:
rewrites them to be standard links to an id with name generated from
the textual form of the wikilink.

This is problematic in that the exact form of the
wikilink (e.g. relative vs absolute) will give a different resulting
name; some consistency will be required in authoring to make
everything work.  However, it works well enough to turn multiple
ikiwiki pages into a single document with internal links.


jgm:
Adding any kind of syntax extension requires more than this.
We'd need it to be hooked into the system for extensions (which can be enabled/disabled one by one).

csrhodes:
I've hooked it in to the extensions, disabled by default, and minimally documented it.

Issue #2354, opened 2015-08-08
Table of contents not linked when generating html from org file


jg:
I'm generating an html file from my org file. I'd like to include a table of contents.

    pandoc -s -S --toc file.org -o file.html

The org file has three simple first level headings.

```
* item 1
* item 2
* item 3
```

Unfortunately the output table of contents is not linked. The links are missing their href attributes.

```
<div id="TOC">
<ul>
  <li><a>item 1</a></li>
  <li><a>item 2</a></li>
  <li><a>item 3</a></li>
</ul>
```

jgm:
@tarleb, what do you think?  The trick is to use `registerHeader`, which does all the work behind the scenes, but you need to define certain instances for OrgParserState for this to work.

tarleb:
Apologies for the late reply.  I didn't find the time yet for this, but I keep it in mind and will take a look when things calm down a little.  It's on my todo list.

jg:
Do we care about ghc 7.10.1 compatibility? CONTRIBUTING.md could use a pragraph about that

jgm:
I added a note about ghc versions and travis.

Btw, travis builds currently fail for ghc 7.10 due to a
cabal version mismatch in running tests, which makes 'cabal
test' fail.  I think this is a cabal bug; do you know how to
make it go away?

+++ jg [Aug 15 15 00:46 ]:

jgm:
Oh, and the answer is yes - we want ghc 7.10.x
compatibility!  Only, for now, it doesn't need to be
-Wall clean -- it's hard to know how to avoid all the
warnings relating to moving Monoid and Applicative stuff
into the Prelude.

+++ jg [Aug 15 15 00:46 ]:

jgm:
Closed by #2360

jgm:
> Btw, travis builds currently fail for ghc 7.10 due to a
cabal version mismatch in running tests, which makes 'cabal
test' fail.  I think this is a cabal bug; do you know how to
make it go away?

Fixed this by using cabal-install-head in the travis build for ghc 7.10.1.

PR #2360, opened 2015-08-15
Org reader: add auto identifiers if not present on headers

jg:
Refs #2354

This should also fix the table of contents (--toc) when generating a html file
from org input


jgm:
Great, thanks!

PR #2369, opened 2015-08-20
`lang` variable is now in BCP47 format

### LINKED ISSUE #1614: lang attribute fits for latex but not for html lang-attribute-value
Hi there,

for documents in german language I use

* **lang: ngerman**
* **mainlang: german**

the first beeing for babel, second for polyglossia (i use mainly xelatex). If I pandoc to html the lang-attribute (ngerman) gets used but **ngerman** is not a valid value for the html-attribute (nor would be german).

html needs something like **de** or **de-DE**.

The documentation says **lang** *language code for HTML or LaTeX documents*.

And therefore the documentation is completely wright, you could use it for the one or the other, not both at the same time. 

perhaps this needs clarification in the docs and/or a different approach. Meanwhile you can always override the yaml attributes in your central file with direct command line switches.

all the best,
christoph

mb21:
closes #1614

### DIFF
--- a/README
+++ b/README
@@ -945,7 +945,19 @@ as `title`, `author`, and `date`) as well as the following:
 :   body of document
 
 `lang`
-:   language code for HTML or LaTeX documents
+:   The `lang` variable should be set by the user to a language
+    code according to [BCP 47] (e.g. `en` or `en-GB`).
+    For some output formats, pandoc will convert it to an approriate
+    format stored in the additional variables `babel-lang`,
+    `polyglossia-lang`, `polyglossia-variant` (LaTeX)
+    and `context-lang` (ConTeXt).
+
+`otherlangs`
+:   Should be set to a list of other languages used in the document
+    in the YAML metadata, according to [BCP 47]. For example:
+    `otherlangs: [en-GB, fr]`.
+    Currently only used by XeTeX through the generated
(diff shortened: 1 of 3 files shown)


jgm:
Note: if the name is changed from `lang` (currently used in html and xml related templates) to `language`, we'll need changes to all those templates too.  Maybe it is better to keep it as `lang`?

mb21:
Yes, that's what I first thought as well, and then everyone started to talk about `language` so I thought the consensus had shifted. @ousia did a bunch of renaming in [pandoc templates pull 101](https://github.com/jgm/pandoc-templates/pull/101), but then everyone who has forked the templates still need to update theirs. So should I change it back to `lang`?

jgm:
I've commented on #1614.

mb21:
changed `language` to `lang`.

Changed ePUB documentation (behavior was already to read out `lang` meta if nothing else found). Note that [BCP 47](https://tools.ietf.org/html/bcp47) includes RFC 5646.

jgm:
Well done!

mb21:
Thanks! Glad you got around to merging it.

mb21:
btw, you should also pull [pandoc templates 114](https://github.com/jgm/pandoc-templates/pull/114)...

Issue #2394, opened 2015-09-12
Pandoc 1.14+ changes to rst/ReST headings break Jupyter/nbconvert


mscuthbert:
(Tag: more discussion needed)

Jupyter's nbconvert (formerly IPython) uses Pandoc to convert to .rst files from their syntax including markdown headers.  Before Pandoc 1.14, markdown headers in either the "===="/"----"/"###" format or the "#"/"##"/"###" format were translated properly to various levels in .rst.  From 1.14 onwards all notebook headers become top-level heading 1 headers in nbconvert.  I believe that this change in the 1.14 release notes changed the behavior:

    Normalize headings to sequential levels (Nikolay Yakimov). This is pretty much required by docutils.

Can Nikolay or someone please take the time to explain the change so that the Jupyter folks can make a fix or use an alternate template of some sort (or could pandoc add an option to use the older behavior?).  I'm new to contributing to either project (both of which I love and admire) so I'm not sure how to help with the situation except to bring it to the attention of both projects.

The relevant nbconvert issue is at https://github.com/jupyter/nbconvert/issues/97

Thank you!

mpickering:
@lierdakil

minrk:
The gist is that nbconvert passes snippets of markdown one at a time to pandoc, so pandoc may get

```markdown
# header 1
some text
```

in one call, then

```markdown
### header 3
```

some time later. What nbconvert needs from pandoc is consistent heading markers for a given input heading level. Is there a way to get this from pandoc ≥ 1.14?

jgm:
I see the problem, and I don't see a workaround.  For
background on the change, see the PR #2079 which links to a
pandoc-discuss discussion.  I'm open to discussion as to how
to improve the situation.

minrk:
@jgm would it be possible to add an option to tell pandoc to consistently render h1 -> `====`, etc. rather than normalizing?

jgm:
@minrk I think we need to normalize in some way, for valid reST.
However, we might be able to do something like this:  before normalizing, detect the top-most header level.  Then, after normalizing, increment all the header levels in the normalized document so that the top-most header level matches what was there before.
@lierdakil - would that still address the issue that motivated this change?

lierdakil:
Sorry, missed notifications. Let me check.

lierdakil:
[...]

PR #2405, opened 2015-09-19
[RST Writer] Don't normalize heading levels below input minimum

lierdakil:
Hopefully helps with #2394

A test included.

### DIFF
--- a/src/Text/Pandoc/Writers/RST.hs
+++ b/src/Text/Pandoc/Writers/RST.hs
@@ -81,7 +81,8 @@ pandocToRST (Pandoc meta blocks) = do
                 (fmap (render colwidth) . blockListToRST)
                 (fmap (trimr . render colwidth) . inlineListToRST)
                 $ deleteMeta "title" $ deleteMeta "subtitle" meta
-  body <- blockListToRST' True $ normalizeHeadings 1 blocks
+  let minLev = findMinHeadingLevel Nothing blocks
+  body <- blockListToRST' True $ normalizeHeadings minLev blocks
   notes <- liftM (reverse . stNotes) get >>= notesToRST
   -- note that the notes may contain refs, so we do them first
   refs <- liftM (reverse . stLinks) get >>= refsToRST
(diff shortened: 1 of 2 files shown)


lierdakil:
@mscuthbert, could you possibly test this fork? I'm especially interested in how would it handle inconsistent heading levels, like this:

```markdown

## Heading 2

text

##### Heading 5

text

### Heading 3

text
```

mscuthbert:
Hi @lierdakil -- thank you! Unfortunately I can't test this because I'm on the El Capitan beta and the rootless problems on building cabal projects (https://github.com/haskell/cabal/issues/2653) are preventing the building from source.  Since I've never developed a Haskell project (didn't know what Cabal was until this download) it's probably going to be beyond my abilities to make a test w/o a binary.

jgm:
You could try building with the new stack tool.
(That may or may not run into the same problem.)
https://github.com/commercialhaskell/stack

+++ Michael Scott Cuthbert [Sep 19 15 12:09 ]:

lierdakil:
Stack will run into the same problem, since this is ghc 7.8 issue. @mscuthbert, you *could* get your hands on ghc 7.10, which, according to [this](https://ghc.haskell.org/trac/ghc/blog/weekly20150721#MacOSXElCapitansupport), should work.

I'd be happy to provide a binary, but I have no access to OSX environment whatsoever, so that would be a bit complicated.

jgm:
I compiled and tested.  Here's the reST output for the input you
provided:

```
Heading 2
---------

text

Heading 5
~~~~~~~~~

text

Heading 3
~~~~~~~~~

text
```

+++ Nikolay Yakimov [Sep 19 15 07:50 ]:

lierdakil:
Thanks, @jgm, but that's something I could've done myself, and it's exactly the expected output. What I'm interested in is how this interacts with nbconvert, and that's not someting I can readily test.  As far as I gather, nbconvert splits input text into sections and feeds those to pandoc. I'm quite positive this input won't produce intuitively expected output in this case...

PR #2412, opened 2015-09-23
Added support for <xref> tag in DocBook reader

frerich:
This patch adds proper support for the DocBook <xref> element which is used to create cross references to other parts of the document - see http://www.docbook.org/tdg/en/html/xref.html.

I didn't see an existing infrastructure to lookup elements by their ID attribute so I introduced a `dbContent` field to the state which can be used to access the DocBook file being read. As it is, the tree is searched for some element matching a given `id` every time an `<xref>` element is found. If this turns out to be too much of a slowdown for documents which make use of a lot of cross references, some simple cache (`M.Map String Element` or so) might help.

The same logic for referencing other elements would probably be useful for similiar elements, e.g. `<link>`.


frerich:
By now, I suspect that I should augment commit 76af74b such that the parsed href is prepended with a hash character (`#`). It appears this is conventional, and at least the HTML output seems to rely on it.

Still, it feels wrong to do it in a reader - but I guess right now there is no way to distinguish cross-references from references to other files in the native model...

jgm:
+++ Frerich Raabe [Sep 24 15 07:23 ]:
>   By now, I suspect that I should augment commit [1]76af74b such that the
>   parsed href is prepended with a hash character (#). It appears this is
>   conventional, and at least the HTML output seems to rely on it.
>
>   Still, it feels wrong to do it in a reader - but I guess right now
>   there is no way to distinguish cross-references from references to
>   other files in the native model...

That's correct.

frerich:
Updated the pull request such that the references have a `#` character prepended.

jgm:
Thanks!

PR #2418, opened 2015-09-26
Give details of locator terms

nickbart1980:


jgm:
I'll tweak this a bit here, and add something to pandoc-citeproc's man page too.

PR #2420, opened 2015-09-27
Update KaTeX JS and CSS versions.

xymostech:
Update the default KaTeX JS/CSS links to the current version. KaTeX v0.5.1 has far more functions and symbols than v0.1.0, so it seems like a better default. I think technically this might break compatibility because we released a breaking change due to the greediness of the `\color` function, but this probably has very little impact.

I have not tested or linted this, but it shouldn't affect the build. I'm not sure if the tests will be affected by this change, I can run them if that is necessary.

### DIFF
--- a/pandoc.hs
+++ b/pandoc.hs
@@ -845,7 +845,7 @@ options =
                   (\arg opt ->
                       return opt
                         { optKaTeXJS =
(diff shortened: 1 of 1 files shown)


jgm:
Thanks

PR #2432, opened 2015-10-04
Docx Reader: Parse soft and non-breaking hyphen elements

### LINKED ISSUE #2415: Conditional breaks are lost when converting from docx
If I have a docx document which contains conditional breaks (press `Ctrl + -` in MS Word 2013), those are lost when converting to markdown or to HTML.

`pandoc -f docx -t html -o out.html input.docx`

I expected that the output contains `&shy;` entities. Instead, nothing at all was added for the conditional breaks, they are just beeing ignored.

hftf:
Fixes #2415.


jgm:
Great, thanks.

hftf:
I think the Docx Writer should get the same feature, but I’m not sure how to code it.

hftf:
> I think the Docx Writer should get the same feature, but I’m not sure how to code it.

Partially addressed for soft hyphens in #3691. (Commenting just as an FYI to link the two issues.)

PR #2435, opened 2015-10-05
Revise variables discussion in README.

adunning:
This adds a few headers to the variables section to separate the sections for various formats and editing templates, and fixes a few other minor details.

Note that there is still some inconsistency in the file in referring to the program as Pandoc or pandoc.

### DIFF
--- a/README
+++ b/README
@@ -124,12 +124,8 @@ will only be included if you use the `-s/--standalone` option.
 Creating a PDF
 --------------
 
-Earlier versions of pandoc came with a program, `markdown2pdf`, that
-used pandoc and pdflatex to produce a PDF.  This is no longer needed,
-since `pandoc` can now produce `pdf` output itself. To produce a PDF, simply
-specify an output file with a `.pdf` extension. Pandoc will create a latex
-file and use pdflatex (or another engine, see `--latex-engine`) to convert it
-to PDF:
+To produce a PDF, specify an output file with a `.pdf` extension.
+Pandoc will use LaTeX to convert it to PDF:
 
     pandoc test.txt -o test.pdf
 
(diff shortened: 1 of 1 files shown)


jgm:
Great - I have a couple tweaks but will make these myself after merging.

adunning:
Thanks for fixing that; I had originally put `polyglossia` at the end of the sentence, but the punctuation was too messy, and I oversimplified.

mb21:
Not sure I agree with the following change:

         Setting the `dir` variable enables bidirectional text
    -    handling in LaTeX and ConTeXt, thus even if the base direction
    -    is the default left-to-right, you should set `dir: ltr` for
    -    documents that also contain some right-to-left script.
    +    handling in LaTeX and ConTeXt: set `dir: ltr` for
    +    documents that also contain some right-to-left script,
    +    even if the base direction is the default left-to-right.

Neither version is particularly easy to read. But upon first reading the second version, I parsed it as "set `dir: ltr` ... even if the base direction is ... left-to-right.", which doesn't make too much sense.

adunning:
I see what you mean. What about this?

```markdown
LaTeX and ConTeXt assume by default that all text is left-to-right.
Setting `dir: ltr` enables bidirectional text handling in a document
whose base direction is left-to-right but contains some right-to-left script.
When using LaTeX for bidirectional documents, only the `xelatex` engine
is fully supported (see `--latex-engine`).
```

mb21:
Yes, that's better :)

Maybe we should put the last sentence in its own paragraph since it's quite important. And make it more explicit:

    When using LaTeX for bidirectional documents, only the `xelatex` engine
    is fully supported (i.e. make sure to include `--latex-engine xelatex`).

adunning:
Should I do that as a new pull request, then? It has also occurred to me that it might make it easier to consult the documentation if we included links to the LaTeX packages to which the readme refers.

adunning:
By the way, am I right in thinking that 'Unicode Bidirectional Algorithm' in the preceding paragraph is supposed to link to <http://www.w3.org/International/articles/inline-bidi-markup/uba-basics>?

jgm:
Yes, a new PR would be welcome.

+++ Andrew Dunning [Oct 07 15 09:29 ]:

mb21:
> It has also occurred to me that it might make it easier to consult the documentation if we included links to the LaTeX packages to which the readme refers.

Good idea, there are some links in the comments in the code in the LaTeX writer, but why not include them in the README as well.

> By the way, am I right in thinking that 'Unicode Bidirectional Algorithm' in the preceding paragraph is supposed to link to http://www.w3.org/International/articles/inline-bidi-markup/uba-basics?

Yes, must have been lost when merging or something.

Thanks!

adunning:
This has been implemented in #2445.

PR #2441, opened 2015-10-07
Change variable to polyglossia-lang.name and .options

### LINKED ISSUE #2437: The variable “polyglossia-otherlangs” should have fields
Currently, no variants can be set for polyglossia’s secondary languages.

I’d suggest that the variable `polyglossia-otherlangs` should have the fields `.name` and `.variant`, and multiple `\setotherlanguage{}` commands rather than one `\setotherlanguages{}` command should be used.

Accordingly, the `default.latex` template should contain:  

~~~
$for(polyglossia-otherlangs)$
\setotherlanguage
$if(polyglossia-otherlangs.variant)$[variant=$polyglossia-otherlangs.variant$]$endif$
{$polyglossia-otherlangs.name$}
$endfor$
~~~

**Alternative: `.options` fields**

Use `polyglossia-mainlang` and `polyglossia-otherlangs` variables, both having `.name` and `.options` fields, where `.options` can have a value of `variant=…`, e.g., `variant=british`.

An `.options` field would make it easier to deal with other polyglossia options, too, e.g.,

`ar-TN` => `\setotherlanguage[locale=tunisia]{arabic}`

`de-AT-1901` => `\setotherlanguage[variant=austrian, spelling=old]{german}`

mb21:
closes #2437

depends on https://github.com/jgm/pandoc-templates/pull/129

### DIFF
--- a/README
+++ b/README
@@ -944,8 +944,7 @@ Language variables
     using a code according to [BCP 47] (e.g. `en` or `en-GB`).
     For some output formats, pandoc will convert it to an appropriate
     format stored in the additional variables `babel-lang`,
-    `polyglossia-lang`, `polyglossia-variant` (LaTeX)
-    and `context-lang` (ConTeXt).
+    `polyglossia-lang` (LaTeX) and `context-lang` (ConTeXt).
 
 `otherlangs`
 :   a list of other languages used in the document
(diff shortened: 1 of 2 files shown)


adunning:
The changes to `lang` need to be copied into the Beamer template, along with a few other things: I can take care of it once this is merged, assuming that there aren't any further changes planned.

mb21:
Good point, I think the beamer template is missing the following two changes as well: https://github.com/jgm/pandoc-templates/pull/114 and https://github.com/jgm/pandoc-templates/pull/120

jgm:
A further PR welcome for the beamer templates.

PR #2450, opened 2015-10-14
Add new LaTeX variables to README.

adunning:


README
```diff
@@ -1105,14 +1105,14 @@ Variables for slides
...
-`colortheme`, `fonttheme`, `innertheme`, `outertheme`
-:   themes for LaTeX [`beamer`] documents
-
 `controls`
-:   show controls in reveal.js slide shows
+:   disables controls in reveal.js slide shows if set to `false`
 
 `progress`
-:   show progress bar in reveal.js slide shows
+:   disables progress bar in reveal.js slide shows if set to `false`
```
adunning: Note, by the way, that the usage of the reveal.js variables is not consistent; not sure whether it would be more user-friendly to require specification of true/false (as controls, progress, center) or to switch from the default with a non-null value (as slideNumber).

PR #2460, opened 2015-10-17
Consider header files when determining whether to use csquotes

AndreasLoow:
If one wants to use csquotes for LaTeX output one currently have to modify the template, which is annoying as one then have to keep one's modified template in sync with the official version.

This pull request offers a solution to this: pandoc should also scan the included header files. Then one can use csquotes without having to modify the template.

### DIFF
--- a/README
+++ b/README
@@ -184,10 +184,10 @@ document contains images), [`color`], [`hyperref`], [`ulem`],
 [`mathspec`], [`polyglossia`] (with `lang`), [`xecjk`], and
 [`bidi`] (with the `dir` variable set). The [`upquote`] and
 [`microtype`] packages are used if available, and [`csquotes`] will
-be used for [smart punctuation] if added to the template. The
-[`natbib`], [`biblatex`], [`bibtex`], and [`biber`] packages can
-optionally be used for [citation rendering]. These are included with
-all recent versions of [TeX Live].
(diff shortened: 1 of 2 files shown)


AndreasLoow:
Strange, AppVeyor failed with:

    cabal: Error: some packages failed to install: 
    digest-0.0.1.2 failed while downloading the package. The exception was: 
    DeleteFile 
    "C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\digest-0.0.1.2-2460\\digest-0.0.1.2\\dist\\dist-sandbox-12ebb133\\build\\Data\\Digest\\Adler32_hsc_make.exe": 
    permission denied (Access is denied.) 
    pandoc-1.15.1 depends on digest-0.0.1.2 which failed to install.

jgm:
Appveyor not infrequently has random errors like this.
It's probably nothing to do with pandoc itself or your
PR.

+++ Andreas Lööw [Oct 18 15 02:50 ]:

AndreasLoow:
I noticed that this hasn't been merged yet. Is there something I should change or is the whole idea bad?

jgm:
I somewhat regret the way pandoc scans templates -- it now seems better to make everything explicit instead of trying to magically do the right thing.  But, given that decision, your modification seems perfectly appropriate.  So I'll merge this.

PR #2461, opened 2015-10-18
Fixed auto_identifiers examples

benweet:
I was trying to figure out how could `Dogs?–in my house?` become `dogs--in-my-house` in the documentation page...


jgm:
Thanks, good improvement.

PR #2472, opened 2015-10-22
Update LaTeX/ConTeXt variable usage in README.

adunning:
Accounts for changes in https://github.com/jgm/pandoc-templates/pull/141.


adunning:
(Note that the actual changes to the template that this describes still need to be merged; let me know if you would like any changes to these.)

PR #2477, opened 2015-10-25
Org reader: allow toggling header args

### LINKED ISSUE #2269: org source block header arguments break syntax highlighting
Hi, 

I searched and haven't seen this reported in other open/closed issues. Let me know if I missing something here.

I am running pandoc as follows

    pandoc test.org --from=org --to=html5 --standalone --highlight-style=espresso > test.html

test.org contains:

    #+begin_src python :noeval
    a = [i for i in range(10)]
    print(a) 
    #+end_src

    #+begin_src python
    a = [i for i in range(10)]
    print(a)
    #+end_src

Code highlighting works *only* if no header arguments are specified after language identifier. Otherwise the result has syntax highlighting.

here's a veiw of the html conversion on similar test.org
example

Is this known? What is the cause?

tarleb:
Org-mode allows to skip the argument of a code block header argument if
it's toggling a value.  Argument-less headers are now recognized,
avoiding weird parsing errors.

The fixes are not exactly pretty, but neither is the code that was
fixed.  So I guess it's about par for the course.  However, a rewrite of
the header parsing code wouldn't hurt in the long run.

Thanks to @jo-tham for filing the bug report.

This fixes #2269.

### DIFF
--- a/src/Text/Pandoc/Readers/Org.hs
+++ b/src/Text/Pandoc/Readers/Org.hs
@@ -513,10 +513,16 @@ rundocBlockClass :: String
 rundocBlockClass = rundocPrefix ++ "block"
 
 blockOption :: OrgParser (String, String)
-blockOption = try $ (,) <$> orgArgKey <*> orgParamValue
+blockOption = try $ do
+  argKey <- orgArgKey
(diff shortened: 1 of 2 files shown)


jgm:
Great!

PR #2478, opened 2015-10-25
Added de-CH-1901, fixed el-polyton

nickbart1980:
el-polyton, not el-poly, see http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry


jgm:
travis is failing with:
```
src/Text/Pandoc/Writers/LaTeX.hs:1076:1: Warning:
    Pattern match(es) are overlapped
    In an equation for ‘toPolyglossia’:
        toPolyglossia ("de" : "CH" : "1901" : _) = ...
```
This more specific pattern match needs to be above the catch-all pattern match that is currently in the previous line:
```
 toPolyglossia ("de":"CH":_)        = ("german", "variant=swiss")
```
Otherwise it will never be triggered, hence the warning.

mb21:
@nickbart1980 thanks, guess it would be nice if you could [rebase](https://git-scm.com/docs/git-rebase) your three commits into one.

    git rebase -i HEAD^3 //select "fix" for the bottom two
    git push --force

nickbart1980:
Thank you. I haven’t been working with the git cli so far though, so I submitted a new PR instead.

Issue #2508, opened 2015-11-09
Add `colorlinks` variable to LaTeX writer


adunning:
The changes to https://github.com/jgm/pandoc-templates/pull/141 depend on a change to the LaTeX writer that automatically enables the `colorlinks` variable if any of `linkcolor`, `citecolor`, `urlcolor`, or `toccolor` are non-null.

(Apologies for not submitting the change myself, but I haven't quite figured out the cleanest way to do this.)

PR #2509, opened 2015-11-09
Update LaTeX/ConTeXt link colour usage in README.

adunning:
Depends on #2508 and https://github.com/jgm/pandoc-templates/pull/141.


adunning:
Let me know if there's anything else you would like me to do with this; this describes the changes now merged into the template.

PR #2532, opened 2015-11-15
Interpret pauses correctly for all headers

### LINKED ISSUE #2530: Unexpected behavior with slides, pauses and HTML
This applies at least to outputting revealjs. It appears that using a lower than slide level header prevents pauses from being interpreted correctly underneath that header, leaving them as ellipses. This behavior seems intended looking at the code, but I can't imagine why. I've made a pull request for this issue if it should be fixed.

michaelbeaumont:
Previously, when using headers below the slide level, pauses are left
uninterpreted into pauses. In my opinion, unexpected behavior but
intentional looking at the code.

Fixes #2530

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -308,11 +308,9 @@ elementToHtml slideLevel opts (Sec level num (id',classes,keyvals) title' elemen
                    $ if titleSlide
                         -- title slides have no content of their own
                         then filter isSec elements
-                        else if slide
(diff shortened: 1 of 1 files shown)


jgm:
Can you give a simple example of where this works and the original goes wrong?

michaelbeaumont:
Using `pandoc --slide-level=2 --from=markdown --to=revealjs test.md`

test.md:
```
# Title

##Subheader

####Subsubheader

part 1

. . .

part 2
```

Maybe I'm not using pauses how they're intended, but I'm not sure why this shouldn't be possible.

Issue #2527, opened 2015-11-13
Docx writer: Optionally insert space or tab before footnote text


conklech:
Currently, the docx writer produces footnotes that contain the footnote reference (i.e. superscript number) followed immediately by the text of the note. It appears like this:

![image](https://cloud.githubusercontent.com/assets/573035/11159521/19091722-8a15-11e5-8730-b3883bddcc2b.png)

By default, inserting a footnote in Word inserts the footnote reference and then a space, so most or many Word documents normally have a single space between reference and the text. The Pandoc reader I use (org-mode) discards whitespace at the beginning of the note, so it's not possible to simply type an extra space in the source document. A literal nonbreaking space in the input produces the desired effect, at the cost of having to remember to manually insert it in every footnote.

![image](https://cloud.githubusercontent.com/assets/573035/11159524/212d9518-8a15-11e5-9fd4-7e60571ff438.png)

Better yet would be to have a tab character following the reference, so that the paragraph style may be used to align footnotes properly. I don't think there's a workaround for that—a literal tab character is ignored as simply whitespace.

![image](https://cloud.githubusercontent.com/assets/573035/11159526/268692da-8a15-11e5-8a2b-cc17bc12bf0b.png)

![image](https://cloud.githubusercontent.com/assets/573035/11159515/0b3395fa-8a15-11e5-9ebc-3886b3eb90dc.png)

The correct default behavior *might* be to insert a literal space between footnote reference and the body of the footnote. But this would change existing behavior, and would be even harder to work around. 

So I propose a flag or other option to control whether nothing, a space, or a tab is inserted after the footnote reference.

Inserting spaces could be implemented as a filter, by prepending a space to each footnote. Would that work for tabs, too?

conklech:
I confirm that Pandoc-produced footnotes now have a space between reference and text. It looks like we were already setting the `xml:space="preserve"` attribute.

I plan to try implementing the alternative tab behavior as an option. I'll submit a pull request if I get there.

PR #2536, opened 2015-11-15
Add --docx-note-tab flag to replace the initial space in footnotes with a tab.

conklech:
This only applies to the docx writer, and implements the alternative behavior from #2527.

I'm open to bikeshedding over the flag name. "footnote" rather than "note" might be better for the user-exposed name. I didn't want to reformat all of `Options.hs`, hence the shorter form.

I updated the README; this should probably be documented elsewhere.

This is what footnotes look like with `--docx-note-tab` and the default template altered to have 0.25" hanging indent:

image

A smaller hanging indent would look nicer for documents without triple-digit footnotes.

Adding that hanging tab to the reference docx's `styles.xml` involves adding this  to the end of the `FootnoteText` style tag:

```
    <w:pPr>
      <w:ind w:left="360" w:hanging="360"/>
    </w:pPr>
```

But note that this style looks terrible with the default spaces, so it shouldn't be the default.

### DIFF
--- a/README
+++ b/README
@@ -744,6 +744,17 @@ Options affecting specific writers
     [character] Default Paragraph Font, Body Text Char, Verbatim Char,
     Footnote Reference, Hyperlink; [table] Normal Table.
 
+`--docx-note-tab`
+
+:   When producing footnotes in docx files, place a tab character rather than a
+    space between the footnote reference (i.e. the footnote number) and the text
+    of the footnote.  If the Footnote Text style in the reference docx has a 
+    hanging indent, this will produce typographically aligned footnotes: the 
+    text will be uniformly aligned, with all footnote references to the left, 
+    as in a well-typeset book.  Correct alignment may also be ensured by 
+    manually modifying the Footnote Text style in the output file to have a
+    small hanging indent. 
+
 `--epub-stylesheet=`*FILE*
 
 :   Use the specified CSS file to style the EPUB.  If no stylesheet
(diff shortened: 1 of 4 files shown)


jgm:
Rather than having another command line option, I wonder
whether it would make more sense always to use a tab,
and to set the tab stop fairly short in the default style,
so it looks about he same as a space?

+++ Christian Conkle [Nov 15 15 15:48 ]:

conklech:
The problem with that is that the tab stop doesn't expand to accommodate longer footnote references. So a tab stop that works well for 1-9 will be too narrow for 10-99. And at some point it'll bump to the next tab stop. We could set several tab stops, so that it'll fall over (sort of) gracefully.

The other problem is that it makes the file un-Wordy, and hard for subsequent editors to work with. (I assume that's a common use case for others as well--my main use of the docx writer is to prepare a semi-finished draft that other lawyers in my firm can work with and I can archive in the format everyone else expects.) If someone adds a new footnote, itll be formatted differently; you have to remember t hit backspace-tab after inserting any new footnote, and fixing footnotes with spaces is fiddly because Word's auto-indent feature kicks in.

In my own usage, I expect to use both the space and tab modes---tabs for "my" documents that I don't expect to be edited by others, and spaces for documents that will circulate and be edited in Word form.

conklech:
Closing in favor of an issue, #2621.  (Also so I can reclaim my `master` branch; sorry!)

PR #2568, opened 2015-12-02
Textile Reader: image attributes

mb21:


src/Text/Pandoc/CSS.hs
```diff
@@ -10,7 +10,7 @@ import Text.Parsec.String
 ruleParser :: Parser (String, String)
 ruleParser = do
     p <- many1 (noneOf ":")  <* char ':'
-    v <- many1 (noneOf ":;") <* char ';' <* spaces
+    v <- many1 (noneOf ":;") <* (optional $ char ';') <* spaces
```
hftf: Isn't only the semicolon in the final rule optional?

src/Text/Pandoc/Readers/Textile.hs
```diff
@@ -535,11 +536,17 @@ link = try $ do
 image :: Parser [Char] ParserState Inlines
 image = try $ do
   char '!' >> notFollowedBy space
-  _ <- attributes -- ignore for now, until we have image attributes
+  (ident, cls, kvs) <- attributes
+  let getAtt k styles = case pickStyleAttrProps [k] styles of
+                          Just v  -> [(k, v)]
+                          Nothing -> []
+  let attr = case lookup "style" kvs of
+               Just stls -> (ident, cls, getAtt "width" stls ++ getAtt "height" stls)
```
hftf: Doesn't this parse the same string multiple times?

mb21: I thought GHC was smart enough to not evaluate the same expression twice as long as it's pure?

hftf: You're probably right, I'm not very used to Haskell.

src/Text/Pandoc/CSS.hs
mb21: Right... so it should be:

    ruleParser :: Parser (String, String)
    ruleParser = do
        p <- many1 (noneOf ":")  <* char ':'
        v <- many1 (noneOf ":;") <* spaces
        return (trim p, trim v)

    styleAttrParser :: Parser [(String, String)]
    styleAttrParser = liftM2 (:)
      (ruleParser <* char ';' <* spaces) $ many
      $ ruleParser <* (optional $ char ';') <* spaces

Or is there a nicer way to write that?

src/Text/Pandoc/Readers/Textile.hs
mb21: I did some [reading](https://wiki.haskell.org/GHC_optimisations#Common_subexpression_elimination) and testing and now think you're actually right. GHC isn't smart enough.

    $ ghci src/Text/Pandoc/CSS.hs
    *Text.Pandoc.CSS> :set +s

    *Text.Pandoc.CSS> let s = concat $ take 10000 $ repeat "height: 40px; width:50%; color: red; background: blue;"

    *Text.Pandoc.CSS> pickStyleAttrProps ["color"] s
    Just "red"
    (0.46 secs, 454,666,920 bytes)

    *Text.Pandoc.CSS> (pickStyleAttrProps ["color"] s, pickStyleAttrProps ["color"] s)
    (Just "red",Just "red")
    (1.05 secs, 908,294,712 bytes)

mb21: @hftf see https://github.com/mb21/pandoc/commit/a9e479d90feae2b237c7d415f1bf754bfcf1a1b8, thanks for the heads-up!

PR #2569, opened 2015-12-03
Add highlighting support to GAP

rgaiacs:
I think this will need changes to https://github.com/jgm/highlighting-kate.

### DIFF
--- a/src/Text/Pandoc/Highlighting.hs
+++ b/src/Text/Pandoc/Highlighting.hs
@@ -124,6 +124,7 @@ langsList =    [("ada","Ada")
                ,("php","PHP")
                ,("xslt","XSLT")
                ,("html","HTML")
+               ,("gap","GAP")
                ]
 
 listingsLangs :: [String]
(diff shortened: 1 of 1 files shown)


jgm:
This change only affects use with the `--listings` option.
To get highlighting for `gap` in general, it would need to be added to highlighting-kate.
What is GAP?  Is there a kate syntax definition file?

alex-konovalov:
@jgm GAP (http://gap-system.org, GitHub: https://github.com/gap-system) is open-source mathematical software system for discrete computational algebra, and it provides [programming language, also called GAP](http://www.gap-system.org/Manuals/doc/ref/chap4.html). 

It happens that Kate supports GAP syntax highlight out-of-box, but highlighting-kate misses gap.xml file. However this one is available in KTextEditor [here](https://quickgit.kde.org/?p=ktexteditor.git&a=history&h=5a797d353ee6bd9916ae514cc5991c526219fef1&f=src%2Fsyntax%2Fdata%2Fgap.xml).

Anyhow, the gap.xml file in KTextEditor is not optimal, and I would like to improve it. Would you be happy for me to submit an PR to  https://github.com/jgm/highlighting-kate with my version of gap.xml (which may also eventually appear in KTextEditor)?

PR #2570, opened 2015-12-03
Image attributes

mb21:


jgm:
Looks like you need to update the rst reader tests.

mb21:
right, fixed that.

hftf:
Should there be unit tests for the tiny CSS module?

jgm:
+++ Ophir Lifshitz [Dec 05 15 12:36 ]:
>   Should there be unit tests for the tiny CSS module?

Wouldn't hurt, if someone wants to write some!

mb21:
@jgm I was looking at the OpenDocument reader when I noticed that the OpenDocument writer didn't support image attributes yet. So I started on the OpenDocument writer, converted it to an `IOStringWriter` until I finally noticed that there was some hack already in place to get image dimensions working in the ODT.hs [transformPicMath](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/ODT.hs#L128) walker which puts the dimensions in the image title field. From the commits it seems this is more historically grown, so should I go ahead with my changes and remove the first half (the image stuff) from that function and put it in the OpenDocument writer? Or would you rather want to keep the OpenDocument writer pure?

jgm:
I'd like to keep the OpenDocument writer pure, yes.

+++ Mauro Bieg [Dec 10 15 05:11 ]:

mb21:
- I kept the OpenDocument writer pure but still made some improvements.
- Probably better if @MarLinn changes the OpenDocument reader since it doesn't support images at all currently.
- Also, for the record: the docx format has really funny way of [doing relative/percentage based widths/heights](http://answers.microsoft.com/en-us/office/forum/office_2010-word/set-relative-size-of-an-image/a22b83c5-d098-4c8c-a68e-f6c92a2ef5eb?db=5&auth=1) which works only on images that are somehow floated. so we currently don't support that.

So I think this pull is good to merge... (The failed test is due to the rebase from master: a SoftBreak is missing)

Issue #1132, opened 2014-01-29
zim output format


ginzel:
Hello,
thank you for pandoc.
Add zim output format please. It looks like DokuWiki. Zim is a flos popular note taking software.
http://zim-wiki.org/manual/Help/Wiki_Syntax.html
Thanks
Hans

jgm:
zimwiki output format has been added.

PR #2577, opened 2015-12-05
Zim writer: initial implementation

Mathnerd314:
This is written for #1132; it has no tests and is still incomplete. I'm opening it now to see what Travis has to say.


Mathnerd314:
#3005, #3011  :+1:

PR #2591, opened 2015-12-16
ICML writer: added figure handling, closes #2590

### LINKED ISSUE #2590: ICML output doesn't retain image caption information
Converting markdown to ICML ignores all captions. I'm not familiar enough with ICML to know whether this is due to a lack in the ICML spec, but losing information on the way doesn't seem like the best way to implement the conversion.

If the ICML spec allows it, the perfect solution would be to retain the captions.

An acceptable solution would be to convert the text into a paragraph after the image, so it could be converted into a caption manually. Now all converted files have to be manually compared to their source to figure out whether the captions have been dropped or whether the images didn't have captions to begin with.

mb21:


jgm:
Great, thanks!

PR #2637, opened 2016-01-10
LaTeX writer: figure label

mb21:


tomduck:
Hi there.  I'm the author of the [pandoc-fignos](https://github.com/tomduck/pandoc-fignos) filter.  I think that this is going to cause a bit of a problem for me.  Pandoc-fignos reads figure attributes and injects a LaTeX \label{} into the abstract syntax tree when the LaTeX format is being used.  With your new patch, I should stop doing that.  Is there any way to detect the pandoc version in a filter so that I can adjust to this new behavior?  Thanks.

jgm:
`Text.Pandoc.pandocVersion` will give you the version.

However, this requires pandoc, not just pandoc-types.
It would be good to export a version from pandoc-types, too.

+++ Thomas J. Duck [Jan 10 16 07:40 ]:

tomduck:
Thanks, @jgm.  My filter is written in python, so I don't think that I have access to `Text.Pandoc.pandocVersion`.

jgm:
+++ Thomas J. Duck [Jan 10 16 13:55 ]:
>   Thanks, [1]@jgm. My filter is written in python, so I don't think that
>   I have access to Text.Pandoc.pandocVersion.

If you're consuming json produced by the pandoc on your
system, then you can just have your python script shell out
to 'pandoc --version' and get the version this way.

tomduck:
> On Jan 10, 2016, at 7:51 PM, John MacFarlane <notifications@github.com> wrote:
> 
> +++ Thomas J. Duck [Jan 10 16 13:55 ]:
> > Thanks, [1]@jgm. My filter is written in python, so I don't think that
> > I have access to Text.Pandoc.pandocVersion.
> 
> If you're consuming json produced by the pandoc on your
> system, then you can just have your python script shell out
> to 'pandoc --version' and get the version this way.

This might not work if one has multiple pandoc versions on the same machine.  For example, I have /usr/local/bin/pandoc-1.5.2, /usr/local/bin/pandoc-1.6-1 and more, all for compatibility testing.  Passing version information along in the JSON would be more reliable.

mb21:
@jgm can this be merged or is this waiting for #2640?

jgm:
Not sure what is best.  I haven't had much time to
review PRs lately, sadly.

+++ Mauro Bieg [Feb 17 16 03:57 ]:

jgm:
@tomduck is this really a problem?  Can't pandoc-fignos simply change the attribute identifier to "" after it injects the label?  And then, even with pandoc 1.17 (assuming this is merged), you won't get an extra label.  mb21's code only has an effect (as far as I can see) if the attribute identifier is non-null.

tomduck:
@jgm Yes, I could do that.  But I think it would be better if filters let pandoc do the processing wherever possible.

In this case pandoc-fignos provides functionality that I expect pandoc is going to gradually replace.  It turns out this replacement is not happening all at once — see mb21’s patch.  If there was a way to detect new behaviours (e.g., by retrieving version information) then pandoc-fignos could stay out of the way where appropriate.

I could use your trick of calling pandoc and asking for the pandoc version, but it seems to me this would be rather fragile.  It would likely break for any pandoc fork.  Having some mechanism to pass version information along solves this problem.

If it can’t be done, no biggie.  But I think that having version information available in the json representation would be an asset to filter writers.


—

Issue #2047, opened 2015-03-31
TEI writer


jgm:
"The Text Encoding Initiative (TEI) is a standard for the representation of textual material in digital form through the means of text encoding.This standard is the collaborative product of a community of scholars, chiefly from the humanities, social sciences, and linguistics who are organized in the TEI Consortium (TEI-C http://www.tei-c.org)."  http://teibyexample.org/

Of course TEI contains lots of rich semantic markup that doesn't correspond to anything in the Pandoc document model.  But it might be useful to people who encode things in TEI if they had a quick way to import documents to a rough (but valid) TEI document that could be progressively enhanced.

markuskiller:
great idea! +1

adunning:
This would be excellent. There are a few journals experimenting with the use of TEI as a base format, such as DH Quarterly and RIDE, and Pandoc support would be a huge help. You will have to choose a customization for output; TEI Simple (https://github.com/TEIC/TEI-Simple) might be suited to this purpose.

jasonseeley:
+1 I've been kinda wanting this, too. I think TEI-Tiny might be more appropriate than Simple, though, especially with its wide adoption. (http://www.tei-c.org/Guidelines/Customization/Lite/)

ousia:
:+1: 

I think _TEI Lite_ would be a better option than _TEI Simple_ because of its widespread adoption.

@jasonseeley, you meant _TEI Lite_, didn’t you?

adunning:
@sebastianrahtz, do you have any thoughts on the best TEI customization to implement here?

sebastianrahtz:
The difference between TEI Lite and TEI Simple in this respect is trivial. Don't worry about it. I'd be very very surprised if the sort of lowest-common-denominator that is  Pandoc's everyday business would trigger complex TEI XML that exposed the differences.  More interesting is whether to use @rend, or@rendition or @style to indicate changes of emphasis (where you can't tell why the emphasis changes). TEI Simple standardizes on @rendition, with a fixed range of values - I might suggest Pandoc follows those. See list in "3.5. Attribute classes" in http://htmlpreview.github.io/?https://github.com/TEIC/TEI-Simple/blob/master/teisimple.html

Hasimir:
[...]

Issue #179, opened 2011-06-10
only valid tags should stay in output, all other occurences of < and > should be transformed into character entities


jgm:
<b>What steps will reproduce the problem?</b>

input: a <valid_tag> and <not a valid tag>

<b>What is the expected output? What do you see instead?</b>

expected: a <valid_tag> and &lt;not a valid tag&gt;

<b>What version of the product are you using? On what operating system?</b>

pandoc 1.2.1 on OSX

<b>Please provide any additional information below.</b>

only valid tags should be left in output, all other occurences of < or >
should be transformed into character entities.

Google Code Info:
Issue #: 176
Author: janh...@googlemail.com
Created On: 2009-11-03T16:45:29.000Z
Closed On: 2009-11-22T18:38:07.000Z

PR #2638, opened 2016-01-10
Add TEI Writer.

c-forster:
Add a TEI Writer (see https://github.com/jgm/pandoc/issues/2047), outputting TEI XML consistent with the [TEI Simple](https://github.com/TEIC/TEI-Simple) subset. 

This writer was developed by modifying the DocBook writer.

The standalone output for this writer depends on a new template, `default.tei`, submitted as part of a [separate pull request](https://github.com/jgm/pandoc-templates/pull/179) (#179) to the `pandoc-templates` repostiory.

### DIFF
--- a/README
+++ b/README
@@ -22,7 +22,7 @@ markup], [Haddock markup], [OPML], [Emacs Org mode], [DocBook],
 [OpenDocument], [ODT], [Word docx], [GNU Texinfo], [MediaWiki markup],
 [DokuWiki markup], [Haddock markup], [EPUB] (v2 or v3),
 [FictionBook2], [Textile], [groff man] pages, [Emacs Org mode],
-[AsciiDoc], [InDesign ICML], and [Slidy], [Slideous], [DZSlides],
+[AsciiDoc], [InDesign ICML], [TEI XML], and [Slidy], [Slideous], [DZSlides],
 [reveal.js] or [S5] HTML slide shows. It can also produce [PDF] output
 on systems where LaTeX, ConTeXt, or `wkhtmltopdf` is installed.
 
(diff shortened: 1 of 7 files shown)


jgm:
Can you squash these changes into one commit that adds the TEI writer?

c-forster:
I *think* that's what I just did; after some thrashing about, I used `git -i rebase` and hope I didn't make too much of a hash of it. (I'm rather a newb to really using git[hub] I fear.)

c-forster:
Updated branch to incorporate upstream changes and pass Travis tests.

jgm:
So is this ready to go?

+++ Chris Forster [Jan 19 16 12:27 ]:

c-forster:
It's not perfect, but this PR, combined with https://github.com/jgm/pandoc-templates/pull/179, will produce TEI output that validates against the DTD (`teisimple.dtd`) contained here: https://github.com/TEIC/TEI-Simple.

(More precisely, I used this Writer and Template to convert the README, and it produces no output with ` xmllint --noout --dtdvalid ./teisimple.dtd readme.tei`).

I think the default template could be expanded to more fully represent the TEI Header (I imagine most of which would be coming from YAML metadata); but this provides a good starting point I think.

If you'd like to give it a fuller test, I could post it to the TEI mailing list and try to get some feedback from TEI folks before merging it.

jgm:
I'll merge it now; it's a good idea to get further feedback, but this is a great start.

jgm:
I added some "old-style" tests: `tests/writer.tei` and `tests/tables.tei`.  Please review these and make sure the test output looks okay.

c-forster:
Checked both old-style tests (thanks for adding those); `writer.tei` validates against the current DTD; and, if you add a valid header, `tables` does too. Thanks!

jgm:
Good.  I think getting feedback from the TEI list is a good
next step.

PR #2665, opened 2016-01-20
LaTeX writer: Allow more flexible table alignment

monofon:
Removing the centering option `[c]` to the `longtable` environment allows LaTeX templates to control the overall table alignment in a document by setting the `longtable` length variables `LTleft` and `LTright`.

Example:

```latex
\setlength\LTleft\parindent
\setlength\LTright\fill
```

results in left aligned tables that respect paragraph indentation.

This patch will not change the default behavior of pandoc because `centering` is the default for tables even if `[c]` is not specified.


jgm:
good idea, I didn't know centering was the default.

PR #2668, opened 2016-01-21
Markdown writer: Use hyphens for yaml metadata block bottom line

monofon:
I use pandoc as a formatter and pretty pretty-printer for markdown documents. Many markdown tools in my pipeline accept YAML metadata blocks delimited top and bottom by three hyphens `---` (for example: [jekyll](http://jekyllrb.com), [mustache](https://mustache.github.io) or [hugo](https://gohugo.io)). Pandoc generated markdown uses three dots `...` to mark the bottom of a metadata block which breaks these tools.
 
This patch changes the markdown writer to use three hyphens to mark the end of a metadata block. 

Some notes:

- The reader remains untouched and will still accept dots as well
- Other tools that rely on the dots may break
- The YAML spec suggests `...` to mark the end of a document in a stream of YAML documents
- This time I did not forget to update the tests

### DIFF
--- a/src/Text/Pandoc/Writers/Markdown.hs
+++ b/src/Text/Pandoc/Writers/Markdown.hs
@@ -116,7 +116,7 @@ plainTitleBlock tit auths dat =
   dat <> cr
 
 yamlMetadataBlock :: Value -> Doc
-yamlMetadataBlock v = "---" $$ (jsonToYaml v) $$ "..."
+yamlMetadataBlock v = "---" $$ (jsonToYaml v) $$ "---"
 
 jsonToYaml :: Value -> Doc
 jsonToYaml (Object hashmap) =
--- a/tests/writer.markdown
+++ b/tests/writer.markdown
@@ -4,7 +4,7 @@ author:
 - Anonymous
 date: 'July 17, 2006'
 title: Pandoc Test Suite
-...
+---
 
 This is a set of tests for pandoc. Most of them are adapted from John Gruber’s
 markdown test suite.


jgm:
This has come up before, and I've always suggested that
people complain to the authors of these other tools.

As you note, the use of three dots to end a YAML stream is
right from the YAML spec:
http://www.yaml.org/spec/1.2/spec.html#id2800132 This makes
it possible to use standard YAML tools to process the
document.

The use of three hyphens is non-standard.  So pandoc's behavior
here is much more sensible.  Moreover, using a `---` has the
bad consequence that the preceding line is highlighted as a
heading in many editors with markdown highlighting.

That's why I'm reluctant to merge this...

+++ monofon [Jan 21 16 04:23 ]:

monofon:
Yes, I can see your point.

However, my understanding of the YAML spec is that a YAML stream is meant to consist only of YAML documents. Using YAML as front matter for a non-YAML document is not covered in the spec. The required behavior of a YAML parser seems more or less undefined. YAML parsers handle this case inconsistently:

- yaml-0.8.15.3 chokes on everything that is not a valid YAML document wether separated by `---` or `...\n---`
- yamljs does not support YAML document streams at all
- http://yaml-online-parser.appspot.com ignores everything but the first yaml document regardless of separator

So my guess is, using `---`  would not violate the YAML spec and most probably wouldn't really break anything implementing it.

Again, I see your point. But I had to try.

jgm:
I'll think about it.  As a matter of practicality, you're
probably right that this would suit most people best.

+++ monofon [Jan 22 16 10:13 ]:

tedhudek:
As John surmised in #2672 GitHub does not handle the triple dot properly.  You guys know a lot more about YAML than I do, but FWIW I vote to accept this PR :+1:

monofon:
Thanks!

MathieuDuponchelle:
This breaks pyyaml , and needs to be worked around, I don't know if pyyaml is actually compliant to the spec but here was my problem:

```
---
foo: 'bar'
---
```

pyyaml.load , which was the function I was previously using as my input always contains one and only one such block, now fails with:

```
ComposerError: expected a single document in the stream
```

I now have to use pyyaml.load_all() , and it gives me a generator that now yields two dictionaries, the second one being None .

So my experience with this patch is that "As a matter of practicality" , it actually makes things worse for me ^^

I would suggest strictly following the spec if that's not the case, otherwise I guess pyyaml isn't compliant, but I'd be surprised if a project with such a name wasn't :P

MathieuDuponchelle:
To test the issue just follow these steps:

```
virtualenv my_silly_pyyaml_test
. my_silly_pyyaml_test/bin/activate
pip install pyyaml
```

Create silly_test.py, put:

```python
import yaml
from yaml.composer import ComposerError

if __name__=='__main__':
    res = yaml.load(u'---\n'
                    'Foo: "bar"\n'
                    '...\n')

    print "Dot bottom line accepted, result:", res

    try:
        res = yaml.load(u'---\n'
                        'Foo: "bar"\n'
                        '---\n')
    except ComposerError:
        print "Hyphen bottom line reject"
```

Then execute it:

```
python silly_test.py
```

jgm:
This is exactly why I resisted this change for so long.
I suspected that people might be using standard YAML tools
which would break.  OP assured that this wasn't the case,
giving an example or two, so I went ahead and did it.

Now, I think, I regret it.

+++ Mathieu Duponchelle [Apr 16 16 06:45 ]:

monofon:
Sorry I caused problems for some people with this pull request. If you feel, that this causes more trouble than it is worth, please don't hesitate to revert back to three dots.

ickc:
May be a command line option to choose whether the markdown writer should end a YAML front matter by `---` or `...`, but default to `...`?

tedhudek:
Sounds reasonable to me. Thanks for working this through.
________________________________

MathieuDuponchelle:
I worked around the issue on my end so I think my opinion is neutral enough, which is that this change should be reverted and pandoc should strictly follow the yaml specification.

The various non-compliant tools cited in the beginning should be made compliant, in the meantime users looking for compatibility between pandoc and these tools should implement their own translation layer (pompous words for a substitute operation I know :)

Just my 2 cents anyway :)

jgm:
I don't like to multiply command-line options beyond
necessity.

I could add an extension, but that also may be
overkill when it's so easy just to do a regex replacement.

ickc:
@jgm said,

> Now, I think, I regret it.

Should this be undone? I thought it is undone until I noticed the markdown writer uses `---` to end.

ickc:
May I ask an update on this? Will the new behavior stay, or will it be reverted?

jgm:
Still undecided.

+++ ickc [Jan 12 17 00:54 ]:
>   May I ask an update on this? Will the new behavior stay, or will it be
>   reverted?

MathieuDuponchelle:
@jgm, this really should be reverted IMHO, the (default) behaviour should be to respect standards.

PR #2673, opened 2016-01-22
Add syntax for creating GFM+YAML frontmatter

tedhudek:
It might be worthwhile to update the thread here as well:

http://stackoverflow.com/questions/22866498/use-pandoc-to-create-yaml-metadata-from-html-meta-tags

### DIFF
--- a/README
+++ b/README
@@ -2410,6 +2410,10 @@ a YAML metadata block will be produced only if the `-s/--standalone`
 option is used.  All of the metadata will appear in a single block
 at the beginning of the document.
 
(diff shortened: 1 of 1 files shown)


jgm:
Thanks, but I'm going to pass on this suggestion.  The README already contains this information, since it indicates what extensions markdown_github includes, and how to enable extensions selectively.  True, this note might be useful to someone who wants to do precisely this -- but with equal justice I could add fifty other paragraphs about using other features with markdown_github or other variants.

PR #2691, opened 2016-01-31
Org reader: Refactor link-target processing

### LINKED ISSUE #2684: org-mode reader does not support `[[file:filename]]` link format for images
Org's default file link format is to include `[[file:filename]]`.  It would be great if Pandoc could support this for conversion of image links.  Currently this require manual intervention to convert to the more Pandoc friendly `[[filename]]`

tarleb:
Cleanup of the code for link target handling.  Most notably, the
canonicalization of a link is handled by a separate function.

This fixes #2684.


tarleb:
I'm not sure what's going wrong with the GHC 7.4 and 7.6 builds, but it appears to be unrelated to the changes I made.  The builds for later GHC versions succeed.

PR #2710, opened 2016-02-13
Add an option to parse-before-combining.

jkr:
This is more of an RFC than a ready-to-go PR. It's my attempt to deal with the issue of footnotes in multiple documents, which is just enough of an issue to keep me from regularly using pandoc/markdown for article-length work. These two commits add an option to parse documents before combining. This allows multiple documents with footnotes starting at `[^1]` to combine as expected, and also allows us to combine files from different input formats, or multiple binary files. The trade-off is that we can't link across multiple files, and auto-identified headers don't always work as expected if the name is repeated. This is an option (`--parse-first` flag) that defaults to off (i.e., default behavior is normal).

For all that, the change is fairly unintrusive. The vast majority of the lines of code are option boilerplate.

More detailed info is in the message for the first commit. Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,87 +1,136 @@
-# This file has been generated -- see https://github.com/hvr/multi-ghc-travis
-language: c
+# Copy these contents into the root directory of your Github project in a file
+# named .travis.yml
+
+# Use new container infrastructure to enable caching
 sudo: false
 
+# Choose a lightweight base image; we provide our own build tools.
+language: c
+
+# Caching so the next build will be fast too.
 cache:
   directories:
-    - $HOME/.cabsnap
-    - $HOME/.cabal/packages
+  - $HOME/.ghc
+  - $HOME/.cabal
+  - $HOME/.stack
 
-before_cache:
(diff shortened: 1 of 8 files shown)


jgm:
I think there might be others who'd like this feature, too.
Maybe worth posting on pandoc-discuss describing the proposed
feature and linking here?

+++ Jesse Rosenthal [Feb 13 16 13:08 ]:

jkr:
Good idea -- note posted to the list.

jkr:
Actually, it occurs to me that Binary readers should probably default to this. In the single-source case (which is all that is allowed for binaries at the moment) it is the same. But this would allow multiple sources (provided they're the same format).

jgm:
+++ Jesse Rosenthal [Feb 14 16 05:13 ]:
>   Actually, it occurs to me that Binary readers should probably default
>   to this. In the single-source case (which is all that is allowed for
>   binaries at the moment) it is the same. But this would allow multiple
>   sources (provided they're the same format).

Yes, that makes sense, I suppose.

jkr:
And the option name seems wrong: too implementation specific. Something more like `--file-scope` (as opposed to document-scope) seems preferable.

PR #2724, opened 2016-02-18
Update LaTeX writer to align images

frasertweedale:
Update the LaTeX writer to align images that have an "align"
attribute whose value is "left", "right" or "center".  If the
attribute is absent or has an unrecognised values, the image is left
alone.

Also update the RST reader to read these horizonal alignment values
from the image directive (the first encountered value applies). See
http://docutils.sourceforge.net/docs/ref/rst/directives.html#image
for details.

### DIFF
--- a/src/Text/Pandoc/Readers/RST.hs
+++ b/src/Text/Pandoc/Readers/RST.hs
@@ -39,7 +39,7 @@ import Text.Pandoc.Shared
 import Text.Pandoc.Parsing
 import Text.Pandoc.Options
 import Control.Monad ( when, liftM, guard, mzero )
-import Data.List ( findIndex, intersperse, intercalate,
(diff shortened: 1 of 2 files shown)


bluebirch:
+1

mb21:
Although I cannot speak for @jgm, I guess for this to be merged, it would have to be implemented in the most commonly used writers (i.e. HTML probably emitting a `style="text-align: left"`) and documented in the `README`.

mb21:
Then again, https://github.com/jgm/pandoc/commit/99b39ffc1726a86e6a6ad45ef6b81a28b46594ef meanwhile set a precedent to simply store alignment in the AST as simply a `align-left` or `align-right` class on the `Image` element...

wallento:
Stumbled over this as a blocker in my work today. Anything I can do to help move it forward?

PR #2725, opened 2016-02-18
Remove stray line from stack.full.yaml

adunning:
The line causes an error with stack 1.0.2:

```
Could not parse '/pandoc-build/pandoc/stack.full.yaml':
AesonException "Error in $.extra-deps: failed to parse field 'extra-deps': failed to parse field extra-deps: expected [a], encountered Null"
See http://docs.haskellstack.org/en/stable/yaml_configuration.html.
```

### DIFF
--- a/stack.full.yaml
+++ b/stack.full.yaml
@@ -12,5 +12,4 @@ packages:
 - '../pandoc-citeproc'
 - '../pandoc-types'
 - '../texmath'
-extra-deps:
 resolver: lts-5.2


jgm:
I think this is already fixed in master.

+++ Andrew Dunning [Feb 18 16 07:08 ]:

adunning:
No, this was caused by the last commit to the file.

jgm:
Oh, sorry. I only fixed it in stack.yaml, not
stack.full.yaml.

+++ Andrew Dunning [Feb 18 16 11:07 ]:

Issue #2744, opened 2016-02-29
ConTeXt, Whitespace, and Line Blocks 


c-forster:
If I have the following markdown:

```markdown
| Fair fa' your honest, sonsie face,
|     Great Chieftain o' the Puddin'-race!
| Aboon them a' ye tak your place,
|            Painch, tripe, or thairm:
| Weel are ye wordy o' a *grace*
|            As lang's my arm.
```
And I convert it to ConTeXt, I get:

```context
Fair fa' your honest, sonsie face,\crlf
~~~~Great Chieftain o' the Puddin'-race!\crlf
Aboon them a' ye tak your place,\crlf
~~~~~~~~~~~Painch, tripe, or thairm:\crlf
Weel are ye wordy o' a {\em grace}\crlf
~~~~~~~~~~~As lang's my arm.
```

With LaTeX, I get:
```latex
Fair fa' your honest, sonsie face,\\
\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}Great
Chieftain o' the Puddin'-race!\\
Aboon them a' ye tak your place,\\
\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}Painch,
tripe, or thairm:\\
Weel are ye wordy o' a \emph{grace}\\
\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}\hspace*{0.333em}As lang's my arm.
```

The ConTeXt writer converts spaces in Line Blocks into nonbreaking spaces (`~`s), which don't get preserved as horizontal space when it is compiled to PDF.

jgm:
Are there any ConTeXt experts out there who can suggest
better output?

+++ Chris Forster [Feb 28 16 19:43 ]:

c-forster:
I corresponded with @denubis via email, who helped me come up with a solution that seems to work and is represented in https://github.com/jgm/pandoc/pull/2745

Denubis:
It may be worth mentioning that ~\hspace[small] repeated can be replaced
with:
[...]

PR #2745, opened 2016-02-29
Context line blocks

c-forster:
Modeled on the LaTeX writer's use of `hspace*{0.333em}`, this PR uses ConTeXt's `hspace[small]` (combined with `~`), and `\strut`s (all necessary to produce the correct amount of whitespace) to output whitespace represented by non-breaking spaces in the pandoc AST.

E.g. the following markdown:

```markdown
| Fair fa' your honest, sonsie face,
|     Great Chieftain o' the Puddin'-race!
| Aboon them a' ye tak your place,
|            Painch, tripe, or thairm:
| Weel are ye wordy o' a *grace*
|            As lang's my arm.
```

Will now produces the following output:

```context
Fair fa' your honest, sonsie face,\crlf
\strut~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]Great
Chieftain o' the Puddin'-race!\crlf
Aboon them a' ye tak your place,\crlf
\strut~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]Painch,
tripe, or thairm:\crlf
Weel are ye wordy o' a {\em grace}\crlf
\strut~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]~\hspace[small]As
lang's my arm.
```

Which compiles to an appropriate looking PDF. 

Fixes https://github.com/jgm/pandoc/issues/2744 

(With thanks to @denubis for recomendations on handling whitespace in ConTeXt.)

### DIFF
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -279,7 +279,29 @@ blockListToConTeXt lst = liftM vcat $ mapM blockToConTeXt lst
 -- | Convert list of inline elements to ConTeXt.
 inlineListToConTeXt :: [Inline]  -- ^ Inlines to convert
                     -> State WriterState Doc
-inlineListToConTeXt lst = liftM hcat $ mapM inlineToConTeXt lst
+inlineListToConTeXt lst = 
+  mapM inlineToConTeXt (fixBreaks $ fixLineInitialSpaces lst)
+    >>= return . hcat
+    -- nonbreaking spaces (~) in ConTeX don't work after line breaks,
+    -- so we turn nbsps after hard breaks to \hspace commands.
+    -- this is mostly used in verse.
+ where fixLineInitialSpaces [] = []
+       fixLineInitialSpaces (LineBreak : Str s@('\160':_) : xs) =
+         LineBreak : fixNbsps s ++ fixLineInitialSpaces xs
+       fixLineInitialSpaces (x:xs) = x : fixLineInitialSpaces xs
+       fixNbsps s = let (ys,zs) = span (=='\160') s
+                    in  [RawInline "context" "\\strut"] ++ replicate (length ys) hspace ++ [Str zs]
+       hspace = RawInline "context" "\\hskip 0.333em "
+       -- linebreaks after blank lines cause problems:
+       fixBreaks [] = []
+       fixBreaks ys@(LineBreak : LineBreak : _) =
+         case span (== LineBreak) ys of
+               (lbs, rest) -> RawInline "context"
+                               ("\\\\[" ++ show (length lbs) ++
+                                "\\baselineskip]") : fixBreaks rest
+       fixBreaks (y:ys) = y : fixBreaks ys
+
+
 
 -- | Convert inline element to ConTeXt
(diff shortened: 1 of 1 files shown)


jgm:
Would this simpler approach work?  Modify the renderer for `Str` so that it checks to see if the string contains two or more consecutive nonbreaking spaces.  If so, use whatever fancy ConTeXt is needed to produce the wider space.  If not, just use `~`.  That approach seems simpler and it would also work with spacing that isn't at the beginning of a line. But there's probably something I'm not seeing here; I don't understand the details of why all that ConTeXt code is needed.

c-forster:
So the solution in this PR (which I put together essentially by rewriting the LaTeX writer) *does* work; when compiled, the PDF it produces ([pandoc_pr2745.pdf](https://github.com/jgm/pandoc/files/167154/pandoc_pr2745.pdf)) works on spaces in the first line (but it represents those spaces differently---b/c of the janky hack way I implemented this). It *does not*, however, respect spaces *within* a line.

That is, with this pull request, the following markdown:

```markdown
|     Initial line, indented *four* spaces.
| No indent.
|   Two space indent.
| Mid          space.
```

Produces this ConTeXt:

```tex
\starttext

~~~~Initial line, indented {\em four} spaces.\crlf
No indent.\crlf
\strut\hskip 0.333em \hskip 0.333em Two space indent.\crlf
Mid space.

\stoptext
```
As the PDF shows, I think this works; but it is inconsistent in how it represents space, **and** it doesn't work for spaces in the middle of a line.

So while it does work for initial lines, I don't this PR represents a good solution to this problem. I also don't think that the initial space problem can be solved the `Str` renderer, because the fundamental ConTeXt problem concerns not the spaces, but the newlines. 

I've actually realized that one *could* just continue to use `~` (nonbreaking space) rather than `hskip` to represent space. The trick (the reason that the current writer doesn't work) is that in addition to the newlines (produced by `crlf`), ConTeXt, for reasons I don't fully understand, also needs a `strut` to respect the spaces. Without a `strut` after the `clrf`, ConTeXt ignores the nonbreaking spaces when it is compiled to PDF (@denubis pointed this out to me).

So, in addition to the `crlf` necessary for a newline, a `strut` is necessary to make the indentation work. A crude, but simpler solution, would be to simply add a `strut` to all emitted `crlf`s (see line 327 of `ConTeXt.hs` on the master branch right now). That would get *initial* spaces working (and, from my quick test, doesn't break any of the existing ConTeXt tests). Ideally, though, following markdown:

```markdown
|     First line, indented 4 spaces.
| Line with mid      space.
|     Line indented 4 spaces.
```

would be output as the following ConTeXt:

```tex
~~~~Indented first line.\crlf
Line with mid~~~~~~space.\crlf
\strut~~~~Line indented 4 spaces.

```
When this is compiled it works for *midspaces* as well. For solving *this* problem (the representation of spaces in the middle of a line in a Line Block), changing the `Str` writer to look for multiple spaces in a row, *would* work, I think (though implementing that would be at the utter limit of my haskell ability).

PR #2757, opened 2016-03-05
Open Document writer: set first level of blockquotes to not use indent

### LINKED ISSUE #2747: ODF output Blockquotes should not have a non-style indent applied
Setting a block quote to the quotation style (already happens) should be enough for block quotes. Currently, pandoc also adds a non-style paragraph left indent of 0.5", which will always completely override the style's indent, rendering styling useless until every blockquote is manually fixed.

jjsheets:
Nested blockquotes start using indents like before. Quotation style is still in use, so the style's indent settings take effect on the first level of blockquotes. This is to fix #2747. Some mechanism to set a starting indent for the second level block quotes and a delta for increasing levels might be useful, but that is beyond my capabilities in Haskell.

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -106,7 +106,7 @@ increaseIndent :: State WriterState ()
 increaseIndent = modify $ \s -> s { stIndentPara = 1 + stIndentPara s }
 
 resetIndent :: State WriterState ()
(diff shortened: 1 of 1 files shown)


jgm:
@jjsheets can you see why the Travis build is failing?

jjsheets:
It's failing due to a test suite fail, I assume because the output now no longer matches the expected output. Unfortunately, I no longer have time to work on fixing this.

PR #2771, opened 2016-03-09
Docx Writer: handle image alt text

### LINKED ISSUE #2754: Accessibility: alt text for images Word documents
It would be great if Pandoc put alt text from images into Word's "Alt Text" Description panel. This would allow authors create accessible documents because the alt text can be used by people reliant on assistive technology (for example, a screen reader) to understand the image. This should happen for inline images and also images placed on their own lines. A markdown example would be:

```
![a pony running through a meadow](images/pony.png)
```

The text "a pony running through a meadow" would be placed into the Description panel of Word's Alt Text panel. WebAIM has an article on this here: http://webaim.org/techniques/word/ Although the screen shots are old, I've just checked the claims about using the Description panel rather than the Title one and they're still correct. Manually adding alt text to images in Word documents is painful, so this feature, if possible to implement, would be very welcome.

mb21:
closes #2754

### DIFF
--- a/src/Text/Pandoc/Writers/Docx.hs
+++ b/src/Text/Pandoc/Writers/Docx.hs
@@ -1103,7 +1103,7 @@ inlineToOpenXML opts (Link _ txt (src,_)) = do
                         M.insert src i extlinks }
               return i
   return [ mknode "w:hyperlink" [("r:id",id')] contents ]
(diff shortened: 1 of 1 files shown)


mb21:
@jgm right, of course, changed to use `stringify`...

jgm:
Thanks!

PR #2818, opened 2016-03-26
Fixes #1762

### LINKED ISSUE #1762: Provide a way to exclude (some) unnumbered headers from TOC
The Pandoc LaTeX Writer adds `\addcontentsline` after any unnumbered section:

```
\section*{Section Heading}\label{section-heading}
\addcontentsline{toc}{section}{Section Heading}
```

This is clearly by design (hence I don't provide a MWE) from lines 662-666 of pandoc/src/Text/Pandoc/Writers/LaTeX.hs:

```
$$ if unnumbered
    then "\\addcontentsline{toc}" <>
        braces (text sectionType) <>
        braces txtNoNotes
else empty
```

This behavior is not correct (for LaTeX). The star in the LaTeX section command (`\section*{}`) removes section number and also removes the entry for the TOC? Why is it explicitly added back? I suppose the answer is that the Pandoc internal data type treats section numbering and TOC entries separately, and other formats (e.g. HTML) expect to see these sections in the TOC. But for that matter, what format besides LaTeX numbers sections anyway? Isn't the unnumber flag (`# Section {-}` or `# Section {.unnumbered}`) intended for LaTeX?

This has come up before on pandoc-discuss: https://groups.google.com/forum/#!searchin/pandoc-discuss/addcontentsline/pandoc-discuss/5yWDF28sDWU/ZzbGM9EU9RkJ

fiddlosopher offers the workaround hack:

```
\renewcommand{\addcontentsline}[3]{}
```

I prefer to deal with it via sed:

```
pandoc -f markdown-auto_identifiers -t latex Document.md | sed -e '/addcontentsline/d' > Document.tex
```

However, this fails on occasionally long subsection titles, which wrap.

I believe the correct patch is to just remove this code (just as it is ignored in LaTeX Reader), or enable it with a switch for compatibility.

chdemko:
See http://stackoverflow.com/questions/2785260/hide-an-entry-from-toc-in-latex

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -251,7 +251,7 @@ elementToLaTeX :: WriterOptions -> Element -> State WriterState Doc
 elementToLaTeX _ (Blk block) = blockToLaTeX block
 elementToLaTeX opts (Sec level _ (id',classes,_) title' elements) = do
   modify $ \s -> s{stInHeading = True}
-  header' <- sectionHeader ("unnumbered" `elem` classes) id' level title'
+  header' <- sectionHeader ("unnumbered" `elem` classes) ("no-toc" `elem` classes) id' level title'
   modify $ \s -> s{stInHeading = False}
   innerContents <- mapM (elementToLaTeX opts) elements
   return $ vsep (header' : innerContents)
@@ -575,7 +575,7 @@ blockToLaTeX HorizontalRule = return $
   "\\begin{center}\\rule{0.5\\linewidth}{\\linethickness}\\end{center}"
 blockToLaTeX (Header level (id',classes,_) lst) = do
   modify $ \s -> s{stInHeading = True}
-  hdr <- sectionHeader ("unnumbered" `elem` classes) id' level lst
+  hdr <- sectionHeader ("unnumbered" `elem` classes) ("no-toc" `elem` classes) id' level lst
   modify $ \s -> s{stInHeading = False}
   return hdr
 blockToLaTeX (Table caption aligns widths heads rows) = do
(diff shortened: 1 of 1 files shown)


tarleb:
Thank you for the PR @chdemko and apologies for the long wait.  The change and code looks good to me, though jgm will have to make the final decision.  Some minor issues remain, those should be addressed first:

  -  Please rebase your branch to the current upstream master.  There has been a minor change in the way `*` are added to commands which conflicts with the proposed changes.  Nothing major though.
  - Since this is an atomic change, it would be nice if the commits where squashed into a single commit.
  - Pandoc code follows the convention of keeping lines below 80 chars; please reformat the code to keep it below this limit.
  - Parentheses around the argument to `not` can be dropped.

Thanks for your patience and for taking the time to compile the PR in the first place.

tarleb:
One additional thought I had was that it might make sense to introduce an additional data-type for section parameters.  I'm always a bit skeptical towards boolean flags and having two boolean flags in the function signature can make the code difficult to read (*which option is which?*).  An additional type would be a bit more verbose but might increase readability.

tarleb:
@jgm: Am I right in thinking that you'd want to see the feature implemented in other writers as well before this can be merged?  We'd also need to document the feature in the MANUAL.

jgm:
I don't know about all writers.  But it would be good to
implement it in at least the most widely used ones (e.g.
HTML, LaTeX, Markdown).  Its limitations would have to
be documented in the MANUAL.

+++ Albert Krewinkel [Sep 23 16 23:31 ]:
>   [1]@jgm: Am I right in thinking that you'd want to see the feature
>   implemented in other writers as well before this can be merged? We'd
>   also need to document the feature in the MANUAL.

utopos:
This is a great feature!!! When it can be expected to be merged into the main branch of pandoc? Thx guys!

jgm:
I was hoping it could be implemented in other formats, not just LaTeX.

utopos:
@jgm Thx for the reply. Unfortunately I'm not familiar with Haskell in order to contribute here for other writers, so I was wondering whether we could make it a latex specific option or @chdemko - do you think you could be so kind to extend this feature over other main writers mentioned by John?

chdemko:
@utopos Unfortunately, I'm not very familiar with Haskell and I don't have the time right now to extend the feature to other writers

PR #2735, opened 2016-02-23
LaTeX Writer: fix polyglossia to babel env mapping

mb21:
allow for optional argument in square brackets, closes #2728

PR #2829, opened 2016-03-31
LaTeX writer: Add missing languages.

adunning:
Updates the list from the hyphenation files at <http://mirror.ctan.org/language/hyph-utf8/tex/generic/hyph-utf8/loadhyph/>.


adunning:
Still some problems with Babel conversion, but this will be better once #2735 is merged.

adunning:
Had forgotten to update the tests, but this is now fixed.

Issue #2862, opened 2016-04-19
Incorrect spacing with org-mode src blocks?


shosti:
Let me know if I'm missing some configuration, but I'm running into some trouble with indentation with markdown_github -> org-mode source block translation: pandoc appears to be adding leading indentation within the source blocks. For instance:

    ```ruby
    x = "something"
    y = "something else"
    ```

when run with `pandoc -f markdown_github -t org`, gets translated into:

    #+BEGIN_SRC ruby
        x = "something"
        y = "something else"
    #+END_SRC

and translating back to markdown_github preserves the spacing (so every back-and-forth round trip adds leading space to the code).

I'm running pandoc 1.17.0.3 on Arch.

tarleb:
Just to be sure: You are referring to an identation error occuring when
translating back-and-forth between *github_markdown* and *org*, correct?

Looks like that's my fault (i.e. a bug in the org reader).
Indentation isn't stripped the way it should be.

    echo "#+BEGIN_SRC ruby\n  hello\n#+END_SRC\n" | pandoc -f org -t native
    [CodeBlock ("",["ruby"],[]) "  hello\n"]

I'll look into it.

shosti:
Yes, it's markdown_github->org (other direction seems fine), sorry said that wrong the first time.

PR #2876, opened 2016-04-25
Ignore leading space in org code blocks

shosti:
Here's my attempt at fixing #2862 (I'm pretty new at Haskell, so let me know if there's a better way...)

### DIFF
--- a/src/Text/Pandoc/Readers/Org.hs
+++ b/src/Text/Pandoc/Readers/Org.hs
@@ -391,6 +391,9 @@ lookupBlockAttribute key =
 
 type BlockProperties = (Int, String)  -- (Indentation, Block-Type)
 
+updateIndent :: BlockProperties -> Int -> BlockProperties
+updateIndent (_, blkType) indent = (indent, blkType)
+
 orgBlock :: OrgParser (F Blocks)
 orgBlock = try $ do
   blockProp@(_, blkType) <- blockHeaderStart
@@ -407,10 +410,22 @@ orgBlock = try $ do
       _         -> withParsed (fmap $ divWithClass blkType)
 
 blockHeaderStart :: OrgParser (Int, String)
-blockHeaderStart = try $ (,) <$> indent <*> blockType
+blockHeaderStart = try $ (,) <$> indentation <*> blockType
  where
-  indent    = length      <$> many spaceChar
   blockType = map toLower <$> (stringAnyCase "#+begin_" *> orgArgWord)
+
+indentation :: OrgParser Int
+indentation = try $ do
+  tabStop  <- getOption readerTabStop
+  s        <- many spaceChar
+  return $ spaceLength tabStop s
+
+spaceLength :: Int -> String -> Int
+spaceLength tabStop s = (sum . map charLen) s
+ where
+  charLen ' '  = 1
+  charLen '\t' = tabStop
+  charLen _    = 0
 
 withRaw'   :: (String   -> F Blocks) -> BlockProperties -> OrgParser (F Blocks)
 withRaw'   f blockProp = (ignHeaders *> (f <$> rawBlockContent blockProp))
@@ -450,7 +465,8 @@ codeBlock blkProp = do
   skipSpaces
   (classes, kv)     <- codeHeaderArgs <|> (mempty <$ ignHeaders)
   id'               <- fromMaybe "" <$> lookupBlockAttribute "name"
-  content           <- rawBlockContent blkProp
+  leadingIndent     <- lookAhead indentation
+  content           <- rawBlockContent (updateIndent blkProp leadingIndent)
   resultsContent    <- followingResultsBlock
   let includeCode    = exportsCode kv
   let includeResults = exportsResults kv
@@ -472,7 +488,7 @@ rawBlockContent (indent, blockType) = try $
   unlines . map commaEscaped <$> manyTill indentedLine blockEnder
  where
    indentedLine = try $ ("" <$ blankline) <|> (indentWith indent *> anyLine)
-   blockEnder = try $ indentWith indent *> stringAnyCase ("#+end_" <> blockType)
+   blockEnder = try $ skipSpaces *> stringAnyCase ("#+end_" <> blockType)
 
 parsedBlockContent :: BlockProperties -> OrgParser (F Blocks)
 parsedBlockContent blkProps = try $ do
--- a/tests/Tests/Readers/Org.hs
+++ b/tests/Tests/Readers/Org.hs
@@ -1054,6 +1054,33 @@ tests =
                        "  where greeting = \"moin\"\n"
            in codeBlockWith attr' code'
 
+      , "Source block with indented code" =:
+           unlines [ "  #+BEGIN_SRC haskell"
+                   , "    main = putStrLn greeting"
+                   , "      where greeting = \"moin\""
+                   , "  #+END_SRC" ] =?>
+           let attr' = ("", ["haskell"], [])
+               code' = "main = putStrLn greeting\n" ++
+                       "  where greeting = \"moin\"\n"
+           in codeBlockWith attr' code'
+
+      , "Source block with tab-indented code" =:
+           unlines [ "\t#+BEGIN_SRC haskell"
+                   , "\tmain = putStrLn greeting"
+                   , "\t  where greeting = \"moin\""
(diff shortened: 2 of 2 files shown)


tarleb:
Looks really good, thanks!  I have a few minor comments which I'll add inline.

src/Text/Pandoc/Readers/Org.hs
```diff
@@ -407,11 +407,21 @@ orgBlock = try $ do
...
+blockHeaderStart = try $ (,) <$> spaceLength <*> blockType
  where
-  indent    = length      <$> many spaceChar
   blockType = map toLower <$> (stringAnyCase "#+begin_" *> orgArgWord)
 
+spaceLength :: OrgParser Int
+spaceLength = try $ do
+  tabStop  <- getOption readerTabStop
+  spaceStr <- many spaceChar
+  let charLen = \c ->
```
tarleb: I personally prefer to define functions like these in a `where` statement as it's less disrupting and a bit cleaner (IMHO).  The lambda function is not really necessary:

    charLen c = case c of
        ' '  -> 1
        '\t' -> tabStop
        _   -> 0

src/Text/Pandoc/Readers/Org.hs
```diff
@@ -407,11 +407,21 @@ orgBlock = try $ do
...
+        case c of
+          ' '  -> 1
+          '\t' -> tabStop
+          _    -> 0
+  return $ sum $ map charLen spaceStr
```
tarleb: My personal preference (and, I believe, a widely accepted convention) is to use `.` instead of `$` wherever possible.  So this would become `return . sum $ map charLen spaceStr` or maybe `return . sum . map charLen $ spaceStr`.

src/Text/Pandoc/Readers/Org.hs
```diff
@@ -472,7 +483,7 @@ rawBlockContent (indent, blockType) = try $
   unlines . map commaEscaped <$> manyTill indentedLine blockEnder
  where
    indentedLine = try $ ("" <$ blankline) <|> (indentWith indent *> anyLine)
-   blockEnder = try $ indentWith indent *> stringAnyCase ("#+end_" <> blockType)
+   blockEnder = try $ skipSpaces *> stringAnyCase ("#+end_" <> blockType)
```
tarleb: I'm torn on whether or not I like this change.  On the plus side, this change reflects how Emacs Org-mode handles code blocks.  On the other hand, checking indentation levels allows one to write nested code blocks like this:

    #+BEGIN_SRC org
        #+BEGIN_SRC haskell
        putStrLn "Hello"
        #+END_SRC
    #+END_SRC

I think Emacs is doing it wrong here, so I'd rather not change it unless there are any good arguments to do so.

src/Text/Pandoc/Readers/Org.hs
```diff
@@ -446,11 +456,12 @@ followingResultsBlock =
                                      *> block)
 
 codeBlock :: BlockProperties -> OrgParser (F Blocks)
-codeBlock blkProp = do
+codeBlock (_, blockType) = do
```
tarleb: The reason I wasn't pattern matching the argument was that this gives a slightly better abstraction level.  Now, if the `BlockProperties` type changes, one would have to change this function as well.

I'm okay with this change, as it leads to slightly simpler code, which is good.  The slightly cleaner, but probably less readable way would be define a new function for changing the indentation level of a `BlockProperty`.

tests/Tests/Readers/Org.hs
```diff
@@ -1054,6 +1054,33 @@ tests =
...
+                       "  where greeting = \"moin\"\n"
+           in codeBlockWith attr' code'
+
+      , "Empty source block" =:
+           unlines [ "  #+BEGIN_SRC haskell"
+                   , "  #+END_SRC" ] =?>
+           let attr' = ("", ["haskell"], [])
+               code' = ""
+           in codeBlockWith attr' code'
+
```
tarleb: Hooray for tests! :tada: :smiley: Much appreciated.

src/Text/Pandoc/Readers/Org.hs
```diff
@@ -407,11 +407,21 @@ orgBlock = try $ do
       _         -> withParsed (fmap $ divWithClass blkType)
 
 blockHeaderStart :: OrgParser (Int, String)
-blockHeaderStart = try $ (,) <$> indent <*> blockType
+blockHeaderStart = try $ (,) <$> spaceLength <*> blockType
  where
-  indent    = length      <$> many spaceChar
   blockType = map toLower <$> (stringAnyCase "#+begin_" *> orgArgWord)
 
+spaceLength :: OrgParser Int
```
tarleb: The "true" functional way would be to split this into two functions, e.g.
`indentation :: OrgParser Int` and `spaceLength :: Int String -> Int`.  The latter function would get the number of spaces per tab and the string as input, returning the converted number of whitspace.

shosti:
I changed the approach a bit based on the comments (`#END_SRC` now has to be indented at the same level as `#BEGIN_SRC`, which is actually how Emacs does it I think). lmk if the new approach looks sane. Thanks!

src/Text/Pandoc/Readers/Org.hs
jgm: +++ Albert Krewinkel [Apr 25 16 06:41 ]:
>   I think Emacs is doing it wrong here, so I'd rather not change it
>   unless there are any good arguments to do so.

Shouldn't we be trying to match Emacs' behavior?  After all,
that is the authoritative standard for org-mode.  One might
have views about how org-mode could have been designed
better, but unless you think Emacs' behavior is actually
buggy (relative to the org-mode documentation), I think
departing from it is a bad idea.

tarleb:
Great, thanks!

@jgm, +1 for merging this once travis is done.

src/Text/Pandoc/Readers/Org.hs
shosti: actually fwiw I think that Emacs' behavior is closer to `skipSpaces`. For instance,

```org
#+BEGIN_SRC org
  #+BEGIN_SRC haskell
  putStrLn "HELLO"
  #+END_SRC
#+END_SRC
```

Gets parsed as follows:

https://gist.github.com/9e1a533d377153cc5604374b53b2ea9a

(i.e., the first `END_SRC` does get interpreted as the end.)

tarleb: I find it very, very tempting to deviate from standard Emacs here, I
feel like nested blocks would be much nicer.  However, you are
absolutely right in that personal sense of æsthetics shouldn't get in
the way of proper functionality.

I checked again, the Org docs don't say anything about nested code
blocks.  The Emacs-way of expressing this would be to use the following
'*' escape syntax.  Which, I just realized, isn't supported by the
reader yet.

    #+BEGIN_SRC org
        *#+BEGIN_SRC haskell
        putStrLn "Hello"
        *#+END_SRC
    #+END_SRC

Welp, one more bug to fix.

@shosti: sorry about leading you in the wrong direction.  Would you
revert the PR to Emacs-comparatible behavior?  Maybe later, if @jgm
agrees, we could add a reader-option to allow nesting, your current
changes would come in handy then.

jgm: +++ Albert Krewinkel [Apr 26 16 05:01 ]:
>   Maybe later, if [3]@jgm agrees, we could add a reader-option to allow
>   nesting, your current changes would come in handy then.

Instead, I would suggest you try to persuade org-mode
maintainers to change Emacs' behavior.

shosti:
OK take 3 :smile: 

Look OK?

shosti:
Also FYI the travis build looks a bit funky: https://travis-ci.org/jgm/pandoc/jobs/125894786 (not sure if that's expected).

src/Text/Pandoc/Readers/Org.hs
tarleb: Just for completeness:  There is little to no chance that indentation-based nesting gets added to Emacs org-mode. Quting from the [mailing list](https://lists.gnu.org/archive/html/emacs-orgmode/2016-05/msg00059.html):

>> Is there any chance that indentation-based nesting will become legit
>> org-mode syntax?
>
> I don't think so. It would introduce unnecessary complexity.

The '*'s in above example should be commas and those are already supported.  Not sure why I thought it should be an asterisk.

Issue #2164, opened 2015-05-20
[Feature request] DocBook 5 support


and3k:
This would be nice. I guess it shouldn’t be very much effort, because DocBook 4.5 and 5 are very similar.

Thanks! :)

Bela

and3k:
PS: my temporary workaround is to use `xsltproc` and the `db4-upgrade.xsl` stylesheet included in DocBook 5

jgm:
Are you talking about input (reader) or output (writer)?

Can you give a summary of the changes from 4.5 to 5 that would affect pandoc?

and3k:
I’m talking about the writer. The reader probably doesn’t care if it’s 4.5 or 5, because they are so similar.

For sure a DocBook 5 template would be needed, because the header is different.

Additionally it should only be minor tweaks, like renaming tags or attributes (`ulink` to `link`, `id` to `xml:id`, …)

A list of all the changes: http://docbook.org/tdg5/en/html/ch01.html#introduction-whats-new

adeluccar:
Input would be cool too. I use pandoc to go from docbook4.5 to epub atm.

IvoBCD:
Had a stab at adding docbook5 writer support; see pull requests
https://github.com/jgm/pandoc/pull/2890 and https://github.com/jgm/pandoc-templates/pull/197

jgm:
Closing since we now have the writer support.

PR #2890, opened 2016-04-29
Docbook5 write support

IvoBCD:
- Extends docbook writer to add support for DocBook 5
- Avoids having to post-process DocBook 4.5 output with db4-upgrade.xsl
- Includes namespace in top-level element
- Use link with xlink:href attribute instead of ulink

Partially addresses feature request https://github.com/jgm/pandoc/issues/2164


IvoBCD:
CI build checks fail because of missing  /data/templates/default.docbook5

PR #2906, opened 2016-05-09
Allow spaces before '!' in MediaWiki table header

roblabla:
Added a test that fails on master pandoc, passes with my PR.


jgm:
Thanks!

PR #2908, opened 2016-05-10
Use the full MathJax configuration to maximize loading speed

ickc:
I used 2 different files for the following comparison:

|	| Time (ms)	| Loading Size (KB)	|  
| ------	| ------	| ------	|  
| MWE	| 381.44	| 391.9	|  
| MWE-full	| 360.75	| 401.0	|  
| primerv7  	| 5050	| 1280	| 
| primerv7-full	| 3840	| 1270	| 

The MWE is really small (`$$E=mc^2$$`), and the primerv7 is very long. i.e. they are used to test the 2 extreme conditions.

`-full` files means the `-full` configuration file is loaded (the proposed change).

You can see from the table that in the minimal case, loading the `-full` configuration only has a penalty of about 9KB out of 400 KB. But the speed increased in longer documents is significant. (and in that case the `-full` also smaller, probably because if the `.js` are loaded separately, the license appended in the beginning of each take up more spaces).

As the original motivation to not using `-full` is to not to load unnecessary files to save data usage. In the latest CHTML option, it seems the difference in size is really negligible. So I guess it is now better to load the `-full` instead.

I attached [the files used in the comparison here](https://github.com/jgm/pandoc/files/256288/MathJax.full.vs.not.full.zip).

Note: I use Safari to tests the size and Chrome to test the time.


jgm:
Sounds like a good idea.

jgm:
PS. In the future, please rebase the PRs so that they don't have merge commits.  It makes it easier to see what's going on later.

ickc:
Thanks for letting me know. Sorry for the ignorance.

jgm:
No, problem, thanks for the contribution!

Issue #2903, opened 2016-05-07
pdf conversion does not work if pdflatex is a batch file


janschulz:
There is a new conda package which installs miktex in a python virtualenv, but to include pdflatex and other commandline commands in the PATH, it creates batch files in a dir in the PATH. Unfortunately, it seems that pandoc expects a exe?

```bat
[pypandoc_35] λ where pdflatex
C:\portabel\miniconda\Scripts\pdflatex.bat

[pypandoc_35] λ pandoc --version
pandoc 1.17.0.2
[...]

[pypandoc_35] λ echo "ok" | pandoc -f markdown -t latex -o text.pdf
pandoc: pdflatex not found. pdflatex is needed for pdf output.

[pypandoc_35] λ set PATH=c:\portabel\miniconda\Library\miktex\miktex\bin\;%PATH%

[pypandoc_35] λ where pdflatex
c:\portabel\miniconda\Library\miktex\miktex\bin\pdflatex.exe
C:\portabel\miniconda\Scripts\pdflatex.bat

[pypandoc_35] λ echo "ok" | pandoc -f markdown -t latex -o text.pdf
```

It also works if I specify the full path to the batch file, so the batch files work:
```bat
[pypandoc_35] λ echo "ok" | pandoc --latex-engine=C:\portabel\miniconda\Scripts\pdflatex.bat -f markdown -t latex -o text.pdf
```

It seems that this can be traced down to the useage of `findExecutable` which seems to only consider `*.exe` files on windows :-( 

https://github.com/haskell/directory/blob/595f519d6f44191c52d00577568c6910635bb3d6/System/Directory.hs#L1047 ->
https://github.com/haskell/directory/blob/1f3e013cd0319ae501fa2c337cb4b0135f793e2a/System/Directory/Internal/Config.hs

Would it be possible to also check for *.bat on windows?

jgm:
findExecutable is from the directory library (and I don't
maintain that, so I can't change its behavior).
See its documentation here:
http://hackage.haskell.org/package/directory-1.2.6.3/docs/System-Directory.html#v:findExecutable
It calls a Windows function to find the executable.


+++ Jan Schulz [May 07 16 09:09 ]:

janschulz:
Would it be possible to work around it by doing the same what the directory library does but check for both `.bat` and `.exe`?

jgm:
`findExecutable` currently does this on Windows:
```hs
    Win32.searchPath Nothing binary exeExtension
```
[...]

PR #2910, opened 2016-05-10
Add appveyor artefacts for pandoc.exe

janschulz:
As discussed in https://github.com/jgm/pandoc/issues/2903#issuecomment-218214394

Example builds in https://ci.appveyor.com/project/JanSchulz/pandoc

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,7 @@ build: off
 before_test:
 - curl -ostack.zip -L --insecure http://www.stackage.org/stack/windows-i386
 - 7z x stack.zip stack.exe
+- git submodule update --init   # to fetch the templates
 
 clone_folder: "c:\\stack"
 environment:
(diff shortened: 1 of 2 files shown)


janschulz:
Oh my, that's the commit history you get when you have no idea about the language and the build tools and try them for the first time and only on the CI platform because your local build doesn't work...

janschulz:
squashed it down to two commits... The build, tests and in the end the artifact works, but there is some red in the build logs...

janschulz:
I think you have to log in at appveyor and enable the project there, too (or I didn't see a PR which was build after this :-)). First build takes almost an hour. second is much faster.

jgm:
OK, great, it built!  I did get a strange error message, though (see attached image).  Do you understand why?  It still seems to have created the zip file.
screen shot 2016-05-12 at 11 21 05 am

janschulz:
I've seen that too. Not sure what is causing this not how to fix it. I've seen appveyor examples where the stack command was used ot copy the result to a predefined dir and the 7z command can then take it from there.

E.g. this:

> --local-bin-path global option. Use to change where binaries get placed on a --copy-bins 

https://github.com/commercialhaskell/stack/blob/ff7a54e3d3b3c4f1a863377d068c65d383093ef9/ChangeLog.md

I'm trying that on https://github.com/janschulz/pandoc/ (master branch) -> https://ci.appveyor.com/project/JanSchulz/pandoc/build/1.0.28

janschulz:
See #2919 for a PR

jgm:
Sorry, made my last comment before seeing that your PR
addresses the build issue.

+++ Jan Schulz [May 12 16 16:28 ]:

PR #2912, opened 2016-05-11
Org reader: basic support for export settings

tarleb:
Org-mode allows to specify export settings via `#+OPTIONS` lines <http://orgmode.org/manual/Export-settings.html#Export-settings>.  Basic support for this is added to the reader.  Only toggling of the simple sub-/superscript syntax is supported as of now, other options are currently ignored.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -393,6 +393,7 @@ Library
                    Text.Pandoc.Readers.Odt.Generic.XMLConverter,
                    Text.Pandoc.Readers.Odt.Arrows.State,
                    Text.Pandoc.Readers.Odt.Arrows.Utils,
+                   Text.Pandoc.Readers.Org.ParserState,
                    Text.Pandoc.Writers.Shared,
                    Text.Pandoc.Asciify,
                    Text.Pandoc.MIME,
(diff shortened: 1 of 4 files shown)


kaushalmodi:
Awesome! Thanks for adding this. Where will this be documented in Pandoc? I don't know Haskell, but by cursorily looking at the code, it looks like it now be able to parse the `#+OPTIONS: ^:nil` option. I have the `org-export-with-sub-superscripts` variable set to `{}` which is equivalent to setting `#+OPTIONS: ^:{}` in all my org files.

So the question is:
- Do I need to add `#+OPTIONS: ^:{}` in all my files so that `a_b` exports as `a_b` and only if I do `a_{b}`, `b` exports as a subscript? *The same applies for the superscript case too.*
- Or is there a `pandoc` commandline option or configuration I will need to do to get this feature.

Thanks.

tarleb:
> Do I need to add `#+OPTIONS: ^:{}` in all my files so that a_b exports as a_b and only if I do a_{b}, b exports as a subscript?

Yes, one will have an `#OPTIONS` line in all org-files in which this setting should apply.  There won't be a command-line switch.

I didn't get around yet to writing wiki pages or documentation for Pandoc's Org-mode features.  It's on my list.

PR #2919, opened 2016-05-12
Better way to find the artifacts on appveyor

janschulz:
Also enable the build again (build: off just prevents a default msbuild based one)

See https://github.com/jgm/pandoc/pull/2910#issuecomment-218842826 for the problem with the artifacts.


janschulz:
Two questions:
* Do you want to add the appveyor badge somewhere, so that people can see the latest build from master to get prebuild binaries?
* Do you test latex conversation? If so I could add the conda miktex packages form conda (setup conda and then `conda install -c conda-forge miktex`)

PR #2927, opened 2016-05-19
Org reader support for ATTR_HTML statements

### LINKED ISSUE #1906: Org reader support for ATTR_HTML
Similar to #1877, it would be nice for the org reader to support `#+ATTR_HTML`. This was [mentioned on pandoc-discuss](https://groups.google.com/forum/#!topic/pandoc-discuss/lD6j8BUNFkQ) a while back, and seems useful for adding arbitrary HTML attributes to non-header elements; take [images, for example](http://orgmode.org/org.html#Images-in-HTML-export):

    #+CAPTION: A black cat stalking a spider
    #+ATTR_HTML: :alt cat/spider image :title Action! :align right
    [[./img/a.jpg]]

tarleb:
Arbitrary key-value pairs can be added to some block types using a `#+ATTR_HTML` line before the block.  Emacs Org-mode only includes these when exporting to HTML, but since we cannot make this distinction here, the attributes are always added.

The functionality is now supported for figures.

This closes #1906.

### DIFF
--- a/src/Text/Pandoc/Readers/Org.hs
+++ b/src/Text/Pandoc/Readers/Org.hs
@@ -37,8 +37,9 @@ import           Text.Pandoc.Error
 import           Text.Pandoc.Options
 import qualified Text.Pandoc.Parsing as P
 import           Text.Pandoc.Parsing hiding ( F, unF, askF, asksF, runF
-                                            , newline, orderedListMarker
-                                            , parseFromString, blanklines
+                                            , anyLine, blanklines, newline
+                                            , orderedListMarker
+                                            , parseFromString
                                             )
 import           Text.Pandoc.Readers.LaTeX (inlineCommand, rawLaTeXInline)
 import           Text.Pandoc.Readers.Org.ParserState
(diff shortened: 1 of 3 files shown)


robstewart57:
@tarleb 

> The functionality is now supported for figures.

Can you confirm that this currently therefore work for `ATTR_HTML` tags of org tables? E.g.

```
#+ATTR_HTML: :class table-dark
| h1 | h2 |
|----|----|
| c1 | c2 |
```

The HTML generated by hakyll (which uses pandoc to parse the org mode syntax) doesn't include the `table-dark` class to `<table>`. I'd expect to see in the generated HTML:

    <table class="table-dark">

Instead I see:

    <table>

Is this because pandoc ignores

    #+ATTR_HTML: :class table-dark

Or because hakyll isn't make use of it?

tarleb:
Thank you for finding this issue. However, this is an over 4 year old pull request. A better place to ask these question would be the pandoc-discuss mailing list.

Table attributes are not supported yet; pandoc's internal representation has been greatly improved in the current development version, so adding this is feasible. Please raise a new issue for us to keep track.

robstewart57:
@tarleb thanks for your reply.

> Please raise a new issue for us to keep track.

Done https://github.com/jgm/pandoc/issues/6441 .

PR #2967, opened 2016-06-07
Don't equate "Latin letters" with ASCII letters

bpj:
Hitherto the documentation for the `ascii_identifiers` extension
has referred to ASCII letters as "latin letters". While this is
correct in a restricted historical sense it is not correct in
Unicode terms. For The Unicode Standard version 8.0.0
[NamesList.txt][] contains 1473 lines which match the (perl)
regular expression `\bLATIN\b.*\bLETTER\b`, which number thus at
least roughly corresponds to the number of characters which count
as Latin letters in Unicode terms. Consequently the term "latin
letters" should be replaced with "ASCII letters". For clarity it
should also be mentioned that non-ASCII digits are stripped.

[NamesList.txt]: ftp://ftp.unicode.org/Public/UNIDATA/UnicodeData.txt

### DIFF
--- a/README
+++ b/README
@@ -3343,8 +3343,8 @@ pointy braces `<...>`.
 #### Extension: `ascii_identifiers` ####
 
 Causes the identifiers produced by `auto_identifiers` to be pure ASCII.
-Accents are stripped off of accented latin letters, and non-latin
-letters are omitted.
+Accents are stripped off of accented Latin letters corresponding to 
+ASCII letters, and non-ASCII letters and digits are omitted.
 
 #### Extension: `mmd_link_attributes` ####
 


bpj:
Perhaps it should be mentioned that non-ASCII letters with strikethroughs like ł  U+0142 LATIN SMALL LETTER L WITH STROKE which don't decompose to an ASCII letter plus combining marks are stripped.

Does Haskell have an equivalent of [Text::Unidecode](https://metacpan.org/pod/Text::Unidecode)? I suppose it wouldn't be too hard to write one using the same algorithm and [the data tables of Text::Unidecode](https://metacpan.org/source/SBURKE/Text-Unidecode-1.27/lib/Text/Unidecode), but it's a huge lot of data. 

@jgm do you think that it would be a good idea to write a Perl filter which ASCIIfies identifiers produced with `-ascii_identifiers` using Text::Unidecode, of course also doing `s/[^a-zA-Z:_]+/-/g` on what Text::Unidecode produces?

jgm:
No, we should not replace "latin letters" with "ASCII
letters."  The docs say:  "Accents are stripped off of
accented latin letters, and non-latin letters are omitted."

ASCII letters don't have accents, so this wouldn't make
much sense with the suggested replacement.  This extension
will change, for example, "föó" to "foo", but e.g. Chinese
letters will just be dropped.

PR #2968, opened 2016-06-07
Docbook writer: Declare xlink namespace in Docbook5 output

IvoBCD:
Declare 'xlink' namespace in Docbook5 XML output, since the xlink namespace is used for external links in Docbook 5.0 output.


jgm:
You'll probably need to adjust tests/writers.docbook5 as well.

IvoBCD:
@jgm But `tests/writers.docbook5` gets its namespace declarations from the `default.docbook5` template, which isn't affected by this change, which only affects docbook output with `--chapter`.

jgm:
Ah, okay.

+++ Ivo Clarysse [Jun 07 16 10:44 ]:

PR #2970, opened 2016-06-08
Textile reader: Added raw link parsing.

### LINKED ISSUE #2969: Support textile bare links
It looks like textile bare links are not supported. That would be nice if you can implement that.

Current behavior:

```sh
$ echo 'https://example.com/my_site#title' | pandoc -f textile -t markdown_github - -o -
https://example.com/my\_site\#title
```

Expected behavior:

```sh
$ echo 'https://example.com/my_site#title' | pandoc -f textile -t markdown_github - -o -
<https://example.com/my_site#title>
```

kayrus:
This improvement allows to parse raw URLs as links

resolves #2969

### DIFF
--- a/src/Text/Pandoc/Readers/Textile.hs
+++ b/src/Text/Pandoc/Readers/Textile.hs
@@ -384,7 +384,8 @@ inline = do
 
 -- | Inline parsers tried in order
 inlineParsers :: [Parser [Char] ParserState Inlines]
-inlineParsers = [ str
+inlineParsers = [ barelink
+                , str
                 , whitespace
                 , endline
                 , code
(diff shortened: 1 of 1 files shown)


kayrus:
There is a problem. If you use

```hs
(rawUrl, escapedUrl) <- (try uri <|> emailAddress)
```

pandoc will convert `user_mail@mail.com` to `[mailto:user_mail@mail.com](user_mail@mail.com)`, which is not correct, it should be `[user_mail@mail.com](mailto:user_mail@mail.com)`

if you omit `emailAddress` it will work with normal e-mail addresses, like `user@mail.com`, but not `user_mail@mail.com`, it will be transformed to `user\_mail@mail.com`. Not sure how exactly should I resolve that.

kayrus:
Decided to swap arguments.

jgm:
Can you point to a textile spec that says that plain URLs will be automatically linked?
I don't want to have pandoc behave differently than normal textile engines in processing textile.

kayrus:
Good point. Unfortunately I found only redmine raw link [explanation](http://www.redmine.org/projects/redmine/wiki/RedmineTextFormattingTextile#External-links)

Anyway, when pandoc converts raw link into markdown raw link - radmine displays it as abroken  link, i.e.: `http://example.com/#text` link will be converted into `http://example.com/\#text` link. Most probably it worth to add extra extension called "redmine"?

jgm:
+++ kayrus [Jun 20 16 12:05 ]:
>   Good point. Unfortunately I found only redmine raw link [1]explanation
>
>   Anyway, when pandoc converts raw link into markdown raw link - radmine
>   displays it as abroken link, i.e.: http://example.com/#text link will
>   be converted into http://example.com/\#text link. Most probably it
>   worth to add extra extension called "redmine"?

Sounds like a bug in redmine - why don't you report it to
them?  If they're autolinking URLs, then they should be
able to handle URLs like that.

kayrus:
It is not a bug, actually if it could be fixed, then redmine would just display `http://example.com/\#text` as a plain text. It is not an option.

kayrus:
@jgm is there an option to implement extension for textile which will convert bare links into hyperlink?

jgm:
I'm not keen on implementing extensions to the readers that
are not supported by standard tools for that format.
But if we did, it could be tied to the existing
autolink_bare_uris extension.

+++ kayrus [Jun 21 16 22:57 ]:

PR #2992, opened 2016-06-21
Org reader: remove partial functions

### LINKED ISSUE #2991: org-reader empty src block results in pandoc: Prelude.minimum: empty list
Hi there,

the latest changes to the org-reader are so cool! Thanks!
I think that a bug sneaked in:
when you create an org file with an empty src block, such as

```
* hi
#+BEGIN_SRC sh

#+END_SRC
```

you get from pandoc
```
$ pandoc -o pdf my.org
pandoc: Prelude.minimum: empty list
```
Not having empty blocks in the file solves the problem though.
I'm using pandoc 1.17.1
Compiled with texmath 0.8.6.3, highlighting-kate 0.6.2.
Thanks however!

tarleb:
Partial functions like `head` lead to avoidable errors and should be
avoided.  They are replaced with total functions.

This fixes #2991.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/Blocks.hs
+++ b/src/Text/Pandoc/Readers/Org/Blocks.hs
@@ -219,7 +219,7 @@ rawBlockContent blockType = try $ do
    stripIndent strs = map (drop (shortestIndent strs)) strs
 
    shortestIndent :: [String] -> Int
-   shortestIndent = minimum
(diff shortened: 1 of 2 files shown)


tarleb:
Incomplete

tarleb:
The travis failure seems to be unrelated to the proposed changes.

Issue #2999, opened 2016-06-26
org-reader: Parse image and table label


michelk:
- The org-reader does not read the image and table label.
- Cross-references are not rendered correctly.

cc @tarleb   

# Minimal example
## Org-file: 

```org
As illustrated in figure [[fig:myfig]], ... and listed in table [[tab:mytab]]

#+CAPTION: My figure.
#+LABEL: fig:myfig
[[blub.png]]


#+CAPTION: My table.
#+LABEL: tab:mytab
 | Name | Value |
 |------+-------|
 | Uno  |     1 |
 | Due  |     2 |

```

## Expected Latex
```latex
As illustrated in figure \ref{fig:myfig}, \ldots{} and listed in table
\ref{tab:mytab}

\begin{figure}[htbp]
\centering
\includegraphics{blub.png}
\caption{My figure}
\label{fig:myfig}
\end{figure}

\begin{longtable}[]{@{}ll@{}}
\caption{My table}\label{tab:mytab}\tabularnewline
\toprule
Name & Value\tabularnewline
\midrule
\endfirsthead
\toprule
Name & Value\tabularnewline
\midrule
\endhead
Uno & 1\tabularnewline
Due & 2\tabularnewline
\bottomrule
\end{longtable}
```

## Actual latex
```latex
As illustrated in figure \url{fig:myfig}, \ldots{} and listed in table
\url{tab:mytab}

\begin{figure}[htbp]
\centering
\includegraphics{blub.png}
\caption{My figure}
\end{figure}

\begin{longtable}[]{@{}ll@{}}
\caption{My table}\tabularnewline
\toprule
Name & Value\tabularnewline
\midrule
\endfirsthead
\toprule
Name & Value\tabularnewline
\midrule
\endhead
Uno & 1\tabularnewline
Due & 2\tabularnewline
\bottomrule
\end{longtable}
```

tarleb:
Thanks for the report.  This is a duplicate of #2496, though info in that threat aren't fully correct any longer.

PR #3001 adds labels for figures.  Labeling tables is not possible at the moment. (@jgm, please correct me if I'm wrong.)

jgm:
Closing in favor of #2496

michelk:
Org-mode

```org
As seen in fig. [[fig:1]]
```

Translates to latex:
```latex
As seen in fig. \url{fig:1}
```

I would expect
```latex
As seen in fig. \ref{fig:1}
```
Is it possible to fix this in the org-reader?

[...]

Issue #2496, opened 2015-11-05
org-mode reader does not support #+LABEL for figures and tables


sebhahn:
First I thought the issue is coming from ox-pandoc, an Emacs plugin for pandoc (I've posted my issue here https://github.com/kawabata/ox-pandoc/issues/12#issuecomment-154106856)

However, it seems that this problem is caused by pandoc. I'm using pandoc 1.15.1.1, pandoc-citeproc 0.8.0.1, emacs 24.5.1 and org-mode 8.3.2

e.g.
`#+CAPTION: insert figure caption here`
`#+NAME: fig:1`
`#+LABEL: fig:1`
`[[img.png]]`

does create

`\begin{figure}[htbp]`
`\centering`
`\includegraphics{img.png}`
`\caption{insert figure caption here}`
`\end{figure}`

and the reference `[@fig:1]` gets correctly translated to `fig.~\ref{fig:1}` (using pandoc-citeproc), but it doesn't work because `\label{fig:1}` is missing in the Latex output? Somehow `#+NAME` needs to be provided as well, otherwise I'll get an error message from pandoc-citeproc.

Exporting with standard org-mode export to latex gives:

`\caption{\label{fig:orgparagraph1} insert figure caption here}`

still not the right label, but at least it is in the latex code

tarleb:
Pandocs handling of figures is limited, labeled figures are unsupported atm. [This patch](https://github.com/tarleb/pandoc/commit/d2a5aee388b1042899167c45f9adcf35ffec1f0d) is probably the closest one can get, even though I'm not entirely sure if it does what you need.

sebhahn:
The result should look like

`\begin{figure}[htbp]`
`\centering`
`\includegraphics{img.png}`
`\caption{insert figure caption here}`
`\label{fig:1}`
`\end{figure}`

tarleb:
The problem is that this depends on issues  #813 and #1299.  People are working on those issues, but their resolution requires substantial changes in the underlying data structures and writers.  It's a complex and laborious task still pending completion.  As a result, there is currently no sane way to get proper figure references.  Sorry.

sebhahn:
I'm looking forward to these new features :smile:

tarleb:
Closing in favor of #813.

PR #3001, opened 2016-06-26
Org reader: support figure labels

tarleb:
Figure labels given as `#+LABEL: thelabel` are used as the ID of the
respective image.  This allows e.g. the LaTeX writer to add proper `\label`
markup.

This fixes half of #2496 and #2999.


michelk:
Great. Thanks a lot!

PR #3003, opened 2016-06-27
Handle NEL character in Markdown input.

wilx:


wilx:
I withdraw the patch.

PR #3005, opened 2016-06-29
Added support for Zim wiki syntax

alexivkin:
[Zim](http://zim-wiki.org/index.html) is a popular desktop wiki that uses a fairly simple text file [syntax] (http://zim-wiki.org/manual/Help/Wiki_Syntax.html). This writer is able to convert whatever pandoc reads into what zim expects. The reader is not needed as Zim uses pandoc to export to markdown and HTML natively.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "data/templates"]
+[submodule "templates"]
 	path = data/templates
-	url = https://github.com/alexivkin/pandoc-templates.git
+	url = https://github.com/jgm/pandoc-templates.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "templates"]
+[submodule "data/templates"]
 	path = data/templates
-	url = https://github.com/jgm/pandoc-templates.git
+	url = https://github.com/alexivkin/pandoc-templates.git
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -363,6 +363,7 @@ Library
                    Text.Pandoc.Writers.Textile,
                    Text.Pandoc.Writers.MediaWiki,
                    Text.Pandoc.Writers.DokuWiki,
+                   Text.Pandoc.Writers.ZimWiki,
                    Text.Pandoc.Writers.RTF,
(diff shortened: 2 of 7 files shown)


jgm:
Thanks.

1.  We need tests before I'd consider merging.  At minimum you should generate `tests/writer.zim`, check it carefully for correctness, and add to the "other writers" list around line 169 of  `tests/Tests/Old.hs` 

2.  On quick glance, the indentFromHTML business seemed strange -- what is this for?

alexivkin:
Added the tests and fixed table syntax a bit. indentFromHTML allows ordered lists to be controlled from raw html code. I.e. if the source file includes <li>, the converted list will add a new item.
This is necessary for conversion from mediawiki for example, that often drops down to straight html for ordered lists, because of the very rigid syntax.

alexivkin:
The builds in this branch will fail until the submodule is updated with the Zim wiki template with the pull 202 [https://github.com/jgm/pandoc-templates/pull/202]. 
I am new to submodules, so I created two separate pulls. If there is a way to do it smoother, let me know.

alexivkin:
The submodule was merged, but the actual zim code was not

jgm:
+++ Alex Ivkin [Jul 04 16 00:18 ]:
>   The submodule was merged, but the actual zim code was not

The code was all merged.  However, I rebased it first,
squashing to a single commit.  That's probably why
github didn't think it was merged.

See commits:
a73c95f61dbf47bbb54345855b07127d9fd82e62
7fdcd9a6e212dc02e0a37f47240e5978683d66a1
b3382cf3777dc121c61937e64890cc9646dc30bf

alexivkin:
Thank you. For the future, what is a proper way to create pull requests that include changes to submodules? Can it be squashed on my side into a single commit between two git repos?

jgm:
You need to submit a separate PR in the submodule
repository, just as you did.  (And indicate on your
other PR that it depends on the submodule change, as
you also did.)

So, what you did was just right.  The only thing I
changed was that I squashed several commits in pandoc
into one.

+++ Alex Ivkin [Jul 12 16 01:53 ]:

PR #3010, opened 2016-07-03
Org reader: support archived trees, headline levels export setting

tarleb:
Support for the `H` (maximum level of headlines) and `arch` (treatment of archived trees) settings is added to the reader.

The document is now parsed into a document tree consisting of headlines and their block contents nodes and subsections as subtrees.  This structure resembles that used by the Emacs Org mode parser and simplifies transformations of the document tree.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -399,6 +399,7 @@ Library
                    Text.Pandoc.Readers.Odt.Arrows.Utils,
                    Text.Pandoc.Readers.Org.BlockStarts,
                    Text.Pandoc.Readers.Org.Blocks,
+                   Text.Pandoc.Readers.Org.ExportSettings,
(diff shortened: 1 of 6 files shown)


tarleb:
I'm a bit unhappy with [this `case` expression](https://github.com/jgm/pandoc/pull/3010/files#diff-1579b2a2f52b1d13d177473e5166c501R140) as it requires the use of `-no-warn-overlapping-patterns` GHC option.  Is there a nicer way of expressing this?

jgm:
On the case expression, does it help to use this form?
```
case () of
    _ | conditionOne -> 1
       | conditionTwo -> 2
       | otherwise -> 3
```

tarleb:
I misread the compiler warning, the faulty code was in one of the case's conditions.  Sorry.  Fixed in #3012.

PR #3011, opened 2016-07-04
Resubmit of the merge 3005

alexivkin:
The submodule was merged, but the actual zim code was not.  This should trigger travis' revalidation.


jgm:
Your code has been merged; I rebased it first, which is why GitHub didn't close this automatically. But if you look at the commit log, you'll see it.

PR #3033, opened 2016-07-20
PoC: GitHub-optimized README

tarleb:
This isn't meant to be merged, but should serve as a basis for further discussions.

The rendered README can be found at
<https://github.com/tarleb/pandoc/blob/github-readme/README.md>


jgm:
I like this idea.

I suggest MANUAL.txt instead of just MANUAL.

Other things will need changing, too.  For example, README -> MANUAL.txt in Makefile (for creating the man page).

One thing I don't like is that, since the first part is duplicated in README and MANUAL.txt, it will need to be revised twice every time something changes.  Maybe there's a way around that.

jgm:
README is also referred to in make_osx_package.sh and in the windows and debian package creation scripts/Makefiles.

jgm:
s/guide-lines/guidelines/

tarleb:
I just couldn't think of a better way than duplicating some content.  I don't really like it either.

I *think* I found and properly replaced all references to the README in makefiles and scripts, though I'll admit that I'm not 100% on the way builds are executed.  Does Travis test this kind of thing?

tarleb:
Please also note that the proposed README links to <http//pandoc.org/MANUAL.html> which of cause doesn't exist yet.  In the new README, I also removed all links to specific sections in the manual as it seems difficult to keep those in a way that's acceptable both for Github rendering and for readers of the plain text file.

PR #3038, opened 2016-07-23
Use http to access online documentation

tarleb:
The link to the online manual used the `https` schema, which doesn't work for this resource.  Changed to `http`.


jgm:
Thanks.  I think the whole README.md change was a good idea!

PR #3045, opened 2016-07-29
Fix the LaTeX and EPUB links

mortonfox:
Parenthesized items following [ ] links need to be escaped. Otherwise, those will be turned into the URLs for those links.


jgm:
GitHub is very much in the minority in allowing this to be a link:
<http://johnmacfarlane.net/babelmark2/?normalize=1&text=%5Bhi%5D+(url)%0A>
and as we can see from this example, there's a good reason not to. Perhaps they could be persuaded to change, but in the mean time your fix is welcome!

PR #3070, opened 2016-08-14
Docx Writer: Implement user-defined styles.

jkr:
This supersedes #3069.

This is an RFC PR (I'll be writing to the mailing list) and shouldn't be considered final.

This would allow users to dynamically control the styles on docx output. People could either write divs by hand or -- more what I had in mind -- script to control indentation or whatever.

A few notes:

1. This requires divs, since we can't add attributes to arbitrary blocks. Unlike the previous PR, it will affect all blocks within the div, since the styles will cascade appropriately.

2. We use a k-v pair in the attribute: the value associated with the key "docx-style" will be passed through. (This takes precedence over the "references" class.)

3. As before, the styles won't do anything if they're not already in your reference.docx. In fact, they'll just show up as "Normal Text", and convert to that upon saving. So this is only useful if you have a reference.docx already.


jgm:
This looks good to me.  You might add to this PR some changes to MANUAL.txt, so this is documented, and of course tests.

jkr:
Bikeshedding: what do you think the key should be? I have `docx-style` in there right now, but it seems like something that could be equally usable for ODT and any other future binary formats (PPT,) would be preferable. `custom-style`, `pandoc-style` both come to mind.

The safest solution -- if we don't want to step on people's pass-through HTML -- would be to have a default key (`custom-style`, `pandoc-style`, whatever) and an option `--custom-style-key`, so that people could reset it if they for some reason wanted to reserve our key. But this does add yet another option, and could easily be added later if it ever becomes an issue.

jkr:
Some new changes: it works with spans as well, and will also inject styles into style.xml if they don't exist. If they are already there, it doesn't mess with them.

Issue #1852, opened 2015-01-02
Task list (e.g. for GSOC)


jgm:
Here is a list of projects that could be done in pandoc:
1.  Use `Text` instead of `String` throughout. This would involve API changes and extensive (but not too difficult) changes to
   - pandoc-types
   - pandoc
   - texmath
   - highlighting-kate
   - pandoc-citeproc
   
   Also worth considering whether it would help to use my custom text-based parser combinators from cheapskate (perhaps with some amplifications) instead of the slower parsec.
2.  <del>Add a Lines (or LineBlock) Block element, and modify readers, writers, and associated code.  Currently we just use a Para with LineBreaks. This is non-ideal, especially when converting _into_ formats allowing  line blocks. (See #1623)</del>
3.  Create a flexible system for labeling objects (images, tables, code) and referring back by number and/or link. Or, more minimally:  handle `\label` and `\ref` better when parsing LaTeX.  (See #813, #165)
4.  Automatic identifiers for images and tables in HTML output. (#208)  Must take care not to break existing documents relying on automatic identifiers for headers. (See also #1806.)
5.  Syntax and (block or inline?) element for anchors (see #1299, #1621, also http://talk.commonmark.org/t/turning-empty-link-definitions-into-anchors/893).
6.  <del>Allow attributes on links, images?  (WIP #1806)</del>
7.  <del>Add a Softbreak inline element to allow preservation of the orginal text's line breaking (#1701).</del> (in progress)
8.  Create a pure Haskell library to parse [commonmark](http://spec.commonmark.org), using the parsing strategy from the commonmark reference implementations (commonmark.js and cmark).  Make it extensible, and use it to add a new pandoc reader for commonmark.
9.  ~~Add a writer for groff ms format for a lightweight route to PDF (#1839).  [Note:  this is done in a branch, `groff_ms`, but it's not merged]~~
10.  Add a ConTeXt reader (#1423) -- should be possible to modify the LaTeX reader.
11.  Add a plain TeX writer (#1541).
12.  Make Format an enumerated type (#547).
13.  ~~Add a GNU `eqn` writer (and reader?) to texmath?~~
14.  pandoc-citeproc needs lots of attention.  The (inherited) code base is overly complex and very hard to understand in places.  It would be great if someone else were to get familiar with it, so more bugs could be fixed.
15.  Writer for the new native Kindle ebook format K8 (#1865).
16.  Find bugs from the tracker and fix them!

mpickering:
[...]

PR #3072, opened 2016-08-15
Read and write LineBlock elements

tarleb:
This is work based on the second point of #1852 and the discussion in #1545.

The LineBlock element is converted to a Para with lines separated by hard line breaks by default.

The following writers have implementations leveraging appropriate format features:

 - AsciiDoc
 - ConTeXt
 - Custom
 - HTML
 - Markdown 
 - Org
 - RST

It's also supported by the following readers:

  - Docbook5
  - RST
  - Markdown
  - Org

The PR depends on jgm/pandoc-types#22.


ousia:
@tarleb, just in case it might help.

I guess a way to implement in the HTML writer would be to ad a `<div class="lineblock">` with the following code in the cascading stylesheet:

``` css
.lineblock{ white-space:pre-line; }
```

I’m not a coder myself, but I guess this may work.

tarleb:
Thanks for the input @ousia.  The problem with that approach is that all HTML fragments would have to contain a CSS block defining the `.lineblock` class.  People using pandoc without the `--standalone` flag would get the wrong layout.

ousia:
@tarleb, I thought of `<div style="white-space: pre-line;">`, but after that I realized that `<br/>` is a much simpler approach.

Sorry for the noise.

jgm:
You could alternatively add a literal style attribute, instead of
using a class that requires CSS.  This might actually be a
cleaner approach than using br tags.

+++ Albert Krewinkel [Aug 17 16 10:34 ]:

tarleb:
The writer methods for Org and ConTeXt currently have problems with overly long LineBlock lines.  I haven't yet found a simple way to solve that.

tarleb:
I learned about `nowrap` from Text.Pandoc.Pretty, so the issues mentioned in the previous comment have been resolved.

The behavior of the Markdown/RST lineblock parser was changed to always parse those line blocks into a single element.  Previously, if blocks contained blank lines, they were parsed into multiple paragraphs.

The Docbook5 writer no longer uses `linegroup`, which is only part of the 5.1 draft.

jgm:
Could you rebase these changes on top of master?
Ideally we'd smush these changes into two or three commits: perhaps one adding LineBlock support to the writers, one to the readers, and one for the changes in Shared.
The commit comments should explain the changes to the various readers and writers.
You should note explicitly the API change (exporting a new function from Shared).  That will help immensely when I write the changelog, otherwise I'm likely to forget to document that.

tarleb:
It seemed right to separate the change in Parsing from the other modifications as well, so there is a fourths commit on top of the other three.  I hope the commit-messages are suitable for the changelog in their present form.

Also, thanks again @ousia for the `pre-line` trick.

src/Text/Pandoc/Shared.hs
```diff
@@ -630,6 +631,11 @@ compactify'DL items =
              | otherwise           -> items
            _                       -> items
 
+-- | Combine a list of lines by adding hard linebreaks.  This is useful
+--   e.g. for rudimentary support of LineBlock elements in writers.
+combineLines :: [[Inline]] -> [Inline]
+combineLines = concat . map (++ [LineBreak])
```
jgm: Do we really want a LineBreak at the end of the last line?
Why not

    combineLines = intercalate [LineBreak]

and this is getting so short that maybe we don't even need a separate function?

src/Text/Pandoc/Parsing.hs
```diff
@@ -718,11 +718,14 @@ lineBlockLine = try $ do
...
 
+blankLineBlockLine :: Stream [Char] m Char => ParserT [Char] st m Char
+blankLineBlockLine = try (char '|' >> blankline)
+
 -- | Parses an RST-style line block and returns a list of strings.
 lineBlockLines :: Stream [Char] m Char => ParserT [Char] st m [String]
 lineBlockLines = try $ do
-  lines' <- many1 lineBlockLine
-  skipMany1 $ blankline <|> try (char '|' >> blankline)
+  lines' <- many1 (lineBlockLine <|> ((:[]) <$> blankLineBlockLine))
```
jgm: If we allow blank lines, we may run into problems in output formats that use paragraphs for line blocks and don't allow blank lines in paragraphs.  But maybe there aren't any such formats?  Just wanted to check.

src/Text/Pandoc/Shared.hs
tarleb: My initial version did use `intercalate`, I do not quite remember why I changed it.  I think it was because rendering results were off in some formats (HTML), while it doesn't do any harm in other formats.  The switch in methods was my reason to introduce this function, exactly for the case where we'd want to change it again.

My current thoughts on it are that I'd rather add a function `linesToPara :: [[Inline]] -> Block` and export that instead.  This would save a tiny bit of boilerplate in the writers, as `combineLines` is almost exclusively used in combination with `Para`. Don't know why I didn't do this already.

src/Text/Pandoc/Parsing.hs
tarleb: AsciiDoc is the prime example here. This could be worked around, but changing the writer would probably take more time. I'll remote the commit from the PR.

jgm:
I looked at the pandoc manual; it doesn't say whether blank
lines are allowed in line blocks, but it does say that the
syntax is borrowed from reStructured Text.

Looking at reStructuredText,
http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#line-blocks
I see that line blocks *can* contain blank lines.

So, I think it would actually be best to allow them to
contain blanks, and figure out how to work around this
in writers where this is a problem.

+++ Albert Krewinkel [Oct 11 16 11:29 ]:

jgm:
I see.  I think that line breaks at the end of paragraphs
are generally bad form, and may even be impossible in some
formats.

If HTML rendering looks bad, it may be a CSS issue.
It would help to have an example of what you mean, if you
can remember.

+++ Albert Krewinkel [Oct 11 16 10:49 ]:

tarleb:
Okay, fixed the AsciiDoc writer (the other writers seem to be working), changed `combineLines` to use `intercalate`, and rebased to master.  I haven't done extensive testing yet, but am fairly confident that things should work.

jgm:
Great work!

Issue #2685, opened 2016-01-29
Support listing formats (-l, --list-formats)


waldyrious:
It would be nice to have an option to list the available input and output formats (like what is displayed at the top of `pandoc -h`). I was surprised this didn't exist, but couldn't find an existing issue or PR proposing this.

By the way, `pandoc --list` seems to hang in the command line, not sure what's happening there. In contrast, both `-l` and `--list-formats` correctly return "pandoc: unrecognized option `--list-formats'".

jgm:
pandoc --list

is interpreted as

pandoc --listings

since that's the only option that starts with --list.  So
it's just waiting for input.

+++ Waldir Pimenta [Jan 29 16 03:21 ]:

waldyrious:
I'm far from a CLI expert, but that behavior sounds odd to me. Was it intended to be that way, or is it just an implementation artifact?

jgm:
The option parsing library we use behaves this way (as do
many other standard option parsing libraries).

It's kind of handy to be able to say `--biblio` instead of
`--bibliography`, for example.

+++ Waldir Pimenta [Jan 29 16 10:03 ]:

jgm:
Since `--help` already lists the input and output formats, why is it important to have another option that (in effect) just emits part of the `--help` output?

waldyrious:
I believe not having the list option violates the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) in two ways: users may be expecting such an option to exist (as did I) and type `pandoc -l` or `pandoc --list`, which is a pretty common format; but most importantly, by typing `pandoc --list` and getting a prompt without any indication of what the program is expecting, the user will be confused and wonder if something broke. From a UX perspective, it seems to me that adding this option would be clearly a net positive.

tarleb:
Thanks for chiming in @waldyrious. It seems so me that the principle of least surprise doesn't apply here, there are many programs where `-l` and `--list` have very different meanings: To give some examples: With ssh the `-l` option is used to specify the user login name, with curl `--list` is short for an FTP specific option, gpg doesn't accept it because it's ambiguous, imagemagick's convert program prints a list of options instead of all supported formats, etc.

However, I do agree that it might be helpful to have all formats listed without the need to scroll back in the terminal window as one has to do with `--help`, so I'm also in favor of adding this.

waldyrious:
Thanks for the LDP link, @tarleb -- I've bookmarked it :)

[...]

Issue #1775, opened 2014-11-22
Provide a way to turn off `--smart` in LaTeX reader?


rjturner:
I have a large LaTeX project and I am using pandoc to convert it to Markdown for future maintenance in Markdown, thus I don't want any difficult-to-type special characters like true em-dashes, elipses, and curly quotes; I just want `---` for em-dash, `...` for ellipses, and `"`/`'` for quotes.

It appears impossible to generate "non-smart" Markdown output from LaTeX (a carefully reading of the Users Guide for `--smart` and `--no-latex-ligatures` would suggest that this is an enhancement request, not a bug report, though I am not certain on this).

Take the em-dash, e.g. "`A---B`". I get:

```
$ echo "A---B" | pandoc -f latex -t markdown
A—B # that's a true em-dash, &#x2014;
```

This is what I expect, at least with some combination of `--smart` and `--no-tex-ligatures` (or "`--no-smart`"?):

```
$ echo "A---B" | pandoc -f latex -t markdown
A---B
```

I think this is reasonable, given:

```
$ echo "A---B" | pandoc -f latex -t latex
A---B

$ echo "A---B" | pandoc -f markdown -t markdown
A---B

$ echo "A---B" | pandoc -f markdown -t latex
A---B
```

automatthew:
I believe this is related:

```latex
He said, ``In US English, prose, dialogue paragraphs preceding those which
continue a speaker's dialogue do not have closing quotation marks.

``It seems Pandoc does not handle this situation correctly.''

Really, I asked?
```

    pandoc -f latex smart.tex -t markdown -o smart.md

```markdown
He said, \`\`In US English, prose, dialogue paragraphs preceding those
which continue a speaker’s dialogue do not have closing quotation marks.

“It seems Pandoc does not handle this situation correctly.”

Really, I asked?
```

This is a blocker for using Pandoc to treat .tex as the master format to produce both ebooks and PDFs for print.  I'm thrilled with Pandoc in every other area.

automatthew:
And then I remembered I hadn't upgraded recently.

With pandoc 1.19.1_1

```markdown
He said, “In US English, prose, dialogue paragraphs preceding those
which continue a speaker’s dialogue do not have closing quotation marks.

“It seems Pandoc does not handle this situation correctly.”

Really, I asked?
```

jgm:
This is done in HEAD.  you can do `-f markdown-smart`

PR #3073, opened 2016-08-16
Add `--list-formats` option

codebje:
This PR adds a -l/--list-formats option to list the available input and output formats, addressing #2685.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -139,7 +139,8 @@ To convert `hello.html` from HTML to Markdown:
 Supported output formats are listed below under the `-t/--to` option.
 Supported input formats are listed below under the `-f/--from` option. Note
 that the `rst`, `textile`, `latex`, and `html` readers are not complete;
-there are some constructs that they do not parse.
+there are some constructs that they do not parse.  All input and output
+formats can be listed using the `-l/--list-foramts` option.
 
 If the input or output format is not specified explicitly, `pandoc`
 will attempt to guess it from the extensions of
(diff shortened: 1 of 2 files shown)


jgm:
Thanks for your contribution!

I'm not sure why this is needed, however, since `--help` already lists the formats.

I'd prefer not to add another command-line option just to print part of the existing help message.  Is there some specific reason for doing this?

Also, this PR seems to affect `--version` output as well.

codebje:
I added this PR to get my feet wet with the pandoc codebase, based on the linked issue.  The `--version` change seemed apropos given the syntax types were also listed there, but I'm happy to either back out that change, or let the whole PR go if you're satisfied with it being in `--help` only.

tarleb:
Mea culpa.  I misinterpreted what was said in the linked issue, causing me to classify it as beginner friendly instead of closing it.  I'm terribly sorry for wasting your time. :disappointed: 

The change itself looks very nice and clean. We'd be more than happy if you were to continue contributing.  If you still feel like it, may I suggest you take a look at #1775 instead?  It's another issue involving command line options but also leads a little more into the depths of pandoc's internals.

Again, thank you for contributing and apologies for the trouble.

jgm:
@codebje - thanks for getting your feet wet, and as @tarleb says, it's a solid PR.  I'm going to comment on the linked issue with some further ideas about changes that I think would be more useful.

codebje:
No problem at all, I'll close this one and have a look at some other stuff :-)

PR #3074, opened 2016-08-16
synchronize spacing of footnotes in help output

waldyrious:
- remove a space between `[` and `*` in the list of input formats, to match the list of output formats
- add space after the `*`s, for improved readability


jgm:
Thanks!

PR #3084, opened 2016-08-22
Support `yaml_metadata_block` extension in more formats

### LINKED ISSUE #1960: Feature request: Ability to specify a YAML metadata file for all reader types
An idea I thought would be useful. Many of the Readers have little or no way to set metadata, and the -M option on the commend line only accepts strings and not arbitrary YAML. Right now you can include a separate file of YAML metadata for Markdown formats (which is simple concatenated with the markdown files during parsing.)

This idea would be to specify a metadata file on the command line ("pandoc -Y metafile.yml" or something), which would be parsed separately and the contents added to the document metadata, regardless of the input file type.

Thanks

tarleb:
This allows to specify metadata using yaml blocks with input formats RST, Org-mode, and LaTeX when the `yaml_metadata_block` extension is specified.  Other than with Markdown, YAML blocks are not fully integrated into the language parser and may only appear at the top of a document.  This should add enough flexibility to pass metadata in the form of a YAML block, yet limits the extend this feature has on language parsing.

This closes #1960.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -830,8 +830,9 @@ Options affecting specific writers
 
 :   Use the specified image as the EPUB cover.  It is recommended
     that the image be less than 1000px in width and height. Note that
-    in a Markdown source document you can also specify `cover-image`
-    in a YAML metadata block (see [EPUB Metadata], below).
+    in a Markdown, RST, Org-mode, or LaTeX source document you can also
+    specify `cover-image` in a YAML metadata block (see [EPUB Metadata],
+    below).
 
 `--epub-metadata=`*FILE*
 
@@ -850,9 +851,10 @@ Options affecting specific writers
     id="BookId">` (a randomly generated UUID). Any of these may be
     overridden by elements in the metadata file.
 
-    Note: if the source document is Markdown, a YAML metadata block
-    in the document can be used instead.  See below under
-    [EPUB Metadata].
+    Note: if the source document supports the `yaml_metadata_block`
+    syntax extension (currently Markdown, RST, Org-mode, and LaTeX), a
+    YAML metadata block in the document can be used instead.  See below
+    under [EPUB Metadata].
 
 `--epub-embed-font=`*FILE*
 
@@ -1081,7 +1083,7 @@ directory (see `--data-dir`, above). *Exceptions:*
   `--reference-docx` to customize the output).
 
 Templates contain *variables*, which allow for the inclusion of
-arbitrary information at any point in the file. Variables may be set
+arbitrary information at any point in the file.  Variables may be set
 within the document using [YAML metadata blocks][Extension:
 `yaml_metadata_block`].  They may also be set at the
 command line using the `-V/--variable` option: variables set in this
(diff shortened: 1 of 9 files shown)


tarleb:
@jgm What do you think about the idea in general? Quite a few people asked about getting YAML blocks for org files in the past.  Do you see a better way to handle this?

jgm:
I'm conflicted about the idea.  Conceptually, there are two ways to think about YAML metadata.  Either (a) it's an integral part of the format or (b) it's separate from the format.  With option (b) you think of documents that start with YAML metadata as really two documents glued together, one a metadata bit and the other the document proper.

With pandoc's Markdown we took option (a), so it feels somewhat inconsistent to me to take option (b) with the other formats.  But of course, with them, option (b) is the *only* option we could take.

There's also a question what to do about formats that have their own ways to specify metadata (as RST does) -- how does this interact with YAML metadata?

tarleb:
Indeed, that conceptual view makes it look like like a bad idea.  I was coming at this from the user perspective, aiming for somewhat uniform behavior, but I know see that it's less consistent than I hoped it would be.

How do you feel about the alternative solutions proposed in #1960, e.g. a command-line option `--yaml-header` which takes an optional format?

There are two main use-cases for this feature: One are markup-formats that allow no or only limited inclusion of metadata (basically all of them), the other are programs like [hakyll](https://jaspervdj.be/hakyll/) which use pandoc but don't use YAML instead of the pandoc's metadata.

> There's also a question what to do about formats that have their own ways to specify metadata (as RST does) -- how does this interact with YAML metadata?

I'd prefer if YAML metadata were treated as default values which can be overwritten.

tarleb:
@jgm, do you think there is a chance to somehow use YAML blocks with arbitrary formats in pandoc 2? Otherwise I'll just close this.

jgm:
I'm still not decided about this; let's leave it open for now.

hoelterhof:
To contribute to this issue with another opinion I would like to say that I agree with tarlebs view. I would like to underline the user perspecitve on pandoc as a flexible and multi propose document converter. Not all languages support their own metadata blocks and somehow YAML is intended to be a universal metadata language. So there is a reason to follow both concepts: a) and b). 

Or let's put it in the view of usability: a user might easily expect YAML to be independent from the native file format, and it is difficult to explain why certain pandoc input formats should not support yaml...

mb21:
While I can see the appeal of simply prepending some yaml metadata to existing rst/org-mode/latex files to quickly get some metadata fields set, doing so would make those files unparsable for the native rst program, `pdflatex` etc. In fact you would have a `.tex` file that isn't actually valid TeX.

Not sure how often-requested this feature is, but would it be too much to ask users to save the metadata in a separate yaml file which can be supplied as an argument as proposed in #1960?

tarleb:
The main benefit of being able to put the YAML block into the file is that both pandoc and tools like hakyll can read it.  Here's one such example file: https://raw.githubusercontent.com/turboMaCk/turboMaCk.github.io/develop/posts/2016-12-21-org-mode-in-hakyll.org> ([rendered version](https://turbomack.github.io/posts/2016-12-21-org-mode-in-hakyll.html)).

(I'm not arguing your general point, which I'm agree with, I just want to provide some context.)

getreu:
I wrote a [pandoc-notetaking](https://github.com/getreu/pandoc-notetaking) tool for organizing notes in files. It uses the YAML header to synchronize the document title and subtitle with the filenames of notes. It relies on `pandoc` for rendering html. This feature #3084 would greatly improve my use case as it would enable the tool to support all markup languages that `pandoc` understands.

saurabhnanda:
Did this get merged? Can I use YAML metadata blocks in RST? (I tried, but it doesn't seem to be working - I'm not sure if I'm doing something wrong).

The docs don't _explicitly_ state that YAML metadata blocks work only with Markdown files (one can make an _implicit_ assumption because all the examples are related to markdown).

tarleb:
Not merged. You'd have to write a small wrapper around pandoc which splits the input into two temporary files, the YAML header and the content. The YAML would then be passed via `--metadata-file`.

PR #3096, opened 2016-08-30
Drop compatibility for GHC < 7.8

jkr:
This removes the Monoid compat, just for housekeeping. It was only necessary for base < 4.5, but that's ghc <= 7.2, which I don't think we're still supporting.

I wasn't able to test this with 7.4 on travis, but I can't see why it wouldn't work. But did want to run this by you before pushing it, in case there's something important I'm missing.


jkr:
I just posted a version to travis with 7.4 back in the travis.yaml, and it checked out. So this should be good.

tarleb:
The only comment I have regards imports grouping. Since many import lists have some loose structure or alphabetical order, simply replacing the old import statement would increase entropy in those cases.  Other than that, I'd be happy to see this change.

tarleb:
Could lines 129 ff in Pretty.hs be removed as well?

jkr:
Good point on Pretty.hs. I'll add that.

Before I start reordering imports, my sense from a quick look is that the only ones that had an order/grouping are yours and the org reader. Does it seem that way to you as well? (I want to make sure there's no ordering principle I'm missing).

@jgm -- do we have any roadmap for 7.4 eol? I know we're not testing with it anymore. Ubuntu LTS 12.04 ends in April 2017, which seems like a good stopping point for desktop users. I guess the question is whether we want to hold on to it through wheezy LTS, which I think goes until 2018. (Not a big deal, but it would allow us to clean up a lot CPP cruft.)

tarleb:
Yeah, Java brainwashing really got me.  I'm not the only one though, Matt (epub) and MarLin (odt) seem to follow similar conventions.

jgm:
I'm okay with giving up on ghc 7.4 support now.

jkr:
> I'm okay with giving up on ghc 7.4 support now.

Okay -- I might do a bit more cleanup while I'm at it, then. Just so I'm clear: the definition of supporting ghc X is passing our travis builds for it, right?

jgm:
I think that's as good a definition as any.
(Travis builds use -Wall -Werror)

If we are giving up on ghc 7.4 we could change the lower bound on base in pandoc.cabal, right?

jgm:
Also if we give up on ghc 7.4 that would free us up to use these extensions:
`MultiWayIf`
`LambdaCase`.

I think we need 7.8 for `PatternSynonyms`, but this might be a compelling reason to give up on 7.6 too.  See https://groups.google.com/d/msg/pandoc-discuss/k9p2CB3TGQ8/jtU-irC0g-oJ where I say (two years ago) that we might be able to require ghc 7.8 in a year or two...

IS there still a compelling reason to support 7.6?  Debian stable has 7.8 now, right?

jkr:
Nope -- stable is still on [7.6](https://packages.debian.org/stable/haskell/ghc).

jkr:
Though that does seem a bit less significant with stack in the mix. The only way you care about the ghc version is if you're running cabal, right? Or is there some official backport channel that would object?

Well, there could be the problem of mismatched pandoc and pandoc-types.

jkr:
Ah, but the problem is that stable doesn't have stack in the repos.

jgm:
I guess I didn't realize that because I just installed the
ghc 7.8 binary downloaded from the haskell site; it works fine
on debian stable.  So, it is definitely possible to compile
pandoc on a debian stable machine, you just have to install
the ghc binary first.

It could also be argued that if you're using stable, you've
already made a decision that keeps you from building some
of the latest and greatest software...

Hard to say.

+++ Jesse Rosenthal [Sep 01 16 06:21 ]:

jkr:
You also supply a static .deb, right? So, in theory, end-users should be able to have access to released versions. And folks who wanted to develop on stable could download a ghc binary or do apti-pinning, which is annoying but doable.

That all being said, my gut feeling is that supporting the version in debian stable is a good gesture. (Though I'm not entirely sure what it's gesturing toward.)

jgm:
Static deb, that is true.

Maybe worth asking on pandoc-discuss if anyone would miss
ghc 7.6 support?

tarleb:
FWIW: [The backports repo](https://backports.debian.org/) became an official part of the Debian ecosystem. [GHC in backports](https://packages.debian.org/jessie-backports/ghc) is currently version 7.10.3.  Apt-pinning ghc to recent versions became simple and painless as a result.

jgm:
All right.  I don't see any compelling reason to continue
supporting ghc < 7.8.  That would open things up for use of
pattern synonyms etc.

+++ Albert Krewinkel [Sep 02 16 02:42 ]:

jkr:
So, practically, that means:

- remove build from `.travis.yaml`
- bump base in `pandoc.cabal`
- change note 10 in `CONTRIBUTING.md`

Anything else? Should we add a note to `INSTALL` as well?

jgm:
A note in INSTALL would be good (just saying which versions
of ghc are supported.)

Would unsupporting 7.6 allow us to make any further
simplifications to CPP or compat?

+++ Jesse Rosenthal [Sep 02 16 04:30 ]:

jkr:
Probably -- I wanted to update the travis file so I could test for it. There are only two compat modules left: one for TagSoup, and one for time. I couldn't figure out the dependencies manually, so I might just bump the lower bounds on a branch and see what happens on travis.

Just to get it down in one place, here are all the MIN_VERSION conditionals in the src tree:

~~~
#if !(MIN_VERSION_base(4,8,0))
#if MIN_VERSION_base(4,8,0)
#if MIN_VERSION_binary(0,7,0)
#if MIN_VERSION_blaze_html(0,5,0)
#if MIN_VERSION_blaze_html(0,5,1)
#if MIN_VERSION_blaze_markup(0,6,3)
#if MIN_VERSION_haddock_library(1,2,0)
#if MIN_VERSION_haddock_library(1,4,0)
#if MIN_VERSION_hslua(0,4,0)
#if MIN_VERSION_http_client(0,4,18)
#if MIN_VERSION_http_client(0,4,30)
#if MIN_VERSION_tagsoup(0,13,0)
#if MIN_VERSION_time(1,5,0)
~~~

jkr:
Okay -- everything seems to work here. I've rebased and I'll push this all. I'll check the other compats and see if I can find other unnecessary CPP stuff in future commits.

jgm:
All right, this should simplify development going forward!

jkr:
Yep, the tagsoup compat can go as well. It's lower-bounded at 0.13.7 and the compat layer checked for min_version 0.13, which means it was always doing the same thing. I'll push that after it goes through the travis tests.

The time compat layer is only used if the "old-locale" flag is passed at build time. I can't find documentation around about this. What's the story around old-locale? I see that it had a deprication notice, which was then removed. The commit message suggests that this was a first step toward getting rid of old-locale. Any reason why we wouldn't do that now?

jkr:
There's no problem on the travis builds with setting the lower bound of `time` to 1.5 -- which would allow us to get rid of the time compat.

That being said, 7.8 *does* have `time` v. 1.4 as one of its [boot libraries](https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory), as do the 7.8-based haskell platforms. `time` 1.5 does build fine with ghc 7.8; it just doesn't ship with it.

So, the question is: does supporting a ghc version mean we have to support its boot libraries? I'd say no, but I just want to see what you think.

tarleb:
> I'd say no

+1

There is [one test](https://github.com/jgm/pandoc-types/blob/master/Text/Pandoc/Builder.hs#L182) for base >= 4.5 in pandoc-types, should we apply the changes to that package too?

jkr:
> There is one test for base >= 4.5 in pandoc-types, should we apply the changes to that package too?

Yes, I think so. Both for parity with pandoc and because from my understanding of the PatternSynonyms extension, we'd actually be using it there and exporting it. Or at least, that's what I got from Matt's original suggestion.

jgm:
Let's stay safe and support boot libraries.  Otherwise it can cause problems for those not using sandboxes.

jkr:
Okay -- I'll put a commented note in the time compat module, then, specifying that we'll support it as long as we support 7.8, but that it can be obsoleted when support for 7.8 is dropped.

PR #3097, opened 2016-08-30
Add \begin{html} environment to LaTeX reader

timheap:
This allows embedding raw HTML blocks in a LaTeX document for conversion
to HTML, Markdown, or other formats that support HTML.

``` latex
\begin{html}
<h1>Hello, World!</h1>
\end{html}
```

This is very useful to me for a project I am working on where I want to write in LaTeX, but need to produce HTML documents as a final result. There are some things, like YouTube embeds, where sticking some raw HTML in the LaTeX document is the simplest solution. The `\begin{html}` environment does not actually exist in LaTeX, nor do I know of something similar that I could have used instead, so I made up my own environment. I understand if pandoc does not want to extend LaTeX in this manner.

Another possible option would be an environment such as `\begin{raw}[format]`, which would allow embedding fragments of any format in a LaTeX document for similar purposes:

``` latex
\begin{raw}[html]
<h1>Hello, world!</h1>
\end{raw}
\begin{raw}[markdown]
# Hello, world!
\end{raw}
```

And thirdly, another option would be to run pandoc over the contents of the raw blocks, so that:

``` latex
\section{LaTeX heading}
\begin{raw}[html]
<h1>HTML heading</h1>
\end{raw}
```

would have an internal representation something like:

``` haskell
[ Header 1 ("",[],[]) "LaTeX heading"
, Header 2 ("",[],[]) "HTML heading"
]
```

I went with the option that I've submitted because it was quick and dirty, and works for what I need. If this option or one of the other options sound like something that you want in pandoc, I am happy make any changes.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -1132,6 +1132,7 @@ environments = M.fromList
   , ("align*", mathEnv para (Just "aligned") "align*")
   , ("alignat", mathEnv para (Just "aligned") "alignat")
   , ("alignat*", mathEnv para (Just "aligned") "alignat*")
+  , ("html", (rawBlock "html") <$> verbEnv "html")
   ]
 
 letterContents :: LP Blocks
--- a/tests/Tests/Readers/LaTeX.hs
+++ b/tests/Tests/Readers/LaTeX.hs
@@ -115,6 +115,11 @@ tests = [ testGroup "basic"
             concat ["^^"++[i] | i <- hex] =?>
             para (str $ ['p'..'y']++['!'..'&'])
           ]
+        , testGroup "Raw HTML"
+          [ "Raw HTML environment" =:
+            "\\begin{html}<div class='test'>Hello, world!</div>\\end{html}" =?>
+            rawBlock "html" "<div class='test'>Hello, world!</div>"
+          ]
         ]
 
 baseCitation :: Citation
--- a/tests/latex-reader.latex
+++ b/tests/latex-reader.latex
@@ -845,4 +845,10 @@ indented.
 
 \$ \% \& \# \_ \{ \}
 
+\section{Raw HTML}
+
+\begin{html}
+  <div class="test">Hello, world!</div>
+\end{html}
+
 \end{document}
--- a/tests/latex-reader.native
+++ b/tests/latex-reader.native
@@ -372,4 +372,6 @@ Pandoc (Meta {unMeta = fromList [("author",MetaList [MetaInlines [Str "John",Spa
  [[Para [Str "And",Space,Str "in",Space,Str "list",Space,Str "items.",Note [Para [Str "In",Space,Str "list."]]]]]
 ,Para [Str "This",Space,Str "paragraph",Space,Str "should",Space,Str "not",Space,Str "be",Space,Str "part",Space,Str "of",Space,Str "the",Space,Str "note,",Space,Str "as",Space,Str "it",Space,Str "is",Space,Str "not",SoftBreak,Str "indented."]
 ,Header 1 ("escaped-characters",[],[]) [Str "Escaped",Space,Str "characters"]
-,Para [Str "$",Space,Str "%",Space,Str "&",Space,Str "#",Space,Str "_",Space,Str "{",Space,Str "}"]]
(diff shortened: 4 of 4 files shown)


adamnyberg:
Any progress on this? This is something that I'd be very helpful for me.

tarleb:
Idea and execution look nice, but introducing a non-existing LaTeX environment seems problematic.  If this was to be included, it would have to be documented in the MANUAL.

 An alternative might be to use a `verbatim` environment and [filters](http://pandoc.org/scripting.html) instead.  You could write a filter that checks for `CodeBlock`s containing HTML tags, and convert the content of those blocks using `readHtml` from module `Text.Pandoc.Readers.Html`.  This doesn't require any changes to pandoc and has the added advantage that the input document will still be valid LaTeX.

tarleb:
Here's a quick example of what I have in mind.  It converts all `CodeBlock`s to HTML, so one would have to add some kind of heuristic if that's unwanted.

``` haskell
#!/usr/bin/env runhaskell
import Data.Default
import Text.Pandoc.JSON
import Text.Pandoc.Readers.HTML ( readHtml )

main = toJSONFilter convertCodeBlocks

convertCodeBlocks :: Block -> [Block]
convertCodeBlocks b =
  case b of
    (CodeBlock _ code) -> case readHtml def code of
                               Left _              -> [b]
                               Right (Pandoc _ bs) -> bs
    _                  -> [b]
```

timheap:
I will test your suggestion with what I need to use it for later, but that looks like an acceptable solution for me at first glance.

tarleb:
The filter-based solution seems to be acceptable.  Please reopen if that is not the case.

timheap:
The project this PR was initially for has finally started, so I had a chance to try your suggested filter. I modified it a bit to suit my needs. It is included below in case it can help someone else:

```haskell
import Data.Char ( isSpace )
import Data.Default ( def )
import Data.List ( dropWhileEnd )
import Data.Maybe ( fromMaybe )

import Text.Pandoc.Error ( PandocError(..) )
import Text.Pandoc.JSON ( Block(CodeBlock, RawBlock), Format(Format), toJSONFilter )
import Text.Pandoc.Readers.HTML ( readHtml )

main = toJSONFilter convertCodeBlocks

type Reader = String -> Maybe [Block]

convertCodeBlocks :: Block -> [Block]
convertCodeBlocks b = case b of
    (CodeBlock _ code) -> fromMaybe [b] $ convertCode code
    _                  -> [b]

-- Given a block of code, try to convert it into something nicer
convertCode :: String -> Maybe [Block]
convertCode code =
    codeBlockType codeHead >>= codeBlockReader >>= runReader codeTail
  where (codeHead, codeTail) = splitWhen (== '\n') $ strip code

-- Get the reader name from the magic header
codeBlockType :: String -> Maybe String
codeBlockType line = if percent == "%%% "
    then Just $ strip typeName
    else Nothing
  where (percent, typeName) = splitAt 4 $ strip line

-- All the known readers and their name from the magic header
codeBlockReader :: String -> Maybe Reader
codeBlockReader "html" = Just rawHtml
codeBlockReader _      = Nothing

-- Make a raw HTML block from some string
rawHtml html = Just [RawBlock (Format "html") html]

runReader :: String -> Reader -> Maybe [Block]
runReader code reader = reader code

-- Strip whitespace from the start and end of a string
strip = dropWhileEnd isSpace . dropWhile isSpace

-- Split an array at a predicate, returning (before, after)
splitWhen :: (a -> Bool) -> [a] -> ([a], [a])
splitWhen pred [] = ([], [])
splitWhen pred (x:xs) = if pred x
    then ([], xs)
    else let (before, after) = splitWhen pred xs in (x:before, after)
```

Use like:

```latex
\begin{verbatim}
    %%% html
    <iframe width="560" height="315" src="https://www.youtube.com/embed/dQw4w9WgXcQ" frameborder="0" allowfullscreen></iframe>
\end{verbatim}
```

I am still learning Haskell, so any improvements to the above are welcome.

tarleb:
Thank you for posting the script.  Short and clear programs like this are good candidates for the [code review stackexchange](https://codereview.stackexchange.com) site, it's oftentimes a nice way to gather constructive criticism.

Two notes regarding library functions: `splitWhen` could also be written using Prelude's `break` or `span` functions.  No need to do it though, it's only used to split up lines, which is what the `lines` function does.

PR #3108, opened 2016-09-06
Add command line option allowing to set type of top-level divisions

tarleb:
Add --parts command line argument.
This only effects LaTeX writer, and only for non-beamer output formats.
It changes the output levels so the top level is 'part', the next
'chapter' and then into sections.

This is a rebased and slightly altered version of PR #1554.  @cwoac, I hope you don't mind the changes.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -661,16 +661,19 @@ Options affecting specific writers
 
 `--chapters`
 
-:   Treat top-level headers as chapters in LaTeX, ConTeXt, and DocBook
-    output.  When the LaTeX document class is set to `report`, `book`,
-    or `memoir` (unless the `article` option is specified), this
-    option is implied.  If `beamer` is the output format, top-level
-    headers will become `\part{..}`.
+:   Deprecated synonym for `--top-level-division=chapter`.
 
-`--parts`
-:   Treat top-level headers as parts in LaTeX output.  The second level
-    headers will be chapters, i.e. `--chapters` is implied.  This does not
-    effect the `beamer` output format.
+`--top-level-division=[section|chapter|part]`
+
+:   Treat top-level headers as the given division type in LaTeX, ConTeXt,
+    DocBook, and  TEI output. The hierarchy order is part, chapter, then section;
+    all headers are shifted such that the top-level header becomes the specified
+    type.  The default is `section`. When the LaTeX document class is set to
+    `report`, `book`, or `memoir` (unless the `article` option is specified),
+    `chapter` is implied as the setting for this option.  If `beamer` is the
+    output format, specifying either `chapter` or `part` will cause top-level
+    headers to become `\part{..}`, while second-level headers remain as their
+    default type.
 
 `-N`, `--number-sections`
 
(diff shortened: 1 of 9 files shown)


cwoac:
I don't mind, but you have left parts out (readme, changelog, etc).

tarleb:
Thanks for the quick feedback!

> but you have left parts out (readme, changelog, etc)

I tried to compile a minimal patch, but the definition of that is of cause a bit subjective.  Apologies if I removed to much.  I won't mind if you'd rather update your own PR, I'd close this one then.  The changelog is now mostly generated from git, so changing it manually probably causes more work than it saves these days.

MANUAL.txt
```diff
@@ -647,6 +647,11 @@ Options affecting specific writers
     option is implied.  If `beamer` is the output format, top-level
     headers will become `\part{..}`.
 
+`--parts`
+:   Treat top-level headers as parts in LaTeX output.  The second level
+    headers will be chapters, i.e. `--chapters` is implied.  This does not
+    effect the `beamer` output format.
```
jgm: spelling: effect -> affect

src/Text/Pandoc/Options.hs
```diff
@@ -363,6 +363,7 @@ data WriterOptions = WriterOptions
   , writerBeamer           :: Bool       -- ^ Produce beamer LaTeX slide show
   , writerSlideLevel       :: Maybe Int  -- ^ Force header level of slides
   , writerChapters         :: Bool       -- ^ Use "chapter" for top-level sects
+  , writerParts            :: Bool       -- ^ Use "part" for top-level sects in LaTeX
```
jgm: I don't really like having two independent booleans for chapters and parts.  It would be better to have a type that can be set to sections, chapters, or parts.  But perhaps this is the best approach for backwards compatibility.

tarleb: > I don't really like having two independent booleans for chapters and parts.

I was afraid you'd say that.  I gave it a shot and added `--top-level-division=[section|chapter|parts]`, deprecating `--chapters`.  The new MANUAL section probably needs the attention of a native speaker.

jgm: Of course, if we touch anything in WriterOptions, we'll
need to increment the second version number (1.18).
So we might consider doing another 1.17.* refresh
before merging this change.

jgm:
@tarleb do you think this is okay to merge now?
It's okay that it's not implemented in every writer, but the MANUAL should clearly document which writers are affected by the new option.
Note that because of the changes to WriterOptions, we'd need to bump the pandoc version to 1.18 if we merged this.

tarleb:
I'll have some time tomorrow, so I can rebase, test, and possibly extend this then.  I think it's ok the way it is now, but I'd rather make sure.

tarleb:
Okay, the branch has been rebased. I also added some unit tests (and found, fixed a bug that way).  As far as I am concerned, this is ready to merge now.

jgm:
Thanks!

PR #3142, opened 2016-10-04
Replace Google Chart API by CodeCogs

ickc:
Proposing to replace Google Chart API by CodeCogs:

Google Chart API is deprecated on April 20, 2015, although they are not taking it offline yet:

>While the dynamic and interactive Google Charts are actively maintained, we officially deprecated the static Google Image Charts way back in 2012. This gives us the right to turn it off without notice, although we have no plans to do so.

But since it can be taken down at any moment, we better switch to alternatives before that.

There are some relevant discussions in adam-p/markdown-here#144 and adam-p/markdown-here#261 regarding alternatives of Google Chart API.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -962,7 +962,7 @@ Math rendering in HTML
 
 :   Render TeX formulas using an external script that converts TeX
     formulas to images. The formula will be concatenated with the URL
-    provided. If *URL* is not specified, the Google Chart API will be used.
+    provided. If *URL* is not specified, the CodeCogs will be used.
     Note:  the `--webtex` option will affect Markdown output
     as well as HTML.
 
@@ -2715,7 +2715,7 @@ Docx
 
 FictionBook2
   ~ If the `--webtex` option is used, formulas are rendered as images
-    using Google Charts or other compatible web service, downloaded
+    using CodeCogss or other compatible web service, downloaded
     and embedded in the e-book. Otherwise, they will appear verbatim.
 
 HTML, Slidy, DZSlides, S5, EPUB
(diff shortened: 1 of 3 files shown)


jgm:
I think I'd prefer to stick with Google Charts until it actually disappears.
I don't know anything about CodeCogs, and for all we know it might disappear before Google Charts!
Anyway, if Google Charts does disappear, you can always specify the CodeCogs url on the command line.

ickc:
The problem is the momment it disappears, it doesn't only affect the documents created after that point (actually it won't, as you pointed out, people will change the url manually before pandoc will be updated), but it affects all documents generated before the that momment. Whatever the purpose of the generated documents is, email or html or etc., the fact it relies on webtex means it is going to be shared through the web. In the case of email (which concerns markdown-here) one cannot go to the recipients to update it. And in the case of html, we all know that for many websites people don't have the resources to constantly go back to maintain it.

So the suggested change is to prevent all those documents (generated between now and the moment it dies) becoming unrenderable in the future.

The user always has an option to choose the URL (as I did in the case of markdown-here which they default to Google Chart API. The developers there said they will update the extension to default to CodeCogs later). But the point of default is to provide the "best" option for people that do not want to research into all the available options and find the best.

Which one is the "best" becomes the controversial and subjective one. To me if a service is officially dead on April 2015 and not being killed yet by the mercy of Google is very fragile. One day CodeCogs might also dies, but at least there wasn't plan to do so.

P.S. **Personally** I really hate a service being killed. The moment Google announced they're killing Google Reader, I jumped ship to Feedly immediately. And for example when Dropbox announced killing Mailbox, I jumped ship to Outlook immediately. I would not wait until it dies and swap to a new service, no matter how good they are (the previous 2 examples are still best in class and non-substitutable), let alone a zombie service (officially dead but kept alive). **But this is just me.**

P.S.2 I knew this is going to be controversial. But rather than spending a lot of time to discuss, I decided to made a pull request. Even if it is rejected at least the decision is made immediately. And when the mind is changed one can always go back here. And by the way, I respect your final decision. It's not important to me as I don't use webtex in pandoc.

ickc:
FYI, GmailTeX already made the move to change from Google Chart API to CodeCogs (and they even added retina images which basically doubled the resolution from the default). The general consesus out there *seems* to be to replace Google Chart API by CodeCogs.

jgm:
OK, I think I got a bit confused, since Google Charts is still alive -- it's just the static png api that is deprecated.

OK, I guess it might make sense to do this change.  I don't know anything about CodeCogs though.
Can you reopen this?

PR #3144, opened 2016-10-04
Replace Google Chart API by CodeCogs

ickc:
I can't reopen #3142. So I open a new one here.

Google Chart API has already been deprecated on April 20, 2015. It is now a "zombie" service:

>While the dynamic and interactive Google Charts are actively maintained, we officially deprecated the static Google Image Charts way back in 2012. This gives us the right to turn it off without notice, although we have no plans to do so.
> from [Image Charts  |  Google Developers](https://developers.google.com/chart/image/)

Info on CodeCogs:

1. [Usage Policy | CodeCogs Equation Editor](http://www.codecogs.com/latex/usage.php): **up to 3,000 equations a day for testing and non-commercial use**
2. [Implement an API call to display LaTeX as inline image - Meta Stack Exchange](http://meta.stackexchange.com/questions/53436/implement-an-api-call-to-display-latex-as-inline-image): from here we see CodeCogs renders better than Google Chart API. (not the reason of the pull request here but a nice bonus.
3. adam-p/markdown-here#144: markdown-here developers has some insightful discussions about replacing Google Chart API by CodeCogs there

I also found a post on usage limit on Google Chart API: [Lifting the limit on calls to the Google Chart API - The official Google Code blog](http://googlecode.blogspot.fr/2008/03/lifting-limit-on-calls-to-google-chart.html), which was 50,000 queries per user per day but lifted practically to 250,000 queries per day.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2715,7 +2715,7 @@ Docx
 
 FictionBook2
   ~ If the `--webtex` option is used, formulas are rendered as images
-    using CodeCogss or other compatible web service, downloaded
+    using CodeCogs or other compatible web service, downloaded
     and embedded in the e-book. Otherwise, they will appear verbatim.
 
 HTML, Slidy, DZSlides, S5, EPUB
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -962,7 +962,7 @@ Math rendering in HTML
 
 :   Render TeX formulas using an external script that converts TeX
     formulas to images. The formula will be concatenated with the URL
-    provided. If *URL* is not specified, the Google Chart API will be used.
+    provided. If *URL* is not specified, the CodeCogs will be used.
     Note:  the `--webtex` option will affect Markdown output
     as well as HTML.
 
@@ -2715,7 +2715,7 @@ Docx
 
 FictionBook2
   ~ If the `--webtex` option is used, formulas are rendered as images
-    using Google Charts or other compatible web service, downloaded
+    using CodeCogss or other compatible web service, downloaded
     and embedded in the e-book. Otherwise, they will appear verbatim.
 
 HTML, Slidy, DZSlides, S5, EPUB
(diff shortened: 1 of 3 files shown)


jgm:
I wonder if the 3000 equation/day limit is going to be a
problem for some users?

Suppose you have a math paper with, say, 100 equations.
More than 30 views per day, and you'd be over the limit,
right?  That might be a reason to stay with Google Chart
for now, given that they have "no plans" to turn it off.

+++ ickc [Oct 04 16 04:41 ]:
>   I also found a post on usage limit on Google Chart API: [6]Lifting the
>   limit on calls to the Google Chart API - The official Google Code blog,
>   which was 50,000 queries per user per day but lifted practically to
>   250,000 queries per day.

ickc:
I didn't present the usage limit correctly. The full sentence is:

>you can generation up to 3,000 equations a day for testing and non-commercial use.

I think the reason I trimmed the sentence yesterday is the wrong grammar that confused me. It should mean "you can generate up to 3,000 equations a day".

There's another line down there:

>Notify us in advance if you need to make more than 20,000 calls in any one day.

If I understand correctly, each time you call a **new** equation, say, `"https://latex.codecogs.com/png.latex?1%21%2B2%21%5Cneq3%21" alt="1!+2!\neq3!" title="1!+2!\neq3!"`, CodeCogs needs to render it first since no one has sent the same request to the server yet. But once they rendered it, they will cache it on their server for future use. So the next time I call `"https://latex.codecogs.com/png.latex?1%21%2B2%21%5Cneq3%21" alt="1!+2!\neq3!" title="1!+2!\neq3!"` again, it doesn't need to generate it, so it doesn't count towards the 3,000 **generations of equation** a day. But now it is the 2nd time I called it, it counts towards the 20,000 calls in any one day.

So the Google Chart API's practical limit of "250,000 queries per day" should be compared to CodeCogs' "20,000 calls in any one day". *But*, CodeCogs has an additional limitation of "you can generate up to 3,000 equations a day".

Using your example, if a paper is wrote to include 100 new equations, that's 100 generations on the day it is published. Then on an average day of 30 visitors, the visitors made 3,000 calls to their server. Both are much lower than the limit.

So I think if it is for personal use only, it is very hard to beat the 3,000 generations limit. The concern arrises in the 2nd reference above, because they were proposing to render all equations generated in stackexchange.com by CodeCogs, which could easily surpasses the 3,000 generations limit because of a lot of users.

So for personal use, the number one should concern should be "20,000 calls in any one day". While it is an order of magnitude lower than Google's. I think it is still reasonably high.

ickc:
This is a sidetrack:

I just discovered a trick to put math in markdown that don't support math. It seems to be intentional as the manual said

>Note: the `--webtex` option will affect Markdown output as well as HTML.

For example, in the `temp.md`:

```markdown
# Important Discovery!

$1+2\neq3!$

Try it!
```

Run this:

```bash
pandoc --atx-headers --webtex=https://latex.codecogs.com/png.latex? -s -o temp-codecogs.md temp.md
```

Then the output becomes:

> # Important Discovery!
> 
> 1+2\\neq3!
> 
> Try it!

It seems to me that there are instances like this that a neat trick is hide in plain sight in the manual. Although I read it before, but I never knew the "note" there meant this. I think the situation could be improved. For now, I will add a wiki page to document this.

PR #3146, opened 2016-10-05
[ODT Parser] Include list's starting value

hubertp-lshift:
Previously the starting value of the lists' items has been
hardcoded to 1. In reality ODT's list style definition can
provide a new starting value in one of its attributes.

Writers already handle the modified start value so no need
to change anything in that area.


src/Text/Pandoc/Readers/Odt/StyleReader.hs
```diff
@@ -578,25 +580,30 @@ readListLevelStyles namespace elementName levelType =
...
                                     )
   where
-  toListLevelStyle _ p s LinfNone         = ListLevelStyle LltBullet p s LinfNone
-  toListLevelStyle _ p s f@(LinfString _) = ListLevelStyle LltBullet p s f
-  toListLevelStyle t p s f                = ListLevelStyle t      p s f
+  toListLevelStyle _ p s LinfNone b         = ListLevelStyle LltBullet p s LinfNone (startValue b)
+  toListLevelStyle _ p s f@(LinfString _) b = ListLevelStyle LltBullet p s f (startValue b)
+  toListLevelStyle t p s f b                = ListLevelStyle t      p s f (startValue b)
+  startValue (Just v)  = if all isDigit v
+                           then read v
```
jgm: Is this potentially unsafe?  Suppose v is the empty string, won't read fail?  (Maybe it's guaranteed that this won't happen in an ODT document, but it's still good practice to handle this case.)

hubertp-lshift: @jgm Should be good now.

jgm:
Great, thanks

Issue #3131, opened 2016-09-26
aeson 1.0.1.0 problem with filters


jgm:
```
pandoc MANUAL.txt -t man -s --template man/pandoc.1.template \
		--filter man/capitalizeHeaders.hs \
		--filter man/removeNotes.hs \
		--filter man/removeLinks.hs \
		--variable version="pandoc 1.17.3" \
		-o man/pandoc.1
pandoc: Error running filter man/removeNotes.hs
Error in $.date[1]: key "c" not present
CallStack (from HasCallStack):
  error, called at pandoc.hs:153:46 in main:Main
make: *** [man/pandoc.1] Error 83
```
This probably has to do with changes to ToJSON/FromJSON instances in aeson 1.0.
Need to resolve this before release.

jgm:
Hm, I can no longer reproduce this. Closing.

jgm:
Ah, no doubt this is because I recompiled pandoc and it used aeson < 1.0 because of a dependency.
It is possible to compile pandoc-types with aeson 1.0.1.0, and you then get this problem using a filter with a version of pandoc that was compiled with aeson 0.11.x.  Probably due to a difference in generic serialization.

jgm:
This appears to be the difference:
pandoc with aeson < 1.0
```
{"t":"Space","c":[]}
```
pandoc with aeson > 1.0
```
{"t":"Space"}
```

jgm:
Here's the relevant changeling note in aeson 1.0.0.0:
```
Remove the contents field encoding with allNullaryToStringTag = False, giving us { "tag" : "c1" } instead of { "tag" : "c1", contents : [] }. The contents field is optional when parsing so this is only a breaking change for ToJSON instances.
```

jgm:
So now we face a choice.  Should we modify pandoc-types so that the old behavior is restored?  This would avoid problems when the pandoc executable is compiled against a different version of aeson than the filter.  Or should we just go with the new (otherwise better) behavior, and tell people to make sure they compile filters against the same version of aeson as pandoc itself?

jgm:
Unfortunately I don't see a way to restore the old behavior.  So I think we're stuck with option 2.

jkr:
A while back you speculated about  writing explicit {To/From}JSON instances instead of deriving. It still seems like that would be a good idea -- it would make aeson's occasional erratic nature would be less of an issue. Of course, that assumes we could decide on a good json spec.

jgm:
[...]

PR #3155, opened 2016-10-08
Remove mmd raw tex

ickc:
According to [MultiMarkdown-5/using_mmd.md](https://github.com/fletcher/MultiMarkdown-5/blob/gh-pages/using_mmd.md#multimarkdown-and-latex), MultiMarkdown expects raw LaTeX to be within HTML comments, *e.g.*,

```html
<!-- \FloatBarrier -->
<!-- \begin{flushright} --><div align="right">
testing flush right.
<!-- \end{flushright} --></div>
```

I tried to compile the above file by `pandoc -f markdown_mmd -o test-pandoc.md test.md` and the HTML comments is not parsed as raw LaTeX.

I don't think pandoc should have a reader for that syntax since by [MultiMarkdown-5/Raw Source](http://fletcher.github.io/MultiMarkdown-5/raw.html), the same syntax could be used for "Raw LaTeX/OpenDocument/etc.".

The pull request is to remove the inclusion of `raw_tex` in `markdown_mmd` variants.

In other words, a raw LaTeX in the pandoc sense would never appears in a valid MultiMarkdown document (for example, if a raw LaTeX in the pandoc sense, say, `\FloatBarrier` is included in a MultiMarkdown document, the TeX output of that would becomes something like `\backslash FloatBarrier`. I tried that in the past.). And it would be better to remove it in the definition of `markdown_mmd` to avoid confusion.

I compiled it and everything looks fine except I also run into issue #3131.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3460,7 +3460,7 @@ variants are supported:
 
 `markdown_mmd` (MultiMarkdown)
 :   `pipe_tables`, `raw_html`, `markdown_attribute`, `mmd_link_attributes`,
-    `raw_tex`, `tex_math_double_backslash`, `intraword_underscores`,
+    `tex_math_double_backslash`, `intraword_underscores`,
     `mmd_title_block`, `footnotes`, `definition_lists`,
     `all_symbols_escapable`, `implicit_header_references`,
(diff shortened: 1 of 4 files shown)


jgm:
Thanks

PR #3156, opened 2016-10-10
Delete googlecode-upload.sh

ickc:
Since Google Code was shut down already, it seems that we don't need this script anymore?


jgm:
Yes, thanks

PR #3157, opened 2016-10-10
Build README.md from the MANUAL.txt and README.md.in

ickc:
Improve upon #3033: the main effort is to avoid maintaining the duplication between `MANUAL.txt` and `README.md`:

- Relevant places in `MANUAL.txt` is updated/modified to facilitate the reuse in `README`

- An auxiliary file `README.md` is used.

- `README.markdown` is built by `make README.markdown` from `MANUAL.txt` and `README.md`

    - it uses the trick that GitHub will pick the first README to show in the main page. Since `a` is smaller than `d`, `README.markdown` comes before `README.md`, and this will be the one used.

- Similar to #3033, it is a proof of concept. `head` and `tail` is used in the `makefile`, which still requires manual input when `MANUAL.txt` is updated and the number of lines are changed. A better logic can be written that target the `Description` and `Authors` sections of `MANUAL.txt`.

- An side-improvement is the links in the 2nd paragraph will be linked to the relevant sections of <pandoc.org/MANUAL.html>. These internal links in `MANUAL.txt` needed to be manually changed to external links in `README.md`. The benefit is one don't need to manually remove the `[]` from the `MANUAL.txt` and those links are also clickable and send the reader to the full manual.

Demo: <https://github.com/ickc/pandoc/tree/github-readme>


ickc:
I forgot to mention I moved those status images to the beginning since this is more common and logical to do.

jgm:
Initial reactions:

1.  I worry that having README.md and README.markdown will
cause too much confusion. (People will modify the wrong one,
etc.)  If we did this, I'd prefer to have something like
README.md.in (which is clearly a template).  Indeed, why
not use an actual pandoc template?

2.  The current method is too brittle, because of the use of
line numbers.  Why not use a filter to grab the Description
section from the manual and do appropriate changes?
(Similar to current man page generation.)

3.  If we do this, the script/filter should insert
HTML comments at the beginning of README.md that instruct
that the file is automatically generated, and that they
should not edit it manually, instead pointing them to
the Descriptions section of the manual and to the template
README.md.in.

4.  I like the idea of having the README.md link externally
to the manual.  Of course, this could be done independently
of the automatic generation idea.

5.  Even with this, changes to the Description have to be
done manually in multiple places, because the Description
is reproduced in pandoc.cabal.  If we really wanted to get
fancy, though, I suppose we could have the filter update
the Description section of pandoc.cabal after converting
md -> haddock.

I'm not sure the added complexity and potential for
confusion is worth it, in this case, but I'm not sure.

ickc:
I agree the use of line no. is not reliable. Just want to get a proof of concept first. The idea I have about getting the line no. automatically is through multiple of `grep -n` and some arithmetic. I think keeping everything (except the `README.md.in`) in the makefile will minimize the footprint of the README.

I took a look at the `pandoc.cabal` and it seems it is a summarized version of the "Description" (lesser paragraphs and each paragraph is more concise). Since that summerizing process has to be manual, it seems it is safe to ignore it here (unless you want them to match, which lengthens that?).

ickc:
Done: the line no. is now automatically calculated through the heading 'Using `pandoc`', 'Description', 'Authors'. As long as these wordings aren't changed, the makefile does not needed to be modified.

Except for the `pandoc.cabal`, the other points are done too. *e.g.* a few HTML comments make it more apparent on what it is doing.

ickc:
Some more comments on the approach I used. And as I kept brainstorming, it becomes quite long. I added some headings. Sorry for the lots of parentheses used!

# A better approach: "pandoc include"?

The current approach is a hack. A more "semantically correct" (or natural, logical?) way to do it seems to be some sort of "pandoc include" (which can currently be done through filter). In that case, there would be some more files: `DESCRIPTION`, `AUTHORS`, `DESCRIPTION_EXCERPT`. Then in `MANUAL.txt`, `markdown.md`, `pandoc.cabal` would "pandoc include" them through a filter.

That suggestion however makes the `MANUAL.txt` not self-contained. Although its purpose is mostly for converting to other formats later, but may be some people will still open it (perhaps when browsing on GitHub.com) directly, as it is a `txt`.

# Generating a GFM version from PFM sources as well?

Along the same line of what this pull request and #3158 is doing, I am very tempted to make a pull request to add `.md` extension to `BUGS`, `COPYRIGHT`, `PROFILING`, perhaps even `changelog` (In a certain sense, it is the "markdown everywhere" spirit that draws some of us into pandoc). But from the history of pull requests adding `.md` to `README`, it suggests extra caution is needed, and perhaps there are some reasons behind the non-existence of file extension.

The spirit behind the thinking are mostly related to GitHub preview. It is quite common for people to read files directly in GitHub.com, and if it is a `.md`, it would be previewed in HTML automatically which makes reading much easier (even in the case of no markdown syntax involved!).

My guess is that the main/only reason a `.md` extension is not used is because it is in PFM. `.md` will be regarded and rendered as GFM on GitHub.com, which causes confusion.

One way to deal with the problem is, for any existing files that is already in markdown, but perhaps using pandoc's exclusive extensions making it not portable, either

1. remove the PFM extensions and take the HCF between GFM and PFM and name it `.md` (`INSTALL` might be an example?)
2. In the case that PFM is absolutely necessary (*e.g.* `MANUAL.txt`), then `pandoc -f markdown -t markdown_github` is used to generate a GFM version of `.md`. *i.e.* treating the `.md` as one of the targets of the sources, similar to `COPYING.rtf` for example.
	1. In this case, the source (e.g. `MANUAL.txt`, `README.md.in`) is put in a folder, say, `src/md`. Every generated `.md` is then started with an HTML comment redirecting anyone wanting to edit it to the said folder.

This suggestion also solved the problem with the "pandoc include" approach above. Since `MANUAL.txt`, `README.md.in` would be in a src folder, and a `.md` version is generated in the root, no one will need to read the sources there on GitHub.

vyp:
Just my opinion but I think option 2 is much too complicated and would just result in confusion. Besides, markdown is fairly easy to read in plain text anyway (because that's the whole point) and you're supposed to be able to edit markdown as well too, so it's a bit strange to tell people not to. I do understand the motivation for the change though but just not sure if it's worth the trouble of generating. :+1:

ickc:
@vyp Good point. I can see the irony and that's overdoing.

So I think when PFM is absolutely necessary, it should be left as is. But may be somewhere it should be indicated clearly it is in PFM, either through a file extension, or perhaps an HTML comment in the file (or mime type?).

For those that PFM use is minimal and not necessary, perhaps it should be converted to vanilla markdown and add a `.md` extension to it (*e.g.* `INSTALL`?)

ickc:
@jgm,

Is the current method ok or still too hacky? Thanks.

jgm:
I think this is too much complexity for what it gets us.  So I'm going to pass on this one.

Eventually, if we wanted to have a GitHub readable manual, we could try rewriting the manual in GitHub-compatible markdown, but using some special conventions for footnotes and definition lists (the main pandoc-specific features we use).  We could then use these conventions when converting the manual to HTML or PDF for the website and to a man page, so we'd still get these features.

ickc:
I opened a separate pull request #3167 contains only the manual modification but not the `make readme.md` automation. I think probably I would keep that branch in my fork and from time to time I can "update" the `README.md` from `MANUAL.txt` and submit a pull request.

Regarding how to do the automation "correctly", I think converting the current pandoc markdown perfect `MANUAL.txt` into a sort of hybrid that works well on native GFM and can be used to convert to pandoc markdown is a bit backward, and can also cause confusion to others who want to make changes and pull request. The lack of footnote in GFM is a major issue and looking into the current 6 footnotes in `MANUAL.txt`, they seems important and should be represented something else.

And from what you said about the GitHub readable manual, it seems the length of the `README.md` isn't a concern (I initially think that the current `README.md` is much shorter than `MANUAL.txt` because of that). In this way, may be we can just do something like `pandoc -t markdown_github -f markdown -o README.md MANUAL.txt` and tweak on the from and to variants until it looks right (*e.g.* if a certain setup can output a GFM that contains HTML footnote, that would be great. From [stackoverflow](http://stackoverflow.com/questions/25579868/how-to-add-footnotes-to-github-flavoured-markdown), it seems HTML footnote is a viable solution.

Do you want me to try create some sort of `pandoc -t markdown_github -f markdown -o README.md MANUAL.txt` and see how well it goes?

ickc:
I actually try `pandoc -f markdown -t markdown_github` and the result is in [ickc/pandoc at github-readme-take3](https://github.com/ickc/pandoc/tree/github-readme-take3).

Looking at the result, it looks almost perfect, except the footnote are not hyperlinked. If you like that I could pull request from there.

jgm:
+++ ickc [Oct 18 16 22:33 ]:
>   I actually try pandoc -f markdown -t markdown_github and the result is
>   in [1]ickc/pandoc at github-readme-take3.
>
>   Looking at the result, it looks almost perfect, except the footnote are
>   not hyperlinked. If you like that I could pull request from there.

Well, the definition lists come across as regular paragraphs
with a line break after the first line, which is not ideal
and won't work well for creating the man page.

Non-linked notes are also bad for the online manual.

ickc:
Oh, I don't mean it to be the source. It is the output. So `MANUAL.txt` is still the source, but both the man and the github `README.md` is generate d by it directly. So then there's no manual editing on `README.md` anymore.

ickc:
> Non-linked notes are also bad for the online manual.

If you mean a table of contents, I just added one. And if you mean the internal links to different sections, it is already there. *e.g.* <https://github.com/ickc/pandoc/tree/github-readme-take3#metadata-blocks>

jgm:
I see.  I don't particularly like having generated files in
the repository, but this is a possibility, I suppose.

+++ ickc [Oct 19 16 04:02 ]:

ickc:
It can be regarded as a demo of pandoc's capbability though. It will be the best example showing what 'from markdown to markdown' can do.

PR #3158, opened 2016-10-10
Use the markdown version of COPYING from GNU

ickc:
I see that in the changelog there are 2 instances:

> - Split LICENSE into COPYING and COPYRIGHT.
> - Converted COPYING to markdown.

This pull request is 2-fold:

1. download the markdown version of the license directly from <https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html>
2. changing the name COPYING to LICENSE.md
    1. making the markdown nature of it explicit
    2. It seems that LICENSE is more appropriate, since the license concerns "copying, distribution and modification" so that COPYING is only part of what is covered.


ickc:
By the way, since

>Split LICENSE into COPYING and COPYRIGHT.

There might be an original reason to avoid using the word LICENSE. But it seems that LICENSE vs COPYRIGHT is pretty clear. The copyright holder holds their copyrights and can change their license whenever they want to, but the one who obtained the code has to obey the LICENSE.

jgm:
I've followed the practices suggested for the GPL by the
FSF:

https://www.gnu.org/licenses/gpl-howto.html

I don't see a strong reason to change.

ickc:
I see. Sorry I didn't know that. I only knew GitHub's standard practice is to use LICENCE. And for the reason explained COPYING doesn't seem to fully explained what it is. But since it is the official recommendation from where the license is coming from, I can't argue with that.

How about the content? The remaining changed I have in this branch is to use the GPL v2 markdown version downloaded directly from <https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html>. From the changelog, it seems it is converted from some other form in the past. And seeing how the COPYING is used in the makefile, it seems to be converted to other forms say HTML from the markdown. So it seems doesn't harm to use the official's markdown version rather than a converted one.

ickc:
Just a quick look on the wordings, it seems the one I just downloaded from GNU has some update on wordings, *e.g.* `GNU Library General Public License instead` becomes `GNU Lesser General Public License instead`. So the new one seems to be preferred.

ickc:
I also added `.md` extension to `COPYING` since it is in Markdown.

ickc:
Preview of the `COPYING.md` on GitHub: <https://github.com/ickc/pandoc/blob/COPY-LICENSE/COPYING.md>

ickc:
I made a change of gitignore for 2 reasons:

1. The previous build has failed. I tested it, but it should actually passed. I suspect may be I make some rapid changes in a short time, so travis complained it took too long for the test.
2. During the test, those files `tests/pandoc-test*` becomes a noise on new files added in the repo.

ickc:
@jgm, to summaries,

- [ ] Free Software Foundation's latest (some wording is changed) markdown version of GPLv2 is used instead.
- [ ] `COPYING.md` is used instead of `COPYING`.
- [ ] `tests/pandoc-test*` is added in `.gitignore`.

Which of these would you not like to see changed? Or does it requires other change (*e.g.* using a different header level: original `COPYING` starts with heading 1, and FSF's markdown version starts with heading 3)?

If none of the changes are desired, feel free to close the pull request.

Thanks!

jgm:
I'm okay with the changes, but can you rebase so we just have one nice atomic commit?
(Actually, the gitignore change should be a separate commit, but the COPYING.md related changes should be lumped together.  I'm also not sure why the tests are creating a file in tests/.  They shouldn't do that under normal circumstances, and  didn't when I tried it...)

ickc:
I revert the `.gitignore` and will test later to see why I have those files shown.

I squashed the commit into one.

By the way, there is a way for admin to squash any pull request in GitHub: (from <https://github.com/blog/2141-squash-your-commits>)

squash

ickc:
I tried again to play with `cabal ...` and `make ...` but I failed to reproduce those auxiliary `tests/pandoc-test*` files. Most probably I made mistake(s) last time.

jgm:
Thanks for squashing.  One more thing -- can you change the commit message?
```
Change every references to COPYING to LICENSE.md instead

remove COPYING

Revert "Change every references to COPYING to LICENSE.md instead"

This reverts commit f121dd7.

overwrite COPYING by the one downloaded from GNU

Add .md extension to COPYING

gitignore the test files from `cabal build`

So that the noise won't show up during the test.

Revert "gitignore the test files from `cabal build`"
```
has a lot of noise from the now squashed history.  A short description of what has changed would be most helpful later when I update the changelog.

Issue #3076, opened 2016-08-16
Pandoc Writer (Docx -> Markdown) should have option to put footnotes immediately after paragraphs


jalustig:
Currently, Pandoc will output Markdown with footnotes at the end of the document. For instance:

    This is a paragraph.[^1]

    This is a second paragraph.[^2] Another sentence.[^3]

    [^1]: One footnote.

    [^2]: Another footnote.

    [^3]: A third note.

It would be great to be able to have the writer be able to optionally output footnotes inline to the document - not to the text itself. That it to say, after each paragraph it will list the footnotes, like this:

    This is a paragraph.[^1]

    [^1]: One footnote.

    This is a second paragraph.[^2] Another sentence.[^3]

    [^2]: Another footnote.

    [^3]: A third note.

Right now, this latter format is a valid Markdown *input* format for Pandoc, but it will not be output in this manner.

It would be particularly useful for diffing documents (or printing them for editing purposes). This way, the local changes are near to each other -- and particularly when text is moved from the body to a footnote or vice-versa, it will be clearer how the document has been edited.

jkr:
I would actually like to see this too. It would make it easier to cut and paste  sections of markdown if the footnotes could be attached to the relevant paragraphs.

jgm:
It would require a command-line option, I suppose.
Or should we implement it as an extension?  It would be an
odd extension, though, only affecting rendering, not
parsing.  Perhaps an option is best?


+++ Jesse Rosenthal [Oct 08 16 06:59 ]:

jgm:
Or should it be the default?

If an option, maybe
`--markdown-footnotes=[interspersed|end]` or something like
that?

+++ Jesse Rosenthal [Oct 08 16 06:59 ]:

jkr:
> Or should it be the default?

Emacs markdown-mode offers three options (for both footnotes and reference definitions): end of document, after current block, and before the next header. For footnotes, end of document is the default, which seems sensible. (It's what I prefer anyway.)

So I think I'd propose the same: `--markdown-footnotes=[end|block|header]`, with `end` as the default. Or, if we want, we could include link references (if reference-links is set).

So, I think that gives us three options:

1. `--markdown-footnotes=[end|block|header]`, no effect on links.

2. `--markdown-footnotes=[end|block|header]`, and change `--reference-links` to something like `--markdown-links=[inline|end|block|header]` (with inline as the default).

[...]

PR #3159, opened 2016-10-12
Specify location for footnotes (and reference links) in MD output

jkr:
As discussed in #3076 

A couple of notes:

1. I've used the option `--reference-location=block|section|document`. I didn't want to specify markdown, since it seemed like this could be used for rst and org as well (though I don't know as much about those formats). If this seems like it's overloading the term "reference," I'm open to changing it.

2. For the `--reference-location=block` option, I've implemented it so that it comes after top-level blocks -- i.e., not after pars in a table cell, or list items. I think this is more in line expectations, and will produce better looking output. Also, per a quick check on babelmark, it will produce output that is more likely to be compliant with other parsers., 

3. The changes to the markdown writer required a reader monad for keeping track of block levels. I took the opportunity to move a few state variables that belonged in the reader monad over to it.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -645,6 +645,13 @@ Options affecting specific writers
 :   Use reference-style links, rather than inline links, in writing Markdown
     or reStructuredText.  By default inline links are used.
 
+`--reference-location = block` | `section` | `document`
+
+:   Specify whether footnotes (and references, if `reference-links` is
+    set) are placed at the end of the current (top-level) block, the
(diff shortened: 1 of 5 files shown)


jkr:
Let me know if you have any comments or proposed changes. When it looks good, I'll push, and we can figure out if other writers should be affected by the option as well.

jgm:
Looks good!
I'll add something to the manual under `--reference-links` as well.

PR #3165, opened 2016-10-17
[odt] images parser

hubertp-lshift:
Provide parser for the ODT images (previously ignored). Also, nested 'draw:frame' elements are handled in order to infer a possible non-empty caption.


hubertp-lshift:
@jgm are the errors some false positives? Doesn't seem to be related.

jgm:
+++ hubertp-lshift [Oct 18 16 02:46 ]:
>   [1]@jgm are the errors some false positives? Doesn't seem to be
>   related.

Yes there are some other issues that I'm trying to clear up now.

PR #3167, opened 2016-10-19
"Merge" MANUAL.txt into README.md

ickc:
Pull request #3157 without the automatic building from MANUAL.txt to
README.md


jgm:
I'd rather not duplicate the list of contributors in README.md.  The less we duplicate from MANUAL to README, the better.

ickc:
I removed them. But if the concern is about 1 more place to manual update the `README.md`, I could use my hacky script from time to time to update `README.md` from `MANUAL.txt` and make pull request, if it's ok.

tarleb:
The reason to put the shields so far down in the readme was that it can distract from the actual content.  Most notably,  GitHub on mobile only shows a brief excerpt of the README, so there is now not much info besides those badges.

Before:

screenshot

After:

screenshot2

ickc:
Actually the one below looks better to me. On such a small screen I'd rather see those badges than text. And I don't think people would surf thorugh GitHub on such a small screen to discover repositories.

Most of the time, I guess people access GitHub on big screens, either a computer or an iPad like I'm doing at the moment. On those screen, the badges look way better on the top.

I guess that's why other people using those badges did the same thing. Besides, I would say the functionality of those badges are lost if they are not on the top.

tarleb:
At least for me, it is not that rare that I browse github on my phone. I
agree that the new version looks better, but pages like that annoy me
because they don't contain not much interesting content. I have to click
through to the README to even see what the package does.  Big screens
are probably the default though, and it does look nicer and is just as
useful on a desktop.

I don't feel strongly about this, just want to make sure the
less-obvious result of this change is understood.

jgm:
Yes, thanks @tarleb for reminding me why it was originally done this way.
I think I'll stick with having the buttons at the top, though -- it's pretty much standard on GitHub, and one advantage is that you immediately see clickable links to the releases.

PR #3168, opened 2016-10-19
[odt] proper formatting of an image with a caption

hubertp-lshift:
Latex Writer only handles captions if the image's title
is prefixed with 'fig:'.


jgm:
Thanks!

PR #3180, opened 2016-10-24
[tex] Positioning of figures

hubertp-lshift:
Adds an option to position figures with captions.
Currently only ODT parses its position information but
other inputs formats can easily support it as well.


hubertp-lshift:
Any objections to having such a feature in the LaTeX writer?

tests/writer.latex
```diff
@@ -923,7 +923,7 @@ or here: <http://example.com/>
 
 From ``Voyage dans la Lune'' by Georges Melies (1902):
 
-\begin{figure}
+\begin{figure}[]
```
jgm: In any case we should avoid empty brackets when there are no options.

jgm:
One concern I have is that `pos` is not a valid HTML attribute, so when this is rendered in HTML it will be invalid.  (In fact, that's already a concern about how it's currently done with the ODT reader.)
Using `data-pos` would solve that, though it seems a bit fussy.
Alternatively, the HTML writer could be taught to ignore the `pos` attribute or something.

hubertp-lshift:
@jgm Thanks, I wasn't aware that HTML accepts all attributes. In the revised one it is prefixed with "pandoc-" to mark it as an internal one (and simply ignored in HTML).

PR #3186, opened 2016-10-26
[odt] Mark text using a fixed-length font as code

hubertp-lshift:
Span blocks that use fixed-length fonts
(i.e. monospaced) should be marked as an inline
code rather than be treated as a normal text.
Code blocks are still not recognized by the
ODT Reader.


jkr:
What happens if the entire document is in a monospace font -- e.g. a manuscript in courier? (Sorry if it's a silly question -- don't know enough about odt internals to know what a "span block" means here.)

hubertp-lshift:
Yeah, so all text fragments will be wrapped in the `texttt` latex command. I'm not sure how that fits in the whole Pandoc design, but I'm guessing that's not desired. The thing is that ODT has no single way of specifying code, neither inlined nor block, so maybe that feature would just have to be documented? It is common that code is written in monospace so would that be a restriction?

jkr:
That doesn't seem ideal, since there are still alot of cases where people might use a mono font for the text (screenplays, some agents and publishers require courier new, professors who are too attached to old MLA). 

It would be safer, I think, if it could work off character styles, similar to what the docx reader does. So we could either use some/all of the built in char styles (`Teletype`, `Source Text`), or our own (`Code`) or both. 

So here, "some" would be read as code (spacing added):

~~~{.xml}
<text:p text:style-name="P1">
Here is 
<text:span text:style-name="Source_20_Text">
  some
</text:span>
code.
</text:p>
~~~

Ideally, the writer would be set up to do the same thing, instead of just putting it into courier with the default char style.

jkr:
Some random examples of sites that recommend that manuscripts be prepared in Courier for submission to publishers:

http://www.ruthnestvold.com/Courier.htm

http://www.shunn.net/format/fonts/

It doesn't seem to be a real rule anymore, but it also seems better not to derive semantics from font choice.

jkr:
By the way, on your earlier point that I didn't address: it's okay if there's no single way. We can choose a sensible subset of styles and say that those are the ones we'll read. If there's enough style parsing in the odt reader (I'm unfamiliar with the internals) we can also do everything inherited. "Source Text", "Teletype", and whatever we have the reader output seems like it would be a sensible enough starting point.

hubertp-lshift:
Ok, that sounds doable. I will work on that tomorrow. Cheers!

jkr:
This looks good. It worked well when I tested it. Would you be able to squash the changes, so there isn't the misleading commit message from before?

By the way, I noticed something strange when I was testing with monospaced font. The ODT reader marks text in certain mono fonts as emphasized. This doesn't seem right to me. It wasn't as a result of this patchset -- but do you know why the reader does this?

hubertp-lshift:
@jkr squashed.

Regarding your second question - I also noticed that accidentally. My guess is that that the `hasEmphChanged` in ContentReader.hs is at fault. More specifically, certain mono fonts have a fixed length pitch and that method is checking if it the pitch has changed by doing `hasChangedM pitch`. It's a wild guess and I haven't tested it enough.

hubertp-lshift:
@jkr I missed the bit about styles.xml, fixed now. But I understand your second question, sorry. Could you elaborate?

jkr:
Sorry if I wasn't clear. If you build with your newest patchset and then convert to odt 

~~~
echo 'foo `code` bar' | stack exec pandoc -- -o out.odt
~~~

and then open the resulting odt and open "View -> Styles and Formatting' and look at character styles, the `code` part will still be listed as being in the "Default Style:" with courier font, as opposed to "Source Code." I"m not sure why this is the case -- I experimented with changing the "Source Text" style to "Teletype" in your code and it works as expected. Unfortunately, I don't know enough about odt to diagnose it completely (and this has been too busy a week to learn).

What you've done so far is obviously a big improvement, so if you poke around and can't figure out what the issue is, I'll just push it and we'll get back to it. But if you can figure out what's making it behave this way, it might be good to do it all in one fell swoop.

hubertp-lshift:
I did some more cleanup (not adding Source_20_Text to doc's styles anmore since it is in styles.xml). Regarding the 'styles and formatting' isn't it because only a fragment of the text is in that style? Normally if the whole paragraph is in that style then you can see it changed.
Also note that the human readable name is 'Inlined code' rather than 'Source text'.

jkr:
Right, but you can switch to character styles (the little "a" thing in the style view). There the character style (i.e., the inline style) shows up as "Default Text".

hubertp-lshift:
Mystery solved: I had to change the `family` from `paragraph` to `text`

hubertp-lshift:
Actually, I'm wondering if we should just completely abandon the `Source_20_Text` style in favour of just `Source_Text`. WDYT?

hubertp-lshift:
That looks like a spurious error

jkr:
looks good to me. I'm happy to push it now. Question, though: why use "Source_20_Text" with the visible name of "inlined code," as opposed to the built in visible "Source Text" which is intended, I think, to be used for that?

jkr:
But if we do keep it as "inlined code", we might as well make the internal name similar, so it doesn't overlap with the builtin styles.

hubertp-lshift:
Yeah, I agree. I changed it to `Source_Text`.

jkr:
Okay -- I pushed it as it is. Thanks for your work and flexibility on this.

But I think my point wasn't quite clear.

LOWriter comes with "Source Text" (internal Source_20_Text) as a default char style, with the intention that this would be used for code. So it would make sense for the reader to accept this as the inline code style, instead of expecting people to create their own style. And for that reason, it makes more sense to me to use the built-in char style for the writer too (since if people add to pandoc output, they will be able to do what they already do with a default ODT doc).

So what I was suggesting that we should read and output the built-in "Source Text"/Source_20_Text since that matches a style that ODT already has, instead of creating our own. (Equivalent to, say, using `<em>` in pandoc html output, instead of `<span class="pandocs-own-code-style">`.)

Anyway -- not certain on this. And what you've added is a big improvement, so there's no rush to modify it. But that's what I was trying to get across before.

hubertp-lshift:
Right. I guess the message just got lost in the translation. I will correct that in another PR.

Issue #2761, opened 2016-03-07
A Need of a Small Caps Syntax That Is Not Explicitly Fake Small Caps


ickc:
I am sure people have been brainstorming and discussing about this. And a quick search showed that almost a decade ago this has been discussed and a syntax like `^^small cap^^` has been suggested but probably it is not attractive enough for people to pick it. In contrast I'm totally new here so probably what I'm going to say don't even worth 0.02 cents. But please bear with me:

I think the more important thing than picking a syntax is the need to have one. Although it is difficult for people to agree on a syntax, I wish we can at least agree a need of a better syntax. In pandoc I think the small cap is the only exception to the rule that it does not look "marked down" but uses HTML syntax, although it understand it and will transform it according to the output format. Probably that makes it most urging to have a new, "marked down" syntax?

I also understand because of the CommonMark movement, probably it's not the best time to define new syntax. But I think at least we should leave it as an open issue pending to be solved sometime in the future when the dust settled?

As to the suggestion of syntax, here's one: `!small cap!`. Seems like it shouldn't has a conflict on existing syntax yet. But it's just my naive suggestion. I think more importantly is we agree we need a better syntax for that.

ousia:
Sorry, @ickc, but I don’t think it makes sense to have an inline element for small caps.

The span element with a small class should be enough (such as in `[small caps]{.sc}`).

The special syntax for divisions and spans is yet to be implemented (#168).

ickc:
Could you elaborate more on why?

And if there's a strong reason such an exception (comparing to all other existing syntax in pandoc markdown) should be made, I suggest to put it in the documentation. It is because I think every newbie reading through the documentation would have this question in there mind (I just finished reading the whole doc on pandoc so I am one of those newbies...).

And I think when we think small cap in the way of "span and class", it is very HTML. But to me small cap is much more similar to bold, italic, etc., in the way that a real font have a bold font for bold, italic font for italic, small cap font for small cap, etc. it is so happen because of this, it can be thought in terms of class and applied a font style with it.

My personal interest in pandoc is primarily LaTeX output with occasional HTML and docx use. In LaTeX a small cap would never been thought in terms of span and class so it is very unnatural. So since pandoc is output format agnostic (not sure if it is a good way to say it, but in the philosophy of pandoc it said something like the original markdown philosophy with one important difference that is multiple output format), it seems a very bad syntax if it is in HTML (current syntax) or indirectly HTML (the example you gave.

My 0.02 cents.

[...]

Issue #1592, opened 2014-08-31
parse small caps from HTML as a class


ousia:
Writing small caps to a class in HTML, textile and markdown has the following benefit.

For HTML there are three ways of getting small caps:
1. `font-variant: small-caps;` is the standard way in CSS2.
2. `font-feature-settings: "smcp";`is the way to enable OpenType features in CSS3.
3. `font-family: MySmallCapsFont;` is the way to get real small caps instead of the fake ones from the first option. Used in many ePub books.

Consider the following: if small caps is a class, the user would be able to adapt the best method with the available fonts. If small caps are hardcoded, there will be no other way than fake small caps.

BTW, small caps are lost when converting to textile ([sample](http://johnmacfarlane.net/pandoc/try/?text=%3Cspan+style%3D%22font-variant%3Asmall-caps%22%3Ea%3C%2Fspan%3E&from=markdown&to=textile)).

Many thanks for your help.

ickc:
I agree using a class too. For backward compatibility, may be a default CSS (`font-variant: small-caps;`) should be given (before custom CSS) in all HTML related templates. Because of the cascading properties, user defined CSS can override this default behavior.

When deciding on which class to use, I suggest keeping it short. Together with the possibility given by the new bracketed spans (pull request #3191), it would make a very short and convenient native markdown small cap syntax.

I suggest using the short form already used in ` font-feature-settings`: `smcp`. This way, it won't be too short/arbitrary to accidentally turn a class (that is not intended to be small caps) into native `SmallCaps`. The syntax combined with bracketed spans will then be `[Small Caps]{.smcp}`.

ickc:
Looking at the code and studying the output of pandoc again, currently the markdown reader pretty much looks for exact match of `<span style="font-variant:small-caps;">Small caps</span>`, that has no id nor class (it allows other attributes, e.g. `foo=bar`. And even then:

```bash
printf '<span foo=bar style="font-variant:small-caps;">Small caps</span>' | pandoc -t native
```

will output

```json
[Para [SmallCaps [Str "Small",Space,Str "caps"]]]
```

i.e. the `foo=bar` information is lost. And the SmallCaps has no sense of attributes.

SmallCaps having no attribute at all is ok in the old syntax. But now if we want to use, say class, to denote small caps instead, then it should have a sense on other attributes as well. In this case, does it mean it requires AST change?

[...]

PR #3191, opened 2016-10-27
Small caps in Bracketed Spans

ickc:
I adapted the relevant code from `spanHtml` into `bracketedSpan` to check if small caps style is used in bracketed spans.

I then update the `MANUAL.txt` accordingly to reflect this use.

I tested and it compile successfully. The code I used to test if the small caps with bracketed span works is:

``` markdown
1. HTML:
    - <span style="font-variant:small-caps;">Small caps</span>
    - <span style="font-variant:small-caps">Small caps</span>
    - <span style="font-variant: small-caps">Small caps</span>
2. markdown:
    - [Small caps]{style="font-variant:small-caps;"}
    - [Small caps]{style="font-variant:small-caps"}
    - [Small caps]{style="font-variant: small-caps"}
```

I don't know anything about the tests yet so I didn't modify any of the test to test this.

Edit: this should partially resolves issue #2761

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2719,6 +2719,10 @@ To write small caps, you can use an HTML span tag:
 (The semicolon is optional and there may be space after the
 colon.) This will work in all output formats that support small caps.
 
+Alternatively, you can also use the new `bracketed_spans` syntax:
+
+    [Small caps]{style="font-variant:small-caps;"}
+
 Math
 ----
 
--- a/man/pandoc.1
+++ b/man/pandoc.1
@@ -3388,6 +3388,14 @@ To write small caps, you can use an HTML span tag:
 .PP
 (The semicolon is optional and there may be space after the colon.) This
 will work in all output formats that support small caps.
+.PP
+Alternatively, you can also use the new \f[C]bracketed_spans\f[] syntax:
+.IP
+.nf
+\f[C]
+[Small\ caps]{style="font\-variant:small\-caps;"}
+\f[]
+.fi
 .SS Math
 .SS Extension: \f[C]tex_math_dollars\f[]
 .PP
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -1758,7 +1758,13 @@ bracketedSpan = try $ do
   guardEnabled Ext_bracketed_spans
   (lab,_) <- reference
   attr <- attributes
-  return $ B.spanWith attr <$> lab
+  let (_,_,keyvals) = attr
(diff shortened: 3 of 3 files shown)


ousia:
@ickc, sorry if I’m not getting everything right (Haskell is really Greek to me). I think the following is an important issue.

Please, don’t hard code CSS code into Markdown syntax. Using a class such as `[small caps]{.small-caps}` is better. So we achieve:

* Cleaner and clearer writing and a more proper Markdown syntax.

* Depending on your scenario, with a class you might use:

    * Another font. _Adobe Reader Mobile_ doesn’t comply with `font-variant:small-caps`, at least in previous versions and this is the software for e-ink readers. And this is also the way to get real small caps when OpenType features aren’t an option.

    * Fake small caps, such as the ones with `font-variant: small-caps;`.

    * With the proper font and rendering engine, the OpenType `smcp` feature (which are real small caps).

Sorry, but I don’t think this is the right approach. Markdown should not mimic CSS or TeX (to name two), but it should use classes in the syntax itself.

Otherwise, OpenType usage would be problematic in this case. Keeping open options (i think) is a much better idea.

ickc:
I am aware of the issue #1592 you described. However, this pull request is to make sure the bracketed span behaves the same way as the html span in the markdown reader. Currently the "font variant" is still the only and recommended small cap syntax in pandoc according to the manual.

ickc:
About the test: actually when I dig a bit deeper it doesn't seem like to have a test of SmallCaps in markdown reader.

ickc:
I just made one more commit to correct the code I wrote in the markdown reader: bracketedSpan functions *truly exactly* the same as spanHtml. Previously I didn't check id and class are empty.

Regarding the test, since the test didn't test for smallcaps, so writing a test for small caps for both bracketed_spans and native_spans should belongs to a separate pull request.

ickc:
Are there any concerns about this pull request? Since what it does is basically to make sure the **`bracketed_span` functions exactly as a `native_span` would**, it seems it shouldn't be too much problem with it?

Until a better small-caps syntax is settled, at least we can make sure we can use `bracketed_span` to specify small-caps in the way `native_span` is already doing?

jgm:
Thanks.  I do like @ousia's suggestion to use a "small-caps" class.
Perhaps the condition could be:  if the span (whether HTMLish or bracket) has *either* the class "small-caps" (and no other attributes) or style="font-variant:small-caps" (and no other attributes), it's treated as a small caps element.

jgm:
Although perhaps @ousia is more concerned with HTML output than with how these things are input.

jgm:
In any case, this discussion should proceed on another (open) issue.

PR #3199, opened 2016-10-31
[odt] Infer tables' header props from rows

hubertp-lshift:
ODT reader simply provided an empty header list
which meant that the contents of the whole table,
even if not empty, was simply ignored.
While we still do not infer headers we at least have
to provide default properties of columns.


jgm:
Thanks!

PR #3218, opened 2016-11-08
[tex] No invalid inlines in sections' options

hubertp-lshift:
Latex doesn't like when hypertargets are put in the
options list of the section. They are not lost since
they were actually duplicated and present also in
the argument list.

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -733,14 +733,16 @@ sectionHeader :: Bool    -- True for unnumbered
 sectionHeader unnumbered ident level lst = do
   txt <- inlineListToLaTeX lst
   plain <- stringToLaTeX TextString $ concatMap stringify lst
-  let noNote (Note _) = Str ""
-      noNote x        = x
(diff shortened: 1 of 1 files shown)


tests/Tests/Writers/LaTeX.hs
```diff
@@ -65,7 +65,7 @@ tests = [ testGroup "code blocks"
             "\\begin{description}\n\\item[foo] ~ \n\\subsection{bar}\n\nbaz\n\\end{description}"
           , "containing image" =:
             header 1 (image "imgs/foo.jpg" "" (text "Alt text")) =?>
-            "\\section{\\texorpdfstring{\\protect\\includegraphics{imgs/foo.jpg}}{Alt text}}"
+            "\\section[]{\\texorpdfstring{\\protect\\includegraphics{imgs/foo.jpg}}{Alt text}}"
```
jgm: It's not ideal getting the empty `[]` here.  I assume that's because you're changing things into `Str ""` rather than just nothing.  Why not make `removeInvalidInline` a function `Inline -> [Inline]`?

PR #3224, opened 2016-11-09
[odt] Infer table's caption from the style's name

hubertp-lshift:
Open document's reader always ignored the possibility
of the caption of the table. This commit
1) treats specially a paragraph with a style named 'Table'
2) does some postprocessing of the paragraphs that combines
 tables followed immediately by captions

The special style is named 'Table' rather than 'TableCaption'
since open office by default uses the former and we want to
remain consistent.


jgm:
This looks good to me.  But can you make it clearer in the commit message that this affects both the reader and the writer, and that it may require changes to a custom reference.odt for those who have given special formatting to TableCaption?

hubertp-lshift:
@jgm Hopefully that one is clearer.

jgm:
Thanks!

Issue #1934, opened 2015-02-10
Pandoc doesn't honor `\newpage` or `\pagebreak` except for PDF output files.


CodeGnome:
## Pagebreaks Don't Work for Most Output Formats
I have a Markdown file that is supposed to have pagebreaks between certain sections. However, Pandoc 1.10.1 isn't honoring the `\newpage` or `\pagebreak` commands when rendering RTF, DOCX, or ODT formatted files. The commands I'm using to invoke pandoc are:

    for format in rtf docx odt; do
        pandoc \
            --smart \
            --normalize \
            --standalone \
            --self-contained \
            -f markdown \
            -t $format \
            --output="${FILE/markdown/$format}" \
            "$FILE"
        echo "Created ${FILE/markdown/$format}"
    done

## PDF Seems to Work
However, the PDF format (which requires a slightly different invocation because it doesn't respect the `-t` flag) seems to respect the pagebreak requests. For example:

    pandoc \
        --standalone \
        --normalize \
        --smart \
        --self-contained \
        --from=markdown \
        --output="${FILE/markdown/pdf}" \
        "$FILE"
    echo "Created ${FILE/markdown/pdf}"

jgm:
Correct, pandoc's internal document model does not currently contain anything corresponding to a page break, so there is no way to convert these.  In principle a PageBreak element could be added.  It's also possible to work around this deficiency using pandoc filters.

CodeGnome:
A PageBreak element would be great, and I'd be happy to use a filter in the meantime. However, I'm not sure what's entailed in doing so. How would I generate a DOCX with forced page breaks using a filtering mechanism?

jkr:
@CodeGnome : see this thread for some hints on setting up a filter for pagebreaks in docx output:

https://groups.google.com/forum/#!searchin/pandoc-discuss/pagebreak/pandoc-discuss/FzLrhk0vVbU/GtSHaI0jddAJ

s7726:
@CodeGnome If your page breaks happen to be prior to a given heading level, you can just set the page break before property for that heading style.

Hi-Angel:
I am also voting for the feature to be added — many formats have something according to a page break *(even in CSS are things like page-break-\*)*.

Hi-Angel:
Hi, I'm just looking through the code in the hope to add the pagebreak, and some features, and I found, well… Does @jgm notice [the two years old pull request](https://github.com/jgm/pandoc-types/pull/9)?

[...]

PR #3230, opened 2016-11-14
Add pagebreaks to Pandoc

hubertp-lshift:
A first stab at adding pagebreaks to Pandoc. Fully implemented for ODT (reader & writer).
Also added a few other writers, when the underlying format supported them out of the box. 

Depends on https://github.com/jgm/pandoc-types/pull/28


wilx:
I wonder this could be implemented using the [FORM FEED](https://en.wikipedia.org/wiki/Page_break#Form_feed) character for textual output formats.

hubertp-lshift:
@wilx for the purpose of latex/odt, maybe. I didn't try it. But I think pagebreaks is such a common (and desired) feature that it deserves its own definition for the purpose of other readers/writers.

jgm:
@hubertp-lshift I've merged this into the typeclass branch (which will become pandoc 2.0).

However, I'm not sure it's right for PageBreak to be an Inline element.  It seems more appropriate as a Block. Can you comment on why you chose to make it an Inline?  (This may be more natural in ODT, but we should also consider other output formats.)

jgm:
I've reverted this.  It needs more thought and discussion.

mb21:
The place for discussion is probably https://github.com/jgm/pandoc/issues/1934

PR #3235, opened 2016-11-16
remove make_travis_yml.hs

ickc:
If I didn't misunderstand the [History for make_travis_yml.hs - jgm/pandoc](https://github.com/jgm/pandoc/commits/master/make_travis_yml.hs) and the [History for .travis.yml - jgm/pandoc](https://github.com/jgm/pandoc/commits/master/.travis.yml), it seems `.travis.yml` is no longer build from `make_travis_yml.hs` 11 months ago.


ickc:
I found that probably [User guide - The Haskell Tool Stack](https://docs.haskellstack.org/en/stable/GUIDE/#travis-with-caching) is where the travis template is coming from. May be a link to that should be included in the `.travis.yml`?

jgm:
I've added a comment to .travis.yml.

PR #3193, opened 2016-10-28
Add a GitHub Pull Request template

ickc:
- [Setting guidelines for repository contributors - User Documentation](https://help.github.com/articles/setting-guidelines-for-repository-contributors/)
  - move CONTRIBUTING.md to `.github/` to declutter the root directory
  - replace `--` by unicode em-dash `—` (explained in the commit itself)
- [Creating an issue template for your repository - User Documentation](https://help.github.com/articles/creating-an-issue-template-for-your-repository/)
  - excerpt from CONTRIBUTING.md
  - As someone who has abused the issue tracker in the past (and sometimes still do), putting the message right at the place people is typing should be most effective.
- [Creating a pull request template for your repository - User Documentation](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/)
  - excerpt from CONTRIBUTING.md
  - As someone who has violated rule no. 1 before (Each patch (commit) should make a single logical change. Not sure how to interpret though. e.g. In this pull request I tried to make a commit for each logical step), putting the message right at the place people is typing should be most effective.

You can see how the issue template and pull request template work in [Demo: ickc/test-dot-github](https://github.com/ickc/test-dot-github):
- try creating an issue
- make a pull request from `pull-request` to `master`

PR #3252, opened 2016-11-24
CONTRIBUTING.md: minor change in wording and punctuation

ickc:
Split off from #3193, let each pull request deal with a slightly different issue:

- em-dash should be used here, not en-dash (which is for, e.g., ranges).
- unicode em-dash is used because GitHub don’t do SmartyPants on dashes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,16 +14,16 @@ for anything relevant.
 Make sure you can reproduce the bug with the latest released version of pandoc
 (or, even better, the development version).
 
-Your report should give detailed instructions for how to reproduce the problem,
-including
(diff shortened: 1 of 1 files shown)


ickc:
Note that the pull request [Job #3983.7 - jgm/pandoc - Travis CI](https://travis-ci.org/jgm/pandoc/jobs/178477698) failed, because `fatal: unable to access 'https://github.com/jgm/pandoc.git/': Could not resolve host: github.com`. It is not the first time it is seen. Perhaps we could move the macOS build to `allow_failure`?

jgm:
>  Perhaps we could move the macOS build to allow_failure?

Better to have occasional false negatives than false positives.

Issue #3197, opened 2016-10-29
top-level-division not replacing chapters with sections in reports


jbertran:
The `--top-level-division=section` option seems to be ineffective with the `report` documentclass.
### Sample input

```
%My title
%Me

# Heading 1
# Heading 2
```
### Command line

`pandoc test.md -o test.pdf --top-level-division=section --variable documentclass=report`

Using pandoc 1.18 installed by downloading the Debian package from github.
### Expected

A report with \section as the top level division
### Obtained

A standard report with \chapter as the top level division, \subsection as the next division level (checked by generating latex instead of pdf).

jkr:
confirmed. @tarleb: the issue seems to come form 754-55 in the LaTeX
writer:

~~~{.haskell}
  let topLevelDivision = min (if book then Chapter else Section)
                             (writerTopLevelDivision opts)
~~~

The effect is that if it has a book template, it ignores
`top-level-division` set to `section`. 

The issue, I think, is that it can't differentiate between `Section` as
the default division, and it being explicitly set. The easiest thing to
do would be to make `writerTopLevelDivision` into a `Maybe Division`, and
only set it `Just Section` if that's explicitly set. So 754-- would
become:

~~~{.haskell}
let topLevelDivision = case writerTopLevelDivison opts of
                         Just div -> div
						 Nothing -> if book then Chapter else Section
~~~

plus whatever other fiddling might be necessary. Does this make sense?

tarleb:
I guess either my thinking was along the lines of "filters can do that if needed". @jbertran, you can get the desired behavior by calling this snippet as a filter (save to a file and pass it as the argument of `--filter`):

``` haskell
#!/usr/bin/env runhaskell
import Text.Pandoc.JSON

main :: IO ()
main = toJSONFilter promoteHeader

promoteHeader :: Block -> Block
promoteHeader (Header n attr title) = Header (n + 1) attr title
promoteHeader b = b
```

I agree with you analysis @jkr, I should have implemented it like that in the first place. Changing it now would mean yet another API change.  @jgm, would you object if we were doing that?

jgm:
Another option of course would be to remove the behavior of setting the default to chapters for book classes, and require people to specify this explicitly on the command line.

jkr:
[...]

PR #3258, opened 2016-11-26
Allow to overwrite top-level division type heuristics

tarleb:
Pandoc uses heuristics to determine the most resonable top-level
division type when emitting LaTeX or Docbook markup.  It is now possible
to overwrite this implicitly set top-level division via the
`top-level-division` command line parameter.

API change (`Text.Pandoc.Options`): the type of the
`writerTopLevelDivision` field in of the `WriterOptions` data type is
altered from `Division` to `Maybe Division`. The field's default value
is changed from `Section` to `Nothing`.

Closes: #3197

### DIFF
--- a/pandoc.hs
+++ b/pandoc.hs
@@ -181,7 +181,7 @@ data Opt = Opt
     , optHtmlQTags         :: Bool    -- ^ Use <q> tags in HTML
     , optHighlight         :: Bool    -- ^ Highlight source code
     , optHighlightStyle    :: Style   -- ^ Style to use for highlighted code
-    , optTopLevelDivision  :: Division -- ^ Type of the top-level divisions
+    , optTopLevelDivision  :: Maybe Division -- ^ Type of the top-level divisions
     , optHTMLMathMethod    :: HTMLMathMethod -- ^ Method to print HTML math
     , optReferenceODT      :: Maybe FilePath -- ^ Path of reference.odt
     , optReferenceDocx     :: Maybe FilePath -- ^ Path of reference.docx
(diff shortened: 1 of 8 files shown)


jgm:
This all looks good, but I just had an idea.

Instead of making the field a Maybe Division, why not just add a Default constructor to the Division type?  Then we'd only need a third-field version bump (unless I'm mistaken).  And, to me, it seems a bit more natural saying that the top level division is "Default" than saying that it's "Nothing."

Default would normally be section but could go to Chapter when a book class is detected.

Thoughts on this?

jgm:
I went ahead and merged this.  Still interested in thoughts about the above alternative, though it's not really clear to me that it's better than using the Maybe.

tarleb:
John MacFarlane <notifications@github.com> writes:

> This all looks good, but I just had an idea.
>
> Instead of making the field a Maybe Division, why not just add a Default
> constructor to the Division type? Then we'd only need a third-field version bump
> (unless I'm mistaken). And, to me, it seems a bit more natural saying that the
> top level division is "Default" than saying that it's "Nothing."
>
> Default would normally be section but could go to Chapter when a book class is
> detected.
>
> Thoughts on this?

I had similar thoughts while implementing this. I almost replaced all
mentions of `Nothing` in the new code with `def`, which would make it
easier to switch between possible representations for a default value. I
decided against it, simply because the additional typeclass could
increase the mental burden for readers of the code. Not sure if that's
actually true.

My reasoning against a "Default" constructor was even more mundane: The
name of the "Division" data-type implies that it could be extened with
more subsection/paragraph constructors and then serve as an alternative
represenation of header levels. I should have chosen a different name
for that type, one that is more in line with the actual use of this type
in the context of "top-level divisions". I'd prefer to rename the
data-type if we were to go the "Default" constructor route. That, of
course, would revert the advantage with regard to API changes.

I actually prefer the "Default" constructor, now that I'm rethinking
this. Do you have reservations against renaming `Division` to
`TopLevelDivision`?

jgm:
+++ Albert Krewinkel [Nov 26 16 13:42 ]:
>   I actually prefer the "Default" constructor, now that I'm rethinking
>   this. Do you have reservations against renaming `Division` to
>   `TopLevelDivision`?

No, that seems appropriate to me!

PR #3267, opened 2016-11-30
travis: use language generic

ickc:
See commercialhaskell/stack#2784:

>Although `language: generic` is undocumented, it is widely used...

It doesn’t give any visible speed boost, if any, probably only shave dozens seconds in boot up time. But since commercialhaskell/stack accepted this pull request, I figure it at least does no harm to make into pandoc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@
 # Use new container infrastructure to enable caching
 sudo: false
 
-# Choose a lightweight base image; we provide our own build tools.
-language: c
+# Do not choose a language; we provide our own build tools.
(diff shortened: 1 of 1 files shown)


ickc:
Oops. I see that travis on jgm/pandoc is failing. It is passing in mine though: [ickc/pandoc - Travis CI](https://travis-ci.org/ickc/pandoc).

jgm:
The travis failure is nothing to do with your commit, which I've merged.
It should be fixed in a half hour or so, when stackage mirrors have texmath 0.9.

PR #3273, opened 2016-12-03
Minor spelling typos in the manual

anthony-geoghegan:
Hi John,

While reading the Pandoc manual on the website, I noticed that the `--list-hightlight-styles` had an extra `t`. I created this pull request to fix this type and another spelling mistake. In a separate commit, I also fixed the case for a few other nouns, e.g., “ASCII”, “Unicode”, “Latin”, “JavaScript”, “CSS” (I used a separate commit in case you don’t want to merge those changes).

I noticed that the same spelling mistakes are in `man/pandoc.1` but I figured that the Troff file is automatically generated from the Markdown file.

Thanks for creating such a great program (and for CommonMark),
Anthony


jgm:
Thanks, that's helpful!

PR #3301, opened 2016-12-12
also wait for LatexMathML when rendering to wkhtmltopdf

oli-obk:


oli-obk:
can't figure this out...

PR #3304, opened 2016-12-12
Speed up Travis

ickc:
- travis use `fast_finish`
- cabal uses `-j`

I added the comments on fast_finish into the file directly for future reference. This will gives the most "speed up" since it will returns the "green light" when the 2 `allow_failure` builds aren't finished (useful for pull request when waiting for the green light).

The cabal `-j` means it will set parallel jobs according to the number of CPUs. It only speed ups when there're multiple dependencies that can be installed at the same time. I noticed that gitit's travis build uses `-j2` while it isn't used in pandoc. So may be there's a reason behind that?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,7 +104,7 @@ install:
     cabal)
       cabal --version
       travis_retry cabal update
-      cabal install --only-dependencies -ffast --enable-tests --enable-benchmarks --force-reinstalls --ghc-options=-O0 --reorder-goals --max-backjumps=-1 $CABALARGS
+      cabal install -j --only-dependencies -ffast --enable-tests --enable-benchmarks --force-reinstalls --ghc-options=-O0 --reorder-goals --max-backjumps=-1 $CABALARGS
       ;;
   esac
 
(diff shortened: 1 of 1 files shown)


ickc:
Another possible "speed up" is to drop the cabal build with HEAD: I noticed that build almost always fails. It might just well be dropped to save travis' computer time (and to devote more thread to other builds by any of the projects under jgm).

ickc:
Another possible drop: `stack nightly osx`, since `stack nightly` is already tested in Ubuntu. As far as testing "if it works for OS X" is concerned, the one `stack 8.0.1 osx` build already shows that.

As far as I know, Travis's OS X builds uses a completely different service, has 1 core only, and is not as reliable as the Ubuntu images. (more down time; and as seen in previous cases, "mysteriously" failed a build).

jgm:
I'd be okay with dropping the osx nightly build, if you want to add that to this PR.
I'd prefer to keep the HEAD build for now.

ickc:
I remove the stack nightly in osx build by commenting it, so it is easier to get it back in the future if needed.

jgm:
OK let's see how much this helps!

Issue #2752, opened 2016-03-02
ODT tables content disappear


matt32106:
Looks like tables content is not converted properly.

Steps to reproduce:
Create a test.odt file with a simple 2x2 table __with some text values in each cell__.
[test.odt file](https://dl.dropboxusercontent.com/u/1307844/test.odt)
Run 
`pandoc -s -o test.tx -t textile -f odt test.odt`

The resulting test.tx file only contains the skeleton of the table but no values:
`||`
`||`

I am using 
**Windows 7 Enterprise**
**LibreOffice
Version: 5.1.0.3**
Build ID: 5e3e00a007d9b3b6efb6797a8b8e57b51ab1f737
CPU Threads: 4; OS Version: Windows 6.1; UI Render: default; 
Locale: fr-FR (fr_FR)

and 
cmd>pandoc --version
**pandoc 1.16.0.2**
Compiled with texmath 0.8.4.1, highlighting-kate 0.6.1.

trying to convert to html instead with 
`pandoc -s -o test.html -t html -f odt test.odt`

produces the following file:

```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
  <meta http-equiv="Content-Style-Type" content="text/css" />
  <meta name="generator" content="pandoc" />
  <title></title>
  <style type="text/css">code{white-space: pre;}</style>
</head>
<body>
<table>
<tbody>
<tr class="odd">
</tr>
<tr class="even">
</tr>
</tbody>
</table>
<p></p>
<p></p>
<p></p>
</body>
</html>
```

as the table is also empty, it probably means that the problem is on the ODT reading side.

matt32106:
trying to convert to html with 
`pandoc -s -o test.html -t html -f odt test.odt`

produces the following file:
```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
  <meta http-equiv="Content-Style-Type" content="text/css" />
[...]

PR #3306, opened 2016-12-13
Test case for bug 2752

hubertp-lshift:
Looks like issue #2752 is bogus but I added the test case just in case.


jgm:
OK, thanks, I'll close that issue.

PR #3310, opened 2016-12-13
Test case for not putting a dot '.' in id.

greut:
There is a problem when a title contains a . somewhere in the middle as
it gets into the generated id. That id cannot be matched by a CSS
selector as a dot marks a classname.

```markdown
#### Level 4 with a '.' (dot)
```

```html
<h4 id="level-4-with-a-.-dot">...
```

If you can give me any hints where to look for doing the fix. I'd be
glad to learn how.

Signed-off-by: Yoan Blanc <yoan@dosimple.ch>


jgm:
You can match a dot with CSS, you just have to backslash-escape it:

```
h1#my\.identifier { color: red; }
```

Periods are perfectly legal in identifiers, and if we substitute them for something else, we make the automatically generated identifier less easy to predict.  I'm willing to consider options here, though, if you have something concrete to suggest.

greut:
My bad! Let's not change anything then. Cheers,

PR #3316, opened 2016-12-15
MANUAL/README: Fixed broken links

ickc:
- ConTeXt garden: trivial editing
- groff man: Apple's link returned 401. But if clicked through, it actually works. Nonetheless, I replace the link to GNU's version since it is the source.
- docx openxml: the link is broken. I replaced it with the best I can find. Please double check it is ok.

By the way, when building the `man`, it got errors. It doesn't affect this commit because the `man` doesn't have those links. The error is:

```bash
$ make man/pandoc.1
pandoc MANUAL.txt -t man -s --template man/pandoc.1.template \
	--filter man/capitalizeHeaders.hs \
	--filter man/removeNotes.hs \
	--filter man/removeLinks.hs \
	--variable version="pandoc 1.19.1" \
	-o man/pandoc.1
pandoc: Error running filter man/capitalizeHeaders.hs
fd:4: hPutBuf: resource vanished (Broken pipe)
make: *** [Makefile:41: man/pandoc.1] Error 83
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Markdown can be expected to be lossy.
 [LaTeX]: http://latex-project.org
 [`beamer`]: https://ctan.org/pkg/beamer
 [Beamer User's Guide]: http://ctan.math.utah.edu/ctan/tex-archive/macros/latex/contrib/beamer/doc/beameruserguide.pdf
-[ConTeXt]: http://contextgarden.net/
+[ConTeXt]: http://www.contextgarden.net/
 [RTF]: http://en.wikipedia.org/wiki/Rich_Text_Format
 [DocBook]: http://docbook.org
 [txt2tags]: http://txt2tags.org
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,13 +83,13 @@ Markdown can be expected to be lossy.
...
 [Haddock markup]: https://www.haskell.org/haddock/doc/html/ch03s08.html
-[groff man]: http://developer.apple.com/DOCUMENTATION/Darwin/Reference/ManPages/man7/groff_man.7.html
+[groff man]: https://www.gnu.org/software/groff/
 [Haskell]: https://www.haskell.org
 [GNU Texinfo]: http://www.gnu.org/software/texinfo/
 [Emacs Org mode]: http://orgmode.org
 [AsciiDoc]: http://www.methods.co.nz/asciidoc/
 [DZSlides]: http://paulrouget.com/dzslides/
-[Word docx]: http://www.microsoft.com/interop/openup/openxml/default.aspx
+[Word docx]: https://msdn.microsoft.com/en-us/library/gg278315.aspx
```
jgm: This points to the SDK for OpenXML.  It would be better to point to the OpenXML documentation or spec.

jgm:
Perhaps the filter is failing because you don't have needed libraries installed?
Try loading `ghci man/capitalizeHeaders.hs` and you should get more info.
You may need to `cabal install pandoc-types`.

README.md
ickc: The openxml link is difficult to find. Here's some info I have:

- Using wayback machine, the last known copy of the original link is in [Microsoft || Open Up](https://web.archive.org/web/20120306024812/http://www.microsoft.com/interop/openup/openxml/default.aspx) from 2012. This can be used to copy keywords to search for possible match (but can't seem to find a match from microsoft.com, so they probably have eliminated the page entirely).

- From within microsoft.com, the best I can find is [Office Open XML, ECMA-376, and ISO/IEC 29500](https://msdn.microsoft.com/en-us/library/office/gg607163(v=office.14%29.aspx).

- From iso.org: [ISO/IEC 29500-1:2016 - Information technology -- Document description and processing languages -- Office Open XML File Formats -- Part 1: Fundamentals and Markup Language Reference](http://www.iso.org/iso/home/store/catalogue_ics/catalogue_detail_ics.htm?csnumber=71691). This one requires a couple clicks and "I agree" to download the spec.

- From coverpages.org: [Cover Pages: Office Open XML File Formats Published as ISO/IEC 29500:2008 Final Standard.](http://xml.coverpages.org/ni2008-11-21-a.html). With some clicks, this one link back to the iso.org with the "I agree" button.

jgm:
We could just link to

https://en.wikipedia.org/wiki/Office_Open_XML

which has links to various other places?

ickc:
Done. I squashed them together too. I see that quite a few other formats link to wikipedia's pages, so it is probably easiest to do so.

jgm:
Great, thanks

PR #3326, opened 2016-12-24
markdown reader: disallow space between inline code and attributes

### LINKED ISSUE #3323: Header attributes are ignored when header contains an inline code block
I'm in version 1.19.1 and have the following output when converting from Markdown to LaTeX:

```
➜  ~ echo '# foo {#bar}' | pandoc -t latex 
\section{foo}\label{bar}
➜  ~ echo '# `foo` {#bar}' | pandoc -t latex
\section{\texorpdfstring{\texttt{foo}}{foo}}\label{foo}
```

In the second example, I expected the section label to be `{#bar}`, as in the first case.

Converting the same Markdown to HTML shows that the attribute is being associated to the code block:

```
➜  ~ echo '# `foo` {#bar}' | pandoc -t html 
<h1 id="foo"><code id="bar">foo</code></h1>
```

Is it the expected behavior?

mb21:
closes #3323

I think the change is worth the risk of possibly breaking a small handful of existing documents... maybe discussion needed?

### DIFF
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -1561,8 +1561,8 @@ code = try $ do
                        (char '\n' >> notFollowedBy' blankline >> return " "))
                       (try (skipSpaces >> count (length starts) (char '`') >>
                       notFollowedBy (char '`')))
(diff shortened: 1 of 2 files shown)


jgm:
I agree.  I don't think many people will have inserted the space, anyway.

PR #3327, opened 2016-12-24
HTML writer: don't process pars with empty RawInline, fixes #1040

### LINKED ISSUE #1040: Don't emit empty paragraphs in HTML writer
For example, if a paragraph only has raw LaTeX (which cannot be written in HTML), an empty paragraph is created when converting to HTML:

```
echo 'Hello

\bigskip

World' | pandoc -f markdown -t html
```

```
<p>Hello</p>
<p></p>
<p>World</p>
```

This is probably just a cosmetic issue as the extra tags don't seem to have any negative effects, but code generated by Pandoc is generally quite nice to look at so it'd be nice if they were gone too.

---

Other HTML tags beside `p` might also be affected if pandoc is creating them implicitly. But empty tags explicitly requested by the user (like below) probably shouldn't be removed, so it's not enough to simply remove all empty tags before rendering the tree.

```
<p></p>
```

```
<div></div>
```

``````
```
```
``````

```
#
```

### PR DESCRIPTION
**Pull request by mb21**

### CODE REVIEW

**Comment by jgm** [comment]
The build failures due to happy should be fixed by the latest update to .travis.yml.
But there is one warning that should be fixed before we merge:
```
/home/travis/build/jgm/pandoc/src/Text/Pandoc/Writers/HTML.hs:460:29: warning: [-Wunused-matches]
    Defined but not used: ‘str’
```

mb21:
of course... btw, I'm getting quite a few warnings in the markdown reader with `stack build --ghc-options="-Wall"`

    .../Readers/Markdown.hs:1891:3: warning: [-Wunused-do-bind]
    A do-notation statement discarded a result of type ‘String’

jgm:
I always use `"-Wall -fno-warn-unused-do-bind"`.

jgm:
Note also that the default ghc options (from the cabal file) include `-Wall -fno-warn-unused-do-bind`.  So you shouldn't need to specify ghc-options separately on stack.

mb21:
ah right, then I must have overlooked it...

mb21:
(fixed and rebased btw)

jgm:
OK, I see the problem with the builds that failed.  Fixed now I think.
I'll merge this, thanks!

PR #3333, opened 2016-12-29
Add '@*' usage

Jmuccigr:
Added to nocite section to add the wildcard functionality.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3377,6 +3377,14 @@ In this example, the document will contain a citation for `item3`
 only, but the bibliography will contain entries for `item1`, `item2`, and
 `item3`.
 
+It is possible to create a bibliography with all the citations, 
+whether or not they appear in the document, by using a wildcard:
+
+    ---
+    nocite: |
+      '@*'
+    ...
+
 For LaTeX or PDF output, you can also use [`natbib`] or [`biblatex`]
 to render bibliography. In order to do so, specify bibliography files as
 outlined above, and add `--natbib` or `--biblatex` argument to `pandoc`
@@ -4093,7 +4101,7 @@ Hubert Plociniczak,
 Ilya V. Portnov,
 Ivo Clarysse,
 J. Lewis Muir,
-Jaime Marquínez Ferrándiz,
+Jaime Marquínez Ferrándiz,
 Jakob Voß,
 James Aspnes,
 Jamie F. Olson,
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3382,7 +3382,7 @@ whether or not they appear in the document, by using a wildcard:
 
     ---
     nocite: |
-      '@*'
+      @*
     ...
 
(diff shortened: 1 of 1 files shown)


MANUAL.txt
```diff
@@ -3377,6 +3377,14 @@ In this example, the document will contain a citation for `item3`
 only, but the bibliography will contain entries for `item1`, `item2`, and
 `item3`.
 
+It is possible to create a bibliography with all the citations, 
+whether or not they appear in the document, by using a wildcard:
+
+    ---
+    nocite: |
```
jgm: Is this right?  I thought that with the `|` style of YAML quoting, you do not need the single quotes, since the single quotes will be taken as literal.  I think they should be omitted (and of course this should be tested)>

MANUAL.txt
```diff
@@ -4093,7 +4101,7 @@ Hubert Plociniczak,
 Ilya V. Portnov,
 Ivo Clarysse,
 J. Lewis Muir,
-Jaime Marquínez Ferrándiz,
+Jaime Marquínez Ferrándiz,
```
jgm: Why the diff on this line?

Jmuccigr: You tell me. No clue.

MANUAL.txt
Jmuccigr: I had a similar thought and tested with the single quotes, which worked. I've been using this in a bash script for a while now, so I was pretty sure it was ok. I was going on your suggestion, actually, @jgm : https://groups.google.com/d/msg/pandoc-discuss/mp0PoYmWtGk/m24UwCeyBgAJ

That said, it seems like the unquoted version works as well...though I'm not sure there aren't circumstances where the quoting is needed.

MANUAL.txt
jgm: Probably in one case it's a combining character accent; in the other case a combined accented character.  The latter (where you just need one backspace to delete both) is probably better; still, this doesn't belong with the other change in this PR, so can you revert this part?

MANUAL.txt
jgm: I'd expect it to work both with and without the single quotes (with the single quotes, you're introducing a citation within quotes, but since it's not printed it doesn't matter).

And of course the single quotes would be needed if you didn't use the `|` quoting mechanism (as was the case in the example you link to).

So, either

```
nocite: |
  @*
```
or
```
nocite: '@*'
```

Jmuccigr:
OK, I had to edit locally to restore the accents, as I couldn't get the web editor to leave the uncombined versions. While I was at it, I got rid of a trailing space I had in the new text. Let me know, @jgm, whether you prefer the quoted or unquoted form of the wildcard.

jgm:
I don't really have a strong preference between

    nocite: '@*'

and

    nocite: |
      *@

The former is shorter; the latter is more like the previous
example in the manual.

Jmuccigr:
OK, went with the more consistent version, instead of the shorter.

jgm:
Thanks!

PR #3358, opened 2017-01-17
Fix sample.lua barfing on Raw data

bumper314:
Fix for "pandoc: user error (Incorrect result type (string expected, got nil))." when the source format contains Raw data.

### DIFF
--- a/data/sample.lua
+++ b/data/sample.lua
@@ -173,8 +173,9 @@ end
 function RawInline(format, str)
   if format == "html" then
     return str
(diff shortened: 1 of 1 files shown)


bumper314:
Test file: http://d.pr/f/HhqU

`pandoc -t data/sample.lua test.org`

jgm:
Thanks!  Could we make it an if...then...else structure instead?
That just makes it clearer what it going on (to me anyway).

PR #3396, opened 2017-01-30
EPUB: improve SVG image output, closes #2766

### LINKED ISSUE #2766: SVG best practice for epub3
I'm following "Epub 3 Best Practices" by Matt Garrish and Markus Gylling to find the best way to embed svg images in an epub3 file.

I'm trying to provide a fallback to PNG with Pandoc, but I'm unable to do so. The following is recommended as the best way to include svg pics in an epub:

```XML
<object data="image01.svg" type="image/svg+xml" width="640" height="480">
<img src="image01.png" alt="..." />
</object>
```

The reason being that you can show the svg for ereaders and browsers that support it and, at the same time, provide a fallback for those that don't.

However, the markdown link format can only create an ``. If there isn't a method to achieve this, it would be great to tell Pandoc in the markdown that a given svg file has a png/jpeg fallback associated to it.

mb21:
This is not ready to merge yet...

Can we use `System.Directory.doesFileExist` instead of `P.fetchItem` somehow? Also, `fetchItem` appears to make pandoc fail if the file is not found even though I'm checking for `:: PandocError` (which I copied from the ICML writer).

### DIFF
--- a/src/Text/Pandoc/Writers/EPUB.hs
+++ b/src/Text/Pandoc/Writers/EPUB.hs
@@ -34,7 +34,7 @@ import qualified Data.Set as Set
 import Data.Maybe ( fromMaybe, catMaybes )
 import Data.List ( isPrefixOf, isInfixOf, intercalate )
 import Text.Printf (printf)
-import System.FilePath ( takeExtension, takeFileName )
+import System.FilePath ( dropExtension, takeExtension, takeFileName )
 import Network.HTTP ( urlEncode )
 import qualified Data.ByteString.Lazy as B
 import qualified Data.ByteString.Lazy.Char8 as B8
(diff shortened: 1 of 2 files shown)


jgm:
+++ Mauro Bieg [Jan 30 17 02:33 ]:
>   This is not ready to merge yet...
>
>   Can we use System.Directory.doesFileExist instead of P.fetchItem
>   somehow?

Since you're in PandocMonad you can only use things defined
there.  Of course, if it turns out we need something like
doesFileExist there we could consider adding a function like
that...

>Also, fetchItem appears to make pandoc fail if the file is not
>   found even though I'm checking for :: PandocError (which I copied from
>   the ICML writer).

Is it failing with a PandocError?

mb21:
> Since you're in PandocMonad you can only use things defined there.

That's what I figured...

> Of course, if it turns out we need something like doesFileExist there we could consider adding a function like that...

I'm not sure how smart the lazy evaluation of `fetchItem` is. But I suspect it will read the whole file into memory. So it feels cleaner to add a `doesFileExist`. Or should it be `doesItemExist` and try to fetch remote resources as well?

> Is it failing with a PandocError?

I think it was, yes. Probably a `PandocFileReadError` (it was outputting ` problem reading media`) The code that caused it was:

    runExceptT $ lift $ P.fetchItem (writerSourceURL opts) src

I have now changed it to using `catchError (do _ <- P.fetchItem ...` but this seems somewhat of a hack...

jgm:
If we end up merging some form of #2211, then we could simply generate the png fallback from the svg and we wouldn't need to check for its existence...

mb21:
Good point. Instead, we could also use `rasterific` in a filter that does the necessary transformations on all referenced SVG images. (Depending on the output format it would fallback to png or pdf.) Then again, from a user perspective it's always nice to have more stuff built in.

mb21:
This doesn't seem to be the right approach anymore. See e.g. https://github.com/jgm/pandoc/issues/4058#issuecomment-424613868

PR #3398, opened 2017-01-31
HTML and DocBook writers: fix internal links with writerIdentifierPrefix opt

### LINKED ISSUE #3397: Option --id-prefix does not update references within the document
I found --id-prefix to be a nice solution to my problem with composing a book of multiple markdown documents but I have the same problem as this guy with references not being updated to point to the new IDs:
https://groups.google.com/forum/#!topic/pandoc-discuss/SQ0_mE2TUvM

In short
```
$ cat test.md
# Markdown Test #

## Introduction ##

This is the intro. See the [Introduction].
```
yields this output
```
$ pandoc --id-prefix=ss- -t docbook  test.md
<sect1 id="ss-markdown-test">
  <title>Markdown Test</title>
  <sect2 id="ss-introduction">
    <title>Introduction</title>
    <para>
      This is the intro. See the
      <link linkend="introduction">Introduction</link>.
    </para>
  </sect2>
</sect1>
```
where the link to Introduction is broken.

Shouldn't --id-prefix append the prefix also to references within the document?

mb21:
fixes #3397

btw, the Haddock writer also uses `writerIdentifierPrefix` for its fake notes, which is not documented in the README... not sure whether to change the writer or the README...

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -933,9 +933,11 @@ inlineToHtml opts inline = do
                         linkText <- inlineListToHtml opts txt
                         slideVariant <- gets stSlideVariant
                         let s' = case s of
-                                      '#':xs | slideVariant == RevealJsSlides
-                                        -> '#':'/':xs
-                                      _ -> s
+                                   '#':xs -> let prefix = if slideVariant == RevealJsSlides
+                                                             then "/"
+                                                             else writerIdentifierPrefix opts
+                                             in  '#' : prefix ++ xs
+                                   _ -> s
                         let link = H.a ! A.href (toValue s') $ linkText
                         let link' = if txt == [Str (unEscapeString s)]
                                        then link ! A.class_ "uri"
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -733,10 +733,10 @@ Options affecting specific writers
 
 `--id-prefix=`*STRING*
 
-:   Specify a prefix to be added to all automatically generated identifiers
-    in HTML and DocBook output, and to footnote numbers in Markdown output.
-    This is useful for preventing duplicate identifiers when generating
-    fragments to be included in other pages.
+:   Specify a prefix to be added to all identifiers and internal links
+    in HTML and DocBook output, and to footnote numbers in Markdown
+    and Haddock output. This is useful for preventing duplicate
+    identifiers when generating fragments to be included in other pages.
 
 `-T` *STRING*, `--title-prefix=`*STRING*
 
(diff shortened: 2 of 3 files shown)


src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -933,9 +933,11 @@ inlineToHtml opts inline = do
                         linkText <- inlineListToHtml opts txt
                         slideVariant <- gets stSlideVariant
                         let s' = case s of
-                                      '#':xs | slideVariant == RevealJsSlides
-                                        -> '#':'/':xs
-                                      _ -> s
+                                   '#':xs -> let prefix = if slideVariant == RevealJsSlides
```
jgm: This fix adds the prefix to all links starting with '#'.  This goes too far, I think.  The `--id-prefix` option as documented is only supposed to modify automatically generated identifiers.  So how do we know if the link is to an automatically generated identifier or a manually inserted one?

jgm: I'm thinking we may need to address this at the reader level, at least in part.  If the readers added a special class to automatically generated identifiers, we could look at that in the writers.  Better, the readers could simply add the id prefix.  This would require exposing the id prefix as a reader option, too, which is a bit ugly.  Maybe the cleanest solution would be to put it in CommonState in PandocMonad, then writers and readers could use it (and pandoc.hs would have to set it inside runIO.)

mb21: That's what I first thought as well, but then I realized the current behaviour breaks manually set links as well:

    $ pandoc --id-prefix=pp-
    # head {#myhead}

    {#myimg}

    [link to head](#myhead) 

    [link to img](#myimg)



    <h1 id="pp-myhead">head</h1>
    <div class="figure">
    

    </div>
    <p><a href="#myhead">link to head</a></p>
    <p><a href="#myimg">link to img</a></p>

That's why I figured the behaviour of `--id-prefix` should be to simply prepend it to _all_ `id`s and all links that start with `#`.  This way, someone using pandoc in a wiki or CMS doesn't have to trust the input document to not clash with `id`s on the rest of the page (users of the wiki might set `id`s manually unaware the admin uses those `id`s on the page). This behaviour sounds even more useful than having only auto-generated links be prefixed and leaving the manual links alone. I don't think the readers should be bothered with this. Unless I'm missing something...?

jgm:
+++ Mauro Bieg [Feb 01 17 04:57 ]:

>   That's why I figured the behaviour of --id-prefix should be to simply
>   prepend it to all ids and all links that start with #. This way,
>   someone using pandoc in a wiki or CMS doesn't have to trust the input
>   document to not clash with ids on the rest of the page (users of the
>   wiki might set ids manually unaware the admin uses those ids on the
>   page). This behaviour sounds even more useful than having only
>   auto-generated links be prefixed and leaving the manual links alone. I
>   don't think the readers should be bothered with this. Unless I'm
>   missing something...?

You have a point.  Well, maybe we should just update the
documentation, then.

mb21:
> Well, maybe we should just update the documentation, then.

Right, I think it's consistent with the code now.

flokli:
This might be related: https://github.com/jgm/pandoc/issues/6329

PR #3403, opened 2017-01-31
Fixes for zimwiki writer

alexivkin:
Improvements for ZimWiki's source view and table markup.


src/Text/Pandoc/Writers/ZimWiki.hs
```diff
@@ -44,20 +44,21 @@ import Data.Default (Default(..))
...
   }
 
 instance Default WriterState where
-  def = WriterState { stItemNum = 1, stIndent = "" }
+  def = WriterState { stItemNum = 1, stIndent = "", stInTable = False }
 
 -- | Convert Pandoc to ZimWiki.
 writeZimWiki :: PandocMonad m => WriterOptions -> Pandoc -> m String
-writeZimWiki opts document = return $ evalState (pandocToZimWiki opts document) (WriterState 1 "")
+writeZimWiki opts document = return $ evalState (pandocToZimWiki opts document) (WriterState 1 "" False)
```
tarleb: The WriterState seems to be def, maybe it would make sense to use that instead?

alexivkin: Def?

tarleb: The `def` value from the `Default` type class.

alexivkin: I have no idea how to do that. Are there any existing examples in the code?

mb21: I think what @tarleb is proposing is simply: `evalState (pandocToZimWiki opts document) def`

alexivkin:
Fixed to def and added 2 more fixes. Not shure why travis checks never complete.

mb21:
@alexivkin e.g. [this travis job](https://travis-ci.org/jgm/pandoc/jobs/201327537) is failing because of the zimwiki test. Try `stack test` to run the tests on your local system... and update the `/test/writer.zimwiki` file if necessary...

alexivkin:
updated the tests

alexivkin:
The tests work for zim, but fail on Latex - completely unrelated to my code. It's something in the previous main branch commits that triggered it.

jgm:
The rebase has gotten pretty hairy.  You added this at a bad time, with lots of churn in the code base.  Sorry this has been difficult.

Suggestion:

1.  remove this PR
2.  copy your src/Text/Pandoc/Writers/ZimWiki.hs, test/writer.zimwiki, and test/table.zimwiki to a safe place.
3.  check out current pandoc master
4.  copy the three files backed up above to their appropriate places in master
5.  git add and commit them.  **Be sure to include a full commit message, describing the changes you've made..."fixed the writer" is not enough.**  I'll need this for the changelog.
6.  stack test
7.  assuming all goes well, create a new PR.  This will be a single clean commit on top of master, with a good description of the changes, and should be mergeable.

alexivkin:
Done. https://github.com/jgm/pandoc/pull/3446
Should have I committed to a branch, instead of master, to avoid trying to catch up with the head?

mb21:
great!

> Should have I committed to a branch, instead of master, to avoid trying to catch up with the head?

commiting to a branch frees the master branch of your own repo to work on other features while you wait for jgm to merge your pull request into the official master. you always have to catch up with master (best to rebase, not merge), but this time you were just unlucky that there were a lot of changes in the meantime.

jgm:
I don't know, perhaps that would have made things easier,
but I think the main problem is the large number of changes
to HEAD in this period.

+++ Alex Ivkin [Feb 14 17 17:01 ]:

PR #3409, opened 2017-02-02
Org.hs: remove misleading comment

ilabdsf:
This comment is likely copied from RST.hs where 'refs' variable indeed exists, but makes no sense here.


tarleb:
Thanks!

PR #3419, opened 2017-02-04
Added missing pipe

ui4j:


jgm:
This isn't right.  The pipe is only needed between header and footer.

PR #3421, opened 2017-02-05
Allow user to specify User-Agent

### LINKED ISSUE #3384: User Agent argument switch if downloading content/images
Is it possible to allow the user to select/input a specific user agent for when pandoc retrieves files or data from a server? 

For instance, I am unable to retrieve images from Facebook, which instead complains that the browser I am using is not supported. Unfortunately, I could not find a way to change the user agent myself. 

Many thanks in advance :)

thenaesh:
Fixes #3384 

When requesting a document from a URL, a user can now specify the User-Agent by doing something like:
```
USER_AGENT="Mozilla/5.0" pandoc -f html -t markdown http://www.fsf.org
```

This should fix the problem of Pandoc appearing as an "unsupported browser" to some websites. :)

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -120,6 +120,11 @@ pandoc will fetch the content using HTTP:
 
     pandoc -f html -t markdown http://www.fsf.org
 
+It is possible to supply a custom User-Agent string when requesting a
+document from a URL, by setting an environment variable:
+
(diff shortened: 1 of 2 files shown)


jgm:
Thanks.  I don't understand the build failure with one ghc version on travis, but we'll figure that out next.

PR #3438, opened 2017-02-12
Org anchor links

ilabdsf:
According to [The Org Manual](http://orgmode.org/manual/Internal-links.html#Internal-links), internal links should contain `#`. Links without `#`
> lead to a text search in the current file

Likely stripping `#` was not intended.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -530,6 +530,7 @@ Test-Suite test-pandoc
                   Tests.Writers.Docbook
                   Tests.Writers.HTML
                   Tests.Writers.Markdown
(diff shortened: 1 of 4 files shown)


mb21:
sounds reasonable, this was introduced with https://github.com/jgm/pandoc/commit/dbf4d77091f8bcde6d43f0945baf4905c18aebfc @tarleb ?

tarleb:
> this was introduced with dbf4d77

I don't recall what my intentions were, but the old code does look wrong.

Thanks @ilabdsf!

Issue #3431, opened 2017-02-09
Add a template variable containing the name of the source file


rohieb:
I keep several Markdown documents in a Git repository which are rendered to MediaWiki through a custom template. In the MediaWiki template, I want to add a "this file is autogenerated" warning, with a link to the source file in the Git repository. For this reason, it would be nice to have a template variable, e.g. `$filename$` or `$jobname$` (inspired by LaTeX), so my template could have:

``` html
<!-- Warning: this file is auto-generated!
     See instead: https://github.com/stratum0/dokumente/$filename$
  -->
```

jgm:
A reasonable request.  But to mention one complexity:  pandoc can take multiple input files on the command line.  It can also parse from stdin, in which case there are no input files.  Both cases would have to be handled as well.

rohieb:
Right! I just used `\jobname` in LaTeX with stdin, which got rendered as `input`, pandoc could do the same. For multiple files, the variable could be an array, like e.g. `$classoption$` already is. The additional `$for$` is sensible overhead.

rohieb:
BTW, I tried to do a pull-request myself, but I couldn't figure out where the template variables are set in the code… Any pointers on that? :)

mb21:
> I couldn't figure out where the template variables are set in the code…

@rohieb https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/LaTeX.hs#L190

mb21:
For the template variable to be available in all formats, I think you'll ned to make sure it is set in [this `doc'`](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/App.hs#L376). The `doc'`  should be of type `Pandoc meta blocks`, so specifically it should be in the `meta` of type [Meta](https://github.com/jgm/pandoc-types/blob/master/Text/Pandoc/Definition.hs#L103). But I'm not sure where the proper place to set it is...

rohieb:
Thanks for your initial pointer :) I didn't want to patch every single
writer, but I was able to work my way up to where the metadata comes
from :) See pull-request #3439 for what I currently have.

Since this is my first time using Haskell in a project, the last two
days were spent for a lot of reading documentation and understanding the
code, and there will probably still be a couple days to get it working…
;-) I'll raise my voice when I think the pull-request is ready.

 - Roland

jgm:
[...]

PR #3439, opened 2017-02-12
add template variables for source and destination filename

rohieb:
see #3431

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1074,6 +1074,22 @@ Some variables are set automatically by pandoc.  These vary somewhat
 depending on the output format, but include metadata fields as well
 as the following:
 
+`sourcefile`, `outputfile`
+:   source and destination filenames, as given on the command line.
+    `sourcefile` can also be a list if input comes from multiple files, or empty
+    if input is from stdin. You can use the following snippet in your template
+    to distinguish them:
+
+        $if(sourcefile)$
+        $for(sourcefile)$
+        $sourcefile$
+        $endfor$
+        $else$
+        (stdin)
+        $endif$
+
+    Similarly, `outputfile` can be `-` if output goes to the terminal.
+
 `title`, `author`, `date`
 :   allow identification of basic aspects of the document.  Included
     in PDF metadata through LaTeX and ConTeXt.  These can be set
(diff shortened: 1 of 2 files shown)


jgm:
I'd prefer:

- `sourcefile`, a list of the input file names (empty if stdin)
- `outputfile`, the output file name

If you want to have it print `(stdin)`, this can be done with a conditional in the template.

```
$if(sourcefile)$
$for(sourcefile)$
$sourcefile$
$endfor$
$else$
(stdin)
$endif$
```

What do you think?

rohieb:
Yes, this makes the code simpler, and simplicity is always a good idea. I've allowed myself to put your template snippet into the manual as an example how to handle the variables.

Looking at the tests, I find it a bit annoying that conversion to markdown now writes the sourcefile and outputfile into the Markdown output – I really only had intended them to be used in user-defined templates. What do you think, does that bother you?

rohieb:
(see [Travis log, L705](https://travis-ci.org/jgm/pandoc/jobs/205154181#L705))

jgm:
+++ Roland Hieber [Feb 24 17 18:34 ]:
>   (see [1]Travis log, L705)

This is precisely why you need to add them as variables,
not as metadata.

rohieb:
On 25.02.2017 12:50, John MacFarlane wrote:
> +++ Roland Hieber [Feb 24 17 18:34 ]:
>> (see [1]Travis log, L705)
> 
> This is precisely why you need to add them as variables,
> not as metadata.

Wait, I'm confused. In [1] you said I should add them to
writerVariables, and I think that's what I did here. Is there some place
else where I should add them?

[1] https://github.com/jgm/pandoc/issues/3431#issuecomment-279236546

jgm:
No, you did the right thing. I jumped to conclusions from the travis output, but in fact this seems to point to a bug somewhere else... I'll investigate.

jgm:
OK, this should be fixed now.

jgm:
I fixed a few more things with subsequent commits.  Glad you unearthed that bug; the Markdown writer was leaking variables in metadata blocks.

rohieb:
Happy to help :smile: But more importantly, happy for your help and patience, @jgm and @mb21! This was a great opportunity for me to get to know Haskell and pandoc more :smiley:

jgm:
We're always on the lookout for new contributors!

+++ Roland Hieber [Feb 25 17 18:05 ]:
>   Happy to help 😄 But more importantly, happy for your help and patience,
>   [1]@jgm and [2]@mb21! This was a great opportunity for me to get to
>   know Haskell and pandoc more 😃

PR #3446, opened 2017-02-15
ZimWiki writer: added source code naming, removed internal formatting from note and table cells

alexivkin:
ZimWiki writer: removed internal formatting from note and table cells because ZimWiki does not support it. Changed source code type mapping to match the one used by the ZimWiki's sourcecode plugin.


jgm:
Thanks!

Issue #3462, opened 2017-02-21
MD-to-ODT-conversion: SVG images not supported properly


nicodecker:
# Problem

A simple markdown document referencing an SVG image can be
converted to ODT. However, the following warnings appear:

```
[pandoc.exe warning] Could not determine image size in `test1.svg': could not determine image type
```

The ODT file is created and the image appears, but the dimensions of the image
are wrong: width and height are always equal (as it can be seen below)

![output](https://cloud.githubusercontent.com/assets/10983264/23175827/df35d286-f860-11e6-9fa0-ab2594f7c8ea.jpg)

# Call

pandoc was called like this:

```
pandoc -f markdown -t odt -o test.odt test.md
```

# Version Information

```
C:\>pandoc --version
pandoc.exe 1.19.2.1
Compiled with pandoc-types 1.17.0.4, texmath 0.9, skylighting 0.1.1.4
Default user data directory: C:\Users\nico\AppData\Roaming\pandoc
Copyright (C) 2006-2016 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
```

# Resources

* Markdown is also very simple and short.

```md
Test
=====

A small test image.

![Test1](test1.svg)

End of markdown file.
```

* The SVG is a very simple test file containing two paths and some text.
  This is an inkscape file, however, the pure SVG version did not work either.

![test1](https://cloud.githubusercontent.com/assets/10983264/23175863/09799898-f861-11e6-9a32-41b73ed850b0.png)

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->

<svg
   xmlns:dc="http://purl.org/dc/elements/1.1/"
   xmlns:cc="http://creativecommons.org/ns#"
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
   xmlns:svg="http://www.w3.org/2000/svg"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:xlink="http://www.w3.org/1999/xlink"
   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
   width="353.07678"
   height="214.60573"
   id="svg2"
   version="1.1"
   inkscape:version="0.91 r13725"
[...]

PR #3464, opened 2017-02-22
make imageSize recognize basic SVG dimensions, see #3462

mb21:
This only looks at the `width` and `height` attributes of the SVG root element which are not mandatory. So it ignores e.g. the `viewBox` attribute, see https://www.w3.org/TR/SVG/coords.html#ViewportSpace and https://css-tricks.com/scale-svg/

Partly resolves https://github.com/jgm/pandoc/issues/3462

### DIFF
--- a/src/Text/Pandoc/ImageSize.hs
+++ b/src/Text/Pandoc/ImageSize.hs
@@ -39,6 +39,7 @@ module Text.Pandoc.ImageSize ( ImageType(..)
                              , Direction(..)
                              , dimension
                              , inInch
+                             , inPixel
                              , inPoints
                              , numUnit
                              , showInInch
@@ -58,6 +59,8 @@ import Data.Default (Default)
 import Numeric (showFFloat)
 import Text.Pandoc.Definition
 import Text.Pandoc.Options
+import qualified Text.Pandoc.UTF8 as UTF8
+import qualified Text.XML.Light as Xml
 import qualified Data.Map as M
 import Control.Monad.Except
 import Data.Maybe (fromMaybe)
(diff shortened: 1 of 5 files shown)


src/Text/Pandoc/ImageSize.hs
```diff
@@ -100,6 +102,8 @@ imageType img = case B.take 4 img of
                      "\xff\xd8\xff\xe0" -> return Jpeg  -- JFIF
                      "\xff\xd8\xff\xe1" -> return Jpeg  -- Exif
                      "%PDF"             -> return Pdf
+                     "<?xm"             -> return Svg -- this is more of a guess
```
jgm: In this case couldn't we have a further test to see if `<svg` is an infix of the bytestring?  Or scan to the next `<` and verify that it's followed by `svg`?

src/Text/Pandoc/ImageSize.hs
```diff
@@ -271,6 +280,23 @@ gifSize img = do
                           }
        _             -> Nothing -- "GIF parse error"
 
+svgSize :: ByteString -> Maybe ImageSize
+svgSize img = do
+  doc <- Xml.parseXMLDoc $ UTF8.toString img
+  let opts = def --TODO: use proper opts instead of def, which simply contains dpi=72
```
jgm: We might as well do it right, even if it requires an interface change.
2.0 is a great opportunity to break things that need breaking.

jgm:
I've made a couple comments, but if you want to go ahead and merge this now, and refine it later, that's fine too.

mb21:
Sure, I'll take a look now...

mb21:
Looking for the first `<` seems like a good idea, thanks for the suggestion.

rebased and added a commit with the interface change.

PR #3468, opened 2017-02-24
Use SVG instead of PNG when using webtex.

doronbehar:
Codecogs supports svg and not only png. In my opinion it's a much better choice and I think `pandoc` should use it as default.

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -1221,7 +1221,7 @@ options =
     , Option "" ["webtex"]
                  (OptArg
                   (\arg opt -> do
-                      let url' = fromMaybe "https://latex.codecogs.com/png.latex?" arg
+                      let url' = fromMaybe "https://latex.codecogs.com/svg.latex?" arg
                       return opt { optHTMLMathMethod = WebTeX url' })
                   "URL")
                  "" -- "Use web service for HTML math"


jgm:
Perhaps @ickc, who suggested CodeCogs originally, would like to comment.

The main worry I see is that svg is not supported in as many output formats, and perhaps not well supported in all browsers and ebook reeaders.  So png seems like a better default.  Of course, it's easy enough to use svg by supplying your own URL as argument to `--webtex`.  I'm willing to be persuaded otherwise, though.

mb21:
SVG support in browsers is [pretty wide-spread](http://caniuse.com/#search=svg) nowadays (IE >= 11), in eBook readers it's more of a mixed bag as far as I know. But maybe it's time to switich the default.

ickc:
png is ubiquitous, but even if svg is quite widely supported now, it is still not ubiquitous. e.g. epub (which shouldn't be s concern because epub shouldn't load external resource anyway) and email. Sadly, many email client still didn't support svg and may even strip it away from the email.

Since pandoc has provided an abundance of choice on Math output, and webtex is argurably the "lowest common denominator" of all (mathematically speaking, it should be highest common factor), keeping the default as png seems to fit the purpose of being highest common factor. Especially considering one can always override the url, the default should be one that's most widely supported and worry-free for beginners.

But even so, such possibility should be made more obvious to beginners, so perhaps the manual can mention if one want to have svg instead, use a certain url instead.

This discussion sparks a related possibility: since svg can naturally be inline element (and small), it can easily be embeded (unlike png requires data uri to be self contained). So may be another option can be given, once activated, pandoc will do IO (made possible by pandoc 2.0) and download the svg from codecogs and inline them in the html output. Similar suggestion (inline svg equation) and effort has been made in @jgm's pandoc-tex2svg as a filter, but because of the external dependency on mathjax-node, it is not without its problem. So may be this is a simple, effective way to implement self-contained svg equations. (a note on quality: since codecogs' is just an svg image, its "typesetting", e.g. the relative position to the lines, will not be as good as MathJax's and MathJax-node's because they aim for typeset-correctness at the very beginning.)

jgm:
+++ ickc [Feb 25 17 02:39 ]:
> e.g. epub (which shouldn't be s concern because
>   epub shouldn't load external resource anyway)

Actually, lots of people use `--webtex` with
`--self-contained` in generating epub.  There aren't a lot
of good alternatives for portable epub with math.

>   But even so, such possibility should be made more obvious to beginners,
>   so perhaps the manual can mention if one want to have svg instead, use
>   a certain url instead.

Yes, good idea to list the required command in the manual.

>   This discussion sparks a related possibility: since svg can naturally
>   be inline element (and small), it can easily be embeded (unlike png
>   requires data uri to be self contained). So may be another option can
>   be given, once activated, pandoc will do IO (made possible by pandoc
>   2.0) and download the svg from codecogs and inline them in the html
>   output.

Or `--self-contained` could be taught to do this when it
encounters a link to an external svg resource.

ickc:
> Actually, lots of people use `--webtex` with
`--self-contained` in generating epub.  There aren't a lot
of good alternatives for portable epub with math.

Right! Missed this combination. Then defaulting at png is still the safest bet.

Another tip is this: often time the need of svg is for retina-display, this can be done with png by requesting the image from codecogs at 2 times the resolution, and display it at 50% (there's another more sophisticated method). I think currently it cannot be done in vanilla pandoc because of the need of 50%.

mb21:
Okay, so maybe it's actually better to leave PNG as the default and just update the manual. See https://github.com/jgm/pandoc/pull/3471

PR #3469, opened 2017-02-24
Fix stale references to `tests` directory

oneeman:
I noticed failures running the benchmarks:

```
pandoc-2.0: benchmarks
Running 1 benchmarks...
Benchmark benchmark-pandoc: RUNNING...
benchmark-pandoc: tests/testsuite.txt: openFile: does not exist (No such file or directory)
Benchmark benchmark-pandoc: ERROR
Completed 2 action(s).

--  While building package pandoc-2.0 using:
      /home/or/programming/haskell/pandoc/.stack-work/dist/x86_64-linux/Cabal-1.24.2.0/setup/setup --builddir=.stack-work/dist/x86_64-linux/Cabal-1.24.2.0 bench benchmark-pandoc
    Process exited with code: ExitFailure 1
```

So I looked and saw stale references to a directory `tests` which should instead be `test`.  Updating the references fixed the benchmark failures.  I then searched through the project to see if there were other references to it, and found and updated a few.

Anyway, this is my first PR to pandoc, so if I'm running foul of any guidelines (e.g. if I should have first created an issue about the benchmark failure and then referenced it in the PR, or whatever) please let me know.

Cheers!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -161,7 +161,7 @@ or, if you're using [stack],
     stack setup
     stack test
 
-The test program is `tests/test-pandoc.hs`.
+The test program is `test/test-pandoc.hs`.
 
 Benchmarks
 ----------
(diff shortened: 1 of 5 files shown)


jgm:
Thanks very much for this!

PR #3480, opened 2017-03-01
Writers: Use gets to access MonadState where possible

ilabdsf:
Replaced  `liftM`, `fmap` and recurring `get >>= return` with dedicated `gets`


jgm:
Great, thanks!
(I think I didn't know about gets when I wrote some of these writers.)

PR #3485, opened 2017-03-04
templates: CSS for .smallcaps, closes #1592

### LINKED ISSUE #1592: parse small caps from HTML as a class

mb21:


jgm:
Thanks for doing this!  I noticed we don't have
```
code{white-space: pre;} 
```
in the epub templates.  Is there any reason for that?

mb21:
I wondered the same thing... don't know, maybe it just was forgotten in https://github.com/jgm/pandoc-templates/commit/9601cc502644458c7ddf71394f8e4be74c464420 ?

Maybe we should also have `pre-wrap;` in all slide show templates, not only [in dzslides](https://github.com/jgm/pandoc-templates/commit/bd5818c52a0fd5f77dceab04f9a208b4100ab2f4)?

Issue #1667, opened 2014-10-05
YAML lang attribute not working in ODT


ousia:
From #1614, i think that `lang` attribute should set document language in (at least) ConTeXt, `.docx` and `.odt`.

Sample file:

```

---
title: Titel
lang: de
...

# Kapitel

Mein Text
```

At least with version 1.12.3.3, language attribute isn’t set for ConTeXt, ODF or OOXML documents.

I think this should be improved.

Many thanks for your excellent work.

ousia:
@jgm, @mpickering, just in case it might help, I attach a list that contains the equivalences between ISO-639 language codes used in HTML and ConTeXt language codes (they differ from LaTeX ones). 

```
af -> af
af-ZA -> af
ar -> ar
ar-DZ -> ar-dz
ar-IQ -> ar-iq
ar-JO -> ar-jo
ar-LB -> ar-lb
ar-MA -> ar-ma
ar-SY -> ar-sy
bg -> bg
bg-BG -> bg
ca -> ca
ca-ES -> ca
cy -> cy
cy-UK -> cy
cz -> cz
cz-CZ -> cz
da -> da
da-DK -> da
de -> de
de-1901 -> deo
de-AT -> de-at
de-CH -> de-ch
de-DE -> de-de
el -> gr
en -> en
en-UK -> uk
en-US -> en
es -> es
es-ES -> es
et -> et
et-EE -> et
eu -> eu
eu-ES -> eu
fi -> fi
fi-FI -> fi
fr -> fr
fr-CA -> fr
fr-FR -> fr
grc -> agr
he -> il
he-IL -> il
hr -> hr
hr-HR -> hr
hu -> hu
hu-HU -> hu
is -> is
is-IS -> is
it -> it
it-IT -> it
jp -> ja
jp-JP -> ja
la -> la
nb -> nb
nb-NO -> nb
nl -> nl
nl-NL -> nl
nn -> nn
nn-NO -> nn
no -> no
no-NO -> no
pl -> pl
pl-PL -> pl
pt -> pt
pt-BR -> pt
pt-PT -> pt
ro -> ro
ro-RO -> ro
ru -> ru
ru-RU -> ru
sk -> sk
sk-SK -> sk
sl -> sl
sl-SL -> sl
sv -> sv
sv-SE -> sv
tr -> tr
tr-TR -> tr
uk -> ua
uk-UA -> ua
vi -> vn
vi-VN -> vn
```

ousia:
[...]

PR #3515, opened 2017-03-19
docx writer: lang meta, see #1667

mb21:
see https://github.com/jgm/pandoc/issues/1667


src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -256,8 +257,29 @@ writeDocx opts doc@(Pandoc meta _) = do
                        )
 
   -- styles
+  let lang = case lookupMeta "lang" meta of
+               Just (MetaInlines [Str s]) -> Just s
```
jgm: Let's also match on `MetaString s` -- if they use the `--metadata` flag on the command line, we'll have a `MetaString`.

mb21: I see, done.

jgm:
Great, thanks!

PR #3535, opened 2017-03-28
lstinline with braces can be used (verb cannot be used with braces)

### LINKED ISSUE #3534: Lstinline with Braces Does Not Work
Following LaTeX code produces a PDF containing highlighted inline Java code. 

```latex
\documentclass{article}
\usepackage[utf8]{inputenc}
\usepackage{listings}

\begin{document}

I want to explain the interface of \lstinline[language=Java]{public class MyClass}.

\end{document}
```
pandoc 1.19.2.1 complain about the LaTeX file:

```
pandoc: 
Error at "input" (line 7, column 84):
unexpected "\n"
expecting "{"
I want to explain the interface of \lstinline[language=Java]{public class MyClass}.
                                                                                   ^
CallStack (from HasCallStack):
  error, called at src/Text/Pandoc/Error.hs:66:13 in pandoc-1.19.2.1-5WoNbO1LsZw5GK0galu1VY:Text.Pandoc.Error
```

schrieveslaach:
I want to resolve #3534

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -601,8 +601,7 @@ inlineCommands = M.fromList $
   , ("thanks", note <$> grouped block)
   , ("footnote", note <$> grouped block)
   , ("verb", doverb)
-  , ("lstinline", do options <- option [] keyvals
-                     dolstinline options)
+  , ("lstinline", dolstinline)
   , ("Verb", doverb)
   , ("url", (unescapeURL <$> braced) >>= \url ->
        pure (link url "" (str url)))
@@ -717,13 +716,12 @@ doverb = do
   marker <- anyChar
   code <$> manyTill (satisfy (/='\n')) (char marker)
 
-dolstinline :: PandocMonad m => [(String, String)] -> LP m Inlines
-dolstinline opts = do
+dolstinline :: PandocMonad m => LP m Inlines
+dolstinline = do
+  options <- option [] keyvals
+  let classes = maybeToList $ lookup "language" options >>= fromListingsLanguage
   marker <- char '{' <|> anyChar 
   codeWith ("",classes,[]) <$> manyTill (satisfy (/='\n')) (char '}' <|> char marker)
-               where classes = case lookup "language" opts >>= fromListingsLanguage of
-                               Just l  -> [l]
-                               Nothing -> []
 
 doLHSverb :: PandocMonad m => LP m Inlines
 doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '|')
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -601,7 +601,8 @@ inlineCommands = M.fromList $
   , ("thanks", note <$> grouped block)
   , ("footnote", note <$> grouped block)
   , ("verb", doverb)
-  , ("lstinline", skipopts *> dolstinline)
+  , ("lstinline", do options <- option [] keyvals
+                     dolstinline options)
   , ("Verb", doverb)
   , ("url", (unescapeURL <$> braced) >>= \url ->
        pure (link url "" (str url)))
@@ -716,10 +717,13 @@ doverb = do
   marker <- anyChar
   code <$> manyTill (satisfy (/='\n')) (char marker)
 
-dolstinline :: PandocMonad m => LP m Inlines
-dolstinline = do
+dolstinline :: PandocMonad m => [(String, String)] -> LP m Inlines
+dolstinline opts = do
   marker <- char '{' <|> anyChar 
-  code <$> manyTill (satisfy (/='\n')) (char '}' <|> char marker)
+  codeWith ("",classes,[]) <$> manyTill (satisfy (/='\n')) (char '}' <|> char marker)
+               where classes = case lookup "language" opts >>= fromListingsLanguage of
+                               Just l  -> [l]
+                               Nothing -> []
 
 doLHSverb :: PandocMonad m => LP m Inlines
 doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '|')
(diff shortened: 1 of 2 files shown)


jgm:
Great!  I think we can make this better, though.

Some thoughts:

* With the braced form, there can be escaped braces, right?
* We should handle the language attribute.  To do this, you can parse options with `option [] keyvals` -- this will give you a `[(String, String)]` association list.  Then, to construct the sort of attributes pandoc uses for language, and convert from listings language names to pandoc names:
```
  return $ codeWith ("",classes,[])
    where classes = case lookup "language" opts >>= fromListingsLanguage of
                       Just l  -> [l]
                       Nothing -> []
```

schrieveslaach:
> With the braced form, there can be escaped braces, right?

I think it is not supported. In this case, you would just use different characters than braces. 

> We should handle the language attribute. To do this, you can parse options with option [] keyvals -- this will give you a [(String, String)] association list. Then, to construct the sort of attributes pandoc uses for language, and convert from listings language names to pandoc names...

I will give it a try.

jgm:
I'm happy to help if you get stuck.

+++ schrieveslaach [Mar 28 17 05:02 ]:
>   I will give it a try.

schrieveslaach:
@jgm, I need some help. My changes won't compile:

```diff
diff --git a/src/Text/Pandoc/Readers/LaTeX.hs b/src/Text/Pandoc/Readers/LaTeX.hs
index d632eb2d..142f2c04 100644
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -601,7 +601,8 @@ inlineCommands = M.fromList $
   , ("thanks", note <$> grouped block)
   , ("footnote", note <$> grouped block)
   , ("verb", doverb)
-  , ("lstinline", skipopts *> dolstinline)
+  , ("lstinline", do options <- option [] keyvals
+                     dolstinline options)
   , ("Verb", doverb)
   , ("url", (unescapeURL <$> braced) >>= \url ->
        pure (link url "" (str url)))
@@ -716,10 +717,14 @@ doverb = do
   marker <- anyChar
   code <$> manyTill (satisfy (/='\n')) (char marker)
 
-dolstinline :: PandocMonad m => LP m Inlines
-dolstinline = do
+dolstinline :: PandocMonad m => [(String, String)] -> LP m Inlines
+dolstinline opts = do
   marker <- char '{' <|> anyChar 
   code <$> manyTill (satisfy (/='\n')) (char '}' <|> char marker)
+  return $ codeWith ("",classes,[]) code
+           where classes = case lookup "language" opts >>= fromListingsLanguage of
+                               Just l  -> [l]
+                               Nothing -> []
 
 doLHSverb :: PandocMonad m => LP m Inlines
 doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '|')
```

Here's the compiler error:

```
src/Text/Pandoc/Readers/LaTeX.hs:724:37: error:
    • Couldn't match type ‘String -> Inlines’ with ‘[Char]’
      Expected type: String
        Actual type: String -> Inlines
    • Probable cause: ‘code’ is applied to too few arguments
      In the second argument of ‘codeWith’, namely ‘code’
      In the second argument of ‘($)’, namely
        ‘codeWith ("", classes, []) code’
      In a stmt of a 'do' block: return $ codeWith ("", classes, []) code
```

I'm pretty sure that I'm using your code snippet wrong.

schrieveslaach:
@jgm, nevermind. I've found the solution.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -601,7 +601,8 @@ inlineCommands = M.fromList $
   , ("thanks", note <$> grouped block)
   , ("footnote", note <$> grouped block)
   , ("verb", doverb)
-  , ("lstinline", skipopts *> dolstinline)
+  , ("lstinline", do options <- option [] keyvals
```
jgm: To make this cleaner, move `options <- option [] keyvals` down into `dolstinline` and remove the `opts` parameter there.

schrieveslaach: I tried to move, but I get an compiler error `Variable not in scope: options :: [([Char], String)]`

```diff
diff --git a/src/Text/Pandoc/Readers/LaTeX.hs b/src/Text/Pandoc/Readers/LaTeX.hs
index 2e40e8eb..11d64340 100644
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -601,8 +601,7 @@ inlineCommands = M.fromList $
   , ("thanks", note <$> grouped block)
   , ("footnote", note <$> grouped block)
   , ("verb", doverb)
-  , ("lstinline", do options <- option [] keyvals
-                     dolstinline options)
+  , ("lstinline", dolstinline)
   , ("Verb", doverb)
   , ("url", (unescapeURL <$> braced) >>= \url ->
        pure (link url "" (str url)))
@@ -717,11 +716,12 @@ doverb = do
   marker <- anyChar
   code <$> manyTill (satisfy (/='\n')) (char marker)
 
-dolstinline :: PandocMonad m => [(String, String)] -> LP m Inlines
-dolstinline opts = do
+dolstinline :: PandocMonad m => LP m Inlines
+dolstinline = do
+  options <- option [] keyvals
   marker <- char '{' <|> anyChar 
   codeWith ("",classes,[]) <$> manyTill (satisfy (/='\n')) (char '}' <|> char marker)
-               where classes = case lookup "language" opts >>= fromListingsLanguage of
+               where classes = case lookup "language" options >>= fromListingsLanguage of
                                Just l  -> [l]
                                Nothing -> []

```

Other function using `options <- option [] keyvals` have `LP m Blocks` in there signature. Does this matter?

jgm: Yes, there are some unexpected things about scope with `where`.

Try this instead:

```haskell
  options <- option [] keyvals
  let classes = maybeToList $ lookup "language" options >>= fromListingsLanguage
  marker <- char '{' <|> anyChar
  codeWith ...as before
```

schrieveslaach: Thanks! 🥇  Done and PR has been updated.

jgm:
Great, thanks!

Issue #3115, opened 2016-09-13
Combining multiple metadata blocks


gfelbing:
Hello,

I am using a shared metadata yaml for all my documents, which contains definitions, which I need in all of my documents, for example:

```

---
documentclass: article
header-includes:
  - \usepackage{tikz}

---
```

Now, I have some document, in which I want to overwrite/extend some values:

```

---
documentclass: IEEEtran
header-includes:
  - \IEEEspecialpapernotice{Some notice}

---
```

I would have expected, that the documentclass (since it is a single-value type) gets overridden and the header-includes (since it is an array) gets extended.
Instead, the values specified in my document got completely ignored.

I compiled the document with the following options:

```
pandoc --standalone --filter pandoc-citeproc -f markdown+yaml_metadata_block+example_lists+implicit_figures+citations -o ./subject/document.pdf ./pandoc/meta.yaml  ./subject/document.md
```

Is there a way to achieve my expected behaviour?

jgm:
When a value is set multiple times in the same document, the
earlier one takes priority.  See the MANUAL:

> A document may contain multiple metadata blocks.  The
> metadata fields will be combined through a *left-biased
> union*:  if two metadata blocks attempt to set the same
> field, the value from the first block will be taken.

Technical explanation for pandoc developers: This is because
the monoid instance for Meta (in Text.Pandoc.Definition)
uses Data.Map.union, which is left-biased.  (Monoid append
is used for each new metadata block in the document; see the
end of yamlMetaBlock in the Markdown reader.)

So the solution for you would be to include the common
metadata after your document rather than before.

gfelbing:
This would work for now.
For List-Typed metadata, you could apply `Data.Map.unionWith` curried with a (++), so arrays would get extendend.

tarleb:
Here is an hacked-together prove of concept of how this could be done using lua filters (requires #3584).  Would this be sufficient to close this issue?

``` lua
--- Define ways in which metadata can be merged.
local merge_methods = {
  replace = function(v1, v2) return v2 or v1 end,
  keep = function(v1, v2) return v1 or v2 end,
  extendlist = function(v1, v2)
    local res
    if type(v1) == "table" and v1.tag == "MetaList" then
      res = v1
    else
      res = pandoc.MetaList{v1}
    end
    if type(v2) == "table" and v2.tag == "MetaList" then
      for i = 1, #v2 do
        res[#res + 1] = v2[i]
      end
    else
[...]

PR #3550, opened 2017-04-02
Lua module: add readers submodule

tarleb:
Plain text readers are exposed to lua scripts via the `pandoc.reader`
submodule, which is further subdivided by format.  Converting e.g. a
markdown string into a pandoc document is possible from within lua:

    doc = pandoc.reader.markdown.read_doc("Hello, World!")

A `read_block` convenience function is provided for all formats,
although it will still parse the whole string but return only the first
block as the result.

Custom reader options are not supported yet, default options are used
for all parsing operations.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -247,9 +247,10 @@ Extra-Source-Files:
                  test/odt/odt/*.odt
                  test/odt/markdown/*.md
                  test/odt/native/*.native
-                 test/lua/strmacro.lua
-                 test/lua/plain-to-para.lua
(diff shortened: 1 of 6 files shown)


jgm:
Can you say a bit about the intended use of this?  An example that motivated this addition?

tarleb:
I had two use-cases in mind: reading and handling additional metadata (#1960, #3115) and creation of custom *include* mechanisms for formats that normally don't support it.

Long term, it could be possible to write conversion scripts, allowing for fully portable programs without much trouble.

PR #3553, opened 2017-04-03
Include \VerbatimFootnotes for highlighted code blocks

nunull:
Updated the LaTeX writer to also include \VerbatimFootnotes in the
preamble for highlighted code blocks. Previously this was only done for
raw code blocks.

This is my first attempt in contributing to pandoc, so my apologies for any mistakes I might have made.


jgm:
Looks great, thanks!

PR #3554, opened 2017-04-03
Add original classes to JS obfuscated links

### LINKED ISSUE #2989: email obfuscation does not properly rebuild the anchor tag
When the default e-mail objuscation rewrites anchor tags, it doesn't seem to copy attributes from the original anchor. So if I had an anchor like:

```
<a class="button blue size-32" href="somewhere">lonely swedish</a>
```

It seems that Pandoc rewrites it to a javascript that inserts:

```
    <a class="email" href="somewhere">lonely swedish</a>
```

And my extra class names are lost.

nunull:
HTML links containing classes originally now preserve them when using javascript email obfuscation.

Fixes #2989

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -476,8 +476,7 @@ obfuscateLink opts attr (renderHtml -> txt) s =
                     else ("'" ++ obfuscateString txt ++ "'",
                           txt ++ " (" ++ name' ++ " at " ++ domain' ++ ")")
               (_, classNames, _) = attr
-              classNamesStr = (if not $ null classNames then " " else "")
-                           ++ intercalate " " classNames
+              classNamesStr = concatMap (' ':) classNames
           in  case meth of
                 ReferenceObfuscation ->
                      -- need to use preEscapedString or &'s are escaped to &amp; in URL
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -475,6 +475,9 @@ obfuscateLink opts attr (renderHtml -> txt) s =
                     then ("e", name' ++ " at " ++ domain')
                     else ("'" ++ obfuscateString txt ++ "'",
                           txt ++ " (" ++ name' ++ " at " ++ domain' ++ ")")
+              (_, classNames, _) = attr
+              classNamesStr = (if not $ null classNames then " " else "")
+                           ++ intercalate " " classNames
```
jgm: Perhaps more economical?
```
classNamesStr = concatMap (' ':) classNames
```
I think that will do the same, without intercalate and the conditional.

Also, a stylistic thing, which may just be a quirk of my own:  I find the `$` in

    if not $ null classnames then " " else ""

to be a bit ugly -- usually I only use the `$` when it is clear to the eye where its scope ends (parentheses or end ofline).  Here the scope ends before `then`, but there's no visual marker for that.  So I'd tend to use `(null classNames)` instead.  Of course what you have is perfectly correct, this is just a stylistic thing.

nunull: Agreed, this is a nice solution I didn't realize was possible.

jgm:
Great!

PR #3556, opened 2017-04-03
Add class to footnote back references

### LINKED ISSUE #1831: footnote back references should have a class in HTML
If pandoc parses the following source:

```
a^[footnote]
```

as 

``` html
<p>a<a href="#fn1" class="footnoteRef" id="fnref1"><sup>1</sup></a></p>
<div class="footnotes">
<hr />
<ol>
<li id="fn1"><p>footnote<a href="#fnref1">↩</a></p></li>
</ol>
</div>
```

There should be a `class="footnoteBack">` for `<a href="#fnref1">↩</a>`. So it would be easier to style it with CSS.

One may want to remove underline from footnote references and footnote back links and not from standard links (so that the latter could be recognized as such).

nunull:
The HTML writer now also adds the class `footnoteBack` to back references
of footnotes. This allows for easier CSS styling.

Fixes #1831

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -1047,7 +1047,7 @@ blockListToNote :: PandocMonad m => WriterOptions -> String -> [Block] -> StateT
 blockListToNote opts ref blocks =
   -- If last block is Para or Plain, include the backlink at the end of
   -- that block. Otherwise, insert a new Plain block with the backlink.
-  let backlink = [Link nullAttr [Str "↩"] ("#" ++ writerIdentifierPrefix opts ++ "fnref" ++ ref,[])]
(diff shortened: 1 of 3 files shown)


jgm:
Thanks!

mb21:
Perhaps we should standardize HTML class names to either camelCase (`footnoteBack`), dash-case (we already have e.g. `comment-start`, `line-block`) or nospacecase (we already have e.g. `smallcaps`, `titleslide`).

jgm:
Not a bad idea, though I'm a bit reluctant to break CSS for
existing documents.

Which of these options would be best, do you think?

+++ Mauro Bieg [Oct 28 17 13:20 ]:

mb21:
I would probably opt for dash-case. It's the most common in HTML/CSS style guides like BEM (component-names there are even required to have a dash in them to distinguish them from nested identifiers that have only meaning inside those components). Some people also reserve camelCase for id/class names that are only there to be used from JavaScript.

jgm:
Want to submit a PR?

I wouldn't feel bad about changing these one day after the
release.

+++ Mauro Bieg [Oct 30 17 08:49 ]:

PR #3569, opened 2017-04-12
Destructuring lua filter functions

tarleb:
Lua filter functions and `Inline` constructors in the `pandoc.lua` module are changed to take the same arguments as those lua functions used for custom writers.  The result is a uniform interface for the lua subsystem.

Example for filtering on `Span`:

``` lua
Span = function(lst, attr)
  return pandoc.Span(lst, attr)
end
```

Performance is improved slightly and lua documentation can be generated via `ldoc`.  The only downside is that `pandoc.lua` becomes larger and less simple.

### DIFF
--- a/data/pandoc.lua
+++ b/data/pandoc.lua
@@ -173,18 +173,23 @@ end
 --- Meta blocks
 -- @function MetaBlocks
 -- @tparam {Block,...} blocks blocks
+
 --- Meta inlines
 -- @function MetaInlines
 -- @tparam {Inline,...} inlines inlines
+
 --- Meta list
 -- @function MetaList
 -- @tparam {MetaValue,...} meta_values list of meta values
+
 --- Meta boolean
 -- @function MetaBool
 -- @tparam boolean bool boolean value
+
 --- Meta map
 -- @function MetaMap
 -- @tparam table a string-index map of meta values
+
 --- Meta string
 -- @function MetaString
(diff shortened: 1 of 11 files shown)


tarleb:
I'm merging this, as I see mostly advantages in this change. GitHub makes it easy to revert merges in case this turns out to be a bad idea.

jgm:
+++ Albert Krewinkel [Apr 12 17 12:46 ]:
>   This is work in progress, opening a PR to check whether it's worth
>   continuing on this path.
>
>   Lua filter functions and Inline constructors in the pandoc.lua module
>   are changed to take the same arguments as those lua functions used for
>   custom writers. The result is a unified interface for the lua
>   subsystem.
>
>   Example for filtering on Span:
>
>Span = function(lst, attr)
>  return pandoc.Span(lst, attr)
>end
>
>   Performance is improved slightly. The only downside I see is that
>   pandoc.lua becomes larger and less simple.

Possible other downsides to consider:

1.  More difference between lua filters and other filters,
    hence harder to port other filters to lua.

2.  More manual revision of the lua interface is now required
    (or is it?) when pandoc-types changes.

tarleb:
I hadn't considered that, the difference to other other filter implementations is a problem. I will turn that back to the previous behavior and focus on giving elements named accessors instead.

How about supporting `SingleQuoted` etc as filter functions, should we keep that?

PR #3584, opened 2017-04-17
Provide method for metadata processing in lua filters

tarleb:
Allow processing of metadata without having to push the whole document to the lua stack.  Additionally, the `pandoc.reader` lua submodule is removed in favor of a simple `read` function which takes a string and format specifier as arguments.

Additional API change: All functions defined directly in the `Text.Pandoc` module are pushed to `Text.Pandoc.Extensions` or one of the newly introduced `Text.Pandoc.Readers` or `Text.Pandoc.Writers` modules.  This is technically not necessary.

This change is intended as a solution to #1960.  The following uses the metadata from `metadata-file.yaml`.

``` lua
return {
  {
    Meta = function(meta)
      local metafile = io.open('metadata-file.yaml', 'r')
      local content = metafile:read("*a")
      metafile:close()
      return pandoc.read(content, "markdowns").meta
    end,
  }
}
```


jgm:
These look fine to me, so go ahead and merge them.

By the way, at some point we'll need a tutorial on how to use the lua filters.

PR #3588, opened 2017-04-18
Add siunitx Support

### LINKED ISSUE #3587: LaTeX: Support `siunitx` Commands
LaTeX provides the package `siunitx` to typeset quantities in a consistent way. Pandoc should provide basic support for this package. For example, following code should be translated:

```latex
\SI[round-precision=2]{1}{m} is equal to \SI{1000}{mm}.

\SI[round-precision=2]{1}[\$]{} is equal to \SI{0.938094}{\euro}
```

Pandoc should parse this input example into following AST:

```haskell
[Para [Str "1m",Space,Str "is",Space,Str "equal",Space,Str "to",Space,Str "1000mm"]
,Para [Str "$1",Space,Str "is",Space,Str "equal",Space,Str "to",Space,Str "0.938094\8364"]]
```

schrieveslaach:
I want to resolve #3587.

Basic command parsing works, but preunit, value, and unit do not get parsed further (e.g. `\euro` is not translated into `€`). I assume I have to use the function `tok`.  @jgm, can you help me to use it correctly?

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -728,16 +728,19 @@ dolstinline = do
 doLHSverb :: PandocMonad m => LP m Inlines
 doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '|')
 
+-- converts e.g. \SI{1}[\$]{} to "$ 1" or \SI{1}{\euro} to "1 €"
 dosiunitx :: PandocMonad m => LP m Inlines
 dosiunitx = do
   skipopts
   value <- char '{' >> (mconcat <$> manyTill inline (char '}'))
-  preunit <- option "" $ char '[' >> (mconcat <$> manyTill inline (char ']'))
+  valueprefix <- option "" $ char '[' >> (mconcat <$> manyTill inline (char ']'))
   unit <- char '{' >> (mconcat <$> manyTill inline (char '}'))
-  return . mconcat $ [preunit, 
-                      (if length preunit == 0 then "" else "\160"),
+  let emptyOr160 "" = ""
+      emptyOr160 _  = "\160"
+  return . mconcat $ [valueprefix, 
+                      emptyOr160 valueprefix,
                       value, 
-                      (if length unit == 0 then "" else "\160"),
+                      emptyOr160 unit,
                       unit]
 
 lit :: String -> LP m Inlines
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -734,7 +734,7 @@ dosiunitx = do
   value <- char '{' >> manyTill anyChar (char '}')
   preunit <- (char '[' >> manyTill anyChar (char ']')) <|> (return "")
   unit <- char '{' >> manyTill anyChar (char '}')
-  pure <$> str $ preunit ++ value ++ unit
+  pure <$> str $ preunit ++ (if length preunit == 0 then "" else "\160") ++ value ++ (if length unit == 0 then "" else "\160") ++ unit
 
 lit :: String -> LP m Inlines
 lit = pure . str
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -732,9 +732,9 @@ doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '
 dosiunitx :: PandocMonad m => LP m Inlines
 dosiunitx = do
   skipopts
-  value <- char '{' >> (mconcat <$> manyTill inline (char '}'))
-  valueprefix <- option "" $ char '[' >> (mconcat <$> manyTill inline (char ']'))
-  unit <- char '{' >> (mconcat <$> manyTill inline (char '}'))
+  value <- char '{' >> (mconcat <$> manyTill tok (char '}'))
+  valueprefix <- option "" $ char '[' >> (mconcat <$> manyTill tok (char ']'))
+  unit <- char '{' >> (mconcat <$> manyTill tok (char '}'))
   let emptyOr160 "" = ""
       emptyOr160 _  = "\160"
   return . mconcat $ [valueprefix, 
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -731,10 +731,14 @@ doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '
 dosiunitx :: PandocMonad m => LP m Inlines
 dosiunitx = do
   skipopts
-  value <- char '{' >> manyTill anyChar (char '}')
-  preunit <- (char '[' >> manyTill anyChar (char ']')) <|> (return "")
-  unit <- char '{' >> manyTill anyChar (char '}')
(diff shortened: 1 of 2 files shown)


schrieveslaach:
I've managed to parse e.g. `\$` as argument (see diff below). But, I need to combine `preunit`, `value`, and `unit` into one value. How can I concat multiple `LP m Inlines` into one? 

```diff
diff --git a/src/Text/Pandoc/Readers/LaTeX.hs b/src/Text/Pandoc/Readers/LaTeX.hs
index f2f6cd8e..3cf1016f 100644
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -731,10 +731,11 @@ doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '
 dosiunitx :: PandocMonad m => LP m Inlines
 dosiunitx = do
   skipopts
-  value <- char '{' >> manyTill anyChar (char '}')
-  preunit <- (char '[' >> manyTill anyChar (char ']')) <|> (return "")
-  unit <- char '{' >> manyTill anyChar (char '}')
-  pure <$> str $ preunit ++ (if length preunit == 0 then "" else "\160") ++ value ++ (if length unit == 0 then "" else "\160") ++ unit
+  value <- char '{' >> (mconcat <$> manyTill inline (char '}'))
+  preunit <- (char '[' >> (mconcat <$> manyTill inline (char ']'))) <|> (lit "")
+  unit <- char '{' >> (mconcat <$> manyTill inline (char '}'))
+  return value -- TODO combine preunit, value, and unit
+ 
 
 lit :: String -> LP m Inlines
 lit = pure . str
```

@jgm, @jkr, @tarleb, any suggestions?

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -726,6 +728,14 @@ dolstinline = do
 doLHSverb :: PandocMonad m => LP m Inlines
 doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '|')
 
+dosiunitx :: PandocMonad m => LP m Inlines
+dosiunitx = do
+  skipopts
+  value <- char '{' >> manyTill anyChar (char '}')
+  preunit <- (char '[' >> manyTill anyChar (char ']')) <|> (return "")
```
tarleb: The pattern `expr <|> return x` can be written as `option x expr`.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -726,6 +728,14 @@ dolstinline = do
...
+  unit <- char '{' >> manyTill anyChar (char '}')
+  pure <$> str $ preunit ++ (if length preunit == 0 then "" else "\160") ++ value ++ (if length unit == 0 then "" else "\160") ++ unit
```
tarleb: This is fine, if only a bit too long for the 80 characters limit that's usually used.

An alternative would be

``` haskell
return . str . mconcat $ [preunit, value, unit]  -- plus the optional spaces
```

Getting the full length of the string is a bit wasteful, as we only need to know if it's empty.  `null` can be used instead.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -726,6 +728,14 @@ dolstinline = do
 doLHSverb :: PandocMonad m => LP m Inlines
 doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '|')
 
+dosiunitx :: PandocMonad m => LP m Inlines
+dosiunitx = do
+  skipopts
+  value <- char '{' >> manyTill anyChar (char '}')
```
schrieveslaach: `braced` does not parse latex commands e.g. `\euro`. So I cannot write `\SI{1.5}{\euro}`.

src/Text/Pandoc/Readers/LaTeX.hs
schrieveslaach: I use the option command now.

`bracketed` does not parse latex commands e.g. `\$`. So I cannot write `\SI{1.5}[\$]{}`.

src/Text/Pandoc/Readers/LaTeX.hs
schrieveslaach: Thanks that advice helped me. Basic parsing works now.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -731,10 +731,14 @@ doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '
...
   skipopts
-  value <- char '{' >> manyTill anyChar (char '}')
-  preunit <- (char '[' >> manyTill anyChar (char ']')) <|> (return "")
-  unit <- char '{' >> manyTill anyChar (char '}')
-  pure <$> str $ preunit ++ (if length preunit == 0 then "" else "\160") ++ value ++ (if length unit == 0 then "" else "\160") ++ unit
+  value <- char '{' >> (mconcat <$> manyTill inline (char '}'))
+  preunit <- option "" $ char '[' >> (mconcat <$> manyTill inline (char ']'))
+  unit <- char '{' >> (mconcat <$> manyTill inline (char '}'))
+  return . mconcat $ [preunit, 
+                      (if length preunit == 0 then "" else "\160"),
```
mb21: `length preunit == 0` can be rewritten as `null preunit`

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -726,6 +728,18 @@ dolstinline = do
 doLHSverb :: PandocMonad m => LP m Inlines
 doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '|')
 
+dosiunitx :: PandocMonad m => LP m Inlines
+dosiunitx = do
+  skipopts
+  value <- char '{' >> (mconcat <$> manyTill inline (char '}'))
+  preunit <- option "" $ char '[' >> (mconcat <$> manyTill inline (char ']'))
+  unit <- char '{' >> (mconcat <$> manyTill inline (char '}'))
+  return . mconcat $ [preunit,
```
mb21: how about:

    preunit <- ...
    unit    <- ...
    let emptyOr160 "" = ""
        emptyOr160 _ = "\160"
    return . mconcat $ [preunit, emptyOr160 preunit, value, emptyOr160 unit, unit]

schrieveslaach: Awesome, thanks!

src/Text/Pandoc/Readers/LaTeX.hs
jgm: `braced` gives you a raw string inside the braces.  If you want LaTeX like `\euro` to be parsed, then use `tok`, which parses a LaTeX token (which might be a group like `{\euro 33}` but could also be e.g. a single digit).
This gives you a list of inlines, rather than a string.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -728,16 +728,19 @@ dolstinline = do
 doLHSverb :: PandocMonad m => LP m Inlines
 doLHSverb = codeWith ("",["haskell"],[]) <$> manyTill (satisfy (/='\n')) (char '|')
 
+-- converts e.g. \SI{1}[\$]{} to "$ 1" or \SI{1}{\euro} to "1 €"
 dosiunitx :: PandocMonad m => LP m Inlines
 dosiunitx = do
   skipopts
   value <- char '{' >> (mconcat <$> manyTill inline (char '}'))
```
jgm: Try

    value <- tok

This will handle escapes, etc., better than what you have.

schrieveslaach: Done.

PR #3589, opened 2017-04-18
Add Support for `glossaries` and `acronym` package

### LINKED ISSUE #3539: LaTeX: Read Glossary Entries
In LaTeX you can use the [glossaries package](http://mirror.unl.edu/ctan/macros/latex/contrib/glossaries/glossariesbegin.pdf) providing and generating glossaries entries and using them in the text. 

```latex
\newacronym{API}{API}{Application Programming Interface}

Many programming languages provide \glspl{API}. Each \gls{API} should provide a documentation.
```

When I convert this LaTeX code with pandoc, pandoc does not print the entries. Is there any way that pandoc will convert the LaTeX into following markdown code?

```markdown
Many programming languages provide Application Programming Interfaces. Each API should provide a documentation.
```

schrieveslaach:
I want to resolve #3539.


mb21:
btw, you can also squash commits, (but only do so as long as they are not merged into master yet!) see https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

so you could squash the clean-up and improve-stuffz commits into the previous ones, then `git push --force` to your feature branch. thanks for contributing to pandoc btw!

schrieveslaach:
@mb21, thanks for the advice. But, I just forgot to do so 😆

schrieveslaach:
Now, pandoc puts the acronym mentions into spans. With that you can write a filter which will resolve the acronyms. For example (fhe file `texmf/Acronyms.sty` contains the definitions):

```python
#!/usr/bin/python3

import os
from panflute import *
import re

acronyms = {}

refcounts = {}

def resolveAcronyms(elem, doc):
    if isinstance(elem, Span) and "acronym-label" in elem.attributes:
        label = elem.attributes["acronym-label"]
        
        if label in acronyms:
            # this is the case: "singular" in form and "long" in form:
            value = acronyms[label]
            
            form = elem.attributes["acronym-form"]
            if label in refcounts and "short" in form:
                if "singular" in form:
                    value = label
                else:
                    value = label + "s"
            
            elif "full" in form or "short" in form:
                # remember that label has been used
                if "short" in form:
                    refcounts[label] = True
                
                if "singular" in form:
                    value = value + " (" + label + ")"
                else:
                    value = value + "s (" + label + "s)"
            
            elif "abbrv" in form:
                if "singular" in form:
                    value = label
                else:
                    value = label + "s"
            
            return Span(Str(value))

def loadAcronyms():
    pattern = re.compile(r"\\newacronym(\[.*\])?\{(?P<label>[A-Za-z]+)\}\{([A-Za-z]+)\}\{(?P<value>[A-Za-z 0-9\-]+)\}")
    
    d = os.path.dirname(__file__)
    filename = os.path.join(d, './texmf/Acronyms.sty')
    with open(filename, 'r', encoding='utf-8') as acronymsFile:
        for line in acronymsFile:
            match = pattern.match(line)
            if match:
                acronyms[match.group('label')] = match.group('value')

def main(doc=None):
    loadAcronyms()
    return run_filter(resolveAcronyms, doc=doc)


if __name__ == "__main__":
    main()
```

jgm:
I think we should either

1) change `acronym-label` and `acronym-form` to `data-acronym-label` and `data-acronym-form`, so they are valid HTML5 attribute names, or

2) change the HTML5 writer so it ignores custom attributes not starting with `data-`.

Both solutions would allow the use of filters; solution 1 would also give useful default HTML output without the need for filters.

This general issue comes up a lot; any thoughts, @tarleb, @jkr, @mb21?

tarleb:
I've been following method 1 in the org reader but actually prefer the second option.  How about a third variant in which the HTML writer adds the `data-` prefix for unknown attributes?

mb21:
Method 1 would basically restrict the attribute names in the pandoc document model to valid HTML values, which I don't think we want in 100% of the cases (although probably 90%).

So method 2 makes a bit more sense to me. If you use `data`-attributes you would be explicitly saying: "I want this to be output to HTML". However, you'd still want valid HTML attributes (like `width`, `height`, `datetime`, `download`, `lang`, etc.) to be passed through on elements which they're valid on – which would require us to keep an up to date list of valid attributes in the code :S

> How about a third variant in which the HTML writer adds the data- prefix for unknown attributes?

I wouldn't like this to be the default, since it's quite unexpected. However, if it were only activated after using an option (like `--data-prefix-unknown-attributes`) it might be a good solution.

schrieveslaach:
I assume that I should follow method one for this pull request, correct?

schrieveslaach:
@tarleb, @jgm, @mb21, I merge the master branch into this branch and the code now uses data attributes. What do you think?

jgm:
I've implemented @tarleb's option 3 in the HTML writer (for HTML 5 only), in commit d441e656db576f266c4866e65ff9e4705d376381.

I don't see any real advantage of ignoring an unknown attribute vs prefixing with `data-`.  The extra `data-` attribute should be harmless enough.

This allows us to use non-`data`-prefixed attributes in spans, e.g. in PR #3639.

mb21:
@jgm not sure about the HTML4 behaviour... technically, you're not allowed to use custom attributes in HTML4 - at all, but I'm sure there are people relying on it. I guess HTML4 export is becoming more of a legacy feature anyway...

jgm:
We could omit simply these attributes in HTML4.  But note,
it's a list of HTML5 attributes; we'd have to add a few
for HTML4.  I guess since HTML4 is a bit legacy, my
inclination is to keep it as it is.

+++ Mauro Bieg [Jul 25 17 05:22 ]:

schrieveslaach:
@jgm, I removed the data prefixes.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -659,6 +659,19 @@ enquote = do
      then singleQuoted <$> withQuoteContext InSingleQuote tok
      else doubleQuoted <$> withQuoteContext InDoubleQuote tok
 
+doAcronym :: PandocMonad m => String -> LP m Inlines
+doAcronym form = do
+  acro <- braced
+  return . mconcat $ [spanWith ("",[],[("acronym-label", toksToString acro), ("acronym-form", "singular+" ++ form)])
```
jgm: This line, and 672, should be wrapped to stay within 80 columns, as mentioned in CONTRIBUTING.md.

schrieveslaach: done

dilawar:
@schrieveslaach Is there any documentation/blog or example on how to use this feature?

schrieveslaach:
@dilawar, unfortunately there is no blog. Basically, pandoc wraps e. g. `\gls{}` commands into spans:

Bash command.

```bash
echo "\gls{API}" | pandoc -f latex -t html
```

HTML output

```html
<p><span data-acronym-label="API" data-acronym-form="singular+short">API</span></p>
``` 

I wrote a `panflute` filter which maps the values of `\newacronym{}` to the actual values. I could share that one as snippet. 

Actually, it would be nice, if pandoc could resolve them automatically. @jgm, what do you think?

PR #3580, opened 2017-04-16
Read image size of Inkscape SVGs

schrieveslaach:
I want to resolve #3579

PR #3593, opened 2017-04-20
Improve SVG image size code

schrieveslaach:
Improves code of reading svg image size (see @mb21's comment in #3580).

### DIFF
--- a/src/Text/Pandoc/ImageSize.hs
+++ b/src/Text/Pandoc/ImageSize.hs
@@ -116,7 +116,11 @@ imageType img = case B.take 4 img of
                      _                  -> mzero
 
 findSvgTag :: ByteString -> Bool
-findSvgTag img = B.null $ snd (B.breakSubstring img "<svg")
+findSvgTag img = prfx `elem` ["<svg ", "<svg\n", "<svg\r"]
+  where
+    prfx = B.take 5 $ last $ B.groupBy openingTag $ B.drop 7 img
+    -- B.groupBy openingTag matches first "<svg" or "<html" but not "<!--"
+    openingTag x y = x == '<' && y /= '!'
 
 imageSize :: WriterOptions -> ByteString -> Either String ImageSize
 imageSize opts img =
--- a/test/command/SVG_logo-without-xml-declaration.svg
+++ b/test/command/SVG_logo-without-xml-declaration.svg
@@ -1,0 +1,32 @@
+<svg viewBox="-50 -50 100 100" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<title>SVG Logo</title>
+<rect id="background" x="-50" y="-50" width="100" height="100" rx="4" fill="#f90"/>
+<rect id="top-left" x="-50" y="-50" width="50" height="50" rx="4" fill="#ffb13b"/>
+<rect id="bottom-right" width="50" height="50" rx="4" fill="#de8500"/>
+<use stroke="#f90" stroke-width="22.6" xlink:href="#a"/>
+<circle r="26"/>
+<use stroke="#000" stroke-width="12" xlink:href="#a"/>
+<g id="a">
+	<g id="b">
+		<g id="c">
+			<circle id="n" cy="-31.6" r="7.1" fill="#fff"/>
+			<path d="m0 31.6v-63.2" stroke="#fff" stroke-width="10"/>
+			<use y="63.2" xlink:href="#n"/>
+		</g>
+		<use transform="rotate(90)" xlink:href="#c"/>
+	</g>
+	<use transform="rotate(45)" xlink:href="#b"/>
+</g>
+<path id="text-backdrop" d="m44.68 0v40c0 3.333-1.667 5-5 5h-79.38c-3.333 0-5-1.667-5-5v-40"/>
+<path id="shine" d="m36 4.21c2.9 0 5.3 2.4 5.3 5.3v18c-27.6-3.4-54.9-8-82-7.7v-10.2c0-2.93 2.4-5.3 5.3-5.3z" fill="#3f3f3f"/>
+<use stroke="#000" stroke-width="7.4" xlink:href="#s"/>
+<g id="svg-text" stroke="#fff" stroke-width="6.4">
+	<g id="s">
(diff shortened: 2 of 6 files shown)


src/Text/Pandoc/ImageSize.hs
```diff
@@ -116,7 +116,11 @@ imageType img = case B.take 4 img of
                      _                  -> mzero
 
 findSvgTag :: ByteString -> Bool
-findSvgTag img = B.null $ snd (B.breakSubstring img "<svg")
+findSvgTag img = prfx `elem` ["<svg ", "<svg\n", "<svg\r"]
+  where
+    prfx = B.take 5 $ last $ B.groupBy openingTag $ B.drop 7 img
+    -- B.groupBy openingTag matches first "<svg" or "<html" but not "<!--"
+    openingTag x y = x == '<' && y /= '!'
```
jgm: What if the document begins with `<?xml`?

mb21: I think the `drop 7` was to take care of that... it's not exactly foolproof though...

Also, `last` should probably be replaced with a total function, e.g. `lastOrEmpty` where:

    let lastOrEmpty xs = foldl (\_ x -> x) [] xs

schrieveslaach: @mb21, you mentioned already in another PR that we should add tests. Do you know how I can add such a test? If we have them we can check those cases.

jgm: Instead of the drop 7, what if you changed
```
openingTag x y = x == '<' && not (y == '!' || y == '?')
```
then you could remove `drop 7`, right?

mb21: @jgm true, so we're now at:

    findSvgTag img = snip `elem` ["<svg ", "<svg\n", "<svg\r"]
      where
        snip = B.take 5 $ lastOrEmpty $ B.groupBy openingTag img
        openingTag x y = x == '<' && not (y == '!' || y == '?')
        lastOrEmpty xs = foldl (\_ x -> x) [] xs

about adding tests: maybe add a file to `/test/command` that tries to load `/test/command/svg-test.svg` etc..?

schrieveslaach: @mb21,  adding commands and checking if the svg will be parsed, seems to be rather difficult. For example, the following command does not verify the image size (same for `-t html`):

```
% pandoc -f latex -t opendocument --default-image-extension=svg
\includegraphics{../svg-test.svg}
^D
<text:p text:style-name="Text_20_body"><draw:frame draw:name="img1"><draw:image xlink:href="../svg-inkscape.0.92" xlink:type="simple" xlink:show="embed" xlink:actuate="onLoad" /></draw:frame></text:p>
```

If I use the option `-t odt`, I need to specify an output file. Verify the image size?

mb21: you could use `-t rtf` or `-t icml` which gives you the warning if the SVG is not recognized... but maybe a unit test just testing the `findSvgTag` function would be better here, but I don't think we have unit tests yet... @jgm ?

jgm: A unit test could be put into, say, tests/Tests/ImageSize.hs.
It would then also be necessary to include this in tests/test-pandoc.hs.

schrieveslaach: @mb21, I added the tests. However, the first test does fail. I assume that the warning will be redirected to stderr?

Additionally, the code example does not compile:

```
pandoc/src/Text/Pandoc/ImageSize.hs:121:23: error:
    • Couldn't match expected type ‘ByteString’ with actual type ‘[t0]’
    • In the second argument of ‘($)’, namely
        ‘lastOrEmpty $ B.groupBy openingTag img’
      In the expression:
        B.take 5 $ lastOrEmpty $ B.groupBy openingTag img
      In an equation for ‘snip’:
          snip = B.take 5 $ lastOrEmpty $ B.groupBy openingTag img
```

jgm: Which code example?

The warning to stderr is just ignored by the current command test infrastructure.

We could think about changing that somehow, but I'm not sure what the best approach is.

Btw I see two test failures on travis, not just one.

jgm:
I took a different, simpler approach, but kept your tests.

jgm:
See 03cb05f4c614f08600bcd8e90a7fd1ca13ae33a2

schrieveslaach:
nice, thanks!

mb21:
The following valid XHTML file would probably be wrongly detected as an SVG now:

    <?xml version="1.0"?>
    <!DOCTYPE html PUBLIC
        "-//W3C//DTD XHTML 1.1 plus MathML 2.0 plus SVG 1.1//EN"
        "http://www.w3.org/2002/04/xhtml-math-svg/xhtml-math-svg.dtd">
    <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
      <head>
        <title>XHTML with inline SVG</title>
      </head>
      <body>
        <h1>XHTML with inline SVG</h1>
        <svg:svg version="1.1" xmlns:svg="http://www.w3.org/2000/svg">
          <svg:rect x="0" y="0" width="100" height="100" />
        </svg:svg>
      </body>
    </html>

So maybe we should at least check that `<svg` is followed by a space or newline?

jgm:
+++ Mauro Bieg [May 21 17 03:59 ]:
>   So maybe we should at least check that <svg is followed by a space or
>   newline?

OK, makes sense.  But now with the `isInfixOf` method we
have six things to check (if we continue to allow SVG
uppercase).  That's kind of inefficient, so maybe it's
best to go back to something closer to the old method:
scanning forward for the first < followed by a letter,
and check for svg + ws.

schrieveslaach:
I think a XML validation is the right to go. Wouldn't be that the right way to check SVGs? You embed SVG into regular HTML (without namespaces). Additionally, a regular text file could contain `<svg` as well.

jgm:
Well, the svgSize function does parse the document with an
XML parser.  We could do the same thing to verify that the
image type is Svg, and this would be the most correct
approach.  But then we're doing two parses to get
the image size for an SVG; one to classify it as an Svg,
one to get the size.  That was my only reservation, but
it's not a big one.


+++ schrieveslaach [May 21 17 09:55 ]:

PR #3604, opened 2017-04-25
Add returnF to Text.Pandoc.Parsing

labdsf:


tarleb:
If we are to make this change (not sure what jgm thinks about it), I suggest to also replace the definition of `returnF` in Org.ParserState with the general definition.

Thinking about it, we could make an even greater change and remove `askF` etc from Org.ParserState as well.  Roughly:

``` haskell
newtype DelayedEval s a = DelayedEval { runDelayed :: Reader s a }
  deriving (Monad, Applicative, Functor)

type F = DelayedEval ParserState
-- in Org.ParserState: type F = DelayedEval OrgParserState
```

This would enable us to use the generic `askF` definition for all readers using the F monad, even if a custom parser state is used.

jgm:
What's the rationale for adding `returnF`?

It saves a few keystrokes, but it also (to my mind) makes it a bit less clear what is happening: the `return $ return val` makes it clear that the parser is returning a monadic value.

tarleb:
> What's the rationale for adding returnF?

My intention when defining `returnF` in the Org reader was to make it clear (to myself) that an F-wrapped monadic value is created. I was confused by `return . return` and wanted to make it obvious that this was all about the F monad.  This still helps me at times, but I'd be ok with removing it altogether. I'm also not sure why I left the type signature as general as it is, should probably have restricted it to the parser and F monad.

tarleb:
I created a new PR #3611 which includes a cherry picked version of @labdsf's commit, but changes to the Markdown reader are reverted.  The newtype unifying all used F monads mentioned above is added as a second commit.

tarleb:
Closing in favor of #3611.

PR #3605, opened 2017-04-25
Simplify linkText in Text.Pandoc.Readers.TWiki

labdsf:
Avoid constructing list of one element ([B.str url]) just to flatten it
back with mconcat.


jgm:
Thanks!

PR #3611, opened 2017-04-27
Created shared F monad functions

tarleb:
Extended version of #3604, supplemented with an approach to build the different F monads used in the readers from a single data type.  Changes to the Markdown reader are kept to a minimum.

### DIFF
--- a/src/Text/Pandoc/Parsing.hs
+++ b/src/Text/Pandoc/Parsing.hs
@@ -108,10 +108,13 @@ module Text.Pandoc.Parsing ( anyLine,
                              applyMacros',
                              Parser,
                              ParserT,
-                             F(..),
+                             F,
+                             Future(..),
                              runF,
                              askF,
                              asksF,
+                             returnF,
+                             trimInlinesF,
                              token,
                              (<+?>),
                              extractIdClass,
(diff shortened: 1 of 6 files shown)


src/Text/Pandoc/Parsing.hs
```diff
@@ -205,18 +208,30 @@ type Parser t s = Parsec t s
 
 type ParserT = ParsecT
 
-newtype F a = F { unF :: Reader ParserState a } deriving (Monad, Applicative, Functor)
+-- | Reader monad wrapping the parser state. This is used to possibly delay
+-- evaluation until all relevant information has been parsed and made available
+-- in the parser state.
+newtype Delayable s a = Delayable { runDelayed :: Reader s a }
```
jgm: Not sure about the nomenclature here.  `Delayable` sounds more like the name of a type class than a type.
My name `F` was meant to stand for `Future`, so perhaps `Future` or `Delayed` would make more sense?
Note that names like `askF` make a lot more sense if it's `F` or `Future`.

tarleb: Thank you, I changed the name multiple times and still wasn't satisfied. *Future* is so much better.

PR #3616, opened 2017-04-28
Markdown writer: Case-insensitive reference links.

### LINKED ISSUE #3615: Inconsistent case-sensitivity for markdown link references
```
% pandoc -v
pandoc 2.0
Compiled with pandoc-types 1.17.0.5, texmath 0.9.4, skylighting 0.3.3

% pandoc -f html -t markdown --reference-links
<a href="a">foo</a>,<a href="b">Foo</a>
^D
[foo],[Foo]

  [foo]: a
  [Foo]: b

% pandoc -f html -t markdown --reference-links | pandoc
<a href="a">foo</a>,<a href="b">Foo</a>
^D
[warning] Duplicate link reference '[Foo]' at line 4 column 1
<p><a href="b">foo</a>,<a href="b">Foo</a></p>
```

Note that the first link now points to "b" rather than "a".

davidar:
Ensure that we do not generate reference links
whose labels differ only by case.

Also allow implicit reference links when the link
text and label are identical up to case.

Closes #3615.

### DIFF
--- a/src/Text/Pandoc/Writers/Markdown.hs
+++ b/src/Text/Pandoc/Writers/Markdown.hs
@@ -66,7 +66,7 @@ import Text.Pandoc.Writers.Math (texMathToInlines)
 import Text.Pandoc.Writers.Shared
 
 type Notes = [[Block]]
-type Ref   = ([Inline], Target, Attr)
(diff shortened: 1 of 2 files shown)


jgm:
Great, thanks!

PR #3618, opened 2017-04-29
Org reader: Avoid creating nullMeta by applying setMeta directly

labdsf:


jgm:
Looks good to me if @tarleb has no objections.

tarleb:
Thanks!

Issue #3314, opened 2016-12-14
Treat Emacs table.el tables in Org mode files as Pandoc's markdown grid tables


unixpipecentipede:
Org mode handles both simple plain ASCII tables (created with `C-c |`)
and Emacs `table.el` tables. The latter are basically the same as
Pandoc's markdown grid tables.

It seems Pandoc's Org reader treats `table.el` tables as Org ASCII
tables, since on conversion to `docx` the output gets mangled when
dealing with rows that contain multi-line paragraphs. The resulting
table arranges every new line of text into --- often several --- new
rows.

On the other hand, Pandoc's markdown documents with grid tables get
what could be called proper treatment, i.e. 1 paragraph (including
multi-lines) == 1 row. This produces somewhat *cleaner*
tables. Division lines are displayed correctly too.

Pandoc's Org reader should distinguish between plain ASCII tables and
`table.el` tables and treat the latter as the markdown reader treats
grid tables.

chunlinyao:
table.el is the only method to support multilines in cell. I have to use `ox-latex` to export when a file contains grid tables.

PR #3620, opened 2017-04-30
[WIP] Add Muse reader

labdsf:
I have created a basic Muse reader.  It supports headings, inline markup and footnotes, but no tables and lists. It is far from full markup support, but it satisfies my needs so I plan to continue development slowly from now on. Hope it gets merged.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -380,6 +380,7 @@ Library
                    Text.Pandoc.Readers.Docx,
                    Text.Pandoc.Readers.Odt,
                    Text.Pandoc.Readers.EPUB,
+                   Text.Pandoc.Readers.Muse,
                    Text.Pandoc.Writers,
                    Text.Pandoc.Writers.Native,
                    Text.Pandoc.Writers.Docbook,
@@ -559,6 +560,7 @@ Test-Suite test-pandoc
                   Tests.Readers.Odt
                   Tests.Readers.Txt2Tags
                   Tests.Readers.EPUB
+                  Tests.Readers.Muse
                   Tests.Writers.Native
                   Tests.Writers.ConTeXt
                   Tests.Writers.Docbook
--- a/src/Text/Pandoc/Readers.hs
+++ b/src/Text/Pandoc/Readers.hs
@@ -59,6 +59,7 @@ module Text.Pandoc.Readers
   , readTWiki
   , readTxt2Tags
   , readEPUB
+  , readMuse
   -- * Miscellaneous
   , getReader
   , getDefaultExtensions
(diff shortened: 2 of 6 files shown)


jgm:
I'd rather not merge until this has support for tables and lists.

labdsf:
@jgm Ok, will make them high priority then.

tarleb:
Regarding tables: My understanding is that Muse supports tables.el and org tables.  The org reader should be able to read tables.el formatted tables as well (#3314). It seems to me that a simple way to do this would be to extract the grid table parser from the markdown reader into a separate module. Same for the org table parser. @jgm, would you be okay with a separate `Table` module, collecting all (shared) table parsers?

jgm:
Text.Pandoc.Parsing already contains gridTableWith.
Can you just use that?

+++ Albert Krewinkel [May 01 17 07:43 ]:

tarleb:
> Text.Pandoc.Parsing already contains gridTableWith. Can you just use that?

Ah, yes. I had missed that.

I generalized the signature of `tableWith` to

``` haskell
tableWith :: (Stream s m Char, HasReaderOptions st, Monad mf)
          => ParserT s st m (mf [Blocks], [Alignment], [Int])
          -> ([Int] -> ParserT s st m (mf [Blocks]))
          -> ParserT s st m sep
          -> ParserT s st m end
          -> ParserT s st m (mf Blocks)
```

i.e. the state is no longer fixed to `ParserState` and table-cell block parsers must return values wrapped in a monad.  This requires a bit of `Identity` wrapping and unwrapping in the RST reader, but seems to work well.

jgm:
Oh, good.  I wonder if your generalization of gridTableWith
would allow it to be used in the Markdown reader?

Currently we don't use it there, mostly because of the
F monad, if I recall.

(But part of the reason is also that in Markdown grid
tables we support alignments now.  Actually, it should
be fairly harmless to allow those in the RST reader, since
an actual RST document wouldn't have alignment markings.)

tarleb:
I've pushed the changes and can take a look into merging the Markdown-specific gridTable parser into the generic parser.

labdsf:
Added tables support. It is aimed to be compatible with Emacs Muse, not Text::Amuse, as Text::Amuse differences are considered a bug: melmothx/text-amuse#14

For example, tables are not required to be indented with a space, although writer indents them now for compatibility with Text::Amuse #3649.

labdsf:
One bug remains: tables that consist only of caption crash because `museToPandocTable` tries to compute `maximum` over empty list.

labdsf:
Added bullet lists. It always produces loose lists, just like A Muse Wiki. AFAIR Emacs Muse distinguishes between tight and loose lists, but I would rather follow A Muse Wiki here and avoid overcomplicating things complexity just for backwards compatibility.

labdsf:
@jgm Tables and lists are now supported. I rebased against master and removed all hlint warnings. No definition lists and Org tables (A Muse Wiki does not support them either, though) yet.

jgm:
Great, thanks!

melmothx:
Text::Amuse does support definition lists https://amusewiki.org/library/manual#toc16 (documented since February 2016).

https://amusewiki.org/git/amw/commit/m/ml/manual.muse?id=73579f4c26aa047742cb319f3661a203b2fc7eb7

labdsf:
@melmothx "A Muse Wiki does not support them either, though" was related to Org tables only.

PR #3623, opened 2017-04-30
organize README

Alok:
split sections into lists to make them easier to read.


jgm:
Thanks for submitting this PR!  I guess this is a matter of taste; I prefer it the way it is now, so I'm not going to merge this. But please don't be discouraged from submitting other PRs to pandoc.

PR #3625, opened 2017-05-01
FB2 writer: Add support for "lang" metadata

labdsf:
I need this when converting from Muse to FB2. Muse uses `#lang` according to its [manual](http://amusewiki.org/library/manual#toc11) and FB2 uses `<lang>` tag, and EPUB writer also uses `lang` metadata, so everything is compatible.

### DIFF
--- a/src/Text/Pandoc/Writers/FB2.hs
+++ b/src/Text/Pandoc/Writers/FB2.hs
@@ -117,8 +117,13 @@ description meta' = do
   bt <- booktitle meta'
   let as = authors meta'
   dd <- docdate meta'
-  return $ el "description"
-    [ el "title-info" (bt ++ as ++ dd)
+  let lang = case lookupMeta "lang" meta' of
+               Just (MetaInlines [Str s]) -> [el "lang" $ iso639 s]
+               Just (MetaString s)        -> [el "lang" $ iso639 s]
+               _                          -> []
+             where iso639 = takeWhile (/= '-') -- Convert BCP 47 to ISO 639
+  return $ el "description" $
+    [ el "title-info" (bt ++ as ++ dd ++ lang)
     , el "document-info" [ el "program-used" "pandoc" ] -- FIXME: +version
     ]
 


mb21:
Pandoc's `lang` [metadata field is in BCP 47](https://github.com/jgm/pandoc/blob/master/MANUAL.txt#L1221) (e.g. `en-GB`), which is basically the ISO 639 language code (e.g. `en`) followed by an optional part indicating the regions/script etc.

Not exactly sure what the Muse people mean with "The language code of the document (2 or 3 letters)", but sounds like the dash and everything following it should be stripped.

labdsf:
@mb21 What Muse uses is irrelevant here, but my Muse documents use something like "ru" instead of "ru-RU". It is really unspecified, directives in Muse are free form, so I am not going to do anything with them in my Muse reader (see PR #3620).

What really matters is what FB2 standard says, [and looks like it uses ISO 639](http://www.fictionbook.org/index.php/%D0%9E%D0%BF%D0%B8%D1%81%D0%B0%D0%BD%D0%B8%D0%B5_%D1%84%D0%BE%D1%80%D0%BC%D0%B0%D1%82%D0%B0_FB2_%D0%BE%D1%82_Sclex#.D0.A1.D0.BF.D0.B8.D1.81.D0.BE.D0.BA_.D0.B2.D0.BE.D0.B7.D0.BC.D0.BE.D0.B6.D0.BD.D1.8B.D1.85_.D1.8F.D0.B7.D1.8B.D0.BA.D0.BE.D0.B2).

So I should probably try to convert BCP 47 to ISO 639 when writing `lang` to FB2.

labdsf:
Also related: https://bugs.launchpad.net/calibre/+bug/990026

labdsf:
Looks like FB2 documents actually use "en" and no "en-US" or "en-GB", so just stripping everything up to "-" is enough. Fixed it.

jgm:
Thanks!

PR #3627, opened 2017-05-02
Add basic \textcolor support to LaTeX reader

schrieveslaach:
In LaTeX you can use the `\textcolor` command to mark text in a different color (see example below). Pandoc does not support this command and the text will be excluded in the result. However, pandoc should at least include the text content. This PR will include the text content of the `\textcolor` as span in the result. 

```latex
\textcolor{red}{Hello World}
```


schrieveslaach:
@jgm, @mb21, do you know, how I can support following LaTeX code?

```
\textcolor{orange}{
\begin{itemize}	
\item Test
\end{itemize}
}
```

jgm:
Would it work to match `\textcolor` as both a block or an inline command?
(As we do currently with `\hspace` for example.)
When used as a block command, the result could be the same, except in a div rather than a span.

schrieveslaach:
@jgm, I'm trying to add a function parsing textcolor as blocks. See d943980. However, I'm not able to compile: 

```
src/Text/Pandoc/Readers/LaTeX.hs:423:5: error:
    • Couldn't match type ‘Many Block’
                     with ‘Text.Parsec.Prim.ParsecT String ParserState m0 Blocks’
      Expected type: LP m0 Blocks
                     -> Text.Parsec.Prim.ParsecT String ParserState m Blocks
        Actual type: Blocks
                     -> Text.Parsec.Prim.ParsecT String ParserState m Blocks
    • The first argument of ($) takes one argument,
      its type is ‘f0 b0’,
      it is specialized to ‘Blocks
                            -> Text.Parsec.Prim.ParsecT String ParserState m Blocks’
      In a stmt of a 'do' block:
        return <$> divWith ("", [], [("style", "color: " ++ color)])
        $ block
      In the expression:
        do { skipopts;
             color <- braced;
             return <$> divWith ("", [], [("style", "color: " ++ color)])
             $ block }
```

What is wrong with my code?

schrieveslaach:
@jgm, is there a way to include the colors in an ODT document? I think, that pandoc has no general support for coloring text, right?

jgm:
+++ schrieveslaach [May 03 17 04:41 ]:
>   [1]@jgm, is there a way to include the colors in an ODT document? I
>   think, that pandoc has no general support for coloring text, right?

That's correct, there's no general support.  One approach
would be to write a filter that checks the Span or Div
for a color style attribute and then substitutes the
appropriate opendocument xml -- whatever that might be.

One might also contemplate adding this support directly
to the writer; I'm not sure what I think about that.

schrieveslaach:
OK, until then I will use a pandoc filter.

jgm:
One drawback with the present method of allowing `\color` to take either block or inline argument (and the same applies to the present treatment of `\hspace`) is that for this input we'll get two blocks instead of one:
```
\color{red}(hi} there
```
I suppose one possible solution would be to add `notFollowedBy inline` or something like that to the block case?

schrieveslaach:
@jgm, fixed.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -414,6 +416,12 @@ blockCommands = M.fromList $
   , "pagebreak"
   ]
 
+blockTextcolor :: PandocMonad m => LP m Blocks
+blockTextcolor = do
+    skipopts 
+    color <- braced
+    divWith ("",[],[("style","color: " ++ color)]) <$> grouped block <* notFollowedBy inline
+
```
jgm: Small problem here:  you should add a `try` before the initial `do`.  Otherwise, you'll get a failure if the `\color` command is parsed and the `notFollowedBy` fails, because the whole parser will have failed after having consumed some text.

schrieveslaach: I assume that you think about following case, right?

```latex
\textcolor{red}{
\begin{itemize}
    \item Test 1
    \item Test 2
\end{itemize}
} some more text
```
I tried to add `try` before, however, this fix is not working (see diff). 

```diff
diff --git a/src/Text/Pandoc/Readers/LaTeX.hs b/src/Text/Pandoc/Readers/LaTeX.hs
index a34be46e..85a77354 100644
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -417,7 +417,7 @@ blockCommands = M.fromList $
   ]
 
 blockTextcolor :: PandocMonad m => LP m Blocks
-blockTextcolor = do
+blockTextcolor = try $ do
     skipopts 
     color <- braced
     divWith ("",[],[("style","color: " ++ color)]) <$> grouped block <* notFollowedBy inline
```

I think, that using `notFollowedBy inline` need to be changed. LaTeX puts _some more text_ into a new paragraph. Pandoc should do the same thing. Do you agree?

jgm: Yeah, this is tricky.  The problem is that we want
```
\textcolor{red}{Hi} there
```
to be parsed as a single paragraph, and
```
\textcolor{red}{
\begin{itemize}
    \item Test 1
    \item Test 2
\end{itemize}
} some more text
```
to be parsed as a colored list followed by a paragraph "some more text."  But if the block parser accepts "Hi" as a block, then it will handle these both the same.

Perhaps the thing to do is to have the block version fail if the contents of the braces can be exhaustively parsed as a sequence of inlines; that will allow the parser to fall through to the inline version.  Not very pretty, though, since we'd be parsing the same contents twice.

Slightly less ugly, maybe:  have the block version attempt to parse the bracketed stuff as inlines; if this succeeds, then parse many inlines following the textcolor command, and emit a single paragraph containing the textcolor command and the following inlines.  This way we don't double-parse.

So, something like
```
blockTextcolor = do
  skipopts
  color <- braced
  let constructor = divWith ("",[],[("style","color: " ++ color)])
  inlineContents <|> blockContents
  where inlineContents = do{ ils <- grouped inline; rest <- inlines; return (para (ils <> rest))}
        blockContents = grouped block
```

jgm: PS. I used {} in inlineContents, because I can't indent in this text box, but you should use indentation probably.

jgm: @schrieveslaach are you planning to make revisions along the lines I suggest above?

schrieveslaach: @jgm, sorry for the delay. I added your suggestions, however, now test first test is failing. I'm not sure why pandoc does not recognise `\textcolor` as inline anymore.

schrieveslaach:
I also added support for `\colorbox` command. 

I'm still not sure why some examples do not work. @jgm, @mb21, any ideas?

schrieveslaach:
@jgm, I'm surprised that you merged this PR because the tests still fail. Are you going to fix them?

jgm:
They passed when I built it.  I made a few changes to your
code in merging, perhaps that did it?

+++ schrieveslaach [Jul 25 17 11:25 ]:

jgm:
Note that the travis builds are broken because of a bug
in cabal, but the stack build should pass.

+++ schrieveslaach [Jul 25 17 11:25 ]:

schrieveslaach:
Yes, that seems so. Thanks for merging.

PR #3631, opened 2017-05-02
Markdown Writer: put space before refLinks, fixes #3630

### LINKED ISSUE #3630: link_attributes vs. reference-links vs. markdown writer
$ pandoc -v
    pandoc 1.19.2.1
    Compiled with pandoc-types 1.17.0.4, texmath 0.9, skylighting 0.1.1.4
    ...

When converting from markdown to markdown a space seems to be missing before the attributes in the reference:

    $ pandoc -f markdown -t markdown --reference-links
    foo{#id:1}
    ^D
    ![foo]

        [foo]: bar.png{#id:1}

If this is (again) sent through Pandoc, the former attributes become part of the image name as in:

     $ pandoc -f markdown -t markdown --reference-links | pandoc -f markdown -t html
     foo{#id:1}
     ^D
     <div class="figure">
     <embed src="bar.png%7B#id:1%7D" />
     <p class="caption">foo</p>
     </div>

mb21:
fixes https://github.com/jgm/pandoc/issues/3630

### DIFF
--- a/src/Text/Pandoc/Writers/Markdown.hs
+++ b/src/Text/Pandoc/Writers/Markdown.hs
@@ -241,7 +241,7 @@ keyToMarkdown opts (label', (src, tit), attr) = do
                 else space <> "\"" <> text tit <> "\""
   return $ nest 2 $ hang 2
             ("[" <> label' <> "]:" <> space) (text src <> tit')
-            <> linkAttributes opts attr
+            <+> linkAttributes opts attr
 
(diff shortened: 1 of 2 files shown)


jgm:
Looks good.  Can you add a regression test, `test/command/3630.md`?

mb21:
right. I knew there was something more to do than write one character...

jgm:
Great, thanks

PR #3632, opened 2017-05-03
LaTeX: Add support for tabularx environment

schrieveslaach:
In LaTeX you can use the `tabularx` environment as an alternative to `tabular` environment. For example:

```latex
\begin{tabularx}{\linewidth}{|b{0.25\linewidth}|c|m{0.25\linewidth}|}
\hline
  Column Heading 1
  & Column Heading 2
  & Column Heading 3 \\
\hline
  Cell 1.1 
  & Cell 1.2
  & Cell 1.3 \\
\hline
  Cell 2.1 
  & Cell 2.2
  & Cell 2.3 \\
\hline
  Cell 3.1 
  & Cell 3.2
  & Cell 3.3 \\
\hline
\end{tabularx}
```

This PR adds support for this environment.


jgm:
This is already an improvement, but it makes me realize that we currently don't parse column width information for any LaTeX tables.  So that could be a further improvement, though more complex.

PR #3642, opened 2017-05-05
Fix keyval funtion: pandoc did not parse options in braces correctly.…

schrieveslaach:
… Additionally, dot, dash, and colon were no valid characters. Therefore, options, for example, of the `lstlsiting` won't be parsed correctly. See following LaTeX code

```latex
\begin{lstlisting}[language=Java, caption={Java Example}, label=lst:Hello-World]
public class World {
    public static void main(String[] args) {
        System.out.println("Hello World");
    }
}
\end{lstlisting}
```

The resulting `CodeBlock` did not contain `label` and `caption` as attributes. 

This PR fixes this issue.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -1095,7 +1095,7 @@ parseListingsOptions options =
 keyval :: PandocMonad m => LP m (String, String)
 keyval = try $ do
   key <- many1 alphaNum
-  val <- option "" $ char '=' >> (braced) <|> (many1 (alphaNum <|> char '.' <|> char ':' <|> char '-' <|> char '\\'))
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -1095,7 +1095,7 @@ parseListingsOptions options =
 keyval :: PandocMonad m => LP m (String, String)
 keyval = try $ do
   key <- many1 alphaNum
-  val <- option "" $ char '=' >> many1 (alphaNum <|> char '.' <|> char '\\')
+  val <- option "" $ char '=' >> (braced) <|> (many1 (alphaNum <|> char '.' <|> char ':' <|> char '-' <|> char '\\'))
   skipMany spaceChar
```
jgm: More idiomatic:
```
  val <- option "" $ char '=' >> braced <|> (many1 (alphaNum <|> oneOf ".:-\\"))
```

schrieveslaach: done

jgm:
Thanks!

Issue #3638, opened 2017-05-04
Merge generic and Markdown-specific grid table parsers


tarleb:
_Note to self: don't try opening an issue from mobile._

The markdown and RST/generic grid table parser have a different methods of compactifying table cells. The generic parser acts in a per-cell level, while the md parser looks at rows. A two element grid table

    +-------+-------+
    | Hello | World |
    +-------+-------+

will give

    [Table [] [AlignDefault,AlignDefault] [0.1111111111111111,0.1111111111111111]
     [[]
     ,[]]
     [[[Para [Str "Hello"]]
      ,[Para [Str "World"]]]]]

when parsed as markdown and

    [Table [] [AlignDefault,AlignDefault] [0.1111111111111111,0.1111111111111111]
     [[]
     ,[]]
     [[[Plain [Str "Hello"]]
      ,[Plain [Str "World"]]]]]

when parsed as rst.  Is this intended behavior which needs to be preserved?

jgm:
I'm pretty sure it would be safe to have these behave the same way.
Most likely the Markdown parser's way is more correct, but I'm not sure, and I didn't quite understand from your brief description why there's a difference in behavior.

tarleb:
Sorry for the bad explanation, should have taken more time instead of creating this issue in passing.

What's happening is that the Markdown parser compactifies rows the same way that lists are compactified: If none of the other cells in that row contain a para, then the cell in the final column will be converted from Para to Plain:

    +--------+-------+
    | - India | Papa |
    +---------+------+

reads the row as

    [[BulletList
      [[Plain [Str "India"]]]]
     ,[Plain [Str "Papa"]]]

while

    +-------+------+
    | India | Papa |
    +-------+------+

gives

    [[Para [Str "India"]]
     ,[Para [Str "Papa"]]]

Note that the block type of the last cell depends on the cell content of the rest of the cells in that row.

I'd argue that the RST behavior is more natural, in that the block type of the second cell is kept constant, regardless of the first cell's content.  When parsed as RST, the above examples result in the rows

    [[BulletList
      [[Plain [Str "India"]]]]
      ,[Plain [Str "Papa"]]

and

    [[Plain [Str "India"]]
    ,[Plain [Str "Papa"]]]

respectively.

tarleb:
The test suite revealed a second difference between the parsers:  The markdown parser allows header cells to contain multiple blocks, while the rst reader will treat even multi-line headers as if they had been given in a single line.  E.g.:

[...]

PR #3643, opened 2017-05-05
Merge grid table parsers

tarleb:
Closes: #3638


jgm:
@tarleb I'm not sure where this stands, since there was a simultaneous discussion on #3638.
Do you think this is ready to merge, or do some of the issues need to be further discussed?

tarleb:
Yes, I believe it's ready to merge, but I wasn't sure if you agreed with the conclusions in #3638. I'm going to write a better commit message describing the changes and then merge it.

jgm:
It looks like the change in compactifying was just a mistake
on my part...

PR #3647, opened 2017-05-07
Markdown writer: missing \n

davidar:
This fixes the behaviour of `echo '<ul><li>foo</li></ul><ul><li>bar</li><li>baz</li></ul>' | pandoc -f html -t markdown-raw_html | pandoc`


jgm:
Thanks.

PR #3652, opened 2017-05-08
Fix #3651.

### LINKED ISSUE #3651: \href raise an error inside \caption with links-as-notes
This seems similar to #2490, but the error is caused by a missing `\protect` before a `\href` (not `\hyperlink`). Adding the `\protect` manually solves the error.

input:

```
pandoc -V links-as-notes -o test.pdf << EOT
# This is a test

![This is an image. It has a [link](http://pandoc.org) in captions.](http://pandoc.org/diagram.jpg)

EOT
```

output:

```
! Argument of \Hy@tempa has an extra }.
<inserted text> 
                \par 
l.77 captions.}

pandoc: Error producing PDF
```

pandoc 1.19.2.1
Compiled with pandoc-types 1.17.0.5, texmath 0.9.4, skylighting 0.1.1.5

wilx:
Declare our redefined `\href` robust.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -155,7 +155,7 @@ $if(graphics)$
 $endif$
 $if(links-as-notes)$
 % Make links footnotes instead of hotlinks:
-\renewcommand{\href}[2]{#2\footnote{\url{#1}}}
+\DeclareRobustCommand{\href}[2]{#2\footnote{\url{#1}}}
 $endif$
 $if(strikeout)$
 \usepackage[normalem]{ulem}


data/templates/default.latex
```diff
@@ -156,6 +156,7 @@ $endif$
 $if(links-as-notes)$
 % Make links footnotes instead of hotlinks:
 \renewcommand{\href}[2]{#2\footnote{\url{#1}}}
+\DeclareRobustCommand{\href}[2]{#2\footnote{\url{#1}}}
```
jgm: Not too familiar with DeclareRobustCommand -- if we have this, do we still need the line before with renewcommand?  It's an odd repetition of the definition.

Also, do you think this is preferable to simply adding a `\protect` in front of `\href` in the source?  Might this `\protect` be necessary for some other cases too, besides links-in-notes?

wilx: Oh, no. I just blindly copied the the line from the other changeset and forgot to remove the prev line. I will fix this ASAP.

wilx: Fixed.

wilx:
> Also, do you think this is preferable to simply adding a `\protect` in front of `\href` in the source? Might this `\protect` be necessary for some other cases too, besides links-in-notes?

I thought about that but then I took a look at the definition of [`\href`](https://github.com/ho-tex/hyperref/blob/master/hyperref.dtx#L7873). It uses `\DeclareRobustCommand*`. I used the no-asterisk version because previously the template also uses no-asterisk version of `\renewcommand`.

Issue #2821, opened 2016-03-27
Support Github-flavored Markdown anchors/id scheme


rkitover:
If you have a section heading like:

``` markdown
## -c cmd
```

It generates the link `#c-cmd` but it **SHOULD** generate the link `#-c-cmd`, so the link does not work.

The command is being run with `-t markdown_github` .

jgm:
+++ Rafael Kitover [Mar 27 16 08:24 ]:
>   If you have a section heading like:
>
>## -c cmd
>
>   It generates the link #c-cmd but it SHOULD generate the link #-c-cmd,
>   so the link does not work.

SHOULD, according to what?  See the documentation for the `auto_identifiers`
extension.

rkitover:
github makes the anchor `#-c-cmd`, so I would expect --toc -t markdown_github to generate a TOC usable on github?

I'm using doctoc for this at the moment, which makes the right TOC links.

jgm:
Is the algorithm github uses for generating automatic
identifiers for headers documented anywhere?

+++ Rafael Kitover [Mar 27 16 12:38 ]:

rkitover:
I have only found this:

http://stackoverflow.com/questions/2822089/how-to-link-to-part-of-the-same-document-in-markdown

and this:

https://gist.github.com/asabaylus/3071099

it seems that the section heading is separated into words, lowercased, and the words are separated with dashes.

pandoc does almost the same thing, but in the case of commandline options like `## -foo` it removes the leading dash while github doesn't. My normal section links worked just fine when I used --toc -t markdown_github .

jgm:
If I recall, the motivation for this was that id attributes in HTML 4 had to begin with a letter, and so couldn't begin with a hyphen. This has been relaxed in HTML 5 (which GitHub is targeting).

We use a single automatic identifier generating scheme for all formats.  Indeed, the identifier is assigned by the *reader*, which doesn't even know whether the ultimate output is going to be HTML5, HTML4, or both or neither.

So it wouldn't be a simple change to allow the hyphen at the beginning of an automatically generated identifier.

jgm:
[Related issue with headers starting with a number](https://groups.google.com/d/msgid/pandoc-discuss/d38ed6d8-6045-4401-9c7e-7466ca0f916a@googlegroups.com?utm_medium=email&utm_source=footer)

[...]

PR #3655, opened 2017-05-08
GitHub Markdown: Don't use dots in header identifiers

asmgf:
GitHub doesn't use dots in header identifiers, but Pandoc does.
So when a header contains dots, links to the header that work on GitHub won't work in a HTML page built using Pandoc.

Here is an example of such Markdown document: https://github.com/asmgf/github-markdown-identifiers/blob/master/README.md


jgm:
Instead of `no_dots_in_identifiers`, I'd rather see an extension that matches all aspects of the GitHub automatic identifier scheme.  See #2821.

jgm:
Thanks for the PR.  I'm closing it for the reason noted above.  We need a general fix for #2821, not just this special case.

PR #3658, opened 2017-05-11
Change FigureWithCaption to CaptionedFigure

### LINKED ISSUE #3656: Using a custom.docx breaks "Figure with Caption" Styling.
This is pretty simple. First create two custom docx reference templates:

```shell
pandoc --print-default-data-file reference.docx > customA.docx
pandoc --print-default-data-file reference.docx > customB.docx
```

Now edit customB.docx in Word — all you need to do is open it, add a space, delete the space and save. Now if you use this markdown and those templates:

```markdown
# Simple test #

![Test image](test.png)  

More text.  
```

```shell
pandoc --reference-docx=customA.docx -o resultA.docx input.md
pandoc --reference-docx=customB.docx -o resultB.docx input.md
```

You find in for resultB.docx in Word that the figure loses its "Figure with Caption" styling (and thus keep with next), and is now "Normal".

The reason is a case issue caused by Word editing the styles on save, here is styles.xml for A vs. B:

screen shot 2017-05-10 at 12 50 54

(LEFT==resultA.docx | RIGHT=resultB.docx) For some reason, Word overwrites the w:styleId = Figure**W**ithCaption with the w:val extrapolation so the w:styleId becomes Figure**w**ithCaption. But the document.xml still uses the uppercase version and thus fails to apply. The easiest fix I assume is to rename the original template so _both_ w:val and w:styleId use an uppercase **W**. I couldn't see other styles that use mixed case in the pandoc template, but assume this applies to any such style.

iandol:
Basic edit to fix #3656

### DIFF
--- a/data/docx/word/styles.xml
+++ b/data/docx/word/styles.xml
@@ -400,8 +400,8 @@
     <w:name w:val="Figure" />
     <w:basedOn w:val="Normal" />
   </w:style>
-  <w:style w:type="paragraph" w:customStyle="1" w:styleId="FigureWithCaption">
-    <w:name w:val="Figure with Caption" />
+  <w:style w:type="paragraph" w:customStyle="1" w:styleId="CaptionedFigure">
+    <w:name w:val="Captioned Figure" />
     <w:basedOn w:val="Figure" />
     <w:pPr>
       <w:keepNext />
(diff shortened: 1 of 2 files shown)


jgm:
Thanks:  can you add the change to the style definition?
This requires editing a file in the data/docx tree (the
repository doesn't actually contain reference.docx; that is
constructed from these pieces).

iandol:
OK, edit of styles.xml added to this patch-1 branch.

jgm:
Thanks!

Issue #3660, opened 2017-05-11
MD --> PDF via XeLaTeX RTL language with tables compile error


lwolfsonkin:
It seems that the Markdown --> PDF (via XeLaTeX) has issue with RTL languages declarations along with the use of tables. Here, with the following example:

```markdown
---
lang: en
otherlangs:
    - he
mainfont: Arial
---

# Test

| header |
|:------:|
| first  |
| second |
| third  |
```

I get the following error:

~~~
! Package bidi Error: Oops! you have loaded package longtable after bidi packag
e. Please load package longtable before bidi package, and then try to run xelat
ex on your document again.

See the bidi package documentation for explanation.
~~~

I believe this to occur because the polyglossia package calls the bidi package implicitly for RTL languages, and the `longtable` package must be called after the `bidi` package.

To fix this locally, I moved the `lang` variable part of the latex template to the first thing before the `bidi` variable handling.

I don't believe this should cause other issues.

jgm:
This should now be fixed. Feel free to comment here if it isn't fixed by #3661.

PR #3661, opened 2017-05-11
modified template.latex to fix XeLaTex being used with tables

lwolfsonkin:
Attempt to fix issue #3660 

In `template.latex`: reordered `lang` variable handling to immediately before `bidi`, in order to avoid errors related to `bidi` being called too early, as `bidi` is loaded by `polyglossia` for RTL languages as described in #3660.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -112,20 +112,6 @@ $endif$
 $if(geometry)$
 \usepackage[$for(geometry)$$geometry$$sep$,$endfor$]{geometry}
 $endif$
-$if(lang)$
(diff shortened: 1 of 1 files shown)


mb21:
sounds good to me, do you know of any issues with this reordering @adunning ?

btw, the [latex-lang test](https://github.com/jgm/pandoc/blob/master/test/writers-lang-and-dir.latex) would have to be updated...

adunning:
That should work, as far as I can see.

jgm:
The failures are due to changes needed in the tests.
To update the tests, you can do `stack test --test-arguments='--accept'` and then very carefully check the result of `git diff` to make sure the changes to the tests are as expected.

jgm:
I'll fix the tests.

PR #3668, opened 2017-05-13
Update copyright notices

tarleb:
This updates the dates of most copyright notices.

I'm not sure this is necessary. The GNU project [suggests it](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html) for GPL licensed software, but most other sources I found on the net say that it's unnecessary, and that just keeping the first year of publication is more than enough.

If this is merged, should I also add missing copyright notices (e.g. for the Docx reader) to COPYRIGHT?

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -32,6 +32,12 @@ statements for these sources are included below.  All are GPL-compatible
 licenses.
 
 ----------------------------------------------------------------------
+src/Text/Pandoc/Writers/Muse.hs
+Copyright (C) 2017 Alexander Krotov
+
+Released under the GNU General Public License version 2 or later.
+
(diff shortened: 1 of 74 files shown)


tarleb:
I added a commit to list all new files and their authors in the COPYRIGHT file. This seems like a nice way to credit the work that went into these files, but the tediousness of keeping the list up to date makes me wonder if this is really necessary. Isn't it relevant for third-party software only?

If legal issues are a concern, adoption of the [AUTHORS certificate](https://github.com/berneout/authors-certificate) could be an option.

jgm:
Thanks. I don't know if it's really necessary; this file goes back to the original debianization and was, at that time, considered good practice.  But things may have changed.

PR #3670, opened 2017-05-15
Implemented #168

Drezil:
This PR implements #168 as described in https://github.com/jgm/pandoc/issues/168#issuecomment-270833294

It should work for normal and nested blocks with both syntaxes described.

There are no tests yet, as i don't know how you want them implemented properly.

A `stack test` shows that it should not break other things:

`All 968 tests passed (81.94s)`

### DIFF
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -746,6 +746,8 @@ birdTrackLine c = try $ do
   char c
   -- allow html tags on left margin:
   when (c == '<') $ notFollowedBy letter
+  -- allow inplace-images on left margin:
+  when (c == '!') $ notFollowedBy (char '[')
   anyLine
 
 --
@@ -782,18 +784,25 @@ blockQuote = do
 
 openingDiv :: PandocMonad m => MarkdownParser m Attr
 openingDiv = do
-  string "; ---"
+  string ":::"
+  skipMany (char ':')
   skipMany spaceChar
-  string "div"
+  mclass <- optionMaybe $ do
+      c <- many1 alphaNum
+      skipMany1 spaceChar
+      skipMany1 (char ':')
+      return c
   attr' <- option nullAttr $ do
     skipMany spaceChar
     attributes
+  let addClass :: Attr -> String -> Attr
+      addClass (id',cs, kv) c = (id', (c:cs), kv)
   anyLine
-  return attr'
+  return $ maybe attr' (addClass attr') mclass -- if we have mclass, add it, otherwise leave it unchanged
 
 closingDiv :: PandocMonad m => MarkdownParser m ()
 closingDiv = do
-  string "; ---"
+  string ":::"
   skipMany spaceChar
   ((char '\n' >> return ()) <|> eof)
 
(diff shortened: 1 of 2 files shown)


Drezil:
(failed travis-check is due to timeout during compilation on their end for 2 of 4 vms)

Drezil:
@jgm anything that is missing or i overlooked before this could be merged?

I am willing to invest some hours into this to get things released as soon and as smooth as possible ;)

Would the writer have to be updated as well to generate the corresponding md? so that `read . write = id` holds?

mb21:
@Drezil It's great to have some working code to play around with and see how the syntax feels like to work with on real documents, however I guess this won't be merged until a syntax is decided for good...

mb21:
> as this AST-Element never gets created in the markdown-parser.

you could always use the HTML div syntax which the markdown-reader parses as a native pandoc div.

btw, I agree that it would be great to decide on a definitive div syntax, but that's up to @jgm...

tarleb:
>  i just think that we need (at least one) way to generate every element in the AST for filters & interop..

Using HTML syntax in markdown works for this:

``` md
<div id="foo" class="bar">
# Slide

Wassup?
</div>
```

This will wrap the elements in a div the AST as long as the *markdown_in_html* extension is enabled (which it is by default).

Drezil:
Yes - but using markdown_in_html it gets parsed as `RawInline` instead of `Div` and thus is unusable by filters in a sane way.

What if i want to switch from reveal.js to latex-beamer as a fallback? HTML won't cut it there.

tarleb:
I don't quite follow.  For me, the above results in 

```
[Div ("foo",["bar"],[])
 [Header 1 ("hello",[],[]) [Str "Hello"]
 ,Para [Str "Moin"]
 ,Header 1 ("tach",[],[]) [Str "Tach"]
 ,Para [Str "Wassup?"]]]
```

when read as pandoc markdown.  Isn't that what you mean?

Drezil:
Oh, i was not aware that this is the supposed syntax for creating AST-divs in pandoc-markdown as it is plain html and nothing "markdown-y".
According to the ["original"](https://daringfireball.net/projects/markdown/syntax#html) html should just remain plain html inside markdown and not be handled in a special way.

One thing that bothers me is that this adds redundancy for the attribute-syntax `{#id .class key=value}` that is present in everything else.. (link, span, image, ..)

With that knowledge the PR is for me personally from "urgent" down to "would be nice to have it soon".

mb21:
> "would be nice to have it soon"

Indeed, and that's why it has been "soon" for 6 years now..

Drezil:
Would it be ok to merge this as an extension? I could rewrite it so that "soon" is not another 6 years... :/

Another small thing for http://pandoc.org/MANUAL.html#non-pandoc-extensions probably?

jgm:
You picked one of the many syntax ideas in that thread.  I'm not sure why you picked that one, but I myself wouldn't favor any syntax that makes you write `div`.

I think there are also some potential syntax issues, assuming we allow "laziness" for the side-marked version  (which we probably should since blockquotes allow it).  With laziness,

```
; --- div {.foo}
Lorem ipsum dolor sit amet, consectetur adipiscing
elit, sed do eiusmod tempor incididunt ut labore et
dolore magna aliqua. Ut enim ad minim veniam.
```

by itself is equivalent to

```
; --- div {.foo}
; Lorem ipsum dolor sit amet, consectetur adipiscing
; elit, sed do eiusmod tempor incididunt ut labore et
; dolore magna aliqua. Ut enim ad minim veniam.
```

But then that creates a kind of ambiguity in cases like

```
; --- div {.foo}
Lorem ipsum dolor sit amet, consectetur adipiscing
elit, sed do eiusmod tempor incididunt ut labore et
dolore magna aliqua. Ut enim ad minim veniam.

; --- div {.bar}
Lorem ipsum dolor sit amet, consectetur adipiscing
elit, sed do eiusmod tempor incididunt ut labore et
dolore magna aliqua. Ut enim ad minim veniam.
; ---                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                            
; ---  
```

Anyway, this isn't the place for discussion of the syntax; that should take place on the issue itself.  I'm just explaining why I'm not keen to merge this at this point, even as an extension.

jgm:
I should also add that tests of the new feature would be a necessity before I'd even consider merging a PR like this.  (I'm not saying to add tests, though, because as mentioned above I have too many reservations about the syntax you're implementing.)

tarleb:
I'm closing this, as a different syntax has been chosen; #168 is closed.
Thank you for contributing! 👍

PR #3671, opened 2017-05-15
Added support for horizontal spacing in LaTeX

WUUUGI:
Latex has very good support for [Horizontal Spacing](https://tex.stackexchange.com/questions/74353/what-commands-are-there-for-horizontal-spacing). Pandoc misses this support. Following should be converted:

```latex
e.\,g.
```

Following html code contains the [six-per-em space](https://en.wikipedia.org/wiki/Whitespace_character#Unicode)

```html
<p>e. g.</p>
```


jgm:
Thanks!

PR #3676, opened 2017-05-16
Txt2Tags parser: newline is not indentation

labdsf:
space parses '\n', while spaceChar parses only ' ' and '\t'


jgm:
Good catch

PR #3677, opened 2017-05-16
Move anyLineNewline to Parsing.hs

labdsf:


src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -614,7 +614,7 @@ hrule = try $ do
 --
 
 indentedLine :: PandocMonad m => MarkdownParser m String
-indentedLine = indentSpaces >> anyLine >>= return . (++ "\n")
+indentedLine = indentSpaces >> anyLineNewline >>= return
```
tarleb: The final `return` can be dropped as well

labdsf: Thanks, updated the patch

Issue #3663, opened 2017-05-12
Add EOL writer option


tajmone:
I've seen that #2097 already mentions the problem of different EOL in pandoc's output with different OSs', but I'd like to make it a formal new feature request here...

Add a writers option to enforce an EOL: either Windows style (`CR`+`LF`), Linux style (`LF`), or Mac style (`CR`). Something on the lines of:

    --eol=[crlf | lf | cr]

Being optional, this new feature won't harm any current users. Furthermore I suggest that the default EOL behavior isn't changed — as @jcm mentioned in #2097, using the OS default EOL might be what most users expect in ordinary usage. But for cross-platform projects, this option would allow enforcing a standard.

I'm not sure if this is relevant to all output formats, but it surely affects markdown, html, and possibly other formats as well — especially inside Git controlled repos, where you get spurious files changes when rebuilding docs in a different OS from the one used to push the contents (ie: contents are the same, eccept for EOL style).

I know that Git allows settings (even on a per-repository, and per file extensions basis) to control how text files EOL are committed. But frankly, being able to control pandoc's EOL from command line options would be much better in projects with automated scripting for docs generation.

PR #3680, opened 2017-05-17
implemented #3663

Drezil:
#3663 sounded reasonable for me so i just implemented it because it was mostly search/replace in the UTF8-Module.

Sadly Mac (cr) is not in System.IO in GHC. Otherwise it should be caught in https://github.com/jgm/pandoc/compare/master...Drezil:issue3663?expand=1#diff-5da4b57212f9458ea23f6f27f62bfd87R971

Also: Errors and other things now get rendered with `\r\n` under windows (due to the usage of `nativeNewline` in those calls) and so the console-output should look better there.

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -65,7 +65,7 @@ import System.Directory (Permissions (..), doesFileExist, findExecutable,
 import System.Environment (getArgs, getEnvironment, getProgName)
 import System.Exit (ExitCode (..), exitSuccess)
 import System.FilePath
-import System.IO (stdout)
+import System.IO (stdout, nativeNewline, Newline(..))
 import System.IO.Error (isDoesNotExistError)
 import Text.Pandoc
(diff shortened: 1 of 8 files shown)


jgm:
How are templates dealt with?  If your templates are stored with LF line breaks, and you select CRLF line breaks for output, do you get a mix of different line break styles, or does everything go CRLF?

jgm:
I answered my own question.

Thanks for this!  I've made a few changes and I'll push this shortly.

Changes:

- instead of changing types of functions exported from Text.Pandoc.UTF8, I now export new functions parameterized on Newline, leaving the old ones as they are. This minimizes API change disruption.

- Added to manual.

- Fixed help text for --eol option.

jgm:
Closing this, as I've now merged a rebased and slightly revised version.

tajmone:
Thanks for implementing this feature @Drezil !

I saw it with some delay because my Internet connection didn't work for a few days.

PR #3683, opened 2017-05-18
Added support for textcquotes in LaTeX Reader

WUUUGI:
**Textcqoutes** (from the [csqoutes](https://www.ctan.org/pkg/csquotes) package) allows  any arbitrary piece of _text_ to be enclosed in quotation marks, additionaly the _cite_ argument specifies the citation.

```latex
\textcquote[prenote][postnote]{key}[punct]{text}tpunct

\textcquote[See][P. 10]{harry:philosophersstone}{You're a wizard, Harry.}
```
Output:
```markdown
"You're a wizard, Harry." [See @harry:philosophersstone P. 10]
```

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -1424,8 +1424,8 @@ dotextcquote :: PandocMonad m => LP m Inlines
 dotextcquote = do
   c <- citation "autocite" NormalCitation False
   skipopts
-  text <- enquote
-  return . mconcat $ [text, "\160", c]
+  t <- enquote
+  return . mconcat $ [t, "\160", c]
 
 -- tables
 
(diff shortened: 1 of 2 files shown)


jgm:
While we're at it, can we support plain `\textquotes`?
Also, I think that the punctuation incorporating feature of `\textcquotes` must be enabled by an option when the package is loaded.  Perhaps we should check for this option?  After all, many locales don't incorporate punctuation.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -1417,6 +1419,14 @@ complexNatbibCitation mode = try $ do
   return $ cite (c{ citationMode = mode }:cits)
            (rawInline "latex" $ "\\citetext" ++ raw)
 
+ -- \textcquote[<prenote>][<postnote>]{<key>}[<punct>]{<text>}<tpunct>
+dotextcquote :: PandocMonad m => LP m Inlines
+dotextcquote = do
+  c <- citation "autocite" NormalCitation False
+  skipopts
+  text <- enquote
+  return . mconcat $ [text, "\160", c]
```
jgm: Why a nonbreaking space?

jgm:
Note also that punctuation incorporation is usually desirable only for footnote citation styles.  Even in American style, we don't usually incorporate punctuation for parenthetical citations.  It looks like csquotes has some sophisticated support for this -- I just skimmed the relevant sections.

Anyway, I think we should either leave this out of pandoc or support it fully.

There's also a question about why double quotes should be used; some locales will want single quotes for outer punctuation, and no doubt this is selectable too with an option on the csquotes package?

PR #3684, opened 2017-05-19
Mediawiki writer: treat links with unknown schemes like wikilinks

tarleb:
MediaWiki wikis frequently have pages with names like `User:John` which
were interpreted as URIs, thus turning internal links into global links.
This is prevented by also checking whether the scheme of a URI is
frequently used (i.e. is IANA registered or an otherwise well-known
scheme).

### DIFF
--- a/src/Text/Pandoc/Shared.hs
+++ b/src/Text/Pandoc/Shared.hs
@@ -81,6 +81,9 @@ module Text.Pandoc.Shared (
                      openURL,
                      collapseFilePath,
                      filteredFilesFromArchive,
+                     -- * URI handling
+                     schemes,
+                     isURI,
                      -- * Error handling
                      mapLeft,
                      -- * for squashing blocks
@@ -104,7 +107,7 @@ import Data.List ( find, stripPrefix, intercalate )
 import Data.Maybe (mapMaybe)
 import Data.Version ( showVersion )
 import qualified Data.Map as M
-import Network.URI ( escapeURIString, unEscapeString )
+import Network.URI ( URI(uriScheme), escapeURIString, unEscapeString, parseURI )
 import qualified Data.Set as Set
 import System.Directory
 import System.FilePath (splitDirectories, isPathSeparator)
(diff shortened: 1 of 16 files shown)


jgm:
While we're at it, would it be worth changing the list of schemes into a Data.Set.Set?
(faster lookup)

Otherwise, looks good to me!

I wonder if there are other places in the good where we might want this function instead of `isURI`?

src/Text/Pandoc/Shared.hs
```diff
@@ -774,6 +777,44 @@ filteredFilesFromArchive zf f =
...
+  "steam", "svn", "tag", "teamspeak", "tel", "telnet", "tftp", "things",
+  "thismessage", "tip", "tn3270", "tv", "udp", "unreal", "urn", "ut2004",
+  "vemmi", "ventrilo", "view-source", "webcal", "ws", "wss", "wtai", "wyciwyg",
+  "xcon", "xcon-userid", "xfire", "xmlrpc.beep", "xmlrpc.beeps", "xmpp", "xri",
+  "ymsgr", "z39.50r", "z39.50s"
+  ]
+
+-- | Check if the string is a valid URL with a IANA or frequently used but
+-- unofficial scheme (see @schemes@).
+isIanaURI :: String -> Bool
```
jgm: Why don't we just call this `isURI`?  `IANA` is misleading anyway because of the unofficial ones.  And if we call it `isURI` this will encourage us to use it throughout the code base, which I assume we'd want?

tarleb:
I just had a similar thought while looking at my code again
(though I was going for `isWellKnownUri`, which is still misleading).

A [trie] is probably the best performing structure for our use case, but
I'm not sure if it implementing/importing it and rewriting the current
algorithms is worth the effort.  Data.Set seems like the best trade-off.

[trie]: https://en.wikipedia.org/wiki/Trie

tarleb:
The new function is now used everywhere in place of `Network.URI.isURI` and the list of schemes is updated to include all those currently accepted by IANA.

jgm:
Great, thanks.
I'm going to push one slight revision to make sure we allow uppercase versions of the schemes.

PR #3687, opened 2017-05-20
Move indentWith to Parsing.hs

labdsf:


labdsf:
Note that Txt2Tags now counts tabs not as one space, but uses `readerTabStop`. It is probably a good change.

I am trying to merge all those auxiliary functions that are general enough into `Parsing.hs` so I don't have to copy-paste code for Muse reader (PR #3620).

src/Text/Pandoc/Parsing.hs
```diff
@@ -260,6 +261,17 @@ anyLine = do
...
+-- | Parse indent by specified number of spaces (or equiv. tabs)
+indentWith :: Stream [Char] m Char
+           => HasReaderOptions st
+           => Int -> ParserT [Char] st m [Char]
+indentWith num = do
+  tabStop <- getOption readerTabStop
+  if (num < tabStop)
+     then count num (char ' ')
+     else choice [ try (count num (char ' '))
+                 , try (char '\t' >> count (num - tabStop) (char ' ')) ]
```
jgm: What if you have, say, two tabs in a row; `indentWith 8` will fail, but should succeed, right?

labdsf: Right, I will fix it. I just moved the version with bug.

Issue #3432, opened 2017-02-10
list-table support for conversion ReST files


lruzicka:
# list-table RST conversion problem

## Problem description

I wanted to convert some `.rst` files and pandoc reported the following error:

~~~~~~~
pandoc: ignoring unknown directive: list-table "source" (line 351, column 1)
~~~~~~~

I went through the source file and realized that it does not recognize the **list-table** directive that is used by reSt to define a table.

## How does that work.

The table designed by the **list-table** directive look like this:

~~~~~~~
.. list-table:: Frozen Delights!
   :widths: 15 10 30
   :header-rows: 1

   * - Treat
     - Quantity
     - Description
   * - Albatross
     - 2.99
     - On a stick!
   * - Crunchy Frog
     - 1.49
     - If we took the bones out, it wouldn't be
       crunchy, now would it?
   * - Gannet Ripple
     - 1.99
     - On a stick!
~~~~~~~

Where:
* `..list-table::` is the beginning of the table data
* *Frozen Delights!* is the title.
* `:widths:` define how wide each column should be
* `:header-rows:` defines the header row (makes it bold)
* each `*` represents a new row
* each `-` represents a column in that row

This would make the following output:

| Treat         | Quantity | Description                                                     |
|---------------|----------|-----------------------------------------------------------------|
| Albatross     | 2.99     | On a stick!                                                     |
| Crunchy Frog  | 1.49     | If we took the bones out, it wouldn't be,crunchy, now would it? |
| Gannet Ripple | 1.99     | On a stick!                                                     |

## Problem solution

It would be nice, if support for understanding tables in this format could be added. Thanks.

jgm:
This should be pretty easy to implement. However,
I note that the RST reader seems to have a bug in
recognizing this kind of list, which would need to
be fixed too.

jgm:
Here's a test case for the bug mentioned above:
```
% pandoc -f rst
* - a
  - b
* - c
  - d
^D
<ul>
<li><ul>
<li>a</li>
</ul>
- b</li>
<li><ul>
<li>c</li>
<li>d</li>
</ul></li>
</ul>
```

gingerhot:
hi, when I  try to convert a ghc user guide file: 

https://github.com/bgamari/ghc-users-guide/blob/master/content/using-warnings.rst

and I have samiliar error messages:

> [pandoc warning] ignoring unknown role :ghc-flag: in "source" (line 1057, column 63)
[...]

PR #3688, opened 2017-05-21
RST Reader: parse list table directive

k16shikano:


src/Text/Pandoc/Readers/RST.hs
```diff
@@ -770,6 +771,26 @@ tableDirective top _fields body = do
                                   aligns' widths' header' rows'
        _ -> return mempty
 
+listTableDirective :: PandocMonad m => String -> [(String, String)] -> String -> RSTParser m Blocks
+listTableDirective top _fields body = do
+  bs <- parseFromString parseBlocks body
+  title <- do { parseFromString (trimInlines . mconcat <$> many inline) top }
+  let rows = takeRows $ B.toList bs
+      numOfCols = length $ rows !! 0
```
jgm: Note, `!!` is a partial function. This will fail with a runtime exception if `rows` is empty.  This case should be handled safely.

src/Text/Pandoc/Readers/RST.hs
```diff
@@ -770,6 +771,26 @@ tableDirective top _fields body = do
...
+      headerRows = case lookup "header-rows" _fields of
+                     Just x -> read x :: Int
```
jgm: `read` is also partial and will crash at runtime if `x` can't be read as an `Int`. We don't want that.  Check out `safeRead` from `Text.Pandoc.Shared`.

src/Text/Pandoc/Readers/RST.hs
```diff
@@ -770,6 +771,26 @@ tableDirective top _fields body = do
...
+                     Nothing -> 0
+  return $ B.singleton $ Table (B.toList title)
```
jgm: Better to use B.table; this will also avoid the need for the `B.toList`s.

src/Text/Pandoc/Readers/RST.hs
```diff
@@ -770,6 +771,26 @@ tableDirective top _fields body = do
...
+                           (replicate numOfCols AlignDefault)
+                           (replicate numOfCols 0)
+                           (if headerRows > 0 then rows !! 0 else replicate numOfCols [])
```
jgm: Again `!!` is partial.

jgm:
Thanks for the PR!  I've made some comments that should be addressed.

Also, it would be good to add some tests. The easiest way to do this would be to add `test/command/3432.md` (see the other files in that directory for guidance).

You are aware of #3432, right?  Your commit message should indicate that this commit closes that issue.

We should also support the attributes that list tables can take, such as `widths` and `header-rows`; see #3432 for an example, and see the RST docs for a full list.

k16shikano:
Sorry for the mess and thanks for your comments. I'll fix them later.

src/Text/Pandoc/Readers/RST.hs
veloduff: I'm not sure if this is the right place for this question, but do you know when this commit will be merged?  I'm right in the middle of producing critical docs, and the lack of list-table for rst is proving to be a big hindrance.

src/Text/Pandoc/Readers/RST.hs
```diff
@@ -775,17 +775,18 @@ tableDirective top _fields body = do
...
 listTableDirective top _fields body = do
   bs <- parseFromString parseBlocks body
-  title <- do { parseFromString (trimInlines . mconcat <$> many inline) top }
+  title <- parseFromString (trimInlines . mconcat <$> many inline) top
   let rows = takeRows $ B.toList bs
-      numOfCols = length $ rows !! 0
-      headerRows = case lookup "header-rows" _fields of
-                     Just x -> read x :: Int
-                     Nothing -> 0
+      headerRow = case rows of
```
jgm: I'd suggest something like
```
(firstRow, restRows) = case rows of
                 (x:xs) -> (x, xs)
                 [] -> ([], [])
```
Then you don't have to do the `drop 1 rows` below.  (This should be a *tiny* bit more efficient.)

test/command/3432.md
```diff
@@ -0,0 +1,53 @@
...
+<td>If we took the bones out, it wouldn't be crunchy, now would it?</td>
+</tr>
+<tr class="odd">
+<td>Gannet Ripple</td>
+<td>1.99</td>
+<td>On a stick!</td>
+</tr>
+</tbody>
+</table>
+```
```
jgm: This looks great.  I'd just suggest adding a few more tests to cover things like:  no header row, header rows > 1, no widths, empty body or inappropriate body (paragraph rather than list).

jgm: PS. You can add further tests in the same file.

k16shikano: Thanks for the suggestion. I added tests for below cases. 

* widths is "auto"
* without header-rows
* without widths and with header-rows bigger than 1 (expecting that only the first row is treated as a header) 
* with an empty cell (= cell with just a white space)
* with a cell having a bulletlist in it

src/Text/Pandoc/Readers/RST.hs
k16shikano: It actually makes the builder part far more concise! Thanks!

test/command/3432.md
jgm: Great!  I hope you'll contribute more in the future.

PR #3699, opened 2017-05-26
Org: Fix reading of citations before punctuation

herwigstuetz:
This PR fixes parsing of `org-ref` cite-keys that are followed by punctuation: Before, the citation key in 

> From cite:pandoc: ...

would incorrectly be parsed as `pandoc:`. This changes `orgRefCiteKey` to ignore all special characters at the end of a cite-key and the key in the example is read as `pandoc` without the colon at the end.

To do so, I also changed `many1Till p end` to fail if not at least one occurrence of `p` would be read before `end` succeeds. I think this only has an effect if `p` is a prefix of `end`, so this should make it more consistent.
But if the current behaviour of `many1Till` is actually as expected, I will make a separate `many1Till'` and leave the original untouched.

The zip file [org-inline.zip](https://github.com/jgm/pandoc/files/1032690/org-inline.zip) contains an example `org-inline.org` file with some edge cases this PR is trying to address. It also contains the natively exported `.tex` file as well as `pandoc` exported `.tex` file before and after the change. Except for one example at the end, the new `pandoc` exported files match the natively exported ones.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/Inlines.hs
+++ b/src/Text/Pandoc/Readers/Org/Inlines.hs
@@ -687,13 +687,13 @@ mathEnd c = try $ do
   return res
 
 
-enclosedInlines :: PandocMonad m => OrgParser m a
+enclosedInlines :: (PandocMonad m, Show b) => OrgParser m a
                 -> OrgParser m b
                 -> OrgParser m (F Inlines)
 enclosedInlines start end = try $
   trimInlinesF . mconcat <$> enclosed start end inline
 
-enclosedRaw :: PandocMonad m => OrgParser m a
+enclosedRaw :: (PandocMonad m, Show b) => OrgParser m a
             -> OrgParser m b
             -> OrgParser m String
 enclosedRaw start end = try $
--- a/src/Text/Pandoc/Readers/Org/Inlines.hs
+++ b/src/Text/Pandoc/Readers/Org/Inlines.hs
@@ -339,8 +339,16 @@ linkLikeOrgRefCite = try $ do
 -- | Read a citation key.  The characters allowed in citation keys are taken
 -- from the `org-ref-cite-re` variable in `org-ref.el`.
 orgRefCiteKey :: PandocMonad m => OrgParser m String
-orgRefCiteKey = try . many1 . satisfy $ \c ->
-                  isAlphaNum c || c `elem` ("-_:\\./"::String)
+orgRefCiteKey =
+  let citeKeySpecialChars = "-_:\\./," :: String
+      isCiteKeySpecialChar c = c `elem` citeKeySpecialChars
+      isCiteKeyChar c = isAlphaNum c || isCiteKeySpecialChar c
+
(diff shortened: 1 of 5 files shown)


tarleb:
Thank you, this looks good.

I had a quick look around and found people writing `many1Till` both as defined [as it was](https://stackoverflow.com/questions/2732832/) and [as you suggest](http://www.csinaction.com/2012/07/06/many1till-custom-generic-parser-combinator-for-parsec/). I guess it's best to put this into a separate PR.

src/Text/Pandoc/Readers/Org/Inlines.hs
```diff
@@ -339,8 +339,16 @@ linkLikeOrgRefCite = try $ do
 -- | Read a citation key.  The characters allowed in citation keys are taken
 -- from the `org-ref-cite-re` variable in `org-ref.el`.
 orgRefCiteKey :: PandocMonad m => OrgParser m String
-orgRefCiteKey = try . many1 . satisfy $ \c ->
-                  isAlphaNum c || c `elem` ("-_:\\./"::String)
+orgRefCiteKey =
+  let citeKeySpecialChars = "-_:\\./," :: String
+      isCiteKeySpecialChar = \c -> c `elem` citeKeySpecialChars
```
tarleb: Shorter:

    isCiteKeySpecialChar c = c `elem` citeKeySpecialChars

src/Text/Pandoc/Readers/Org/Inlines.hs
```diff
@@ -339,8 +339,16 @@ linkLikeOrgRefCite = try $ do
...
+      isCiteKeyChar = \c -> isAlphaNum c || c `elem` citeKeySpecialChars
```
tarleb: How about

``` haskell
isCiteKeyChar c = isAlphaNum c || isCiteKeySpecialChar c
```

herwigstuetz:
Thanks, I've applied your suggestions.

herwigstuetz:
Regarding `many1Till`: Should I just make a new PR containing [Parsing: Fail `many1Till p end` if `p` doesn't succeed at least once](https://github.com/jgm/pandoc/pull/3699/commits/627b91e9deb530cac8f4805f7c58c9341b17aa58), or should I rather add a new `many1Till'`?

jgm:
I agree that this change to `many1Till` is needed, so go ahead and make the change rather than defining a new function.

jgm:
@herwigstuetz if it wouldn't be too much trouble, I'd propose rebasing your changes into two commits, one for the `many1Till` change, the other for the org reader changes.

Also, I think the commit message of the `many1Till` changes would be more illuminating if it said something like "Check for the end condition before parsing anything", and it would be helpful to give an example of a parser that has different behavior with the old and new definitions.

Thanks!

herwigstuetz:
@jgm I have rebased the commits and tried to make the commit messages more understandable, as you suggested.

jgm:
Great, many thanks!

Issue #863, opened 2013-05-23
Add Support for Vimwiki Syntax


sk8ingdom:
The syntax is defined here: https://code.google.com/p/vimwiki/wiki/Syntax
It's very close to the Google wiki syntax defined here: https://code.google.com/p/support/wiki/WikiSyntax

Is this possible? Are there other vimwiki users that want to take advantage of pandoc??

Thanks!

HeroicAutodidact:
I also want this! Seems close to existing options!

ycpei:
I want this too. It would be helpful for me to use Jekyll with vimwiki. I'm also interested in implementing it if help is needed.

ycpei:
I'm going to start working on this. Has there been any progress on it here or somewhere else?

LarsEKrueger:
In the mean time, here's a kludge: https://github.com/LarsEKrueger/pandoc-vimwiki

It's a pandoc filter to support some of the extra markup.

ycpei:
I have written a Vimwiki reader. It is a preliminary version, and I have listed the progress at the end of this post. I also wrote a simple test. The files are `/test/vimwiki-reader.{wiki,html}`.

I'm rather new to Haskell so any feedbacks are welcome :)

I couldn't find a dev branch so the PR is submitted to master. Please let me know what I can do to get it merged. Thanks.

 progress:
* block parsers:
    * [X] header
    * [X] hrule
    * [X] comment
    * [X] blockquote -- currently only accepting four spaces rather than a mix of spaces and tabs
    * [X] preformatted -- need to implement preformatted with attributes
    * [X] displaymath - a bit buggy
    * [X] bulletlist / orderedlist - a bit buggy with nested lists -- currently not calculating mixed tabs and spaces indentations.
        * [ ] orderedlist with 1., i., a) etc identification.
        * [ ] todo lists -- see https://github.com/LarsEKrueger/pandoc-vimwiki
    * [X] table
        * [ ] centered table -- there does not seem to be a table builder in Pandoc.Builder that accepts attributes
        * [O] colspan and rowspan -- pandoc limitation, see issue #1024
    * [X] paragraph
    * [ ] definition list
* inline parsers:
    * [X] bareURL
    * [X] strong
    * [X] emph
    * [X] strikeout
    * [X] code
    * [X] link
        * [ ] with thumbnails -- pandoc limitation? can't find builder of link with thumbnails
    * [X] image -- to implement images with attributes - same as in preformatted
[...]

PR #3705, opened 2017-05-28
Vimwiki reader

ycpei:
See #863.

This is a preliminary version.

Descriptions and comments about progress etc. are in the issue thread.


src/Text/Pandoc/Parsing.hs
```diff
@@ -952,7 +952,7 @@ readWithM :: (Monad m)
           -> String                   -- ^ input
           -> m (Either PandocError a)
 readWithM parser state input =
-    mapLeft (PandocParsecError input) `liftM` runParserT parser state "source" input
+    mapLeft (PandocParsecError input) `liftM` runParserT parser state "source" input
```
kanishka-azimi: Remove this change?

src/Text/Pandoc/Readers/Vimwiki.hs
```diff
@@ -0,0 +1,398 @@
...
+hrule :: PandocMonad m => VwParser m Blocks
+displayMath :: PandocMonad m => VwParser m Blocks
+para :: PandocMonad m => VwParser m Blocks
+mixedList :: PandocMonad m => VwParser m Blocks
+table :: PandocMonad m => VwParser m Blocks
+blockQuote :: PandocMonad m => VwParser m Blocks
+preformatted :: PandocMonad m => VwParser m Blocks
+
+header = try $ do
+  sp <- many spaceChar
```
kanishka-azimi: Maybe restrict to just space (not tab or carriage return) here?

src/Text/Pandoc/Readers/Vimwiki.hs
```diff
@@ -0,0 +1,398 @@
...
+  sp <- many spaceChar 
+  eqs <- many1 (char '=')
+  spaceChar
+  let lev = length eqs
+  guard $ lev <= 6
+  contents <- trimInlines . mconcat <$> manyTill inline (try $ spaceChar >> (string eqs) >> many spaceChar >> (char '\n')) 
+  attr <- registerHeader ("", (if sp == "" then [] else ["justcenter"]), []) contents
+  return $ B.headerWith attr lev contents
+para = try $ do
+  contents <- trimInlines . mconcat <$> many1 inline
```
kanishka-azimi: maybe a many1till with delimiters similar to TWiki?

src/Text/Pandoc/Readers/Vimwiki.hs
```diff
@@ -0,0 +1,398 @@
...
+     then return mempty
+     else return $ B.para contents
+hrule = try $ do
+  string "----" >> many (char '-') >> newline
+  return B.horizontalRule
+
+comment :: PandocMonad m => VwParser m ()
+comment = try $ do
+  many spaceChar >> string "%%" >> many (noneOf "\n") >> (lookAhead newline)
+  return ()
```
kanishka-azimi: Might be able to simplify such that comment is an line instead of a block level element. This might allow you to delegate matching leading whitespace to the whitespace inline parser. Just guessing, not certain yet.

test/Tests/Old.hs
```diff
@@ -99,6 +99,10 @@ tests = [ testGroup "markdown"
           , test "reader" ["-r", "mediawiki", "-w", "native", "-s"]
             "mediawiki-reader.wiki" "mediawiki-reader.native"
           ]
+        , testGroup "vimwiki"
+          [ test "reader" ["-r", "vimwiki", "-w", "html5", "-s"]
```
jgm: The test output should be "native", not "html5" (that way we only test the reader, not also the HTML writer).
See how it's done above with e.g. mediawiki.

jgm: I see above that `test/.vimwiki-reader.wiki.swp` was committed; undoubtedly this is a mistake.

pandoc.cabal
```diff
@@ -282,6 +283,8 @@ Flag old-locale
 Library
   Build-Depends: base >= 4.7 && < 5,
                  syb >= 0.1 && < 0.8,
+                 split == 0.2.3.2,
+                 HUnit == 1.6.0.0,
```
jgm: Why a dependency on HUnit?
We have this dependency in the test suite; it shouldn't be needed in normal library code.

pandoc.cabal
```diff
@@ -282,6 +283,8 @@ Flag old-locale
 Library
   Build-Depends: base >= 4.7 && < 5,
                  syb >= 0.1 && < 0.8,
+                 split == 0.2.3.2,
```
jgm: Avoid dependencies on exact versions.
If we really need split, let's depend on as wide a range as possible, at the least `>= 0.2 && < 0.3`.

src/Text/Pandoc/Readers/Vimwiki.hs
```diff
@@ -0,0 +1,398 @@
...
+    * [X] `TODO:` mark
+    * [ ] placeholders: %title and %date -> metadata, %template -> template
+--}
+
+module Text.Pandoc.Readers.Vimwiki ( readVimwiki
+                                 ) where
+import Control.Monad.Except (throwError)
+import Control.Monad (guard)
+import Data.Default 
+import Data.Functor.Identity
```
jgm: Do you actually use this?
It's only in base starting with 4.8, which is why some of the travis builds failed.
But I don't see you using `runIdentity` anywhere in the code.

src/Text/Pandoc/Readers/Vimwiki.hs
```diff
@@ -0,0 +1,398 @@
...
+                , blockQuote
+                , preformatted
+                , displayMath 
+                , table
+                , para
+                ]
+  report $ ParsingTrace (take 60 $ show $ B.toList res) pos
+  return res
+
+header :: PandocMonad m => VwParser m Blocks
```
jgm: Stylistically, we prefer to have the type signatures right above the function definitions.

pandoc.cabal
```diff
@@ -283,8 +283,7 @@ Flag old-locale
 Library
   Build-Depends: base >= 4.7 && < 5,
                  syb >= 0.1 && < 0.8,
-                 split == 0.2.3.2,
-                 HUnit == 1.6.0.0,
+                 split >= 0.2 && <= 0.3,
```
jgm: You only use one function from split, `splitOn`.
You could replace `splitOn ":"` with `splitBy (==':')`, using `splitBy` from `Text.Pandoc.Shared`.
Then `split` can be removed from cabal dependencies (generally, we should avoid adding new dependencies unless it's really needed).

jgm:
Your build is failing because on travis we treat warnings as errors.
You can do that too by selecting `--pedantic` with your `stack` invocation.
That will allow you to test locally.

You might also consider running `hlint` which can offer helpful suggestions.

ycpei:
@jgm Thanks for the suggestions.

jgm:
@ycpei let us know when you think this might be ready to merge, and I'll take another look.

ycpei:
@jgm OK I'll let you know. There's a small bug with bullet and ordered lists I need to fix, as well as some missing features to add. None of these will be major changes though.

ycpei:
Not done yet - please ignore this commit.

ycpei:
I think this may be ready to merge now. Let me know if there are any bugs or other problems.

jgm:
test/vimwiki-reader.html seems not to be used; can it be removed?

jgm:
Can you make sure that the code is wrapped to within 80 columns (as requested in CONTRIBUTING.md)?

ycpei:
@jgm test/vimwiki-reader.html was the output of `pandoc -f vimwiki -t html -s --css data/vimwiki.css -o vimwiki-reader.html` as a test easy for a human eye to see that the html output of the pandoc reader is the same as the `vimwiki` plugin. It has been removed now.

I have enforced the 80 column rule.

I also added support for diary links. I only added it now because it relies on configs in the vim plugin of vimwiki. Since there is a default config I'm using that, namely the diary links like `[[diary:yyyy-mm-dd]]` points to `diary/yyyy-mm-dd.html`.

src/Text/Pandoc/Readers/Vimwiki.hs
```diff
@@ -0,0 +1,661 @@
...
+  
+procLink :: String -> String
+procLink s = let (x, y) = divByHash "" s in (procLink' x) ++ y
+
+divByHash :: String -> String -> (String, String)
+divByHash s1 s2
+  | s2 == ""        = (s1, s2)
+  | head s2 == '#'  = (s1, s2) 
+  | otherwise       = divByHash (s1 ++ [head s2]) (tail s2)
+
```
jgm: If the point of this function is to split a string on a prefix not containing `#` and a suffix starting with `#`, why not use the Prelude function `break`?

```haskell
procLink s = procLink' x ++ y
  where (x, y) = break (=='#') s
```

src/Text/Pandoc/Readers/Vimwiki.hs
```diff
@@ -0,0 +1,661 @@
...
+
+listItemContent :: PandocMonad m => VwParser m Blocks
+listItemContent = try $ do
+  w <- option mempty listTodoMarker
+  x <- plainInlineML' w
+  y <- many p2
+  z <- p6
+  return $ mconcat $ x:y++[z]
+p2 :: PandocMonad m => VwParser m Blocks
```
jgm: You should put a blank line between the two function defs.

src/Text/Pandoc/Readers/Vimwiki.hs
ycpei: I didn't know `break` - thanks for pointing it out.

src/Text/Pandoc/Readers/Vimwiki.hs
ycpei: Done.

jgm:
Thanks very much for this contribution.  There may be further problems that appear, but I think this is in good shape to merge.

ycpei:
Thanks very much for merging. I'll keep an eye on issues related to this reader. Also feel free to @ me whenever a problem appears.

Issue #997, opened 2013-09-25
Headers inside <div> tags do not appear in the table of contents (html, texinfo)


joostkremers:
As discussed in this thread:

https://groups.google.com/forum/#!topic/pandoc-discuss/ZOuYDaq3WcQ

headers that are inside `<div>` tags in Markdown input are not included in the table of contents. This happens at least in Texinfo and in HTML output (the latter with the `--toc` flag).

The issue is caused by the new handling of `<div>` tags in Markdown input, which are converted into `Div` elements, causing headers inside them to appear below the top level, while the relevant writers assume that headers are always at the top level.

In Pandoc 1.11, the following Markdown text:

```
% Title
% Author

<div id="main">

# Header 1

Text. Text.

# Header 2

Text. Text.

</div>
```

would yield HTML output in which the TOC appears _outside_ `<div id="main">`, which allowed the body of the text to be styled and placed separately from the TOC. For me, it would be preferable if this behaviour would be restored.

LarsEKrueger:
It seems possible to resort the output of hierarchicalizeWithIds inside hierarchicalize. Would that have any undesirable consequences? If not, I would try to write a sorter and send a PR.

Also, would it hurt to make the matched pattern there from

```
hierarchicalizeWithIds ((Div ("",["references"],[])
                         (Header level (ident,classes,kvs) title' : xs)):ys) =
```

to something more general, i.e. allowing non-empty ids and checking if `references` is just one of the classes? This would allow to apply a style to a part of the generated HTML file.

PR #3712, opened 2017-05-30
Shared.hierarchicalize: fix the "header-in-div" bug (#997)

LarsEKrueger:
I think I found a heuristic that works for divs with and without ids. Since we can't preserve the div at all (that would require a rewrite of either hierarchicalize or the HTML writer, I guess), I reasoned that a div with an id is something special and we should preserve its id at least.

### DIFF
--- a/src/Text/Pandoc/Shared.hs
+++ b/src/Text/Pandoc/Shared.hs
@@ -575,6 +575,9 @@ rebuildTree _ _ = undefined -- These cases should not happen.
 -- The div will be a child of the previous header/section and the header
 -- inside the div will become a child of the previous section, regardless its
 -- depth. The two other functions correct this.
+-- If a header is placed in a div without an identifier, it will keep its own
+-- identifier. If it is placed in a div with an identifier, the header takes
+-- the identifier of the div.
 hierarchicalize :: [Block] -> [Element]
 hierarchicalize blocks = rebuildTree [] $ flatten $ S.evalState (hierarchicalizeWithIds blocks) []
 
@@ -596,6 +599,10 @@ hierarchicalizeWithIds ((Header level attr@(_,classes,_) title'):xs) = do
 hierarchicalizeWithIds ((Div ("",["references"],[])
                          (Header level (ident,classes,kvs) title' : xs)):ys) =
   hierarchicalizeWithIds ((Header level (ident,("references":classes),kvs)
+                           title') : (xs ++ ys))
+hierarchicalizeWithIds ((Div ("",_,[])
+                         (Header level (ident,classes,kvs) title' : xs)):ys) =
+  hierarchicalizeWithIds ((Header level (ident,classes,kvs)
                            title') : (xs ++ ys))
 hierarchicalizeWithIds ((Div (ident,_,[])
                          (Header level (_,classes,kvs) title' : xs)):ys) =
--- a/src/Text/Pandoc/Shared.hs
+++ b/src/Text/Pandoc/Shared.hs
@@ -491,21 +491,10 @@ inlineListToIdentifier =
  where nbspToSp '\160'     =  ' '
        nbspToSp x          =  x
 
-
--- | Convert a hierarchical list of Elements into a list of sections. All
--- blocks remain with their parent sections. If the list starts with blocks,
--- they will be kept at the start.
-flatten :: [Element] -> [Element]
-flatten [] = []
-flatten (b@(Blk _) : es) = b : flatten es
-flatten (Sec d ident attr cont children : es) =
-  Sec d ident attr cont blkChildren : flatten secChildren ++ flatten es
-  where
-  (blkChildren,secChildren) = partition isBlk children
-
-  isBlk :: Element -> Bool
-  isBlk (Blk _) = True
-  isBlk _ = False
+-- | Check if an Element is a Blk. Required for filtering blocks and sections.
+isBlk :: Element -> Bool
+isBlk (Blk _) = True
+isBlk _ = False
 
 -- | Stack of lists of Elements. Each item on the stack is a list of
 -- semi-processed sections.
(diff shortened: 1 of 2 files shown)


jgm:
Can you explain in English what the precise change is?

LarsEKrueger:
In plain English:
If you have a markdown input like
```
# Header 1
## Header 2
<div id="something">
# Header 3
</div>
```
after the change hierarchicalize will create the correct tree
```
[ Sec "Header 1"
  [ Sec "Header 2" [ ]
  ]
, Sec "Header 3" []
]
```

thus solving issue #997.

This is done by two fixes:
1. Handling divs with ids in a similar manner as the "references" block.
2. Correcting the output of hierarchicalize by putting the Secs in a list (depth-first) and then rebuilding the tree from that list. Details of each steps are to be found in the comments.

jgm:
I've got some concerns about this.

(1)  The code needed for this is surprisingly long and complex.

(2)  This code will cause classes on divs to be ignored/discarded (even if there's an empty class on the contained header element) and will cause identifiers on headers to be ignored/discarded in favor of identifiers on containing divs.  Note that the latter could cause breakage of internal links, and the former could cause unexpected styling changes.  Example:
```html
<div id="foo" class="purple">
<h1 id="foobar" class="polkadotted">Title</h1>
</div>
```
Your code converts this to something like
```html
<section id="foo" class="polkadotted">
<h1>Title</h1>
</section>
```
We've lost an identifier, and any link to `#foobar` is broken.  We've also lost the information that this is `purple`.  Perhaps we could solve the class problem by taking the union.  There's no good solution to the identifier problem, though.

(3)  I can see allowing hierarchicalize to look past the divs in
```html
<div id="section1">
<h1>section title</h1>
<div id="subsection-1.1">
<h2>subsection title</h2>
</div>
</div>
```
but I don't see a compelling reason why we should treat
```html
<div id="section1">
<h1>section title</h1>
</div>
<div id="subsection-1.1">
<h2>subsection title</h2>
</div>
```
the same way (most of the complexity in the code comes from this, since once can handle the first case above using small changes in pattern matching).

LarsEKrueger:
You are perfectly right with all the points. Let me address them one by one:

1. 
    1) It might be possible to merge flatten and hierarchicalizeWithIds. I wanted to keep the changes as minimal as possible vs. keeping the resulting code as short as possible. I'll try over the next days to merge them together. Not sure if this more readable.
    2) There is very little I can do about the complexity of rebuildTree. You have to cover the three cases and you have to cover them for the in-between items and the end-of-list items differently. I can try to reduce the code duplication and the `@`s in the pattern, but the number of cases is already minimal.
    3) Half of the patch are comments explaining the logic. I'm not skimping on those.
2.
    1) I'm aware of that, but there is little I can think of besides adding a `Div`-like constructor to Element and rewriting most of the `elementTo<whatever>` functions in *all* writers. A patch like this would have to address the discrepancies between the various writers as well: e.g. `Writer.HTML` uses `[Element]` for both the ToC and the actual document. Thus we need to preserve the Div in `[Element]` to be able to render the body of the document. In contrast, `Writer.Markdown` writes the ToC from `[Element]` and the body of the document from `[Block]`.
    2) Taking union of classes would also be incorrect. You'd get a purple and a polkadotted title, instead of a purple part of the document with a polkadotted title.
3. Divs are allowed to break the hierarchy of headers. Both your cases are valid and should be treated equally. Again, I fail to see an alternative to adding a `Div`-like constructor to `Element` and rewriting the writers. If you see one, please advise. I'm happy to incorporate your suggestions. I do, however, shy away from rewriting 32 full-fledged writers at once. Not for the work, but for the bugs.

jgm:
I still don't see why we should worry about supporting
```html
<div id="section1">
<h1>section title</h1>
</div>
<div id="subsection-1.1">
<h2>subsection title</h2>
</div>
```
To me, this looks like an explicit signal that the h2 is not to be considered as a logical part of the section containing the h1.  And I might choose to use this structure to give myself an exception to the normal div-creation caused by `--section-divs`.  Maybe I want to use an h2 inside a specially marked div for a sidebar, for example.  You wouldn't want that treated like a normal h2.

If we don't need to handle this case, then things get much simpler, it seems to me.

Here's a thought. `hierarchicalize` and the `Section` data structure date back to before we had native Div elements in the AST.  Now that we have Divs, why not rewrite `hierarchicalize` so that it simply inserts Divs as needed?  (Of course, where there is already a Div in the right place, it can just leave this.)  This would simplify things quite a bit in the writers.  And not all of the 32 writers use hierarchicalize:  only ConTeXt, Docbook, EPUB, HTML, JATS, LaTeX, Markdown, OPML, RTS, TEI; LaTeX only uses it for slides, EPUB for toc).

LarsEKrueger:
While I agree that your last example has somewhat fishy semantics, it's still something that could show up and has to be handled. So, we don't get around the complexity, even if it's just for generating an error message.

Regarding your idea: So `hierarchicalize` would become:

```
hierarchicalize :: hierarchicalize :: [Block] -> [Block]
```
and `Div` takes the role of `Sec`? That's still tricky for two reasons:
1. We have to distinguish between system-injected divs (those that take the role of a `Sec`) and user-created divs.
2. We have to store the depth somewhere. I don't think abusing the attribute field for that is even close to a good idea.

My understanding was that `Sec` is something like a hierarchical `Header`. So we could redefine `Header` as

```
  Header Int [Int] Attr [Inline] [Block]
```

with the two additional fields being the section number list [1,4,1] and the children from `Sec`. The parsers would generate them as empty fields first and call `hierarchicalize` at the end to fill them. Your example could not be expressed anymore and would cause an error in `hierarchicalize`. The writers might even get shorter as the ToC generation could be moved completely to Shared (i.e. filter the headers and turn them into lists of lists of bullet points).

This change probably affects probably all readers and writers.

jgm:
Closing because #997 was closed long ago by another fix.

PR #3714, opened 2017-05-31
Odt reader: remove dead code

tarleb:
The ODT reader contained a lot of general code useful for working with
arrows. However, a lot of these utils weren't used and are hence removed.

It might make sense to keep this dead code around if the ODT reader is to
be extended or reworked in the near future, but it's probably cleaner to revert 
this commit selectively if parts of the code become needed again.

cc @MarLinn

### DIFF
--- a/src/Text/Pandoc/Readers/Odt/Arrows/State.hs
+++ b/src/Text/Pandoc/Readers/Odt/Arrows/State.hs
@@ -59,10 +59,6 @@ withState           :: (state -> a -> (state, b)) -> ArrowState state a b
 withState            = ArrowState . uncurry
 
 -- | Constructor
-withState'          :: ((state, a) -> (state, b)) -> ArrowState state a b
-withState'           = ArrowState
-
--- | Constructor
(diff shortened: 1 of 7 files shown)


MarLinn:
TL;DR: Drop it like it's hot.

Well if you look at how the arrows are actually used you can see how I mostly choose them because I didn't know about Applicative back then. There was a lot I learned while and after writing that code. I'm absolutely not proud of it. As I see it today, the best way forward for the reader would be to rewrite it with an existing XML library that's both Applicative and a bit more high-level than the one I used. If I remember correctly the one I used didn't even support encodings or namespaces by itself. This would also allow to drop even more of the generic stuff. But that's a rather big undertaking, therefore it's on my radar, but nothing I expect to come to very soon. And if I do, the arrow stuff won't be necessary. Others are free to keep expanding on the current implementation, but the structure I left behind isn't very friendly, so I guess they'd come to a similar conclusion.
Sorry.

jgm:
Thanks!

jgm:
I love it when we can remove code.

PR #3719, opened 2017-06-02
Add keywords metadata to docx core.xml document properties

iandol:
Hi, I don't know haskell so possibly this is wrong, but DOCX stores keywords in <cp:keywords> in core.xml, and this should be easy to add from the pandoc metadata (I copy and paste the author code). As far as I can tell (no clear documentation, just a few refs), keywords should be separated with a comma.


jgm:
Thanks!

jgm:
Whoops, there's no `docKeywords`.  I will fix this in a subsequent commit.

PR #3723, opened 2017-06-04
Transplant jgm/pandoc-templates#236 to Pandoc repository.

wilx:


jgm:
This also needs an update to MANUAL.txt -- under `Variables for slides` I suppose, though it should be mentioned that this one is only implemented for Beamer.

wilx:
I have documented the variable and its possible values.

jgm:
Thanks!

PR #3727, opened 2017-06-07
Support for \faCheck and \faClose

schrieveslaach:
This PR adds support for some fontawesome commands. 

Check: ✓ 
Close: ✗

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -681,6 +681,9 @@ inlineCommands = M.fromList $
   , ("nohyphens", tok)
   , ("textnhtt", ttfamily)
   , ("nhttfamily", ttfamily)
+  -- fontawesome
+  , ("faCheck", lit "\10003")
+  , ("faClose", lit "\10007")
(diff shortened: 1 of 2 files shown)


jgm:
What LaTeX package enables these?

schrieveslaach:
Sorry, I forget to include the package name. It is [fontawesome](https://www.ctan.org/tex-archive/fonts/fontawesome?lang=en).

jgm:
Looks good.   I'm not sure whether we should be supporting semi-obscure packages like this, but in this case at least there seems little risk of a conflict with the names.

schrieveslaach:
What do mean by “semi-obscure”?

jgm:
+++ schrieveslaach [Jun 11 17 23:51 ]:
>   What do mean by “semi-obscure”?

My point was that we should be cautious about hard-coding
interpretation of macros defined in uncommonly used
packages, since some people may have their own macros with
the same name (or other packages may define them).  Mostly
we try to support core latex + commonly used macro packages.

I have no idea how commonly used fontawesome is, maybe it's
not semi-obscure!

PR #3739, opened 2017-06-15
Relax time and process in preparation for GHC 8.2

niteria:
GHC 8.2 is very likely to ship with process-1.6.0.0
and time-1.8.0.1.

Consult:
https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory


jgm:
Thanks!

PR #3765, opened 2017-06-26
HTML reader: Use the lang value of <html> to set the lang meta value.

bucklereed:


jgm:
Good addition, thanks!

PR #3791, opened 2017-07-09
HTML reader: parse <main> like <div role=main>.

bucklereed:
Doing the reverse transformation in the HTML writer, and probably other writers too, might make sense, but this is a very good semantic match as it is.


bucklereed:
The Travis failure looks to be unrelated to this pull request; HEAD's currently failing, too.

bucklereed:
Whoops, this isn't ready to merge yet anyway; it doesn't deal with closing <p> correctly, and still warns about skipped </main>. I'll fix it up and push a better version.

bucklereed:
I think I fixed everything with that push, so unless there are some concerns I forgot about or didn't know in the first place, I think this should be ready. (I don't know how to get Travis to re-check it now that HEAD is green...)

jgm:
I think if you rebase against current master, it should run the tests again.

jgm:
Thanks!

PR #3800, opened 2017-07-14
Added TikiWiki support.

rlpowell:
Added TikiWiki support, including tests and documentation.

It's probably not like *complete*, but it works pretty well, handles all
the basics (and some not-so-basics).


src/Text/Pandoc/Readers/TikiWiki.hs
```diff
@@ -0,0 +1,663 @@
+{-# LANGUAGE RelaxedPolyRec, FlexibleInstances, TypeSynonymInstances #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE FlexibleContexts #-}
+
+{- |
+   Module      : Text.Pandoc.Readers.TikiWiki
+   Copyright   : Copyright (C) 2017 Robin Lee Powell
+   License     : MIT (but large portions are copied from Text.Pandoc.Readers.TWiki , which is GPLv2)
```
jgm: License needs to be GPLv2 -- that is the standard license for pandoc, and this is derivative from the TWiki reader which is GPLv2.

src/Text/Pandoc/Readers/TikiWiki.hs
```diff
@@ -0,0 +1,663 @@
...
+  when (verbosity >= INFO) $ do
+    trace (printf "line %d: %s" (sourceLine pos) (take 60 $ show $ B.toList res))
+  return res
+
+blockElements :: PandocMonad m => TikiWikiParser m B.Blocks
+blockElements = choice [ -- separator
+                       table
+                       , hr
+                       , header
+                       -- , verbatim
```
jgm: Are these commented out parts TWiki features that don't exist in TikiWiki (in which case they should just be deleted) or TikiWiki features that you haven't implemented yet?  In the latter case, how hard would it be to implement them and get something more complete?

src/Text/Pandoc/Readers/TikiWiki.hs
rlpowell: *grumble*notabiggplfan*grumble*

Done.

src/Text/Pandoc/Readers/TikiWiki.hs
rlpowell: They are one of:

- already implemented under a different name ("hr" vs. "separator")
- don't exist in tikiwiki ("noautolink")
- are so badly supported in TikiWiki that I won't touch them, and you shouldn't either (blockquotes)

They have all been deleted.

rlpowell:
Sorry for the delay; wanted to make sure my *own* tests (in my hakyll blog repo) passed.  :)  Just fixed up so that the tests I added will all pass now (I somehow dropped the final closing ]).

jgm:
Great, thanks!

PR #3817, opened 2017-07-26
Sorted the list of supported HTML5 attributes and removed duplicates.

Wandmalfarbe:
The list of supported HTML5 attributes introduced in commit https://github.com/jgm/pandoc/commit/d441e656db576f266c4866e65ff9e4705d376381 had lots of duplicates in it.


jgm:
That's very helpful, thanks!

PR #3823, opened 2017-08-01
[WIP] Parse siunitx num command

schrieveslaach:
I want to add support for the `\num[〈options〉]{〈number〉}` command of the [siunitx](http://ftp.dante.de/tex-archive/macros/latex2e/exptl/siunitx/siunitx.pdf) package.

```latex
\num{123}     \\
\num{1234}    \\
\num{12345}   \\
\num{0.123}   \\
\num{0,1234}  \\
\num{.12345}  \\
\num{3.45d-4} \\
\num{-e10}
```

Current code does not work yet. I need some help in order to parse the numbers correctly.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -600,6 +600,11 @@ dosiunitx = do
                       emptyOr160 unit,
                       unit]
 
+dosinum :: PandocMonad m => LP m Inlines
+dosinum = do
+  skipopts
+  tok
+
 lit :: String -> LP m Inlines
 lit = pure . str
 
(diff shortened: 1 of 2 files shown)


jgm:
Instead of having these numbers appear as `Str` elements, as in your test output, why not make them `Math` elements -- then you can use LaTeX math constructs for superscripts, etc.

ickc:
Note that this might not be a good behavior for some people though. I didn't realized pandoc's LaTeX reader is already doing something to SIUnitX's environment (where this commit adds more).

The reason is, because MathJax's third party extension (already on MathJax's official distribution with a custom config) includes support of SIUnitX. So if one want to say, convert LaTeX to HTML with mathjax, this kind of behavior actually does more harm than good.

Also, not to mention LaTeX round trip.

Might be a better approach is for the LaTeX reader to read these to math equation first, and the writers to deal with this particular case.

PR #3844, opened 2017-08-08
Muse writer: insert two blanklines between lists of the same type

labdsf:


jgm:
I merged this without looking at the test failure.
Looks like you need to update the muse writer test results (or the code).

jgm:
```
       basic:                                                     FAIL (0.78s)
          
          --- writer.muse
          +++ /Users/jgm/src/pandoc/.stack-work/dist/x86_64-osx/Cabal-1.24.2.0/build/pandoc/pandoc --quiet --data-dir ../data testsuite.native -r native -w muse --columns=78 --variable pandoc-version= -s
          + 317 foo
          + 318 And nested without indentation:
          - 317 fooAnd nested without indentation:
          + 322 bar
          + 323 Interpreted markdown in a table:
          - 321 barInterpreted markdown in a table:
          + 385 foo
          + 386 This should just be an HTML comment:
          - 383 fooThis should just be an HTML comment:
```

labdsf:
I just updated test results, just a bit late. Will submit another PR then.

`<div>`s are now on separate line because I always insert at least "blanklines 0" between block elements.

PR #3848, opened 2017-08-17
slidy uses https instead of http

ickc:
The reason behind this pull request is that if I deliver my presentation over https, the browsers (tested on Safari and Chrome) will block the access of non-https files like the http css and js used in slidy.

Edit: although `slidy-url` can be used to change this manually, but I think defaulting to https is good practice enough for a pull request.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1320,7 +1320,7 @@ including all [reveal.js configuration options].
 
 `slidy-url`
 :   base URL for Slidy documents (defaults to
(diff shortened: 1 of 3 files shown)


ickc:
By the way, I checked briefly and it seems that this is the only such instance http is used.

And I checked the js and css urls generated, <https://www.w3.org/Talks/Tools/Slidy2/scripts/slidy.js>, <https://www.w3.org/Talks/Tools/Slidy2/styles/slidy.css> works.

jgm:
Makes sense!
Thanks.

PR #3855, opened 2017-08-20
forward $lang$ and $subtitle$ attributes

getreu:
$lang$ is needed for non-English documents.
Many articles and books have a $subtitle$.

### DIFF
--- a/data/templates/default.docbook5
+++ b/data/templates/default.docbook5
@@ -11,7 +11,9 @@ $if(lang)$
 $endif$ >
   <info>
     <title>$title$</title>
+$if(subtitle)$
     <subtitle>$subtitle$</subtitle>
+$endif$
 $if(author)$
     <authorgroup>
 $for(author)$
--- a/data/templates/default.docbook5
+++ b/data/templates/default.docbook5
@@ -6,8 +6,12 @@ $if(mathml)$
   xmlns:mml="http://www.w3.org/1998/Math/MathML"
 $endif$
   xmlns:xlink="http://www.w3.org/1999/xlink" >
+$if(lang)$ 
+      xml:lang="$lang$"
+$endif$ >
   <info>
     <title>$title$</title>
+    <subtitle>$subtitle$</subtitle>
 $if(author)$
     <authorgroup>
 $for(author)$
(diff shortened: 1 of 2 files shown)


data/templates/default.docbook5
```diff
@@ -6,8 +6,12 @@ $if(mathml)$
   xmlns:mml="http://www.w3.org/1998/Math/MathML"
 $endif$
   xmlns:xlink="http://www.w3.org/1999/xlink" >
+$if(lang)$ 
+      xml:lang="$lang$"
+$endif$ >
```
jgm: This doesn't look quite right - this attribute will be placed outside the `>` in line 8, and we'll get an additional `>`.
Also, these attributes should line up.

data/templates/default.docbook5
```diff
@@ -6,8 +6,12 @@ $if(mathml)$
...
   <info>
     <title>$title$</title>
+    <subtitle>$subtitle$</subtitle>
```
jgm: Perhaps this should be made conditional, so we don't get an empty `<subtitle>` element if there's no subtitle?

jgm:
See comments; note also that writer tests will need to be updated when there is a template change.

jgm:
And thank you for the PR!  I see you are a first-time contributor.
To update the writer tests, it suffices to modify `test/writer.docbook5`.
Should the docbook4 template also be updated?

data/templates/default.docbook5
getreu: Why not! But an empty subtitle should not harm either.
I made it optional.

data/templates/default.docbook5
getreu: Yes, sorry. Corrected.

getreu:
> To update the writer tests, it suffices to modify test/writer.docbook5.

Ok, needs some time.

> Should the docbook4 template also be updated?

Maybe yes, someone could do it. I do not have a docbook 4.5 toolchain to test.

getreu:
Everything seems to be OK now.

jgm:
Great, I've merged it, thanks.  I may make one slight change, moving the lang attribute up so we don't have that lonely close `>` on a line by itself.

getreu:
Yes, the xml will look better.

PR #3869, opened 2017-08-23
Name change OSX -> macOS

ickc:
fix commit c96b64e

This commit finishes remaining osx to macOS change, as well as replacing MacOS with macOS.

The reason for the later one is because the "correct" casing of macOS is like that. Apple styles it to looks like iOS, watchOS, tvOS, etc. And unfortunately they all start with a lowercase letter, making propercasing (or even title-casing) odd.

Edit: I add 1 more commit to fix the casing of Linux, UNIX, and Windows when they are addressed as the name of the OSes. (I noticed them when they are addressed alongside macOS.)


jgm:
Great, thanks

PR #3871, opened 2017-08-24
Generalize Text.Pandoc.Parsing.lineBlockLines

labdsf:
I need the same function with '>' for Muse reader


labdsf:
Found out I don't need continuations, better do my own implementation

PR #3873, opened 2017-08-24
Allow setting Japanese fonts when using LuaLaTeX

wilx:
...by using the `luatexja-fontspec` and `luatexja-preset` packages. Use
existing `CJKmainfont` and `CJKoptions` template variables. Add
`luatexjafontspecoptions` for `luatexja-fontspec` and `luatexjapresetoptions`
for `luatexja-preset`.

This resolves jgm/pandoc-templates#227.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -63,10 +63,21 @@ $else$
   \setmathfont[$for(mathfontoptions)$$mathfontoptions$$sep$,$endfor$]{$mathfont$}
 $endif$
 $endif$
+  \ifxetex
 $if(CJKmainfont)$
     \usepackage{xeCJK}
     \setCJKmainfont[$for(CJKoptions)$$CJKoptions$$sep$,$endfor$]{$CJKmainfont$}
 $endif$
+  \fi
+  \ifluatex
+$if(luatexjapresetoptions)$
+    \usepackage[$for(luatexjapresetoptions)$$luatexjapresetoptions$$sep$,$endfor$]{luatexja-preset}
+$endif$
+$if(CJKmainfont)$
+    \usepackage[$for(luatexjafontspecoptions)$$luatexjafontspecoptions$$sep$,$endfor$]{luatexja-fontspec}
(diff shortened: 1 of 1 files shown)


data/templates/default.latex
```diff
@@ -63,10 +63,21 @@ $else$
...
+  \ifluatex
+$if(luatexjapresetoptions)$
+    \usepackage[$for(luatexjapresetoptions)$$luatexjapresetoptions$$sep$,$endfor$]{luatexja-preset}
+$endif$
+$if(CJKmainfont)$
+    \usepackage[$for(luatexjafontspecoptions)$$luatexjafontspecoptions$$sep$,$endfor$]{luatexja-fontspec}
+    \setmainjfont[$for(CJKoptions)$$CJKoptions$$sep$,$endfor$]{$CJKmainfont$}
+$endif$
+  \fi
 \fi
```
jgm: Note, this way of doing it gives you
```
\ifxetex
\fi
\ifluatex
\fi
```
when the CJKmainfont and luatexjapresentoptions variables are not set (as is the default).
It would be better to make all of this conditional on CJKmainfont.  (Or you might need two conditionals, one with CJKmainfont, one with luatexjapresetoptions.)

Note also that the writer tests may need updating (though the need for this will probably go away if you make these things conditional, since the writer tests don't set these variables).

jgm:
Great. Should the same changes be made to default.beamer?

jgm:
Actually, I see that there are quite a few differences between default.beamer and default.latex. Perhaps we should try to integrate the two templates into one template with some conditionals.  I can't recall why we had two in the first place.

wilx:
I am not sure. There seem to be various issues or were at least in the past with Beamer and fonts. I am not confident enough to modify the Beamer template. I do not use it.

adunning:
Thanks for sorting this!

The reason for the separate Beamer template is that the package behaves a bit differently from the basic LaTeX document classes; most notably, it loads `hyperref` independently. It would certainly be possible to merge the two templates together, but the result would be more complex.

Issue #3877, opened 2017-08-25
Colorlinks option failing when producing a beamer presentation


dpuleri:
Here is the version I am running on macOS Sierra
```
pandoc -v
pandoc 1.19.2.1
Compiled with pandoc-types 1.17.0.5, texmath 0.9, skylighting 0.1.1.4
Default user data directory: /Users/daniel/.pandoc
Copyright (C) 2006-2016 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
```

When using beamer and the colorlinks option the following error is produced:
```
! Package xcolor Error: Undefined color `Maroon'.

See the xcolor package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.77 \end{frame}

pandoc: Error producing PDF
```

I realize there is a previous issue (#2680) that discusses this, but I believe that was in regard to the regular LaTeX output and not with the beamer option enabled.

Reproducible with:
```
pandoc -t beamer -V colorlinks -o test.pdf << EOT
# This is a test

So [This is a test].

[Test link](http://test.com).

EOT
```

A quick fix to get the command to run is by specifying the colors manually through the `linkcolor` and `urlcolor` variables:
```
pandoc -t beamer -V colorlinks -V linkcolor=blue -V urlcolor=blue -o test.pdf << EOT
# This is a test

So [This is a test].

[Test link](http://test.com).

EOT
```

However, there seems to be a limited amount of colors you can use through this option. The only ones that have worked for me are blue, white, and black. No fancy colors :frowning_face:.

adunning:
Good catch. The problem is that we're passing the options for colour names to the `color` package, but Beamer loads `xcolor` instead of its own accord and overrides this. For now, you can fix it by adding `-V classoption=dvipsnames`, but I will fix this in the template.

NikosAlexandris:
Is it the version of pandoc I use that still gives this error when, for example, using 
```
urlcolor: blue
```

in the header of a markdown - beamer presentation?

I use the solution provided by @dpuleri to work-around this error (i.e. using `-V colorlinks -V linkcolor=blue -V urlcolor=blue` in the  command line).

PR #3879, opened 2017-08-26
Combine LaTeX/Beamer templates

### LINKED ISSUE #3878: Synchronize or combine LaTeX/Beamer templates
Some of the recent changes to the LaTeX template need to be merged into the Beamer template, notably the use of `unicode-math` and `luatexja`.

Before working through this, would it indeed make sense to determine whether the two templates should be combined, as suggested in #3873? I haven't tested it, but it should theoretically be possible if the writer set a `beamer` variable to true and several conditionals were included to include and exclude several parts. This would make the template more complex to customize, but it would also make it easier to maintain within the codebase.

adunning:
In this template, the sections needed by Beamer (and that need to be taken out) are taken care of using conditionals with the `beamer` variable, already set by the writer as it turns out. Closes #3878.

This is meant to show what this looks like in practice, to facilitate a decision on whether the simplified internal maintenance outweighs the complexity added to the template. It works with some sample documents I have to hand, but I haven't tested it extensively with the Beamer-specific variables.

Should more conditionals be added to prevent variables not applicable to Beamer from being passed on (`linestretch`, `geometry`, `lof`, `lot`)?


adunning:
This also now closes #3877. While the `dvipsnames` option could be sent along for Beamer only to avoid loading `xcolor`, this ensures consistency and allows for a wider selection of colours with `svgnames`.

jgm:
We should incorporate the changes in commit 732005456e2b28150943a5a4e11bca6e1566f309, if it still seems a good idea to merge the two templates.

adunning:
Done. After using it for a couple of weeks, my feeling is that the changes don't unnecessarily increase the complexity of the template, and indeed merging the files makes it much easier to maintain a custom template if one uses both standard LaTeX and Beamer.

jgm:
Thanks for doing this!

adunning:
You're most welcome; thanks for the idea!

PR #3881, opened 2017-08-28
RST reader: treat preceding and following blank lines equally in line blocks

labdsf:
Test is checked to be consisted with http://rst.ninjs.org/


jgm:
Thanks.  I'm puzzled as to why I wrote it like that in the first place; there might be a reason, but I can't think of one now.

Issue #3876, opened 2017-08-25
Setup.hs broken by Cabal 2.x


peti:
I cannot compile pandoc-1.19.2.1's `Setup.hs` file with GHC 8.2.1 because of the following issue:

~~~
Setup.hs:42:3: error:
    • Couldn't match expected type ‘ComponentLocalBuildInfo
                                    -> PreProcessor’
                  with actual type ‘PreProcessor’
    • In the expression:
        PreProcessor
          {platformIndependent = True,
           runPreProcessor = mkSimplePreProcessor
                               $ \ infile outfile verbosity
                                   -> do let ...
                                         ....}
      In the expression:
        \ _ lbi
          -> PreProcessor
               {platformIndependent = True,
                runPreProcessor = mkSimplePreProcessor
                                    $ \ infile outfile verbosity -> ...}
      In the expression:
        ("hsb", 
         \ _ lbi
           -> PreProcessor
                {platformIndependent = True,
                 runPreProcessor = mkSimplePreProcessor
                                     $ \ infile outfile verbosity -> ...})
   |
42 |   PreProcessor {
   |   ^^^^^^^^^^^^^^...

Setup.hs:45:39: error:
    Data constructor not in scope: FlagName :: [Char] -> FlagName
   |
45 |       do let embedData = case lookup (FlagName "embed_data_files")
   |                                       ^^^^^^^^
~~~

mb21:
have you tried with current master?

jgm:
We're focused on 2.0, which will build with Cabal 2.0.

As for backporting a build fix to 1.19.x, I'm not sure how to do it.  Is there a way to make sections in the Setup.hs file conditional on the Cabal version used?  Otherwise I don't know how to deal with the API change.  Or maybe we just add `< 2.0` to Cabal-Version, if that can be done?

peti:
> Or maybe we just add `< 2.0` to Cabal-Version, if that can be done?

I believe that restriction belongs into the setup-dependencies stanza.

jgm:
+++ Peter Simons [Sep 02 17 11:37 ]:
>     Or maybe we just add < 2.0 to Cabal-Version, if that can be done?
>
>   I believe that restriction belongs into the setup-dependencies stanza.

The problem is that if you use that stanza, a very recent
version of Cabal is required.  That is why I've resisted
adding it; I'd like pandoc to be buildable with ghc 8.2.1+
with the build tools that come with the compiler.

[...]

PR #3899, opened 2017-09-06
Fix build with cabal 2.x

peti:
- Setup.hs does not compile with Cabal 2.x, so we require an earlier version
  via setup-depends.

- The following packages need newer versions with GHC 8.2.1 and had their
  constraints relaxed accordingly: executable-path, process, syb, and time.

Resolves https://github.com/jgm/pandoc/issues/3876.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -251,16 +251,20 @@ Flag old-locale
   Description:   Use old-locale and time < 1.5
   Default:       False
 
+custom-setup
+  setup-depends: base,
+                 Cabal < 2
+
 Library
   Build-Depends: base >= 4.7 && <5,
-                 syb >= 0.1 && < 0.7,
+                 syb >= 0.1 && < 0.8,
                  containers >= 0.1 && < 0.6,
                  unordered-containers >= 0.2 && < 0.3,
                  array >= 0.3 && < 0.6,
                  parsec >= 3.1 && < 3.2,
                  mtl >= 2.2 && < 2.3,
                  filepath >= 1.1 && < 1.5,
-                 process >= 1 && < 1.5,
+                 process >= 1 && < 1.7,
                  directory >= 1 && < 1.4,
                  bytestring >= 0.9 && < 0.11,
                  text >= 0.11 && < 1.3,
@@ -298,7 +302,7 @@ Library
      Build-Depends: old-locale >= 1 && < 1.1,
                     time >= 1.2 && < 1.5
   else
-     Build-Depends: time >= 1.5 && < 1.7
+     Build-Depends: time >= 1.5 && < 1.9
   if flag(network-uri)
      Build-Depends: network-uri >= 2.6 && < 2.7, network >= 2.6
   else
@@ -498,14 +502,14 @@ Test-Suite test-pandoc
      Hs-Source-Dirs: prelude
      Other-Modules:  Prelude
   Build-Depends:  base >= 4.2 && < 5,
-                  syb >= 0.1 && < 0.7,
+                  syb >= 0.1 && < 0.8,
                   pandoc,
                   pandoc-types >= 1.17 && < 1.18,
                   bytestring >= 0.9 && < 0.11,
                   text >= 0.11 && < 1.3,
                   directory >= 1 && < 1.4,
                   filepath >= 1.1 && < 1.5,
-                  process >= 1 && < 1.5,
+                  process >= 1 && < 1.7,
                   skylighting >= 0.1.1.3 && < 0.2,
                   Diff >= 0.2 && < 0.4,
                   test-framework >= 0.3 && < 0.9,
@@ -515,7 +519,7 @@ Test-Suite test-pandoc
                   HUnit >= 1.2 && < 1.6,
                   containers >= 0.1 && < 0.6,
                   ansi-terminal >= 0.5 && < 0.7,
-                  executable-path >= 0.0 && < 0.1,
+                  executable-path >= 0.0 && < 0.4,
                   zip-archive >= 0.2.3.4 && < 0.4
   Other-Modules:  Tests.Old
                   Tests.Helpers
@@ -553,7 +557,7 @@ benchmark benchmark-pandoc
      Other-Modules:  Prelude
   Build-Depends:   pandoc,
                    base >= 4.2 && < 5,
-                   syb >= 0.1 && < 0.7,
+                   syb >= 0.1 && < 0.8,
                    criterion >= 1.0 && < 1.2
   Ghc-Options:   -rtsopts -Wall -fno-warn-unused-do-bind
   Default-Language: Haskell98


jgm:
Two questions:

1.  Does the use of `custom-setup` mean that this requires a very recent version of cabal?  Or is this a non-issue because the stanza will be ignored by versions that don't support `custom-setup` (and these versions will be linked against Cabal < 2 anyway)?

2.  Is there any way to make the Setup script work with both Cabal < 2 and Cabal 2, e.g. using CPP?  That would be preferable, going forward.  (And if it's not possible, this is going to be a serious pain point for the whole ecosystem.)

peti:
Older versions of Cabal will simply ignore the unknown stanza. Typically, there will be a warning:

    Warning: pandoc.cabal: Ignoring unknown section type: custom-setup

This is not an issue for people who are using a recent version of `cabal-install` with an old compiler, though, since `cabal-install` always uses its matching Cabal version to interpret the Cabal file. So everyone with `cabal-install 1.24.x` or newer won't even see the warning.

The only way I'm aware of to make `Setup.hs` build with both Cabal 1.x and 2.x is to use `cpp` as outlined in https://github.com/jgm/pandoc/issues/3876#issuecomment-327138557.

jgm:
Thanks -- so, if I understand correctly, Cabal 1.24 and
higher set MIN_VERSION_Cabal, and if it's unset, that
means we have Cabal < 1.24?

jgm:
I've merged this in the r1.19.2.2 branch.  If travis doesn't complain, I'll release to Hackage.

ilovezfs:
```
==> cabal update
Downloading the latest package list from hackage.haskell.org
==> cabal sandbox init
Writing a default package environment file to
/private/tmp/pandoc-20170907-2260-ewc3bh/pandoc-1.19.2.2/cabal.sandbox.config
Using an existing sandbox located at
/private/tmp/pandoc-20170907-2260-ewc3bh/pandoc-1.19.2.2/.cabal-sandbox
==> cabal install --jobs=8 --max-backjumps=100000 --only-dependencies
clang: warning: -Wl,-headerpad_max_install_names: 'linker' input unused
clang: warning: argument unused during compilation: '-L/usr/local/lib'
clang: warning: argument unused during compilation: '-L/System/Library/Frameworks/OpenGL.framework/Versions/Current/Libraries'
Resolving dependencies...
cabal: Cannot select only the dependencies (as requested by the
'--only-dependencies' flag), the packages are required by a dependency of one
of the other targets.
Note: when using a sandbox, all packages are required to have consistent
dependencies. Try reinstalling/unregistering the offending packages or
recreating the sandbox.
/usr/local/Homebrew/Library/Homebrew/build.rb (Formulary::FormulaLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/python.rb
/usr/local/Homebrew/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: cabal install --jobs=8 --max-backjumps=100000 --only-dependencies
1. raise
2. ignore
3. backtrace
4. irb
5. shell
Choose an action: 5
When you exit this shell, you will return to the menu.
bash-3.2$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.2.1
bash-3.2$ cabal --version
cabal-install version 2.0.0.0
compiled using version 2.0.0.2 of the Cabal library 
bash-3.2$ 
```
I think something is still not right :(

jgm:
cabal update
    cabal install pandoc-1.19.2.2

works fine for me with cabal 2.

Does it fail for you?  (You seem to be building from a
source directory; are you sure you have the right source?)

jgm:
I take that back, but it failed in a completely different
way.

Building library for hslua-0.4.1..
[1 of 2] Compiling Scripting.Lua.Raw ( dist/build/Scripting/Lua/Raw.hs, dist/build/Scripting/Lua/Raw.o )
[2 of 2] Compiling Scripting.Lua    ( dist/build/Scripting/Lua.hs, dist/build/Scripting/Lua.o )
/usr/bin/ld.gold: fatal error: -pie and -r are incompatible
collect2: error: ld returned 1 exit status
`gcc' failed in phase `Linker'. (Exit code: 1)
cabal: Leaving directory '/tmp/cabal-tmp-25604/hslua-0.4.1'

+++ John MacFarlane [Sep 08 17 11:04 ]:
>   cabal update
>   cabal install pandoc-1.19.2.2
>
>works fine for me with cabal 2.
>
>Does it fail for you?  (You seem to be building from a
>source directory; are you sure you have the right source?)
>

ilovezfs:
@jgm note that it's the `cabal install --only-dependencies` step that fails. Not `cabal install pandoc`, which is OK. This is actually a cabal bug (https://github.com/haskell/cabal/issues/4456), and the only way to work around it effectively (until/unless the cabal bug is fixed) is to get pandoc actually working with Cabal 2, not dependent on Cabal<2.

ilovezfs:
@jgm assuming you do have it working with Cabal 2, here's how to interleave support for both Cabal<2 and Cabal>=2: http://cabal.readthedocs.io/en/latest/developing-packages.html#backward-compatibility-and-custom-setup

jgm:
The dev version of pandoc (to be pandoc 2.0) works fine
with Cabal < 2 and Cabal > 2, without CPP (because I've
simplified Setup.hs).

Since I'm trying to get that version released,
I'm not sure how much effort I'm willing to invest
getting the older version working with Cabal >= 2.
But I'd consider a PR if anyone else wants to do this.

+++ ilovezfs [Sep 08 17 11:28 ]:

jgm:
I think I've fixed it.

jgm:
OK, 1.19.2.3 released, hopefully this one fixes it.

+++ ilovezfs [Sep 08 17 18:13 ]:

ilovezfs:
@jgm thanks! That did the trick. We're now able to build pandoc with GHC 8.2.1 in Homebrew thanks to this fix :) See https://github.com/Homebrew/homebrew-core/pull/17851.

ilovezfs:
@jgm hmm spoke too soon. The `--only-dependencies` thing is fixed but the build ultimately failed during the actual install
```
[92 of 93] Compiling Text.Pandoc.Readers.EPUB ( src/Text/Pandoc/Readers/EPUB.hs, dist/dist-sandbox-2d431bd4/build/Text/Pandoc/Readers/EPUB.o )
[93 of 93] Compiling Text.Pandoc      ( src/Text/Pandoc.hs, dist/dist-sandbox-2d431bd4/build/Text/Pandoc.o )
Preprocessing executable 'pandoc' for pandoc-1.19.2.3..
Building executable 'pandoc' for pandoc-1.19.2.3..
[1 of 2] Compiling Paths_pandoc     ( dist/dist-sandbox-2d431bd4/build/pandoc/autogen/Paths_pandoc.hs, dist/dist-sandbox-2d431bd4/build/pandoc/pandoc-tmp/Paths_pandoc.o )
[2 of 2] Compiling Main             ( pandoc.hs, dist/dist-sandbox-2d431bd4/build/pandoc/pandoc-tmp/Main.o )

pandoc.hs:95:3: error:
    Data constructor not in scope: VERSION_texmath :: [Char]
   |
95 |   VERSION_texmath ++ ", skylighting " ++ VERSION_skylighting
   |   ^^^^^^^^^^^^^^^
cabal: Leaving directory '.'
cabal: Error: some packages failed to install:
pandoc-1.19.2.3-APgsIoRCJF61yUuw2pwnXl failed during the building phase. The
exception was:
ExitFailure 1
/usr/local/Homebrew/Library/Homebrew/build.rb (Formulary::FormulaLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/python.rb
/usr/local/Homebrew/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: cabal install --jobs=8 --max-backjumps=100000 --prefix=/usr/local/Cellar/pandoc/1.19.2.3
```

Full log https://gist.github.com/ilovezfs/4a236cbba7401c62f9c40a772dcfc6cb

ilovezfs:
@jgm same result on Hackage's matrix:
https://matrix.hackage.haskell.org/package/pandoc#GHC-8.2/pandoc-1.19.2.3

ilovezfs:
@jgm did you have a chance to reproduce the build issue?

jgm:
+++ ilovezfs [Sep 10 17 06:04 ]:
>   [1]@jgm did you have a chance to reproduce the build issue?

No, and my attempt at a travis build keeps timing out.

But I gather from the log you linked to that the problem is
that the macros VERSION_texmath, VERSION_skylighting, etc.
are no longer defined by Cabal?  Is this a documented
change?  How to work around it?

ilovezfs:
@jgm hmm I'm wondering if it's a consequence of
```
+#if MIN_VERSION_Cabal(2,0,0)
+ppBlobSuffixHandler = ("hsb", \_ lbi _ ->
+#else
 ppBlobSuffixHandler = ("hsb", \_ lbi ->
+#endif
```

ilovezfs:
@jgm OK, I got some help with it on IRC:
```
7:38 AM <hvr> ilovezfs_: w/o having looked at the code: does the executable depend directly on 'texmath'?
7:39 AM <hvr> if not, it shouldn't assume there's is a VERSION_texmath definition
7:39 AM <hvr> (that was a bug/oversight in previous versions of Cabal which got fixed)
7:40 AM <ilovezfs_> texmath is in the library build-depends:
7:41 AM <hvr> ilovezfs_: that's not enough; dependencies are per-component and this is taken more seriously w/ newer cabal lists
7:41 AM <hvr> libs
```

ilovezfs:
@jgm see https://github.com/jgm/pandoc/pull/3910

jgm:
Got it!  I've added skylighting and texmath to the
build-depends for the executable, and released 1.19.2.4.
Hopefully this will do the trick.

+++ ilovezfs [Sep 10 17 07:43 ]:

ilovezfs:
@jgm yup that did the trick! 1.19.2.4 shipped in Homebrew, built with GHC 8.2.1 and Cabal 2, and --only-dependencies works as expected. Thanks a bunch!

https://github.com/Homebrew/homebrew-core/commit/f0ea5102a1bf13117fefe6dd0f930c04f3e76f84

PR #3905, opened 2017-09-09
Handle skylighting’s new code output

### LINKED ISSUE #3838: line wrap in html breaks `numberLines` code line numbering
When printing html it is useful to wrap lines of code; however the supported option for adding line numbers leads to issues such as the following:

> ``` {.haskell .numberLines}
> extraLongFunctionWithVeryLongNames :: a -> a
> extraLongFunctionWithVeryLongNames = whenTheLineGetsTooLong . itWraps . whenPrinted .  ifTheCSSOption "white-space" . isSetTo "pre-wrap"
> 
> whenTheLineGetsTooLong :: a -> a
> whenTheLineGetsTooLong = id
> ```
> 
> becomes
> 
> line-numbers-misaligned

The fundamental issue is the markdown is converted to the following html

```html
<body>
<div class="sourceCode"><table class="sourceCode haskell numberLines"><tr class="sourceCode"><td class="lineNumbers"><pre>1
2
3
4
5
</pre></td>
<td class="sourceCode"><pre><code class="sourceCode haskell">
<span class="ot">extraLongFunctionWithVeryLongNames ::</span> a <span class="ot">-&gt;</span> a
extraLongFunctionWithVeryLongNames <span class="fu">=</span> whenTheLineGetsTooLong <span class="fu">.</span> itWraps <span class="fu">.</span> whenPrinted <span class="fu">.</span> ifTheCSSOption <span class="st">&quot;white-space&quot;</span> <span class="fu">.</span> isSetTo <span class="st">&quot;pre-wrap&quot;</span>

<span class="ot">whenTheLineGetsTooLong ::</span> a <span class="ot">-&gt;</span> a
whenTheLineGetsTooLong <span class="fu">=</span> id
</code></pre></td>
</tr></table></div>
</body>
```

With `numberLines` the code is inserted into a one row, two column table, with the left column a simple list of newline-separated line numbers. 
Without `numberLines` there is no indication what code corresponds to a line, in the html output — indeed, in the native AST line breaks are represented as `\n` but this information is lost on conversion to html.

I suspect the fix for this would be to rewrite the html writer so that `numberLines` wraps each line in a div, say (this could then be prefixed with a line number using css) or something more similar to github, where the code is presented in a table but each line has its own row.

What is the best approach?

---

Using
```
pandoc 1.19.2.1
Compiled with pandoc-types 1.17.0.5, texmath 0.9.4, skylighting 0.3.3
```

dbaynard:
-   Bump dependency on skylighting (Edit: to >=0.4)
-   Fix tests requiring the old code output (Edit: Note, CI will fail until updated skylighting is available, per dependency bump above.)

This fixes #3838 

The details are discussed in jgm/skylighting#14

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -553,7 +553,7 @@ test-suite test-pandoc
                   filepath >= 1.1 && < 1.5,
                   hslua >= 0.8 && < 0.9,
                   process >= 1.2.3 && < 1.7,
-                  skylighting >= 0.3.3 && < 0.4,
+                  skylighting >= 0.4 && < 0.5,
                   temporary >= 1.1 && < 1.3,
                   Diff >= 0.2 && < 0.4,
                   tasty >= 0.11 && < 0.12,
--- a/test/Tests/Writers/HTML.hs
+++ b/test/Tests/Writers/HTML.hs
@@ -32,7 +32,7 @@ tests :: [TestTree]
 tests = [ testGroup "inline code"
           [ "basic" =: code "@&" =?> "<code>@&amp;</code>"
           , "haskell" =: codeWith ("",["haskell"],[]) ">>="
-            =?> "<code class=\"sourceCode haskell\"><span class=\"fu\">&gt;&gt;=</span></code>"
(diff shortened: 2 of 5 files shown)


dbaynard:
An example

# html

## Old version

old-html

## New version

new-html

# pdf (via prince)

## Old version

old-pdf

## New version

new-pdf

jgm:
I've done this separately, after modifying skylighting so it doesn't put the source line divs in inline HTML output.

PR #3910, opened 2017-09-10
executable: depend on texmath directly

ilovezfs:
to satisfy Cabal >=2, which doesn't like only having the dependency
declared indirectly via the pandoc library.


jgm:
We needed skylighting too; I did this separately, but thanks for the PR.

ilovezfs:
Thanks @jgm :heart:

PR #3926, opened 2017-09-17
List splicing for lua filters

tarleb:
Support splicing of element lists in lua filters.

The `Walkable` instances for `[Block]` and `[Inline]` are added as orphaned instances in *Text.Pandoc.Lua*. They could also be added to *Text.Pandoc.Walk*, but this would require proper implementations of `query`, leading to code duplication and bloat with, as far as I can see, no clear advantage. I was wondering if it could make sense to split `Walkable` and to put query into a separate `Queryable` typeclass?

The `query`-less instances already contain a lot of code duplication. The only way I found to avoid this was by using the IncoherentInstances pragma, which seemed like a bad trade-off.

Side-remark / question: Would it be ok to add Applicative to the signature of `walkM`?

      walkM :: (Monad m, Applicative m, Functor m) => (a -> m a) -> b -> m b

This would allow for some (IMHO) slightly nicer code here and there.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -105,11 +105,22 @@ element filtering function.  In other words, filter entries will
 be called for each corresponding element in the document,
 getting the respective element as input.
 
-The element function's output must be an element of the same
-type as the input. This means a filter function acting on an
-inline element must return an inline, and a block element must
-remain a block element after filter application. Pandoc will
-throw an error if this condition is violated.
+The return of a filter function must one of the following:
+
+-   nil: this means that the object should remain unchanged.
+-   a pandoc object: this must be of the same type as the input
+    and will replace the original object.
+-   a list of pandoc objects: these will replace the original
+    object; the list is merged with the neighbors of the orignal
+    objects (spliced into the list the original object belongs
+    to); returning an empty list deletes the object.
+
+The function's output must result in an element of the same type
+as the input. This means a filter function acting on an inline
+element must return either nil, an inline, or a list of inlines,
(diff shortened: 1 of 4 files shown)


jgm:
I don't mind adding Applicative.

I am worried about the orphan instances, though. I think that's a bad thing.

Can you say more about why adding the new instances of Walk would require doing something unpleasant with query?

jgm:
I suppose that if this is just for use in the lua filters, and the new instances don't get exported, the orphan instances might not be problematic.

tarleb:
The thing that bothered me is that there is a lot of code duplication which I was unable to reduce. E.g., the implementation of `query` for `Walkable [Block] Inline` is exactly the same as that for `Walkable Block Inline`.

Now that you said it, I'm very unsure whether the orphan instances might become problematic for users of pandoc as a library. Code duplication might be the lesser evil; I'll move all instances into Text.Pandoc.Walk if you agree.

jgm:
+++ Albert Krewinkel [Sep 17 17 15:12 ]:
>   The thing that bothered me is that there is a lot of code duplication
>   which I was unable to reduce. E.g., the implementation of query for
>   Walkable [Block] Inline is exactly the same as that for Walkable Block
>   Inline.

I think that's acceptable, if there's no way to simplify.
Go ahead and move it to Text.Pandoc.Walk.  Once it's there
I'll see if I can think of any way to simplify; if not,
we'll live with the duplication.  It will be handy to have
these instances available for other purposes as well.

PR #3935, opened 2017-09-26
WIP: add header colours

bollenn:
Add options to specify table header styles

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -176,6 +176,10 @@ $highlighting-macros$
 $endif$
 $if(tables)$
 \usepackage{longtable,booktabs}
+\newcommand{\BeforeTableStyle}{}
+\newcommand{\TableHeaderRowStyle}{}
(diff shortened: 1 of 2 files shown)


tarleb:
Thank you for this PR and apologies for not answering sooner.

This is nice and clean code. I'm only worried that it might be too specific to a certain use-case. As far as I see, the same result could be achieved via a filter (making heavy use of the `RawInline` element). @ickc maintains a collection of pandoc filters, I believe a filter like this would make a very nice addition.

bollenn:
Would be very interested in such a filter approach... going to check that out this week...

bollenn:
@jgm finally took a look at the pandoc filters, am not sure if that approach is future proof as I would have to completely overwrite the https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/LaTeX/Table.hs in the filter I believe...

PR #3938, opened 2017-09-27
Enable TOC display for DOCX in YAML metadata

mjfs:


mb21:
Haven't looked this through, but is this consistent with https://github.com/jgm/pandoc/commit/75f4e41d7d292e011a83d06efebc356060ea812b ?

mjfs:
Proposed changes are actually implementation of 75f4e41d7d292e011a83d06efebc356060ea812b for DOCX output since `toc` boolean flag in YAML metadata was otherwise ignored within DOCX writter.

jgm:
I don't think it makes sense to try to make `toc: true` in metadata activate the table of contents in a docx, unless we are going to do the same for every format.  Too confusing for users otherwise.

mjfs:
@jgm: I agree that it would be confusing if only one writer would be implementing this feature, but since I was needing this functionality in DOCX outputs it seemed like a logical place for me to start.

In a standalone documents it is a lot more intuitive to enable the table of content display within the document itself than via command line parameter. Since https://github.com/jgm/pandoc/commit/75f4e41d7d292e011a83d06efebc356060ea812b enabled setting `toc` to `true`, I saw this as a great opportunity to enable populating table of content as well.

When choosing standalone mode some (most?) writers already output `toc` placeholders (e.g. HTML) when `toc` is set to `true` in metadata. One could argue that populated table of content in this case is only following the principle of least surprise.

Issue #3940, opened 2017-09-28
Support R filters


andrie:
Pandoc already supports filters in many languages.  However, filter written in R doesn't work, since any `filter.R` should be run by `Rscript` .

I have a fix and will make a pull request shortly.

mb21:
fixed by https://github.com/jgm/pandoc/commit/2f03f389e8df0c52248d0767a8372184d65e4ba5

PR #3941, opened 2017-09-28
Support R filters #3940

andrie:
This triggers `Rscript` when encountering a filter ending in `*.R` or `*.r`.

#3940


jgm:
Thanks.

PR #3946, opened 2017-09-30
Provide make target to update lua module docs

tarleb:
The pandoc module documentation in doc/lua-filters.md was automatically
generated from `data/pandoc.lua`.  A make target is provided which uses
a lua filter to update the documentation.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -1,39 +1,42 @@
-% Pandoc Lua Filters
-% Albert Krewinkel, John MacFarlane
-% August 31, 2017
+---
+author:
+- 'Albert Krewinkel, John MacFarlane'
+date: 'August 31, 2017'
+title: Pandoc Lua Filters
+---
 
 # Introduction
 
 Pandoc has long supported filters, which allow the pandoc
 abstract syntax tree (AST) to be manipulated between the parsing
-and the writing phase.  Traditional pandoc filters accept a JSON
+and the writing phase. Traditional pandoc filters accept a JSON
 representation of the pandoc AST and produce an altered JSON
-representation of the AST.  They may be written in any
+representation of the AST. They may be written in any
 programming language, and invoked from pandoc using the
 `--filter` option.
 
 Although traditional filters are very flexible, they have a
-couple of disadvantages.  First, there is some overhead in
-writing JSON to stdout and reading it from stdin (twice,
-once on each side of the filter).  Second, whether a filter
-will work will depend on details of the user's environment.
-A filter may require an interpreter for a certain programming
-language to be available, as well as a library for manipulating
-the pandoc AST in JSON form.  One cannot simply provide a filter
-that can be used by anyone who has a certain version of the
-pandoc executable.
+couple of disadvantages. First, there is some overhead in
+writing JSON to stdout and reading it from stdin (twice, once on
+each side of the filter). Second, whether a filter will work
+will depend on details of the user's environment. A filter may
+require an interpreter for a certain programming language to be
+available, as well as a library for manipulating the pandoc AST
+in JSON form. One cannot simply provide a filter that can be
+used by anyone who has a certain version of the pandoc
+executable.
 
 Starting with pandoc 2.0, we have made it possible to write
-filters in lua without any external dependencies at all.
-A lua interpreter and a lua library for creating pandoc filters
-is built into the pandoc executable.  Pandoc data types
-are marshalled to lua directly, avoiding the overhead of writing
+filters in lua without any external dependencies at all. A lua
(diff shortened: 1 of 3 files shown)


doc/lua-filters.md
```diff
@@ -62,12 +65,12 @@ pandoc manual, MANUAL.txt, and versions of the same filter
 written in compiled Haskell (`smallcaps`) and interpreted Python
 (`smallcaps.py`):
 
-| Command                                          | Time  |
-|--------------------------------------------------|------:|
-| `pandoc MANUAL.txt`                              | 1.01s |
-| `pandoc MANUAL.txt --filter ./smallcaps`         | 1.36s |
-| `pandoc MANUAL.txt --filter ./smallcaps.py`      | 1.40s |
-| `pandoc MANUAL.txt --lua-filter ./smallcaps.lua` | 1.03s |
+  Command                                               Time
```
jgm: We should probably rerun these timings, since this was done before the change allowing returning a list.  (I think that had a small performance penalty.)

ickc:
Interesting. I think an HTML comment should be added to `doc/lua-filters.md` that warn people that this is generated and so should not be edited (I was about to edit it when I was batch-proofreading stuffs but luckily I didn't find anything to change there).

And I actually don't understand exactly how `doc/lua-filters.md` is generated. It seems that the content of `doc/lua-filters.md` plus `data/pandoc.lua` are used by `tools/update-lua-docs.lua` to update `doc/lua-filters.md`. So in a sense `doc/lua-filters.md` is not really generated? From the code it seems only the section `module-pandoc` is auto-updated? Then may be an HTML comment just before that section is sufficient.

And by the way, why `doc/lua-filters.md` is a phony target in the `Makefile`?

ickc:
By the way, this file has pandoc only markdown extensions. So it seems that similar to `MANUAL.txt`, this should have extension `.txt` to avoid people reading it on GitHub to see a half-rendered document.

This brings to the next question that how to distribute this document? I think it is quite natural to put it in the user guide, perhaps in the appendix. Since I don't see a make target of the user guide in the Makefile, perhaps it was the plan all along. May be we should add the building of the User Guide in the Makefile as well? This will also help people to have a custom build of the User Guide (perhaps they want a different font size / font, etc.).

And then perhaps `MANUAL.txt` should also be put under the `doc/` directory.

tarleb:
Let me start by saying that I'm embarrassed by the current state of this. Updating the 'module pandoc' section programmatically is a cludge. There must be a better way, but I coudnl't quite find it yet.  This is also why I'd not invest too much energy into polishing the current setup.

You are absolutetly correct that only a single section is generated. Adding a comment before that section seems like an exellent idea.

> And by the way, why doc/lua-filters.md is a phony target in the Makefile?

Whenever doc/lua-filters.md is newer than all of its the dependencies, it would not be possible to regenerate it. So if I edit doc/pandoc.lua and *then* edit doc/lua-filters.md, I'd still want the make target to be executed.

ickc:
>Whenever doc/lua-filters.md is newer than all of its the dependencies, it would not be possible to regenerate it. So if I edit doc/pandoc.lua and then edit doc/lua-filters.md, I'd still want the make target to be executed.

That make sense. `-B` could help with that too.

There’s a similar issue in the case of `MANUAL.txt` and `README.md`, where the later one contain excerpts of the former one. I once wrote a hacky make command that uses some UNIX utilities to auto-generates `README.md` but it is too hacky to be accepted.

In your case that you already have a script, you might want to put everything (as a string?) in the script, or as a template besides the script, and generates the whole output, rather than having it depends on itself.

Anyway, this is great work. And it’s only v0.3, so there’s time to polish things. This can even be a good example on using pandoc and lua filter per se (for example the use of to markdown from markdown).

jgm:
+++ Albert Krewinkel [Oct 27 17 00:56 ]:
>   Let me start by saying that I'm embarrassed by the current state of
>   this. Updating the 'module pandoc' section programmatically is a
>   cludge. There must be a better way, but I coudnl't quite find it yet.
>   This is also why I'd not invest too much energy into polishing the
>   current setup.

One option would be to have a separate file, doc/lua-api.md,
which is autogenerated.

We could just concatenate the files when generating the
online documentation, so it wouldn't make any difference
to the public-facing documentation.

jgm:
+++ ickc [Oct 27 17 06:38 ]:
>   By the way, this file has pandoc only markdown extensions. So it seems
>   that similar to MANUAL.txt, this should have extension .txt to avoid
>   people reading it on GitHub to see a half-rendered document.
>
>   This brings to the next question that how to distribute this document?

I plan to include it as a separate document under the
Documentation tab in the website (like the ebook
doc currently).

PR #3951, opened 2017-10-03
Load Google Font using HTTPS by default

greut:
Otherwise they won't show up in current version of firefox/chromium.

https://greut.github.io/slides-devweb/00-presentation.html


jgm:
Thanks!

Issue #3743, opened 2017-06-18
MathJax and reveal.js


dotlambda:
When using MathJax with reveal.js, the output looks rather ugly:
![2017-06-18-020850_1920x1080_scrot](https://user-images.githubusercontent.com/6806011/27257125-9e0aa876-53ba-11e7-9565-f384d0c641bc.png)
This is taken from [example16d.html](https://pandoc.org/demo/example16d.html) from https://pandoc.org/demos.html.

But there is a reveal.js [plugin](https://github.com/hakimel/reveal.js/#mathjax) that generates better output:
![2017-06-18-000217_1920x1080_scrot](https://user-images.githubusercontent.com/6806011/27257138-282c6af8-53bb-11e7-89da-e2aba9b0022d.png)
This was generated using the same command, i.e.
```sh
pandoc -s --mathjax -i -t revealjs SLIDES -o example16d.html
```
but a modified `default.revealjs` template that includes
```js
dependencies: [
  // ...
  { src: '$revealjs-url$/plugin/math/math.js', async: true },
  // ...
]
```
I think that the we should handle the `--mathjax` option this way when used in combination with reveal.js.

jgm:
Thanks, I didn't know about this.

PR #3962, opened 2017-10-09
Use the math plugin in reveal.js template with --mathjax option

ugonj:
Currently, when compiling to a reveal.js presentation with the --mathjax option, a link to the mathjax library is added, in a similar fashion to the other html templates. However, the proper way of using Mathjax with reveal.js is using the math plugin, otherwise the math is not displayed properly.

This is the bug referenced in issue [3743](https://github.com/jgm/pandoc/issues/3743)

This commit contains a trivial patch to the revealjs template to implement the correct behaviour.


ugonj:
nevermind. I patched an old version. Looks like this is already fixed...

PR #3963, opened 2017-10-09
docx writer - activate evenAndOddHeaders from reference doc

### LINKED ISSUE #3901: docx writer - "different odd/even pages" is not selected even if template has it on
First of all thank you for making this amazing tool.

I'm trying to set up a markdown -> docx pipeline, using **pandoc 1.18** on windows 7.

I'm tweaking the docx template (reference-docx), but I just realized that odd/even pages selection (header/footer) on my template is not honored in the generated docx (thus the odd style is applied to the whole document).

**WORKAROUND:**
Thankfully the even/odd styles are not lost and I can manually go to the header of the generated docx and select "different odd/even pages" and I will get the expected result.

Ideally the generated docx should have this option selected if the reference-docx had it.

agusmba:
Fixes #3901 by checking for the evenAndOddHeaders mark in the
reference doc, and copying it to the resulting docx if present.

### DIFF
--- a/src/Text/Pandoc/Writers/Docx.hs
+++ b/src/Text/Pandoc/Writers/Docx.hs
@@ -549,6 +549,7 @@ writeDocx opts doc@(Pandoc meta _) = do
                      , "w:consecutiveHyphenLimit"
                      , "w:hyphenationZone"
                      , "w:doNotHyphenateCap"
+                     , "w:evenAndOddHeaders"
                      ]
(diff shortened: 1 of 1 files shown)


agusmba:
It seems one of the builds got stuck:

> The job exceeded the maximum time limit for jobs, and has been terminated.

jgm:
Thank you!

PR #3965, opened 2017-10-11
FB2 Writer: format LineBlock as poem

labdsf:
Previously writer produced one paragraph with `<empty-line/>` elements,
which are not allowed inside `<p>` according to FB2 schema.

[reStructuredText Markup Specification](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#line-blocks) suggests using line blocks for verse and Muse reader parses <verse> as LineBlocks, so it makes sense to convert LineBlocks to `<poem>` in FB2 writer.

Also Linux FBReader has problems parsing current output and stops parsing `<p>` after first `<empty-line/>`. It was "fixed" in mobile FBReader at my [request](https://github.com/geometer/FBReaderJ/issues/465), but now that I look at XML schema it looks like it is a bug in pandoc. And with new output we get actual line breaks instead of empty lines (two line breaks) on display.


jgm:
Thanks!

PR #3980, opened 2017-10-17
Latex reader: Skip spaces in image options

bfirsh:
Fixes things like this:

```
\includegraphics[width=12cm, height = 5cm]{foo.png}
```


jgm:
Excellent, and thanks for the tests!

PR #3984, opened 2017-10-19
Added a few abbreviations and sorted the list.

Wandmalfarbe:


data/abbreviations
```diff
@@ -1,28 +1,54 @@
...
+Mr.
+Mrs.
+Ms.
+No.
+Nov.
+Oct.
+Okt.
 p.
+Ph.D.
+PhD.
```
tarleb: I've only seen *Ph.D.* or *PhD*, but not *PhD.*. Is this commonly used?

Wandmalfarbe: I don't think it's commonly used so I removed it. I also removed the german abbreviations except `Sept.`.

Wandmalfarbe:
I left `Sept.` in because it is the English abbreviation for September: <https://english.stackexchange.com/questions/408701/abbreviation-for-september>
<http://public.oed.com/how-to-use-the-oed/abbreviations/#s>

I will add `Jun.`, `Jul.`, and `Sep.` for consistency with the other three letter abbreviated months.

tarleb:
Interesting, didn't know. Thanks!

Issue #3488, opened 2017-03-05
LaTeX-to-HTML conversion omits contents of \textsf{}


nschneid:
For example, on the [Try page](http://pandoc.org/try/):

```tex
\newcommand{\fnf}[1]{\textsc{\textsf{#1}}}
\newcommand{\fnr}[1]{\textbf{\textsf{#1}}}
\newcommand{\ggg}[1]{\textsf{#1}}
\newcommand{\lex}[1]{\textit{#1}}

\fnf{Abcd}

\fnr{Efgh}

\ggg{Ijkl}

\textsf{Mnop}

\lex{Zzzzz}
```

converted to HTML results in

```html
<p><span><span style="font-variant: small-caps;"></span></span></p>
<p><span><strong></strong></span></p>
<p><span></span></p>

<p><span><em>Zzzzz</em></span></p>
```

The same holds true for converting a full LaTeX file where these macros are used. (Version 1.19.2.1.)

1. Should `\textsf{}` be supported out-of-the-box?
2. If a LaTeX command is unknown, how should it be handled in HTML?
3. Would it be possible to have an alternate treatment of macros—e.g., specify them as classes in an HTML span, like `<span class="fnr">Efgh</span>`? It could be left to the user to define appropriate CSS.

jgm:
>    1. Should \textsf{} be supported out-of-the-box?

Yes, at least its contents should come through unscathed,
and probably there should be a span with an attribute.

This should be a simple change to the LaTeX reader.

>    2. If a LaTeX command is unknown, how should it be handled in HTML?
>    3. Would it be possible to have an alternate treatment of macros—e.g.,
>       specify them as classes in an HTML span, like <span
>       class="fnr">Efgh</span>? It could be left to the user to define
>       appropriate CSS.

Unfortunately, this wouldn't work well, because many latex
commands have contents that aren't meant to be printed
(e.g. \ref{}, \label{}, many others).

If we defaulted to including all of their contents (in a
span or not), the default output would be very messy.

Pandoc 2.0 will at least give good warnings about things it
is ignoring.

jgm:
I've just made changes that give the following result for your input:
``` html
<p><span><span class="smallcaps"><span class="sans-serif">Abcd</span></span></span></p>
<p><span><strong><span class="sans-serif">Efgh</span></strong></span></p>
<p><span><span class="sans-serif">Ijkl</span></span></p>
<p><span class="sans-serif">Mnop</span></p>
[...]

PR #3988, opened 2017-10-22
LaTeX Reader: support \lettrine

mb21:
Lettrine, e.g. as in https://tex.stackexchange.com/a/52744/33952

Not sure what's up with all the empty spans (this is something we have in https://github.com/jgm/pandoc/issues/3488 as well).

I was thinking of adding a class to the span, but this can probably be done just as well using CSS pseudo-elements like `p::first-letter`.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -1320,7 +1320,7 @@ inlineCommands = M.union inlineLanguageCommands $ M.fromList $
   , ("i", lit "i")
   , ("\\", linebreak <$ (do inTableCell <- sInTableCell <$> getState
                             guard $ not inTableCell
-                            optional (bracketed inline)
+                            optional opt
                             spaces))
   , (",", lit "\8198")
   , ("@", pure mempty)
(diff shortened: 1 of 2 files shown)


mb21:
I thought about getting rid of the `[warning] Skipped '\noindent'` as well, but that seemed to break the test added in https://github.com/jgm/pandoc/commit/f4a452f89174828fea77614a4b6a067fc4675ba5, what I tried was:

    inlineCommands = ...
      , ("noindent", pure mempty)


    ```
    % pandoc -f latex -t native
    foo

    \noindent
    bar
    ^D
    [Para [Str "foo"]
    ,Para [Str "bar"]]
    ```

jgm:
I'd rather have separate PRs for conceptually separate changes, rather than three unrelated changes wrapped into one.

On `\lettrine`:  At the minimum, if we parse `\lettrine`, we need to preserve the information that this is a lettrine (perhaps `lettrine` as a class on the first span would do it, and maybe the spacing information could be preserved as an attribute on this).  This would allow one to recover the lettrine presentation in HTML or another format (perhaps using a filter).

On bold math:  `\bf` and `\mathbf` are already supported by `texmath`.  But of course texmath will only handle them if they're in a math mode section (e.g. inside `$`s).  If they actually work outside math mode (which would be a surprise to me), then the best way to handle them would be to parse `\bf` plus its argument as a `Math` element.

mb21:
Sure, I've redone`\lettrine` with a class added and removed the bold-math commit, which would belong to https://github.com/jgm/texmath/pull/26 I guess..?

jgm:
+++ Mauro Bieg [Oct 22 17 11:31 ]:
>   Sure, I've redone\lettrine with a class added and removed the bold-math
>   commit, which would belong to [1]jgm/texmath#26 I guess..?

It depends.  Can the `\bf` command be used outside of math
mode?  (It might be if it is defined with `\ensuremath`.)
If so, we need to handle it in pandoc by simply putting it
and its argument into Math.  It is already handled by
texmath on the writer side.

PR #4002, opened 2017-10-28
Creole reader

swilde:
I added an "old" style test, based on the "official" test case document from 
http://www.wikicreole.org/wiki/Creole1.0TestCases

I manually reviewed the matching pandoc native document -- which is a wearisome and error prone task, but I'm pretty confident represents the expected, conforming result.

cheers, 
sascha


jgm:
Terrific, thanks!

PR #4004, opened 2017-10-29
Write FB2 lists without nesting blocks inside <p>

labdsf:
According to FB2 XML schema `<empty-line />` cannot be placed inside
`<p>`. Linux FBReader can't display such paragraphs, e.g. any "loose"
lists produced by pandoc prior to this commit.  Besides that,
FB2 writer placed `<p>` inside `<p>` when writing nested lists,
this commit fixes the bug.

Also this commit removes leading non-breaking space from ordered
lists for consistency with bullet lists.

Definition lists are not affected at all.


labdsf:
There are some obvious bugs remaining like inconsistency between bullet lists and ordered lists. Nested ordered lists retain parent markers, while nested bullet list don't, so when you nest ordered lists inside bullet lists, indentation is broken. Also indentation with nonbreaking spaces does not display correctly due to nonmonospace fonts in FBReader. Maybe it works with something like `fbless`.

Definition lists are broken too it seems and the whole function `paraToPlain` should be removed at some point once they are fixed, I just have not tested them yet.

Lack of support for lists and some standardized way to create markers in FB2 is just a problem of the format so I had to stop at some point, there are just too many bugs to fix at a time. At least at this point `pandoc` produces lists that FBReader can display.

jgm:
Great, thanks!

PR #4008, opened 2017-10-31
FB2 writer: make bullet lists consistent with ordered lists

labdsf:
Previously bullet lists interacted in odd way with ordered lists.
For example, bullet lists nested in ordered list had incorrect
indentation. Besides that, indentation with spaces is not rendered
by FBReader and fbless. To avoid this problem, bullet lists are
indented by appending bullets to marker just the same way it is
done for ordered lists.

### DIFF
--- a/test/writer.fb2
+++ b/test/writer.fb2
@@ -261,23 +261,23 @@
 <p>Nested</p>
 </title>
 <p>• Tab</p>
-<p>◦ Tab</p>
-<p>* Tab</p>
+<p>• • Tab</p>
+<p>• • • Tab</p>
 <p>Here’s another:</p>
 <p>1. First</p>
 <p>2. Second:</p>
-<p>   • Fee</p>
-<p>   • Fie</p>
-<p>   • Foe</p>
+<p>2. • Fee</p>
+<p>2. • Fie</p>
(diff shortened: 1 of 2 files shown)


test/writer.fb2
```diff
@@ -261,23 +261,23 @@
...
-<p>* Tab</p>
+<p>• • Tab</p>
+<p>• • • Tab</p>
 <p>Here’s another:</p>
 <p>1. First</p>
 <p>2. Second:</p>
-<p>   • Fee</p>
-<p>   • Fie</p>
-<p>   • Foe</p>
+<p>2. • Fee</p>
```
jgm: I don't know what this looks like in the reader, but it seems like odd output (not natural looking).
Is there a way to force indentation -- perhaps by using nonbreaking spaces?

labdsf: Indentation with nonbreaking spaces, as used for bullet lists, does not work in FBReader and fbless. I don't know why they are used in the first place, maybe there are some readers which display them. Besides that, most readers don't use monospace fonts, so indentation will not be the same even if spaces are displayed. So I decided to just follow what pandoc does for ordered lists, at least it works even if ugly.

PR #4009, opened 2017-10-31
HTML Writer: consistently use dashed class-names

mb21:
see #3556

I went through the HTML writer and corrected those I could find. However, I left `smallcaps` and `literatehaskell` as is, since it feels like that could also be one word, or what do you think?


mb21:
Unfortunately, we also already have `comment-start`, `line-block`, `toc-section-number`, `header-section-number` and maybe others.

PR #4018, opened 2017-10-31
Creole Reader: fix lists with triling white space

swilde:
I found and fixed a problem with my creole parser: when there was trailing white space in list items, the next item was not correctly detected.

In addition I added another test for no-wiki (verbatim code) blocks, to ensure they are not affected by an similar problem (they are not).  And made some minor corrections of typos and code formatting...


jgm:
Great, thanks.

PR #4024, opened 2017-11-01
Remove obsolete reference to standalone beamer template

Jmuccigr:


jgm:
The man page is auto generated from MANUAL.txt.

PR #4040, opened 2017-11-04
INSTALL.md: update links to Linux distributions

rohieb:
Some links were outdated. Link to the package search page if no separate page for pandoc exists, and use HTTPS where possible.

All other links have been clicked once and deemed okay.


jgm:
Thanks!

Issue #4041, opened 2017-11-04
Differentiate between footnotes and endnotes


labdsf:
Some formats, like OpenDocument and Docx have support for footnotes and endnotes. LaTeX has `endnotes` package. FB3 format will have an option for endnotes: https://github.com/gribuser/FB3/blob/master/fb3_links.xsd

Now converting Docx to ODT converts all notes to *end*notes. Converting ODT to Docx converts all of them to *foot*notes.

I suggest to add footnote/endnote marker to AST to avoid losing this information for ODT and Docx. For LaTeX read footnotes as footnotes and add support for `endnotes` package later. For wiki-like formats it has to be footnotes for compatibility so users converting markdown to PDF via LaTeX are not surprised when they see their notes placed at the end of the document.

labdsf:
Definition even has a comment saying "footnote or endnote": 
https://github.com/jgm/pandoc-types/blob/f0731bf2106f0cd38c34a25c59f16ccc9082fff7/Text/Pandoc/Definition.hs#L266

I want to add a `NoteType` similar to `QuoteType` and `MathType` and add it to Note constructor. If there are no objections, I can try to do it myself and submit a PR for pandoc-types.

jgm:
Any discussion of changes to Note in the AST should also consider these issues: #1603, #2053.
Fixing those issues would seem to require storing the note reference separately from the note contents.

labdsf:
Only #1603 requires storing note reference separately from the note contents. Great idea, as it makes sense to refer to the same footnote multiple times sometimes, but I rarely see it in real texts. Also, as pointed out in #2053 discussion, referring to previous footnote in LaTeX can only be done manually with `\footnotemark`. I think #1603 is a completely separate issue which can be postponed.

As for #2053, it is just a problem with Markdown reader. It is correct that trying to convert it to PDF via LaTeX will result in missing footnotes, but it is a problem of LaTeX writer or LaTeX itself. 

What I see in real texts is endnotes referenced from footnotes (also discussed in #2053, @Kedrigern even suggested adding endnotes), but I see no problem with them after making AST changes that I suggest. LibreOffice does not support endnotes inside footnotes, manually produced `.fodt` document with endnote placed inside footnote does not load, but it is the bug of LibreOffice I think. LaTeX document with endnote inside footnote compiles correctly.

labdsf:
Pull request submitted: https://github.com/jgm/pandoc-types/pull/34

labdsf:
[...]

PR #4042, opened 2017-11-05
Endnotes

link2xt:
This is for issue #4041 and depends on https://github.com/jgm/pandoc-types/pull/34

So far I did some straightforward changes, trying to relpace `Note` with `Note FootNote` where possible. For some reason Lua tests fail.

Reproduce with: `stack test -j 1 --test-arguments '-p Lua' 2>&1 | less`

Can someone familiar with lua (@tarleb ?) help with fixing this? Looks like something should be done about `function Note`, for example replacing it with `function FootNote` and `function EndNote` like it is done with `Math`.


tarleb:
Don't have time to look at it right now, but the most likely explanation is that you'll have to edit the `M.Note` function in `data/pandoc.lua`.

labdsf:
@tarleb Thanks, trying to fix it.

labdsf:
Fixed, all tests pass. Now going to add endnotes to Docx and ODT.

labdsf:
Reporting progress:
Still need to fix OpenDocument reader and Docx writer.
Also maybe add some tests for OpenDocument or add EndNote to generic writer test at least.

labdsf:
All done, need review.

Especially for Docx writer. Documents it produces can be loaded by LibreOffice, that is all I have tested. Code is mostly a copy of footnotes, I think it can be hlinted and pruned after merge. I don't understand the reason for `defaultFootnotes`, so I left default endnotes list empty. Judging from comments, looks like a kludge for some MS Office glitch? OOXML standard part 1 is 5000+ pages, TLDR.

Also, I will have to switch from my repo back to official one, for that we need a new version of pandoc-types, right? So first https://github.com/jgm/pandoc-types/pull/34 is to be merged and a new version assigned, then I can fix first commit in this PR.

labdsf:
Added support for Muse (`{1}` for endnotes). Now ODT, Docx and Muse support endnotes both for reading and writing. CI failures are due to the fact that I use custom version of pandoc-types.

@jgm How do I proceed to get this PR merged? I think it is ready.

src/Text/Pandoc/Writers/Custom.hs
```diff
@@ -232,7 +232,9 @@ inlineToCustom (Link attr txt (src,tit)) =
 inlineToCustom (Image attr alt (src,tit)) =
   callFunc "Image" (Stringify alt) src tit (attrToMap attr)
 
-inlineToCustom (Note contents) = callFunc "Note" (Stringify contents)
+inlineToCustom (Note FootNote contents) = callFunc "FootNote" (Stringify contents)
```
jgm: This approach will break existing custom writers that use Note.
A better approach would be `callFunc "Note" "footnote"` and `callFunc "Note" "endnote"`, I think.
This would allow current writers to work without being changed.  @tarleb, any thoughts?

tarleb: That seems like a nice solution.  We'd get full backwards compatibility if we pass the note type after the content:

```haskell
inlineToCustom (Note noteType contents) = callFunc "Note" (Stringify contents) noteType
```

Lua won't mind the extra parameter if the function just expects contents. This would pass the node type as "FootNote" and "EndNote*.

labdsf: Fixed

labdsf:
@jgm ping, just updated PR

Linking #4213 as related

labdsf:
Rebased it once again on top of refactored Muse reader.

@jkr How do I update binary Docx files with the new test framework? All binary tests are failing mostly due to added endnote style. Manual testing shows produced files with endnotes are ok.

jkr:
Alexander <notifications@github.com> writes:
> @jkr How do I update binary Docx files with the new test framework?
> All binary tests are failing mostly due to added endnote style. Manual
> testing shows produced files with endnotes are ok.

Just run the tests with `--accept`. If you're using stack, that would be
`stack test --test-arguments="--accept"`. That will produce new docx
files. When you commit, just make sure that you say that you confirmed
they work, and mention which version of Word you checked them in.

jkr:
Oh, and if you test against a non-existent word file, it will create it.

More info here: https://ro-che.info/articles/2017-12-04-golden-tests

labdsf:
Updated once again, `stack` test passes, Muse round-trip test too.

@jkr I checked golden tests with LibreOffice, don't have any MS Office around.

@tarleb Lua doesn't work as expected. I am trying to run the following Lua filter, but it doesn't turn footnotes into endnotes:
```
return {
  {
    Note = function (elem)
      if elem.notetype == "FootNote" then
        elem.notetype = "EndNote"
      end
      return elem
    end,
  }
}
```
Also, is there anything that should be done to JSON filters compatible?

tarleb:
@labdsf Thanks, LGTM. I'm only wondering why it is `FootNote` and `EndNote` instead of `Footnote` and `Endnote`.

labdsf:
@tarleb 
> I'm only wondering why it is `FootNote` and `EndNote` instead of `Footnote` and `Endnote`.

Not sure, most likely because I looked at `SingleQuote`/`DoubleQuote` and `DisplayMath`/`InlineMath` while implementing it. This way quote, math and note types have capitalized "quote", "math" and "note" in their names, it is more consistent.

On the other hand, "footnote", "endnote" and "sidenote" are single words in English, while "doublequote" and "inlinemath" are not.

I still prefer the CamelCase style, because it allows to add something like "PopupNote", "MarginNote" etc.

jgm:
I still prefer Footnote and Endnote, because these are single words;
the CamelCase looks wrong here, though I see your thinking.

jgm:
This is a massive PR, so it's difficult to examine thoroughly.

Can you clarify which readers and writers actually support the distinction between endnotes and footnotes?  (I assume that the other readers and writers just treat all notes as footnotes?)

labdsf:
@jgm Endnotes are supported in both readers and writers for ODT, Docx and Muse (with Amusewiki syntax: "{1}", "{2}" instead of "[1]" and "[2]" for footnotes), besides Native. All the other readers and writers simply ignore the type and treat them as footnotes.

Later it would be nice to add support for LaTeX, at least for the writer, but that requires using additional packages, so I have left it out for now. HTML/EPUB can get support once #4213 is resolved.

link2xt:
@jgm 
Just rebased it again to merge with Muse changes. Any updates?

johannesburg:
I'm wondering if this PR is still being worked on? It would be great to be able to specify whether notes are endnotes and footnotes, especially for `.docx` files

PR #4043, opened 2017-11-05
update commands to extract deb archive on Linux

salim-b:
Since Pandoc v2.0 the xz instead of the gz format is used for the data archive inside the deb (`data.tar.xz`) and the old commands to extract a gzip archive fail.


jgm:
Thanks

PR #4046, opened 2017-11-06
fix copy-paste slip

salim-b:
Sorry, a copy-paste slip sneaked in while fixing the commands in the first commit :grimacing:


jgm:
I think this is irrelevant now, in light of the changes I just made in INSTALL.md

PR #4049, opened 2017-11-07
MANUAL: consolidate math documentation

mb21:
This is basically two things

1. merge HTML math section of "tex_math_dollars" into "Math rendering in HTML" section (there was a lot of overlap and the split was confusing for the reader)
2. reorder math rendering options by relevance (maybe some of them should also be discontinued?)

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1075,53 +1075,37 @@ Citation rendering
 Math rendering in HTML
 ----------------------
 
-`-m` [*URL*], `--latexmathml`[`=`*URL*]
-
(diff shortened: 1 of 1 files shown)


jgm:
Great!

Does `--latexmathml` even work any more?  We used to include the js automatically, but we don't any more, so it won't do much unless you explicitly specify a path to a local copy.  (But there isn't a warning or anything.)  We should probably have removed this option in pandoc 2, and also jsmath, maybe gladtex too.  I suppose we could deprecate them.

mb21:
I haven't tested any, but I don't think there's a reason to still support `latexmathml`, `jsmath`, `gladtex` and `mimetex`.

PR #4074, opened 2017-11-15
Extend isEnabled to WriterOptions to remove code duplication

labdsf:
`enabled` from CommonMark.hs was duplicate of isEnabled from Options.hs


jgm:
This looks good, it just needs a better commit message.

Since you're changing the API, you need to say clearly in the commit message what the API changes are.  (New class, generalization of isEnabled, isDisabled.)  This will eventually go in the changeling.

labdsf:
@jgm fixed

jgm:
Great, this change was a good idea.

PR #4077, opened 2017-11-16
Lua filters: preload text module

tarleb:
The `text` module is preloaded in lua. The module contains some UTF-8
aware string functions, implemented in Haskell.  The module is loaded on
request only, e.g.:

    text = require 'text'
    function Str (s)
      s.text = text.upper(s.text)
      return s
    end

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -327,6 +327,7 @@ library
                  scientific >= 0.2 && < 0.4,
                  vector >= 0.10 && < 0.13,
                  hslua >= 0.9 && < 0.10,
+                 hslua-module-text >= 0.1.2 && < 0.2,
                  binary >= 0.5 && < 0.9,
                  SHA >= 1.6 && < 1.7,
                  haddock-library >= 1.1 && < 1.5,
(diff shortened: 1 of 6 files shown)


tarleb:
It's a tiny package, so always loading it on startup would be possible.  However, it seems safer and more backwards compatible to not pollute the global namespace.

tarleb:
Functions provided by that module: `lower`, `upper`, `reverse`, `len`, and `sub`.

jgm:
Looks good.  Will wait til you've released the library before merging (then it will work with cabal builds).

One thought:  I suppose it's too radical an idea to inject these functions into string, so that string.upper "just works"?

tarleb:
It's now [published](https://hackage.haskell.org/package/hslua-module-text-0.1) on hackage.

> One thought: I suppose it's too radical an idea to inject these functions into string, so that string.upper "just works"?

This would be nice, but I'd not be very comfortable doing so unless we also replace all pattern-handling functions like `find` and `match`. I don't expect to get around to that in the near future.

jgm:
We should also add documentation to the `lua-filters.md` document.  (This could be done in a separate request if you prefer.)

Issue #4071, opened 2017-11-14
Lua filters: make pandoc.Attr more user friendly


fiapps:
I was unable to figure out on my own how to pass attributes to `pandoc.Attr`, and in the reply to my [StackExchange question](https://stackoverflow.com/questions/47290793/pandoc-lua-filters-how-to-specify-attributes-for-span-element/) I was invited to open an issue to discuss a better interface.

I tried to do this: `attrs = pandoc.Attr("", {}, {lang = "es-SP"})`.
The correct usage is (as of Pandoc 2.0.1.1): `attrs = pandoc.Attr("", {}, {{"lang", "es-SP"}})`. It was explained that the reason for this non-intuitive syntax is that it could be important to preserve the order of elements, and Lua provides no guarantees about the order of keys in a table.

fiapps:
My first comment on this is that the non-intuitive syntax wouldn't have been a problem for me if it had been documented, but "table containing string keys and values" wasn't clear enough for me to understand that I was meant to pass `{{"k1", "v1"}, {"k2", "v2"}, …}`. Sample code would have also been sufficient.

As for a more intuitive interface, I'm not sure what to suggest. My first thought was that an ordered list of {key, value} pairs is fine as an internal representation, but the constructor `pandoc.Attr` should accept a table like `{k1 = "v1", k2 = "v2"}`. This would work perfectly well for me, but I soon realized it would fail to preserve attribute order in other use cases. Suppose one is adding an attribute to elements with existing attributes. If the attributes are converted from an internal ordered table to a table indexed by key at the API level, and then back to an ordered table for internal purposes, the original order will be lost.

I suppose `pandoc.Attr` could accept either a table indexed by string keys or an ordered table of pairs, but that might be complicated to implement.

Thinking about users potentially manipulating attribute lists makes me think that with the current representation, it would be cumbersome for users to perform operations like getting the value of an attribute or setting an attribute that may already exist. Perhaps helper functions would be appropriate if these turn out to be common operations.

pkulchenko:
Maybe accept a regular (unordered) table and another (optional) table that only includes the keys in the right order?

tarleb:
Thanks for raising the issue, @fiapps.  You are right about documentation, an example would be helpful.

[...]

PR #4080, opened 2017-11-19
data/pandoc.lua: enable table-like behavior of attributes

tarleb:
Attribute lists are represented as associative lists in Lua. Pure
associative lists are awkward to work with. A metatable is attached to
attribute lists, allowing to access and use the associative list as if
the attributes were stored in as normal key-value pair in table.

Note that this changes the way `pairs` works on attribute lists. Instead
of producing integer keys and two-element tables, the resulting iterator
function now returns the key and value of those pairs.  Use `ipairs` to
get the old behavior.

Warning: the new iteration mechanism only works if pandoc has been
compiled with Lua 5.2 or later (current default: 5.3).

Closes: #4071

### DIFF
--- a/test/Tests/Lua.hs
+++ b/test/Tests/Lua.hs
@@ -7,9 +7,9 @@ import Test.Tasty (TestTree, localOption)
 import Test.Tasty.HUnit (Assertion, assertEqual, testCase)
 import Test.Tasty.QuickCheck (QuickCheckTests (..), ioProperty, testProperty)
 import Text.Pandoc.Arbitrary ()
-import Text.Pandoc.Builder (bulletList, doc, doubleQuoted, emph, header,
-                            linebreak, para, plain, rawBlock, singleQuoted,
-                            space, str, strong, (<>))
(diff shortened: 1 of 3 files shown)


test/Tests/Lua.hs
```diff
@@ -83,6 +83,14 @@ tests = map (localOption (QuickCheckTests 20))
       "uppercase-header.lua"
       (doc $ header 1 "les états-unis" <> para "text")
       (doc $ header 1 "LES ÉTATS-UNIS" <> para "text")
+
+  , testCase "Attribute lists are convenient to use" $
+    let kv_before = [("one", "1"), ("two", "2"), ("three", "3")]
+        kv_after  = [("one", "eins"), ("three", "3"), ("five", "5")]
+    assertFilterConversion "Attr doesn't behave as expected"
```
jgm: missing `in` here?

tarleb:
Argh. Refactored without running the tests afterwards 🙄

PR #4083, opened 2017-11-21
Change JSON instances for Opt to TemplateHaskell

jaspervdj:
The `Generic` JSON instances for `Text.Pandoc.App.Opt` seem to tickle a
particulary bad quadratic complexity case (Generics complexity is worse
than quadratic with respect to the number of fields in the datatype).
This is with GHC-8.2.1, I didn't test it using 8.0 but I assume it is
similar.

Using `Generic`, compilation of the `Text.Pandoc.App` module takes
minutes and often gets killed due to out of memory on slower machines
with "only" 8GB of accessible memory.  This is particularly annoying to
me since it means I cannot build pandoc on Travis.

TemplateHaskell is a little uglier, but the module seems to compile
within a few seconds, and compilation doesn't take more than 1GB of
memory.

Should I also change the other JSON instances throughout the codebase
for consistency?

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE CPP                 #-}
 {-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell     #-}
 {-# LANGUAGE TupleSections       #-}
 {-
 Copyright (C) 2006-2017 John MacFarlane <jgm@berkeley.edu>
@@ -44,8 +45,8 @@ import qualified Control.Exception as E
 import Control.Monad
 import Control.Monad.Except (catchError, throwError)
 import Control.Monad.Trans
-import Data.Aeson (FromJSON (..), ToJSON (..), defaultOptions, eitherDecode',
-                   encode, genericToEncoding)
+import Data.Aeson (defaultOptions, eitherDecode', encode)
+import Data.Aeson.TH (deriveJSON)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as B
 import Data.Char (toLower, toUpper)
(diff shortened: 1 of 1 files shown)


jgm:
Thanks!

I guess I'm inclined to leave the other instances alone, if they're not causing a problem.

jgm:
Although it might be worth a try if there's some way of measuring the effect on compile time and memory consumption!

jaspervdj:
Here are some numbers, simply done on my personal laptop with `time --verbose`.  As you can see, converting the remaining instances also makes an impact, although less dramatic than the `Opt` instances.

1. Generic instances:

        Command being timed: "stack build"
        User time (seconds): 416.65
        System time (seconds): 4.36
        Percent of CPU this job got: 100%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 6:57.72
        Maximum resident set size (kbytes): 5105856
        Major (requiring I/O) page faults: 2727
        Minor (reclaiming a frame) page faults: 2454687
        Voluntary context switches: 54224
        Involuntary context switches: 1078

2. Opt converted to TemplateHaskell:

        Command being timed: "stack build"
        User time (seconds): 327.77
        System time (seconds): 3.68
        Percent of CPU this job got: 101%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 5:27.19
        Maximum resident set size (kbytes): 2052216
        Major (requiring I/O) page faults: 1346
        Minor (reclaiming a frame) page faults: 1680365
        Voluntary context switches: 42785
        Involuntary context switches: 922

3. All JSON instances converted to TemplateHaskell:

        Command being timed: "stack build"
        User time (seconds): 316.37
        System time (seconds): 3.44
        Percent of CPU this job got: 101%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 5:15.43
        Maximum resident set size (kbytes): 1866248
        Major (requiring I/O) page faults: 1289
        Minor (reclaiming a frame) page faults: 1624889
        Voluntary context switches: 41279
        Involuntary context switches: 661

jgm:
I am considering removing this flag.  I don't see a proven benefit.
Compare timings for
https://github.com/jgm/pandoc/commit/3f28e332bf5608a54c54bd11c77778e46f625764/checks?check_suite_id=255050941
(derive_json_via_th True)
and
https://github.com/jgm/pandoc/commit/b235a187dc47e3d0afc8942f4e19c99527e1fca8/checks?check_suite_id=255450243
(derive_json_via_th False)

I don't see any significant differences on any ghc version since 8.0.  But there's a cost in more code to maintain and test.  Thoughts?

jgm:
Looking back, it seems this may only affect low-memory machines, so maybe that's why the difference isn't visible in our CI tests.  Still, I'm going to change the default to False.

Issue #4081, opened 2017-11-19
Provide convenience functions for working with lists


tarleb:
Operations on lists, like checking whether an element has a certain class, currently requires users to awkwardly iterate over the lists.  Working with lists should be made more convenient.

jgm:
Would it make sense to export a `list` module, similar to `text`, with things like `reverse`, `filter`, and `map`?

Alternatively (or in addition), for the common case of checking a class, we could add a method on `Attr`, so you could do `el.attributes.has_class("foo")`.

tarleb:
Maybe [penlight's List class](https://stevedonovan.github.io/Penlight/api/classes/pl.List.html) could be used. Penlight aims to be kind of a standard library in Lua.
I've yet to evaluate whether using it is feasible and desireable.

jgm:
I'm reluctant to include penlight, it's kind of a big
library, isn't it? (by lua standards).  That's why I thought
of just exposing some Haskell list manipulation functions.

+++ Albert Krewinkel [Nov 19 17 21:27 ]:

tarleb:
John MacFarlane <notifications@github.com> writes:

> I'm reluctant to include penlight, it's kind of a big
> library, isn't it? (by lua standards).

I thought it might be possible to include only the List module, but I
see now that it depends on other modules, pulling in a considerable
fraction of the rest of penlight. I agree that this would be too
heavy-weight.

> That's why I thought of just exposing some Haskell list manipulation
> functions.

Unfortunately, hslua isn't yet flexible enough to support function
polymorphism in an appropriate way: there is no general `LuaValue` type
as of now. We'd have to restrict the module functions to concrete values
like `String -> [String] -> Bool`; this seems highly impractical. Adding
a LuaValue type is on my todo list, but that list has become quite long
lately…

I guess I'd prefer if we wrote a small list module in lua instead. This
would come with the obvious downsides of having more lua code to test
and maintain, but maybe we could split it of into a separate library.
Maybe that's something that could go into the @pandoc organisation?

jgm:
+++ Albert Krewinkel [Nov 20 17 21:44 ]:
>   I guess I'd prefer if we wrote a small list module in lua instead. This
>   would come with the obvious downsides of having more lua code to test
>   and maintain, but maybe we could split it of into a separate library.

Sounds okay to me.  But maybe this needs to go in the pandoc
repo, since we'll want this code to be made available to all
lua filters.

tarleb:
[...]

PR #4099, opened 2017-11-26
Add basic lua List module

tarleb:
The List module is automatically loaded, but not assigned to a global
variable. It can be included in filters by calling `List = require
'List'`.

Lists of blocks, lists of inlines, and lists of classes are now given
`List` as a metatable, making working with them more convenient. E.g.,
it is now possible to concatenate lists of inlines using Lua's
concatenation operator `..` (requires at least one of the operants to
have `List` as a metatable):

    function Emph (emph)
      local s = {pandoc.Space(), pandoc.Str 'emphasized'}
      return pandoc.Span(emph.content .. s)
    end

Closes: #4081

### DIFF
--- a/data/List.lua
+++ b/data/List.lua
@@ -1,0 +1,110 @@
+--[[
+List.lua
+
+Copyright © 2017 Albert Krewinkel
+
+Permission to use, copy, modify, and/or distribute this software for any purpose
+with or without fee is hereby granted, provided that the above copyright notice
+and this permission notice appear in all copies.
+
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+THIS SOFTWARE.
+]]
+
+---
+-- Lua functions for pandoc scripts.
+--
+-- @author Albert Krewinkel
+-- @copyright © 2017 Albert Krewinkel
+-- @license MIT
+local M = {
+  _VERSION = "0.1.0"
+}
+
+------------------------------------------------------------------------
+-- Metatable for lists
+-- @type List
+local List = {}
+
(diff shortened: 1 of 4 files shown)


tarleb:
I don't consider this ready to merge yet. I just wanted to have something worth discussing.

jgm:
Is there a reason the module name is capitalized here, but
not in other places (e.g. pandoc, text)?

tarleb:
The module is also a class, which we capitalize in existing code. We could keep the module name lowercase, importing it into an capitalized var, but I felt that capitalizing the module would be the lesser evil.

tarleb:
I opted for functions which (slightly) resemble those in lua's `string` module and JavaScript's Array prototype.  A Common Lisp inspired function name was used where additional flexibility is desired (`find_if`). The library contains only functions which I considered essential, more can be added later.  From my point of view, this is ready to merge now.

The module loading code needs more work, but I'm planning to refactor this anyway. I'll address it in a separate commit/request.

data/List.lua
```diff
@@ -0,0 +1,110 @@
...
+end
+
+--- Returns a (shallow) copy of the list.
+function List:clone ()
+  local lst = setmetatable({}, getmetatable(self))
+  List.extend(lst, self)
+  return lst
+end
+
+--- Appends the given list to the end of this list.
```
jgm: This comment is wrong, isn't it?

data/pandoc.lua
```diff
@@ -262,7 +264,7 @@ M.DefinitionList = M.Block:create_constructor(
 M.Div = M.Block:create_constructor(
   "Div",
   function(content, attr)
-    return {c = {attr or M.Attr(), content}}
+    return {c = {attr or M.Attr(), List:new(content)}}
```
jgm: I'm curious why this change is needed?

jgm:
I've made a couple small comments in the code.

My main request is: can we also have a `filter` function that returns a new list with the elements of the old list that match a condition?

data/List.lua
tarleb: You are right, that's very wrong.

data/pandoc.lua
tarleb: It's just for convenience, making it possible to write `div.content:find_if(fn)` rather than `List.find_if(div.content, fn)`. Users won't have to `require 'List'`. It doesn't appear to affect performance in a significant way.

tarleb:
Added `filter` and fixed the comment.  I also renamed the module to `pandoc.List` to avoid future name clashes.

johnd0e:
```
local M = {
  _VERSION = "0.1.0"
}
```
This variable is not used. Intentionally?
Or maybe `_VERSION` should go into `List` table?

And question: where can we find documentation for this module, sample of usage etc?

tarleb:
> Or maybe _VERSION should go into List table?

Thanks, fixed.

There is no generated documentation yet. I guess we could either add it to lua-filters, or provide a link to a ldoc generated file. The former seems the better option, I'll check what needs to be done for this.

tarleb:
This is weird. My last changes, including the filter function, didn't make it to master. I must have pushed to the wrong branch.  I'll fix it.

Issue #4086, opened 2017-11-21
YAML Headers And HTML Metadata Issues


tajmone:
I'm facing some general issues with YAML headers in markdown documents.

For example, if I try to provide a specific string for `author-meta` (ie: have a different value than for `author`, to have different strings showing the document and its metadata) which contains an email address:

``` yaml
author-meta: John Doe <john.doe@dudemail.com>
```

... the email addressed get converted in a `mailto:` link:

``` html
<meta name="author" content="John Doe <a href="mailto:john.doe@dudemail.com">john.doe@dudemail.com</a>" />
```

If I try to escape the angle brackets:

``` yaml
author-meta: John Doe \<john.doe@dudemail.com\>
```
... , they get converted to HTML entities:

``` html
<meta name="author" content="John Doe &lt;john.doe@dudemail.com&gt;" />
```

These are the expected behaviours in markdown document, but pandoc should handle differently assignment of string to variable which concern HTML metadata or header stuff. Maybe the second case is correct, but chances are that the value of `author` meta key (unlike `description`) should allow angle brackets for machine parsing since it's not intended to be shown in browsers — but I'm not 100% sure.

Another problem I'm facing is trying to use double quotes within a `title` string — I've tried escaping the double quotes, enclosing the whole string in single quotes, etc., but to no avail. It might be just that I'm missing out some basic YAML rules here, but I have the impression that something changed here with pandoc v2 because in the past I managed to escape more or less any character I wanted to  see through.

jgm:
If you really just want to set a template variable, don't
put it in metadata; use the `-V` option on the command line.

Or, if you want a metadata value that is not going to be
interpreted as Markdown, you can use the new raw inline syntax:

    author-meta: '`John Doe <john.doe@dudemail.com>`{=html}'

>   Another problem I'm facing is trying to use double quotes within a
>   title string

An example would be helpful.  YAML escaping rules can be a
bit complex, but it really depends on the particular case.
Often no escaping is needed, but if you start with a double
quote character, you may need single quotes round the whole
thing.

tajmone:
Thanks a lot, great tip — I wasn't aware of the new raw inline syntax (still trying to get a grip on the tons of new features).

I'd like also to mention an issue I stumbled in regarding dates. If I set in YAML:

``` yaml
date: April 20, 2017
```

[...]

PR #4101, opened 2017-11-27
make normalizeDate more forgiving

mb21:
also parse two-digit days, e.g. "April 20, 2017"

replace some `%d` with `%e`, from [docs](https://www.stackage.org/haddock/lts-8.21/time-1.6.0.1/Data-Time-Format.html#v:formatTime):

> `%d`: day of month, 1-padded to two chars, 01 - 31
> `%e`: day of month, space-padded to two chars, 2 - 31

this was brought up in #4086

### DIFF
--- a/src/Text/Pandoc/Shared.hs
+++ b/src/Text/Pandoc/Shared.hs
@@ -291,7 +291,7 @@ normalizeDate s = fmap (formatTime defaultTimeLocale "%F")
              parseTime defaultTimeLocale
 #endif
         formats = ["%x","%m/%d/%Y", "%D","%F", "%d %b %Y",
-                    "%d %B %Y", "%b. %d, %Y", "%B %d, %Y",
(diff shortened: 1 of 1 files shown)


jgm:
Will `%e` accept a zero-padded day, like 04-02-17?  If not, then this change may break some existing documents.  (In that case, we could throw in the `%d` versions as well, I suppose.)

wilx:
If complex dates parsing is a topic, I have good experience with IBM's ICU library and parsing of dates based on [CLDR](http://cldr.unicode.org/). Though I have done that in Perl and not in Haskell.

mb21:
> Will `%e` accept a zero-padded day

yes, the following works:

    > parseTimeM True defaultTimeLocale "%e %B %Y" "02 April 2017" :: Maybe Day
    Just 2017-04-02

jgm:
Great.

Issue #4096, opened 2017-11-26
Default Theme is Kate not Pygments


tajmone:
**Pandoc v2.0.3**

Documentation mentions "pygments" as being the default style:

```
--highlight-style=STYLE|FILE
    Specifies the coloring style to be used in highlighted source code.
    Options are pygments (the default), kate, monochrome, breezeDark,
    espresso, zenburn, haddock, and tango.
```

But the command `pandoc --print-default-data-file default.theme` outputs the "kate" theme, not "pygments":

```json
{
    "metadata" : {
        "revision" : 1,
        "name" : "Default",
        "author" : "Kate Authors",
        "license" : "MIT",
        "read-only" : true
    },
    "text-styles": {
        "Normal" : {
            "text-color" : "#1f1c1b",
            "selected-text-color" : "#ffffff",
            "bold" : false,
            "italic" : false,
            "underline" : false,
            "strike-through" : false
        },
        "Keyword" : {
            "text-color" : "#1f1c1b",
            "selected-text-color" : "#ffffff",
            "bold" : true
        },
[...]
```

... pygments doesn't use the `"#1f1c1b` color for keywords, kate does.

tajmone:
**NOTE** — I'm not sure if the above is a mistake or intended; my understanding is that `pandoc --print-default-data-file default.theme` should output a template file covering all the possible syntax elements, so that it might be used as a starting point to build a custom style.

As of now, the documentation suggests that the above command will output the default theme used by pandoc.

Maybe the metadata of the [`default.theme`](https://github.com/jgm/pandoc/blob/2.0.3/data/default.theme) file could be changed so that it's clearer that it's a template to build-on: 

```json
{
    "metadata" : {
        "revision" : 1,
        "name" : "Default",
        "author" : "Kate Authors",
        "license" : "MIT",
        "read-only" : true
    },
```

Also, does the current `default.theme` cover all possible skylighting syntax elements? This is important to build a theme that covers all possible language syntaxes. So far I haven't found any documentation listing and explaining all the syntax elements covered by skylighting — except by peeking at the source code.

It would be useful to add to pandoc documentation the full list of skylighting syntax elements, with some comments regarding their scope, and possibly also their HTML equivalent class names — it would make it easier to create custom CSS stylesheets, instead of using pandoc styles, for handling different styling for different languages in a same document.

mb21:
[...]

PR #4106, opened 2017-11-28
Default Theme Boilerplate

tajmone:
Cut down contents of `default.theme` to entries which are actually used by pandoc.

This commit turns `default.theme` into a boilerplate for creating custom themes, which is what most users probably expect from the `pandoc --print-default-data-file default.theme` command.

(See Issue #4096)

### DIFF
--- a/data/default.theme
+++ b/data/default.theme
@@ -1,177 +1,231 @@
 {
     "metadata" : {
         "revision" : 1,
-        "name" : "Default",
-        "author" : "Kate Authors",
-        "license" : "MIT",
-        "read-only" : true
+        "name" : "theme boilerplate",
+        "author" : "your name",
+        "license" : "public domain"
     },
     "text-styles": {
         "Normal" : {
-            "text-color" : "#1f1c1b",
-            "selected-text-color" : "#ffffff",
-            "bold" : false,
-            "italic" : false,
-            "underline" : false,
-            "strike-through" : false
+            "text-color" : "#ffffff",
+            "background-color" : "#333333",
+            "bold" : true,
+            "italic" : true,
+            "underline" : true
         },
         "Keyword" : {
-            "text-color" : "#1f1c1b",
-            "selected-text-color" : "#ffffff",
-            "bold" : true
+            "text-color" : "#ffffff",
+            "background-color" : "#333333",
+            "bold" : true,
+            "italic" : true,
+            "underline" : true
         },
         "Function" : {
-            "text-color" : "#644a9b",
-            "selected-text-color" : "#452886"
+            "text-color" : "#ffffff",
+            "background-color" : "#333333",
+            "bold" : true,
+            "italic" : true,
+            "underline" : true
         },
         "Variable" : {
-            "text-color" : "#0057ae",
-            "selected-text-color" : "#00316e"
+            "text-color" : "#ffffff",
+            "background-color" : "#333333",
+            "bold" : true,
+            "italic" : true,
+            "underline" : true
         },
         "ControlFlow" : {
-            "text-color" : "#1f1c1b",
-            "selected-text-color" : "#ffffff",
-            "bold" : true
+            "text-color" : "#ffffff",
+            "background-color" : "#333333",
+            "bold" : true,
+            "italic" : true,
+            "underline" : true
         },
         "Operator" : {
-            "text-color" : "#1f1c1b",
-            "selected-text-color" : "#ffffff"
+            "text-color" : "#ffffff",
+            "background-color" : "#333333",
+            "bold" : true,
+            "italic" : true,
+            "underline" : true
         },
         "BuiltIn" : {
-            "text-color" : "#644a9b",
-            "selected-text-color" : "#452886",
-            "bold" : true
+            "text-color" : "#ffffff",
+            "background-color" : "#333333",
+            "bold" : true,
+            "italic" : true,
+            "underline" : true
         },
         "Extension" : {
-            "text-color" : "#0095ff",
-            "selected-text-color" : "#ffffff",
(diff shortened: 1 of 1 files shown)


jgm:
I'd rather have the default.theme be a workable theme (i.e., reasonable settings for colors, underlining, etc.).  I'd suggest using the colors and other settings from the current (kate) default.theme, and subtracting all the extra stuff.

jgm:
I noticed that there are some fields pandoc supports that Kate doesn't (e.g. for line numbers).
We can use the skylighting library's ToJSON instance to convert any style to JSON.  Here is pygments:
```json
{"background-color":null,"text-styles":[["KeywordTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#007020","bold":true}],["DataTypeTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#902000","bold":false}],["DecValTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#40a070","bold":false}],["BaseNTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#40a070","bold":false}],["FloatTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#40a070","bold":false}],["CharTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#4070a0","bold":false}],["StringTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#4070a0","bold":false}],["CommentTok",{"background-color":null,"underline":false,"italic":true,"text-color":"#60a0b0","bold":false}],["OtherTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#007020","bold":false}],["AlertTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#ff0000","bold":true}],["FunctionTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#06287e","bold":false}],["ErrorTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#ff0000","bold":true}],["WarningTok",{"background-color":null,"underline":false,"italic":true,"text-color":"#60a0b0","bold":true}],["ConstantTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#880000","bold":false}],["SpecialCharTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#4070a0","bold":false}],["VerbatimStringTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#4070a0","bold":false}],["SpecialStringTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#bb6688","bold":false}],["ImportTok",{"background-color":null,"underline":false,"italic":false,"text-color":null,"bold":false}],["VariableTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#19177c","bold":false}],["ControlFlowTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#007020","bold":true}],["OperatorTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#666666","bold":false}],["BuiltInTok",{"background-color":null,"underline":false,"italic":false,"text-color":null,"bold":false}],["ExtensionTok",{"background-color":null,"underline":false,"italic":false,"text-color":null,"bold":false}],["PreprocessorTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#bc7a00","bold":false}],["AttributeTok",{"background-color":null,"underline":false,"italic":false,"text-color":"#7d9029","bold":false}],["DocumentationTok",{"background-color":null,"underline":false,"italic":true,"text-color":"#ba2121","bold":false}],["AnnotationTok",{"background-color":null,"underline":false,"italic":true,"text-color":"#60a0b0","bold":true}],["CommentVarTok",{"background-color":null,"underline":false,"italic":true,"text-color":"#60a0b0","bold":true}],["InformationTok",{"background-color":null,"underline":false,"italic":true,"text-color":"#60a0b0","bold":true}]],"text-color":null,"line-numbers":"#aaaaaa"}
```

jgm:
It occurs to me that we could just have pandoc programatically spit out the JSON for any selected style.  Something like:

```
pandoc --print-highlight-theme pygments
```
Then we could just get rid of `default.theme` entirely.

tajmone:
> It occurs to me that we could just have pandoc programatically spit out the JSON for any selected style. 

This would be a much better approach, I agree. The default theme then would just be pygments (as expected), even if no name is provided. That would be an "example", rather than a template; an example that can easily be compared to its CSS output in the final HTML docs (rather than a more abstract boilerplate). But it would still need to contain only the JSON fiels that affect pandoc, of course, otherwise the confusion persists.

We'd better leave boilerplates and templates to user-run projects and collections, like mine, and limit pandoc to output its built-in templates and styles — otherwise the reasoning of minimum templates/boilerplates could apply also to HTML, and all other sort of formats.

> I noticed that there are some fields pandoc supports that Kate doesn't (e.g. for line numbers).

I didn't manage to access these from JSON (the `line-numbers` key seems to work only if the color of `Normal` isn't set, else it's overriden by the latter). Also, line-numbers background color is not consistent with that of the code.

Other fields from the `editor` section have strange side-effects in HTML5 CSS: 1px borders seem to get their colors from some of these entries — I did some tests, by changing color values to see where they'd turn up in the final doc CSS, but didn't manage to isolate all cases. 

I haven't worked much with pandoc except makrdown and HTML5, so I have no idea which other formats support syntax highlighting and how themes work with these. As for HTML5, the issues mentioned above regarding line numbers color and bg-color seem to need some fixing.

jgm:
I've implemented --print-highlight-style

Try building the latest. This also brings in some
skylighting changes related to line numbers.

If you still notice problems after that, let me know.

+++ Tristano Ajmone [Nov 28 17 19:02 ]:

PR #4123, opened 2017-12-05
Stop using overlapping instances

### LINKED ISSUE #4122: Consider removing overlapping instances
`PandocMonad` has a very general instance and one for `ParsecT` that overlaps it. This can be avoided using the `DefaultSignatures` extension. If you make the `MonadTrans` method implementations defaults, then you can write empty instance declarations for the usual transformers and a custom one for `ParsecT`.

treeowl:
Use `DefaultInstances` instead, with one instance per transformer.

Fixes #4122

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -301,6 +301,7 @@ library
                  containers >= 0.4.2.1 && < 0.6,
                  unordered-containers >= 0.2 && < 0.3,
                  parsec >= 3.1 && < 3.2,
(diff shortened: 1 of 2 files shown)


treeowl:
The only potential downside is that users of some fancy transformers may need orphan instances, which are of course evil.

tarleb:
Closing, as the corresponding issue was closed as well. Discussions for the necessity of the proposed change should be directed to #4122.

PR #4145, opened 2017-12-12
JATS reader [WIP]

hamishmack:
This PR includes:

* Basic support for reading JATS files.  Largely based on the HTML reader code.
* Changes to JATS writer to improve round tripping from pandoc native to JATS and back to native.
* Some tests for both the JATS reader and writer.
* default.nix (useful when working on pandoc with Nix).


jgm:
Thanks for this!

There's a lot of code duplication with the HTML reader.

Since JATS is an XML format, I think it would make much more sense to use the DocBook reader as a model for your JATS reader.  This uses a proper XML parser. The HTML reader uses a "tag soup" parser, so it can handle all kinds of crazy HTML, but this isn't what you want for an XML format.

The tooling for nix should ideally be in a separate PR.

hamishmack:
> There's a lot of code duplication with the HTML reader.

> Since JATS is an XML format, I think it would make much more sense to use the DocBook reader as a model for your JATS reader. This uses a proper XML parser, and you could customize it to fit the JATS format exactly. The HTML reader uses a "tag soup" parser, so it can handle all kinds of crazy HTML, but this isn't what you want for an XML format.

I'll give that a go.  Some of the HTML features made debugging JATS parsing a bit of a nightmare to debug (I left the code for implicit closing of tags in much longer than I should have).

> The tooling for nix should ideally be in a separate PR (and should definitely be in separate commits from the other changes).

> The JATS writer changes should also be in separate commits, and ideally a separate PR.

Yeah it is a bit of a monster.  Don't merge this request.  I'll keep updating though as it will probably be handy as an overview and to check that all the pieces fit.

I'll break it up into nicer chunks.  I've made a start with #4146.

jgm:
+++ Hamish Mackenzie [Dec 12 17 04:37 ]:
>   I'll give that a go. Some of the HTML features made debugging JATS
>   parsing a bit of a nightmare to debug (I left the code for implicit
>   closing of tags in much longer than I should have).

Great.  I'll change the title of this to indicate "work in progress."

mb21:
Closing this in favour of https://github.com/jgm/pandoc/pull/4177

PR #4148, opened 2017-12-12
fig, table-wrap & caption Divs for JATS writer

hamishmack:
Support writing `<fig>` and `<table-wrap>` elements with `<title>` and
`<caption>` inside them by using Divs with class set to on of
fig, table-wrap or cation.  The title is included as a Heading
so the constraint on where Heading can occur is also relaxed.

Also leaves out empty alt attributes on links.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -635,6 +635,7 @@ test-suite test-pandoc
                   Tests.Writers.ConTeXt
                   Tests.Writers.Docbook
                   Tests.Writers.HTML
+                  Tests.Writers.JATS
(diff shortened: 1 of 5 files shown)


hamishmack:
This adds a bit of code duplication in `blockToJATS` with the version of the function that supports titles starting with `fig`.    Do we still need that version or is it ok to only support using a Div?  If we do I can refactor out the code duplication.

jgm:
Yes, we still need to support turning a Para containing just an image whose title starts with `fig:` into a figure.

jgm:
Great, thanks.

PR #4151, opened 2017-12-13
Delete removed -S option from command in epub.md

georgerbr:
Because `--smart/-S` has been removed.
Maybe ` -f markdown+smart` shoud also be added?


jgm:
Thanks.  `markdown+smart` isn't needed because smart is default for markdown.

PR #4153, opened 2017-12-13
Unify lua initalization

tarleb:
The same init file (`data/init`) that is used to setup the Lua interpreter for Lua filters is also used to setup the interpreter of custom writers.lua.


jgm:
Seems like a good idea to me!

Issue #1800, opened 2014-12-09
XWiki Writer


mpickering:
The syntax can be see [here](http://platform.xwiki.org/xwiki/bin/view/Main/XWikiSyntax).

saigkill:
Is it possible to integrate it into any next release? That would be really great :-)

PonderingGrower:
+1

azarbayejani:
+1

rkumarp:
+1

HS-157:
+1

mb21:
pull requests welcome :) [this pull](https://github.com/jgm/pandoc/pull/3489/commits) is a good template...

marcbowes:
👍

ksmoran:
+1 !

mb21:
It doesn't help to "plus one"... what we need here is a pull request... if you have time it's not that hard... if you don't know haskell, http://learnyouahaskell.com/chapters is a good start.

dchenbecker:
Starting work on it now. Might have a first draft for review next week.

dchenbecker:
Started a PR with a (very) rough cut. I've tested locally and it seems to do what I want, but I need to work on unit tests and better support for various formatting options. I'm brand new to XWiki, so if someone who has more experience with XWiki could provide feedback on where I can better handle things, it would be appreciated.

geofflangenderfer:
@dchenbecker how hard do you think it would be to write a xwiki -> markdown parser?

PR #4167, opened 2017-12-17
Issue #1800 Add XWiki Support

dchenbecker:
This is very rudimentary for now. Based on DocuWiki/MediaWiki, but with less
support for special formatting because it's unclear what additional support
XWiki has for this.

### DIFF
--- a/src/Text/Pandoc/Writers/XWiki.hs
+++ b/src/Text/Pandoc/Writers/XWiki.hs
@@ -25,7 +25,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
    Stability   : alpha
    Portability : portable
 
-Conversion of 'Pandoc' documents to DokuWiki markup.
+Conversion of 'Pandoc' documents to XWiki markup.
 
 XWiki:  <http://www.xwiki.org/>
 XWiki Syntax:  <http://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiSyntax/>
--- a/src/Text/Pandoc/Writers/XWiki.hs
+++ b/src/Text/Pandoc/Writers/XWiki.hs
@@ -34,12 +34,14 @@ XWiki Syntax:  <http://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Feat
 module Text.Pandoc.Writers.XWiki ( writeXWiki ) where
 import Control.Monad.Reader (ReaderT, asks, local, runReaderT)
 import Data.List (intercalate)
+import qualified Data.Set as Set
 import Data.Text (Text, pack)
 import Text.Pandoc.Class (PandocMonad, report)
 import Text.Pandoc.Definition
 import Text.Pandoc.Logging
 import Text.Pandoc.Options
 import Text.Pandoc.Shared (escapeURI, isURI, linesToPara)
+import Text.Pandoc.Writers.MediaWiki (highlightingLangs)
 
 data WriterState = WriterState {
   listLevel :: String -- String at the beginning of items
@@ -92,9 +94,8 @@ blockToXWiki (Header level _ inlines) = do
   let eqs = replicate level '='
   return $ eqs ++ " " ++ contents ++ " " ++ eqs ++ "\n"
 
--- TODO: Optionally support code highlight plugin
-blockToXWiki (CodeBlock _ str) = do
-  return $ "{{{\n" ++ str ++ "\n}}}"
+-- XWiki doesn't appear to differentiate between inline and block-form code, so we delegate
+blockToXWiki (CodeBlock attrs str) = inlineToXWiki (Code attrs str)
 
 -- TODO: Figure out how to handle this better
 blockToXWiki (BlockQuote blocks) =
@@ -177,8 +178,12 @@ inlineToXWiki (Quoted DoubleQuote lst) = do
   contents <- inlineListToXWiki lst
   return $ "\"" ++ contents ++ "\""
 
-inlineToXWiki (Code _ contents) = do
-  return $ "{{\n" ++ contents ++ "\n}}"
+inlineToXWiki (Code (_,classes,_) contents) = do
+  let at  = Set.fromList classes `Set.intersection` highlightingLangs
+  return $
+    case Set.toList at of
+      [] -> "{{" ++ contents ++ "}}"
+      (l:_) -> "{{code language=\"" ++ l ++ "\"}}" ++ contents ++ "{{/code}}"
 
 inlineToXWiki (Cite _ lst) = inlineListToXWiki lst
 
--- a/src/Text/Pandoc/Writers/XWiki.hs
+++ b/src/Text/Pandoc/Writers/XWiki.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE OverloadedStrings #-}
 {-
 Copyright (C) 2008-2017 John MacFarlane <jgm@berkeley.edu>
 
@@ -33,9 +34,10 @@ XWiki Syntax:  <http://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Feat
 
 module Text.Pandoc.Writers.XWiki ( writeXWiki ) where
 import Control.Monad.Reader (ReaderT, asks, local, runReaderT)
-import Data.List (intercalate)
+import Data.Monoid
 import qualified Data.Set as Set
-import Data.Text (Text, pack)
+import qualified Data.Text as Text
+import Data.Text (Text, concat, init, intercalate, pack, replace, unlines, unwords)
 import Text.Pandoc.Class (PandocMonad, report)
 import Text.Pandoc.Definition
 import Text.Pandoc.Logging
@@ -44,7 +46,7 @@ import Text.Pandoc.Shared (escapeURI, isURI, linesToPara)
 import Text.Pandoc.Writers.MediaWiki (highlightingLangs)
 
 data WriterState = WriterState {
-  listLevel :: String -- String at the beginning of items
+  listLevel :: Text -- String at the beginning of items
 }
 
 type XWikiReader m = ReaderT WriterState m
@@ -52,20 +54,20 @@ type XWikiReader m = ReaderT WriterState m
 -- | Convert Pandoc to XWiki.
 writeXWiki :: PandocMonad m => WriterOptions -> Pandoc -> m Text
 writeXWiki _ (Pandoc _ blocks) =
-  let env = WriterState { listLevel = [] }
+  let env = WriterState { listLevel = "" }
   in do
   body <- runReaderT (blockListToXWiki blocks) env
-  return $ pack body
+  return $ body
 
 -- | Concatenates strings with line breaks between them.
-vcat :: [String] -> String
+vcat :: [Text] -> Text
 vcat = intercalate "\n"
 
-blockListToXWiki :: PandocMonad m => [Block] -> XWikiReader m String
+blockListToXWiki :: PandocMonad m => [Block] -> XWikiReader m Text
 blockListToXWiki blocks =
   fmap vcat $ mapM blockToXWiki blocks
 
-blockToXWiki :: PandocMonad m => Block -> XWikiReader m String
+blockToXWiki :: PandocMonad m => Block -> XWikiReader m Text
 
 blockToXWiki Null = return ""
 
@@ -78,21 +80,21 @@ blockToXWiki (Plain inlines) =
 
 blockToXWiki (Para inlines) = do
   contents <- inlineListToXWiki inlines
-  return $ contents ++ "\n\n"
+  return $ contents <> "\n\n"
 
 blockToXWiki (LineBlock lns) =
   blockToXWiki $ linesToPara lns
 
 blockToXWiki b@(RawBlock f str)
-  | f == Format "xwiki" = return str
+  | f == Format "xwiki" = return $ pack str
   | otherwise           = "" <$ report (BlockNotRendered b)
 
 blockToXWiki HorizontalRule = return "\n----\n"
 
 blockToXWiki (Header level _ inlines) = do
   contents <- inlineListToXWiki inlines
-  let eqs = replicate level '='
-  return $ eqs ++ " " ++ contents ++ " " ++ eqs ++ "\n"
+  let eqs = Text.replicate level "="
+  return $ eqs <> " " <> contents <> " " <> eqs <> "\n"
 
 -- XWiki doesn't appear to differentiate between inline and block-form code, so we delegate
 blockToXWiki (CodeBlock attrs str) = inlineToXWiki (Code attrs str)
@@ -101,42 +103,42 @@ blockToXWiki (CodeBlock attrs str) = inlineToXWiki (Code attrs str)
 blockToXWiki (BlockQuote blocks) =
   blockListToXWiki blocks
 
-blockToXWiki (BulletList contents) = blockToXWikiList "*" contents
+blockToXWiki (BulletList contents) = blockToXWikiList "*" $ contents
 
-blockToXWiki (OrderedList _ contents) = blockToXWikiList "1" contents
+blockToXWiki (OrderedList _ contents) = blockToXWikiList "1" $ contents
 
 blockToXWiki (DefinitionList items) = do
   lev <- asks listLevel
-  contents <- local (\s -> s { listLevel = listLevel s ++ ";" }) $ mapM definitionListItemToMediaWiki items
-  return $ vcat contents ++ if null lev then "\n" else ""
+  contents <- local (\s -> s { listLevel = listLevel s <> ";" }) $ mapM definitionListItemToMediaWiki items
+  return $ vcat contents <> if Text.null lev then "\n" else ""
 
 -- TODO: support more features
 blockToXWiki (Table _ _ _ headers rows') = do
   headers' <- mapM (tableCellXWiki True) headers
   otherRows <- mapM formRow rows'
-  return $ unlines (unwords headers':otherRows)
+  return $ Data.Text.unlines (Data.Text.unwords headers':otherRows)
 
-formRow :: PandocMonad m => [[Block]] -> XWikiReader m String
+formRow :: PandocMonad m => [[Block]] -> XWikiReader m Text
 formRow row = do
   cellStrings <- mapM (tableCellXWiki False) row
-  return $ unwords cellStrings
+  return $ Data.Text.unwords cellStrings
 
 
-tableCellXWiki :: PandocMonad m => Bool -> [Block] -> XWikiReader m String
+tableCellXWiki :: PandocMonad m => Bool -> [Block] -> XWikiReader m Text
 tableCellXWiki isHeader cell = do
   contents <- blockListToXWiki cell
   let cellBorder = if isHeader then "|=" else "|"
-  return $ cellBorder ++ contents
+  return $ cellBorder <> contents
   
 
-inlineListToXWiki :: PandocMonad m => [Inline] -> XWikiReader m String
+inlineListToXWiki :: PandocMonad m => [Inline] -> XWikiReader m Text
 inlineListToXWiki lst =
-  fmap concat $ mapM inlineToXWiki lst
+  fmap Data.Text.concat $ mapM inlineToXWiki lst
 
 
-inlineToXWiki :: PandocMonad m => Inline -> XWikiReader m String
+inlineToXWiki :: PandocMonad m => Inline -> XWikiReader m Text
 
-inlineToXWiki (Str str) = return $ escapeXWikiString str
+inlineToXWiki (Str str) = return $ escapeXWikiString $ pack str
 
 inlineToXWiki Space = return " "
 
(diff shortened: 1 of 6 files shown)


SimplGy:
Way to get after it, Derek. I'd love to have an xwiki writer and hope this gets merged in!

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,243 @@
...
+{- |
+   Module      : Text.Pandoc.Writers.XWiki
+   Copyright   : Copyright (C) 2008-2017 John MacFarlane
+   License     : GNU GPL, version 2 or above
+
+   Maintainer  : Derek Chen-Becker <dchenbecker@gmail.com>
+   Stability   : alpha
+   Portability : portable
+
+Conversion of 'Pandoc' documents to DokuWiki markup.
```
marcbowes: Stale comment? ("DokuWiki")

mb21:
Thanks!

What do you mean with "special formatting"?

Also, I assume XWiki merits its own writer and shouldn't be a variant of the mediawiki writer? (If there's a lot of code duplication, it could be done like e.g. `markdown_mmd`)

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: Fixing.

dchenbecker:
For special formatting I meant things like html tags. I do see now that there's support for inline element attributes (via the `(% <attr>="..." %)` tag, and I just found the section on escaping HTML tags directly, so I can start working on that. In terms of code duplication, this will look a lot like DocuWiki and MediaWiki, so if you think it's worth factoring that into a common set of functions I can take a stab at it, but it's going to be a bigger chunk of work and I would prefer to get something that works generally before iterating on improvements.

mb21:
What I meant was to modify the `Mediawiki.hs` file to check for extensions (similarly to how the `Markdown.hs` checks whether `Ext_*` are defined. (The `blockToXWiki (Header level _ inlines) =` caught my eye because it's the same as the Mediawiki one.) But on a closer look, the two writers seem to be different enough, so I guess it makes more sense the way you've started it.

btw, the following is a good test case (check the generated `writer.xwiki` _very_ carefully before checking in):

    pandoc -s -f native test/testsuite.native -o test/writer.xwiki

mb21:
P.S. maybe you want to use [Data.Text](https://www.stackage.org/haddock/lts-10.0/text-1.2.2.2/Data-Text.html) instead of `String`? Add `{-# LANGUAGE OverloadedStrings #-}` to the top, then it should be mostly the same except:

- `pack body` -> `body`
- `++` -> `<>`

Alternatively, especially if you have to indent a lot of things, [Text.Pandoc.Pretty](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Pretty.hs) may be a better choice? Lots of writers already use it...

dchenbecker:
I tried using OverloadedStrings and everything blew up. I really like Haskell, but Text/String is a perpetual source of confusion and frustration. If there's a strong reason for using it I'll take another stab at it.

dchenbecker:
OK, took another stab at String -> Text and got it working. I'm notsure what I'm doing wrong, but hoogle didn't return relevant results for "<>", so it took me a while to sort out what I actually needed (Data.Monoid). Kind of ironic that Monoids are one of my favorite FP concepts, but I couldn't figure out the syntax around them...

mb21:
cool, btw. because `<>` is an infix function, you’d have to hoogle `(<>)`

dchenbecker:
OK, after way too long I have some time to work on this again. I've made some additional fixes and checked in a writer.xwiki file. The formatting is about 95% correct, but getting that last 5% is going to take some thinking and better understanding of the code, so I would like to at least ship what I have so far. 

If the changes look good, I need to resolve a small conflict. Would you prefer that I just merge and fix conflicts, or do you want all my work rebased/squashed into a single commit?

jgm:
I'd prefer rebase/squash.

Can you comment on what is in that last 5%?

dchenbecker:
The last 5% currently includes these items:

* Custom bullet formatting
* Properly handling loose lists (numbering is currently reset)
* Footnotes

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+
+data WriterState = WriterState {
+  listLevel :: Text -- String at the beginning of items
+}
+
+type XWikiReader m = ReaderT WriterState m
+
+-- | Convert Pandoc to XWiki.
+writeXWiki :: PandocMonad m => WriterOptions -> Pandoc -> m Text
+writeXWiki _ (Pandoc _ blocks) =
```
jgm: The `in do` here, while perfectly legal, is not very idiomatic.  Usually one would do this instead:
```
writeXWiki _ (Pandoc _ blocks) = do
   let env = WriterState { listLevel = "" }
   body <- runReaderT (blockListToXWiki blocks) env
   return $ body
```

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+
+blockListToXWiki :: PandocMonad m => [Block] -> XWikiReader m Text
+blockListToXWiki blocks =
+  fmap vcat $ mapM blockToXWiki blocks
+
+blockToXWiki :: PandocMonad m => Block -> XWikiReader m Text
+
+blockToXWiki Null = return ""
+
+-- TODO: handle this
```
jgm: What handling does this need?
If there is a way to mark internal anchors in xwiki, one could check the id attribute and set an anchor.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+-- TODO: support more features
+blockToXWiki (Table _ _ _ headers rows') = do
+  headers' <- mapM (tableCellXWiki True) headers
+  otherRows <- mapM formRow rows'
+  return $ Data.Text.unlines (Data.Text.unwords headers':otherRows)
+
+formRow :: PandocMonad m => [[Block]] -> XWikiReader m Text
+formRow row = do
+  cellStrings <- mapM (tableCellXWiki False) row
+  return $ Data.Text.unwords cellStrings
```
jgm: You may know this, but it's usual to do something like
```
import qualified Data.Text as Text
```
and then `Text.unwords` works.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+tableCellXWiki :: PandocMonad m => Bool -> [Block] -> XWikiReader m Text
+tableCellXWiki isHeader cell = do
+  contents <- blockListToXWiki cell
+  let cellBorder = if isHeader then "|=" else "|"
+  return $ cellBorder <> contents
+  
+
+inlineListToXWiki :: PandocMonad m => [Inline] -> XWikiReader m Text
+inlineListToXWiki lst =
+  fmap Data.Text.concat $ mapM inlineToXWiki lst
```
jgm: You may know this, but it's more usual to use the infix form:
```
Data.Text.concat <$> mapM inlineToXWiki lst
```
Since `Data.Text.concat` is monoid mconcat, you could also just do
```
mconcat <$> mapM inlineToXWiki lst
```

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+
+inlineToXWiki :: PandocMonad m => Inline -> XWikiReader m Text
+
+inlineToXWiki (Str str) = return $ escapeXWikiString $ pack str
+
+inlineToXWiki Space = return " "
+
+inlineToXWiki LineBreak = return "\n"
+
+-- FIXME: better understand this
```
jgm: We can't render a SoftBreak as a newline, because a newline is interpreted as a hard break in xwiki -- or so I gather from your other code.  Hence it becomes a space.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+
+blockToXWiki (OrderedList _ contents) = blockToXWikiList "1" $ contents
+
+blockToXWiki (DefinitionList items) = do
+  lev <- asks listLevel
+  contents <- local (\s -> s { listLevel = listLevel s <> ";" }) $ mapM definitionListItemToMediaWiki items
+  return $ vcat contents <> if Text.null lev then "\n" else ""
+
+-- TODO: support more features
+blockToXWiki (Table _ _ _ headers rows') = do
```
jgm: Will this work even for complex table cells containing multiple blocks?
Might want to fall back to raw HTML for complex tables.
See other wiki writers.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+inlineToXWiki (Subscript lst) = do
+  contents <- inlineListToXWiki lst
+  return $ ",," <> contents <> ",,"
+
+-- TODO: Not supported. Maybe escape to HTML?
+inlineToXWiki (SmallCaps lst) = do
+  contents <- inlineListToXWiki lst
+  return contents
+
+inlineToXWiki (Quoted SingleQuote lst) = do
```
jgm: These are for "smart quotes" - does xwiki interpret straight quotes as curly automatically?
If not, might want to use curly quotes for Quoted.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+inlineToXWiki (Cite _ lst) = inlineListToXWiki lst
+
+-- FIXME: optionally support this (plugin?) 
+inlineToXWiki (Math _ str) = return $ "{{formula}}" <> (pack str) <> "{{/formula}}"
+
+inlineToXWiki il@(RawInline frmt str)
+  | frmt == Format "xwiki" = return $ pack str
+  | otherwise              = "" <$ report (InlineNotRendered il)
+
+-- TODO: Handle anchors
```
jgm: It would be good to handle internal links.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+     _  -> return $ "[[" <> label <> ">>" <> (pack src) <> "]]"
+
+inlineToXWiki (Image _ alt (source, tit)) = do
+  alt' <- inlineListToXWiki alt
+  let
+    titText = pack tit
+    params = intercalate " " [
+        if Text.null alt' then "" else "alt=\"" <> alt' <> "\"",
+          if Text.null titText then "" else "title=\"" <> titText <> "\""
+        ]
```
jgm: I don't think this is quite what you want.  If alt is null and titText is "hi", for example, you'll get `intercalate " " ["", "hi"]` which has a leading space. If they're both null, you'll get a string with a space.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+  return $ "[[image:" <> (pack source) <> (if Text.null params then "" else "|| " <> params) <> "]]"
+
+inlineToXWiki (Note contents) = blockListToXWiki contents
+
+-- FIXME: support attrs
```
jgm: In particular, it would be good to support anchors.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+  alt' <- inlineListToXWiki alt
+  let
+    titText = pack tit
+    params = intercalate " " [
+        if Text.null alt' then "" else "alt=\"" <> alt' <> "\"",
+          if Text.null titText then "" else "title=\"" <> titText <> "\""
+        ]
+  return $ "[[image:" <> (pack source) <> (if Text.null params then "" else "|| " <> params) <> "]]"
+
+inlineToXWiki (Note contents) = blockListToXWiki contents
```
jgm: Is there a more graceful fallback?  (Maybe just printing `[`...`]` around the note's contents would help identify it as a note.)

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+  
+-- Utility method since (for now) all lists are handled the same way
+blockToXWikiList :: PandocMonad m => Text -> [[Block]] -> XWikiReader m Text
+blockToXWikiList marker contents = do
+  lev <- asks listLevel
+  contents' <- local (\s -> s { listLevel = listLevel s <> marker } ) $ mapM listItemToXWiki contents
+  return $ vcat contents' <> if Text.null lev then "\n" else ""
+  
+
+listItemToXWiki :: PandocMonad m => [Block] -> XWikiReader m Text
```
jgm: If this doesn't handle loose lists (you say numbering restarts), then one solution might be to strip trailing newlines from the items, so they don't end up with blank lines in them even if paragraphs are used. This will still break when the items have multiple blocks in them, of course, but it would be a lot better in common cases.

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -0,0 +1,256 @@
...
+                              => ([Inline],[[Block]])
+                              -> XWikiReader m Text
+definitionListItemToMediaWiki (label, items) = do
+  labelText <- inlineListToXWiki label
+  contents <- mapM blockListToXWiki items
+  marker <- asks listLevel
+  return $ marker <> " " <> labelText <> "\n" <>
+    intercalate "\n" (map (\d -> (Data.Text.init marker) <> ": " <> d) contents)
+
+escapeXWikiString :: Text -> Text
```
jgm: Is this really all that needs escaping?
```
What if I write a dash--like this--and I don't want emphasis?
```

test/writer.xwiki
```diff
@@ -0,0 +1,862 @@
...
+
+
+
+----
+
+= Paragraphs =
+
+Here’s a regular paragraph.
+
+
```
jgm: There are two blank lines between each paragraph.  Is that needed in xwiki?  If not it should be avoided.

Note that Text.Pandoc.Pretty might be worth using instead of just generating Text.

test/writer.xwiki
```diff
@@ -0,0 +1,862 @@
...
+> A list:
+> 
+> 
+> 1. item one
+> 1. item two
+> 
+> Nested block quotes:
+> 
+> 
+> > nested
```
jgm: Does this style of nested block quote work in xwiki?

test/writer.xwiki
```diff
@@ -0,0 +1,862 @@
...
+
+
+
+{{code}}
+<hr />
+{{/code}}
+
+Hr’s:
+
+
```
jgm: Why all the blank lines here?

jgm:
I've added a bunch of comments.  I don't think this is ready to merge into master yet, but it's getting a lot closer!

dchenbecker:
Appreciate all of the comments! I'll see what I can do to start tackling them.

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: I've added this, but there doesn't seem to be a test case in testsuite.native where a Div has an id set (they're all ""):

```
9 matches for "Div" in buffer: testsuite.native
    231:,Div ("",[],[])
    234:,Div ("",[],[])
    235: [Div ("",[],[])
    236:  [Div ("",[],[])
    238: ,Div ("",[],[])
    253:,Div ("",[],[])
    260:,Div ("",[],[])
    261: [Div ("",[],[])
    262:  [Div ("",[],[])
```

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: Can you give me an example of what a `Link` to an internal reference would look like? I don't see anything in the testsuite.native file.

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: I think I have this fixed, but there don't appear to be any Spans at all in testsuite.native

test/writer.xwiki
dchenbecker: These are RawBlocks containing HTML markup. I tried using the `html` macro, which works for some items, but for most items in testsuite.native I get:

```
Failed to execute the [html] macro. Cause: [When using the HTML macro inline, you can only use inline HTML content. Block HTML content (such as tables) cannot be displayed.
```

Previously I was reporting `BlockNotRendered`, which seems like it's probably the appropriate response.

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: Are you saying I should escape things like "--", "**", and "//"?

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: Thanks, that seems to work better.

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: I think I have this fixed, but there aren't any examples like this in testsuite.native

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: XWiki doesn't support raw html, unfortunately. I found this out while trying to fix RawBlock (below). I can take a look at how DokuWiki does non-raw table handling and see if I can adapt it.

tarleb:
@dchenbecker this seems to be related to 7e389cb3dbdc11126b9bdb6a7741a65e5a94a43e. Try adding `{-# LANGUAGE NoImplicitPrelude #-}` to the top of the file, then import prelude manually:

``` haskell
module Text.Pandoc.Writers.XWiki ( writeXWiki ) where
import Prelude
```

dchenbecker:
I've addressed most of the top-level comments, but I need some guidance on cases that aren't currently covered in the `testsuite.native` file.

src/Text/Pandoc/Writers/XWiki.hs
mb21: The following markdown should result in a sensible xwiki representation with `pandoc input.md -t xwiki`

```
# title {#myid}

some internal link that [links to the tile](#myid)
```

kevinszatkowski:
@dchenbecker, Thanks for taking this on, I needed to convert media wiki to xwiki and your addition has been a lifesaver.  I did run into an issue with tables using the attached
[test-table.txt](https://github.com/jgm/pandoc/files/2810677/test-table.txt) as input the output didn't format correctly (not super hard to fix by hand).  Everything else so far has worked like a champ, thanks again!

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: OK, confirmed this is working with a quick `text.html` file:

```html
<div id="test">
This is div 3
</div>

<a href="#test">Link to test div</a>
```

produces:

```xwiki
{{ id name="test" /}}This is div 3
[[Link to test div>>#test]]
```

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: Test produces reasonable output, but only if the user adds the "id" extension:

https://extensions.xwiki.org/xwiki/bin/view/Extension/Id%20Macro

There's no default support for arbitrary anchors in xwiki.

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: Tested with a simple input html:

```html
<span id="foo">This is fooey</span>
```

produces:

```xwiki
{{ id name="foo" /}}This is fooey
```

Again, dependent on https://extensions.xwiki.org/xwiki/bin/view/Extension/Id%20Macro

src/Text/Pandoc/Writers/XWiki.hs
dchenbecker: Had to make a minor fix for this.

Tested with simple HTML input:

```html
<p>This--and not this--is ok.</p>
<p>So **is this**?</p>
<p>And what //about this//?</p>
```

produces:

```xwiki
This~-~-and not this~-~-is ok.

So ~*~*is this~*~*?

And what ~/~/about this~/~/?
```

dchenbecker:
I think I addressed most of the issues other than the raw HTML and the table formatting issue that @kevinszatkowski submitted. I got a weird build error, so I rebased onto current master, and I'll try to take a stab at fixing the table formatting.

dchenbecker:
> @dchenbecker, Thanks for taking this on, I needed to convert media wiki to xwiki and your addition has been a lifesaver. I did run into an issue with tables using the attached
> [test-table.txt](https://github.com/jgm/pandoc/files/2810677/test-table.txt) as input example the output didn't format correctly (not super hard to fix by hand). Everything else so far has worked like a champ, thanks again!

@kevinszatkowski  I switched to using "\\" for hard breaks, which I think fixes the table formatting issue. If you have a moment, could you test it out?

dchenbecker:
I have no idea what's going on with the most recent build failures. I haven't touched that code, and at least GHC 8.6 is passing. I tried rebasing onto master to see if I was maybe missing a commit, but that didn't help.

mb21:
I think the last commit on current master broke the tests...

jgm:
It's an update to cmark-gfm-hs that broke things.

dchenbecker:
OK, build passes and all issues are addressed, other than raw HTML, which I don't see a solution for because XWiki doesn't support it.

src/Text/Pandoc/Writers/XWiki.hs
jgm: I think a better approach would be
```
params = mconcat [ if Text.null alt' then "" else " alt=\""...
```
That is, just include the space at the beginning of the attribute if it's not null. That saves you an intercalate and a filter.

jgm:
This looks good!  If you want to tweak the code later on, feel free to submit additional PRs.  But this is definitely worth adding in its current state.

dchenbecker:
Awesome, I'll keep tweaking, and I'll start thinking about what it would take for reader support :)

geofflangenderfer:
@dchenbecker @jgm do y'all think it would be tough to write a xwiki -> markdown parser?

dchenbecker:
@geofflangenderfer I'll have to look at what's involved in writing the parser side of things, but the writer wasn't as hard as I initially thought it would be. I'll be upfront: between work and kids it's going to be slow going

geofflangenderfer:
I may give it a shot if I'm feeling ambitious :)

PR #4170, opened 2017-12-18
improve formatting of formulas in OpenDocument

oltolm:
This commit greatly improves the formatting of formulas in ODT documents.

Formulas are wrapped in a `span` element. Two formula styles are generated. One for display math and one for inline math.

An inline math formula is marked as text. See "Text mode" in [Math Guide](https://wiki.documentfoundation.org/Documentation/Publications#LibreOffice_Math_Guide).

Display math is placed in a paragraph of its own, which is centered. A new paragraph style is generated for each formula.

### DIFF
--- a/src/Text/Pandoc/Writers/ODT.hs
+++ b/src/Text/Pandoc/Writers/ODT.hs
@@ -56,7 +56,7 @@ import Text.Pandoc.XML
 import Text.TeXMath
 import Text.XML.Light
 
-data ODTState = ODTState { stEntries :: [Entry]
+newtype ODTState = ODTState { stEntries :: [Entry]
                          }
 
 type O m = StateT ODTState m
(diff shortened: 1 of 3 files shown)


mb21:
This improves how the generated math looks visually? Sounds like a welcome change...

I guess you'll have to adjust the failing tests (and possibly add a new one that covers you new code).

oltolm:
I fixed the test.

src/Text/Pandoc/Writers/ODT.hs
```diff
@@ -224,17 +224,36 @@ transformPicMath _ (Math t math) = do
...
+                                      else [("draw:style-name","fr1")
+                                           ,("text:anchor-type","as-char")]) $
              selfClosingTag "draw:object" [("xlink:href", dirname)
                                         , ("xlink:type", "simple")
                                         , ("xlink:show", "embed")
                                         , ("xlink:actuate", "onLoad")]
 
 transformPicMath _ x = return x
+
+isTextMode :: Bool -> B.ByteString
```
jgm: Just as a stylistic matter, I'd expect a function named `isTextMode` to return a Bool.
Maybe `inTextMode` or `withTextMode` would be a better name?

src/Text/Pandoc/Writers/ODT.hs
```diff
@@ -87,7 +87,7 @@ pandocToODT opts doc@(Pandoc meta _) = do
                                 P.readDataFile "reference.odt"
   -- handle formulas and pictures
   -- picEntriesRef <- P.newIORef ([] :: [Entry])
-  doc' <- walkM (transformPicMath opts) $ walk fixDisplayMath doc
+  doc' <- walkM (transformPicMath opts) $ walk wrapMath $ walk fixDisplayMath doc
```
jgm: We should think about whether some of these transformations should go in the OpenDocument writer rather than here.  (We had to put tarnsformPicMath here, originally at least, because OpenDocument writer was pure. Now that we have PandocMonad we could even think about putting it all in the OpenDocument writer.)

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -446,7 +474,9 @@ inlineToOpenDocument o ils
      | writerWrapText o == WrapPreserve
                   -> return $ preformatted "\n"
      | otherwise  ->return space
-    Span attr xs  -> withLangFromAttr attr (inlinesToOpenDocument o xs)
+    il@(Span attr xs)-> do
+                       S.when (isDisplayMath il) (setDisplayMath True)
```
jgm: This seems to set `stDisplayMath` to True at the first occurrence of display math; it then stays True throughout the rest of the document -- at least I don't see anywhere where it gets reset. Is this the intent?  AS noted above, I'm confused about what work this is doing.

src/Text/Pandoc/Writers/Shared.hs
```diff
@@ -213,6 +214,15 @@ fixDisplayMath (Para lst)
                          not (isDisplayMath x || isDisplayMath y)) lst
 fixDisplayMath x = x
 
+wrapMath :: Block -> Block
```
jgm: This seems to be used only in the ODT writer, so it should go there.

src/Text/Pandoc/Writers/ODT.hs
oltolm: I removed `wrapMath`. It was not needed anymore.

src/Text/Pandoc/Writers/ODT.hs
oltolm: I renamed it to `documentSettings`. That's the name of the element it creates.

src/Text/Pandoc/Writers/OpenDocument.hs
oltolm: The idea was to put display math into an own paragraph and center it. I have changed the implementation so that it is not needed anymore. Now I set the `anchor-type` of the formula to `paragraph`. which is what pandoc did before. Now I put the properties like `style:vertical-pos` and `style:vertical-rel` into a style, because these properties are not supported directly on a `draw:frame`. The document produced by pandoc was not compliant to the spec, I think.

src/Text/Pandoc/Writers/Shared.hs
oltolm: `wrapMath` has been removed.

oltolm:
Thank you for the review. I have improved the code.

This commit greatly improves the formatting of formulas in ODT documents. An inline math formula is marked as text. See "Text mode" in [Math Guide](https://wiki.documentfoundation.org/Documentation/Publications#LibreOffice_Math_Guide). Two formula styles are generated. One for display math and one for inline math. Pandoc now generates documents which better comply to the OpenDocument spec.

jgm:
Thanks!  Looks good.

jgm:
PS. I did a before-and-after test, and this did really improve things.

PR #4177, opened 2017-12-20
Add Basic JATS reader based on DocBook reader

hamishmack:


mb21:
I guess @jgm will have a closer look at this, but looks good to me. A few minor things:

- why not add the commit mentioned in https://github.com/stencila/pandoc/issues/4 to this pull? (or even squash it to one commit)
- the indentation is sometimes a bit funny, but then again, pandoc doesn't have a style guide...
- Maybe add a `jats-reader.native` and check it very carefully, i.e.:

      pandoc -s -f native test/testsuite.native -t jats | pandoc -s -f jats -t native -o test/jats-reader.native

hamishmack:
> why not add the commit mentioned in stencila#4 to this pull? (or even squash it to one commit)

I have added it to this PR, but left it as a separate commit since.  I left it out initially since it updates both the reader and the writer.

jgm:
Great, thanks for this!

jgm:
What would be cool to add some time would be `element-citation` parsing, producing both a Cite inline and a corresponding entry in the `references` field of metadata.

jgm:
> What would be cool to add some time would be element-citation parsing, producing both a Cite inline and a corresponding entry in the references field of metadata.

I've already started this -- needs filling, but it already gives acceptable results on some jats samples.

PR #4178, opened 2017-12-20
JATS writer: Make <p> optional in <td> and <th>

hamishmack:
If the contents are single `Plain` block then do not wrap them with
a <p> element.

### DIFF
--- a/src/Text/Pandoc/Writers/JATS.hs
+++ b/src/Text/Pandoc/Writers/JATS.hs
@@ -316,6 +316,9 @@ tableItemToJATS :: PandocMonad m
                    -> Bool
                    -> [Block]
                    -> JATS m Doc
+tableItemToJATS opts isHeader [Plain item] =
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Writers/JATS.hs
```diff
@@ -405,8 +408,11 @@ inlineToJATS opts (Link (ident,_,kvs) txt ('#':src, _)) = do
              [("alt", stringify txt) | not (null txt)] ++
              [("rid", src)] ++
              [(k,v) | (k,v) <- kvs, k `elem` ["ref-type", "specific-use"]]
-  contents <- inlinesToJATS opts txt
```
jgm: This change seems unrelated to the main change of this commit.  Please split into two separate commits, and ideally two separate PRs, one for each logical change.

jgm:
The table change seems fine!  But can you disentangle it with the unrelated link change?

PR #4184, opened 2017-12-21
HTML Reader: be more forgiving about figcaption

### LINKED ISSUE #4183: HTML Reader: Regression for figure without figcaption
HTML `figure`s without a corresponding `figcaption` should fall back to the `alt` text of the `img` (as was the behaviour prior to #3813) rather than having no caption at all.

```
$ pandoc -v
pandoc 2.0.5
$ echo '<figure><img src="foo" alt="bar"></figure>' | pandoc -f html -t markdown
![](foo)
$ echo '<img src="foo" alt="bar">' | pandoc -f html -t markdown
![bar](foo)
```

A related issue is that `figcaption` is ignored when it contains block tags like `div` or `p`:

```
$ echo '<figure><img src="foo" alt="bar"><figcaption><p>baz</p></figcaption></figure>' \
  | pandoc -f html -t markdown
![](foo)
$ echo '<figure><img src="foo" alt="bar"><figcaption><div>baz</div></figcaption></figure>' \
  | pandoc -f html -t markdown
![](foo)
$ echo '<figure><img src="foo" alt="bar"><figcaption><span>baz</span></figcaption></figure>' \
  | pandoc -f html -t markdown
![baz](foo)
```

mb21:
fixes #4183

### DIFF
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -74,8 +74,8 @@ import Text.Pandoc.Options (
     ReaderOptions (readerExtensions, readerStripComments),
     extensionEnabled)
 import Text.Pandoc.Parsing hiding ((<|>))
-import Text.Pandoc.Shared (addMetaField, crFilter, escapeURI, extractSpaces,
-                           safeRead, underlineSpan)
+import Text.Pandoc.Shared (addMetaField, blocksToInlines', crFilter, escapeURI,
+                           extractSpaces, safeRead, underlineSpan)
 import Text.Pandoc.Walk
 import Text.Parsec.Error
 import Text.TeXMath (readMathML, writeTeX)
@@ -588,8 +588,11 @@ pFigure = try $ do
   skipMany pBlank
   let pImg  = (\x -> (Just x, Nothing)) <$>
                (pOptInTag "p" pImage <* skipMany pBlank)
-      pCapt = (\x -> (Nothing, Just x)) <$>
-               (pInTags "figcaption" inline <* skipMany pBlank)
+      pCapt = (\x -> (Nothing, Just x)) <$> do
+                skipMany pBlank
+                bs <- pInTags "figcaption" block
+                skipMany pBlank
+                return $ blocksToInlines' $ B.toList bs
       pSkip = (Nothing, Nothing) <$ pSatisfy (not . matchTagClose "figure")
   res <- many (pImg <|> pCapt <|> pSkip)
   let mbimg = msum $ map fst res
--- a/test/command/4183.md
+++ b/test/command/4183.md
@@ -1,0 +1,32 @@
+```
+% pandoc -f html -t native
+<figure>
+  <img src="foo" alt="bar">
+</figure>
+^D
+[Para [Image ("",[],[]) [] ("foo","fig:")]]
+```
+
+```
+% pandoc -f html -t native
+<figure>
+  <img src="foo" alt="bar">
+  <figcaption>
+    <div>
+      baz
+    </div>
+  </figcaption>
+</figure>
+^D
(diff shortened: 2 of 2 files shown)


mb21:
update: got this working...

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -588,17 +588,23 @@ pFigure = try $ do
   skipMany pBlank
   let pImg  = (\x -> (Just x, Nothing)) <$>
                (pOptInTag "p" pImage <* skipMany pBlank)
-      pCapt = (\x -> (Nothing, Just x)) <$>
-               (pInTags "figcaption" inline <* skipMany pBlank)
+      pCapt = (\x -> (Nothing, Just x)) <$> do
+                skipMany pBlank
+                bs <- pInTags "figcaption" block
+                skipMany pBlank
+                let getInlines (Plain xs) = xs
```
jgm: Couldn't we just use `blocksToInlines` from `Text.Pandoc.Shared`?

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -588,17 +588,23 @@ pFigure = try $ do
...
   let mbimg = msum $ map fst res
   let mbcap = msum $ map snd res
   TagClose _ <- pSatisfy (matchTagClose "figure")
-  let caption = fromMaybe mempty mbcap
   case B.toList <$> mbimg of
-       Just [Image attr _ (url, tit)] ->
-         return $ B.para $ B.imageWith attr url ("fig:" ++ tit) caption
+       Just [Image attr alt (url, tit)] ->
+         let caption = fromMaybe alt mbcap
+         in  return $ B.para $ B.imageWith attr url ("fig:" ++ tit) $ B.fromList caption
```
jgm: As noted in the linked issue, I'm not convinced we should do this.  The HTML doesn't have a caption, so why should we produce an image with a caption?  The tricky thing is that we're overloading `alt`; it will still be used for the image's alt text, which just disappears in the current setup.  I'm not for the alt attribute disappearing, but I'm also hesitant about creating a caption where there wasn't one before.

mb21:
Thanks for the feedback, I agree.

One thing: `Text.Pandoc.Shared` should also export `blocksToInlines'`, don't you think? The two places where `blocksToInlines` is already used, `fromList` is immediately called on it. Should I make another pull for that?

jgm:
+++ Mauro Bieg [Dec 22 17 11:23 ]:
>   Thanks for the feedback, I agree.
>
>   One thing: Text.Pandoc.Shared should also export blocksToInlines',
>   don't you think? The two places where blocksToInlines is already used,
>   fromList is immediately called on it. Should I make another pull for
>   that?

Sure.

mb21:
btw, I've rebased this and changed according to your comments.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -588,8 +588,11 @@ pFigure = try $ do
   skipMany pBlank
   let pImg  = (\x -> (Just x, Nothing)) <$>
                (pOptInTag "p" pImage <* skipMany pBlank)
-      pCapt = (\x -> (Nothing, Just x)) <$>
-               (pInTags "figcaption" inline <* skipMany pBlank)
+      pCapt = (\x -> (Nothing, Just x)) <$> do
```
jgm: You probably want to put a try around this one.
Otherwise it may fail after consuming some blanks, if there's no figcaption.

jgm: `try (skipMany pBlank *> pInTags "figcaption" inline <* skipMany pBlank)`

jgm: Actually, do you need the `skipMany pBlank`?  Won't this be handled by `pSkip`?

mb21:
right... thanks!

Issue #4196, opened 2017-12-25
Lua filter can't be used on pandoc-citeproc output


jkr:
Native lua filters are used before external filters. This means that you can't use a lua filter on the output of other filters.

This is an issue for me because I have some filters that fix up `pandoc-citeproc` output (removing repeated parenthetical citations in MLA output). While my use is a bit niche, I can imagine more mainstream use-cases: a filter styling the `refs` Div that `pandoc-citeproc` produces.

Any thoughts on the best way to deal with this? Should `pandoc-citeproc` be a special case? I can always do two passes with a JSON intermediate, but that seems to go against the spirit of using the lua filters in the first place. (Of course, for the time being, I can keep using my old haskell filter for this special case.)

jkr:
Another example that I came across while converting my filters. One targeted LaTeX style uses a list environment for its works-cited list, so I need a filter that can run through the bibliography entries and put an `\item` in front of them (as well as wrapping the whole thing in `\begin{citedlist}`/`\end{citedlist}`). So, again, this is a filter that needs to run on the other side of `pandoc-citeproc`.

jgm:
@tarleb, any thoughts?  I actually think it might make sense
for all lua filters to run after regular filters, because
one of the main things that will be difficult to achieve in
a lua filter is pandoc-citeproc, and Jesse is certainly
right that we might want to manipulate its output.

Another option would be revising things so that filters
and lua-filters could be mixed in any order, and would
be run in the order presented on the command line.  That
would take a bit of re-plumbing in Text.Pandoc.App.

+++ Jesse Rosenthal [Dec 25 17 15:35 ]:

tarleb:
I'm not sure. This sounds sensible, but the opposite order can be useful
as well, as evident from this discussion:
https://groups.google.com/d/topic/pandoc-discuss/-l-1QkPoG9k/discussion
And the complexity inherent to special cases always scares me.

I guess my preferred solution would be to provide a way to run
pandoc-citeproc from within lua filters. E.g., if there was a simple way
to transform a document to json, one could do

    json = ??doc_to_json?? (doc)
    doc_with_refs_json = pandoc.pipe('pandoc-citeproc', {}, json)
    doc_with_refs = pandoc.read(doc_with_refs_json, 'json')

We could provide a helper function for this.

Or another idea: adding a command line switch '--disable-json-filters'
and a Lua function `run_json_filters(doc)` would allow users to gain
[...]

PR #4197, opened 2017-12-25
Work in Progress: Lua run json filter

tarleb:
Provide a lua function `pandoc.utils.run_json_filter` to run a JSON filter from within a lua filter or custom writer.

Intended as a solution for #4196.


tarleb:
I'll have to make some changes to hslua to implement this cleanly. Closing.

PR #4204, opened 2017-12-27
MANUAL introduce dedicated extensions section

mb21:
We've started using extensions in other formats than markdown more and more recently. However from the questions I've seen, user awareness hasn't followed suit. I think we should therefore update the documentation accordingly.

The following was the best I could come up with...


jgm:
Good idea

PR #4212, opened 2017-12-28
clarify default html output for --section-divs

iredwards:
In Pandoc v2 using --section-divs and -t html results in <section>s, not <div>s, by default, as was the case for v1.9.
This change to the Manual emphasizes that you must use -t html4 if you want divs, otherwise you get sections.


jgm:
Thanks, I agree this is clearer.

PR #4214, opened 2017-12-29
MANUAL.txt simplify and add more structure

mb21:
This structures the `Description` section and streamlines some of the wording, which was sometimes outdated.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3312,45 +3312,6 @@ For the most part this should give the same output as `raw_html`,
 but it makes it easier to write pandoc filters to manipulate groups
 of inlines.
 
-#### Extension: `fenced_divs` ####
-
-Allow special fenced syntax for native `Div` blocks.  A Div
-starts with a fence containing at least three consecutive
-colons plus some attributes. The attributes may optionally
-be followed by another string of consecutive colons.
-The attribute syntax is exactly as in fenced code blocks (see
-[Extension: `fenced_code_attributes`]).  As with fenced
-code blocks, one can use either attributes in curly braces
-or a single unbraced word, which will be treated as a class
-name.  The Div ends with another line containing a string of at
-least three consecutive colons.  The fenced Div should be
-separated by blank lines from preceding and following blocks.
-
-Example:
-
-    ::::: {#special .sidebar}
-    Here is a paragraph.
-
-    And another.
-    :::::
-
-Fenced divs can be nested.  Opening fences are distinguished
-because they *must* have attributes:
-
-    ::: Warning ::::::
-    This is a warning.
-
-    ::: Danger
-    This is a warning within a warning.
-    :::
-    ::::::::::::::::::
-
-Fences without attributes are always closing fences.  Unlike
-with fenced code blocks, the number of colons in the closing
-fence need not match the number in the opening fence.  However,
-it can be helpful for visual clarity to use fences of different
-lengths to distinguish nested divs from their parents.
-
 #### Extension: `raw_tex` ####
 
 In addition to raw HTML, pandoc allows raw LaTeX, TeX, and ConTeXt to be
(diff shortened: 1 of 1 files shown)


mb21:
Also:
- should we move the `fenced_divs` section down to a `Divs and Spans` section (in place of the current http://pandoc.org/MANUAL.html#spans) ?
- The TOC has grown to be quite intimidating, maybe make it collapsed per default (using JS) or place it after the Description section, or float it?
- What's up with https://pandoc.org/scripting.html ? Should this be a redirect to `/filters.html` ? I can't see how this is still online with https://github.com/jgm/pandoc-website/commit/5729a21ebe6d00f5e6e7cd892d624a9eccb54f2f

jgm:
+++ Mauro Bieg [Dec 29 17 10:10 ]:
>     * What's up with [2]https://pandoc.org/scripting.html ? Should this
>       be a redirect to /filters.html ? I can't see how this is still
>       online with [3]jgm/pandoc-website@5729a21


I fixed this (redirect).

MANUAL.txt
```diff
@@ -30,21 +33,20 @@ Simple], [Muse], [PowerPoint] slide shows and [Slidy], [Slideous],
...
-or output format.)
-
-In contrast to most existing tools for converting Markdown to HTML, which
-use regex substitutions, pandoc has a modular design: it consists of a
-set of readers, which parse text in a given format and produce a native
-representation of the document, and a set of writers, which convert
+Pandoc's enhanced version of Markdown includes syntax for [tables],
+[definition lists], [metadata blocks], [`Div` blocks][Extension:
+`fenced_divs`], [footnotes] and [citations], embedded
+[LaTeX][Extension: `raw_tex`] (incl. [math]), [Markdown inside HTML
```
jgm: Let's just say "including" instead of "incl."

MANUAL.txt
```diff
@@ -109,60 +111,44 @@ Markdown can be expected to be lossy.
...
+By default, pandoc produces a document fragment. To produce a standalone
+document (e.g. a valid HTML file including `<head>` and `<body>`),
+use the `-s` or `--standalone` flag:
 
     pandoc -s -o output.html input.txt
 
 For more information on how standalone documents are produced, see
-[Templates], below.
-
-Instead of a file, an absolute URI may be given.  In this case
```
jgm: The part about using URIs instead of files has disappeared.  Why?  Is that documented anywhere else?

MANUAL.txt
```diff
@@ -109,60 +111,44 @@ Markdown can be expected to be lossy.
...
 
 To convert `hello.html` from HTML to Markdown:
 
     pandoc -f html -t markdown hello.html
 
-Supported output formats are listed below under the `-t/--to` option.
-Supported input formats are listed below under the `-f/--from` option. Note
-that the `rst`, `textile`, `latex`, and `html` readers are not complete;
-there are some constructs that they do not parse.
+Supported input and output formats are listed below under [Options].
```
jgm: The original text said where to look under Options for this information, and I think we should continue to do that. Let's also mention `pandoc --list-input-formats` and `pandoc --list-output-formats`.

MANUAL.txt
```diff
@@ -228,6 +199,29 @@ If `wkhtmltopdf` is used, then the variables `margin-left`,
...
+representation: instead of `-o test.pdf`, use for example `-s -o test.tex`
+to output the generated LaTeX. You can then test it with `pdflatex test.tex`.
+
+When using LaTeX, the following packages need to be available
+(they are included with all recent versions of [TeX Live]):
+[`amsfonts`], [`amsmath`], [`lm`], [`unicode-math`],
+[`ifxetex`], [`ifluatex`], [`listings`] (if the
+`--listings` option is used), [`fancyvrb`], [`longtable`],
+[`booktabs`], [`graphicx`] and [`grffile`] (if the document
+contains images), [`hyperref`], [`xcolor`] (with `colorlinks`), [`ulem`], [`geometry`] (with the
```
jgm: overlong line

MANUAL.txt
jgm: Ah, I see it now below.  OK!

jgm:
Great!  I've got a few tweaks but I can make them myself after merging.

PR #4219, opened 2018-01-02
fixes for docx writer

stalmon:
docx output has two bugs:

- if a subtitle is defined, the subtitle was appended to the title string (output as pStyleM "Title" paragraph), and in addition to that written as new paragraph (pStyleM "Subtitle") after the title. This leads to double subtitle. As the subtitle is added as a separate paragraph, there is no need to append it to the title

- Word 2016 seems to default to a header format in tables, if there is no "firstRow" definition. setting it explicitly to "0" is compatible with Word 2013 and fixes it for Word 2016


jgm:
Great, thank you!

PR #4221, opened 2018-01-04
Ms writer: Added papersize variable

phikal:
I believe the title is self-describing. Should I also write some documentation, since this variable, and some others mentioned in the template aren't explicitly found in the manual?


jgm:
Yes, adding some documentation of these variables to the manual would be good (since we have something like that for latex and some other formats).

PR #4223, opened 2018-01-05
ConTeXt writer: Use xtables instead of Tables

### LINKED ISSUE #2102: ConTeXt writer: \starttable is deprecated
Note that the `\starttable` command is deprecated:

http://wiki.contextgarden.net/Tables_Overview

The preferred way of typesetting tables is to use "natural tables":
- http://wiki.contextgarden.net/TABLE
- http://www.pragma-ade.com/general/manuals/enattab.pdf

hmenke:
Fixes #2102.

This is a work in progress.  Because until about 5 minutes ago I had never written a single line of Haskell, I have no idea how to do even basic things, like passing a value to a function.  More changes will follow once I figure out how things work.  Linting from Haskell professionals is more than welcome!

### DIFF
--- a/src/Text/Pandoc/Pretty.hs
+++ b/src/Text/Pandoc/Pretty.hs
@@ -69,6 +69,7 @@ module Text.Pandoc.Pretty (
      , inside
      , braces
      , brackets
+     , maybeBrackets
      , parens
      , quotes
      , doubleQuotes
@@ -491,6 +492,11 @@ braces = inside (char '{') (char '}')
 -- | Puts a 'Doc' in square brackets.
 brackets :: Doc -> Doc
 brackets = inside (char '[') (char ']')
+
+-- | Apply 'brackets' to 'Doc' if boolean is true.
+maybeBrackets :: Bool -> Doc -> Doc
+maybeBrackets False = id
+maybeBrackets True = brackets
 
 -- | Puts a 'Doc' in parentheses.
 parens :: Doc -> Doc
--- a/src/Text/Pandoc/Pretty.hs
+++ b/src/Text/Pandoc/Pretty.hs
@@ -69,7 +69,6 @@ module Text.Pandoc.Pretty (
      , inside
      , braces
      , brackets
-     , maybeBrackets
      , parens
      , quotes
      , doubleQuotes
@@ -492,11 +491,6 @@ braces = inside (char '{') (char '}')
 -- | Puts a 'Doc' in square brackets.
 brackets :: Doc -> Doc
 brackets = inside (char '[') (char ']')
-
--- | Apply 'brackets' to 'Doc' if boolean is true.
-maybeBrackets :: Bool -> Doc -> Doc
-maybeBrackets False = id
-maybeBrackets True = brackets
 
 -- | Puts a 'Doc' in parentheses.
 parens :: Doc -> Doc
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -264,7 +264,8 @@ blockToConTeXt (Table caption aligns widths heads rows) = do
                                  zipWith colDescriptor widths aligns)
     headers <- if all null heads
                   then return empty
-                  else tableRowToConTeXt heads
+                  else liftM ("\\startxtablehead[topframe=on,bottomframe=on]" $$) $
+                       liftM ($$ "\\stopxtablehead") $ tableRowToConTeXt heads
     captionText <- inlineListToConTeXt caption
     rows' <- mapM tableRowToConTeXt rows
     return $ "\\startplacetable" <> brackets (
@@ -272,9 +273,11 @@ blockToConTeXt (Table caption aligns widths heads rows) = do
         then "location=none"
         else ("caption=" <> braces captionText))
       ) $$
-      "\\startxtable" $$
+      "\\startxtable[frame=off]" $$
       headers $$
+      "\\startxtablebody" $$
       vcat rows' $$
+      "\\stopxtablebody" $$
       "\\stopxtable" $$
       "\\stopplacetable" <> blankline
 
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -55,6 +55,8 @@ data WriterState =
               , stOptions          :: WriterOptions -- writer options
               }
 
+data Tabl = Xtb | Ntb deriving (Show, Eq)
+
 orderedListStyles :: [Char]
 orderedListStyles = cycle "narg"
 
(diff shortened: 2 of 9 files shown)


hmenke:
I don't know why the CI is failing... Sorry.

jgm:
First, welcome to Haskell, and pandoc!

Second, look at the CI log:
```

src/Text/Pandoc/Writers/ConTeXt.hs:263:9: Warning:
    Defined but not used: ‘colDescriptors’
<no location info>: 
Failing due to -Werror.
```
In the travis builds we turn on `-Werror` (which isn't enabled by default), so that the compiler will fail if there are any warnings.  In this case there is a warning--see above.  You're defining `colDescriptors` but you never use it, so (unless this is a mistake) you can simply remove the definition.  Watch for warnings on your local builds to avoid CI failures.

hmenke:
Thank you for the warm welcome and additional guidance!
The latest commit builds without warnings for me, so I hope the CI will be fixed.

hmenke:
On my machine `cabal build` and `cabal test` pass without warnings and errors.
I guess I'm done porting the code and it is now ready for review.
Please let me know if you are happy with the output format.

DaveJarvis:
Although it appears to be backwards compatible with the old behaviour for `\starttable`, will [line 257](https://github.com/jgm/pandoc/pull/4223/files#diff-8942905958822b03e09a0d97548bb117R257) override any custom settings to the contrary? For example, is it possible to format all of the tables as follows, without having to change the .tex file:

table-header

jgm:
I'm not a ConTeXt expert, so I can't really comment informatively on the change (other than to note that it's more verbose than the old table).  If the table format we currently use is deprecated, then I think we need to do this.  @adunning @DaveJarvis @ousia commented on the other issue -- maybe they'd like to look at the changes to the test file and see if the new format looks good?

DaveJarvis:
The changes look good -- my only concern is whether it is possible to change the header style across all tables in the generated `.tex` document without post-processing (e.g., using `\setupxtables` prior to `\starttext`).

hmenke:
@DaveJarvis I suspect that this is just not possible with xtables currently.  I have posted a question on the mailing list about that.  https://mailman.ntg.nl/pipermail/ntg-context/2018/090524.html

DaveJarvis:
Merging these changes for the next release would be smashing.

hmenke:
@jgm Do not merge yet please.
@DaveJarvis I'm working on poor-man's global setups.
```latex
\setupxtable[head][...]
\setupxtable[body][...]
\setupxtable[foot][...]
```

hmenke:
@DaveJarvis With the following Markdown input
```
| Field             | Entity              | Column                 | Notes            |
|-------------------|---------------------|------------------------|------------------|
| Level             | COGNITIVE_LEVEL     | LEVEL                  | Via ASSESSMENT   |
| **Count** / Total | ASSESSMENT_ANSWER   | ASSESSMENT_QUESTION_ID | Count the number |
| Count / **Total** | ASSESSMENT_QUESTION | ASSESSMENT_QUESTION_ID | Count the number |
```
and executing
```
pandoc -f markdown -t context -s -o test.tex test.md
```
I get this (shortened) output
```latex
% ...
\setupxtable[frame=off]
\setupxtable[head][topframe=on,bottomframe=on]
\setupxtable[foot][bottomframe=on]


\starttext

\startplacetable[location=none]
\startxtable
\startxtablehead[head]
\startxrow
\startxcell[align=right] Field \stopxcell
\startxcell[align=right] Entity \stopxcell
\startxcell[align=right] Column \stopxcell
\startxcell[align=right] Notes \stopxcell
\stopxrow
\stopxtablehead
\startxtablebody[body]
\startxrow
\startxcell[align=right] Level \stopxcell
\startxcell[align=right] COGNITIVE_LEVEL \stopxcell
\startxcell[align=right] LEVEL \stopxcell
\startxcell[align=right] Via ASSESSMENT \stopxcell
\stopxrow
\startxrow
\startxcell[align=right] {\bf Count} / Total \stopxcell
\startxcell[align=right] ASSESSMENT_ANSWER \stopxcell
\startxcell[align=right] ASSESSMENT_QUESTION_ID \stopxcell
\startxcell[align=right] Count the number \stopxcell
\stopxrow
\stopxtablebody
\startxtablefoot[foot]
\startxrow
\startxcell[align=right] Count / {\bf Total} \stopxcell
\startxcell[align=right] ASSESSMENT_QUESTION \stopxcell
\startxcell[align=right] ASSESSMENT_QUESTION_ID \stopxcell
\startxcell[align=right] Count the number \stopxcell
\stopxrow
\stopxtablefoot
\stopxtable
\stopplacetable

\stoptext
```
It doesn't look super clean because your material is not properly breakable but you could easily adjust hyphenation settings through the global setups now.
test

hmenke:
@jgm @DaveJarvis Please let me know whether all requirements are met.

DaveJarvis:
The output looks exactly as expected, thank you.

hmenke:
Latest CI failure is not my fault.

adityam:
I agree that pandoc should not use `\starttables`. Natural tables (`\bTABLE ... \eTABLE`) or Extreme Tables (`\startxtables`) are better alternatives. I personally prefer natural tables because, in the context of automatic document generation, it is much easier to tweak their output. To give a concrete example, with the xtable implementation, it is not possible to set the style for the first column. 

The consensus on the context mailing list appears to be that xtables are better because they are simpler and faster. Personally, I don't agree with this assessment. I never use tables that are longer than a page, so speed is never a concern for me, so I still exclusively use natural tables.

Nonetheless, if pandoc is changing the table mechanism, then the new mechanism should be the favoured mechanism in ConTeXt, which appears to be xtables. The xtables syntax is verbose enough that it is possible to write a `xtb-to-ntb` module (similar to `ntb-to-xtb` module provided with xtables), so that all the xtable macros are mapped to natural tables.

src/Text/Pandoc/Writers/ConTeXt.hs
```diff
@@ -252,33 +252,42 @@ blockToConTeXt HorizontalRule = return $ "\\thinrule" <> blankline
...
+tableRowToConTeXt aligns widths cols = do
+  cells <- mapM tableCellToConTeXt $ zip3 aligns widths cols
+  return $ ("\\startxrow") $$ vcat cells $$ "\\stopxrow"
+
+tableCellToConTeXt :: PandocMonad m => (Alignment, Double, [Block]) -> WM m Doc
+tableCellToConTeXt (align, width, blocks) = do
+  cellContents <- blockListToConTeXt blocks
+  let colWidth = if width == 0
+        then ""
+        else ",width=" <> braces (text (printf "%.2f\\textwidth" width))
```
adityam: This is ugly Haskell code. A better option is to use something like (sorry, it has been ages since I last wrote Haskell)
```
let colWidth = if width = 0 
    then Nothing
    else  Just $ "width=" <> braces ( ... )
```
Then on line 289, one could use

```
.... brackets (intercalate ", " [halign, fromMaybe "" colWidth])  <>
```

hmenke: Sorry, this is the first bit of Haskell I wrote *ever*.  Your suggestion doesn't work (as is)
```
src/Text/Pandoc/Writers/ConTeXt.hs:285:32:
    Could not deduce (Data.String.IsString [a0])
      arising from the literal ‘"align=right"’
    from the context (PandocMonad m)
      bound by the type signature for
                 tableCellToConTeXt :: PandocMonad m =>
                                       (Alignment, Double, [Block]) -> WM m Doc
      at src/Text/Pandoc/Writers/ConTeXt.hs:278:23-79
    The type variable ‘a0’ is ambiguous
    Relevant bindings include
      halign :: [a0] (bound at src/Text/Pandoc/Writers/ConTeXt.hs:284:7)
    Note: there is a potential instance available:
      instance Data.String.IsString [Char] -- Defined in ‘Data.String’
    In the expression: "align=right"
    In a case alternative: AlignLeft -> "align=right"
    In the expression:
      case align of {
        AlignLeft -> "align=right"
        AlignRight -> "align=left"
        AlignCenter -> "align=middle"
        AlignDefault -> "align=right" }

src/Text/Pandoc/Writers/ConTeXt.hs:290:16:
    Couldn't match expected type ‘Doc’ with actual type ‘[a0]’
    Relevant bindings include
      halign :: [a0] (bound at src/Text/Pandoc/Writers/ConTeXt.hs:284:7)
    In the first argument of ‘brackets’, namely
      ‘(intercalate ", " [halign, fromMaybe "" colWidth])’
    In the first argument of ‘(<>)’, namely
      ‘brackets (intercalate ", " [halign, fromMaybe "" colWidth])’

src/Text/Pandoc/Writers/ConTeXt.hs:290:55:
    Couldn't match type ‘Doc’ with ‘[a0]’
    Expected type: Maybe [a0]
      Actual type: Maybe Doc
    Relevant bindings include
      halign :: [a0] (bound at src/Text/Pandoc/Writers/ConTeXt.hs:284:7)
    In the second argument of ‘fromMaybe’, namely ‘colWidth’
    In the expression: fromMaybe "" colWidth
```

adityam:
I'll need to install a working Haskell distribution to test. I'll do that 
early next week.

hmenke:
@adityam Thanks for your help.  You should now be able to directly push to my branch.

DaveJarvis:
The output contains `[location=none]` on each table written to the document:

``` .tex
\startplacetable[location=none]
```

This seems to preclude the possibility of changing the table orientation and location via setups.

See also:

* http://wiki.contextgarden.net/Floating_Objects
* http://www.pragma-ade.com/general/manuals/xtables-mkiv.pdf
* http://wiki.contextgarden.net/xtables

Is it possible to remove the `[location=none]` without breaking the xtables?

DaveJarvis:
On that same token:

``` .tex
\startxcell[align=right,width={0.50\textwidth}]
```

* The `align=right` seems to be superfluous? (The default is `flushleft` aka `right`, isn't it?) This isn't a huge issue because suppressing the default value would take more code.
* The computation of `width={0.50\textwidth}` also precludes the possibility of changing the the columns widths throughout the document.

Not sure if there's a way to suppress the calculation for the width. If `xtable`s automatically pick the ideal width for all columns (such that the table fits within the document's `\textwidth`), then hard-coding the `\textwidth` does not seem necessary and takes away the possibility of forcing a different style (e.g., 30% for the first column and 70% for the second column through the document).

hmenke:
> Is it possible to remove the `[location=none]` without breaking the xtables?

This actually has nothing to with xtables but only with caption placement.  We could omit `location=none`
```latex
\startplacetable[location=none]
```
to always print a (possibly empty) caption, but I see no reason to break with the previous behaviour here.

> The `align=right` seems to be superfluous?

That is correct, `align=right` is superfluous in most cases.  I adjusted the code to omit default alignment.

> The computation of `width={0.50\textwidth}` also precludes the possibility of changing the the columns widths throughout the document.

The width calculation I have only kept from the original Tables approach in pandoc.  I think the width is only forwarded from the reader, i.e. if your input specifies a certain column width it will be kept in the ConTeXt writer (probably what you'd want).  The markdown input below for example will not contain any `width=...` in the output.

```
| Field             |              Entity | Column                 | Notes            |
|-------------------|---------------------|------------------------|------------------|
| Level             | COGNITIVE_LEVEL     | LEVEL                  | Via ASSESSMENT   |
| **Count** / Total | ASSESSMENT_ANSWER   | ASSESSMENT_QUESTION_ID | Count the number |
| Count / **Total** | ASSESSMENT_QUESTION | ASSESSMENT_QUESTION_ID | Count the number |
:This is the caption
```

src/Text/Pandoc/Writers/ConTeXt.hs
hmenke: `intercalate` doesn't work on elements of type `Doc` because it is essentially `concat $ intersperse` and `concat` is not defined for `Doc`.  One has to use `hcat $ intersperse` instead.

DaveJarvis:
Source:

``` .html
<table>
  <tr>
    <td><strong>ID</strong></td>
    <td>ID-1</td>
  </tr>
  <tr>
    <td><strong>Steps</strong></td>
    <td>
      <div>
        <ol>
          <li>One</li>
          <li>Two</li>
          <li>Three</li>
        </ol>
      </div>
    </td>
  </tr>
</table```

Generates:

``` .tex
\startplacetable[location=none]
\startxtable
\startxtablebody[body]
\startxrow
\startxcell[align=right,width={0.50\textwidth}] {\bf ID} \stopxcell
\startxcell[align=right,width={0.50\textwidth}] ID-1 \stopxcell
\stopxrow
\stopxtablebody
\startxtablefoot[foot]
\startxrow
\startxcell[align=right,width={0.50\textwidth}] {\bf Steps} \stopxcell
\startxcell[align=right,width={0.50\textwidth}] 

\startitemize[n,packed]
\item
  One
\item
  Two
\item
  Three
\stopitemize

 \stopxcell
\stopxrow
\stopxtablefoot
\stopxtable
\stopplacetable
```

Not sure why the `{0.50\textwidth}` is generated?

hmenke:
That's from the HTML reader.
https://github.com/jgm/pandoc/blob/78b142b880730bceaae7b9c0491e7f645d133380/src/Text/Pandoc/Readers/HTML.hs#L520-L524
I don't really feel like touching that code though.

mb21:
The `isSimple` is `True` iff there are only empty or `Plain` cell contents (no `Para` or `Div` etc.) Only if there might be multiline content the columns get widths assigned...

DaveJarvis:
Thank you. I've opened a new issue to address the table widths.

https://github.com/jgm/pandoc/issues/4234

hmenke:
@jgm In my opinion it's ready to merge

src/Text/Pandoc/Pretty.hs
```diff
@@ -492,6 +493,11 @@ braces = inside (char '{') (char '}')
 brackets :: Doc -> Doc
 brackets = inside (char '[') (char ']')
 
+-- | Apply 'brackets' to 'Doc' if boolean is true.
```
jgm: I'd prefer not to add this to the Pretty module.  You only use it once, anyway, in the ConTeXt writer, and I think the code would be more intelligible without it.  (Or just define maybeBrackets in the writer where you use it.)

jgm:
@hmenke - I've added a small comment about the change to Text.Pandoc.Pretty. Also, should we have some tests for the natural table output mode?  Otherwise, looks good!

hmenke:
@jgm I have fixed your suggestion.  I also tried to add a test for the Natural Tables but could not figure out how to do it.  Is there a guide somewhere?

jgm:
You could add a test to test/Tests/Writers/ConTeXt.hs.

+++ Henri Menke [Jan 15 18 20:43 ]:

hmenke:
@jgm I've added a test for Natural Tables.

jgm:
Great, thanks for this contribution to pandoc!

xeruf:
Thanks! Unfortunately, neither with this nor with the `ntb` extension does ConTeXt export respect colspan, tested working input from LaTeX and HTML.

hmenke:
I just tried with the `colspan` example from https://www.w3schools.com/tags/att_td_colspan.asp but it doesn't look like Pandoc's intermediate representation knows about `colspan` so it's impossible for the ConTeXt writer to support that
```console
$ pandoc -f html -t native test.html 
[Plain [Str "Month"]
,Plain [Str "Savings"]
,Plain [Str "January"]
,Plain [Str "$100"]
,Plain [Str "February"]
,Plain [Str "$80"]
,Plain [Str "Sum:",Space,Str "$180"]]
```

jgm:
Pandoc can handle colspan if you're using a relatively recent version:
```
% pandoc -t native -f html
<table>
  <tr>
    <th>Month</th>
    <th>Savings</th>
  </tr>
  <tr>
    <td>January</td>
    <td>$100</td>
  </tr>
  <tr>
    <td>February</td>
    <td>$80</td>
  </tr>
  <tr>
    <td colspan="2">Sum: $180</td>
  </tr>
</table>
^D
[ Table
    ( "" , [] , [] )
    (Caption Nothing [])
    [ ( AlignDefault , ColWidthDefault )
    , ( AlignDefault , ColWidthDefault )
    ]
    (TableHead
       ( "" , [] , [] )
       [ Row
           ( "" , [] , [] )
           [ Cell
               ( "" , [] , [] )
               AlignDefault
               (RowSpan 1)
               (ColSpan 1)
               [ Plain [ Str "Month" ] ]
           , Cell
               ( "" , [] , [] )
               AlignDefault
               (RowSpan 1)
               (ColSpan 1)
               [ Plain [ Str "Savings" ] ]
           ]
       ])
    [ TableBody
        ( "" , [] , [] )
        (RowHeadColumns 0)
        []
        [ Row
            ( "" , [] , [] )
            [ Cell
                ( "" , [] , [] )
                AlignDefault
                (RowSpan 1)
                (ColSpan 1)
                [ Plain [ Str "January" ] ]
            , Cell
                ( "" , [] , [] )
                AlignDefault
                (RowSpan 1)
                (ColSpan 1)
                [ Plain [ Str "$100" ] ]
            ]
        , Row
            ( "" , [] , [] )
            [ Cell
                ( "" , [] , [] )
                AlignDefault
                (RowSpan 1)
                (ColSpan 1)
                [ Plain [ Str "February" ] ]
            , Cell
                ( "" , [] , [] )
                AlignDefault
                (RowSpan 1)
                (ColSpan 1)
                [ Plain [ Str "$80" ] ]
            ]
        , Row
            ( "" , [] , [] )
            [ Cell
                ( "" , [] , [] )
                AlignDefault
                (RowSpan 1)
                (ColSpan 2)
                [ Plain [ Str "Sum:" , Space , Str "$180" ] ]
            ]
        ]
    ]
    (TableFoot ( "" , [] , [] ) [])
]
```

hmenke:
Oh, okay. I'll have a look at some in the future then.

xeruf:
That would be fantastic, with this feature in place `pandoc` could aid me greatly.

PR #4227, opened 2018-01-07
Run JSON filters from Lua filters

tarleb:
Allow running JSON filters from within Lua.  This is useful when only a temporary sub-document should be processed. E.g., [chapter wise bibliography](https://groups.google.com/d/topic/pandoc-discuss/N4Hmr8wErBs/discussion) could be created using this filter.

``` lua
local List = require 'pandoc.List'
local utils = require 'pandoc.utils'

-- flatten should probably be provided by pandoc.utils
local function flatten (lst)
  local res = List:new{}
  for i, el in ipairs(lst) do
    if el.t == 'Sec' then
      res[#res + 1] = pandoc.Header(el.level, el.label, el.attr)
      res:extend(flatten(el.contents))
    else
      res[#res + 1] = el
    end
  end
  return res
end

function Pandoc (doc)
  local res = List:new{}
  local sections = utils.hierarchicalize(doc.blocks)
  for _, sec in ipairs(sections) do
    -- create temporary sub-documents for each (top-level) section
    local section_doc = pandoc.Pandoc(flatten({sec}), doc.meta)
    -- resolve citations in sub-document using pandoc-citeproc
    local filtered = utils.run_json_filter('pandoc-citeproc', section_doc)
    res:extend(filtered.blocks)
  end
  return pandoc.Pandoc(res, doc.meta)
end
```

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -1433,6 +1433,37 @@ functions.
         print(table.concat(elements[1].numbering, '.')) -- 0.1
         print(table.concat(elements[2].numbering, '.')) -- 0.2
 
+[`run_json_filter (doc, filter[, args])`]{#utils-run_json_filter}
+
+:   Filter the given doc by passing it through the a JSON filter.
+
+    Parameters:
+
+    `doc`:
+    :   the Pandoc document to filter
+
+    `filter`:
+    :   filter to run
+
+    `args`:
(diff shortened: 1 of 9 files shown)


tarleb:
Another use-case would be to use [different CSL for different sections](https://stackoverflow.com/questions/47551051/how-to-use-a-different-csl-to-every-chapter-in-bookdown).

Issue #3275, opened 2016-12-04
Add a style for tables in docx writer


choies1:
When I convert HTML table to docx using pandoc (ver 1.18), I would like to change the table style for MS-word(docx).

I used the following Pandoc command for conversion (HTML table to docs)
```
pandoc -f html -t docx -S --mathjax conversion_test.html -o conversion_test.doc
```

HTML table code
```html
<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>Account</th>
      <th>Name</th>
      <th>Rep</th>
      <th>Manager</th>
      <th>Product</th>
      <th>Quantity</th>
      <th>Price</th>
      <th>Status</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>0</th>
      <td>714466</td>
      <td>Trantow-Barrows</td>
      <td>Craig Booker</td>
      <td>Debra Henley</td>
      <td>CPU</td>
      <td>1</td>
      <td>30000</td>
      <td>presented</td>
    </tr>
    <tr>
      <th>1</th>
      <td>714466</td>
      <td>Trantow-Barrows</td>
      <td>Craig Booker</td>
      <td>Debra Henley</td>
      <td>Software</td>
      <td>1</td>
      <td>10000</td>
      <td>presented</td>
    </tr>
    <tr>
      <th>2</th>
      <td>714466</td>
      <td>Trantow-Barrows</td>
      <td>Craig Booker</td>
      <td>Debra Henley</td>
      <td>Maintenance</td>
      <td>2</td>
      <td>5000</td>
      <td>pending</td>
    </tr>
  </tbody>
</table>
```
Web browser result:

![html_output2](https://cloud.githubusercontent.com/assets/10408785/20864238/d6fab7c6-ba2b-11e6-8135-3a6fd80bd464.png)

MS Word result (after conversion):

![docx_output](https://cloud.githubusercontent.com/assets/10408785/20864245/0155c04c-ba2c-11e6-936c-79c02ce39318.png)

Howe can I change MS-word(docx) table styles as follows when I convert HTML table to docx using pandoc ?

![docx_output2](https://cloud.githubusercontent.com/assets/10408785/20864249/2b001ece-ba2c-11e6-9c2e-5ad6921e4235.png)

jgm:
Here are the official instructions.  Create a reference.docx:

    pandoc --print-default-data-file reference.docx > myref.docx

[...]

PR #4236, opened 2018-01-10
Docx Writer: add borders in table cell

Meow-z:


cj1324:
Link #3275

Meow-z:
image

mb21:
AFAIK this shouldn't be needed since we have http://pandoc.org/MANUAL.html#custom-styles-in-docx-output

agusmba:
I agree with @mb21 , the possibility to define your default table style should be enough (you can define one with borders if you like).
Of course it will be nice when we are also able to specify a different custom table style in our Markdown as we can do already with divs etc.

agusmba:
@Meow-z specifically have a look at `--reference-doc=FILE` under http://pandoc.org/MANUAL.html#options-affecting-specific-writers

You only need to modify the "Table" table style.

PR #4237, opened 2018-01-10
Add header and footer parameters for wkhtmltopdf

newmana:
Add the ability to use the header and footer parameters in wkhtmltopd.


mb21:
This can already be done using `--pdf-engine-opt`, right?

newmana:
The existing way of passing parameters means that they are all in the same place, you can ensure there’s a flag and value and you have the ability to give a default value (not needed in this case).

newmana:
At the very least maybe add to the documentation that all other parameters to wkhtmltopdf should go through the second channel.

PR #4242, opened 2018-01-10
Add beamer option to show notes on second screen

bersace:
Hi,

To use speaker mode in pdfpc or pympress, i need pandoc to render notes on *second screen*. I suggest to add a beamer specific option for this, since this behaviour is completely specific to beamer.

What do you think of this ?

Thanks for pandoc !

Regards,

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1,6 +1,6 @@
 % Pandoc User's Guide
 % John MacFarlane
-% January 7, 2018
+% January 10, 2018
 
 Synopsis
 ========
@@ -1541,6 +1541,10 @@ LaTeX variables are used when [creating a PDF].
 
 `natbiboptions`
 :   list of options for natbib.
+
+`notes2screen`
+:   In beamer, show notes on second screen. Value can be 'top', 'left', 'right' or
+    'bottom'.
 
 `pagestyle`
 :   An option for LaTeX's `\pagestyle{}`. The default article class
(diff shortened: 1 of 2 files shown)


bersace:
I'm wondering how to document that. Can you give me a clue ? Thanks !

bersace:
Ok, i got it. It's documented in MANUAL :-)

mb21:
Please use pandoc-discuss for questions like this...

jgm:
@mb21 It looks like this is a PR, with some code, not a question.  The PR should be considered.

MANUAL.txt
```diff
@@ -1542,6 +1542,10 @@ LaTeX variables are used when [creating a PDF].
 `natbiboptions`
 :   list of options for natbib.
 
+`notes2screen`
```
jgm: Would it make more sense to call this just `notes`?
Or is there also some way to display notes on the same screen in beamer?

bersace: @jgm I didn't find a best name. notes can be on separate slides.

Maybe we could just add a raw option `shownotes` which is actually just a passthru for `\setbeameroption` ?

Do you have a better idea ? I'm ok with `notes` if you confirm this choice.

jgm: Maybe for maximum generality, something like

```
$if(beameroptions)$
$for(beameroption)$$beameroption$$sep$,$endfor$
$endif$
```

then you could have in your YAML
```
beameroptions:
- show notes on second screen=top
```

bersace: @jgm that's perfect ! Just like classoptions. I'll do this.

bersace:
Closing in favor fo #4359 . Thanks @jgm for the review.

Issue #3675, opened 2017-05-16
md-->reST. Block quote is included in the code block


pashango2:
convert the markdown below to reST, the quoted block will be included in the code block.

````
```python
print("hello")
```
> block quote
````

Output reST code.
```
.. code:: python

    print("hello")

    block quote
```

To terminate the directive explicitly with reST, use an empty comment(..).
The following reST should be output.

```
.. code:: python

    print("hello")

..
    block quote
```

pandoc version: 1.17.1

PR #4259, opened 2018-01-15
in RST writer insert comment between lists and quotes, closes #4248

### LINKED ISSUE #4248: an header in a bullet point followed by a quoted paragraph leads to a syntax error in RST
given the following native document:

```
[BulletList
 [[Header 2 ("bullet",[],[]) [Str "bullet",Space]]]
,BlockQuote [Para [Str "quoted"]]]
```

when converted via Pandoc the result is:

```
-  .. rubric:: bullet
      :name: bullet

    quoted
```

which raises a syntax error when passed for example through `rst2html`, because `quoted` is interpreted as the content of the `rubric` directive wich does not allow content.

note that `quoted` starts aligned with the second dot in the directive block. it's anyway considered _indented_ by `rst2html`.

the [documentation](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#directives) states that:

> Simple directives may not require any content. If a directive that does not employ a content block is followed by indented text anyway, it is an error. If a block quote should immediately follow a directive, use an empty comment in-between (see Comments below).

so something like the following seems to be the right way to write the document:

```
-  .. rubric:: bullet
      :name: bullet
   ..

    quoted
```

Pandoc version 2.1

i am working at a pull request where the empty comment is always added to close a rubric

danse:
Here it is, as you asked. `fixBlocks` here can easily become parametric if desired

### DIFF
--- a/src/Text/Pandoc/Writers/RST.hs
+++ b/src/Text/Pandoc/Writers/RST.hs
@@ -353,9 +353,20 @@ blockListToRST' :: PandocMonad m
                 -> [Block]       -- ^ List of block elements
                 -> RST m Doc
 blockListToRST' topLevel blocks = do
+  -- insert comment between list and quoted blocks, see #4248
+  let fixBlocks (b1:b2@(BlockQuote _):bs)
+        | isList b1 = b1 : commentSep : b2 : fixBlocks bs
+        where
+          isList (BulletList _)     = True
+          isList (OrderedList _ _)  = True
+          isList (DefinitionList _) = True
+          isList _                  = False
+          commentSep  = RawBlock "rst" ""
+      fixBlocks (b:bs) = b : fixBlocks bs
+      fixBlocks [] = []
   tl <- gets stTopLevel
   modify (\s->s{stTopLevel=topLevel, stLastNested=False})
-  res <- vcat `fmap` mapM blockToRST' blocks
+  res <- vcat `fmap` mapM blockToRST' (fixBlocks blocks)
   modify (\s->s{stTopLevel=tl})
   return res
 
--- a/src/Text/Pandoc/Writers/RST.hs
+++ b/src/Text/Pandoc/Writers/RST.hs
@@ -57,7 +57,6 @@ data WriterState =
               , stHasRawTeX   :: Bool
               , stOptions     :: WriterOptions
               , stTopLevel    :: Bool
-              , stLastNested  :: Bool
               }
 
 type RST = StateT WriterState
@@ -68,7 +67,7 @@ writeRST opts document = do
   let st = WriterState { stNotes = [], stLinks = [],
                          stImages = [], stHasMath = False,
                          stHasRawTeX = False, stOptions = opts,
-                         stTopLevel = True, stLastNested = False}
+                         stTopLevel = True }
   evalStateT (pandocToRST document) st
 
 -- | Return RST representation of document.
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Writers/RST.hs
```diff
@@ -353,9 +353,24 @@ blockListToRST' :: PandocMonad m
...
 blockListToRST' topLevel blocks = do
+  let transformLigature b1 b2@(BlockQuote _)
+         | isListBlock b1 = [b1, commentSep, b2]
+         | otherwise      = [b1, b2]
+        where
+          isListBlock (BulletList _)     = True
+          isListBlock (OrderedList _ _)  = True
+          isListBlock (DefinitionList _) = True
+          isListBlock _                  = False
+          commentSep  = RawBlock "rst" ""
```
jgm: I'm confused: shouldn't the comment be `".."` and not `""`?

src/Text/Pandoc/Writers/RST.hs
```diff
@@ -353,9 +353,24 @@ blockListToRST' :: PandocMonad m
...
+      transformLigature b1 b2 = [b1, b2]
```
jgm: This way of writing `transformLigature` causes a new list to be built up for every two elements -- that seems inefficient.

Also, what if the sequence is: BlockQuote, List, BlockQuote? If I understand the logic, `fixBlocks` won't insert a comment between the second and third elements in this case...

Why not something simpler like:
```
fixBlocks (b1:b2:bs)
  | isList b1 && isBlockQuote b2 = b1 : commentsep : b2 : fixBlock bs
fixBlocks (b:bs) = b : fixBlocks bs
fixBlocks [] = []
```

src/Text/Pandoc/Writers/RST.hs
danse: I thought so as well, but tests proved me wrong. Off the top of my mind i imagine that a `RawBlock` renders an empty comment by default. Maybe this is because directives are similar to comments so a custom directive should be started with the same characters that would start an empty comment

src/Text/Pandoc/Writers/RST.hs
danse: > Also, what if the sequence is: BlockQuote, List, BlockQuote? If I understand the logic, fixBlocks won't insert a comment between the second and third elements in this case...

yes you are right i didn't do a good job with pattern-matching there.

> Why not something simpler like ...

cool i updated to adopt your version. `transformLigature` was an attempt at generalising `fixBlocks`. i am not familiar with the use of a single guard in a function body, and i didn't find any reference about its usage online. what happens when the condition is not met? i assume that the next pattern match (at the function signature level) is evaluated.

let me know whether anything else is missing

jgm:
+++ Francesco Occhipinti [Jan 17 18 17:32 ]:

>   cool i updated to adopt your version. transformLigature was an attempt
>   at generalising fixBlocks. i am not familiar with the use of a single
>   guard in a function body, and i didn't find any reference about its
>   usage online. what happens when the condition is not met? i assume that
>   the next pattern match (at the function signature level) is evaluated.

Correct.

src/Text/Pandoc/Writers/RST.hs
```diff
@@ -353,9 +353,20 @@ blockListToRST' :: PandocMonad m
...
 blockListToRST' topLevel blocks = do
+  -- insert comment between list and quoted blocks, see #4248
+  let fixBlocks (b1:b2@(BlockQuote _):bs)
+        | isList b1 = b1 : commentSep : b2 : fixBlocks bs
+        where
+          isList (BulletList _)     = True
+          isList (OrderedList _ _)  = True
+          isList (DefinitionList _) = True
+          isList _                  = False
+          commentSep  = RawBlock "rst" ""
```
jgm: Looking at the writer's code, I still don't see why this gives you `..`.
What happens if you use `RawBlock "rst" ".."`?

danse: Tests fail with a duplicated `..` line. Let's see whether i can find the code

danse: nope i can't find it. as far as i understand [this](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/RST.hs#L236) is the line responsible for rendering a raw block.

maybe the `..` is actually added by [this line](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/RST.hs#L368). if that's true i probably got the code wrong and have to understand that part of the rendering of a quoted block but it's evening here, i'll do tomorrow with a fresher mind

danse: inserting the comment separator before the quote seems to affect `stLastNested` so that `blockToRST'` adds [what looks like a comment](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/RST.hs#L368) before the quote. [this commit](https://github.com/jgm/pandoc/commit/ca77f0a95e03cace027a235ebbc1effa99ea030a) connects this logic with the same problem we are trying to solve now. removing the empty `RawBlock`, my test fails. i have to understand how the status evolves

danse: basically [`blockToRST'`](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/RST.hs#L362-L386) is a version of `fixBlocks` which uses status to terminate the previous block in order to avoid syntax errors. ideally we would have a single function going through the blocks to perform this check, but i'm having an hard time figuring out the meaning of `stLastNested` in order to unify the triggering conditions for an empty comment.

the logic around `stLastNested` seem to be contained mostly in `blockToRST'` linked above. the flag seems to be set most of the times but it's resetted for some specific blocks that we don't want to be followed by an empty comment, apparently.

i believe that the empty raw block before the quote leads to the flag being set and the comment to be added by `blockToRST'`, but i don't understand why lists block don't cause the same to happen.

i will try extending the test cases related to that code in order to get a better understanding while improving the software

danse: once understood the conditions within `blockToRST'` we could be able to make it stateless moving it to `fixBlocks`. it could be more readable and manageable

danse: i've not been able to understand why lists don't fall into the [default pattern match](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/RST.hs#L379) into `blockToRST'`, if they did issue #4248 wouldn't exist. `RawBlocks` matched that pattern i believe, that's why adding a comment would turn out to be redundant in my test.

danse:
so i couldn't move the logic from `fixBlocks` into `blockToRST'`, but i tried moving the logic from `blockToRST'` into` fixBlocks` and it works, all tests pass also those added for #3675. it looks like it makes more sense now :)

jgm:
Looks good, good catch on getting rid of stLastNested.

jgm:
And thanks!

danse:
Thanks to you, Pandoc is powering [Italy](https://github.com/italia) right now and we are proud to be contributing back :+1:

PR #4261, opened 2018-01-15
ImageSize: add derived Eq instance to Dimension

jkr:
@jgm: Any objections? It will allow me to derive Eq on some picture props in the Powerpoint writer. (With given caveat that pixel equality depends on dpi).

### DIFF
--- a/src/Text/Pandoc/ImageSize.hs
+++ b/src/Text/Pandoc/ImageSize.hs
@@ -83,6 +83,7 @@ data Dimension = Pixel Integer
                | Inch Double
                | Percent Double
                | Em Double
+               deriving Eq
 
 instance Show Dimension where
   show (Pixel a)      = show   a ++ "px"


jgm:
My only reservation is that with the derived instance, `Centimeter 10.0` won't be equal to `Millimeter 100.0`.  But maybe this is desirable?

jkr:
I think it would probably be desirable, since it would distinguish between comparing `Dimension` (i.e. something set in attributes) and measurement (which depends on dpi in some cases, meaning of percentage). It might or might not be meaningful that one thing is in inches, the other is in cm. The measurement might be the same, the dimension different.

jgm:
OK, no objections then.

+++ Jesse Rosenthal [Jan 15 18 10:37 ]:

jkr:
Cool -- it's part of another series of changes that I don't quite have working yet, so I'll add it with those.

PR #4262, opened 2018-01-16
HTML reader: Fix col width parsing

n3fariox:
Rather than take user input, and place a "0." in front, actually
calculate the percentage to catch cases where small column sizes are
needed.

[Example File and Results](https://gist.github.com/n3fariox/c8b3ef2e23d81479c8bbce0db1276411)


jgm:
Thanks.

PR #4269, opened 2018-01-16
Add image class to HTML figure. Closes #3928.

### LINKED ISSUE #3928: Link attributes on implicit figures
Hi, figures in their own paragraph get marked up to become implicit figures wrapped in a `<figure>` element. Images can also receive link attributes. When inline it makes sense to apply the atribute to the ``, but I don't think this assumption is ideal for implicit figures:

```
>  pandoc

![Fig. 1](fig.png){.nofloat}

<figure>
<img src="fig.png" alt="Fig. 1" class="nofloat" /><figcaption>Fig. 1</figcaption>
</figure>
```

I think it would be more useful to treat the containing `<figure>` itself as the target for the attributes for implicit figures:

```html
<figure class="nofloat" >
<img src="fig.png" alt="Fig. 1" /><figcaption>Fig. 1</figcaption>
</figure>
```

The implicit figure should treat image and caption as one "unit".

madgen:
Add the class attribute associated with the image to the enclosing
figure with a `-figure` suffix as suggested by @mb21. Closes #3928.

Example input:
```markdown
![a](b){.myClass1 .myClass2 #id}
```

HTML output without the change:
```html
<figure>
<img src="b" alt="a" id="id" class="myClass1 myClass2" /><figcaption>a</figcaption>
</figure>
```

HTML output with the change:
```html
<figure class="myClass1-figure myClass2-figure">
<img src="b" alt="a" id="id" class="myClass1 myClass2" /><figcaption>a</figcaption>
</figure>
```

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -634,12 +634,13 @@ figure opts attr@(_,classes,_) txt (s,tit) = do
   capt <- if null txt
              then return mempty
              else tocapt `fmap` inlineListToHtml opts txt
-  let figureClasses = unwords . map (++ "-figure") $ classes
+  let figureClasses = fromString . unwords $
+        [ "figure" | not html5 ] ++ map (++ "-figure") classes
   return $ if html5
-              then H5.figure ! A.class_ (fromString figureClasses) $ mconcat
-                    [nl opts, img, capt, nl opts]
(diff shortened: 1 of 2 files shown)


madgen:
@jgm: It seems the CI has failed without an actual error (perhaps due to Travis outage yesterday). Could you rerun it please?

madgen:
Is there any chance I can get some feedback on this?

Issue #4111, opened 2017-12-01
Verbatim'd tildes break when exporting to PDF with --listings enabled


lvh:
Pandoc version:

```
> pandoc --version
pandoc 2.0.3
Compiled with pandoc-types 1.17.3, texmath 0.10, skylighting 0.4.3.2
Default user data directory: /Users/lvh/.pandoc
Copyright (C) 2006-2017 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
```

Reproducer markdown file:

```
> cat PandocBugRepro.md
`~`
```

Can't produce a PDF if `--listings` is enabled:

```
> pandoc --listings PandocBugRepro.md -o PandocBugRepro.pdf
Error producing PDF.
! Package Listings Error: lstinline ended by EOL.

See the Listings package documentation for explanation.
Type  H <return>  for immediate help.
 ...

l.61 \passthrough{\lstinline!~!}

```

Works fine without `--listings`:

```
> pandoc PandocBugRepro.md -o PandocBugRepro.pdf
```

Got pandoc from homebrew. Reproduced on 2.0.2, 2.0.1.1. I found this in a document that's been around for a while, and that definitely used to work. Perhaps it's a 2.x change?

OS is fully updated macOS High Sierra:

```
> uname -a
Darwin plotka 17.2.0 Darwin Kernel Version 17.2.0: Fri Sep 29 18:27:05 PDT 2017; root:xnu-4570.20.62~3/RELEASE_X86_64 x86_64
```

LaTeX installation is MacTeX 2016. Will update to 2017 to see if that fixes it, but as mentioned above: I'm pretty sure this used to work and I haven't updated MacTex in a while...

jamtur01:
I have the same issue with listings like:

```latex
\passthrough{\lstinline!__metrics_path__!}
```

Adding _ to the escaped characters resolved it.

```haskell
let str' = escapeStringUsing (backslashEscapes "\\{}%~_") str
```

jgm:
OK, I added `_` as well.

+++ James Turnbull [Dec 01 17 22:04 ]:

dotlambda:
How about adding `&` as well?

PR #4271, opened 2018-01-16
LaTeX writer: escape & in lstinline

dotlambda:
I get an error when I've got a backticked `&` in my document and `--listings` enabled.
```
! Improper alphabetic constant.
<to be read again> 
                   \unskip 
l.69 \passthrough{\lstinline!&!}
```
Strangely, it only seems to happen inside tables like
```
col1 col2
---- ----
`&`  abc
```
Also see #4111.


mb21:
What command are you using? What pandoc version? Does it go away with `latex_macros` disabled?

jgm:
Thanks!

PR #4273, opened 2018-01-16
ConTeXt writer: Print also "tex" and "latex" raw elements

### LINKED ISSUE #3969: Context output ignores raw_tex when it is alone in a paragraph
I have just updated to the latest (source) version of pandoc and have found that since V 1.19.2.1 the treatment of raw_tex has changed.

It seems that a tex command on a line of its own is now ignored in context, but if anything else is on the line then it is included.

If I build the file (test.md)

```
\placefloats

\placefloats

Here \placefloats 

```

with 

```
pandoc.exe -f markdown+raw_tex+latex_macros -t context -o test.tex test.md
```

I get
```
Here \placefloats 
```

When I run it using V 1.19.2.1 I get

```
\placefloats

\placefloats

Here \placefloats
```
as I would expect. The latex writer does include them and the RawInline elements are visible in the json output so it seems that it is a problem with the context writer.

hmenke:
Fixes #3969


hmenke:
Hm, this needs more thought...

PR #4291, opened 2018-01-22
ConTeXt writer: xtables: correct wrong usage of caption

### LINKED ISSUE #4289: ConTeXt: startplacetable "caption" vs "title"
Pandoc Version: 2.1.1
Commandline: pandoc -t context -s test.md -o test.tex && context test.tex

```
+-----+-----+
| Foo | Bar |
+-----+-----+

Table: Some Caption
```

Result: no caption visible in PDF
Expected: a caption under the table.

Reason:
In the generated TeX it reads `\startplacetable[caption={Some Caption}]` while it should be `\startplacetable[title={Some Caption}]`.

hmenke:
Fixes #4289


jgm:
Thanks for the quick fix.

PR #4292, opened 2018-01-23
ConTeXt writer: Print also tex and latex raw elements

### LINKED ISSUE #3969: Context output ignores raw_tex when it is alone in a paragraph

hmenke:
Fixes #3969 

This blindly prints out any raw blocks which are marked `tex` or `latex` in the ConTeXt writer.  The rationale behind this is that people probably do not write inline markup and then have ConTeXt *and* LaTeX output generated from this.  I think this is a reasonable approach because the conversion of foreign inline markup is highly non-deterministic and as a user I wouldn't want to rely on pandoc reading my mind and doing the right thing.

This is of course open for discussion and I would especially like to hear voices of people who *do* use inline markup and multiple output.
@alnjxn @aksdb @ousia @DaveJarvis @adunning @PeterDavidson

### DIFF
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -188,6 +188,8 @@ blockToConTeXt (CodeBlock _ str) =
   return $ flush ("\\starttyping" <> cr <> text str <> cr <> "\\stoptyping") $$ blankline
   -- blankline because \stoptyping can't have anything after it, inc. '}'
 blockToConTeXt (RawBlock "context" str) = return $ text str <> blankline
+blockToConTeXt (RawBlock "latex" str) = return $ text str <> blankline
+blockToConTeXt (RawBlock "tex" str) = return $ text str <> blankline
 blockToConTeXt b@(RawBlock _ _ ) = do
   report $ BlockNotRendered b
   return empty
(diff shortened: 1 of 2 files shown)


mb21:
I agree with printing `RawBlock "tex" str`. But outputting something that's explicitly marked `RawBlock "latex" str` in the ConTeXt writer feels wrong. (The suggested approach in #3969 is to change the reader to parse to `tex` instead of `latex`.)

jgm:
+++ Mauro Bieg [Jan 23 18 09:12 ]:
>   I agree with printing RawBlock "tex" str. But outputting something
>   that's explicitly marked RawBlock "latex" str in the ConTeXt writer
>   feels wrong. (The suggested approach in [1]#3969 is to change the
>   reader to parse to tex instead of latex.)

Agreed, I prefer the suggested approach in #3969.

DaveJarvis:
@PeterDavidson wrote:
> but it would seem wrong to include clearly identified latex in a context writer.

@hmenke wrote:
>  the conversion of foreign inline markup is highly non-deterministic

Agreed, as well. There's no way for Pandoc to determine what commands are intended for what TeX processor. Raw TeX should be output verbatim.

Aside, I avoid inlining TeX in Markdown because it seems like mixing content and presentation.

jgm:
Closing this, we should instead change the markdown reader so it uses "tex" instead of "latex" consistently.

PR #4295, opened 2018-01-23
ConTeXt writer: new section syntax and --section-divs

### LINKED ISSUE #2609: Support for start/stop style headings with ConTeXt writer?
At the moment, I'm only able to coax pandoc into generating ConText documents with headings like:  

```
\section[foo]{Foo}
```

Would it be possible to also support the support the (newer, nesting) start/stop syntax for headings in ConTeXt IV? Hence:

```
\startsection[title={Foo}]
  Lorem ipsum, bananas foster, eels.
\stopsection
```

Or, for give me if this is so: is there already a way to get pandoc to do this?

hmenke:
Fixes #2609.

This PR introduces the new-style section headings.
```latex
\section[my-header]{My Header} -> \section[title={My Header},reference={my-header}]
```
On top of this, the ConTeXt writer now supports the `--section-divs` option to write sections in the fenced style
```latex
\startsection[title={My Header},reference={my-header}]

contents...

\stopsection
```

@pmazaitis @ousia @DaveJarvis

### DIFF
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -159,8 +159,9 @@ elementToConTeXt :: PandocMonad m => WriterOptions -> Element -> WM m Doc
 elementToConTeXt _ (Blk block) = blockToConTeXt block
 elementToConTeXt opts (Sec level _ attr title' elements) = do
   header' <- sectionHeader attr level title'
+  footer' <- sectionFooter attr level
   innerContents <- mapM (elementToConTeXt opts) elements
-  return $ vcat (header' : innerContents)
+  return $ header' $$ vcat innerContents $$ footer'
 
 -- | Convert Pandoc block element to ConTeXt.
 blockToConTeXt :: PandocMonad m => Block -> WM m Doc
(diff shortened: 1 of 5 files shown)


hmenke:
This is not yet ideal.  This piece of markdown
```markdown
# Section
## Subsection
### Subsubsection
#### Subsubsubsection
##### Subsubsubsubsection
###### Subsubsubsubsubsection
####### Subsubsubsubsubsubssection
```
translates to
```latex
\section[title={Section},reference={section}]

\subsection[title={Subsection},reference={subsection}]

\subsubsection[title={Subsubsection},reference={subsubsection}]

\subsubsubsection[title={Subsubsubsection},reference={subsubsubsection}]

\subsubsubsubsection[title={Subsubsubsubsection},reference={subsubsubsubsection}]

\

\
```
The problem is not so much in the spurious backslashes but rather in the fact that even with the original behaviour section levels beyond 5 were silently discarded.

What should be the desired behaviour here?  I would go for simply generating all those lower levels even if they are not predefined in ConTeXt.  If the user wants to use such deep nesting they should just adapt their preamble.

DaveJarvis:
> What should be the desired behaviour here? I would go for simply generating all those lower levels even if they are not predefined in ConTeXt. If the user wants to use such deep nesting they should just adapt their preamble.

There are a number of possibilities:

* Define `\stop` and `\start` environments in the standalone preamble for the number of subsections encountered in the document that go beyond ConTeXt's default depth? This way producing a standalone document still works. It's a weird edge case.
* Exit with an error code stating that the number of nested levels is not yet supported.
* Introduce a `\startsubject` section, though this leaves a few issues: the subject sections are unnumbered and they wouldn't appear in the table of contents by default.
* Create custom title levels so `######### Title` generates a preamble with `\definehead[section9][section]` (where `9` indicates the title level depth).

hmenke:
@DaveJarvis 

1. Writing defintions for the deeply nested sections to the preamble is certainly possible but only works for standalone document which include the preamble.  Also it is not the prettiest solution.
2. Definitely no.
3. You can use pandoc also with unnumbered sections so this would clash.
4. Isn't this the same as 1.?

I would just generate `subsub...subsub` and let the user worry about that.  As you already mentioned, it's a weird edge case.

DaveJarvis:
> I would just generate subsub...subsub and let the user worry about that. There is also the subsub module...

That's a good idea. Perhaps output a warning, then, directing the user to see the subsub module in the event they have nested [too deeply](https://youtu.be/PfhZB7rQ7iA?t=11)?

jgm:
+++ Henri Menke [Jan 23 18 15:20 ]:
>   What should be the desired behaviour here? I would go for simply
>   generating all those lower levels even if they are not predefined in
>   ConTeXt. If the user wants to use such deep nesting they should just
>   adapt their preamble.

That sounds okay to me.

hmenke:
@jgm Done.

jgm:
Thanks, merged!

Issue #1843, opened 2014-12-28
Extracting more document structure from docx files


elliottslaughter:
Many thanks for the introduction of the docx reader in 1.13! This improves some of my workflows significantly. Unfortunately, I have a few use cases that do not work well under the current docx reader.

Typically, these cases are the result of extensive use of templates where all of the styles are based on the "Normal" style, as opposed to being on, say, "Heading 1" something more semantically appropriate. As a result, Pandoc is able to extract minimal semantic information about the text. This is completely understandable, but nevertheless frustrating, because the user must manually re-apply the desired markup to the resulting Markdown.

An initial proposal might be to include more of this information in the output, i.e. essentially tagging all the output content with the name of the style(s) applied to it. So for example, a paragraph styled as "AChapterTitle" might result in the following Markdown:

```
<div class="AChapterTitle">

This is a chapter title

</div>
```

The downside of doing this is of course that it makes the resulting Markdown that much messier (though arguably this only affects files with extensive use of user-defined non-semantic styles). But the upshot is that it makes it possible for the user to write a filter---which knows the mapping from template styles to Markdown constructs---to clean up the file and restore the markup to a more semantic state. This approach adds some implementation complexity to the docx reader, but otherwise doesn't complicate Pandoc itself much if at all.

A more sophisticated alternative might be to pull the filter into Pandoc itself; i.e. to provide an API with special support for Word documents in the docx reader, and then provide a special --input-filter flag in Pandoc which would connect via this new API to the filter, which would then be able to look at the structure of the Word document and tell the docx reader directly what styles correspond to which semantic properties. This obviously adds additional implementation complexity to both the docx reader and to Pandoc, but has the advantage of avoiding information loss in the initial conversion to Markdown.

Or perhaps there might be alternative ways to extract this information. I would be happy with any approach that avoids throwing away the information entirely.

elliottslaughter:
@jkr Your diff would be sufficient for my needs. It exposes enough information, and it actually isn't *that* verbose compared to some of the HTML I was looking at from the same source. If something along these lines makes it into the next release, I'll be satisfied.

elliottslaughter:
@jkr Actually, I lied. Turns out that the document has inline styles as well. This ends up being a double-whammy because the style is question is a templated used for inline code, and in addition to not getting the style itself, I'm also missing some leading spaces used as indentation. Is there any way the approach could be tweaked to handle this as well?

jgm:
[...]

Issue #2106, opened 2015-04-22
Custom styles for ODT (and maybe LaTeX and RST) writers


matthijskooijman:
Currently, you can customize formatting by providing some customized styles in a reference odt/docx file to pandoc. These styles are copied into the generated document and applied to selected elements in the document.

However, this only supports a limited, hard-coded list of styles. E.g. the 'Heading 1' style is applied to level-1 headings, the 'Text body' style to regular paragraphs, the 'Bullet Symbols' style to bullets. On top of this, I'd like to be able to specify other styles to use, on a block-by-block basis.

My usecase is that I'm writing a book, and my publisher requires the text to be delivered in docx or odt format, using a set of custom styles with non-standard style names. I have two challenges:
1. Regular styles have different names, so I need to use e.g. "Publisher Heading 1" for first-level headings.
2. Some parts of my document must use completely different styles, which have no equivalent in the pandoc AST (e.g. notes for the layouter must use the "Layout Notes" style).

If I have some way to specify custom style names in the AST, I can use a pandoc filter to introduce the proper style names in the AST and have the writers use the right styles. Ideally, I'd be able to specify the style names for challenge 2 above directly in my markdown source.

Attributes seem like a good fit for this, so that would work out to something like this in markdown:

```
# My Header {style="Fancy Header 1"}
```

Note that this is already supported by the current markdown parser, and results in this JSON:

```
[{"unMeta":{}},
 [{"t":"Header","c":[1,["my-header",[],[["style","Fancy Header 1"]]],
 [{"t":"Str","c":"My"},{"t":"Space","c":[]},{"t":"Str","c":"Header"}]]}]
]
```

However, converting this to ODT or DOCX drops the attributes completely.

I'm aware that not all elements currently allow specifying styles, but that's another discussion. What I'd like to see is that, for all elements that currently support attributes, the "style" attribute be interpreted as a style name in the ODT and DOCX writes.

ODT and DOCX have a distinction between paragraph and character styles, but it seems that using the paragraph styles  for block elements and character styles for inline elements is sufficient?

[...]

PR #4299, opened 2018-01-24
RFC: Verbose docx

jkr:
As discussed on #1843:

We still resolve dependent styles, but we optionally wrap elements in div/span tags derived from paragraph/character style names. This is controlled by `--verbose-docx` which defaults to off.

~~~
$ pandoc test/docx/german_styled_lists.docx -t markdown
-   One level of the list.

-   Second level of the list.

    -   Next level of the list

-   Back to the top level.
~~~

~~~
$ pandoc --verbose-docx test/docx/german_styled_lists.docx -t markdown
-   ::: {.AufzhlungStrich}
    One level of the list.
    :::

-   ::: {.AufzhlungStrich}
    Second level of the list.
    :::

    -   ::: {.AufzhlungStrich}
        Next level of the list
        :::

-   ::: {.AufzhlungStrich}
    Back to the top level.
    :::
~~~

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1974,6 +1974,16 @@ input formats
 output formats
 :  `markdown`, `docx`, `odt`, `opendocument`, `html`
 
+#### Extension: `styles` #### {#ext-styles}
+
+Read all docx styles as divs (for paragraph styles) and spans (for
+character styles) regardless of whether pandoc understands the meaning
+of these styles. This can be used with [docx custom
+styles](#custom-styles-in-docx). Disabled by default.
+
+input formats
+:  `docx`
+
 #### Extension: `amuse` ####
 
 In the `muse` input format, this enables Text::Amuse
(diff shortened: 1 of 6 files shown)


jkr:
As mentioned in the issue, the option name should probably be changed. `--read-all`? `--keep-unknown`?

mb21:
What about using an extension like `-f docx+styles` instead of an option?

jkr:
Sure, that would be fine too -- but I have to admit that I've never been quite sure of the semantic difference between options and extensions. It's an extension if it only covers one input/output format? Or something deeper?

jgm:
One difference is that extensions can be different on input and output formats. When `--smart` was an option, it affected both; now that it's an extension, you can enable it on input without enabling it on output, for example.

nokome:
Just to +1 this. 

I have done some limited testing on this branch and it worked well (using a filter implemented in Node.js to handle certain custom styles). 

One thing I was wondering about is whether instead of using a wrapping `Div` e.g.

```html
<div data-custom-style="TextBody">
<p>Paragraph 1</p>
</div>
<div data-custom-style="TextBody">
<p>Paragraph 2</p>
</div>
```
is it possible to apply the paragraph styles to the paragraphs themselves? e.g.

```html
<p data-custom-style="TextBody">Paragraph 1</p>
<p data-custom-style="TextBody">Paragraph 2</p>
```

Also, I agree that `-f docx+styles` seems more appropriate than `--verbose-docx`.

jkr:
Okay -- I think I'm going to go with `-f docx+styles`, default off. @jgm, does that sound good to you?

jgm:
Sounds okay - just make sure it's appropriately documented.
(And, does it only affect docx input?  Is there a sensible
way to make it affect output?)

+++ Jesse Rosenthal [Feb 22 18 14:57 ]:
>   Okay -- I think I'm going to go with -f docx+styles, default off.
>   [1]@jgm, does that sound good to you?

jkr:
I did a fair amount of documentation. I also described a bit how to make
a better docx -> edit-in-markdown -> docx workflow (using the input docx
as a reference-doc and putting the custom-styles back into the output).

I'm thinking of making a standalone doc describing that in more detail
at some point soon.

The only way I could think of making it apply to output is to make the
use of custom-styles dependent on it. But that would probably only make
sense if we made it disabled by default for the sake of symmetry (`-f
docx+styles -t docx+styles`). Otherwise it could be pretty
confusing. But that would change existing CLI behavior.

We could also expand it to other writers: i.e.,

-t latex+styles:

    ::: {custom-style=foo}

    bar

    :::

becomes
    
    \begin{foo}
    bar
    \end{foo}

I know there's a long thread on generalized custom-styles that somewhere
and this is one of the proposed solutions, so this might be a way toward
giving some users what they want there.

jgm:
+++ Jesse Rosenthal [Feb 26 18 17:59 ]:
>   We could also expand it to other writers: i.e.,
>   -t latex+styles:
>   ::: {custom-style=foo}
>   bar
>   :::
>   becomes
>   \begin{foo}
>   bar
>   \end{foo}
>   I know there's a long thread on generalized custom-styles that
>   somewhere
>   and this is one of the proposed solutions, so this might be a way
>   toward
>   giving some users what they want there.

It might be worth opening a discussion on pandoc-discuss
about this, or perhaps on the linked issue about custom
latex environments, or both.

mb21:
Indeed, that discussion is https://github.com/jgm/pandoc/issues/2106

Not sure if the new `+styles` extension here is the same as the discussed `+environments` there?

PR #4304, opened 2018-01-25
MANUAL.txt: self-contained implies standalone

quite:


jgm:
The man page is generated from MANUAL.txt, so you should modify MANUAL.txt instead.
(The man page can then be generated using `make man/pandoc.1`.)

quite:
Argh, did the same somewhere else recently

jgm:
thanks!

Issue #4302, opened 2018-01-25
[feature request] autolinebreak for fenced code under skylighting


jannick0:
Pandoc 2.0.6 / skylighting 0.5.0.1:

It appears that for fenced code blocks with long lines there is no autolinebreak feature (for style pygments, at least).

**Example:**
`pandoc -f markdown -t latex -o long-code-line.pdf long-code-line.md`
~~~~
long-code-line.md:
~~~
   this is a very long long long long long long long long long long long long long line which is not broken
~~~
~~~~

NB: Options of the LaTeX [listings package](https://ctan.org/pkg/listings):
- breaklines (boolean)
- breakatwhitespace (boolean)
- breakautoindent
- linewidth

Apologies if there is a trivial solution to this issue making this feature request obsolete.

jgm:
I don't think this can be implemented in skylighting itself, because skylighting has no way of knowing how much space each token takes and hence when the end of the line is reached.

If there's some latex magic that can cause auto wrapping inside a custom verbatim environment defined using fancyvrb, then we could use that.

jannick0:
These edits in the preamble would work, too:
~~~tex
\usepackage{fvextra}
\DefineVerbatimEnvironment{Highlighting}{Verbatim}{breaklines,commandchars=\\\{\}}
~~~

georgerbr:
> These edits in the preamble would work, too:
> 
> ```tex
> \usepackage{fvextra}
> \DefineVerbatimEnvironment{Highlighting}{Verbatim}{breaklines,commandchars=\\\{\}}
> ```

How about prepending to `verbatim` instead of creating a custom environment?

```tex
% single line spacing for code
\usepackage{etoolbox}
\BeforeBeginEnvironment{verbatim}{\def\baselinestretch{1}}
```

jannick0:
`baselinestretch` changes the spacing in **everything** of the document **after** its has been triggered. That is not the aim of the proposed feature, namely line beaks of long lines in fenced code blocks. So I can not see how this would help solving the issue.

BTW - my suggestion would amend an already existing 'custom' environment created by the current version of `pandoc`. The adjustment is tiny, and it fully addresses the issue.

Or am I missing anything here, since I am lightyears away from being a LaTeX expert?

georgerbr:
[...]

PR #4314, opened 2018-01-27
[latex] forced linebreaks for fenced code using

jannick0:
Per suggestion in #4302. The patch makes sure that long lines of fenced code are broken (LaTeX). The solution requires the additional package [fvextra](https://ctan.org/pkg/fvextra) which nicely collaborates with `fancyvrb`.

@jgm I hope this is the correct place to make these innocent edits. I am not sure if additional commits are needed in the documentation area.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -211,7 +211,7 @@ When using LaTeX, the following packages need to be available
 (they are included with all recent versions of [TeX Live]):
 [`amsfonts`], [`amsmath`], [`lm`], [`unicode-math`],
 [`ifxetex`], [`ifluatex`], [`listings`] (if the
-`--listings` option is used), [`fancyvrb`], [`longtable`],
+`--listings` option is used), [`fancyvrb`], [`fvextra`], [`longtable`],
 [`booktabs`], [`graphicx`] and [`grffile`] (if the document
 contains images), [`hyperref`], [`xcolor`] (with `colorlinks`),
 [`ulem`], [`geometry`] (with the `geometry` variable set),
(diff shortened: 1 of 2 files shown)


jgm:
I'm a bit reluctant to depend on a semi-obscure package like fvextra.
Maybe we could load it if it's available, and only apply this command in that case?
Then people would get the breaking behavior if they have fvextra available, otherwise not.
Or we could put the whole thing under a variable, so people could select `break-code-lines`.

jannick0:
I am not sure what you mean by 'semi-obscure', while the package `fvextra` with precious features is available at ctan and is downloadable for everyone like `verbatim` and `fancyvrb`. Anyways, a flag like `break-code-lines` would be just fine.

Another possible workaround until feature implementation: Insert this header snip (via `--include-in-header`)
```tex
\usepackage{fvextra}
\RecustomVerbatimEnvironment{Highlighting}{Verbatim}{breaklines,commandchars=\\\{\}}
```
Thanks.

jgm:
+++ jannick0 [Jan 28 18 19:58 ]:
>   I am not sure what you mean by 'semi-obscure', while the package
>   fvextra with precious features is available at ctan and is downloadable
>   for everyone like verbatim and fancyvrb. Anyways, a flag like
>   break-code-lines would be just fine.

For e.g. debian users, it can be inconvenient if pandoc uses
a package that isn't in texlive-latex-recommended, because
then you may need to install texlive-latex-extra just for
this...

Not a huge issue, perhaps, but I've tried to restrict to
commonly used packages.

jannick0:
Understood. Then a switch for the user's choice and taste looks doable which would not force her / him to apply the package.

Issue #4320, opened 2018-01-29
table contents are not affected by the wrap option


danse:
[this test](https://github.com/italia/pandoc/commit/11cb8a60a6974605d7dbd97ee42e66d262e3bf29) features a table with a very long word wrapped in inline markup. when the corresponding native file gets converted into RST, the output from pandoc is:
```
+---------------------------------+------------------------------------+
| one                             | two                                |
+=================================+====================================+
| ports                           |     **thisIsGoingToBeTooLongAnyway |
|                                 | **                                 |
+---------------------------------+------------------------------------+
```
which contains a syntax error due to the way the inline markup gets wrapped to the next line breaking the block indentation.

the test linked above contains a solution proposal. users could add `--wrap=none` from the command line in order to produce longer tables without errors, giving priority to syntax soundness over readability

i imagine that another possible area of improving is the wrapping logic, but i wouldn't know how to correctly wrap those lines preserving the markup

danse:
reading [the code](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/Shared.hs#L245-L261) i found out that, while `wrap` seems not to be considered when rendering a table, the `columns` option can be used to affect the column width of tables. so currently a large number could be passed to `columns` as a workaround to prevent syntax errors

jgm:
Yes, I suppose we could have wrap=none affect table cells as
well, but then we run into incompatibilities since many
table formats specify the relative cell widths (relative to
--columns).  We either have to wrap or truncate the text or
automatically increase the column width for just that table
(I suppose you had in mind the last).

danse:
yes i had in mind to extend tables, as wrapping is cause of the error and truncating would damage the document too much. while gaining a better understanding of gridTable i thought it was a good idea to [add comments to the code and factor out some entities](https://github.com/italia/pandoc/blob/4320/src/Text/Pandoc/Writers/Shared.hs#L242-L283). i hope that it will guide the next person reading this logic. since i should touch gridTable i would take the chance to refactor it to a way that is more readable in my opinion.

i think that widths could be calculated with guards on conditions related to options and table specs. i would add a guard whose returned widths would be calculated from rows and headers in order to have a table without wrappings.

then the widths would be zipped with the column cells and passed to `blocksToDoc`, outside the recursion used currently. if this refactoring seems too invasive and not worth it, i'll add the logic for "full widths" within the current structure

danse:
[...]

PR #4326, opened 2018-01-31
4320

danse:
here is a solution for #4320. i organised the changes in two commits but i can rebase if desired :)


danse:
rebased

PR #4335, opened 2018-02-01
Remove extraneous, significant whitespace in JATS writer output

nokome:
The JATS writer produces more readable XML by using indentation. However, it was inappropriately indenting the content of elements which may contain `#PCDATA`, thereby adding extraneous, significant whitespace. In addition, it had wrapping turned on which added more whitespace.

I checked all the uses of `inTagsIndented` in `src/Text/Pandoc/Writers/JATS.hs` against the [JATS spec](https://jats.nlm.nih.gov/publishing/tag-library) an fixed those tags which can have `#PCDATA` content:

tag | status
:---| :-----
`<caption>` | OK
`<def-item>` | OK
`<def>` | OK
`<disp-quote>` | OK
`<fn>` | OK
`<label>` | fixed
`<list-item>` | OK
`<p>` | fixed
`<ref-list>` | OK
`<tbody>` | OK
`<td>` | fixed
`<term>` | fixed
`<th>` | fixed
`<thead>` | OK
`<tr>` | OK


src/Text/Pandoc/Writers/JATS.hs
```diff
@@ -69,11 +69,8 @@ docToJATS opts (Pandoc meta blocks) = do
   let (backblocks, bodyblocks) = partition isBackBlock blocks
   let elements = hierarchicalize bodyblocks
   let backElements = hierarchicalize backblocks
-  let colwidth = if writerWrapText opts == WrapAuto
-                    then Just $ writerColumns opts
-                    else Nothing
   let render'  :: Doc -> Text
-      render'  = render colwidth
+      render'  = render Nothing
```
jgm: Thanks for this PR!  One comment on this line. This disables wrapping entirely.  I don't think we need to do anything that drastic.  Text.Pandoc.Pretty provides a `nowrap` combinator which can locally disable wrapping when that is absolutely needed.  But normal wrapping, which occurs on Space elements, should be innocuous, so you should only need the `nowrap` in a few special cases (e.g. verbatim).

nokome: Thanks for the review @jgm. I've reinstated the wrapping now.

nokome:
@jgm : I reinstated wrapping with 10eb42d which broke the tests but which illustrates the issue of introducing extra whitespace when both indentation and wrapping enabled as currently implemented. Here's an example from the breaking test in `test/writer.jats`:

#### Without wrapping:
```xml
<sec id="paragraphs">
  <title>Paragraphs</title>
  <p>Here’s a regular paragraph.</p>
  <p>In Markdown 1.0.0 and earlier. Version 8. This line turns into a list item. Because a hard-wrapped line in the middle of a paragraph looked like a list item.</p>
  <p>Here’s one with a bullet. * criminey.</p>
  <p>There should be a hard line break<break />here.</p>
</sec>
```

#### With wrapping:
```xml
<sec id="paragraphs">
  <title>Paragraphs</title>
  <p>Here’s a regular paragraph.</p>
  <p>In Markdown 1.0.0 and earlier. Version 8. This line turns into a list
  item. Because a hard-wrapped line in the middle of a paragraph looked like a
  list item.</p>
  <p>Here’s one with a bullet. * criminey.</p>
  <p>There should be a hard line break<break />here.</p>
</sec>
```

Note that because the `<p>` is a child of `<sec>` that an extra 2 significant spaces are inserted at the start of each wrapped line.

I'm afraid I don't have a good enough knowledge of Haskell or Pandoc to be able to fix this myself.

jgm:
So, is it the case that in JATS (unlike DocBook or HTML),
there is a semantic difference between two spaces and one
space inside a `<p>` element?

Where is this documented?

+++ Nokome Bentley [Feb 26 18 00:31 ]:
>   Note that because the <p> is a child of <sec> that an extra 2
>   significant spaces are inserted at the start of each wrapped line.

nokome:
You're right, I wrongly assumed there was a semantic difference but that might not be the case. I did some searching but found it difficult to find clear documentation on this matter in the JATS standard https://jats.nlm.nih.gov/archiving/tag-library/1.1d1/. 

The best I found was the "Whitespace Handling" section in https://www.ncbi.nlm.nih.gov/books/NBK425547/ which says:

> XML processors may remove all insignificant white space but must retain one space where whitespace is significant

Which implies that processors usually combine more than one space into one space. Pinging @michael and @oliver---- who may be able to shed some light on the matter.

Either way, as you say a user can turn off wrapping locally, so I'm happy to defer to your judgement and adjust  the test case so that it with wrapping on.

jgm:
I'd prefer it if the test case had wrapping, because that would be consistent with the other writer tests.
This also allows us to make sure that whitespace is treated appropriately even when there's wrapping.

nokome:
@jgm: I have reintroduced wrapping into the test case and all tests are now passing.

jgm:
Great, thanks for the patch!

PR #4348, opened 2018-02-04
Update default.nix to current nixpkgs-unstable for hslua-0.9.5

jarlg:
A month ago bdb911550c2894925be6893a5d93fca484448dd4 bumped `hslua` to version 0.9.5, but the change was not incorporated into `default.nix`. Consequently, `nix-build` fails while complaining about conflicting versions. Updating to a more recent nixpkgs channel resolves this, and the previously specified overrides are no longer necessary.

The revision was taken from [nixos.org/channels/nixpkgs-unstable/git-revision](nixos.org/channels/nixpkgs-unstable/git-revision) and the sha256 hash can be double-checked by running `nix-prefetch-git https://github.com/nixos/nixpkgs.git --rev 42b9b8f7c8687cb26e69c3559e0e1346fb0e680f` and inspecting the output.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -9,16 +9,11 @@
 { pkgs ?
     import ((import <nixpkgs> {}).pkgs.fetchFromGitHub {
       owner = "NixOS"; repo = "nixpkgs";
-      rev = "1354099daf98b7a1f79e6c41ce6bfda5c40177ae";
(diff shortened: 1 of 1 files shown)


jgm:
Thanks.  I don't use nix, so any updates to default.nix are
going to have to come from people who do.

Issue #4355, opened 2018-02-08
"TeX capacity exceeded" with `--pdf-engine=xelatex -V lang:ar`


mustafa0x:
```
$ pandoc --pdf-engine=xelatex -M mainfont="Amiri" -V lang:ar test.md -o test.pdf
Error producing PDF.
! TeX capacity exceeded, sorry [input stack size=5000].
<to be read again> 
                   \use_i:nn 
l.69 \end{document}
```

Version:
```bash
$ pandoc -v
pandoc 2.1.1
Compiled with pandoc-types 1.17.3, texmath 0.10.1, skylighting 0.6
```

Steps (using a fresh Ubuntu VPS with 16gb of ram):
```bash
# Install xetex
apt install texlive-xetex texlive-lang-arabic texlive-math-extra

# Install Amiri font
apt-add-repository universe
apt update
apt install fonts-hosny-amiri

# Create test file
echo 'هذه تجربة' > test.md
```

---

The issue doesn't seem to be `xelatex`, as I ran `xelatex ar.tex` on the following and it created the PDF file:
```tex
\documentclass[12pt,fleqn,titlepage,twoside,a4paper]{book}
\usepackage[centering,includeheadfoot,margin=1in]{geometry}
\usepackage{arabxetex}
%\newfontfamily{\arabicfont}[Script=Arabic,Scale=1.5]{Amiri}

\parindent = 0pt

\begin{document}

\begin{arab}[utf]
هذه تجربة
\end{arab}

\end{document}
```

---

Adding `-V dir:rtl` gave me another issue, but I doubt it's related:

```
$ pandoc --pdf-engine=xelatex -M mainfont="Amiri" -V lang:ar -V dir:rtl test.md -o test.pdf
Error producing PDF.
! LaTeX Error: Option clash for package bidi.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.65   \fi
```

jgm:
Using --verbose will give you the intermediate tex that
pandoc is using; this might help to diagnose the problem.

mustafa0x:
```tex
\PassOptionsToPackage{unicode=true}{hyperref} % options for packages loaded elsewhere
\PassOptionsToPackage{hyphens}{url}
%
\documentclass[arabic,]{article}
\usepackage{lmodern}
\usepackage{amssymb,amsmath}
\usepackage{ifxetex,ifluatex}
[...]

PR #4357, opened 2018-02-09
Provide `bidi` package's option using `\PassOptionsToPackage`.

wilx:
This avoids option clash when `polyglossia` loads it first and then it is loaded again for XeLaTeX when `latex-dir-rtl` defined.

This fixes `Option clash for package bidi` issue observed in #4355.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -2,11 +2,9 @@
 \PassOptionsToPackage{hyphens}{url}
 $if(colorlinks)$
 \PassOptionsToPackage{dvipsnames,svgnames*,x11names*}{xcolor}
-$endif$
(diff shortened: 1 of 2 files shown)


wilx:
The build failure is only stack nightly failing due to dependencies:

```
Configuring pandoc-2.1.1...
Error:
    The following packages are broken because other packages they depend on are missing. These broken packages must be rebuilt before they can be used.
installed package haddock-library-1.4.5 is broken due to missing package haddock-library-1.4.5-JMywL3gQR7h8vuSGU4Yu5w-attoparsec
```

fauno:
i can confirm this works, thanks!

PR #4370, opened 2018-02-15
remove multiple inline styles when writing RST, closes #4368

### LINKED ISSUE #4368: nested inline styles turn into extra RST text
a block like `[Emph [Str "first", Strong [Str "second"]]]` gets translated by pandoc into `*first\ **second***` which is not an accurate RST translation because [RST does not support nested inlines](http://www.sphinx-doc.org/en/stable/rest.html)

this does not represent a syntax error but it causes extra text to be displayed to the user when RST is converted to other formats like HTML

we propose to ignore the innermost styles, in order to preserve the style given to a broader section of the text

danse:


danse:
closing because we [want to do better](https://github.com/jgm/pandoc/issues/4368#issuecomment-366205802)

danse:
build failed but that seems unrelated to the changes here

danse:
i just tested this on more files and i spotted some errors, i will turn the errors into unit tests and then require again

Issue #3930, opened 2017-09-18
pandoc does not use different sizes if same image is included several times (md -> docx)


agusmba:
The same image included several times appears with the same size as the first image (markdown to docx conversion):

Tested with Pandoc versions **1.19.2.1** and **nightly** from 18/09/2017
OS: Windows 7 Enterprise.

I'll edit to attach a small markdown file and image.
Converted with:

```bash
$ pandoc -s pandoc-issue-xxx.md -o pandoc-issue-3930.docx
```

Markdown attachment is not supported, I'll zip it. but here it is in source:

```markdown
# Pandoc issue #3930

The same image included several times appears with the same size as the first image
 (tested in markdown to docx conversion):

Image 5cm wide: `![test-img](./test-img.png){width=5cm}`

![test-img](./test-img.png){width=5cm}

Image 10cm wide: `![test-img2](./test-img.png){width=10cm}`

![test-img2](./test-img.png){width=10cm}

The same happens if the image is jpeg.

```

[pandoc-issue-3930.zip](https://github.com/jgm/pandoc/files/1311146/pandoc-issue-3930.zip)

Screenshot of the resulting incorrect docx:

![image](https://user-images.githubusercontent.com/3104020/30549313-bf14a774-9c94-11e7-864e-1204f55d0580.png)

agusmba:
This probably has low severity.

As a workaround, I could use a copy of the image if I really needed to use different sizes for the same image, but still this could be a problem in some corner cases.

Issue #3798, opened 2017-07-13
Sizing of EMF files incorrect and unchanging.


parchd-1:
Images in EMF format exported to docx do not have the correct size or aspect ratio, and it does not seem possible to set the size. This was previously reported in #2720 but the reporter closed the issue before it was resolved.

Example markdown document (also in attached zip with emf and output docx [problem.zip](https://github.com/jgm/pandoc/files/1146421/problem.zip)):

```markdown
No width given:

![testimg](./squarecircle.emf)

With percentage:

![2testimg](./squarecircle.emf){width=100%}

With units:

![3testimg](./squarecircle.emf){#bah width=6cm}

With units:

![4testimg](./squarecircle.emf){width=6in height=3in}

```

The size / proportions of the image seem to be extractable from the file:
From the attached file, opening it with less I see "Screen=10205x13181px, 216x279mm" and "Drawing=576.0x288.0px, 152.4x76.2mm"

jgm:
We'd need to add support for EMF to Text.Pandoc.ImageSize.
It seems that the size information is fairly easy to extract, so that's good.

parchd-1:
I have also found https://stackoverflow.com/questions/3391087/extracting-width-height-dimensions-from-emf-image-files#3391180 which links to https://msdn.microsoft.com/en-us/library/cc230514.aspx. They might also be helpful.

parchd-1:
I had a little look at Text.Pandoc.ImageSize and the docx writer, so I think the answer to this is probably no, but would this be possible to do using a filter
 I'm a bit desperate for it, although I will use high-dpi png files in an emergency this would obviously be much better. Filters I can cope with, especially thanks to panflute, but haskell not so much.

mb21:
> I had a little look at Text.Pandoc.ImageSize and the docx writer, so I think the answer to this is probably no

the answer to what?

you can certainly do it in a filter as well, but it would be preferably to have this in pandoc itself in the long run...

> but haskell not so much

see e.g. http://learnyouahaskell.com :)

parchd-1:
> the answer to what?
the answer to the whether it was possible in a filter. Sadly I didn't accomplish it, so in the meantime I'll have to go with massive file sizes from the pngs.

parchd-1:
I did work out how to do this in a filter in the end -  I was having a dense moment.

[...]

PR #4375, opened 2018-02-18
EMF Image size support

andymac-2:
ref #3798

Correct aspect ratio for *.emf files when exporting to docx. There are two issues however unrelated to this PR:

1. If using multiple of the same image in a docx, the docx will resize every image (not just .emf) to the size of the first image. The size descriptions of the other copies are ignored. (#3930)
2. .emf files exported by Inkscape are non-compliant to the specification of emf, so don't use these images for testing. 

The only emf files I could find that weren't generated by Inkscape were included by default on windows.


jgm:
Terrific, thanks!

PR #4383, opened 2018-02-21
Fix Docx writer: Same image can now be individually resized

andymac-2:
ref #3930

Fixes a bug wherein including the same image multiple times in a `.docx` will cause all of the images to be the same size, regardless of the size that you specify in the input.

This PR currently breaks 2 tests, however, and I am fairly confident that the tests themselves are broken. The tests which break fail because the description of the image and title text are now different with each copy of the same image. Previously, they would be a copy of the title text and description of the first instance of an image, like the size of the image. I would like some confirmation of this before I change any "golden" files however.


jkr:
You can rebuild the golden tests by passing `--accept` to the test script (so if you're using stack, `stack test --test-arguments "-p Docx --accept"`. Then just make sure to commit the changed golden files in their own commit with a line that you checked them in Word (mentioning Word version and OS), that they weren't corrupted, and that they had the expected output.

(Maybe you were going to do this anyway, but could you also squash down the other commits into logical steps with more descriptive commit messages?)

PR #4392, opened 2018-02-22
Add option to specify table env. in LaTeX output

pajowu:
This PR adds a command line switch (`--latex-table-environment`) for switching between possible table environments. Currently pandoc is hardcoded to use `longtable`, which allows for multi-page-tables. This adds support for the tabularx environment (allowing columns that span over the while remaining table-width) and gives the possibility to easily add new environments in the future.

**NOTE**: The package used in the template could also be `ltablex`, which is a combination of longtable and tabularx (allows multi-page tables with auto-spanning columns).

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -921,7 +921,7 @@ Options affecting specific writers {.options}
 
 :   Use the [`listings`] package for LaTeX code blocks. The package
     does not support multi-byte encoding for source code. To handle UTF-8
-    you would need to use a custom template. This issue is fully 
+    you would need to use a custom template. This issue is fully
     documented here: [Encoding issue with the listings package].
 
 `-i`, `--incremental`
@@ -1148,10 +1148,18 @@ Options affecting specific writers {.options}
     If used multiple times, the arguments are provided with spaces between
     them. Note that no check for duplicate options is done.
 
+`--latex-table-environment=longtable`|`tabularx`
+
+:   Use the specified table environment when producing LaTeX output.
+    The default is [`longtable`] which allows table to spread over multiple pages.
+    [`tabularx`] allows for paragraph-like columns whose width automatically expand to fill the width of the environment.
+
 [Dublin Core elements]: http://dublincore.org/documents/dces/
 [ISO 8601 format]: http://www.w3.org/TR/NOTE-datetime
 [Encoding issue with the listings package]:
   https://en.wikibooks.org/wiki/LaTeX/Source_Code_Listings#Encoding_issue
+[`longtable`]: https://ctan.org/pkg/longtable
+[`tabularx`]: https://ctan.org/pkg/tabularx
 
 Citation rendering {.options}
 ------------------
@@ -1598,6 +1606,10 @@ LaTeX variables are used when [creating a PDF].
 :   An option for LaTeX's `\pagestyle{}`. The default article class
     supports 'plain' (default), 'empty', and 'headings'; headings puts
     section titles in the header.
+
+`table_environment`
+:   The environemnt given by `--latex-table-environment`
+
 
 [`article`]: https://ctan.org/pkg/article
 [`report`]: https://ctan.org/pkg/report
@@ -4484,7 +4496,7 @@ the [Beamer User's Guide] may also be used: `allowdisplaybreaks`,
 Background in reveal.js and beamer
 ----------------------------------
 
-Background images can be added to self-contained reveal.js slideshows and 
+Background images can be added to self-contained reveal.js slideshows and
 to beamer slideshows.
 
 For the same image on every slide, use the  configuration
@@ -4492,9 +4504,9 @@ option `background-image` either in the YAML metadata block
 or as a command-line variable. (There are no other options in
 beamer and the rest of this section concerns reveal.js slideshows.)
 
-For reveal.js, you can instead use the reveal.js-native option 
-`parallaxBackgroundImage`. You can also set `parallaxBackgroundHorizontal` 
-and `parallaxBackgroundVertical` the same way and must also set 
+For reveal.js, you can instead use the reveal.js-native option
+`parallaxBackgroundImage`. You can also set `parallaxBackgroundHorizontal`
+and `parallaxBackgroundVertical` the same way and must also set
 `parallaxBackgroundSize` to have your values take effect.
 
 To set an image for a particular reveal.js slide, add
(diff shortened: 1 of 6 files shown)


pajowu:
Is there something blocking this PR?

MANUAL.txt
```diff
@@ -1054,8 +1054,16 @@ Options affecting specific writers
     If used multiple times, the arguments are provided with spaces between
     them. Note that no check for duplicate options is done.
 
+`--latex-table-environment=longtable`|`tabularx`
+
+:   Use the specified table environment when producing LaTeX output.
+    The default is [`longtable`] which allows table to spread over multiple tables.
```
jgm: Do you mean "over multiple pages"?

MANUAL.txt
```diff
@@ -1054,8 +1054,16 @@ Options affecting specific writers
...
+    [`tabularx`] allows for columns which autospread over the entire linewidth.
```
jgm: I wasn't sure exactly what was meant by "autospread over the entire linewidth."

jgm:
Sorry, I just haven't had time to review it.  Can you say more about the motivation for the change?

themightyoarfish:
One motivation from my perspective is that `longtable` is broken in the presence of floats on the same page (see e.g.[here](https://tex.stackexchange.com/questions/326882/figure-image-is-overflowing-bottom-of-page) or [here](https://tex.stackexchange.com/questions/276699/the-longtable-environment-pushes-content-below-it-into-the-bottom-margin-of-a-pa#277542))

In my case, this happily pushes sections into the bottom margin and the space between pages, so they are lost. 

This is a known bug in `longtable`, but unfortunately it also breaks pandoc use in cases, so an option for this would be great.

themightyoarfish:
Could this be moved along? I don't have a workaround for this right now.

pajowu:
I updated the documentation according to your Remarks and fixed all merge conflicts. The PR should be mergable again

Regarding my motivation: `longtable` inserts empty pages or overflows the page when float objects are on the same page. This makes `longtable` unusable for us

MANUAL.txt
```diff
@@ -1148,10 +1148,18 @@ Options affecting specific writers {.options}
     If used multiple times, the arguments are provided with spaces between
     them. Note that no check for duplicate options is done.
 
+`--latex-table-environment=longtable`|`tabularx`
+
+:   Use the specified table environment when producing LaTeX output.
+    The default is [`longtable`] which allows table to spread over multiple pages.
+    [`tabularx`] allows for paragraph-like columns whose width automatically expand to fill the width of the environment.
```
jgm: Lines should be hard-wrapped to fit within 80 columns.

jgm: I'm not sure I quite understand this description: "paragraph-like columns whose width automatically expand to fill the width of the environment."  Maybe better to say something about what should dictate this choice?  For example: If you have long tables that may need to spans multiple pages, use `longtable`.  If you want to create floating tables (and for better compatibility with other floating elements on the page), use `tabularx`.

MANUAL.txt
```diff
@@ -1599,6 +1607,10 @@ LaTeX variables are used when [creating a PDF].
     supports 'plain' (default), 'empty', and 'headings'; headings puts
     section titles in the header.
 
+`table_environment`
+:   The environemnt given by `--latex-table-environment`
```
jgm: "environment"

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -850,6 +850,17 @@ options =
                      UTF8.hPutStr stdout (usageMessage prg options)
                      exitSuccess ))
                  "" -- "Show help"
+    , Option "" ["latex-table-class"]
```
jgm: Shouldn't this be `latex-table-environment`, given the documentation?

src/Text/Pandoc/Options.hs
```diff
@@ -168,6 +169,11 @@ data ReferenceLocation = EndOfBlock    -- ^ End of block
                        | EndOfDocument -- ^ at end of document
                        deriving (Show, Read, Eq, Data, Typeable, Generic)
 
+-- | Class of tables to use in LaTeX output
+data LatexTableEnvironment = Longtable    -- ^ Use longtable for tables
+                     | Tabularx     -- ^ Use tabularx tables
```
jgm: This should align as in the other examples in the code.

src/Text/Pandoc/Options.hs
```diff
@@ -204,7 +210,8 @@ data WriterOptions = WriterOptions
   , writerReferenceLocation :: ReferenceLocation    -- ^ Location of footnotes and references for writing markdown
   , writerSyntaxMap         :: SyntaxMap
   , writerPreferAscii       :: Bool           -- ^ Prefer ASCII representations of characters when possible
-  } deriving (Show, Data, Typeable, Generic)
+  , writerLatexTableEnvironment   :: LatexTableEnvironment
```
jgm: Note clearly the API change in the commit comment.

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -822,13 +835,22 @@ getCaption txt = do
...
+  text (latexTableEnvironment tableEnvironment) <> "}"
+
+toColDescriptor :: (Integer, Alignment) -> String
+toColDescriptor (index, align) =
   case align of
          AlignLeft    -> "l"
          AlignRight   -> "r"
          AlignCenter  -> "c"
-         AlignDefault -> "l"
+         AlignDefault -> if index == 0 then "l" else "X"
```
jgm: Why is "l" used for the first column and then "X"?

jgm:
Thanks for the update. I've made some comments to your patch.

Currently there are no tests. There should be tests.

Also, your patch assumes that the only differences between the environments lie in the column specs and environment name.  Have you tested things like captions, footnotes in table cells, and so on, to make sure that this is the case?

jgm:
tabu scared me off with this note in the documentation:


> The tabu package patches internal commands of many packages and is liable to stop working if those internal interfaces change. Unfortunately the original author appears to be out of contact and the package had not been updated for some years.

Grandement:
I would prefer longtabu as well instead of longtable. The definition of longtabu is close to longtable and I guess it is not a big deal to add it with a optional argument to pandoc. The tabu packge is still working well :). With longtabu automatic linebreaks in a cell are possible due to the column X definition which would be a benifit for pandoc.

PR #4396, opened 2018-02-23
Docx writer: Fix issue regarding same image included multiple times

andymac-2:
ref #3930 

The Docx writer was storing a record of the XML generated for each image, and if an image was included several times, the same XML was used. This fix guarantees that new XML is generated each time and only keeps the image ID and image path fields the same. This allows custom sizing and descriptions of the same image included more than once.

Test cases are included.


jkr:
Pushed. Thanks!

PR #4412, opened 2018-02-27
add option to space list items with blank lines in RST, closes #4410

### LINKED ISSUE #4410: loose RST lists are parsed as tight lists
blank lines between items in an RST list [are optional](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#bullet-lists), and currently not added by pandoc. in cases when they are desired, rather than adding them manually, we would like to have a writer option that changes the default behaviour

danse:


danse:
while i was there i found out that `bulletListItemToRST` could be expressed in terms of `orderedListItemToRST`

danse:
oops this option cannot be used until i'll add it to `App.hs`, i will amend

PR #4432, opened 2018-03-06
Remove rawInlineOr because `\ref` and `\label` should be parsed in any case even with raw_tex extension

schrieveslaach:


jgm:
I'd like to hear more about the motivation for this.
It seems to me that someone might want to pass through the raw tex if they're targeting tex.
The pandoc replacement won't be as good, will it?

schrieveslaach:
I think that `raw_tex` returns the raw latex source of unknown commands. As pandoc is getting better to parse the `\label` and `\ref` commands, such as parsing labels of sections, I think `\label` and `\ref` should not be treated as raw latex commands. If you provide the extension `raw_tex`on the command line, pandoc parses lists not as raw block and IMHO pandoc should parse `\label` and `\ref` in the same way.

jgm:
But look at the current output:

    % pandoc -t native -flatex-raw_tex
    \label{mylabel}

    See \ref{mylabel}.
    ^D
    [Para [Span ("mylabel",[],[("label","mylabel")]) [Str "[mylabel]"]]
    ,Para [Str "See",Space,Link
    ("",[],[("reference-type","ref"),("reference","mylabel")]) [Str
    "[mylabel]"] ("#mylabel",""),Str "."]]

If you're targeting pdf or a tex format, this will give you
much worse results than passing through the raw tex.
Of course, you could process this with a filter to get what
you want, but that's an extra step and not at all built in.

Consider also that raw latex `\ref` commands can refer to
equation numbers in raw math environments, where pandoc
wouldn't pick up the label at all...

jgm:
Marc Schreiber <notifications@github.com> writes:

> The problem I encounter relates to `latex+raw_tex`. For example (notice the `+`):
>
> ```bash
> echo '\label{mylabel}    See \ref{mylabel}.' | pandoc -f latex+raw_tex -t native
> ```
>
> results in:
>
> ```
> [Para [RawInline (Format "latex") "\\label{mylabel}",Space,Str "See",Space,RawInline (Format "latex") "\\ref{mylabel}",Str "."]]
> ```
>
> Now, `\ref` and `\label` are `RawInline`s... 


Right, but why is this bad?  You should only use `+raw_tex` if your
output format is going to be able to use raw LaTeX, and in this case
having the RawInlines is fine, no?

schrieveslaach:
I'm using `+raw_tex` to parse commands unknown to pandoc as `RawBlock`/`RawInline`. These blocks/inlines will be process by a custom filter. However, if I do so, I need to handle, such as `[RawInline (Format "latex") "\\label{mylabel}"]` by myself. 

I think that the current behavior is somehow inconsistent, for example, if I run `echo '\begin{itemize} \item \textbf{1} \end{itemize}' | pandoc -f latex+raw_tex -t native` pandoc will parse all commands as native blocks while `\ref` and `\label` will be parsed as raw.

jgm:
> I think that the current behavior is somehow inconsistent, for example, if I run echo '\begin{itemize} \item \textbf{1} \end{itemize}' | pandoc -f latex+raw_tex -t native pandoc will parse all commands as native blocks while \ref and \label will be parsed as raw.

The principle is that we fall back to raw (when +raw_tex is available) in cases where parsing as native Pandoc structure would give essentially the same output when rendered back out to tex, so nothing is lost.  When our native Pandoc constructs don't reproduce what the raw tex can do, preserving the raw tex gives better results when the output format can handle raw tex.  That's the idea, anyway.

schrieveslaach:
OK, I understand. 
But, shouldn't it be possible to extend the LaTeX writer to work that way, such as that following command produce the same output?

```bash
echo "\\label{x} \ref{x}" | pandoc -f latex+raw_tex -t latex
# Output: \label{x} \ref{x}
echo "\\label{x} \ref{x}" | pandoc -f latex -t latex
# Desired output: \label{x} \ref{x}
```

The latter command produces `\hypertarget` and `\hyperlink` LaTeX command.

jgm:
That's a possibility; if we modified the LaTeX writer to
give the same output, then we could always parse these
to native pandoc structures.  I'm not sure how difficult
this would be, or how reliable.

Issue #4313, opened 2018-01-26
Converting from Markdown to DokuWiki preserves backslashes/escapes in table.


RobertZenz:
When I'm converting from Markdown to DokuWiki, backslashes which are used to create paragraphs inside table cell are preserved.

Example:

    ----------------------------------------------------------------------------
    Column             Column
    -----------------  ---------------------------------------------------------
    Value              This is some text.\
                       \
                       And this is a new paragraph.
    ----------------------------------------------------------------------------

HTML (is as expected):

    <table>
        <colgroup>
            <col style="width: 25%" />
            <col style="width: 75%" />
        </colgroup>
        <thead>
            <tr class="header">
                <th style="text-align: left;">Column</th>
                <th style="text-align: left;">Column</th>
            </tr>
        </thead>
        <tbody>
            <tr class="odd">
                <td style="text-align: left;">Value</td>
                <td style="text-align: left;">This is some text.<br />
                        <br />
                        And this is a new paragraph.</td>
                </tr>
        </tbody>
    </table>

DokuWiki:

    ^Column ^Column ^
    |Value |This is some text.\\\\ \\\\ And this is a new paragraph. |

As far as I can see, these should not be preserved.

jgm:
+++ RobertZenz [Jan 26 18 12:26 ]:
>   DokuWiki:
>^Column ^Column ^
>|Value |This is some text.\\\\ \\\\ And this is a new paragraph. |
>
>   As far as I can see, these should not be preserved.

What do you mean exactly?  Is this invalid DokuWiki?  How
does this display in DokuWiki?

Note:  if you want multiple paragraphs in a Markdown table,
you should use grid tables. That's the only format that
allows multiple blocks in cells.

RobertZenz:
It is displayed with the backslashes when rendered by DokuWiki:

    This is some text.\\
    \\
    And this is a new paragraph.

But it should display without them, I'd say. So it shouldn't be converted to 4 backslashes, but only two.

I just tried with grid tables, and that works. It's a lot more type-work, though.

jgm:
Can you confirm that
```
^Column ^Column ^
|Value |This is some text.\\ \\ And this is a new paragraph. |
```
will render as expected in DokuWiki (with line breaks in the cell)?

RobertZenz:
Sorry for the long delay, yes, I can confirm that the two-backslash variant is rendering correctly.

[...]

PR #4445, opened 2018-03-12
DokuWiki writer: rewrite backSlashLineBreaks

mb21:
After fixing #4313, I felt confident enough to rewrite `backSlashLineBreaks` to be more readable and more efficient. Behaviour should be the same as `backSlashLineBreaks . vcat` previously.

Maybe @jkr can have a quick look, just to be sure?

### DIFF
--- a/src/Text/Pandoc/Writers/DokuWiki.hs
+++ b/src/Text/Pandoc/Writers/DokuWiki.hs
@@ -366,12 +366,16 @@ isSimpleBlockQuote bs  = all isPlainOrPara bs
 vcat :: [String] -> String
 vcat = intercalate "\n"
 
-backSlashLineBreaks :: String -> String
-backSlashLineBreaks cs = reverse $ g $ reverse $ concatMap f cs
-  where f '\n' = "\\\\ "
-        f c    = [c]
-        g (' ' : '\\':'\\': xs) = xs
-        g s                     = s
+-- | For each string in the input list, convert all newlines to
+-- dokuwiki escaped newlines. Then concat the list using double linebreaks.
+backSlashLineBreaks :: [String] -> String
+backSlashLineBreaks ls = vcatBackSlash $ map escape ls
+  where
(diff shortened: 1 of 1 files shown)


jgm:
My suggestion would be to fire up quickCheck and test that `backSlashLineBreaks xs == oldBackSlashLineBreaks (vcat xs)`.

mb21:
Good idea! Here it is:

    import Test.QuickCheck
    import Data.List (intercalate)

    vcat :: [String] -> String
    vcat = intercalate "\n"

    oldBackSlashLineBreaks :: String -> String
    oldBackSlashLineBreaks cs = reverse $ g $ reverse $ concatMap f cs
      where f '\n' = "\\\\ "
            f c    = [c]
            g (' ' : '\\':'\\': xs) = xs
            g s                     = s

    old = oldBackSlashLineBreaks . vcat

    new :: [String] -> String
    new ls = vcatBackSlash $ map escape ls
      where
        vcatBackSlash = intercalate "\\\\ \\\\ " -- simulate paragraphs.
        escape ('\n':[]) = "" -- remove trailing newlines
        escape ('\n':cs) = "\\\\ " ++ escape cs
        escape (c:cs)    = c : (escape cs)
        escape []        = []

    prop_backslashapp :: [String] -> Bool
    prop_backslashapp ls = old ls == new ls

    test = quickCheck prop_backslashapp

Unfortunately, there are quite a few differences, e.g.:

    λ> old ["", ""]
    ""
    λ> new ["", ""]
    "\\\\ \\\\ "
    λ> old ["a", "a"]
    "a\\\\ a"
    λ> new ["a", "a"]
    "a\\\\ \\\\ a"

I'm not sure which behavior is correct. Also: those are lists that I'm quite sure `mapM (blockToDokuWiki opts) blocks'` would never generate. Namely, I'm quite certain that all useful lists would have a newline at the end of each element string...

PR #4478, opened 2018-03-20
better interline space for footnote references

ousia:


mb21:
Can you describe why this is "better"? It doesn't seem to be what we want: https://jsfiddle.net/Ljbm14ec/

Also: why use percent instead of em?

ousia:
@mb21, please check the ePub I have just sent to the mailing list.

About JSFiddle, I have never used it. But I’m afraid that they mix both `text-top` and `text-bottom` (<https://www.w3.org/TR/CSS2/visudet.html#propdef-vertical-align>).

Look, `text-top` is located at the bottom of the line:

text-top-bottom

And `text-bottom` is placed at the top of the line:

text-bottom-top

I think percentages are a better way to define relative sizes in fonts, especially in this case. I think is a clearer way to express the proportion in decrease to the main font.

ousia:
Sorry, I forgot. I think it is better because it doesn’t increase the line height.

Here you have the screenshot:

both-options

The superscript in the first paragraph is set with `text-top`. The superscript in the second paragraph is set with `super`.

Although they have exactly the same relative font size, `vertical-align: super` increases the height for that particular line.

I hope it is clear now why I consider it is a better approach.

mb21:
I see the issue now, doesn't look like a very good default... then again, that's exactly the styling a browser would apply to a `<sup>` element as well (which we had [before](https://github.com/jgm/pandoc/commit/4560447041a44d3eea9e442f9b5aa20911b92ed8)).

The jsfiddle was meant to illustrate that `text-top` breaks on large line heights. Better [solutions](https://stackoverflow.com/questions/1530685/html-sup-tag-affecting-line-height-how-to-make-it-consistent):

    a.footnote-ref { vertical-align: top; font-size: 0.6em; }
    a.footnote-ref { vertical-align: top; position: relative; top: -0.5em; }

and yet [another](https://css-tricks.com/snippets/css/prevent-superscripts-and-subscripts-from-affecting-line-height/):

```
a.footnote-ref { vertical-align: baseline; position: relative; top: -0.4em; }
```

Maybe it depends on the font and the ebook reader which one is best...?

ousia:
> I see the issue now, doesn't look like a very good default... then again, that's exactly the styling a browser would apply to a `<sup>` element as well (which we had before).

I don’t think so. I used the CSS code contributed in the patch to style the `<sup>` element. (I don’t use it anymore, since I replace the numbers with real superscript glyphs).

> Maybe it depends on the font and the ebook reader which one is best...?

I don’t know. I’m only able to check it with the reader I have. (I wonder whether the three options you mention would work in an old readers, such as the ones from _Sony_.) 

I’m closing the pull request myself. I can contribute a new patch, but I need to know first which approach is better for you.

mb21:
I'm not quite sure which approach is best either... maybe you can open an issue until we can get some people to test it on a few ereaders?

ousia:
Just opened #4487.

PR #4479, opened 2018-03-21
MANUAL.txt fix raw content example

mb21:
Otherwise I got:

    [WARNING] Could not parse YAML metadata at line 4 column 3: found character that cannot start any token


jgm:
Thanks.

Issue #1995, opened 2015-03-06
ePub pop-up footnotes break in iBooks


anthonynelzin:
When reading an ePub book generated by Pandoc with iBooks for iOS, popup footnotes won't open. Here's why. Pandoc writes the footnotes as:

	<a href="#fn1" class="footnoteRef" id="fnref1" epub:type="noteref"><sup>1</sup></a>

When the `<sup>` tag is inside the `<a>` tag, popup footnotes break (for that matter, if any tag is inside the `<a>` tag, the footnote won't work). This is a fairly well documented problem -- here on Github, on the Adobe InDesign forums, or [on Liz Castro's website](http://www.pigsgourdsandwikis.com/2012/05/creating-pop-up-footnotes-in-epub-3-and.html) for example. In order for the footnote to properly work in iBooks, the markup should be :

	<sup><a href="#fn1" class="footnoteRef" id="fnref1" epub:type="noteref">1</a></sup>

Another tweak would go a long way in improving the reliability and legibility of Pandoc-generated ePubs: allowing the user to specify if the footnotes block should be visible (`<section class="footnotes">`, good for books designed to be distributed on many plateforms) or not (`<aside class="footnotes">`, good for books designed to be distributed on the iBooks Store only, or for the iBooks version of the book, as iBooks hides asides by defaults).

Hope it helps!

mpickering:
So you suggest it would be best to extract any tags which are inside links to outside the link? This is not just specific to `sup`?

anthonynelzin:
I've seen it happen with spans, for example. Adobe also told me that “Formatting in footnote text is not supported in pop up footnotes. This is a limitation of iBooks”: putting a `<b>` or an `<i>` or anything else in the `<a>` tag break the footnote.

nkalvi:
Did some quick tests with iBooks on Mac and iPad. Interestingly enough popup footnotes work as expected on iBooks on Mac. Formatting seems to 'break' them in iOS version; I'm wondering whether it will be fixed in future versions (since it works on Mac).

BTW, when output with epub3 option, 'footnotes' section is created (it can be hidden in CSS, as show below). I'm wondering whether style can be used instead of hard-coding *superscript* for footnote references, at least for epubs.

Following seems to work fine in iBooks on Mac and iOS.

Manually removed *\<sup\>* tag:
```
<p>Footnotes</p>
<p>Extension: footnotes</p>
<p>Pandoc’s markdown allows footnotes, using the following syntax:</p>
[...]

PR #4491, opened 2018-03-24
Several fixes for EPUB footnotes

### LINKED ISSUE #4489: Footnotes section should have epub:type="footnotes" in EPUB
As noted in [the accessibility guidelines](https://idpf.github.io/a11y-guidelines/content/xhtml/notes.html#xhtm080):
> Notes that are grouped at the end of a section do not need to be individually tagged as `aside`, but should be grouped in a section carrying the appropriate `footnotes` or `endnotes` semantic

Location to update:

https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/HTML.hs#L481-L484

Porges:
With the current EPUB output (after converting with Kindle previewer and copying to Kindle...), the pop-over footnotes will often have problems:

1. Several footnotes run together and appear as one when clicking on a footnote.
2. Some footnotes do not work at all with pop-over and the link will jump you directly to the footnote section.

Here's an example showing the footnotes running together:
Multiple footnotes run together

I believe that both of these problems are caused by having the footnote back-link at the end of the footnote. Changing the link to be at the start of the footnote fixes this. (I don't know why this causes problems with the Kindle software.)

Here's the same footnote after making this change:
 Single footnote at a time

Putting the '↩' back-link at the start of the footnote doesn't make much sense, so I changed it to a numeric back-link instead. This was based on the examples given in both the [Amazon Kindle Publishing Guidelines](https://kindlegen.s3.amazonaws.com/AmazonKindlePublishingGuidelines.pdf) and the [EPUB3 Accessibility Guidelines](https://idpf.github.io/a11y-guidelines/content/xhtml/notes.html#xhtm080-ex04).

However, making the back-link numeric meant that the ordered list that Pandoc outputs for footnotes would duplicate the numbering provided in the back-link. To avoid this (and implement something suggested on #1995), I've also changed the footnote output to use `<aside>` (on HTML5/EPUB3) or fall back to `<div>` (on HTML4/EPUB2). Using asides is something that is "strongly recommended" by Amazon's guidelines, and is what Apple recommends in the [iBooks Asset Guide](https://help.apple.com/itc/booksassetguide/en.lproj/static.html#itccf8ecf5c8). (There is an additional feature provided by iBooks which is that the asides are hidden from the main flow of the text.) So we seem to be able to solve two issues at once 🙂

I've also checked the new output in Calibre, Edge, Kindle Previewer, and iBooks to confirm that it works as expected.

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -490,8 +490,7 @@ footnoteSection opts notes = do
   return $
     if null notes
        then mempty
-       else nl opts >> container (nl opts >> hrtag >> nl opts >>
-              H.ol (mconcat notes >> nl opts) >> nl opts)
+       else nl opts >> container (nl opts >> hrtag >> mconcat notes >> nl opts)
 
 -- | Parse a mailto link; return Just (name, domain) or Nothing.
 parseMailto :: String -> Maybe (String, String)
@@ -1166,22 +1165,21 @@ inlineToHtml opts inline = do
 
 blockListToNote :: PandocMonad m => WriterOptions -> String -> [Block] -> StateT WriterState m Html
 blockListToNote opts ref blocks =
-  -- If last block is Para or Plain, include the backlink at the end of
+  -- If first block is Para or Plain, include the backlink at the start of
   -- that block. Otherwise, insert a new Plain block with the backlink.
-  let backlink = [Link ("",["footnote-back"],[]) [Str "↩"] ("#" ++ "fnref" ++ ref,[])]
-      blocks'  = if null blocks
-                    then []
-                    else let lastBlock   = last blocks
-                             otherBlocks = init blocks
-                         in  case lastBlock of
-                                  (Para lst)  -> otherBlocks ++
-                                                 [Para (lst ++ backlink)]
-                                  (Plain lst) -> otherBlocks ++
-                                                 [Plain (lst ++ backlink)]
-                                  _           -> otherBlocks ++ [lastBlock,
-                                                 Plain backlink]
+  let backlink = [Link ("",["footnote-back"],[]) [Str (ref ++ ".")] ("#" ++ "fnref" ++ ref,[]), Space]
+      blocks'  = case blocks of
+                  [] -> []
+                  (firstBlock:otherBlocks) ->
+                    case firstBlock of
+                      (Para lst) -> Para (backlink ++ lst) : otherBlocks
+                      (Plain lst) -> Plain (backlink ++ lst) : otherBlocks
+                      _ -> Plain backlink : blocks
+
   in  do contents <- blockListToHtml opts blocks'
-         let noteItem = H.li ! prefixedId opts ("fn" ++ ref) $ contents
+         html5 <- gets stHtml5
+         let noteElement = if html5 then H5.aside else H.div
+         let noteItem = noteElement ! prefixedId opts ("fn" ++ ref) $ contents
          epubVersion <- gets stEPUBVersion
          let noteItem' = case epubVersion of
                               Just EPUB3 -> noteItem ! customAttribute "epub:type" "footnote"
--- a/test/command/4235.md
+++ b/test/command/4235.md
@@ -5,8 +5,6 @@ This.^[Has a footnote.]
 <p>This.<a href="#foofn1" class="footnote-ref" id="foofnref1"><sup>1</sup></a></p>
 <section class="footnotes">
 <hr />
-<ol>
-<li id="foofn1"><p>Has a footnote.<a href="#foofnref1" class="footnote-back">↩</a></p></li>
-</ol>
+<aside id="foofn1"><p><a href="#foofnref1" class="footnote-back">1.</a> Has a footnote.</p></aside>
 </section>
 ```
--- a/test/writer.html4
+++ b/test/writer.html4
@@ -534,16 +534,14 @@ Blah
 <p>This paragraph should not be part of the note, as it is not indented.</p>
 <div class="footnotes">
 <hr />
-<ol>
-<li id="fn1"><p>Here is the footnote. It can go anywhere after the footnote reference. It need not be placed at the end of the document.<a href="#fnref1" class="footnote-back">↩</a></p></li>
-<li id="fn2"><p>Here’s the long note. This one contains multiple blocks.</p>
+<div id="fn1"><p><a href="#fnref1" class="footnote-back">1.</a> Here is the footnote. It can go anywhere after the footnote reference. It need not be placed at the end of the document.</p></div>
+<div id="fn2"><p><a href="#fnref2" class="footnote-back">2.</a> Here’s the long note. This one contains multiple blocks.</p>
 <p>Subsequent blocks are indented to show that they belong to the footnote (as with list items).</p>
 <pre><code>  { &lt;code&gt; }</code></pre>
-<p>If you want, you can indent every line, but you can also be lazy and just indent the first line of each block.<a href="#fnref2" class="footnote-back">↩</a></p></li>
-<li id="fn3"><p>This is <em>easier</em> to type. Inline notes may contain <a href="http://google.com">links</a> and <code>]</code> verbatim characters, as well as [bracketed text].<a href="#fnref3" class="footnote-back">↩</a></p></li>
-<li id="fn4"><p>In quote.<a href="#fnref4" class="footnote-back">↩</a></p></li>
-<li id="fn5"><p>In list.<a href="#fnref5" class="footnote-back">↩</a></p></li>
-</ol>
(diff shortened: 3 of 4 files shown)


Porges:
It looks like there are only reader tests at the moment?

mb21:
It's great to see someone taking a look at the ePub output and how it's rendered on a kindle.

The existing behvarour of the HTML (and consequently EPUB) writer puts the footnotes in an `ol` in a `section` at the end of the document, which makes perfect sense for a document to be printed, or rendered without javascript or fancy on-hover/on-click CSS. As such, and also for backwards-compatibility, I guess we should at least preserve an option to keep the old behaviour. I also totally agree with adding `epub:type="footnotes"`. Then there are two changes which you mention that might be independent of each other?

- Moving the backling to the front: does this alone fix the kindle-problem of mangling several footnotes together? If so, do you have a guess to why? Because naively I would expect the human reader to want to encounter the backlik _after_ having read the footnotes, not _before_ (i.e. how does this affect the rendering in other epub readers?)
- Using an `aside` for foornotes seems to make sense if the aside is placed in the body text as well, not at the end. (Maybe you can link to Kindle/Apple recommendations?) Placing the footnotes in an `aside` in the middle of the text flow might make sense as an option, in addition to the existing way. The new way might be toggled with the `inline_notes` extension, which currenlty only the markdown reader uses, but the HTML writer could make use of it as well.

P.S. I had assumed EPUB writer tests would be in `/test/epub/` (e.g. `-f native formatting.native -o formatting.epub) but seeing that your pull request doesn't fail tests maybe they're not properly called in the test suite...?

Porges:
@mb21 I also [posted on pandoc-discuss](https://groups.google.com/forum/m/#!topic/pandoc-discuss/1hmkJFyE6sE) to explain the changes in more detail, some of your questions are answered there. I'll try to answer the others later today.

mb21:
Maybe let's try to keep the discussion to one place (I guess this pull request since we're already here)...

P.S. I guess it would make most sense if the position/display of the backlink could be overridden using CSS...

Porges:
Yes, I'll merge that info back into the PR (just haven't done it yet)

ousia:
@Porges, would it be possible that you could upload sample documents (for both ePub2 and ePub3) so that changes could be checked before they are merged?

Porges:
Yes, I also need to rebase as #4489 was fixed by c997f112 so is no longer needed in this PR.

Porges:
Updated original comment with more context from my pandoc-discuss message.

Porges:
> Moving the backlink to the front: does this alone fix the kindle-problem of mangling several footnotes together? If so, do you have a guess to why? Because naively I would expect the human reader to want to encounter the backlink after having read the footnotes, not before (i.e. how does this affect the rendering in other epub readers?)

Yes, this alone fixes the problem, and no, I don't know why. My best guess is that it's looking for the back-link as a marker for where the footnote starts and then reading onward from there, but that's not based on much evidence...

> Using an `aside` for footnotes seems to make sense if the aside is placed in the body text as well, not at the end. (Maybe you can link to Kindle/Apple recommendations?) Placing the footnotes in an `aside` in the middle of the text flow might make sense as an option, in addition to the existing way. The new way might be toggled with the `inline_notes` extension, which currently only the markdown reader uses, but the HTML writer could make use of it as well. (Ah yes, or indeed better the `--reference-location` option...?)

Yes, I had thought putting `aside`s inline, but there are some issues with doing it. For example the Kindle software doesn't hide `aside`s by default (but iBooks does) so we'd have to ship additional CSS to make this consistent, or you'd have footnotes appearing in the main flow of the text.

What I was trying to do was to fix the initial problem with minimal impact on other facilities. Adding support to `--reference-location` (or maybe `--footnote-location`) would be good but would require more work (I don't think `inline_notes` should affect output, it's only an input-side thing). I'd like to fix the display problem by itself first and then look at more flexibility later.

----

**Aside**: (😉) One thing I did realize while making this change is that the way that Pandoc currently outputs footnotes in EPUB should really be called "endnotes". [DPUB-ARIA is *very* explicit about this](https://www.w3.org/TR/dpub-aria-1.0/#doc-footnote) (note that `doc-footnote` is tied to `epub:type="footnote"`, so the semantics should be the same):

> The `doc-footnote` role is only for representing individual notes that occur within the body of a work. For collections of notes that occur at the end of a section, see `doc-endnotes`.

----

So the "wholistic" solution to footnotes might be to add `--footnote-location` (which perhaps defaults to `--reference-location` but may be set independently). The valid values are only `block` or `section`, `document` is not supported (or is treated as `section`).
  * `block` would place the footnotes into an `aside` after the current block, and Pandoc will have custom CSS to hide it (except they need to be shown when printing, etc).
  * `section` would place the footnotes at the end of the EPUB file (i.e. whatever `--epub-chapter-level` splits the document into), and these would be marked as _endnotes_ with `epub:type` and ARIA role, and not using `aside`.

Porges:
Okay, to further complicate matters, the EPUB 3 Accessibility Guidelines are now deprecated and the new site is the DAISY  Accessible Publishing Knowledge Base, which now has different examples for footnotes and endnotes!

The [guidelines](http://kb.daisy.org/publishing/docs/html/notes.html) are quite clear now:

> * Identify footnotes within the text using the `aside` element. [WCAG 1.3.1]
> * Group footnotes or endnotes at the end of a section or work in lists. [WCAG 1.3.1]
> * Link note references to their notes. [WCAG 1.3.1]
> * Backlink notes at the end of a section or work to their references. [WCAG 1.3.1]

For footnotes: 
> The backlink precedes the note text so that the user will be aware that such a link is available.

For endnotes:
> The backlink follows the note text to avoid any conflicts with automatic list numbering. The link is placed in its own element so that the user does not have to listen to the note before being able to activate it.

----

So the new "wholistic" plan is something like: add `--footnote-location` (which defaults to `--reference-location` but may be set independently). The valid values are only `block` or `section`, `document` is not supported (or is treated as `section`). [Alternately we call the options "footnote" and "endnote"?]

* `section` (which is very close to the current placement method):
  1. footnotes go at end of current EPUB file (i.e. whatever --epub-chapter-level splits the document into)
  2. placed in `ol`/`li`,
  3. back-link at end,
  4. with DPUB-ARIA/EPUB "endnote" semantics.

* `block` (which would be the new placement method):
  1. footnotes go after current block,
  2. placed in `aside`,
  3. back-link at the start,
  4. with DPUB-ARIA/EPUB "footnote" semantics.

----

However, this doesn't fix Kindle display for endnotes which have back-links at the end. I'll try experimenting some more and see if I can come up with some HTML which makes Kindles work again _without_ having the back-link at the start for the endnote.

mb21:
Sounds promising... just a few notes:

> I don't think `inline_notes` should affect output, it's only an input-side thing

Actually, we've started using extensions in different input and output, e.g. we could do `-t epub+inline_notes`. But maybe you're right that in this case it's better to use the existing `--reference-location` option.

> However, this doesn't fix Kindle display for endnotes which have back-links at the end.

Does this also depend on how you convert the ePUB to `mobi`? Using calibre or the official amazon tool?
 Also, couldn't the user just use `--reference-location=block` if he/she want to convert to `mobi` afterwards?

Porges:
> Does this also depend on how you convert the ePUB to `mobi`? Using calibre or the official amazon tool?

No, the problem exists with either method.

> Also, couldn't the user just use `--reference-location=block` if he/she want to convert to `mobi` afterwards?

This is true, but it would be nice to have it fixed in either case.

Porges:
When I get time/energy I'm going to try implementing the approach outlined in the [last comment](https://github.com/jgm/pandoc/pull/4491#issuecomment-378802283).

jgm:
@Porges thanks for your work on this.  I wonder if we could start with something simple. The guidelines say:

- Identify footnotes within the text using the aside element. [WCAG 1.3.1]
- Group footnotes or endnotes at the end of a section or work in lists. [WCAG 1.3.1]
- Link note references to their notes. [WCAG 1.3.1]
- Backlink notes at the end of a section or work to their references. [WCAG 1.3.1]
- For footnotes, the backlink precedes the note text so that the user will be aware that such a link is available.

According to this, it's fine to put the footnotes at the end of the section in lists, which is what we currently do.  So, all we really need to do is change the backlink to precede the text, and make sure all the links and cross-references and containing sections conform to guidelines, right?  We could think about doing fancier stuff later, and it might interact with a current PR adding an Endnote type.  But for now I'd prefer to do the simplest thing that would work.

Porges:
Closing this because I think the design & discussion has moved on @jgm?

jgm:
I was keeping this open because I thought we should at least do the following:

> So, all we really need to do is change the backlink to precede the text, and make sure all the links and cross-references and containing sections conform to guidelines, right?

If you'd prefer, though, we could make a new issue with this.

Porges:
@jgm, sure, feel free to take the code I had if needed. I'm just trying to clear up my old PRs and don't have time to contribute to this at the moment.

PR #4506, opened 2018-03-27
Removed unused extensions

Anabra:
Removed most unused extensions from the modules in the src directory. This has been done using the OrganizeExtensions refactoring of the Haskell-Tools framework.

### DIFF
--- a/src/Text/Pandoc/Readers.hs
+++ b/src/Text/Pandoc/Readers.hs
@@ -1,7 +1,4 @@
-{-# LANGUAGE NoImplicitPrelude #-}
-{-# LANGUAGE FlexibleInstances   #-}
-{-# LANGUAGE GADTs               #-}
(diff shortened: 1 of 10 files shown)


Anabra:
Looks like the pre8 GHC doesn't like the changes. Could add a conditional block with FlexibleContexts in it, or just FlexibleContexts itself.

jgm:
Thanks! I hadn't heard of this tool.  Can you give a link to it?

Comments:

- We should keep NoImplicitPrelude, even where it is not strictly needed, because we're using a custom prelude and importing it explicitly.

- I prefer having these pragmas one per line as before, rather than all on one line.

Anabra:
This is a little demo for the tool: http://www.haskelltools.org/
Since then, the OrganizeExtensions refactoring has improved quite a bit. Some other features have been added as well.

jgm:
Great, thanks for doing this!

PR #4531, opened 2018-04-05
CommonMark writer: don't emit HTML for small caps without raw_html

### LINKED ISSUE #4528: CommonMark writer: small caps are rendered with <span>, even without raw_html
```
$ cat test2.tex
This has \textsc{small caps} text.
$ pandoc -f latex -t commonmark-raw_html test2.tex 
This has <span class="smallcaps">small caps</span> text.
```

I grepped the CommonMark spec and it seems that `<span>` will be interpreted as raw HTML, which seems against the user's stated intent.

I think that the Markdown writer's behaviour, which is to capitalise the small-caps span, is much better in this case:

```
$ pandoc -f latex -t markdown_strict-raw_html test2.tex 
This has SMALL CAPS text.
```

quasicomputational:
Closes #4528.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3381,8 +3381,14 @@ Markdown allows it, but it has been made an extension so that it can
 be disabled if desired.)
 
 The raw HTML is passed through unchanged in HTML, S5, Slidy, Slideous,
-DZSlides, EPUB, Markdown, Emacs Org mode, and Textile output, and suppressed
-in other formats.
+DZSlides, EPUB, Markdown, CommonMark, Emacs Org mode, and Textile
+output, and suppressed in other formats.
+
+In the CommonMark format, if `raw_html` is enabled, superscripts,
+subscripts, strikeouts and small capitals will be represented as HTML.
+Otherwise, plain-text fallbacks will be used. Note that even if
+`raw_html` is disabled, tables will be rendered with HTML syntax if
+they cannot use pipe syntax.
 
 #### Extension: `markdown_in_html_blocks` ####
 
--- a/src/Text/Pandoc/Writers/CommonMark.hs
+++ b/src/Text/Pandoc/Writers/CommonMark.hs
@@ -45,7 +45,7 @@ import Network.HTTP (urlEncode)
 import Text.Pandoc.Class (PandocMonad)
 import Text.Pandoc.Definition
 import Text.Pandoc.Options
-import Text.Pandoc.Shared (isTightList, linesToPara, substitute)
+import Text.Pandoc.Shared (isTightList, linesToPara, substitute, capitalize)
 import Text.Pandoc.Templates (renderTemplate')
 import Text.Pandoc.Walk (query, walk, walkM)
 import Text.Pandoc.Writers.HTML (writeHtml5String, tagWithAttributes)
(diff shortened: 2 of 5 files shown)


jgm:
I'm not sure about this one.  Note that we currently fall back to HTML for superscript and subscript, for example.  The current behavior for SmallCaps is just like this.  But maybe we should provide fallbacks for all these cases?

I'm wondering whether someone might want to disable *passthrough* of raw HTML while still having the canned `<sup>` and so on.

quasicomputational:
RE `<sup>` and allies: it's a fair point, but I think that the small caps case can be separated out since there's a good enough non-HTML rendering. I'll also note that the PR's behaviour on small caps matches the Markdown writer's behaviour (except for being ignorant of `native_spans`).

I definitely see your point about wanting to control raw passthrough separately to using/not using HTML renderings. Admittedly that can be done with a filter, but keeping stray markup out is important enough that it should be convenient to do.

Here's a proposal for version 2.0: gate all the HTMLish renderings on `native_spans` || `raw_html`. Steal the remaining non-HTML fallbacks in the Markdown writer for the CommonMark writer, since that's already there.

quasicomputational:
I've implemented the plan outlined in the previous comment. It might be better to add another extension rather than abusing `native_spans`, possibly, but I'm also not convinced that this is a great evil.

I couldn't get the simple plain text sub/superscripts for numbers to work: following the pattern of the Markdown writer, I tried rendering the superscripts and then going character by character, but the CommonMark renderer adds a newline at the end which complicates everything; just blindly chomping whitespace seems dangerous.

quasicomputational:
I got the fancy Unicode sup/super-script numbers working. I think there are two remaining issues here:
* Is the `native_spans` extension being appropriately abused, or should a new one be added?
* `(markdown|commonmark)+native_spans-raw_html` will not round-trip, since the writer doesn't know about `<sup>`, `<sub>` and `<s>`.

quasicomputational:
@jgm, is this still a potentiality? Sorry if I'm being a bother here, but I'd like to either get this sorted or to know definitively that this isn't going to work. I'm plenty willing to hack on this more, but to me the PR looks in decent shape as it is and would be useful, so I'll need a pointer on which direction to take it if there are problems.

jgm:
Sorry to have taken so long to get back to this.
I think this looks good.  But I don't understand why `native_spans` should be used in the commonmark writer.  (That would make sense, I think, only if the commonmark reader parsed `<span>...</span>` as a pandoc `Span` element.)  Why not just gate on `raw_html`?

jgm:
I'm really undecided about this.  Note that we currently fall back to raw HTML for tables that can't fit in pipe tables, and there doesn't seem to be any alternative to that.  So, we can't say that `-t commonmark-raw_html` will give you output without raw HTML, even with these changes.  So these changes make the situation a bit more confusing:  setting `-raw_html` has the effect of (a) blocking things that originated as user-supplied raw HTML and (b) also blocking some, but not all of the HTML fallbacks produced by the writer itself.  Overloading `native_spans` seems even more confusing to me.  I do think there's a use case for people who want to suppress random user-supplied HTML but don't mind the curated HTML supplied by the writer, but it could be argued that this should be handled by putting `-raw_html` on the reader.

quasicomputational:
I'd consider `commonmark-raw_html` producing any HTML to be a bug. A quick look at the code doesn't reveal any other places where that's happening; do you have any in mind?

So I'll rip out the `native_spans` stuff again and have the only extension gating be on `raw_html` (and `strikeout` for `Strikeout`, and `native_spans` for `Span` elements only)?

jgm:
quasicomputational <notifications@github.com> writes:

> I'd consider `commonmark-raw_html` producing any HTML to be a bug. A quick look at the code doesn't reveal any other places where that's happening; do you have any in mind?

Look at blockToNodes for Table.  We fall back to
writeHtml5String.  It's hard to know what else to do
in this case, if we can't use a pipe table OR html.

quasicomputational:
I added a note to the docs about the HTML table caveat, since I figure that's the best we can do. Perfect non-HTML output is evidently not happening without some bright ideas here, but I think that this PR is a definite improvement on the status quo.

quasicomputational:
Thanks! This was evidently a lot more fiddly than I originally suspected, so the review was very useful.

Issue #4527, opened 2018-04-05
Converting from LaTeX to CommonMark leaves in \maketitle, even without raw_tex


quasicomputational:
The extra puzzling thing is that the Markdown writer doesn't do this:

```
$ cat test.tex 
\documentclass{article}

\title{Foo.}

\begin{document}

\maketitle

Hello.

\end{document}
$ pandoc -f latex-raw_tex -t commonmark-raw_tex test.tex
\maketitle

Hello.

$ pandoc -f latex -t markdown_strict-raw_tex test.tex
Hello.
$ pandoc --version
pandoc 2.1.3
Compiled with pandoc-types 1.17.4.2, texmath 0.10.1.1, skylighting 0.6
Default user data directory: /home/sam2/.pandoc
Copyright (C) 2006-2018 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
```

quasicomputational:
There are two bugs going on here, I think, and they're both necessary to observe this bug.

First, the LaTeX reader is emitting raw blocks even without `raw_tex`:

```
$  printf '\\foo{}\n\nhello \\foo{} bar' | pandoc -f latex-raw_tex -t native
[RawBlock (Format "latex") "\\foo{}"
,Para [Str "hello",Space,Str "bar"]]
```

As can be seen, the raw inlines are correctly omitted.

Second, the CommonMark and RST writers aren't ignoring the raw blocks. Here's the relevant code from `Text.Pandoc.Writers.CommonMark`:

```
blockToNodes opts (RawBlock fmt xs) ns
  | fmt == Format "html" && isEnabled Ext_raw_html opts
              = return (node (HTML_BLOCK (T.pack xs)) [] : ns)
  | fmt == Format "latex" || fmt == Format "tex" && isEnabled Ext_raw_tex opts
              = return (node (CUSTOM_BLOCK (T.pack xs) T.empty) [] : ns)
  | otherwise = return ns
```

That's got an operator precedence bug in its logic. The corresponding code for RawInline is correctly bracketted. I'll make up a PR for that.

quasicomputational:
Upon review, the RST raw logic is very different to the Markdown/CommonMark raw logic, defaulting to inclusion. That's probably a separate bug, and there's also the LaTeX reader bug, so merging the PR won't suffice to fully clear up the bugs here.

quasicomputational:
[...]

PR #4533, opened 2018-04-05
CommonMark writer: correctly ignore LaTeX raw blocks when not raw_tex

quasicomputational:
Issue #4527.


quasicomputational:
No clue why Travis is unhappy.

jgm:
Thanks! The 8.4.1 failure has to do with lua support and is unrelated to this patch.

PR #4536, opened 2018-04-06
Add publishers option to latex template

bvolkmer:
Add the publishers command as option to the latex template.
These are included on the title page in KOMA document classes.


karland:
I suppose this PR can also be closed due to the discussion in PR #5372.

Issue #2424, opened 2015-09-28
fb2 produced by pandoc doesn't validate


vbwagner:
PAndoc 1.12.4.2 produces incorrect FB2 metainfomramation.

Consider following example

```
  ---
  author: Victor Wagner
  title: The example fb2
  language: en
  abstract: The example FB2
  ---
  Chapter One
  ==========
   This is a text of  the chapter 0ne
```

I expect that command

```
  pandoc -o example.fb2 example.mkd 
```

Would produce file which would pass validation using

```
  xmllint --schema FictionBook2.1xml example.fb2
```

Unfortunately, I recieve lots of errors.

example.fb2:2: element book-title: Schemas validity error : Element '{http://www.gribuser.ru/xml/fictionbook/2.0}book-title': This element is not expected. Expected is ( {http://www.gribuser.ru/xml/fictionbook/2.0}genre ).
example.fb2:2: element program-used: Schemas validity error : Element '{http://www.gribuser.ru/xml/fictionbook/2.0}program-used': This element is not expected. Expected is ( {http://www.gribuser.ru/xml/fictionbook/2.0}author ).
example.fb2:2: element annotation: Schemas validity error : Element '{http://www.gribuser.ru/xml/fictionbook/2.0}annotation': This element is not expected. Expected is one of ( {http://www.gribuser.ru/xml/fictionbook/2.0}epigraph, {http://www.gribuser.ru/xml/fictionbook/2.0}section ).

I.e. 
1. No genre specified. FB2 requires at least one genre.
2. author should go before book-title, not vice versa
3. Document creation date is required in the document-info
4. id is required
5. annotation tag should go into the title-info and contain an abstract, not into first page in the body.

jgm:
Pinging @astanin who wrote the FB2 writer

astanin:
I'll look into this.

@vbwagner What is this FictionBook2.1xml schema file that you use? Is it the same as [FictionBook2.xsd](http://gribuser.ru/xml/fictionbook/2.0/xsd/FictionBook2.xsd)?

vbwagner:
FictionBook 2.1 is updated version, which includes support for tables, and 'code' (preformatted fixed witdth text). Sometime ago I'v complained that pandoc formats something as 'code' element, which doesn't present in the FictionBook 2 element. and was advised by LiveJournal user sastanin 
(weren't it you?) to switch to 2.1 version. 

[...]

PR #4539, opened 2018-04-07
Add FB2 reader

labdsf:
Not ready, PR is just to make sure nobody else starts from scratch

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -304,6 +304,8 @@ extra-source-files:
                  test/fb2/images-embedded.html
                  test/fb2/images-embedded.fb2
                  test/fb2/test-small.png
+                 test/fb2/reader/*.fb2
+                 test/fb2/reader/*.native
                  test/fb2/test.jpg
                  test/docx/*.docx
                  test/docx/golden/*.docx
@@ -448,6 +450,7 @@ library
                    Text.Pandoc.Readers.Odt,
                    Text.Pandoc.Readers.EPUB,
                    Text.Pandoc.Readers.Muse,
+                   Text.Pandoc.Readers.FB2,
                    Text.Pandoc.Writers,
                    Text.Pandoc.Writers.Native,
                    Text.Pandoc.Writers.Docbook,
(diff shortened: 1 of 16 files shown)


labdsf:
@jgm 
This PR is ready for merge.

What is not supported yet (as listed in the comment in the code):

- Tables: nothing is produced for them at all. FBReader does not display them anyway (while CoolReader does) and they are rarely used
- Named styles (have not seen them used anyway, don't know what they are for)
- ID attribute for some elements (e.g. paragraphs, should add <span> to them in this case probably)

Reader errs out on `writer.fb2`, but that is the problem of FB2 writer, see issue #2424 Problems are `<empty-line/>` inside paragraph (fixed by #4522, not merged), nested `<cite>` and markup outside paragraphs (after "Interpreted markdown in a table"). Once these 3 places are manually corrected, reader parses the file successfully.

jgm:
OK, great. I merged the other change as well.

jgm:
I'm getting:
```
% pandoc -f fb2 test/writer.fb2 
Couldn't parse FB2 file: unexpected element cite in cite.
```
Can't FB2 handle nested block quotes?
If not, is there a more graceful way of handling this?

labdsf:
FB2 schema does not allow `<cite>` in `<cite>`. It should be fixed in writer instead.

In reader it may be better to output "skipped content" warning instead of error. The problem is that it accepts Parsec's SourcePos. But I don't use Parsec in FB2 reader and I don't know the position in file.

I may add yet another `LogMessage` constructor, similar to `DocxParserWarning String`. Is it a good solution?

murrrf:
There are some problems with the conversion from fb2.

1. The <id> tag specified by the author, since version 2.2, is not recognized by the conversion, causing the message "Could not parse FB2 file: unexpected element id in author."
2. The sections of the fb2-file are converted to <div class = "section"> </ div>, and this text needs to be deleted manually.

labdsf:
@murrrf Please file new issues instead of commenting in closed PRs, it is easier to track that way.

> The tag specified by the author, since version 2.2, is not recognized by the conversion, causing the message "Could not parse FB2 file: unexpected element id in author."

IDs are not parsed indeed, there is even a TODO in the source file.

> The sections of the fb2-file are converted to </ div>, and this text needs to be deleted manually.

You can remove them with a filter, not manually. It is not a bug, FB2 may contain section with an ID but without a title, and it is the only way to represent it.

PR #4554, opened 2018-04-16
flatten nested inlines following readability criteria, closes #4368

### LINKED ISSUE #4368: nested inline styles turn into extra RST text

danse:
after the refactoring we can apply transformations to inlines in a bottom-up direction, thus it becomes possible to transform inlines flattening them according to the the readability criteria discussed in #4368

### DIFF
--- a/test/writer.rst
+++ b/test/writer.rst
@@ -615,21 +615,21 @@ This is *emphasized*, and so *is this*.
 
 This is **strong**, and so **is this**.
 
-An *`emphasized link </url>`__*.
+An `emphasized link </url>`__.
 
-***This is strong and em.***
+**This is strong and em.**
 
-So is ***this*** word.
+So is **this** word.
 
-***This is strong and em.***
+**This is strong and em.**
 
-So is ***this*** word.
+So is **this** word.
 
 This is code: ``>``, ``$``, ``\``, ``\$``, ``<html>``.
 
-[STRIKEOUT:This is *strikeout*.]
+[STRIKEOUT:This is strikeout.]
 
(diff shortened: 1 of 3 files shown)


test/writer.rst
```diff
@@ -649,8 +649,8 @@ Smart quotes, ellipses, dashes
 
 ‘He said, “I want to go.”’ Were you alive in the 70’s?
 
-Here is some quoted ‘``code``’ and a “`quoted
-link <http://example.com/?foo=1&bar=2>`__”.
+Here is some quoted ‘``code``’ and a `quoted
```
jgm: Is there really any reason to do this flattening for Quoted?
The original lines in this test are perfectly valid RST, are they not?

danse: good point, i didn't realise that quotes are not RST inlines. i modified the `combine` function as you suggested

danse:
thanks, work on this has been long. Unfortunately testing this once again on a set of documents i found out that it has a serious issue: the flatten function drops inline parents that are empty.

the related test and the fix are in #4603. i wonder whether it makes sense to add empty images to the `test/writer.*` document, as it's a case of valid empty parent which can sometimes be overlooked

PR #4569, opened 2018-04-20
DocBook Reader: Read Latex math as output by asciidoctor

expipiplus1:
`informalequation` can appear at the top level, in which case it is converted
into a standalone paragraph.

This is an example of some docbook Latex math:

It's duplicated between a `alt` and `mathphrase` tags.

```
<informalequation>
<alt><![CDATA[m = \max\left( \left| { {\partial z_f} \over {\partial x_f} } \right|,
               \left| { {\partial z_f} \over {\partial y_f} } \right|
       \right).]]></alt>
<mathphrase><![CDATA[m = \max\left( \left| { {\partial z_f} \over {\partial x_f} } \right|,
               \left| { {\partial z_f} \over {\partial y_f} } \right|
       \right).]]></mathphrase>
</informalequation>
```

### DIFF
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -16,7 +16,7 @@ import Text.Pandoc.Builder
 import Text.Pandoc.Class (PandocMonad)
 import Text.Pandoc.Options
 import Text.Pandoc.Shared (crFilter, safeRead)
-import Text.TeXMath (readMathML, writeTeX)
+import Text.TeXMath (readMathML, readTeX, writeTeX)
 import Text.XML.Light
 
 {-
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1058,25 +1058,44 @@ parseInline (Elem e) =
             descendantContent name = maybe "???" strContent
                                    . filterElementName (\n -> qName n == name)
 
+-- | Extract a math equation from an element
+--
+-- asciidoc can generate Latex math in CDATA sections.
+--
+-- Note that if some MathML can't be parsed it is silently ignored!
```
jgm: "silently ignored" sounds non-ideal.
When we can't parse tex math, we:
a) issue a warning,
b) include the raw tex.
(See the LaTeX reader.)  b) might make less sense here -- nobody wants to have raw MathML in their document -- but we could still do a), right?

jgm: I went ahead and merged  this, but it would be great if you could do another PR that adds a warning when math is being ignored, using our standard mechanism for doing that.

expipiplus1:
Thanks for merging this!

The ignoring behaviour for mathml was there before, but uncommented. I can
have a look at adding a warning anyway next week.

PR #4571, opened 2018-04-20
Haddock Writer: Use proper haddock math escapes

expipiplus1:
As implemented here: https://github.com/haskell/haddock/pull/465/commits/3f50b955324bd4b42f88a421f0203bc46a3ccf64

### DIFF
--- a/src/Text/Pandoc/Writers/Haddock.hs
+++ b/src/Text/Pandoc/Writers/Haddock.hs
@@ -251,8 +251,8 @@ inlineToHaddock _ (Str str) =
   return $ text $ escapeString str
 inlineToHaddock _ (Math mt str) =
   return $ case mt of
-    DisplayMath -> cr <> "\\\\[" <> text str <> "\\\\]" <> cr
-    InlineMath  -> ("\\(" <> text str <> "\\)")
+    DisplayMath -> cr <> "\\[" <> text str <> "\\]" <> cr
+    InlineMath  -> "\\(" <> text str <> "\\)"
 inlineToHaddock _ il@(RawInline f str)
   | f == "haddock" = return $ text str
   | otherwise = do
--- a/src/Text/Pandoc/Writers/Haddock.hs
+++ b/src/Text/Pandoc/Writers/Haddock.hs
@@ -45,7 +45,6 @@ import Text.Pandoc.Options
 import Text.Pandoc.Pretty
 import Text.Pandoc.Shared
 import Text.Pandoc.Templates (renderTemplate')
-import Text.Pandoc.Writers.Math (texMathToInlines)
 import Text.Pandoc.Writers.Shared
 
 type Notes = [[Block]]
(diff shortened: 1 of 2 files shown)


wilx:
@expipiplus1 You need to adjust test cases to match the new output. See the Travis CI results.

expipiplus1:
Hmm, it seems as though the output isn't strictly an improvement...

I certainly prefer `/α/ ∧ /ω/` over `\(\alpha \wedge \omega\)`

This is the test diff:

```
          --- writer.haddock
          +++ /home/travis/build/jgm/pandoc/sourcedist/dist/build/pandoc/pandoc --quiet testsuite.native -r native -w haddock --columns=78 --variable pandoc-version= -s
          + 458 -   \(2+2=4\)
          + 459 -   \(x \in y\)
          + 460 -   \(\alpha \wedge \omega\)
          + 461 -   \(223\)
          + 462 -   \(p\)-Tree
          - 458 -   2 + 2 = 4
          - 459 -   /x/ ∈ /y/
          - 460 -   /α/ ∧ /ω/
          - 461 -   223
          - 462 -   /p/-Tree
          + 464     \\[\frac{d}{dx}f(x)=\lim_{h\to 0}\frac{f(x+h)-f(x)}{h}\\]
          + 465 -   Here’s one that has a line break in it: \(\alpha + \omega \times x^2\).
          - 464     $$\\frac{d}{dx}f(x)=\\lim_{h\\to 0}\\frac{f(x+h)-f(x)}{h}$$
          - 465 -   Here’s one that has a line break in it: /α/ + /ω/ × /x/2.
```

expipiplus1:
It's obviously a big improvement for big blocks of Latex though:

In action here: https://hackage.haskell.org/package/vulkan-2.0.0.0/docs/Graphics-Vulkan-Core10-CommandBufferBuilding.html#v:vkCmdSetDepthBias

jgm:
I suppose we could make the haddock output sensitive to the writerHTMLMathMethod, which is set by flags like `--mathjax`.  This would allow haddock users the option of "faking it with unicode" or including raw latex math.

expipiplus1:
Ah, that is a mistake. I think it should be \\( too.

It would be nice perhaps if there was some simple metric of how faithful the non latex math renders. Lots of the haddock I read is just in ascii form and keeping that clear is useful.

expipiplus1:
Actually, I think that's something to delegate to the tex writer. It could replace things like `\alpha` with `α`.

jgm:
Pandoc uses the texmath for math conversion.  One conversion mode
is to unicode plain text.  This is what is currently used.
However, it will fail when the math gets moderately complex
(e.g. fractions, which can't easily be represented in a unicode
string). For the most uniform results, it's best to use latex
output throughout, probably.

jgm:
I merged this manually, thanks!

expipiplus1:
Thank!

PR #4573, opened 2018-04-21
Haddock Writer: Support Haddock tables

expipiplus1:


expipiplus1:
These are a fairly recent feature in haddock.

It might be good to workaround/wait for a fix for https://github.com/haskell/haddock/issues/801 before merging.

expipiplus1:
I think that this can be closed now, superseded by be12ae3bca, @jgm, is this the case?

jgm:
be12ae3 was March 16, 2018.  This PR was submitted April 20, 2018.
Perhaps this PR didn't take into account the recent change?
But be12ae3 does indeed add table support to the haddock writer, so if you agree and there's nothing I'm missing, I'll be happy to close this.

expipiplus1:
@jgm You are quite right, that PR emulated table support in haddock by using code blocks.

I have rebased this PR. Sorry it took a while to get round to this.

jgm:
Thanks.  I understand now.
There are two commits here: the first changes `gridTable` in `T.P.Writers.Shared` and this would be a breaking API change requiring a version bump to 2.10.  But I didn't see why this is necessary for the Haddock table change. Can you explain?

expipiplus1:
IIRC it's there because Haddock doesn't handle intra-word splitting very well, for example the table 

```
+--------+
| A      |
| ligned |
+--------+
```

Wouldn't be rendered into HTML by Haddock as a cell containing `Aligned`, but as a cell containing `A ligned`. I'm not sure what the desired behavior is for the other writers, so I added this parameter to keep them the same. 

To avoid breaking the API I suppose a function `gridTableNoStrictWrapping` could be added and both that and `gridTable` could call out to an internal function taking this `strictWrapping` parameter.

jgm:
Well, none of the formats we support would render this table as a cells containing `Aligned`.
Since you contributed your original patch there have been changes to that grid table system (if I recall) that try to avoid this kind of thing, so that part of your PR is probably no longer needed.
Can you remove that part from this PR?  We can reconsider this sort of change later, but it's not really haddock-specific.

expipiplus1:
@jgm, you're quite right, I have rebased without the table wrapping patch.

jgm:
thanks!

expipiplus1:
Thanks!

PR #4574, opened 2018-04-21
Haddock Writer: Use markdown style links where possible

expipiplus1:


expipiplus1:
@jgm is it possible for you to have a look at this please?

tarleb:
Hey @expipiplus1, thanks for your contribution! The likely reason that this PR has not received feedback yet is that the continuous integration tests fail. There is usually an implicit expectation that pull request should be mergeable before someone takes the time to review. In this case, the failure is due to a compiler warning, as the value `linktext` is no longer used. You'll probably have to update the tests, too.

Other than that, this looks good to me, as[Haddock supports Markdown links](https://haskell-haddock.readthedocs.io/en/latest/markup.html#links) nowadays.

jgm:
My main substantive questions are:

- Has Haddock supported markdown-style links
  long enough that we can assume this will work for
  people?
- Are markdown-style links now the preferred option for
  Haddock?
- Pandoc used to use `<http://example.com description>`.  
  Did this ever work? Does it still work?  Has it been
  deprecated in favor of markdown-style links?

expipiplus1:
Sorry for leaving this so long!

- I can say that Haddock has supported this since at least Apr 21, 2018 :)
- I'm not sure if they're preferred, I can't remember precisely why I wanted this, but there was some deficiency in the existing haddock links, like wonky parsing of spaces or `<` `>` in the description, something like that.

I'd be happy to tidy the PR if there aren't objections, otherwise happy to have this closed :)

jgm:
Well, I guess I'll close this -- but if compelling reasons can be given for switching to md style links, a new issue can be opened explaining what they are.

expipiplus1:
sounds good to me!

PR #4577, opened 2018-04-21
MANUAL: consolidate input/output format documentation

mb21:
These are two related commits, both changing the input/output formats documentation in the MANUAL.

The first, consolidates the list of input/output formats in one place (namely, the options). There were a couple of formats missing there. Also, my hope is that people will now read the part of the intro about how pandoc works with the AST etc., if they don't have to scroll over the looong list of formats first.

The second commit uses the unordered list for the formats instead, and orders them alphabetically. This makes it much easier to read, at the expense of taking a bit more space (but scrolling is free on computers...).

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -219,21 +219,39 @@ General options
 
 `-f` *FORMAT*, `-r` *FORMAT*, `--from=`*FORMAT*, `--read=`*FORMAT*
 
-:   Specify input format.  *FORMAT* can be `native` (native Haskell),
-    `json` (JSON version of native AST), `markdown` ([Pandoc's
-    Markdown]), `markdown_strict` (original unextended
-    [Markdown]), `markdown_phpextra` ([PHP Markdown Extra]),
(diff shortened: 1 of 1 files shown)


mb21:
> we currently generate the README.md from this section, and I think it's useful for the list to occur there

ah, sure. Well, the formats could also be added to the README.md differently: e.g. by extracting them from the `-f` / `-t` options (or even from `pandoc --list-input-format`).

Another approach would be to leave the formats in the preamble, but change the formatting to the unordered list. This would address at least the first of the two problems this pull is trying to solve, namely:

1. make input/output formats more readable: a bullet list is just so much easier to parse for the human mind than that unwieldy paragraph with all its commas, brackets and links (have a look at the generated HTML in a browser to see the effect).
2. have one canonical place in the MANUAL where the formats are listed, to remove redundancy for the reader, as well as for us when updating the documentation (as I mentioned, we've managed to forget a couple of formats in the options part which were listed in the preamble – not good.)

jgm:
I'll just merge this and change the filter that creates the README.md, so it collects the formats from --to and --from.

mb21:
okay great :)

PR #4619, opened 2018-05-02
catch IO errors when writing media files, closes #4559

### LINKED ISSUE #4559: fatal error when an ODT image cannot be fetched
when an ODT file contains an image that cannot be fetched and the `--extract-media` option is used, a runtime error prevents pandoc from writing any converted output. The following line is printed:

    pandoc: media/.: openBinaryFile: inappropriate type (Is a directory)

danse:
if we do not catch these errors, any malformed entry in a media bag
could cause the loss of a whole document output. an example of
malformed entry is an entry with an empty file path.

for the error logging i looked for a way to preserve information from the IO error. An IO error like `media/.: openBinaryFile: inappropriate type (Is a directory)` can give helpful troubleshooting information to the users in case of errors in a path or problems with permission

### DIFF
--- a/src/Text/Pandoc/Class.hs
+++ b/src/Text/Pandoc/Class.hs
@@ -477,6 +477,14 @@ liftIOError f u = do
          Left e  -> throwError $ PandocIOError u e
          Right r -> return r
 
+-- | Show potential IO errors to the user continuing execution anyway
+logIOError :: IO () -> PandocIO ()
+logIOError f = do
+  res <- liftIO $ tryIOError f
+  case res of
+    Left e -> report $ IgnoredIOError (E.displayException e)
+    Right _ -> pure ()
+
 instance PandocMonad PandocIO where
   lookupEnv = liftIO . IO.lookupEnv
   getCurrentTime = liftIO IO.getCurrentTime
@@ -862,7 +870,7 @@ writeMedia dir mediabag subpath = do
        Just (_, bs) -> do
          report $ Extracting fullpath
          liftIOError (createDirectoryIfMissing True) (takeDirectory fullpath)
-         liftIOError (\p -> BL.writeFile p bs) fullpath
+         logIOError $ BL.writeFile fullpath bs
 
 adjustImagePath :: FilePath -> [FilePath] -> Inline -> Inline
 adjustImagePath dir paths (Image attr lab (src, tit))
--- a/src/Text/Pandoc/Logging.hs
+++ b/src/Text/Pandoc/Logging.hs
@@ -85,6 +85,7 @@ data LogMessage =
   | InlineNotRendered Inline
   | BlockNotRendered Block
   | DocxParserWarning String
+  | IgnoredIOError String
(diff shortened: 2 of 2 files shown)


wilx:
@danse You need to fix the warning. See Travis CI output.

danse:
oops thanks for the heads up @wilx i fixed it let's wait for the new build

jgm:
Under these conditions, will we still get a CouldNotFetchResource warning, or just the IgnoredIOError?

I think it would be best if we still got CouldNotFetchResource as well.

danse:
Yes we still get the warning which is triggered by an independent part of the software, within `fetchMediaResource` if i remember correctly. Using the test file from #4559, the output with this change is the following

```
[WARNING] Could not fetch resource './ObjectReplacements/Object 1': replacing image with description
[WARNING] IO Error (ignored): media/.: openBinaryFile: inappropriate type (Is a directory)
```

danse:
the warning is triggered when the resource is not found. but then the ODT reader fetches the resource independently and stores an empty path in the media bag. when we try to use the path later on we get the error that is ignored in this change

jgm:
OK, good.  I'm still not sure what I think about this, though.
One could make a case for the view that this is a bug in pandoc,
and should therefore lead to an error rather than a warning.
Pandoc should handle this case better.

jgm:
On reflection, not all IO errors will be pandoc bugs (a file might not have the right permissions, e.g.), so let's do this.

src/Text/Pandoc/Class.hs
```diff
@@ -477,6 +477,14 @@ liftIOError f u = do
          Left e  -> throwError $ PandocIOError u e
          Right r -> return r
 
+-- | Show potential IO errors to the user continuing execution anyway
+logIOError :: IO a -> PandocIO ()
+logIOError f = do
+  res <- liftIO $ tryIOError f
+  case res of
+    Left e -> report $ IgnoredIOError (E.displayException e)
+    Right _ -> pure ()
```
jgm: Why not change type to `IO a -> PandocIO a` and change this line to
```
Right x -> return x
```
That makes the function more general.
Either that, or change the type to `IO () -> PandocIO ()`.

danse: `IO a -> Pandoc a` would require a default value when the operation fails, so i opted for `IO () -> Pandoc ()`, i amended and rebased

danse:
> One could make a case for the view that this is a bug in pandoc, and should therefore lead to an error rather than a warning. Pandoc should handle this case better

I agree that this is a fault in pandoc, yet the conversion process can continue. Why would we want a fault to be fatal, rather than preserving what is unaffected? We could rename `IgnoredIOError` to `RecoverableError`, it might help us to keep in mind the idea. It could be shown to the user as `IO Error: <description>, trying to continue anyway...`. It sounds funny but it's clear. I agree that, if we had more logging levels, we would use the error level rather than the warning level for this one

PR #4623, opened 2018-05-03
LaTeX: Repeat caption of a table consistently.

### LINKED ISSUE #4558: caption for table with header is not repeated on pagebreak in latex output
Hey there,

I noticed, that a long table is broken into pages, but the caption is not repeated **if** the table has a header. **If not** the caption is repeated. Two questions:

1. Is this intended behavior?
2. How do I force repetition of the caption for tables with headers?

[min_example.zip](https://github.com/jgm/pandoc/files/1918842/min_example.zip) attached. The template is the latest default from 7f97b04862ffbe2bb9efaa648eee284f7949d20d.

[This markdown](https://pastebin.com/spRMxxHR) creates [tables.pdf](https://github.com/jgm/pandoc/files/1918821/tables.pdf).

Best, Jonas

wilx:
This fixes #4558.

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX/Table.hs
+++ b/src/Text/Pandoc/Writers/LaTeX/Table.hs
@@ -28,7 +28,7 @@ import Text.Pandoc.Definition
 import Text.DocLayout
   ( Doc, braces, brackets, cr, empty, hcat, hsep, isEmpty, literal, nest
   , text, vcat, ($$))
-import Text.Pandoc.Shared (splitBy, tshow)
+import Text.Pandoc.Shared (deNote, splitBy, tshow)
 import Text.Pandoc.Walk (walk, query)
 import Data.Monoid (Any(..))
 import Text.Pandoc.Writers.LaTeX.Caption (getCaption)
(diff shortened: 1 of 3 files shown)


src/Text/Pandoc/Writers/LaTeX/Table.hs
```diff
@@ -193,9 +208,12 @@ tableToLaTeXLongtable colDesc mkHead mkRow capt thead tbodies tfoot = do
...
                   , foot'
                   ]
-             else [ foot'
+             else (if hasBottomCaption && not (isEmpty repeatedCapt)
+                      then [ repeatedFoot, "\\endfoot" ]
+                      else []) <>
+                  [ foot'
                   , "\\endlastfoot"
-                  ,  vcat rows'
+                  , vcat rows'
```
jgm, line 216: The repeated head is already incorporated in `head'`, whereas the repeated foot isn't already incorporated in `foot'`, and hence needs adding here. To me, this makes the logic less clear. It would be better to do it the same way in both places.

Note also that lns 211-13 could be simplified using guards:
```
[repeatedFoot, "\\endfoot" | hasBottomCaption && not (isEmpty repeatedCapt)]
```

jgm, line 216: The repeated head is already incorporated in `head'`, whereas the repeated foot isn't already incorporated in `foot'`, and hence needs adding here. To me, this makes the logic less clear. It would be better to do it the same way in both places.

Note also that lns 211-13 could be simplified using guards:
```
[repeatedFoot, "\\endfoot" | hasBottomCaption && not (isEmpty repeatedCapt)]
```

PR #4638, opened 2018-05-09
groff (man) Reader

### LINKED ISSUE #1484: Groff man reader
Is there a reason why pandoc doesn't support conversions from groff manpages to other formats? What would be involved in building a parser?

Yanpas:
Fixes #1484

Example of generated poll (2) man page -> html:
image


jgm:
This is welcome. But how complete is it?  To be included in pandoc, it should handle all the main document elements that pandoc supports:  emphasis, links, inline code, various types of lists, headers, code blocks, block quotes, tables, and so on.  It should also include automated tests.  It's too much of a maintenance hassle to merge an incomplete parser.

Yanpas:
It is complete in the way that it parses kernel.org man files. Groff format is pretty limited so it doesn't have links, lists etc.


I was comparing the results with man2html utility (exists in debian-based distros repos)

PS is merge possible if this reader will be marked as 'beta'? Anyway it can be merged, but be disabled.

Here are format man pages:
* https://linux.die.net/man/7/groff
* https://linux.die.net/man/7/man

mb21:
Can it roundtrip? i.e. read the `man-reader.man` file generated by the following?

    pandoc test/writer.native -f native -t man -o test/man-reader.man

If so, maybe add that as a test... along with the `test/man-reader.native` `-f man -t native` generates...

jgm:
Yan Pashkovsky <notifications@github.com> writes:

> It is complete in the way that it parses kernel.org man files. Groff format is pretty limited so it doesn't have links, lists etc.

Correct about links.  Only sort of correct about lists.  There are
standard ways of making lists in groff man: see the pandoc man page, for
example, which is generated from MANUAL.txt and has definition lists,
bullet lists, and ordered lists.  It would be nice if such things could
be recognized and parsed as the corresponding sorts of lists. Tables are
also supported (via tbl); there's an example of this in the pandoc man
page too (search for `.TS`).

Anyway, a good thing to aim for would be approximate round-tripping with
the man writer.

I hope you can understand why I'm reluctant to merge this until it's
more complete.  I've had bad experiences in the past merging
half-completed work and then, essentially, having to finish it myself.
Why don't we mark this PR "[WIP]" and when you've got it to a more
complete stage, we can review it?  Since you're doing a whole module, it
should be relatively easy to keep it rebased on master.  (Changes
elsewhere in pandoc are unlikely to affect this.)

More comprehensive test coverage is also a must!

Anyway, thanks very much for working on this; it will be a really
useful contribution to pandoc.

Yanpas:
> I hope you can understand why I'm reluctant to merge this until it's more complete.

I understand you and agree with you point of view. I'll keep working in this branch.

Can you do quick code review please? Maybe I'm using some pandoc, parsec or even Haskell 
 APIs antipatterns (e.g. I've seen that tests use builder functions, not data constructors). BTW I plan to add tokenization phase to the reader, since parsing text directly to AST makes the code messy.

> Can it roundtrip?

Technically yes but in fact the quality is poor
image

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,284 @@
...
+  isCode (Code _ _) = True
+  isCode _ = False
+
+parseMan :: PandocMonad m => ManParser m Pandoc
+parseMan = do
+  blocks <- createParas <$> many (parseMacro <|> parseLine)
+  parserst <- pState <$> getState
+  return $ Pandoc (stateMeta parserst) blocks
+
+-- | Read man (troff) from an input string and return a Pandoc document.
```
jgm: In pandoc we typically put the main function (readMan, here) at the top of the source file.

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,284 @@
...
+                else [CodeBlock nullAttr (intercalate "\n" $ codeToStr <$> cds)]
+        paras = [Para (intersperse (Str " ") ncds)]
+    in cbs ++ paras ++ plainInlinesToPara oth
+
+  isCode (Code _ _) = True
+  isCode _ = False
+
+parseMan :: PandocMonad m => ManParser m Pandoc
+parseMan = do
+  blocks <- createParas <$> many (parseMacro <|> parseLine)
```
jgm: This architecture assumes that each line with a macro will correspond to one Block.  I don't think that's going to work in all cases (e.g. lists, tables).

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,284 @@
...
+roffInline rst str
+  | null str && (not $ inCodeBlock rst) = []
+  | inCodeBlock rst = [Code nullAttr str]
+  | otherwise       = case fontKind rst of
+    Regular    -> [Str str]
+    Italic     -> [Emph [Str str]]
+    Bold       -> [Strong [Str str]]
+    ItalicBold -> [Emph [Strong [Str str]]]
+
+parseLine :: PandocMonad m => ManParser m Block
```
jgm: Instead of having the parsers return a Block, you should have them return Blocks (from Text.Pandoc.Builder).  All of the readers are structured this way now.  Use functions from Text.Pandoc.Builder to create the return values, and use monoidal mappend to concatenate them.  A 'Blocks' is essentially a wrapped sequence of Block elements (perhaps empty).  This gives you a lot more flexibility:  a parser can return no blocks (mempty) or several.

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,284 @@
...
+    parseLineParts :: PandocMonad m => ManParser m [Inline]
+    parseLineParts = do
+      lnpart <- many $ noneOf "\n\\"
+      ManState {rState = roffSt} <- getState
+      let inls = roffInline roffSt lnpart
+      others <- backSlash <|> return []
+      return $ inls ++ others
+    
+    backSlash :: PandocMonad m => ManParser m [Inline]
+    backSlash = do
```
jgm: I didn't look at this in detail, but I wasn't sure it handled things like `\f[CB]`, which is valid groff.

Yanpas:
Thanks for the feedback, I'll continue improving this PR

Yanpas:
I've made a progress on this PR:
* Splitted lexing and parsing, the code has become leaner. I've introduced `readWithMTokens` function since pandoc function only support `[Char]` type parsers
* Handle more escape sequences
* Ordered/Bullet list support

Here is what's left:
* Add more tests
* strong/emphasized text in headers support (macro parsing needs rework)
* Ensure that all posix and linux man files are parsed successfully.

Here is what I currently obtain after executing `pandoc.exe -f man .\test\writer.man -o dist\round.html` (rename pdf to html)
[round.html.pdf](https://github.com/jgm/pandoc/files/2042542/round.html.pdf)

jgm:
Excellent to hear about the progress!

jgm:
This has come pretty far -- I'm just wondering if it's still in progress?

Yanpas:
Currently this reader parses nested lists, urls and everything that was mentioned as mandatory previously. Formally this pr lacks unit tests. But I need to perform a lot of stress tests (several hundred linux and posix man pages). I ve noticed that some of them either a bit malformed, either I didn't get roff syntax right (I think i should find some roff grammar definition).

Sincerely I am very busy at the moment, so hope to finish everything by the middle of the fall.

jgm:
Sounds great. Thanks for the update!

wilx:
How are the stress tests going?

Yanpas:
Currently I have limited free time

jgm:
Note: I've added an unexported module, Text.Pandoc.Groff, to hold shared groff functions. Currently it just has a function to escape non-ASCII characters.

Yanpas:
Good news, posix (1149 files) and linux (2353) man files are now parsed successfully. Only one linux file (zic.8) seems to be broken, it has line with two dots `..`. Man files were extracted from ubuntu packages

Yanpas:
That means without parsing error. I've looked at htmls of 20 random man files and they look nice, very similar to man2html ones.

I've also noticed that there are a lot of link-like files which contain only one line for example `.so man3/rpc.3`. Such files generate empty html (without html, head and body).

jgm:
That sounds like great progress.  Do you think it's
ready to merge?

Yanpas:
Yes, I guess so. I will add unit tests for the grammar and it would be merge-ready.

jgm:
Sounds good. I'll wait for the tests, then, before reviewing for merge.

Yanpas:
I've added tests, I think it's time to code review :)

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,552 @@
...
+   Copyright   : Copyright (C) 2018 Yan Pashkovsky
+   License     : GNU GPL, version 2 or above
+
+   Maintainer  : Yan Pashkovsky <yanp.bugz@gmail.com>
+   Stability   : WIP
+   Portability : portable
+
+Conversion of man to 'Pandoc' document.
+-}
+module Text.Pandoc.Readers.Man (readMan) where --testFile
```
jgm: Not sure what `--testFile` is doing here?

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,552 @@
...
+readMan opts txt = do
+  eithertokens <- readWithM lexMan def (T.unpack $ crFilter txt)
+  case eithertokens of
+    Right tokenz -> do
+      let state = def {stateOptions = opts} :: ParserState
+      eitherdoc <- readWithMTokens parseMan state tokenz
+      case eitherdoc of
+        Right doc -> return doc
+        Left e -> throwError e
+    Left e       -> throwError e
```
jgm: for simple Either matches like l. 129, you might consider using `either`:
```haskell
either throwError return eitherdoc
```
You could even do:
```haskell
either throwError return <$> readWithMTokens parseMan state tokenz
```
and avoid the named `eitherdoc`.

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,552 @@
...
+
+  where
+
+  readWithMTokens :: PandocMonad m
+          => ParserT [ManToken] ParserState m a  -- ^ parser
+          -> ParserState                         -- ^ initial state
+          -> [ManToken]                       -- ^ input
+          -> m (Either PandocError a)
+  readWithMTokens parser state input =
+    mapLeft (PandocParsecError . (intercalate "\n") $ show <$> input) `liftM` runParserT parser state "source" input
```
jgm: Try to stay within 80 columns.

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,552 @@
...
+-- String -> ManToken function
+--
+
+lexMan :: PandocMonad m => ManLexer m [ManToken]
+lexMan = many (lexComment <|> lexMacro <|> lexLine <|> lexEmptyLine)
+
+parseMan :: PandocMonad m => ManParser m Pandoc
+parseMan = do
+  let parsers = [ try parseList, parseTitle, parsePara, parseSkippedContent
+                , try parseCodeBlock, parseHeader, parseSkipMacro]
```
jgm: These parsers return a `Block`.  In the other pandoc readers, we return a `Blocks`, generated using the functions in Text.Pandoc.Builder.  (Ditto for `Inline`).  This is recommended for several reasons:  (a) it will make changes to the underlying types less painful to implement, (b) you automatically get things like merging of adjacent Space elements and elimination of Null blocks, (c) your parsers can return empty sequences (`mempty`) or sequences containing more than one block or inline.

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,552 @@
...
+  parserst <- getState
+  return $ Pandoc (stateMeta parserst) (filter (not . isNull) blocks)
+
+  where
+
+  isNull Null = True
+  isNull _    = False
+
+eofline :: Stream s m Char => ParsecT s u m ()
+eofline = (newline >> return ()) <|> eof
```
jgm: You can also write `void newline` (using `void` from `Control.Monad`) or `() <$ newline`.

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,552 @@
...
+
+spacetab :: Stream s m Char => ParsecT s u m Char
+spacetab = char ' ' <|> char '\t'
+
+-- TODO handle more cases
```
jgm: Can you document what is missing, more specifically?

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,552 @@
...
+              ]
+
+  escFont :: PandocMonad m => ManLexer m EscapeThing
+  escFont = do
+    char 'f'
+    font <- choice [ letterFont
+          , char '(' >> anyChar >> anyChar >> return Regular
+          , try (char '[' >> letterFont >>= \f -> char ']'  >> return f)
+          , try $ string "[BI]" >> return ItalicBold
+          , char '[' >> many letter >> char ']'  >> return Regular
```
jgm: This will miss some font specifications one can find in the wild.
E.g., `\f[BC]` for bold + typewriter font.  ItalicBold isn't the only possible font combination.
Would it make more sense to have independently combinable components, and use a set?

src/Text/Pandoc/Readers/Man.hs
```diff
@@ -0,0 +1,552 @@
...
+  escUnknown :: PandocMonad m => ManLexer m EscapeThing
+  escUnknown = do
+    c <- anyChar
+    pos <- getPosition
+    logOutput $ SkippedContent ("Unknown escape sequence \\" ++ [c]) pos
+    return ENothing
+
+currentFont :: PandocMonad m => ManLexer m FontKind
+currentFont = do
+  RoffState {fontKind = fk} <- getState
```
jgm: A better idiom here is:
```haskell
   fk <- fontKind <$> getState
```
In fact you could just write:
```haskell
currentFont = fontKind <$> getState
```

src/Text/Pandoc/Readers/Man.hs
Yanpas: Placed this debug function to eliminate warnings. I'll remove this comments

src/Text/Pandoc/Readers/Man.hs
Yanpas: There are a lot of escapes from man/7/groff, I've included only which occur in existant man files or able to handle

jgm:
> BTW I'm trying to implement logggin CouldNotLoadIncludeFile message in the lexing phase. I've packed ParserState inside RoffState, but it seems like readWithM doesn'y do any IO on modified state. Do you know how to fix that?

Probably you just need a `lift` to get to the inner monad.  `lift . logMessage`.

Yanpas:
I've replaced `logMessage` with `logOutput` and now see the logs 🤔

jgm:
I built this locally.  First thing I tried was the pandoc man page pandoc.1, and I got these warnings:
```
[INFO] Skipped 'Unknown escape sequence \@' at line 1264 column 3
[INFO] Skipped 'Unknown escape sequence \@' at line 1270 column 3
[INFO] Skipped 'Unknown escape sequence \@' at line 1276 column 3
[INFO] Skipped 'Unknown escape sequence \@' at line 1282 column 3
[INFO] Skipped 'Unknown escape sequence \@' at line 3281 column 32
[INFO] Skipped 'Unknown escape sequence \@' at line 3283 column 35
[INFO] Skipped 'Unknown escape sequence \@' at line 3286 column 14
[INFO] Skipped 'Unknown escape sequence \@' at line 3291 column 4
[INFO] Skipped 'Unknown escape sequence \@' at line 3292 column 4
[INFO] Skipped 'Unknown escape sequence \@' at line 3296 column 4
[INFO] Skipped 'Unknown escape sequence \@' at line 3305 column 4
[INFO] Skipped 'Unknown escape sequence \@' at line 3307 column 8
[INFO] Skipped 'Unknown escape sequence \@' at line 4142 column 37
[INFO] Skipped 'Unknown escape sequence \@' at line 4395 column 7
[INFO] Skipped 'Unknown escape sequence \@' at line 4421 column 26
[INFO] Skipped 'Unknown escape sequence \@' at line 4945 column 51
[INFO] Skipped 'Unknown escape sequence \@' at line 4955 column 21
[INFO] Skipped 'Unknown escape sequence \@' at line 4955 column 51
[INFO] Skipped 'Unknown escape sequence \@' at line 4957 column 16
[INFO] Skipped 'Unknown escape sequence \@' at line 4959 column 16
[INFO] Skipped 'Unknown escape sequence \@' at line 4959 column 28
[INFO] Skipped 'Unknown escape sequence \@' at line 4982 column 46
[INFO] Skipped 'Unknown escape sequence \@' at line 4988 column 25
[INFO] Skipped 'Unknown escape sequence \@' at line 4996 column 3
[INFO] Skipped 'Unknown escape sequence \@' at line 4998 column 3
[INFO] Skipped 'Unknown escape sequence \@' at line 5026 column 7
[INFO] Skipped 'Unknown escape sequence \@' at line 5026 column 17
[INFO] Skipped 'Unknown escape sequence \@' at line 5029 column 3
[INFO] Skipped 'Unknown escape sequence \@' at line 5044 column 7
[INFO] Skipped 'Unknown escape sequence \@' at line 6105 column 44
```
Sure enough, the `@` characters were missing in the code blocks.
So, this escape, at least, needs to be handled.

jgm:
Also, the INFO messages shouldn't have appeared because I wasn't using `--verbose`.
That's because you use `logOutput` directly.  You should use `logMessage` so the log level is respected (we'll need to figure out how to wrangle the types).

jgm:
OK, I see.  Changing `logOutput` to `logMessage` we get:
```
Could not deduce (HasLogMessages RoffState)
        arising from a use of ‘logMessage’
```
In the markdown parser, we want to avoid issuing messages that pertain to paths we're going to parse partially and backtrack on, so we use `logMessage` which updates a state variable (and then we print the log messages at the end, after parsing).

IF you don't have this kind of backtracking, you can just use `report` from Text.Pandoc.Class.

Yanpas:
will stick to `report` function.

Cannot find any mention how to interpret `\@` (http://man7.org/linux/man-pages/man7/groff.7.html). Is it standard?

mb21:
@Yanpas, I think `\@` simply means `@`. Apparently it should be possible to escape it. See e.g. https://github.com/jgm/pandoc/blob/master/man/pandoc.1

Yanpas:
I mean it violates standard, is it ok?

jgm:
Hm, it works (experimentally), but maybe we should be using `\[at]` in the man writer?

jgm:
I'd say it's okay not to support it if it's not standard.
i'll change the writer, though.

jgm:
Are you supporting escapes of the form `\[at]`, `\[ff]`, `\['A]`, `\[oq]`, and so on?
Documented in `groff_char.7`

Yanpas:
Currently no, but it's easy to add them

jgm:
I think this is in good enough shape to merge.  We'll still want to make adjustments, but let's do that in separate PRs.  This is a great addition to pandoc!

Yanpas:
That was interesting and challenging task! Thanks for the assistance.

jgm:
@Yanpas I've added some issues to the tracker for further improvements.  I'm happy to take these on, but if you want to do any of them just indicate on the issue that you're working on it.

PR #4641, opened 2018-05-09
data/templates/default.latex:  `\usepackage{iftex}` for conditionals.

RandomDSdevel:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I wasn't able to clean up _all_ of the conditional processing done based on engine used, but hopefully this reads a bit better nonetheless.  (The odd line out is number 43, as there doesn't seem to be a LaTeX package that allows one to say…:  

```latex
\if{\LuaTeX{} \and \XeTeX{}}
  …
\then
  …
\else
  …
```

…, at least as far as I can tell.)

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -34,15 +34,15 @@ $if(linestretch)$
 \setstretch{$linestretch$}
 $endif$
 \usepackage{amssymb,amsmath}
-\usepackage{ifxetex,ifluatex}
+\usepackage{iftex}
 \usepackage{fixltx2e} % provides \textsubscript
-\ifnum 0\ifxetex 1\fi\ifluatex 1\fi=0 % if pdftex
+\ifPDFTeX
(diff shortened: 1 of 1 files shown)


jgm:
While I agree that the code simplification is nice, I tend to prefer packages that are more commonly available.  ifxetex and ifluatex are part of texlive-latex-base in debian.  iftex is in texlive-generic-extra.  For a lot of people, this template change would require installing an additional package.

RandomDSdevel:
@jgm:  

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Oh, sorry, I didn't think about that, even for MacTeX's BasicTeX distribution (I'm not sure if the same caveat applies there, as I use the full MacTeX distribution myself, but it wouldn't be out of the realm of possibility.)  I'll close this PR, then.

Issue #4657, opened 2018-05-16
Source map for code blocks


ehildenb:
I found this discussion on pandoc-discuss: https://groups.google.com/forum/#!searchin/pandoc-discuss/source$20line$20numbers$20for$20code%7Csort:date/pandoc-discuss/pDs41Da4KjA/SKt35OWyhxAJ

Did anything ever become of it? We use `pandoc` for literate programming in the K language without having to write a literate programming language (Markdown sources, generate K files). See examples at https://github.com/kframework/evm-semantics.

Currently we use `pandoc-tangle`, a filter at https://github.com/ehildenb/pandoc-tangle, for selecting code blocks to be included in each resulting K file (basically `pandoc-tangle` is a `Writer` for code). For each other line of input, blank lines are generated to make the target K files have the same line numbers as the source markdown. This is a bit fragile, and assumes that people write "sane-ish" markdown.

A better solution would be to have access to the source information of each `CodeBlock` as attributes. I have a patch which adds source attributes to `CodeBlock`s in the Markdown reader, along with a command-line flag `--insert-source-attrs` and reader option `readerInsertSourceAttr`. I'm working on tests, let me know if this has a chance getting merged, or if a different approach should be taken.

I've thought about whether it makes sense to add these source attributes for all `Block` which contain `Attr`, and would be willing to do that as well, though I don't think it will be useful for literate programming using Markdown.

jgm:
I can see the use of this.  It might make more sense to develop it as a markdown extension, so you'd do `-f markdown+source_attributes` or something like that (`+code_source_attributes`?).

Adding source attributes to everything is impossible, since only a few things currently have attributes.  Adding just to code blocks seems a bit ad hoc -- it's useful for you, but how generally useful is it?  So I'm not sure what to think about this.  (Of course, I wish I'd designed for source position information from the beginning, and I am doing that in the brand new markdown parsing library I'm working on, commonmark-hs.)

gpoore:
This would be very useful, even if only `CodeBlock`s get source information. I'm currently working on a program that will allow code blocks in arbitrary languages to be executed and optionally include `stdout` and `stderr` in the filtered markdown. Without something like this, I'll be parsing the `--trace` output or using regex to try to determine source information.

The one additional thing that would be helpful for my case is source information for inline `Code`.

ehildenb:
@gpoore have you seen Panpipe? https://hackage.haskell.org/package/panpipe

jgm:
[...]

PR #4659, opened 2018-05-17
[WIP] Markdown reader: extension to insert source code position attributes

ehildenb:
Optionally inserts source-code position attributes for `CodeBlock` in Markdown
sources. The extension is `Ext_code_block_source_position`.

This reads the parser position (`SourcePos name line col`) at the beginning of
parsing a `CodeBlock` and adds the attributes
`[("sourceName", name), ("sourceLine", line), ("sourceColumn", col)]` if the
extension is turned on.

Addressed #4657 

WIP Part
========

Currently, the `sourceName` is returning nothing useful (just the word `source`). It would be nice if it actually returned the name of the source file for generating things like line directives for files generated from multiple literate sources: https://gcc.gnu.org/onlinedocs/cpp/Line-Control.html

Any ideas @jgm? I could either (1) omit the `sourceName` attribute, or (2) figure out what the original file name is (using `stdin` or `-` for standard input). I would prefer ~(1)~(2) for the above reason.


ehildenb:
I'm not a huge fan of inserting source directives which may make the parser more fragile. I think I would prefer a structured representation of it, especially since we have all that information in hand at some point, but I guess that would require much more extensive surgery.

I like the single `sourcepos` attribute, though I'm a bit worried about losing the structure of the information. I think I would prefer something like `STARTLINE:STARTCOL:ENDLINE:ENDCOL` so it's easier to say "split on colons" in a Lua filter.

Should I:

1.   Investigate parsing files individually?
2.   Call it good without having `SOURCENAME` for the moment?

PR #4665, opened 2018-05-21
Adding a variable romanfont

OvidiusCicero:
It would be nice to have a possibility to set the romanfont in xelatex/lualatex over a variable since mainfont actually oversets sansfont in headings

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -57,6 +57,9 @@ $endfor$
 $if(mainfont)$
     \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$]{$mainfont$}
 $endif$
+$if(romanfont)$
+	\setsansfont[$for(romanfontoptions)$$romanfontoptions$$sep$,$endfor$]{$romanfont$}
+$endif$
 $if(sansfont)$
     \setsansfont[$for(sansfontoptions)$$sansfontoptions$$sep$,$endfor$]{$sansfont$}
 $endif$
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -58,7 +58,7 @@ $if(mainfont)$
     \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$]{$mainfont$}
 $endif$
 $if(romanfont)$
-	\setsansfont[$for(romanfontoptions)$$romanfontoptions$$sep$,$endfor$]{$romanfont$}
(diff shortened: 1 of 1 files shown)


jgm:
MANUAL.txt should also be updated (search for `mainfont` and `mainfontoptions`).

adunning:
@OvidiusCicero Would you mind providing an example of how you're using this? I can't reproduce the situation that you describe in the original commit, and am wondering if there is a better way to solve it.

There is no documentation of `romanfont` in the [fontspec manual](http://texdoc.net/pkg/fontspec). Setting both `mainfont` and `romanfont` with the standard classes gives unexpected results, with `romanfont` overwriting `mainfont` but only after inheriting its scale, which isn't likely the desired effect.

screen shot 2019-01-11 at 16 06 10

```markdown
pandoc -o test.pdf --pdf-engine=xelatex << EOT

---
title: Font Family Test
fontsize: 12pt
mainfont: Helvetica
romanfont: Palatino
sansfont: Gill Sans
monofont: Courier
---

# Fonts

The quick fox --- \the\fontdimen6\font\relax

\normalfont

The quick fox --- \the\fontdimen6\font\relax

\rmfamily

The quick fox --- \the\fontdimen6\font\relax

\sffamily

The quick fox --- \the\fontdimen6\font\relax

\ttfamily

The quick fox --- \the\fontdimen6\font\relax

EOT
```

jgm:
@adunning if you think this was a misguided change, then (after getting clarification from @OvidiusCicero) it would be good to open a new issue or PR, so this doesn't get lost.

adunning:
You're seeing that because the `scrartcl` document class sets its headings in the sans-serif font, but by default Pandoc uses the `article` document class. You can set the document class as a variable to get the same result as your LaTeX file:

```markdown
pandoc -o libertine.pdf -V mainfont="Linux Libertine" -V sansfont="Linux Biolinum" -V documentclass="scrartcl" --pdf-engine=xelatex << EOT

# Test should be in Linux Biolinum

Text, should be in Linux Libertine

EOT
```

If that solves your problem, I hope you wouldn't have any objections to removing the `romanfont` variable?

OvidiusCicero:
Ah okay, that makes sense

I find `romanfont` to have clearer semantics in some cases but I do not object

adunning:
Thank you! This will prevent it from unexpectedly changing the font size on someone.

There's no question that `romanfont` certainly does sound better, and makes much more sense with KOMA-Script in particular. It's strange that the option is completely undocumented.

Issue #3703, opened 2017-05-28
east_asian_line_breaks Extension for Org-mode input


wenxin-wang:
The `east_asian_line_breaks` extension, which eliminates `SoftBreak` between two East Asian wide characters, currently works only for [the markdown reader](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Readers/Markdown.hs#L377). Org-mode needs the same handling.

Here's a simple example
```org
* Example
你
好
```

I don't know haskell, but from the code it seems to me that `softBreakFilter` is rather generic and could be adopted by the Org reader.

The pandoc version I'm using is
```
pandoc 1.19.2.1
Compiled with pandoc-types 1.17.0.5, texmath 0.9.4, skylighting 0.1.1.5
```

Thank you for making pandoc!

jgm:
Indeed, this would be simple to add.
We could, with a bit of replumbing, even make the extension work for
all the readers.

wenxin-wang:
Here's a [python filter(https://gist.github.com/wenxin-wang/daa084356d0f96cf3fb4d38266e6f9a0) that does the same thing (at least I hope), which could serve as a temporary solution. Please correct me if there's any mistake in the code.

jgm:
Here's a thought.  Currently we do the transformation in the reader, so the AST has no SoftBreak elements between east asian characters.

This means that you can't meaningfully do `--wrap=preserve` with the `east_asian_line_breaks` option.

What if we removed the `east_asian_line_breaks` extension and instead had a fourth option for `--wrap`:  `--wrap=auto|asian|none|preserve`.

This would operate in the writer, not the reader, and would simply omit softbreak elements (not replacing with space).  (This could most conveniently be implemented by adding a filter right before the writer.)

I can think of one reason why one might want to keep this feature in the reader:  certain constructions are sensitive to spaces (e.g. _ emphasis), and we might want the reader to act as if the line breaks between two east asian characters are not spaces at all.  However, the current implementation (which is a filter AFTER parsing) does not do this.  If we wanted this, we'd need to integrate the feature more directly in the parsing.

jgm:
I made commit 774075c3e22ab2ad35e2306a5b98e30da512b310 as a first step.

jgm:
OK, I now think my thinking above was confused.  If you want to use `--wrap=preserve`, you can just leave off `east_asian_line_breaks`.  The two don't need to be made compatible.

jgm:
[...]

PR #4674, opened 2018-05-26
emphasize Ext_east_asian_line_breaks will not take effect when called as libraries

kaizhang91:
According to https://github.com/jgm/pandoc/issues/3703, `Ext_east_asian_line_breaks` will not take effect when readers/writers are called as libraries. And I think this is a nice solution for universal conversion with `Ext_east_asian_line_breaks`. However, I have spent hours to find out why my reader options don't take effect, until discover your discussion in https://github.com/jgm/pandoc/issues/3703. So I think it would be great if we add some description about the limitation.


jgm:
Thanks!

PR #4682, opened 2018-06-01
regression: path to --pdf-engine works again

### LINKED ISSUE #4681: --pdf-engine full path of the engine not working
According to the documentation:

`--pdf-engine=pdflatex|lualatex|xelatex|wkhtmltopdf|weasyprint|prince|context|pdfroff`

If the engine is not in your PATH, the full path of the engine may be specified here.
```
pandoc README.md -o out.pdf --pdf-engine=/usr/bin/xelatex
pdf-engine /usr/bin/xelatex not known
```
without full path it works:

```
pandoc README.md -o out.pdf --pdf-engine=xelatex
```
pandoc version used:
```
pandoc -v
pandoc 2.2.1
Compiled with pandoc-types 1.17.4.2, texmath 0.11, skylighting 0.7.0.2
```
It was working (with and without full path) in version 1.19.1
`/usr/bin/xelatex` exists and is executable.

mb21:
fixes #4681

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -155,9 +155,9 @@ pdfWriterAndProg mWriter mEngine = do
     where
       go Nothing Nothing       = Right ("latex", "pdflatex")
       go (Just writer) Nothing = (writer,) <$> engineForWriter writer
-      go Nothing (Just engine) = (,engine) <$> writerForEngine engine
+      go Nothing (Just engine) = (,engine) <$> writerForEngine (takeBaseName engine)
       go (Just writer) (Just engine) =
-           case find (== (baseWriterName writer, engine)) engines of
+           case find (== (baseWriterName writer, takeBaseName engine)) engines of
                 Just _  -> Right (writer, engine)
                 Nothing -> Left $ "pdf-engine " ++ engine ++
                            " is not compatible with output format " ++ writer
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/App.hs
```diff
@@ -155,7 +155,7 @@ pdfWriterAndProg mWriter mEngine = do
     where
       go Nothing Nothing       = Right ("latex", "pdflatex")
       go (Just writer) Nothing = (writer,) <$> engineForWriter writer
-      go Nothing (Just engine) = (,engine) <$> writerForEngine engine
+      go Nothing (Just engine) = (,engine) <$> writerForEngine (takeBaseName engine)
       go (Just writer) (Just engine) =
            case find (== (baseWriterName writer, engine)) engines of
```
jgm: I think we also need
```
find (== (baseWriterName writer, takeBaseName engine)) engines
```
here, right?

mb21: good point, I updated the pull...

jgm:
I also noticed that makePDF (in Text.Pandoc.PDF) pattern matches
on the engine passed in.  So, if an engine with a full path
is passed in, it won't match.  This probably needs to be fixed
too.  (You wouldn't notice it for latex, since the fallthrough
case does take the base name, but you should see an issue with
wkhtmltopdf or another program if a full path is specified.)

mb21:
Indeed, I tested only with `latex`... rebased and hopefully we've caught them all now?

jgm:
Thanks.  Some nice types would probably make this more solid, but this is good for now!

Issue #4403, opened 2018-02-25
LaTex/Beamer add hook before \begin{frame}... and after \end{frame}


NMarkgraf:
I think it would be a great idea to add the possibility to call a hook macro in latex/beamer right before and after the creation of a frame.

This would make it possible to translate something like

```
## Exercise: a new quiz {hook=exercisemode}

Just an  example ...

## No exercise 

Nothing else
```

to an latex output with might change the behavior frame title styling etc:

```
\exercisemodetrue
\begin{frame}
\frametitle{%
\protect\hypertarget{exercise-a-new-quiz}{%
 Exercise: a new quiz }}

Just an example ...
\end{frame}
\exercisemodefalse

\begin{frame}
\frametitle{%
\protect\hypertarget{o-exercise}{%
 No exercise  }}

Nothing else ...
\end{frame}

```

I need something like this desperately to change the frame title style of a slide when it shows exercises..

Thanks
Norman

cagix:
Do you really need to enclose the complete frame in the exercise mode, i.e. do you need `\exercisemodetrue` to appear **before**  `\begin{frame}`  and  `\exercisemodefalse` to appear **after** `\end{frame}`? 

If it would suffice to wrap just the **content** of your frame into the  exercise mode, you could use a simple lua-filter like

```lua
local exercise = false

function Header(h)
    local exercisemode = pandoc.Null()

    -- handle end of previous excercise mode
    if exercise then
        exercise = false
        exercisemode = pandoc.RawBlock("latex", "\\exercisemodefalse")
    end

    -- handle current header: start exercise if attribut `hook` is set to `exercisemode`
    if h.attributes["hook"] == "exercisemode" then
        exercise = true
        return { exercisemode, h, pandoc.RawBlock("latex", "\\exercisemodetrue") }
    end

    -- else return header
    return { exercisemode, h }
end
```

This filter would produce the following TeX from your example (using `--slide-level=2`):
```latex
\begin{frame}{%
\protect\hypertarget{exercise-a-new-quiz}{%
Exercise: a new quiz}}

\exercisemodetrue

Just an example \ldots{}

\exercisemodefalse

\end{frame}

\begin{frame}{%
\protect\hypertarget{no-exercise}{%
No exercise}}

Nothing else

\end{frame}
```

mb21:
> If it would suffice to wrap just the content of your frame into the exercise mode

then you could also you write raw TeX inside the markdown

NMarkgraf:
[...]

PR #4696, opened 2018-06-09
Beamer: Allow "noframenumbering" option

raymondEhlers:
As noted [here](https://tex.stackexchange.com/a/49805) ([beamer commit here](https://github.com/josephwright/beamer/commit/ff70090f36b631667b472cfe675fc3514fe46f7e)), `noframenumbering` is an undocumented, but long existing option to disable frame numbering for a particular slide. This is useful to avoid numbering backup slides (the natural approach of defining an appendix doesn't work due to the issue described [here](https://github.com/jgm/pandoc/issues/4403)). This PR makes the trivial change to support this option by adding it to the available `frameoptions`.

(I also noted the "standout" option in the manual while adding "noframenumbering")

Thanks for developing such an excellent tool!

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -4438,7 +4438,7 @@ introducing the slide:
 All of the other frame attributes described in Section 8.1 of
 the [Beamer User's Guide] may also be used: `allowdisplaybreaks`,
 `allowframebreaks`, `b`, `c`, `t`, `environment`, `label`, `plain`,
-`shrink`.
+`shrink`, `standout`, `noframenumbering`.
 
 Background in reveal.js
 -----------------------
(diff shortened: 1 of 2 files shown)


MANUAL.txt
```diff
@@ -4438,7 +4438,7 @@ introducing the slide:
 All of the other frame attributes described in Section 8.1 of
 the [Beamer User's Guide] may also be used: `allowdisplaybreaks`,
 `allowframebreaks`, `b`, `c`, `t`, `environment`, `label`, `plain`,
-`shrink`.
+`shrink`, "standout", "noframenumbering".
```
jgm: These should be surrounded by backticks `` ` `` rather than double quotes `"`, for consistency with the others.

raymondEhlers: Thanks - fixed!

jgm:
Thanks

PR #4699, opened 2018-06-12
README: Citation styles link updated

wiefling:
https://citationstyles.org/styles/ will give a 404 error. https://citationstyles.org/authors/ should be the new link with the same content.


jgm:
thanks!

Issue #4705, opened 2018-06-13
Support for Slack Formatting


omarish:
It'd be great if we supported writing to Slack. I'll often write up a few paragraphs of text in my preferred markdown editor, only to lose formatting because slack formatting differs from Markdown. Moreover, Slack does not plan to support markdown.

https://api.slack.com/docs/message-formatting

omarish:
Nice, ok. I looked through the spec and added a list of formatting types we might support.

mb21:
Yes, I had listed those already, with the exception of links/URLs which are not supported...

omarish:
RE: urls - I wonder if there's a way we could incorporate this? https://api.slack.com/docs/message-formatting#linking_to_urls

omarish:
@mb21 here's a very rough draft of the writer. The code isn't great yet since this is my first haskell project and I'm mostly pattern matching. Thoughts so far?

https://github.com/jgm/pandoc/pull/4706

mb21:
Closing this, since Slack is really not a full document format.

Use a custom writer instead, like @marish's https://github.com/omarish/pandoc-md-to-slack

PR #4706, opened 2018-06-13
[WIP] Prototype for slack writer

omarish:
https://github.com/jgm/pandoc/issues/4705

I based this on the Textile Writer, which is why there's so much extra code in there. I'll remove it.

Text formatting to support (via @mb21's suggestion):

## Features to Support

- [ ] Emphasized `_hello world_`
- [ ] Bold `*hello world*`
- [ ] Strikethrough `~hello world~`
- [ ] Ordered Lists
- [ ] Unordered Lists
- [ ] Blockquotes `>`
- [ ] Monospaced (single and triple-backtick)
- [ ] Code blocks (with language)
- [ ] URLs

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -484,6 +484,7 @@ library
                    Text.Pandoc.Writers.Math,
                    Text.Pandoc.Writers.Shared,
                    Text.Pandoc.Writers.OOXML,
+                   Text.Pandoc.Writers.Slack,
                    Text.Pandoc.Lua,
                    Text.Pandoc.PDF,
                    Text.Pandoc.UTF8,
@@ -685,6 +686,7 @@ test-suite test-pandoc
                   Tests.Writers.FB2
                   Tests.Writers.Powerpoint
                   Tests.Writers.OOXML
+                  Tests.Writers.Slack
   ghc-options:  -rtsopts -Wall -fno-warn-unused-do-bind -threaded
   default-language: Haskell2010
   other-extensions: NoImplicitPrelude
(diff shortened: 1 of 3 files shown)


mb21:
A few points:

- It seems you have mostly copied the Textile writer. That's no good, a module should contain as little code as possible, no more.
- A more comprehensive test would be to convert `test/writer.native` to your target format and check that file in.
- @jgm will still have to decide whether outputting slack is useful enough to take on the maintenance burden... it's really not much of a format since it can represent around 10% of what pandoc markdown can.

jgm:

> - @jgm will still have to decide whether outputting slack is useful enough to take on the maintenance burden... it's really not much of a format since it can represent around 10% of what pandoc markdown can.


Have you considered creating a custom lua writer?
That would be easier than a full compiled-in writer.

omarish:
@jgm good call on the custom lua writer - I'll try going in that direction and update this PR accordingly.

mb21:
@omarish the idea of custom writers is that they don't need to be merged into the code base but can be maintained separately, see http://pandoc.org/MANUAL.html#custom-writers

omarish:
@mb21 ah, got it! That's a great idea and probably the right solution, given the scope of what I'm trying to do. Thanks for the help.

omarish:
Yup, ended up implementing it as a lua filter. Thanks again for the help.

micthiesen:
@omarish have a link to your lua filter? I'm looking to make something similar

omarish:
@micthiesen sure thing -- here it is: https://github.com/omarish/pandoc-md-to-slack

micthiesen:
Thanks 🙂

PR #4708, opened 2018-06-14
AsciiDoc Writer: escape square brackets at start of line

### LINKED ISSUE #4545: square brackets at the start of the line not escaped in asciidoc output
When content is enclosed in the input in square brackets spanning the whole line, asciidoc will treat it as a an option command for the following block since it is not escaped (using asciidoctor as renderer).

````
Test 1

[my text]

Test 2
````
will render as

> Test 1
> 
> Test 2

The issue is that the line disappears completely from the rendered output, making it difficult to spot this kind of conversion errors.

There seem to be a few ways to fix this (unfortunately, `\[my text]` does not work and will render the backslash):
````
{empty}[my text]
&#91;my text]
pass:[[my text]]
````
See also [here](https://stackoverflow.com/questions/40972082/literal-asterisk-in-the-beginning-of-the-line-in-asciidoc-to-html-conversion). The {empty} variant seems to be the safest one.

There are also a few other cases decribed in the [asciidoctor reference](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#block-id-role-and-options). Strangely enough, the single-line variants can be escaped with a backslash.

**Edit:** There's also passthroughs, which also seem to work well:
````
++[++my text]
+[my text]+
````

mb21:
closes #4545

### DIFF
--- a/src/Text/Pandoc/Writers/AsciiDoc.hs
+++ b/src/Text/Pandoc/Writers/AsciiDoc.hs
@@ -44,7 +44,7 @@ import Data.Aeson (Result (..), Value (String), fromJSON, toJSON)
 import Data.Char (isPunctuation, isSpace)
 import Data.List (intercalate, intersperse, stripPrefix)
 import qualified Data.Map as M
-import Data.Maybe (fromMaybe, isJust)
+import Data.Maybe (fromMaybe, isJust, listToMaybe)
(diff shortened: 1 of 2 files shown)


mb21:
Updated and good to merge, I think.

jgm:
Thanks!

PR #4723, opened 2018-06-21
TikiWiki reader: improve list parsing

### LINKED ISSUE #4722: Tikiwiki reader should parse lists that have no space after list marker
when evaluating tikwiki, an unordered list is not evaluated.

example:

*list
**will not
**evaluate

mb21:
- remove trailing Space from list items
- also parse lists that have no space after list marker (fixes #4722)

@rlpowell what do you think?

### DIFF
--- a/src/Text/Pandoc/Readers/TikiWiki.hs
+++ b/src/Text/Pandoc/Readers/TikiWiki.hs
@@ -22,6 +22,7 @@ import Prelude
 import Control.Monad
 import Control.Monad.Except (throwError)
 import qualified Data.Foldable as F
+import Data.List (dropWhileEnd)
 import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 import qualified Data.Text as T
@@ -319,7 +320,7 @@ listItem = choice [
 bulletItem :: PandocMonad m => TikiWikiParser m (ListNesting, B.Blocks)
 bulletItem = try $ do
   prefix <- many1 $ char '*'
-  many1 $ char ' '
+  many $ char ' '
   content <- listItemLine (length prefix)
   return (LN Bullet (length prefix), B.plain content)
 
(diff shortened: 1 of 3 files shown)


rlpowell:
Assuming that test/command/4722.md is run as an automated test, this
looks excellent to me, thank you.

src/Text/Pandoc/Readers/TikiWiki.hs
```diff
@@ -344,9 +344,13 @@ listItemLine nest = lineContent >>= parseContent
       return $ filterSpaces content ++ "\n" ++ Data.Maybe.fromMaybe "" continuation
     filterSpaces = reverse . dropWhile (== ' ') . reverse
     listContinuation = string (replicate nest '+') >> lineContent
+    removeSpace x = if B.toList x == [Space]
+                    then []
+                    else [x]
     parseContent x = do
       parsed <- parseFromString (many1 inline) x
-      return $ mconcat parsed
+      let ins = init parsed ++ removeSpace (last parsed)
```
jgm: Generally I think we should avoid using partial functions like `init` and `last` whenever possible.  These crash with a runtime error when the list is empty.  Even if you know the list won't be empty when you write this code, sometimes a later change will cause the assumption to be violated, so it's good to code defensively.

mb21:
@jgm, right... removed partial functions...

src/Text/Pandoc/Readers/TikiWiki.hs
```diff
@@ -344,9 +344,14 @@ listItemLine nest = lineContent >>= parseContent
       return $ filterSpaces content ++ "\n" ++ Data.Maybe.fromMaybe "" continuation
     filterSpaces = reverse . dropWhile (== ' ') . reverse
     listContinuation = string (replicate nest '+') >> lineContent
+    removeTrailingSpace []     = []
```
jgm: Could `dropWhileEnd` from Data.List be used instead?

src/Text/Pandoc/Readers/TikiWiki.hs
```diff
@@ -344,9 +344,14 @@ listItemLine nest = lineContent >>= parseContent
...
+    removeTrailingSpace (x:[]) = if x == B.space
```
jgm: Assuming this isn't replaced with `dropWhileEnd`, this seems unnecessarily wordy.  Why not
```
removeTrailingSpace [B.space] = ...
```

mb21:
Yes! `dropWhileEnd` was what I was looking for... thanks.

PR #4733, opened 2018-06-25
CommonMark reader: Handle ascii_identifiers extension

wanders:


jgm:
Thanks for the PR!  This looks good.
My only request is to rename the test file; `gfm.txt` is not descriptive enough.
The best thing would be to create an issue that this PR closes, and name the test file using the number of that issue.

wanders:
Thanks for your input! Branch has been updated according to your comments (I think)

jgm:
Great, thanks for this!

Issue #4743, opened 2018-06-29
No easy way to use color font for emojis in html output


wanders:
The gfm input handles `:slightly_smiling_face:` style emojis by converting them to the corresponding unicode code point. This however makes it hard to use a separate font for them in html output (e.g a special color emoji font such as https://www.google.com/get/noto/help/emoji/ ).

If they were wrapped in `<span>` tags with a special class it would be trivial.

mb21:
Not saying the `span` tag is a bad idea, but I guess you could also use CSS fallback fonts for [specific unicode ranges](https://24ways.org/2011/creating-custom-font-stacks-with-unicode-range/) (or [character positions](https://css-tricks.com/css-basics-fallback-font-stacks-robust-web-typography/#article-header-id-0)).

wanders:
@mb21 Interesting. I wasn't aware that was possible. Can that be mad to work work with combining characters? As an example the `:zero:` emoji (:zero:) which is an ordinary ascii zero followed by U+FE0F and U+20E3. Having a separate font for the standard ascii zero codepoint will obviously cause other problems.

mb21:
@wanders I don't know, you'd have to try it out and see how different browsers handle it I guess...

jgm:
Instead of modifying the HTML writer, as suggested in the PR, and adding a new option (which I like to avoid whenever possible), what about just having the markdown reader put all emojis in pandoc Span elements?  e.g.
```
Span ("",["emoji"],[("title",":smile:")]) [Str "\128516"]
```
This would have the desired result in HTML output, and in most other formats the span would be ignored.  But in textual formats (like Markdown) we could emit the ASCII fallback when the `--ascii` option is specified.  (Currently emojis break the `--ascii` guarantee.)

Some things I'd like comment on:

- I'm not sure the `title` attribute is the right place to put the fallback.  Alternatively one could use `data-emoji`, or use an additional class prefixed with `emoji-` (though this is harder to extract programmatically).
- I'm not sure if the fallback value should be `:smile:` or `smile`.
- Maybe there are drawbacks I haven't considered to always including the span?

wanders:
@jgm Thanks, I'll try to hack it up when I get and chance so it can be tried out.

wanders:
@jgm Find it implemented here: https://github.com/wanders/pandoc/commit/1f384637c5150b0d75876cf324e0c730dc3e96f7

I think this feels much cleaner than doing it in the html writer.
[...]

PR #4759, opened 2018-07-07
Wrap emojis in span nodes

wanders:
This is the alternative implementation as discussed in issue #4743 that instead does the wrapping unconditionally on the reader side  (#4744 is the pull request that did it in html writer)


jgm:
Great.  One request.  I'd rather have the Emoji module continue to export the original `emojis`, as well as the new `emojiToInline`, since removing an exported symbol is a breaking API change.

wanders:
@jgm Ah, of course, didn't realize it could be used as a module by externals.  Branch updated.

jgm:
Great, thanks.

Issue #4750, opened 2018-06-30
RST output: Generate simple-style tables if possible


kaushalmodi:
See [the original request on Google Groups][1] for details.

Here's  my original message copy/pasted from that thread:

---

I am trying to use Org mode to document few Nim (https://nim-lang.org) packages. Nim has an internal HTML documentation generator that parses ReST markup docs and doc-strings.

So I am doing:

```
pandoc foo.org -o foo.rst
```

to export Org docs to ReST.

This works mostly fine except that the Nim document generator doesn't parse the grid-style ReST tables generated by Pandoc.

Example:

This style of ReST tables is parsed fine by the Nim doc generator:

```
  =======  =========
  Option   Meaning
  -------  ---------
  ``+``    All numbers (including positive ones) are preceded by a sign.
  ``-``    Only negative numbers are preceded by a sign.
  *SPACE*  Negative numbers are preceded by a sign, positive numbers are preceded by a space.
  =======  =========
```

But this style (grid-style) is not:

```
    +-----------------------------------+-----------------------------------+
    | Option                            | Meaning                           |
    +===================================+===================================+
    | ``+``                             | All numbers (including positive   |
    |                                   | ones) are preceded by a sign.     |
    +-----------------------------------+-----------------------------------+
    | ``-``                             | Only negative numbers are         |
    |                                   | preceded by a sign.               |
    +-----------------------------------+-----------------------------------+
    | *SPACE*                           | Negative numbers are preceded by  |
    |                                   | a sign, positive numbers are      |
    |                                   | preceded by a space.              |
    +-----------------------------------+-----------------------------------+
```

Here is the Org source:

```
  | Option    | Meaning                                                                              |
  |-----------+--------------------------------------------------------------------------------------|
  | =+=       | All numbers (including positive ones) are preceded by a sign.                        |
  | =-=       | Only negative numbers are preceded by a sign.                                        |
  | /SPACE/   | Negative numbers are preceded by a sign, positive numbers are preceded by a space.   |
```

Is there a way to choose the former style of ReST table when exporting from Org?

If not, would you please add this feature?

Thanks!

---

To which, @jgm replied:

> In principle, we could change the RST writer so it produces the simple tables when the contents of the table are simple enough (no block structure, no overly long lines). 

[1]: https://groups.google.com/forum/m/#!msg/pandoc-discuss/GDzFYzr_Od4/ojMGuMy0CAAJ

kaushalmodi:
[...]

PR #4780, opened 2018-07-18
rST writer: add option for list tables, closes #4564

### LINKED ISSUE #4564: Option to use list-tables in rST writer.

danse:
I moved some code from the body of `blockToRST` to a `gridTableRST` function which exposes the same interface as `listTable`.

I wrote `mapTable` and `joinTable` thinking that they could end up to the `Shared` module some day. They are straightforward but i think that using them makes the code more readable.

I also fixed the help string for the `reference-links` option.

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -417,6 +417,7 @@ convertWithOpts opts = do
           , writerSectionDivs      = optSectionDivs opts
           , writerExtensions       = writerExts
           , writerReferenceLinks   = optReferenceLinks opts
+          , writerListTables       = optListTables opts
           , writerReferenceLocation = optReferenceLocation opts
           , writerDpi              = optDpi opts
           , writerWrapText         = optWrapText opts
@@ -581,6 +582,7 @@ data Opt = Opt
     , optLogFile               :: Maybe FilePath -- ^ File to write JSON log output
     , optFailIfWarnings        :: Bool    -- ^ Fail on warnings
     , optReferenceLinks        :: Bool    -- ^ Use reference links in writing markdown, rst
+    , optListTables            :: Bool    -- ^ Use list tables in writing rst
     , optReferenceLocation     :: ReferenceLocation -- ^ location for footnotes and link references in markdown output
     , optDpi                   :: Int     -- ^ Dpi
     , optWrapText              :: WrapOption  -- ^ Options for wrapping text
@@ -653,6 +655,7 @@ defaultOpts = Opt
     , optLogFile               = Nothing
     , optFailIfWarnings        = False
     , optReferenceLinks        = False
+    , optListTables            = False
     , optReferenceLocation     = EndOfDocument
     , optDpi                   = 96
     , optWrapText              = WrapAuto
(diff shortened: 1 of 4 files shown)


danse:
This is a non-breaking change, so according to the [versioning policy](https://pvp.haskell.org/) it must affect only C in A.B.C, if i understand correctly.

When you say _carefully described_ you mean a line for the changelog and the release notes, or a paragraph somewhere in the manual?

danse:
Note that, differently from what is discussed in #4564, list tables are _not_ the default option, here. I figured it's better to introduce it as a new option for now, it will be easy to change the default behaviour later if we will want to

jgm:
Francesco Occhipinti <notifications@github.com>
writes:

> This is a non-breaking change, so according to the [versioning policy](https://pvp.haskell.org/) it must affect only C in A.B.C, if i understand correctly.

That's correct.

> When you say _carefully described_ you mean a line for the changelog and the release notes, or a paragraph somewhere in the manual?

The commit message should precisely describe the API
change and signal that it's an API change. That way I
can put this in the changelog when it's time to do a
release. The API isn't described in the manual, so
no changes are needed there.

danse:
i edited the commit message and force-pushed. while writing it occurred to me that, being an option that only involves one format, maybe it could be exposed as an extension in order not to pollute the space of global options. it could be used with `-t rst+list-tables`. i cannot assess the possibility of list tables to be adopted for other markup languages

kaushalmodi:
@danse Seeing that you are tweaking the table option in RST writer, would you have time to implement exporting of simple_table too? (https://github.com/jgm/pandoc/issues/4750)

Thanks!

danse:
hi @kaushalmodi, the simple table syntax alternative doesn't seem to add much value in terms of maintainability in comparison with other styles, at least in my opinion. Anyway i will submit the idea to my team and check whether is there any interest, thanks

link2xt:
> i cannot assess the possibility of list tables to be adopted for other markup languages

Muse format has both list and grid tables. I have not implemented grid tables yet though.

jgm:
I like the idea of making this an extension.  The main
difficult I see is that, in this case, we'd want the
extension to have different defaults for reading and
writing, and that isn't currently possible.  That is,
the default should be to properly detect list tables
when reading RST, but I don't think the default should
be to emit list tables when writing RST. Perhaps it
could be documented that the extension only affects
the writer (and only RST and maybe muse). That would
be a bit unusual, but we may already have other
extensions like that.

Francesco Occhipinti <notifications@github.com>
writes:

> i edited the commit message and force-pushed. while writing it occurred to me that, being an option that only involves one format, maybe it could be exposed as an extension in order not to pollute the space of global options. it could be used with `-t rst+list-tables`. i cannot assess the possibility of list tables to be adopted for other markup languages

danse:
so if i understand correctly, extensions are expected to affect reading and writing. is there a place where we keep a glossary? it would be great for refining these concepts. about list tables, if they are one of the allowed syntaxes, then we want them to pass through the parser anyway. i guess that this principle applies to all readers

kwshaw1:
Is this a functional solution yet? Can I get a version of Pandoc that generates list tables instead of grid tables when converting to RST?

danse:
Hi @kwshaw1, this solution worked and was merged in [this fork](https://github.com/italia/pandoc/commits/master), which hasn't been updated in a while

petterreinholdtsen:
I would love to see list-table write support in pandoc. Need it for a Norwegian national standard.  Any hope to have support for it in an official pandoc version any time soon?

krzychb:
@danse another happy pandoc user here :+1: 
Grid tables, especially when big and in big quantity are far less convenient to use than the list tables.
Adding to several good comments in https://github.com/jgm/pandoc/issues/4564, unlike for the grid tables, I like the list tables for their flexibility when it comes to text translation and editing.
Our life will be even better having the list table support in pandoc :heart: 
Thank you for this great conversion tool.

danse:
Superseeded by a [corresponding updated pull request](https://github.com/jgm/pandoc/pull/7059)

Issue #4782, opened 2018-07-19
`pandoc_title_block` extension does not work while using input format `gfm`


griffin-stewie:
## How to reproduce

```sh
pandoc sample.md -f markdown_github+pandoc_title_block -t json
```

Non-empty `meta` object is contained in json when I ran command above.

```sh
pandoc sample.md -f gfm+pandoc_title_block -t json
```
However `meta` object is empty when I ran command above.

My understanding is that `gfm` and `markdown_github` are same format.

`pandoc_title_block` extension is not  the only problem, `yaml_metadata_block` extension does not work as well.

## Environment

macOS: 10.13.6

```
$ pandoc -v
pandoc 2.2.2
Compiled with pandoc-types 1.17.5.1, texmath 0.11.0.1, skylighting 0.7.2
```

pandoc is installed by homebrew.

## Sample markdown

```md
%  'This is the title'
% Jack White; Mike Brown
% 2018/06/23

# Header Level1

## Header Level2

- AAAAA
- BBBBB
```

mb21:
> My understanding is that gfm and markdown_github are same format.

Nope. From the MANUAL:

> `markdown_github`, which provides deprecated and less accurate support for Github-Flavored Markdown; please use `gfm` instead, unless you need to use extensions other than `smart`.

Since you need extensions, please use `markdown_github` until the new `gfm` reader is implemented...

PR #4783, opened 2018-07-19
MANUAL: clarify gfm vs markdown_github

mb21:
Hopefully, this wording is clearer? There was at least [someone](https://github.com/jgm/pandoc/issues/4782) thinking `gfm` and `markdown_github` were synonyms. And a few other people asking why `gfm` doesn't support certain extensions.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -225,9 +225,9 @@ General options
     - `epub` ([EPUB])
     - `fb2` ([FictionBook2] e-book)
     - `gfm` ([GitHub-Flavored Markdown]),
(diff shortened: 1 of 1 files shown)


jgm:
Looks like an improvement to me.
One thing, though:  since the original text was added, we've added support for a few other extensions, besides `smart`, in `gfm`; `gfm_auto_identifiers`, `strikeout`, `pipe_tables`, `autolink_bare_uris`, `raw_html`, `emoji`, `hard_line_breaks`.

mb21:
True, it's corrected now.

PR #4804, opened 2018-07-31
RST writer: keep the subtitle among the other metadata, closes #4803

### LINKED ISSUE #4803: The subtitle is omitted when writing rST
In the case of rST the subtitle is rendered within the `$title` variable rather than being handled as a separate template field. Besides this, the rendering of the subtitle is fragile due to a [pattern match](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/RST.hs#L86) which causes subtitles to be discarded in the many cases when the structure of the metadata field does not match the pattern. For instance given the following template and native document, no subtitle is rendered:

```
 rotterdam (master)*$ cat templates/template.rst 
$if(title)$
$title$

$endif$
$if(subtitle)$
$subtitle$

$endif$
 rotterdam (master)*$ cat sottotitolo.native 
Pandoc (Meta {unMeta = fromList [("subtitle",MetaInlines [Str "Un",Space,Str "breve",Space,Str "sottotitolo",Space,Str "che",Space,Str "spiega",Space,Str "di",Space,Str "cosa",Space,Str "si",Space,Str "tratta"]),("title",MetaInlines [Str "Titolo",Space,Str "del",Space,Str "documento"])]})
[Para [Str "Questo",Space,Str "documento",Space,Str "di",Space,Str "test",Space,Str "serve",Space,Str "a",Space,Str "provare",Space,Str "il",Space,Str "sottotitolo."]
,Para [Str "Altro",Space,Str "paragrafo."]]
 rotterdam (master)*$ pandoc sottotitolo.native -t rst --template template --data-dir . -s
====================
Titolo del documento
====================

 rotterdam (master)*$ 
```

I found that title and subtitle are enclosed in `MetaInlines` at least when coming from a `.docx` like [sottotitolo.docx](https://github.com/jgm/pandoc/files/2245151/sottotitolo.docx), so i would at least add a pattern match for `MetaInlines` and drop the `Plain` container.

I also found confusing not to be able to use `$subtitle$` in a template, and it's important for my team to render a subtitle within an [highlight directive](http://docutils.sourceforge.net/docs/ref/rst/directives.html#highlights) rather than as a bordered title like it's currently hardcoded in the rST writer, so i would change this logic in order to have `$subtitle$` as a regular template variable.

I'd be happy to learn about reasons which might have caused the choices above and could affect the way we are planning to address this

danse:
in case of input formats like ODT or DOCX this doesn't change the
existing interface. this change will cause the subtitle not to be
rendered within the `$title$` template variable in the case of input
formats which parse the subtitle as a plain block

### DIFF
--- a/src/Text/Pandoc/Writers/RST.hs
+++ b/src/Text/Pandoc/Writers/RST.hs
@@ -82,14 +82,11 @@ pandocToRST (Pandoc meta blocks) = do
                     else Nothing
   let render' :: Doc -> Text
       render' = render colwidth
-  let subtit = case lookupMeta "subtitle" meta of
-                    Just (MetaBlocks [Plain xs]) -> xs
-                    _                            -> []
-  title <- titleToRST (docTitle meta) subtit
+  title <- titleToRST $ docTitle meta
   metadata <- metaToJSON opts
                 (fmap render' . blockListToRST)
                 (fmap (stripEnd . render') . inlineListToRST)
-                $ B.deleteMeta "title" $ B.deleteMeta "subtitle" meta
+                $ B.deleteMeta "title" meta
   body <- blockListToRST' True $ case writerTemplate opts of
                                       Just _  -> normalizeHeadings 1 blocks
                                       Nothing -> blocks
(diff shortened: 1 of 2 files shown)


jgm:
Currently pandoc implements the official rST method for indicating the subtitle, described here:
http://docutils.sourceforge.net/docs/user/rst/quickstart.html#document-title-subtitle
This PR would change that behavior, and there's no way to recover it in the template, since the subtitle needs to have the right sized border.

PR #4806, opened 2018-08-01
Fix broken link for cabal-install in INSTALL.md

ChanHoHo:
Fixed the broken link for cabal-install under the Quick cabal method section in the INSTALL.md.


jgm:
Thanks!

Issue #4801, opened 2018-07-30
Markdown reader: Semantics of space between # and header title changed after v1.19.2.4?


krlmlr:
I have read the release notes (searched for instances of "Markdown reader") but haven't found any hint. Is this intended?

```
$ echo "#header" | pandoc
<h1 id="header">header</h1>
$ echo "#header" | /usr/lib/rstudio/bin/pandoc/pandoc
<p>#header</p>

$ pandoc -v | head -n 1
pandoc 1.19.2.4
$ /usr/lib/rstudio/bin/pandoc/pandoc -v | head -n 1
pandoc 2.2.1
```

agusmba:
Probably.
Space after # is standard in commonmark, and in pandoc's markdown (if I'm reading the [manual](http://pandoc.org/MANUAL.html#extension-space_in_atx_header) correctly).

If you want the old behavior try with

```
$ echo "#header" | pandoc -f markdown-space_in_atx_header
<h1 id="header">header</h1>
```

krlmlr:
Thanks! I missed that, sorry for the noise.

krlmlr:
Thanks. How do I submit a PR to the `releases.html` page? Apart from typos, I think this is really a behavioral change (too), because the default has changed.

jgm:
It's jgm/pandoc-website
But you should simultaneously check changelog in jgm/pandoc

PR #4808, opened 2018-08-02
Describe required space after hash as breaking change

krlmlr:
Follow-up to #4801. Happy to update the website too once we agree on the wording.

### DIFF
--- a/changelog
+++ b/changelog
@@ -2610,6 +2610,11 @@ pandoc (2.0)
       (since the two-space version of a line break doesn't work inside ATX
       headers, and neither version works inside Setext headers).
 
+    + A space between the opening # and the header is now required for
+      pandoc and GitHub markdown (but not the other flavors). Disable
+      the `space_in_atx_header` extension (#3512) to restore the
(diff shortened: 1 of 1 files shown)


jgm:
The change is already described in the changelog here:
```
  * Add `space_in_atx_header` extension (#3512).  This is enabled by default
    in pandoc and GitHub markdown but not the other flavors.
    This requires a space between the opening #'s and the header
    text in ATX headers (as CommonMark does but many other implementations
    do not).  This is desirable to avoid falsely capturing things like

        #hashtag

    or

        #5
```

jgm:
I don't have objections to adding the proposed text to the changes for the markdown reader, but I wasn't sure if you'd missed the above.

krlmlr:
When I noticed the new behavior, I skimmed through the release notes, and looked more carefully at the "Changes in behavior" section for the Markdown reader. I don't know what "ATX" means and missed that item. With the new text there's some redundancy, this feature is described from two different angles. I hope that's not too bad.

PR #4815, opened 2018-08-03
Better error message on `-t pdf -o out.pdf`

### LINKED ISSUE #1155: Improve error messages about `pandoc --to pdf`
Pandoc can produce pdfs, but there is no pdf writer. Instead, in order to produce pdfs, you have to use the latex writer with an output file name that ends in `.pdf`. I understand the technical reasons for this behavior. Unfortunately, the user interface is not perfect if people accidently use a `--to pdf` flag:

```
`>echo hello | pandoc --to pdf
pandoc: Unknown writer: pdf

> echo hello | pandoc --to pdf -o hello.pdf
pandoc: cannot produce pdf output with pdf writer
```

The first error message suggests that pandoc cannot produce pdfs, and the second error message suggests that there is a pdf writer, but it is broken. Both error messages fail to help users achieve their goal, presumably to create a pdf with pandoc.

I see two options for how to improve the behavior:
1. These error messages could explain how to produce a pdf. Approximately:
   
   ```
   Unknown writer: pdf.
   To create a pdf with pandoc, use the latex writer and an output file name:
   pandoc --to latex -o filename.pdf
   ```
2. These invocations could try to produce a pdf, essentially "faking" a pdf writer. The first invocation would have to fail because there's no output file name:
   
   ```
   `>echo hello | pandoc --to pdf
   pandoc: Specify output file name for pdf with -o filename.pdf
   ```
   
   The second invocation would have the same as `pandoc --to latex -o hello.pdf`.

mb21:
closes #1155 (again)

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -142,6 +142,13 @@ engines = map ("html",) htmlEngines ++
 pdfEngines :: [String]
 pdfEngines = ordNub $ map snd engines
 
+pdfIsNoWriterErrorMsg :: String
+pdfIsNoWriterErrorMsg =
+  "To create a pdf using pandoc, use " ++
+  "-t latex|beamer|context|ms|html5" ++
+  "\nand specify an output file with " ++
+  ".pdf extension (-o filename.pdf)."
+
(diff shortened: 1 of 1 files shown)


jgm:
That seems like an improvement!

PR #4835, opened 2018-08-14
Support "toc-title" in the beamer template.

Steap:
It is a bit awkward to have a title for every frame, but not for the one
that holds the table of contents. Allow users to specify a title if they
wish.


jgm:
Good idea! Thanks.

PR #4847, opened 2018-08-21
ODT reader: better conversion of code inlines and blocks, closes #4841

### LINKED ISSUE #4841: More style names could be detected as code by the ODT reader
In order to detect code formatting in an ODT document, pandoc currently recognises the `Source_Text` value for the `style-name` attribute, but this is not the value used by the styles i have available with Libreoffice 5. I can use _Preformatted Text_ for paragraphs and _Source Text_ among the "character styles" that is inlines. They get encoded as `Preformatted_20_Text` and `Source_20_Text` values for the style names.

danse:
this pull request adds some style names to code detection and most importantly it enables the conversion of code blocks. in order to make it work i added a customised version of the `<>` operator for Inlines called `mappend'`. this is not very elegant, maybe you have some ideas about how to improve this part


jgm:
Perhaps @MarLinn would like to chime in about your question?

MarLinn:
Thank you for tackling this feature.

If I remember correctly, the function `read_plain_text` you're changing is the one producing _all_ the inlines, regardless of if they are in a code block or not. So changing it like this to preserve all double-spaces, spaces after a line break etc. should influence _all_ output, not just code blocks. I'm not calling that wrong, but it might be quite an unexpected change for some users.

The new `mappend'` doesn't look elegant indeed, but the only other way I can think of would be to use a different datatype for the inlines in code blocks – which would need its own `Monoid` instance with effectively this `mappend`, so… yeah, no gain.

So the approach looks good, but maybe `read_plain_text` should accept an `mappend` as a parameter that changes depending on if it's used inside code blocks or not? Or something along those lines.

MarLinn:
Pondering about this a bit further: I think the reader state should get a new field saying "am I in a code block" or more generally "should I preserve whitespace". `constructPara` would set that to `PreserveWhiteSpace` on encountering a code block and `read_plain_text` would honor it by choosing the appropriate `mappend`. Similarly to how the state already keeps track of how deep in a list the reader is via `currentListLevel`.

Bonus: one day, a command line parameter could set "preserve whitespace" on the top level to get as exactly a copy as possible.

danse:
i tried to apply your idea in https://github.com/jgm/pandoc/pull/4852, please let me know what you think about it

danse:
superseded by #4852

PR #4851, opened 2018-08-22
Preserve leading and trailing spaces inside links.

### LINKED ISSUE #4845: Space inside of a tag being removed
Have an interesting issue regarding spaces in links. When converting from HTML to doc. if you have a trailing space inside of the link, it strips it but does not add it outside of the link. For example:

`This is a test sentence with a link <a href="https://www.google.ca">inside </a>of it`
The sentence will convert to this: This is a test sentence with a link [inside](https://www.google.ca)of it.
The sentence should be: This is a test sentence with a link [inside](https://www.google.ca) of it.
Ran the standard pandoc config, pandoc -s input.html -o output.docx

This is happening on the latest version of Pandoc on Ubuntu 12.04

The document converts fine and works except for the above issue.

leungbk:
Closes #4845.

### DIFF
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -745,7 +745,7 @@ pLink = try $ do
   let uid = fromMaybe (T.unpack $ fromAttrib "name" tag) $
                maybeFromAttrib "id" tag
   let cls = words $ T.unpack $ fromAttrib "class" tag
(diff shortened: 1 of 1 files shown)


link2xt:
@leungbk I have actually reverted your patch (by force-pushing, so it is still displayed as merged in).

The problem with your patch is that you do not actually *extract* spaces. You preserve them, but they are left inside the link. That is inconsistent with how we handle `<em>`.

leungbk:
Oh, oops. Thanks for the feedback.

PR #4852, opened 2018-08-22
ODT reader: detect more code styles and parse code blocks, closes #4841

### LINKED ISSUE #4841: More style names could be detected as code by the ODT reader

danse:
this uses a `mappendPreserve` function provided by a new version
of pandoc-types


danse:
it requires https://github.com/jgm/pandoc-types/pull/47

MarLinn:
This looks good to me. Thanks!

danse:
Great, i guess that merging could take a while because of the change in pandoc-types

jgm:
I would much prefer a local change that doesn't touch pandoc-types. Instead of adding all this logic to preserve consecutive Space elements, why not simply have the plain text reader parse things as `Code` rather than `Str`, `Space`, etc. when "preserve whitespace" is set?  Alternatively, could try using `Str " "` instead of `Space`.  Note that `str " " <> str " " == str "  "`.

danse:
i don't get the advantage using `Str " "` rather than `Space`, since you say that both are collapsed. i changed pandoc-types assuming that `mappendPreserve` can be useful in other cases, and in some filters. for example i wrote a filter that reads elements in a div having some attributes and turns them into code ... there i would have the same problem

jgm:
`Str " "` is not collapsed.  Oddly, the markdown rendering above collapsed it!  Seems to be a bug.

jgm:
Anyway, using `Code` makes more sense than `Str`.

jgm:
I don't see how the sort of filter you're describing would be useful unless the readers generated consecutive `Space` elements, which they shouldn't.

jgm:
OK, I remember now that commonmark collapses interior spaces in code spans, and that I've planned for a long time to change that: https://talk.commonmark.org/t/leading-and-trailing-white-spaces-in-code-blocks/628/18.

danse:
>>> I don't see how the sort of filter you're describing would be useful unless the readers generated consecutive Space elements, which they shouldn't.

thanks, i hadn't considered this, clearly my filter has a big flaw. so our data model does not allow for consecutive spaces, i wonder whether the advantages are more than the disadvantages. with your last comment about Commonmark, are you suggesting that we want to find a common way to preserve spaces?

jgm:
Our data model does allow consecutive spaces in Code contexts
(Code and CodeBlock), just not in regular text.  I
think that's fine.  My suggestion was just to use Code
here instead of Str and Space, since that's designed
for literal content.

The commonmark issue is kind of separate.  Pandoc does
allow consecutive spaces in Code spans; commonmark
currently doesn't, but that will change.

Francesco Occhipinti <notifications@github.com>
writes:

danse:
Thanks, i agree with your approach and started working on such a version, if getting the types right will take me too much time i might ask for suggestions :)

danse:
regardless of how the parser is implemented, we do not want to touch pandoc-types for this so i'm closing this pull request

PR #4853, opened 2018-08-22
Add support for $toc-title$ to LaTeX (and PDF).

Wandmalfarbe:
This PR adds support for a [custom toc title](https://pandoc.org/MANUAL.html#variables-set-by-pandoc) to the default LaTeX template (and resulting PDFs). It also updates the manual for the option `toc-title`.


jgm:
looks good!

PR #4855, opened 2018-08-24
PDF: fix reference to rsvg-convert

terceiro:
When rsvg-convert is not available, pandoc would tell the user to check for rsvg2pdf instead


jgm:
That was confusing! Thanks.

PR #4858, opened 2018-08-26
Add "default" syntax highlighter language

dbader:
@jgm Thanks for your amazing work on Pandoc, it's an incredible tool!

This adds [skylighting's `default` syntax description](https://github.com/jgm/skylighting/blob/master/skylighting-core/xml/default.xml) as a syntax highlighter language. 

Currently, code blocks rendered with skylighting and "plain text" code blocks lead to different outputs (background color settings from the highlighting theme are not applied):

screenshot 2018-08-26 14 16 17

This change allows rendering code blocks containing plain text (no syntax highlighting) with the same background and text color settings used by skylighting-styled by setting their language to `default`.

About the `default` language:

> This just highlights everything as Normal Text; use it when you don't want highlighting but you do want the other formatting added by this library to be consistent with that of highlighted code. [Source](https://github.com/jgm/skylighting/commits/master/skylighting-core/xml/default.xml)

### DIFF
--- a/src/Text/Pandoc/Highlighting.hs
+++ b/src/Text/Pandoc/Highlighting.hs
@@ -149,6 +149,7 @@ langsList =
   ("command.com","command.com"),
   ("comsol","Comsol"),
   ("csh","csh"),
+  ("default","default"),
   ("delphi","Delphi"),
   ("elan","Elan"),
   ("erlang","erlang"),


jgm:
Note that this change would only affect pandoc's parsing of the LaTeX `listings` environment; the mapping you've modified correlates "listings" names with skylighting names.

Since `default` isn't a listings name, this doesn't seem right.

When pandoc is compiled against a version of skylighting including `default`, then `default` will automatically be available as a syntax name.  However, there is not yet a released version of skylighting that includes `default`; it's only in master so far....

jgm:
I will try to do a new skylighitng release soon so this will be available.

dbader:
@jgm Ah gotcha, my bad. Thanks for taking a look!

Issue #3239, opened 2016-11-17
image height in % does not retain image aspect ratio for odt output


kwijk:
### Reproducibility
When an image link is included in markdown input one can give the image width and image height. Providing only the image width should keep the image aspect ratio in the output dcument. It does when the unit is 'cm' or 'px'. It does not when unit is '%'.
 * add to a markdown document an inline reference to an image
    \[A Screenshot](screenshot.png){width=15%}
 * call pandoc myfile.md -o myfile.odt

==> The image in the odt file has its image scaled ok horizontaly, but it does not retain its ascpect ratio. Using {width=15cm} does work as expected.

### expected behaviour
When image width is given explicitly as in {width=x%} without giving the image height I expect the aspect ratio to be kept the same.

### version
pandoc 1.18

jgm:
Here's the relevant code:
``` haskell
       let dims =
             case (getDim Width, getDim Height) of
               (Just w, Just h)              -> [("width", show w), ("height", show h)]
               (Just w@(Percent _), Nothing) -> [("width", show w), ("style:rel-height", "scale")]
               (Nothing, Just h@(Percent _)) -> [("style:rel-width", "scale"), ("height", show h)]
               (Just w@(Inch i), Nothing)    -> [("width", show w), ("height", show (i / ratio) ++ "in")]
               (Nothing, Just h@(Inch i))    -> [("width", show (i * ratio) ++ "in"), ("height", show h)]
               _                             -> [("width", show ptX ++ "pt"), ("height", show ptY ++ "pt")]
             where
               ratio = ptX / ptY
               getDim dir = case (dimension dir attr) of
                              Just (Percent i) -> Just $ Percent i
                              Just dim         -> Just $ Inch $ inInch opts dim
                              Nothing          -> Nothing
```
As you can see, when a percent is used, we use `style:rel-height="scale"`.  That sounds as if it should do what you want, but apparently not?

nelas:
Hi, I just ran into this issue. Using `%` does not maintain image aspect ratio for ODT. Markdown input:

    ![Figure legend.](Fig.png){ width=100% }

ODT output:

    <draw:frame draw:name="img1" svg:width="100.0%" style:rel-height="scale">

Somehow this is not enough for ODT. It seems that it needs both `svg:width` and `svg:height` to draw the figure properly. Removing `rel-height` and adding an explicit `height %` renders the image correctly (no surprise there, though):

[...]

PR #4881, opened 2018-09-07
Fix percentage image scaling in ODT

pyssling:
Image scaling in ODT was broken when a width was set to
a percentage. The width was passed to the svg:width field
as a pecentage, which is not correct according to the ODT
standard.

Instead the real dimensions should be passed as width and
height and the style:rel-width attribute should be set to the
percentage while style:rel-heigh attribute should be set to
"scale". The converse is true if a percentage height is given.

This is now fixed and documents produced are now properly
scaled.

### DIFF
--- a/src/Text/Pandoc/Writers/ODT.hs
+++ b/src/Text/Pandoc/Writers/ODT.hs
@@ -189,8 +189,8 @@ transformPicMath opts (Image attr@(id', cls, _) lab (src,t)) = catchError
        let dims =
              case (getDim Width, getDim Height) of
                (Just w, Just h)              -> [("width", show w), ("height", show h)]
-               (Just w@(Percent p), Nothing) -> [("width", show w), ("height", show (p / ratio) ++ "%")]
-               (Nothing, Just h@(Percent p)) -> [("width", show (p * ratio) ++ "%"), ("height", show h)]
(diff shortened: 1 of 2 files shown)


pyssling:
A small comment just: There don't appear to be unit tests for the ODT writer. I'm not competent enough in haskell to write one, but if somebody does I would be happy to contribute the test-case for this.

jgm:
Thanks!  It's a bit tricky creating unit tests for the binary-format writers.  If someone wanted to pursue it, one could use the current docx writer unit tests as a model (test/Test/Writer/Docx).

mb21:
@pyssling Have you tested in LibreOffice as well? See https://github.com/jgm/pandoc/issues/3239#issuecomment-283121140

pyssling:
@mb21 I have in fact only tested in LibreOffice. Works like a charm. 

I mostly read the OASIS ODT standard docs though, they are pretty clear.

pyssling:
@mb21 Ok, I had a quick look at the code and the problem is I think you always need all four attributes, the standard at least couples rel-width with rel-height, i.e. if you set rel-width to 70% and want a proportional image you need to set rel-height to "scale". The old implementation tried to take short-cuts from what I can see by only using two tags.

PR #4901, opened 2018-09-14
MANUAL.txt epub:type attribute

mb21:
Documents https://github.com/jgm/pandoc/commit/7318bc91ce58bb6c39e556e334f278e590439c3f.


jgm:
Good idea!

PR #4905, opened 2018-09-15
Docx writer: add MetaString case for abstract, subtitle

### LINKED ISSUE #4900: * There is no abstract outputed into docx with command option "--metadata=abstract:METADATA"?
The command option "--metadata=toc-title:toc" works fine. But there is no abstract outputed into docx with command option "--metadata=abstract:METADATA".

mb21:
fixes #4900


jgm:
I wonder if there's a recurring pattern there which we could abstract out into Writers.Shared?

mb21:
Good point, see https://github.com/jgm/pandoc/pull/4907

PR #4907, opened 2018-09-16
use lookupMeta* functions wherever possible

mb21:
~~depends on https://github.com/jgm/pandoc-types/pull/51~~

### DIFF
--- a/src/Text/Pandoc/Readers/Org/DocumentTree.hs
+++ b/src/Text/Pandoc/Readers/Org/DocumentTree.hs
@@ -43,7 +43,6 @@ import Text.Pandoc.Readers.Org.BlockStarts
 import Text.Pandoc.Readers.Org.ParserState
 import Text.Pandoc.Readers.Org.Parsing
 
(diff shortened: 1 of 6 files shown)


mb21:
Okay, this is now changed to include the new functions in `Text.Pandoc.Writers.Shared` instead of pandoc-types.

mb21:
@jgm, due to https://github.com/jgm/pandoc/commit/d7263a7e5fb1ba53465099c6f1be7a85e05564f8, there was a conflict again. You added `metaValueToString` there, analogous to the `metaValueToInlines` functions which I removed in favour of the `lookupMetaInlines`. (I don't really feel strongly either way, but most usage makes this variant simpler.)

So I rebased once more. Would be really good to merge this, since otherwise people will continue adding stuff that transforms the metadata inconsistently to string/inline/block/bool, etc. like just happened now.

jgm:
@mb21, sorry about that.  I'll merge!

mb21:
Thanks!

PR #4911, opened 2018-09-17
MANUAL: add anchors to option descriptions

bpj:
This is useful for linking to option descriptions in the manual, e.g. on
the mailing list, Stack Exchange etc.

The anchors are implemented as a span wrapping each option signature
line with and id attribute of the form `#option-<option-name>` where
`<option-name>` is the *first* long option name on the line minus the
leading double dash, e.g.

    [`-t` *FORMAT*, `-w` *FORMAT*, `--to=`*FORMAT*, `--write=`*FORMAT*]{#option-to}

I must confess that I didn't put a lot of work into the edit itself; I
just ran the following Perl substitution command on the *Options*
section:

    s/^((?=`--?\w[-\w]*).*?`--(\w[-\w]+).*)/[$1]{#option-$2}/

Due to the layout of the *Options* section this should do it correctly.

I have looked through the diff and it seems like everything is as
expected.

I have also checked that there are no duplicate matches against the
regex `#option-[-\w]+`.

I haven't added any links to these anchors from elsewhere in the manual
at this time, but plan to do so if and when this PR is accepted.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -214,7 +214,7 @@ Options
 General options
 ---------------
 
-`-f` *FORMAT*, `-r` *FORMAT*, `--from=`*FORMAT*, `--read=`*FORMAT*
+[`-f` *FORMAT*, `-r` *FORMAT*, `--from=`*FORMAT*, `--read=`*FORMAT*]{#option-from}
 
 :   Specify input format.  *FORMAT* can be:
 
@@ -258,7 +258,7 @@ General options
     their names.  See `--list-input-formats` and `--list-extensions`,
     below.
 
-`-t` *FORMAT*, `-w` *FORMAT*, `--to=`*FORMAT*, `--write=`*FORMAT*
+[`-t` *FORMAT*, `-w` *FORMAT*, `--to=`*FORMAT*, `--write=`*FORMAT*]{#option-to}
 
 :   Specify output format.  *FORMAT* can be:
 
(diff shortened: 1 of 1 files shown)


jgm:
I think it is a good idea to have these anchors, but I think I'd rather generate them automatically using a lua filter when the website is built, rather than cluttering up the source.

bpj:
@jgm I've written an attempt at such a filter [here][].

It assumes that option definition lists are wrapped in divs with a certain class. While this smakes it necessary to add some stuff to the document I think it is safest so that definition lists outside the options section aren't modified inadvertently, since the only way to selectively decide which definition list terms to modify is by pattern matching.

[here]: https://gist.github.com/bpj/5513fc3fc4e33fc6330b693bcd9e4198

jgm:
Great, I've added a modified version of your filter to
the pandoc website tools. It works:

https://pandoc.org/MANUAL.html#option--output

I made a few changes. Using divs doesn't play well
with automatic tables of contents.  I had it check
for definition lists immediately under sections with
class .options.

I also make the ids include double --, with the
thought that these are less likely to collide with anything.

Benct Philip Jonsson <notifications@github.com>
writes:

bpj:
Just out of curiosity, how do you make the filter see what comes after a heading?

jgm:
Ack, looks like I didn't push my changes to the
pandoc-website repository, or I'd link you to the
script.  But the basic trick is to iterate through the
outer-level blocks and set an `inoption` variable
to true or false depending on the last header.

Benct Philip Jonsson <notifications@github.com>
writes:

bpj:
Ah, the traditional approach!  I thought I had missed some global variable or something! :-)

Anyway I'll close this now.

jgm:
I've pushed the script to pandoc-website, if you want
to have a look.


Benct Philip Jonsson <notifications@github.com>
writes:

Issue #4912, opened 2018-09-17
Convert from reStructuredText: unable to recognize embedded URIs


ivangalkin:
Hi All,

pandoc (tested versions 2.2.3.2 and 2.3) is unable to parse embedded URIs of RST input files in some cases. For perfectly fine RST input (see the [official reference](http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets)) and the valid hyperlinks like...

```reStructuredText
`Forwarding references
<http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4164.pdf>`_
```

... pandoc prints a warning... 

```
[WARNING] Reference not found for 'forwarding references <http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4164.pdf>' at line 16 column 70
```

.. and produces a wrong output. E.g. HTML

```HTML
<a href="">Forwarding references &lt;http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4164.pdf&gt;</a>
```

How to reproduce:

I observed this error as I tried to convert the clang-tidy documentation from reStructuredText to HTML in the batch manner (clang online documentation is generated by means of sphinx).

1. cd /tmp && svn co http://llvm.org/svn/llvm-project/clang-tools-extra/trunk
2. cd /tmp/trunk/docs/clang-tidy/checks
3. create and execute the following bash script
```BASH
FILES=*.rst
for f in $FILES
do
  filename="${f%.*}"
  (>&2 echo "Converting $f to $filename.html")
  `pandoc $f -f rst -t html -o $filename.html`
done
```

4. this will produce the following warnings

Alternatively you just can call pandoc manually for each single file from the list below. All affected files are also available under http://llvm.org/svn/llvm-project/clang-tools-extra/trunk/docs/clang-tidy/checks/ .

Thank you!

```
Converting abseil-no-namespace.rst to abseil-no-namespace.html
[WARNING] Reference not found for 'the full abseil compatibility guidelines <https:// abseil.io/about/compatibility>' at line 21 column 33
Converting android-comparison-in-temp-failure-retry.rst to android-comparison-in-temp-failure-retry.html
[WARNING] Reference not found for 'a convenience macro <https://www.gnu.org/software/libc/manual/html_node/interrupted-primitives.html>' at line 12 column 83
Converting bugprone-move-forwarding-reference.rst to bugprone-move-forwarding-reference.html
[...]

PR #4914, opened 2018-09-18
parse rST inlines containing newlines closing #4912

danse:
this eliminates a regression error introduced after pandoc 2.1.1,
affecting rST inline parsing. see the issue for details


danse:
i move `endline` to `inlineContent`, after `str` as it was originally

jgm:
Thanks.

PR #4921, opened 2018-09-20
ODT Writer: Improve table header row style handling

pyssling:
This changes the way styles for cells in the header row
and normal rows are handled in ODT tables.

Previously a new (but identical) style was generated for
every table, specifying the style of the cells within the table.

After this change there are two style definitions for table cells,
one for the cells in the header row, one for all other cells.

This doesn't change the actual styles, but makes post-processing
changes to the table styles much simpler as it is no longer
necessary to introduce new styles for header rows and there are
now only two styles where there was previously one per table.

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -403,8 +403,8 @@ blockToOpenDocument o bs
                       else withParagraphStyle o "Table" [Para c]
         th <- if all null h
                  then return empty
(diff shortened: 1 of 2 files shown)


jgm:
I think the CI failure was due to a warning from a change of mine to the context writer.
I've pushed a fix; if you rebase on master it will probably work this time.

pyssling:
Ok, this seems to build now. Hadn't seen that there are in fact a few small unit tests for the opendocument writer.

jgm:
Great, thanks

PR #4923, opened 2018-09-21
ConTeXt writer: Fix floats

### LINKED ISSUE #4915: ConTeXt writer: Use xtables instead of Tables #4223
Re: https://github.com/jgm/pandoc/pull/4223#issuecomment-356103589

> to always print a (possibly empty) caption, but I see no reason to break with the previous behaviour here.

My understanding is that the `location` attribute controls how the table is inserted relative to the text, although it could also control placement of the caption. (If so, this is going to be a deeper issue.)

With the hard-coded setting of `location=none`, the following output is produced:

location-none

Clearly the intent is incorrect, as what follows in the text is not a table but the following subsection. This can be changed in multiple ways. Currently I use:

```
sed -i "s/location=none/location=force/g" proposal.tex
```

This generates the expected output (widows and orphans notwithstanding):

location-force

It would be convenient if there was a way to override the value of `none` as per the [complete list of options](https://wiki.contextgarden.net/Command/placefigure).

hmenke:
In #4915 it was voiced that it would be nice to fix tables and figures rather than having them float around the page.  I'm not really sure whether that is a wise idea, from a document design point of view, but I hope that people have good reasons for horrible page breaks.

Fixes #4915

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2060,6 +2060,18 @@ In the `context` output format this enables the use of [Natural Tables
 Natural tables allow more fine-grained global customization but come
 at a performance penalty compared to extreme tables.
 
+#### Extension: `nofloat` ####
+
+Some document processing system, mostly those derived from TeX, place
+tables and images in so called “inserts”.  These inserts float around
+the page to achieve optimal placement of elements in the text with
+regards to pagebreaking.  Sometimes however, this behavior can be
+undesirable and floats should be fixed at the position at which they
+appear in the source.  If you are sure that you know what you are
+doing and are certain that you want to take the risk of extremely poor
+pagebreaks, you can add this extension to the writer.  Currently only
+the ConTeXt writer is supported.
+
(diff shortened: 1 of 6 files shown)


jgm:
> I'm not really sure whether that is a wise idea, from a document design point of view

I'm not sure either, and will hold off on merging for now.

In LaTeX we went with longtable because otherwise we'd just end up with tables that ran over the edge of the page.  longtable is a non-floating environment, so this also caused tables not to float. But of course this can led to suboptimal layout in many cases.  It's a continuing issue which we should solve somehow.

hmenke:
I think a good solution would be to let it float if there is a caption (because it can then be referred to), and fix it if there is no caption.

In ConTeXt, Extreme Tables and Natural Tables can break across pages.

hmenke:
Closing this for now.  Maybe one day I have an epiphany how to do this correctly.

PR #4925, opened 2018-09-21
add variable hyperrefoptions to default latex template to pass options to the hyperref package

tolot27:
Introduces a new variable `hyperrefoptions` which can be used to pass options to the hyperref package. For instance, it allows to specify `hyperrefoptions: linktoc=all` in a YAML block to enable linked page numbers in the table of content.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -1,4 +1,4 @@
-\PassOptionsToPackage{unicode=true}{hyperref} % options for packages loaded elsewhere
+\PassOptionsToPackage{unicode=true,$for(hyperrefoptions)$$hyperrefoptions$$sep$,$endfor$}{hyperref} % options for packages loaded elsewhere
 \PassOptionsToPackage{hyphens}{url}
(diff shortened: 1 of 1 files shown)


data/templates/default.latex
```diff
@@ -1,4 +1,4 @@
-\PassOptionsToPackage{unicode=true,$for(hyperrefoptions)$$hyperrefoptions$$sep$,$endfor$}{hyperref} % options for packages loaded elsewhere
+\PassOptionsToPackage{unicode=true$if(hyperrefoptions)$,$for(hyperrefoptions)$$hyperrefoptions$$sep$,$endfor$$endif$}{hyperref} % options for packages loaded elsewhere
```
jgm: Why not
```
unicode=true$for(hyperrefoptions)$,$hyperrefoptions$$endfor$
```

tolot27: This is much cleaner and I wasn't aware of. Thx.

Issue #4716, opened 2018-06-19
[bug] when lstset{mathescape=true}, lstinline cannot work well


bubifengyun:
Watch source https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/LaTeX.hs?#L1048#L1053

when 

```latex
\lstset{mathescape=true}
```

code in two `$` should not be converted, such as 
`$x_1$` should be 

<pre>
\passthrough{\lstinline!$x_1$!}
</pre>

but in fact it is,

<pre>
\passthrough{\lstinline!$x\_1$!}
</pre>

I think, when in mathescape=true, it should not convert the strings between two `$`s. 

https://github.com/jgm/pandoc/issues/4713

https://stackoverflow.com/questions/50898498/how-to-insert-in-lstinline/50919878#50919878

bubifengyun:
Sorry, I don't know hs language, I don't know how to edit source directly.

bubifengyun:
I had pointed the source code, pandoc version 2.2.1. in the source code, it forget to   process when mathescape = true. so every special string would be converted. but string between $ and $, as math express should not be converted. I will make a small bug code. sorry.

mb21:
I still don't quite understand, since you provide a lot of information, but the example is neither minimal (best would be a simple `echo 'my input' | pandoc -t latex` ) nor do you specify the expected output... but maybe jgm can figure out what you're getting at, since he wrote the LaTeX writer.

bubifengyun:
@jgm  I think it is simple enough, even the source code, I had already circled. 

before complete escape, take if it is mathescape, then can take convert in double $?

bubifengyun:
I upload a simple example, Thank you.

[latex_lstinline.zip](https://github.com/jgm/pandoc/files/2128152/latex_lstinline.zip)

I hope output like this.
![_ _20180622231115](https://user-images.githubusercontent.com/4437214/41784515-84017a28-7672-11e8-9280-2c981da61d77.png)

bubifengyun:
I had ref to https://github.com/dotlambda/pandoc/commit/730796ee314d42477fab216621b8e44539c94656#comments I think the person who add this feature can understand this bug. Thank you.

bubifengyun:
I had contacted with the listings package author, here is their reply.

```
Re: how about add lstget in listings package
Hoffmann, Jobst于2018年9月21日 星期五 下午21:01发送给bubifengyun...查看2个附件
详细信息
关闭按钮
[...]

PR #4926, opened 2018-09-22
Update LaTeX.hs

bubifengyun:
Solve bug https://github.com/jgm/pandoc/issues/4716,

now, the old text can works, but it need `\usepackage{etoolbox}` for `\ifdef`.
if user want to use in mathescape=true for `listings` package, the following can be added to the template when needed.

```latex
\makeatletter
\newcommand{\ifmathescape}[2]{%
  \lst@ifmathescape #1\else #2\fi
}
\makeatother
\let\ifdisableEscapeString\ifmathescape
```

Sorry, I don't know haskwell language very well. Can it work?

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -1051,7 +1051,10 @@ inlineToLaTeX (Code (_,classes,_) str) = do
         -- (defined in the default template) so that we don't have
         -- to change the way we escape characters depending on whether
         -- the lstinline is inside another command.  See #1629:
-        return $ text $ "\\passthrough{\\lstinline" ++ listingsopt ++ [chr] ++ str' ++ [chr] ++ "}"
+        -- works when lstset{mathescape=true}, See #4716
+        let originalLstinline = "\\lstinline" ++ listingsopt ++ [chr] ++ str ++ [chr]
+        let escapeLstinline = "\\lstinline" ++ listingsopt ++ [chr] ++ str' ++ [chr]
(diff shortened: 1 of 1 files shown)


jgm:
I'd rather use `\ifmathescape` directly, rather than `\ifdisableEscapedString`.

jgm:
We can put the definition of `\ifmathescape` into the default template -- conditionally on the use of listings.  This way we don't have to check to see if it's defined.

I suspect that even when mathescape=true, some nontrivial escaping is needed.  We just need to be sensitive to math bits.  What if you have e.g. `^ $x^2$` in a listing?  The first `^` would need to be escaped, but the second wouldn't.  So this requires nontrivial parsing of the contents of the code span, to separate out the math bits.

jgm:
Finally, note that any changes would have to come with changes to the tests, so the CI passes.

Since there are too many issues with this PR, I'm going to close it.  You may open a new one if you think you can address them all.

bubifengyun:
> We can put the definition of `\ifmathescape` into the default template -- conditionally on the use of listings. This way we don't have to check to see if it's defined.
> 
> I suspect that even when mathescape=true, some nontrivial escaping is needed. We just need to be sensitive to math bits. What if you have e.g. `^ $x^2$` in a listing? The first `^` would need to be escaped, but the second wouldn't. So this requires nontrivial parsing of the contents of the code span, to separate out the math bits.

It cannot fit all cases, but now it can solve current `mathescape=true` problem. In the future, may we could solve this `nontrivial escaping` by adding another condition. 
This is my solution. 
1. We can add another definition in default template, like `\newcommand\lstinlinekeepchar{$}`. 
2. When in `mathescape=true` condition, between `\lstinlinekeepchar`, we disable escape string, else escape string.
3. When in `mathescape=false` condition, escape string, and remove `\passthrough` command.

bubifengyun:
In command, we can get whether it use lstinline.

```bash
if not, 
      now code OK. 
else, 
       ifmathescape 
             some code, 
       else
            use \$,$ check escape.

PR #4927, opened 2018-09-22
Translations/ru

tie:
I suppose that you've copy-pasted the Russian translations from babel[1] package. Actually, it has two versions — `\captionsrussian@ancient` and `\captionsrussian@modern`, the former contains translations with pre-Revolutionary (1918) orthography.

This PR replaces the ancient Russian with modern. :)

[1]: http://ctan.math.utah.edu/ctan/tex-archive/macros/latex/contrib/babel-contrib/russian/russianb.ldf


jgm:
Thanks!

PR #4938, opened 2018-09-27
Parse empty argument array in inline org-mode src blocks.

leungbk:
This allows `src_emacs-lisp[]{(+ 1 1)}` to be handled like `src_emacs-lisp{(+ 1 1)}`, which wasn't the case previously.


jgm:
Thanks! I suppose another option would be to have `enclosedByPair` (using `many`) and `enclosedByPair1` (using `many1`).  The other uses of `enclosedByPair` could then be changed to `enclosedByPair1`.  This seems a bit clearer to me than the solution here (which is certainly acceptable). @tarleb any preferences?

tarleb:
I second @jgm's suggestion of introducing an `enclosedByPair1` function. I don't feel strongly about this, so we'll merge this right away. Another PR for that change would be most welcome.

Thanks!

PR #4944, opened 2018-10-02
OpenDocument writer: Implement figure numbering in captions

pyssling:
Figure captions are now numbered 1, 2, 3, ... The format in
the caption is "Figure 1: <caption>" and so on.

This is necessary in order for libreoffice to generate an
Illustration Index (Table of Figures) for included figures.

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -411,9 +411,21 @@ blockToOpenDocument o bs
       figure attr caption source title | null caption =
         withParagraphStyle o "Figure" [Para [Image attr caption (source,title)]]
                                   | otherwise    = do
+        id' <- gets stImageId
         imageDoc <- withParagraphStyle o "FigureWithCaption" [Para [Image attr caption (source,title)]]
-        captionDoc <- withParagraphStyle o "FigureCaption" [Para caption]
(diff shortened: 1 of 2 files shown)


jgm:
Instead of hard-coding the word "Figure," you can use the new translations.
See `translateTerm` and `setTranslations` in `Text.Pandoc.Class`.

pyssling:
I can look at doing it, but as far as I can see no translations have been used this far in any writers, so I really don't know how the API is intended to be used. Would you consider merging this and then adding translations? I plan to do the same numbered captions for tables and then I could use your example for that.

jgm:
Translations are used in the LaTeX reader.

Up til now we've avoided using any English words like
"Figure" in any of the writers (though LaTeX adds such
words automatically), precisely because we want to
avoid English-centrism.  But with the translation
API, we could start doing so.

If you want I could merge this and then add the needed
code (which is simple).

pyssling:
If you could I would very much appreciate it.

I haven't coded much Haskell since university in 2001 and let's say that I've gone through a steep learning curve the last few weeks.

mb21:
@pyssling see https://github.com/jgm/pandoc/commit/58ae017075cc9dbaacce85866b6fc04a02fa1ab4 and feel free to ask questions :-) (btw. I found http://learnyouahaskell.com really helpful)

PR #4949, opened 2018-10-04
OpenDocument writer: Implement table caption numbering

pyssling:
Implement table caption numbering with a format
"Table 1: <caption>".

Translations are enabled and numbering is consecutive for
captioned tables, uncaptioned tables are not enumerated.

Captioned figures are now also numbered consecutively
and uncaptioned figures are not enumerated.

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -52,7 +52,7 @@ import Text.Pandoc.Options
 import Text.Pandoc.Pretty
 import Text.Pandoc.Shared (linesToPara)
 import Text.Pandoc.Templates (renderTemplate')
-import Text.Pandoc.Translations (Term(Figure))
+import qualified Text.Pandoc.Translations as Term (Term(Figure, Table))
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -46,13 +46,13 @@ import Data.Text (Text)
...
                           setTranslations, toLang)
-import Text.Pandoc.Definition
+import Text.Pandoc.Definition as Definition
 import Text.Pandoc.Logging
 import Text.Pandoc.Options
 import Text.Pandoc.Pretty
 import Text.Pandoc.Shared (linesToPara)
 import Text.Pandoc.Templates (renderTemplate')
-import Text.Pandoc.Translations (Term(Figure))
+import Text.Pandoc.Translations as Term (Term(Figure, Table))
```
jgm: Minor stylistic preference:  I think it would be nicer to do `import qualified ... as Term` and drop the need for `Definition.Table`.  But I wouldn't insist on this.

pyssling:
All done!

PR #4950, opened 2018-10-05
Add partial support for `#+EXCLUDE_TAGS` option.

### LINKED ISSUE #4284: Org-mode reader should take into account EXCLUDE_TAGS option
The Emacs Org-mode exporter uses the `EXCLUDE_TAGS` option in the document header to customize the tag which will be used to disable the exporting of headers. This is very useful in complex setups because it allows to select which part of the document should not be exported in a more powerful way than just using `noexport`. Currently, the use of `noexport` is hard-coded in the function `isNoExportTag` at https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Readers/Org/DocumentTree.hs#L175.

It should be nice to have either a direct support of `EXCLUDE_TAGS` or at least a pandoc-specific way to disable headers.

leungbk:
Closes #4284.

If one or more of the specified tags contains a non-tag character
like `+`, Org-mode will not treat that as a valid tag, but will
nonetheless continue scanning for valid tags. That behavior is not
replicated in this patch; entering `cat+dog` as one of the entries in
`#+EXCLUDE_TAGS` and running the file through Pandoc will cause the
parser to fail and result in the only excluded tag being the default, `noexport`.
I couldn't think of a particularly clean way of handling this particular edge case.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/DocumentTree.hs
+++ b/src/Text/Pandoc/Readers/Org/DocumentTree.hs
@@ -43,6 +43,7 @@ import Text.Pandoc.Readers.Org.BlockStarts
 import Text.Pandoc.Readers.Org.ParserState
 import Text.Pandoc.Readers.Org.Parsing
 
+import qualified Data.Set as Set
 import qualified Text.Pandoc.Builder as B
 
 --
@@ -72,9 +73,6 @@ documentTree blocks inline = do
       , headlineContents = initialBlocks'
       , headlineChildren = headlines'
       }
-
-newtype Tag = Tag { fromTag :: String }
-  deriving (Show, Eq)
 
 -- | Create a tag containing the given string.
 toTag :: String -> Tag
@@ -153,7 +151,7 @@ headline blocks inline lvl = try $ do
 
    headerTags :: Monad m => OrgParser m [Tag]
    headerTags = try $
-     let tag = many1 (alphaNum <|> oneOf "@%#_") <* char ':'
+     let tag = orgTagWord <* char ':'
      in map toTag <$> (skipSpaces *> char ':' *> many1 tag <* skipSpaces)
 
 -- | Convert an Org mode headline (i.e. a document tree) into pandoc's Blocks
@@ -163,15 +161,17 @@ headlineToBlocks hdln = do
   let tags = headlineTags hdln
   let text = headlineText hdln
   let level = headlineLevel hdln
+  shouldNotExport <- hasDoNotExportTag tags
   case () of
-    _ | any isNoExportTag tags -> return mempty
+    _ | shouldNotExport -> return mempty
     _ | any isArchiveTag  tags -> archivedHeadlineToBlocks hdln
     _ | isCommentTitle text    -> return mempty
     _ | maxLevel <= level      -> headlineToHeaderWithList hdln
     _ | otherwise              -> headlineToHeaderWithContents hdln
 
-isNoExportTag :: Tag -> Bool
-isNoExportTag = (== toTag "noexport")
+hasDoNotExportTag :: Monad m => [Tag] -> OrgParser m Bool
+hasDoNotExportTag tags = containsExcludedTag . orgStateExcludedTags <$> getState
+  where containsExcludedTag s = any (`Set.member` s) tags
 
 isArchiveTag :: Tag -> Bool
 isArchiveTag = (== toTag "ARCHIVE")
--- a/src/Text/Pandoc/Readers/Org/Meta.hs
+++ b/src/Text/Pandoc/Readers/Org/Meta.hs
@@ -52,6 +52,7 @@ import Data.Char (toLower)
 import Data.List (intersperse)
 import Data.Maybe (fromMaybe)
 import qualified Data.Map as M
+import qualified Data.Set as Set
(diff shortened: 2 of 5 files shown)


src/Text/Pandoc/Readers/Org/DocumentTree.hs
```diff
@@ -163,15 +161,17 @@ headlineToBlocks hdln = do
...
     _ | any isArchiveTag  tags -> archivedHeadlineToBlocks hdln
     _ | isCommentTitle text    -> return mempty
     _ | maxLevel <= level      -> headlineToHeaderWithList hdln
     _ | otherwise              -> headlineToHeaderWithContents hdln
 
-isNoExportTag :: Tag -> Bool
-isNoExportTag = (== toTag "noexport")
+isNoExportTag :: Monad m => [Tag] -> OrgParser m Bool
+isNoExportTag tags = containsExcludedTag <$> (orgStateExcludedTags <$> getState)
+  where containsExcludedTag s = any ((flip Set.member) s) tags
```
jgm: Generally I think it's more readable to use ``any (`Set.member` s) tags``.

src/Text/Pandoc/Readers/Org/Meta.hs
```diff
@@ -190,6 +192,15 @@ parseFormat = try $ replacePlain <|> replaceUrl <|> justAppend
    rest            = manyTill anyChar         (eof <|> () <$ oneOf "\n\r")
    tillSpecifier c = manyTill (noneOf "\n\r") (try $ string ('%':c:""))
 
+excludedTagSet :: Monad m => OrgParser m (Set.Set Tag)
+excludedTagSet = do
+  skipSpaces
+  (Set.fromList <$> (map Tag <$>
```
jgm: Is this equivalent to `(Set.fromList . map Tag) <$>`?  I think that's more readable.

src/Text/Pandoc/Readers/Org/ParserState.hs
```diff
@@ -113,6 +117,7 @@ data OrgParserState = OrgParserState
                                            -- specified here.
   , orgStateEmphasisPostChars    :: [Char] -- ^ Chars allowed at after emphasis
   , orgStateEmphasisNewlines     :: Maybe Int
+  , orgStateExcludedTags          :: Set.Set Tag
```
jgm: Align `::`.

leungbk:
OK, thanks for the feedback. I've applied your suggestions. One of the Travis CI jobs seems to have failed for an unrelated reason (it's like this on the `master` branch, too).

src/Text/Pandoc/Readers/Org/Meta.hs
```diff
@@ -190,6 +192,15 @@ parseFormat = try $ replacePlain <|> replaceUrl <|> justAppend
    rest            = manyTill anyChar         (eof <|> () <$ oneOf "\n\r")
    tillSpecifier c = manyTill (noneOf "\n\r") (try $ string ('%':c:""))
 
+excludedTagSet :: Monad m => OrgParser m (Set.Set Tag)
+excludedTagSet = do
+  skipSpaces
+  ((Set.fromList . map Tag) <$>
```
tarleb: I believe this will fail if there is trailing whitespace after the last tag. A common idiom is to use `many1 (foo *> skipSpaces1) *> newline`.

src/Text/Pandoc/Readers/Org/DocumentTree.hs
```diff
@@ -163,15 +161,17 @@ headlineToBlocks hdln = do
...
+    _ | shouldNotExport -> return mempty
     _ | any isArchiveTag  tags -> archivedHeadlineToBlocks hdln
     _ | isCommentTitle text    -> return mempty
     _ | maxLevel <= level      -> headlineToHeaderWithList hdln
     _ | otherwise              -> headlineToHeaderWithContents hdln
 
-isNoExportTag :: Tag -> Bool
-isNoExportTag = (== toTag "noexport")
+hasDoNotExportTag :: Monad m => [Tag] -> OrgParser m Bool
+hasDoNotExportTag tags = (containsExcludedTag . orgStateExcludedTags) <$> getState
```
tarleb: Could we omit the braces here?

src/Text/Pandoc/Readers/Org/Meta.hs
tarleb: I believe some of the parentheses here could be omitted.

tarleb:
@link2xt could you take a look at the CI failure? It seems to be caused by a comment in the Muse reader which confuses haddock.

jgm:
Brian Leung <notifications@github.com> writes:

> OK, thanks for the feedback. I've applied your suggestions. One of the Travis CI jobs seems to have failed for an unrelated reason (it's like this on the `master` branch, too).

I've fixed that issue now.

src/Text/Pandoc/Readers/Org/Meta.hs
leungbk: Thanks. I used `many [...] skipSpaces [...]` instead since a blank input to `#+EXCLUDE_TAGS` in Emacs results in `noexport` no longer being an excluded tag.

jgm:
Great, thanks!

PR #4958, opened 2018-10-08
Add chromeos install instructions

Ewpratten:
This PR adds instructions on how to install pandoc on a chromeos device.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -45,6 +45,13 @@
 
         tlmgr install collection-fontsrecommended
 
+## Chromeos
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -45,6 +45,13 @@
 
         tlmgr install collection-fontsrecommended
 
+## Chromeos
```
jgm: Thanks!  Minor stylistic requests:

- blank line after the header
- hard-wrap lines to stay within 80 columns
- Shouldn't it be 'Chrome OS' instead of 'Chromeos'?

mb21: Maybe also link directly to `http://skycocker.github.io/chromebrew/`

Issue #4957, opened 2018-10-07
Lua custom writers can only access metadata variables in one place


Mihara:
This is not a question, but rather, a feature request.

I'm working on a custom writer for an obscure format, which requires a lot of configuration variables, tweakable per-document, normally passed through metadata, because I don't have anywhere else to put them.

The problematic part about that is that the tweakables have to define the behavior of inlines, i.e. if a particular metadata variable is true, `Str` needs to behave in a certain way, while if it's not, it needs to behave in an entirely different way.

Based on my very limited understanding of Haskell, I am more or less convinced that by the time the first Lua function is called, the AST is complete, and all the metadata variables should be known to *something*. Nevertheless, the only way to get them in the custom writer code appears to be as parameters to the `Doc` Lua function. I can't even intercept `Meta` to track the variables as they come in through the tree and save them somewhere the inlines can get at it, like the documentation says I could do with a Lua-based filter, because `Custom.hs` does not call a function with that name.

`Doc`, however, runs too late in the process, after every inline has already completed its work. This necessitates a really crude solution: Inlines leave *extra* markup in place of the markup detected, which `Doc` then has to act upon through `gsub`.

I've been up and down the pandoc internals exposed to Lua code, but couldn't find a way to acquire the metadata when it's actually needed short of trying to walk the tree myself, and it doesn't seem like `Custom.hs` will let me do that, either.

Could something be done about that?

tarleb:
If you are ok with a hacky solution involving Lua filters, consider the following approach.

As you noted, the first thing converted by the custom writer will be the first block. This is a fact we can (mis)use to signal any data that is needed by the writer. Probably the best suited is a `RawBlock` element, as it is usually rare and doesn't contain nested inlines.  We can add a writer function like this:

```lua
function RawBlock (format, str)
  print(format, str)
  if format == 'meta-hack' and str == 'yes' then
    _G.Str = function (s) return 'yes' end
    return ''
  end
  return str
end
```

This will replace the `Str` function used with one replacing all strings with just *yes*.

So now we just need to prepend a RawBlock with the information we want.  This can be done using a Lua filter.

``` lua
function Pandoc (el)
  local blocks = el.blocks
  if el.meta['meta-hack'] then
    table.insert(el.blocks, 1, pandoc.RawBlock('meta-hack', 'yes'))
  end
  return el
end
```

This will prepend a signaling RawBlock iff the metadata contains a (truthy) value of name `meta-hack`.

[...]

PR #4967, opened 2018-10-13
Custom writer: give full access to doc in optional Setup function

tarleb:
Custom writers can specify an optional `Setup` function. The function
takes the full Pandoc document as input and should not return any value.
Users can use this function to configure the writer depending on the
given document's content or its metadata.

Closes: #4957


tarleb:
I had a change of mind and replaced the *Setup* function with a global variable. I think it's less confusing and easier to use. @jgm, please revert if you disagree with that change.

PR #4972, opened 2018-10-16
Add docs about customizing pandoc

### LINKED ISSUE #3288: Add a manual (or manual section): Customizing Pandoc
Topics:

- templates
- reference.docx/odt
- filters (including documentation of the JSON serialization format and [AST definition](https://github.com/jgm/pandoc-types/blob/master/Text/Pandoc/Definition.hs#L94))
- custom lua writers
- custom syntax highlighting (syntaxes, styles)

mb21:
closes #3288

### DIFF
--- a/doc/customizing-pandoc.md
+++ b/doc/customizing-pandoc.md
@@ -1,0 +1,64 @@
+This document provides a quick overview over the various ways to
+customize pandoc's output. Follow the links to learn how to use each
+approach.
+
+[Templates](/MANUAL.html#templates)
+
+:   Pandoc comes with a template for (almost) every output format. A
+    template is a plain text file, that contains for example the line
+    `$body$`. That variable is replaced by the document's body text on
+    output.
+
+    There are many other variables, like `title`, `header-includes`,
+    etc. that are either set automatically, or that you can set using
+    [YAML metadata blocks](/MANUAL.html#extension-yaml_metadata_block),
+    [`--metadata`](/MANUAL.html#option--metadata) (which properly escape
+    things), or `--variable` (which does no escaping). You can also
+    generate your own template (e.g. `pandoc -D html > myletter.html`)
+    and customize that file, for example by introducing new variables.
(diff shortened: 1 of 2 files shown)


jgm:
Great!

jgm:
I've merged this and modified it, adding lots of TODOs.
I think it needs quite a bit more before we can put it on line, but this is a great start.

mb21:
Thanks for the quick merge!

> adding lots of TODOs

Ha, I see them. In my mind, this was more of an introductory document, to give an overview and merely linking to the other existing resources, not duplicating what we already have in the MANUAL or somewhere else. (Seems like we already have quite some duplication between `filters.html` and `lua-filters.html`...) Then again, I can definitely see the value of examples to new users...

It's also possible that the time has come to split up the MANUAL into several pages, and integrate the other files in `doc/` into that, along the lines of https://readthedocs.org. An example: https://solidity.readthedocs.io Not sure how easy it would be to make the proper redirects etc, though.

jgm:
> Ha, I see them. In my mind, this was more of an introductory document, to give an overview and merely linking to the other existing resources, not duplicating what we already have in the MANUAL or somewhere else. (Seems like we already have quite some duplication between `filters.html` and `lua-filters.html`...) Then again, I can definitely see the value of examples to new users...

I think having some examples is going to be useful for
helping people see how flexible pandoc is.  EVen
though it would duplicate things you can find
elsewhere, I see some value in having a readable
document just on the customization possibilities.

> It's also possible that the time has come to split up the MANUAL into several pages, and integrate the other files in `doc/` into that, along the lines of https://readthedocs.org. An example: https://solidity.readthedocs.io Not sure how easy it would be to make the proper redirects etc, though.

Yes, the manual has become a bit cumbersome, it's
true. On the other hand, it's handy being able to
search a single file -- and the man page is the same,
which is also nice.

It's also true that breaking links would be a big
cost.

PR #4986, opened 2018-10-20
report include is not supported

Yanpas:
There are a lot of such files in linux docs. Ideally that should be replaces with `insertIncludedFile` ...

### DIFF
--- a/src/Text/Pandoc/Readers/Man.hs
+++ b/src/Text/Pandoc/Readers/Man.hs
@@ -34,7 +34,7 @@ Conversion of man to 'Pandoc' document.
 module Text.Pandoc.Readers.Man (readMan) where
 
 import Prelude
-import Control.Monad (liftM, void, mzero)
+import Control.Monad (liftM, mzero, void, when)
 import Control.Monad.Except (throwError)
 import Data.Char (isHexDigit, chr)
 import Data.Default (Default)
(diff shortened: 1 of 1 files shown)


Yanpas:
Couple of offtopics:

There is a comment here `skipMany $ try $ string "\\\n"  -- TODO why is this here?`. here I handle hypothetical case like this
```
.MACRO\
\
arg1 arg2 etc
```
I don't remember whether I've met it, I searched all man files and didn't found anything similar, only these cases, but they are handled perfectly without this line (man pread.2):
```
.BI "ssize_t pwrite(int " fd ", const void *" buf ", size_t " count \
", off_t " offset );
```

I've also noticed that pandoc began to fail on lines like this (sigaction.2 , line 929):
```
.IR "struct sigcontext".
```
Seems like this man is incorrect, but it's quiet common to see patterns where there is one character after closing quote

jgm:
Yan Pashkovsky <notifications@github.com> writes:

> I've also noticed that pandoc began to fail on lines like this (sigaction.2 , line 929):
> ```
> .IR "struct sigcontext".
> ```
> Seems like this man is incorrect, but it's quiet common to see patterns where there is one character after closing quote

I think this is #4995.

jgm:
Yan Pashkovsky <notifications@github.com> writes:

> I've also noticed that pandoc began to fail on lines like this (sigaction.2 , line 929):
> ```
> .IR "struct sigcontext".
> ```
> Seems like this man is incorrect, but it's quiet common to see patterns where there is one character after closing quote

I've just fixed this.

PR #4991, opened 2018-10-21
EPUB writer: use metadata field `css` instead of `stylesheet`

### LINKED ISSUE #4990: Generating epub does not honour css in metadata file
I have `blue.css`:

    * { color: blue !important; }

`doc.md`:

    ---
    title: foo
    css: blue.css
    ---

Doing `pandoc -s -o test.epub doc.md` does not embed (or link to) `blue.css` from the epub. If I generate html, it gets blue indeed.

If I generate using `pandoc -s -o test.epub --css=blue.css doc.md` the my custom css does take effect, entirely replacing the default built-in epub css (https://github.com/jgm/pandoc/blob/master/data/epub.css).

This is pandoc 2.3.1.

mb21:
closes #4990

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -976,8 +976,8 @@ Options affecting specific writers {.options}
     include multiple files. They will be included in the order specified.
 
     A stylesheet is required for generating EPUB.  If none is
-    provided using this option (or the deprecated `stylesheet`
-    metadata field), pandoc will look for a file `epub.css` in the
+    provided using this option (or the `css` or `stylesheet`
+    metadata fields), pandoc will look for a file `epub.css` in the
     user data directory (see `--data-dir`).  If it is not
     found there, sensible defaults will be used.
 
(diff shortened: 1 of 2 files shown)


jgm:
For backwards compatibility, why don't we look first in `css` and then (if nothing found) `stylesheet`?

mb21:
Right, rebased with that change...

MANUAL.txt
```diff
@@ -976,8 +976,8 @@ Options affecting specific writers {.options}
     include multiple files. They will be included in the order specified.
 
     A stylesheet is required for generating EPUB.  If none is
-    provided using this option (or the `stylesheet` metadata
-    field), pandoc will look for a file `epub.css` in the
+    provided using this option (or the deprecated `stylesheet`
+    metadata field), pandoc will look for a file `epub.css` in the
```
jgm: I think this is a bit misleading:  if you use `css` in the metadata, that's not the same as using `--css` command line option.  Better would be:  "provided using this option (or the `css` or `stylesheet` metadata field)"

PR #4993, opened 2018-10-21
Org reader: clean up

leungbk:
Implement changes previously suggested by jgm and tarleb.


tarleb:
Thanks, much appreciated!

PR #5007, opened 2018-10-22
Add viz. to abbreviations file

nfleisher:
A less commonly used abbreviation, but one that is always used sentence-internally and where we want a non-breaking space after.

### DIFF
--- a/data/abbreviations
+++ b/data/abbreviations
@@ -46,5 +46,6 @@ Sept.
 Sgt.
 Sr.
 St.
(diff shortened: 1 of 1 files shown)


jgm:
I think I left `viz.` out because I thought it was pretty much always followed by a comma.  It appears, though, that this may be a matter of American usage.
https://english.stackexchange.com/questions/302709/when-should-viz-be-followed-by-a-comma

PR #5013, opened 2018-10-24
LaTeX reader: get rid of dependency on Safe module

link2xt:


link2xt:
Closed since this commit actually changes the behavior, if `query headerLevel doc'` contains only [2], the result should be 2, not 1. It is probably worth adding a test.

PR #5015, opened 2018-10-24
Lua: allow access to pandoc state

tarleb:
Lua filters and custom writers now have read-only access to most fields
of pandoc's internal state via the global variable `PANDOC_STATE`.


tarleb:
Currently missing: documentation, `__pairs` method to allow iterating over all fields.

jgm:
Looks good.  I'll wait for the documentation etc.

PR #5024, opened 2018-10-27
Cleanup Text.Pandoc.Lua module

tarleb:
Expose more useful functions via `Text.Pandoc.Lua`.

The `Text.Pandoc.Filter` is now also exposed, as importing `Text.Pandoc.App` just for its filtering features seems like overkill. This also helps with testing.


jgm:
This looks great to me.

Issue #5026, opened 2018-10-27
Roff reader fails on empty table description


Yanpas:
Code:
```
.TS
;
lb lb
l l.
Function	Description
_
```
File `linuxmans/stdio.3:183`

Yanpas:
```
.TS
lb2 lb2 lb2 lb2 lb1
lb l l l l l.
```
produces:
```
Error at "source" (line 696, column 3):
unexpected "2"
expecting "|", " ", "\t", lf new-line, ",", white space or "."
lb2 lb2 lb2 lb2 lb1
  ^
Failed to convert man7/mount_namespaces.7
```

Yanpas:
From [man-pages-posix-2013-a.tar.xz](https://mirrors.edge.kernel.org/pub/linux/docs/man-pages/man-pages-posix/)/man0p/complex.h.0p:
```
Error at "source" (line 2, column 1):
unexpected MMacro "nf" [] "source" (line 2, column 1)
MMacro "nf" [] "source" (line 2, column 1)
^
Failed to convert /mnt/d/dev/mandocset/resource/man-pages-posix-2013-a/man0p//complex.h.0p
```
Suppose it fails on line 195:
```
T{
.nf
\fIcerf\fR()
\fIcerfc\fR()
\fIcexp2\fR()
T}!T{
.nf
```

jgm:
I just fixed the empty table options bug.  The other issues should ideally have been reported separately, but we can keep them here for now.

jgm:
I usually just recompile the whole thing.  I should probably do what you advise, but of course recompiling runs the test suite and it's good to know nothing has broken.

jgm:
Perhaps open another issue about block-level content (e.g. nf) inside table cells?

jgm:
We can now get through everything in the kernel.org man pages without a parse error.

Yanpas:
> Perhaps open another issue about block-level content (e.g. nf) inside table cells?

I have tested pandoc on my archlinux on all **posix** and **linux** mans. Only one file failed:
```
Error at "input" (line 477, column 5):
unexpected "2"
expecting lf new-line, ",", white space or "."
Failed to convert man1p//stty.1p
```
From line 474
```
.TS
center tab(@) box;
cB cB | cB cB | cB cB
lf5 2 l 6 | lf5 2 l 6 | lf5 2 l.
\&^c@Value@^c@Value@^c@Value
```
[...]

PR #5027, opened 2018-10-27
man fix tableColFormat digit on the end

Yanpas:
relates to #5026 
Fixes cases like:
```
3-a/man3p//basename.3p

Error at "source" (line 105, column 3):
unexpected "5"
expecting lf new-line, ",", white space or "."
lf5 | lf5.
```

### DIFF
--- a/src/Text/Pandoc/Readers/Roff.hs
+++ b/src/Text/Pandoc/Readers/Roff.hs
@@ -433,12 +433,13 @@ tableColFormat = do
                 else return ""
       return $ x : num
     pipeSuffix' <- option False $ True <$ string "|"
+    skipMany (letter <|> digit)
     return $ CellFormat
              { columnType     = c
              , pipePrefix     = pipePrefix'
              , pipeSuffix     = pipeSuffix'
              , columnSuffixes = numsuffixes ++ suffixes }
- 
+
 -- We don't fully handle the conditional.  But we do
 -- include everything under '.ie n', which occurs commonly
 -- in man pages.  We always skip the '.el' part.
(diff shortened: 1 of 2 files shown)


Yanpas:
After 1st commit all posix files are not parsed due to nf macros. But one file failed to parse on this line:
```
Error at "source" (line 27, column 14):
unexpected "("
expecting lf new-line, ",", white space or "."
lw(15) | cf5w(15) | lw(40).
             ^
Failed to convert /mnt/d/dev/mandocset/resource/man-pages-posix-2013-a/man0p//tar.h.0p
```

src/Text/Pandoc/Readers/Roff.hs
```diff
@@ -433,12 +433,13 @@ tableColFormat = do
                 else return ""
       return $ x : num
     pipeSuffix' <- option False $ True <$ string "|"
+    skipMany (letter <|> digit)
```
jgm: I don't think we should skip these at this (tokenization) phase.

jgm: Perhaps the better solution is to move line 424 (numsuffixes) to right before 434.  That is, the numerical suffix comes after all other suffixes.  I wasn't clear from the tbl manual whether it worked this way, but it seems from your examples that maybe it does...

jgm:
So far we're only parsing "plain" tables with inline-level cell contents, so we fail on a table containing `.nf`. But I've fixed things so we get a regular parse failure (falling back to TABLE and a message about skipping the content) rather than a parse error.

jgm:
It would be easy to allow block-level content in cells, but then we'd need to have widths in addition to alignments.  Widths could be derived from the format spec or we could just set to equal widths for now.

jgm:
Looks like a digit can occur in any order.

jgm:
It was easy to fix this with these examples, thanks!

PR #5035, opened 2018-10-31
Vimwiki reader: convert parser from [Char] to Text

link2xt:


jgm:
I'd like to avoid the tostring dependency.  The package doesn't seem to be widely used (just a few packages depend on it).  The build matrix also shows failures building recent versions on ghc 8.6.1.
I think we can get by without it.

link2xt:
@jgm 
Well, there are several solutions:
1. We can implement minimal ToString just for String and Text internally.
2. We can use https://hackage.haskell.org/package/tagsoup-0.14.7/docs/Text-StringLike.html but it is not stable (see the warning).
3. We can change `PandocParsecError` and `withRaw` to use `Text`.

I am going to try to implement the third solution.

Also, `toString` is the same as to `unpack . fromString`.

jgm:
OK, now I've taken a closer look at why you need
ToString; I didn't appreciate that you were needing
to generalize withRaw.

I imagine a ToString implementation would be quite
short a simple, so maybe we could just implement
itself and put it in Shared?

Modifying withRaw to require Text is a bad plan,
unless we are changing all the parsers to use Text,
because it means we need to convert the remainder
of a String input into Text every time we use it,
and that will have bad performance implications
I think.

link2xt:
@jgm Maybe the proper way is to get rid of `withRaw` in Vimwiki reader.

link2xt:
@jgm I have updated PR to remove `tostring` dependency.

jgm:
This looks good. Actually, even if you don't need `withRaw` for this module, generalizing it in the way you suggest would be a good idea.

We should also make sure the git log clearly marks any API changes as such -- readWithM is exported, isn't it?  And ToString is also a new export.

link2xt:
@jgm I have reworded git messages to mark API changes.

As for `withRaw`, it is not related to this PR anymore and it needs further discussion. Generalizing simply with `toString` requires converting the whole input from `Text` to `String`. It is better to write some kind of `withRawText` and use specialized functions such as `Text.lines`, but I would rather avoid using this function in new parsers.

PR #5036, opened 2018-10-31
Update Links to DocBook 4 and 5

nemaniarjun:


mb21:
I seem to get a 404 on those links (after a redirect to https). Btw. not sure we want to link directly to the schema definition... as we don't for the other formats.

README.md
```diff
@@ -81,8 +81,8 @@ It can convert *to*
   - `beamer` ([LaTeX beamer](https://ctan.org/pkg/beamer) slide show)
   - `commonmark` ([CommonMark](http://commonmark.org) Markdown)
   - `context` ([ConTeXt](http://www.contextgarden.net/))
-  - `docbook` or `docbook4` ([DocBook](http://docbook.org) 4)
-  - `docbook5` (DocBook 5)
+  - `docbook` or `docbook4` ([DocBook 4](http://docbook.org/schema/4x.html))
```
wilx: `schema` wrong, `schemas` right.

jgm:
I'd prefer to link to the docbook.org top-level, which has links to the sort of documentation that tells users what docbook is and how to use it, than to the schema.  Thanks, though, for the suggestion.

PR #5041, opened 2018-11-03
Split App module

tarleb:
Extract sub-modules from App.  The App module is very large and takes a long time to compile. This can be delay re-compilations during development, as App will always be recompiled when the library is updated. Breaking-off small parts reduces compile times and can make the module more accessible.


jgm:
Good idea

jgm:
Both cabal new-build and stack build do this.

tarleb:
Oh, I wasn't aware either.

Tangentially related, what do you think about us adding a new flag which allows to choose deriving via TH vs deriving via Generics?

jgm:
Albert Krewinkel <notifications@github.com> writes:

> Oh, I wasn't aware either.
>
> Tangentially related, what do you think about us adding a new flag which allows to choose deriving via TH vs deriving via Generics?

That would be okay with me.  I'd still really like to
understand why these files are always being
recompiled, but google hasn't turned up anything.

tarleb:
I can't say for sure, but there seems to be a difference in how cabal determines the need for recomputation if a file changes. For normal files, this is based on whether a depencency's interface file had changed. This doesn't seem true for files containing TH.

Example: `echo "\n" >> src/Text/Pandoc/Highlighting.hs` causes recompilation of Text.Pandoc.Options and other TH containing modules, but not of other reverse dependencies.

tarleb:
I hope this gets answered: https://stackoverflow.com/q/53135728

tarleb:
Quoting Michael Snoyman:

> It seems like you've determined that, semantically, these two versions of the file are the same (with or without a newline). However, that's not always the case; the file-embed library would generate two different ByteStrings. There's no way for GHC, Cabal, Stack, or any other build tool I can think of to determine what is or is not a semantically meaningful change, and therefore must recompile in all cases.

jgm:
Albert Krewinkel <notifications@github.com> writes:

> Quoting Michael Snoyman:
>
>> It seems like you've determined that, semantically, these two versions of the file are the same (with or without a newline). However, that's not always the case; the file-embed library would generate two different ByteStrings. There's no way for GHC, Cabal, Stack, or any other build tool I can think of to determine what is or is not a semantically meaningful change, and therefore must recompile in all cases.

I commented on the stackoverflow sight.  Wondering
whether we can fix this problem by use
addDependentFile...

jgm:
What's weird is that I no longer see this behavior,
either with stack or cabal!

If I modify Text.Pandoc.App, then just that module
gets recompiled.

tarleb:
For me, running `echo "\n" >> src/Text/Pandoc/Highlighting.hs` will cause recompilation of T.P.Options, T.P.Filter, and T.P.App.CommandLineOptions, which all transitively depend on T.P.Highlighting. Doing the same with `src/Text/Pandoc.hs` will only cause CommandLineOptions to be recompiled. I guess this makes sense, as none of the other modules depend on Text.Pandoc.

tarleb:
Of all modules containing template haskell, CommandLineOptions seems to be the most problematic: It's relatively large and will apparently be recompiled every time we change a file that goes into Text.Pandoc.

We could extract `Opt` and its instances into a dedicated module, thus minimizing the amount of code which needs to be recompiled. It doesn't solve the problem, but it contains all TH code to small files.

jgm:
Albert Krewinkel <notifications@github.com> writes:

> We could extract `Opt` and its instances into a dedicated module, thus minimizing the amount of code which needs to be recompiled. It doesn't solve the problem, but it contains all TH code to small files.

I think this would solve the problem, wouldn't it,
because unlike T.P.App.CommandLine, the dedicated
module defining Opt wouldn't depend on other pandoc
modules, so it wouldn't get recompiled when they
change.

Seems like a good idea.

Btw, I just pushed a commit that fixes Filter so it
is sensitive to the json_via_th flag.

tarleb:
Thanks for fixing my oversight in Filter. I didn't realize that Opt had so few dependencies. That's really nice.

PR #5043, opened 2018-11-04
CommonMark writer: respect --ascii

quasicomputational:


test/command/ascii.md
```diff
@@ -51,3 +51,39 @@ pandoc -t markdown-smart --ascii
...
+^D
+hello &mldr; ok? &mldr; bye
+```
+
+```
+% pandoc -f commonmark+smart -t commonmark-smart --ascii --wrap=none
+"hi"...dog's breath---cat 5--6
+^D
+&OpenCurlyDoubleQuote;hi&rdquor;&mldr;dog&rsquor;s breath&mdash;cat 5&ndash;6
+```
```
jgm: Just noticed that we get `&OpenCurlyDoubleQuote` instead of `&ldquo`, and `&rdquor` instead of `&rdquo`.
I guess this is because the more obscure entities come later in the association list provided by tagsoup, so when we convert this to a map, the earlier entries are overwritten.  This can be fixed by writing our own list->map which either prefers entries earlier in the list, or perfers shorter entity names.

jgm: Update: the list is alphabetically ordered, so I suppose preferring shorter entity names would be the only reasonable way.

jgm: I've just pushed a change that causes shorter entity names to be preferred when there's a choice.

quasicomputational: Thanks! Yeah, I'd noticed it too, but I got a bit bogged down in the quagmire of wondering if there should be a canonical entity name for each character and how to determine that, etc, and I decided to just punt and accept whatever entity names pandoc was producing. Going with the shortest one is likely a great heuristic here.

PR #5054, opened 2018-11-07
Docx image and code block bookmarks

lierdakil:
This is prompted by https://github.com/lierdakil/pandoc-crossref/issues/147

Long story short, pandoc creates docx bookmarks for divs/spans with `id` set but fails to do so for images and code blocks. I'm pretty sure it's an oversight, so this PR adds those.

I also moved the bookmark-related code into a new utility function, since copy-pasting that code two more times didn't seem like a very good solution.


jgm:
Was an oversight - thanks!

Issue #5076, opened 2018-11-14
Pandoc 2.4 regression: (@) and odt


AugustH:
Sample:

```
text

(@) text

    text

text
```

compile with:
    pandoc -f markdown test.md -o test.odt

output changed with pandoc 2.4 (on linux mint, libreoffice 6.0.6): the numbered paragraph is indented

mb21:
Output changed from what version? Isn't the new output better?

Maybe this was this change back in 2017: https://github.com/jgm/pandoc/commit/34b9bee5a4eff0d4b76a2d9ac153105a4da363ed#diff-75ef96f1820e7e194afee0dd28277dd6 ?

AugustH:
sorry, change was from 2.3 to 2.4 -- and the output looks differs from latex/pdf and docx. And it doesn't look good either  :-( 

Example from one of my documents:

![bad](https://user-images.githubusercontent.com/11654032/48490278-5c9c9600-e825-11e8-8c3d-757cfbeb15e8.png)
 vs.

![godd](https://user-images.githubusercontent.com/11654032/48490568-febc7e00-e825-11e8-9687-fe5a02cfcaab.png)

jgm:
This probably has to do with the following changelog entry:

```
  * Text.Pandoc.Writers.OpenDocument: Improve bullet/numbering
    alignment (#4385). This change eliminates the large gap we used
    to have between bullet and text, and also ensures that numbers
    in numbered lists will be right-aligned.
```

See #4385 and commit f82d574d14c9ad71e76b8dfcbadac89ab4d10501.
Needs more fine tuning, it appears.

jgm:
```
% diff test2221/content.xml test24/content.xml 
9c9,11
<         <style:list-level-properties text:space-before="0.0in" text:min-label-width="0.4in" />
---
>         <style:list-level-properties text:list-level-position-and-space-mode="label-alignment" fo:text-align="right">
>           <style:list-level-label-alignment text:label-followed-by="listtab" text:list-tab-stop-position="0.5in" fo:text-indent="-0.1in" fo:margin-left="0.5in" />
>         </style:list-level-properties>
```
where 2221 is pandoc 2.2.2.1, 24 is pandoc 2.4.

jgm:
Fixed by #5095, I believe.

AugustH:
just checked with 2.5: as bad as 2.4 -- should the fix the commit be in pandoc 2.5 ?

jgm:
[...]

Issue #4385, opened 2018-02-21
.odt bullets misplaced horizontally


kuba-orlik:
Using: 

```
# echo "* element 1
* element 2" | pandoc --from=markdown --to=odt > odt.odt
```
Gives the following result:

![selection_085](https://user-images.githubusercontent.com/2697916/36501074-71426bd0-1746-11e8-8a56-ce9d1f904b04.png)

^ The bullet is too far to the left, leaving an awkard space.

The default spacing in Libreoffice is much nicer, and gives more ergonomic results:

![selection_088](https://user-images.githubusercontent.com/2697916/36501080-747dca7e-1746-11e8-9049-e284bb888d39.png)

I'm not saying that pandoc should make everything look same as Libreoffice does by default, but that in this particular case bringing the bullet closer to the text of the list element kind of makes sense.

The style difference basically boils down to "Indent" and "Width of numbering" style properties for list elements. In Libreoffice, it's 0.25''/ 0.25'': 

![selection_087](https://user-images.githubusercontent.com/2697916/36501177-b58b6080-1746-11e8-9a3a-844d5fb9d521.png)

Whereas the odt output from pandoc gives 0.00''/ 0.40''

My pandoc version:

```
pandoc 2.0.6
Compiled with pandoc-types 1.17.3.1, texmath 0.10.1.1, skylighting 0.5.0.1
```

kuba-orlik:
To make the case for readability - compare the usability of these two outputs:

![selection_089](https://user-images.githubusercontent.com/2697916/36501537-aa57b60e-1747-11e8-9c3d-609193f3c521.png)

![selection_090](https://user-images.githubusercontent.com/2697916/36501542-acf370a6-1747-11e8-8058-62f3c0f4fa2c.png)

Jmuccigr:
If I make an odt file with your input and then add my own list after it, here's what they look like in the content.xml file from the odt (after xml tidy):

```xml
      <text:list xml:id="list2755162617" text:style-name="L1">
        <text:list-item>
          <text:p text:style-name="P1">one</text:p>
        </text:list-item>
        <text:list-item>
          <text:p text:style-name="P1">two</text:p>
        </text:list-item>
      </text:list>
      <text:list xml:id="list971299886" text:style-name="L2">
        <text:list-item>
          <text:p text:style-name="P3">One</text:p>
        </text:list-item>
        <text:list-item>
          <text:p text:style-name="P3">two</text:p>
        </text:list-item>
[...]

PR #5095, opened 2018-11-21
ODT writer: Fix list indentation

pyssling:
Previously lists were indented by half an inch on the first line
for each level of nesting. This resulted in lists that looked like
this:
```
1.      The first line of the list point text
the second line of the same list point.
```

Fix this and bring style into line with libreoffice standards:
```
    1.  The first line of the list point text
        the second line of the list point text.
```


mb21:
Does this intend to resolve https://github.com/jgm/pandoc/issues/5076 ?

If it looks good for #4385 as well now, could you also update the failing tests?

pyssling:
I guess I discovered this independently. I'll fix the failing tests.

jgm:
Thanks!

PR #5102, opened 2018-11-26
simplify INSTALL.md

mb21:
There have been quite a few users asking questions, that were simply using the outdated Linux version from their package manager.

I've also seen people new to the commandline getting confused with the zips, and whether they should move the binary in the zip to the macOS `Applications` folder. That's why I switched the order of the `zip` and `brew` install methods.

We could also consider adding some JavaScript that constructs a direct link to the right installer, depending on the User Agent String of the browser.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,5 +1,16 @@
 # Installing pandoc
 
+The simplest way to get the latest pandoc release is to use the graphical installer. Look out for the following name for your platform:
+
+- Windows: `*-windows-x86_64.msi` (unless you're still on 32-bit),
+- macOS: `*-macOS.pkg`
+- Linux: `*-amd64.deb`
+
+**[Download the latest installer][download page]**
+
+Alternative ways to install pandoc, and additional information
+for each platform, is listed below.
+
 ## Windows
 
   - There is a package installer at pandoc's [download page].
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -1,5 +1,16 @@
 # Installing pandoc
 
+The simplest way to get the latest pandoc release is to use the graphical installer. Look out for the following name for your platform:
```
jgm: Calling it a graphical installer isn't quite right - at least for the deb.  Binary installer?
Also, maybe break the line earlier; it seems rather long.

INSTALL.md
```diff
@@ -47,15 +58,15 @@
 
 ## Linux
 
-  - First, try your package manager.
-    Pandoc is in the [Debian], [Ubuntu], [Slackware],
+  - Older versions of pandoc can be installed using most package
```
jgm: It's not strictly accurate to say that the package managers have older versions.
Arch and NiXOS, for example, are usually pretty current.  (gentoo too, I imagine.)
It depends also on the rate of release of pandoc.

mb21:
I addressed the comments... you probably can squash and merge.

jgm:
Thanks. I also made the link a big blue button.

PR #5111, opened 2018-11-29
Text.Pandoc.Process: update pipeProcess

tarleb:
The implementation of `pipeProcess` was rewritten to fix sporadic
failures caused by prematurely closed pipes.


tarleb:
[IIUC](https://stackoverflow.com/questions/26692284), the error occured because the process failed and closed the pipe before Haskell was done writing the input to the file. Ignoring the resulting error fixes the problem.

Lua's `pipe` should have handled the error better; hslua currently handles exceptions poorly and should be fixed.

Issue #4592, opened 2018-04-25
some code files omit explicit copyright and licensing header


jonassmedegaard:
The following files, arguably containing enough creative work to be copyright-protected, lack explicit copyright and licensing header within the files themselves:

 benchmark/weigh-pandoc.hs
 data/bash_completion.tpl
 data/sample.lua
 src/Text/Pandoc/CSS.hs
 src/Text/Pandoc/Data.hs
 src/Text/Pandoc/Lua/Filter.hs
 ~~src/Text/Pandoc/Readers/DocBook.hs~~
 src/Text/Pandoc/Readers/Docx/Combine.hs
 src/Text/Pandoc/Readers/Docx/StyleMap.hs
 src/Text/Pandoc/Readers/Docx/Util.hs
~~src/Text/Pandoc/Readers/EPUB.hs~~
~~src/Text/Pandoc/Readers/JATS.hs~~
~~src/Text/Pandoc/Readers/OPML.hs~~
 test/Tests/*
 test/lua/test-pandoc-utils.lua
 test/test-pandoc.hs                                
 trypandoc/trypandoc.hs                             

While the file COPYRIGHT covers these files, that is less reliable than explicit statements in a comment at the top of each file, as is done in general.

Please consider adding copyright and licensing headers for above files.

PR #5112, opened 2018-11-30
Add missing copyright notices

tarleb:
This fixes most issues mentioned in #4592.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,5 +1,5 @@
 Pandoc
-Copyright (C) 2006-2018 John MacFarlane <jgm at berkeley dot edu>
+Copyright (C) 2006-2019 John MacFarlane <jgm at berkeley dot edu>
 
 With the exceptions noted below, this code is released under the [GPL],
 version 2 or later:
@@ -37,7 +37,7 @@ The modules in the `pandoc-types` repository (Text.Pandoc.Definition,
 Text.Pandoc.Builder, Text.Pandoc.Generics, Text.Pandoc.JSON,
 Text.Pandoc.Walk) are licensed under the BSD 3-clause license:
 
-Copyright (c) 2006-2018, John MacFarlane
+Copyright (c) 2006-2019, John MacFarlane
 
 All rights reserved.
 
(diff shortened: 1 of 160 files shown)


tarleb:
Years in test files might be a bit off whenever the file was moved around such that git lost track of its origin (mostly test files).

jgm:
I'm of two minds about this.  I actually don't much like all this boilerplate in the source files, and especially in a trivial little thing like the bash completion script.  It's not strictly necessary, since we have a COPYRIGHT file covering everything, and it's one more thing that can get out of sync and need to be updated.  Maybe someone could comment about why this is important.  I do think it makes sense to consistently put these headers on the source files in src/, since we're generally doing that, but I'm more skeptical about the value of adding them in tests/ etc.

tarleb:
CC @jonassmedegaard

tarleb:
I tend to agree, these headers usually annoy me. Maybe we can just use haddock module headers, they are less verbose and contain all the necessary info. Could automatic generation of copyright statements from these headers serve as a solution?

jonassmedegaard:
I recommend to read the section "Why license notices" at https://www.gnu.org/licenses/gpl-howto.html
- and then have in mind that it is _software_ licenses which means they apply to crative works of sotware (i.e. not README files and likely don't apply to e.g. template files if they contain little if any creative changes).

jonassmedegaard:
My understanding (as a non-lawyer but passionate about semantics of licensing) is that a) copyright statements have no legal value and is added only as a courtesy for downstream users to aid their potential detective work in locating true legal origins of creative works, and that b) listing copyright notices and license statements inline (instead of or in addition to project-wide) is also not legally required.

The whole logic is backwards: Nothing is required - but if the user cannot easily locate a) who owns the rights for the works, and b) how the owner(s) license their rights, then the user is given a hard time in an eventual court case.  So if the reason you license code under copyleft is to help users enjoy the freedoms you grant them, then I strongly recommend that you follow the GNU guidelines as they are intended to limit the risk of giving your users headaches and surprises later on!

jgm:
I think the haddock module headers are sufficient to meet the goal of alerting people who come across the file of the copyright and license. Granted, they don't have the text of the GPL, but that's not hard to find.  Example of haddock module header:
```
{- |
   Module      : Text.Pandoc.Readers.CommonMark
   Copyright   : Copyright (C) 2015-2018 John MacFarlane
   License     : GNU GPL, version 2 or above

   Maintainer  : John MacFarlane <jgm@berkeley.edu>
   Stability   : alpha
   Portability : portable

Conversion of CommonMark-formatted plain text to 'Pandoc' document.

CommonMark is a strongly specified variant of Markdown: http://commonmark.org.
-}
```
Removing the GPL boilerplate that we currently have above these would streamline things a bit, I think.  I really can't see why it's necessary, given the motivations outlined in the GNU document.

jonassmedegaard:
In the end that's your call, John :-)

As I wrote, I believe *nothing* is required, really - and I agree Pandoc has already pleanty of hints of its licensing, even if not in the form preferred by the GNU project.

tarleb:
I've updated the PR to use haddock headers with copyright and licensing information instead of the usual licensing boilerplate header.

tarleb:
I was playing it safe for files under src, and for main programs. I'm interpreting your comment in that you'd prefer to use module headers in these cases as well, which is fine by me. I'll update the PR.

jgm:
Yes, I think it makes sense to get rid of this
boilerplate everywhere, except perhaps when there's
no module header.

tarleb:
Should be good now.

jgm:
Great!

PR #5115, opened 2018-11-30
Improve installation notes on associated software

adunning:
Includes explanation of how to install related tools with package managers (since the method of installing `rsvg-convert` is not obvious).


jgm:
All good changes, thanks!

adunning:
Thank you!

Issue #5044, opened 2018-11-05
LaTeX reader: support \underline{text}


mabundis:
I reported on pandoc-discuss that:
– I ran Pandoc ver 2.3.1 on a Mac OS X 10.11.6, converting LaTeX(.tex)  to MS Word (.docx).
– ANY use of the LaTeX command \underline{. . .} returned a omission/deletion of the entire '. . .' text, in the .docx output. Example:
     \underline{most-primitive}
     \underline{\textit{analytic} base} *(an embedded example)*
     both return ' '.
– I expected the { . . . } text to be returned underlined.
– In checking on jgm/pandoc on GitHub, I see that item #2264 reported that same issue, but that it was closed.
– Conversely, I was asked by John McFarlane to submit a change request:
> Pandoc doesn't have a native representation for 
> underlined text in its data model, but we could parse 
> this as a Span with class "underline", just as we do 
> with <u> in HTML.  Could you add an issue requesting 
> this to the bug tracker (jgm/pandoc on GitHub)? 

– I see the suggested solution given in #2264 is to run pandoc with '--parse - raw', but I've not tried this approach to see if it actually works.
– it would be nice to see \underline{ . . . } incorporated into standard pandoc functioning as this seems like a fairly standard need in working with text.

mb21:
This is implemented as of pandoc 2.7.3

```
$ pandoc -f latex -t native 
\underline{\textit{analytic} base}

\underline{most-primitive}
^D

[Para [Span ("",["underline"],[]) [Emph [Str "analytic"],Space,Str "base"]]
,Para [Span ("",["underline"],[]) [Str "most-primitive"]]]
```

Issue #53, opened 2011-06-10
default language for codeblocks


jgm:
see the text below for a description of the enhancement, which i posted
recently to pandoc-discuss.

a diff is attached; please feel free to use/modify/ignore it.

note that i have not updated README to mention the new option

---

i love the new syntax-highlighting of codeblocks, but the one downside
for me is having to use delimited-codeblock sytax and the language
annotation for all my code snippets, where previously the indented-
code syntax was so simple and readable.

but it was easy (what nice code!) to add a --default-codeblock-class
option to pandoc, which is then used to populate a new writer option
that specifies the classes to use for a codeblock when none have been
specified (which is the result of an indentation-style codeblock).

now all my code is magically highlighted, without changing the
markdown at all!  (and for the rare case of code that shouldn't be
highlighted, i can use the delimited syntax to give some non-assigned
"language" like "ignore".) 

Google Code Info:
Issue #: 87
Author: butt...@gmail.com
Created On: 2008-10-03T23:05:03.000Z
Closed On: 2009-12-01T06:56:47.000Z

aubertc:
Note: the name of the option that ended up being added is "`--indented-code-classes`", not "`--default-code-classes`".
This issue is discussed [here](https://github.com/jgm/pandoc/issues/2104).

PR #5135, opened 2018-12-08
Added native underline support

0C4:
Note: pandoc-types pull request #53 added a "Underline" inline constructor. 

Committing to pandoc updates that allows for reading/writing underline markup, where supported. Avoids the span-underline hack that was being used previously. Tests cases were added for underlining. All test pass using current ghc/cabal. Closes issue #5044.


0C4:
The above pull-request is twined with pantoc-types pull-request #53 since a new underline inline constructor was added. Check above will fail without the update to pandoc-types.

mb21:
Thanks @0C4, 

Unfortunately, you seem to have missed the following quote in the issue:

> Pandoc doesn't have a native representation for
underlined text in its data model, but we could parse
this as a Span with class "underline", just as we do
with in HTML.

For example:

    $ echo '<span class="underline">foo</span>' | pandoc -f html -t native
    [Plain [Span ("",["underline"],[]) [Str "foo"]]]

That's because changing the AST has the potential to break existing pandoc filters, so it's rarely done. And if it needs to be done, it's in a major release. So there is a change this pull request will be used when other AST changes are done, but this will be minimum a few months, possibly years... @jgm?

P.S. unrelated changes (like typos) should be done in separate pull requests... also saw at least one `undefined` in there, which would have to be replaced...

0C4:
Makes sense; I see looking at the project history. Applying all the above now.

mb21:
Closing this pull request... but please feel free to make a new one with the things discussed above in mind.

PR #5137, opened 2018-12-09
ICML writer: support custom-styles

mb21:
see #2106 

This turned out to be a lot easier than expected. For better or worse (this time for the better), the ICML writer implements styles simply as a list of Strings that are being passed around.


mb21:
@jgm, just wanted to make sure you had seen this as well...

jgm:
Thanks!

PR #5139, opened 2018-12-10
Implement task lists

### LINKED ISSUE #3051: Markdown checkboxes
It would be nice if Pandoc's GFM supported checkboxes, either through an extension or native to the GFM.
- [ ] example unchecked
- [x] example checked

```
 - [ ] example unchecked
 - [x] example checked
```

I hope I'm not just being thick here, I couldn't find anything about it in the manual, and no one appears to have mentioned it in the issue tracker.

mb21:
closes #3051 

rationale:

- basis for the markdown syntax is [GFM task list items spec](https://github.github.com/gfm/#task-list-items-extension-)
- checkboxes in HTML output are [disabled on GitHub](https://blog.github.com/2014-04-28-task-lists-in-all-markdown-documents/) as well

This currently does not create an extra span, but matches directly on the unicode characters. The code seems reasonably concise to me, but if you think going with a span is a better solution, I can change it of course...

~~TODO: adjust GFM reader/writer~~

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2612,6 +2612,13 @@ If default list markers are desired, use `#.`:
     #.  two
     #.  three
 
+#### Extension: `task_lists` ####
+
+Pandoc supports task lists, using the syntax of GitHub-Flavored Markdown.
+
+    - [ ] an unchecked task list item
+    - [x] checked item
+
 
 ### Definition lists ###
 
@@ -4223,7 +4230,7 @@ variants are supported:
 :   `pipe_tables`, `raw_html`, `fenced_code_blocks`, `auto_identifiers`,
     `gfm_auto_identifiers`, `backtick_code_blocks`,
     `autolink_bare_uris`, `space_in_atx_header`,
-    `intraword_underscores`, `strikeout`, `emoji`,
+    `intraword_underscores`, `strikeout`, `task_lists`, `emoji`,
     `shortcut_reference_links`, `angle_brackets_escapable`,
     `lists_without_preceding_blankline`.
 
@@ -4254,7 +4261,7 @@ Also, `raw_tex` only affects `gfm` output, not input.
 :   `pipe_tables`, `raw_html`, `fenced_code_blocks`, `auto_identifiers`,
     `gfm_auto_identifiers`, `backtick_code_blocks`,
     `autolink_bare_uris`, `space_in_atx_header`,
-    `intraword_underscores`, `strikeout`, `emoji`,
+    `intraword_underscores`, `strikeout`, `task_lists`, `emoji`,
     `shortcut_reference_links`, `angle_brackets_escapable`,
     `lists_without_preceding_blankline`.
 
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -167,6 +167,7 @@ data Extension =
     | Ext_subscript           -- ^ Subscript using ~this~ syntax
     | Ext_superscript         -- ^ Superscript using ^this^ syntax
     | Ext_styles              -- ^ Read styles that pandoc doesn't know
+    | Ext_task_lists          -- ^ Parse certain list items as task list items
     | Ext_table_captions      -- ^ Pandoc-style table captions
     | Ext_tex_math_dollars    -- ^ TeX math between $..$ or $$..$$
     | Ext_tex_math_double_backslash  -- ^ TeX math btw \\(..\\) \\[..\\]
@@ -215,6 +216,7 @@ pandocExtensions = extensionsFromList
   , Ext_strikeout
   , Ext_superscript
   , Ext_subscript
+  , Ext_task_lists
   , Ext_auto_identifiers
   , Ext_header_attributes
   , Ext_link_attributes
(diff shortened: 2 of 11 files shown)


mb21:
Fixed the code for GHC 7...

jgm:
Nicer LaTeX output would be
```latex
\begin{itemize}
\item[$\square$] Foo bar
    
    baz
\item[$\boxtimes$] ok
\end{itemize}
```
which renders like this:
screen shot 2018-12-17 at 10 00 01 am

mb21:
Yes, I've changed it to `\item[$\square$]`, followed by a newline for consistency with how we render normal lists...

jgm:
This looks good so far.  questions:

1. What about gfm/commonmark support?  It's a bit awkward that this doesn't work with the gfm reader; that is what people are most likely to use anyway with this feature.

    gfm/commonmark support should be quite easy to add.  In the clause for LIST at line 113 of the Commonmark reader, you'd just need to add another function in the pipeline, e.g. `(handleTaskList . setTightness . addBlocks opts)`.  This would be operating on a pandoc AST.

    But now, it makes sense to factor out this `handleTaskList` code and use it in the markdown reader as well, instead of the custom parsing code.  That is, parse as normal into `Str "[x]", Space, ...` and then just revise the block.

2. What about the markdown and gfm/commonmark writers?  Shouldn't they be recognizing the unicode checklist symbols at the beginning of lists (if the extension is set) and replacing with `[x] ` etc.?

3. I *think* I'm okay with leaving out the span.  This would mean that if one has a list item that starts with the modal logic box operator, it might be accidentally construed as a task list, but this would normally be in math mode.

mb21:
I left the CommonMark parts as a todo because I wasn't sure what was the best way. Thanks for the pointers, I'll take a look in the next couple of days... I guess factoring out the `handleTaskList` is worth the second pass to eliminate code duplication?

About the "modal logic box operator": we're actually not using the U+25FB ◻WHITE MEDIUM SQUARE, but instead:

U+2610  ☐  BALLOT BOX
U+2612  ☒  BALLOT BOX WITH X

jgm:

> I left the CommonMark parts as a todo because I wasn't sure what was the best way. Thanks for the pointers, I'll take a look in the next couple of days... I guess factoring out the `handleTaskList` is worth the second pass to eliminate code duplication?

The performance difference should be minimal
(unmeasurable I'd guess).  We're not doing a pass over
the whole AST, just that one list item.

mb21:
I think this is actually starting to look much better!

I wasn't sure whether `handleTaskListItem` should be `[Block] -> [Block]` or `Blocks -> Blocks`. But since we still support GHC 7, we cannot use pattern synonyms on `Seq`, so I went for the lists where we can just pattern match...

Note that I also changed the CommonMark Writer to output `Raw Inline/Block "markdown"`. Not sure whether this should also be `"commonmark", "gfm"`, or similar.

jgm:
I agree, this is looking good!

I think the names `taskListItemFromMd` and `taskListItemToMd` are a bit odd, since these are AST transformations and not conversions to and from markdown.  I wonder if `taskListItemToAscii` and `taskListItemFromAscii` or something like that would make more sense?  (Alternatively, `taskListItemFromUnicode`, `taskListItemToUnicode`.)

jgm:
One more thought.  Looking at our other extensions, I think `task_lists` would be a more consistent name for the extension than `tasklist`.  (Compare `example_lists`, `definition_lists`, etc.)

mb21:
I adopted the name from the GFM spec linked above, but if you think `task_lists` is better I can change it...

jgm:
Yes, I think we should try to be internally consistent with naming conventions of extensions.

jgm:
That is:  `task_lists` is better.

mb21:
Yes, makes sense. It's changed.

jgm:
I forgot to document this change properly in the git log message, so I made a trivial follow-up change with documentation. For future reference, API changes should always be carefully document in the commit messages, so they can be indicated in the changelog.

mb21:
Right... but good to see this merged into master!

agusmba:
Thanks for implementing this!
The default docx behavior doesn't look great though.

image

There is an issue with the spacing, and another with long lines. Both could be solved by using specific numbering styles with the proper bullets:

image

Should I open a separate issue for it? (I was reluctant since this isn't yet in a released pandoc version)

mb21:
@agusmba We currently don't do anything in the docx writer to handle task lists. This is just the unicode fallback. I added the feature mainly for HTML output. But feel free to open a new issue for the docx output, best with an example docx xml snippet of how the output should look like.

agusmba:
I'll do that. Thanks!!

alok:
Does this recognize `- []` as a task list? (no space between the square brackets)

jgm:
Alok Singh <notifications@github.com> writes:

> Does this recognize `- []` as a task list? (no space between the square brackets)

No.

alok:
@jgm Would that functionality be accepted? I'd be happy to put in the PR since it looks like a one line change. I don't think it really interferes with anything else since I don't see what else `- []` is used for.

mb21:
As mentioned above, this was modeled after GitHub tasklists. There, it needs a space as well, so I would keep it that way.

Issue #5131, opened 2018-12-07
load parskip only if not handled by document class


Konfekt:
Regarding the default inclusion of `parskip` at https://github.com/jgm/pandoc-templates/blob/7fed34ab85d9c89b7f17707f7e97b4daa75276fb/default.latex#L162 , how about making this conditional on the documentclass being different from one of the KOMA classes (in TeX, this can be tested for by `\@ifundefined{KOMAClassName}{}{...}`) so that KOMA does not warn about the incompatibilities arising when using both, KOMA and `parskip` at the same time?

According to https://tex.stackexchange.com/questions/133003/using-the-parskip-package-i-find-the-space-between-subtitles-ugly it is better to solely rely on KOMA instead.

Konfekt:
Thank you for pointing the `indent` option out. I made a PR.

mb21:
@Konfekt Thanks, but you made to pull to the wrong repo. From that repo's README:

> This repository is a subtree of pandoc; it will be updated each time there is a new pandoc release. However, changes to templates are made first in pandoc; thus, issues and pull requests should be made pandoc's issue tracker rather than here.

Konfekt:
Thank you, I hope the PR now applies to the correct repo.

Konfekt:
True, thank you, that was a self-referential pull request to my own repo. Now it is fixed: https://github.com/jgm/pandoc/pull/5143

PR #5143, opened 2018-12-10
let KOMA document class handle parskip when applicable

Konfekt:
As discussed in https://github.com/jgm/pandoc/issues/5131 it is better to let KOMA document classes handle parskip itself.

Please check the indentation of the added code, which might not be the style that should be adhered to.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -159,12 +159,13 @@ $endif$
 }{}
 $if(indent)$
 $else$
-\IfFileExists{parskip.sty}{%
-\usepackage{parskip}
-}{% else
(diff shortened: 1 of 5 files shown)


jgm:
As you can see, there are a few test golden files that need to be adjusted for the template change.

Konfekt:
All affected golden test files were adjusted and the last test passed.

jgm:
Excellent - many thanks for this!

Konfekt:
Hold on, there was an `makeatletter` missing. The new PR https://github.com/jgm/pandoc/pull/5145/ remedies that omission.

PR #5144, opened 2018-12-10
fix various links in INSTALL.md; bump required stack version

maxthomas:
odd formatting at the end of the document was making markdown
choke on link rendering; deleting that fixed the issues with all
links.

also update version: 1.6.5 produced an error;
1.7.0 and higher seems OK

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,17 +23,17 @@ under the heading for your operating system.
     a zip file that contains pandoc's binaries and
     documentation.  Simply unzip this file and move the binaries
     to a directory of your choice.
-    
+
   - Alternatively, you can install pandoc using
     [Chocolatey](https://chocolatey.org):
-    
+
         choco install pandoc
-    
+
     Chocolatey can also install other software that integrates with Pandoc.
     For example, to install `rsvg-convert` (from [librsvg], covering formats
     without SVG support), [Python] (to use Pandoc filters), and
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -393,12 +393,12 @@ To run just the markdown benchmarks:
 [NixOS]: https://nixos.org/nixos/packages.html
 [Slackware]: https://www.slackbuilds.org/result/?search=pandoc&sv=
 [Ubuntu]: https://packages.ubuntu.com/pandoc
-[download button]: https://github.com/jgm/pandoc/releases/latest {.btn .btn-block .btn-primary}
```
jgm: This is valid pandoc markdown, and it's needed to make the download link on <http://pandoc.org/installing.html> a big, visible button.  GitHub markdown doesn't understand this, though, so it won't render right on GitHub.  If we have to choose, I'd say it's more important for it to render correctly on pandoc.org.

We could replace the markdown link with some raw HTML, though I don't know if this would survive GitHub's rendering chain either...

jgm: Another option would be to change this to INSTALL.txt, so GitHub won't attempt to render it as Markdown.

maxthomas: hey there! thanks for the feedback. makes sense: i found the document via github and obviously thought the sole source of truth was living there, but had no idea it was rendered elsewhere.

happy to consider whichever other option, just thought it was a quick win (but obviously not so simple!)

PR #5148, opened 2018-12-14
Update data file abbreviations

qlx4r:
add "etc."


jgm:
Probably not a good idea to add `etc.`.

When `etc.` occurs in normal writing, it is either followed by a punctuation mark (usually a comma) or it ends a sentence.  In neither case do we need to convert a following space to a nonbreaking space. In the second case, it would be positively wrong to do that.

PR #5175, opened 2018-12-25
Remove unused HasHeaderMap

link2xt:
It is updated by some readers, but never actually used.


jgm:
Yes, I guess `stateHeaderMap` got supplanted by `stateHeaderKeys`.
Good catch.

Note that this also changes the API by removing `stateHeaderMap` from `ParserState`.

PR #5184, opened 2018-12-27
FIX #5183 : dokuwiki : remove automatic ':' prefix before internal image links

### LINKED ISSUE #5183: Dokuwiki Writer : wrong path for internal images
Hi !

Thanks for this wonderful software, we're producing thousands pages of PostgreSQL trainings with it !

Here's a report  for what looks to me like a bug in the dokuwiki writer:

the markdown syntax `` is converted into `{{:foo.png}}`

See https://pandoc.org/try/?text=!%5B%5D(images%2Ffoo.png)&from=markdown&to=dokuwiki`

The colon character (`:`) is a marker for namespaces in the dokuwiki syntax  (see  https://www.dokuwiki.org/namespaces ) and `{{:foo.png}}` means the files is located in the root namespace. This is similar to an absolute path. 

The markdown `` means the file is located in the same relative path that the document. Therefore it should be translated as `{{foo.png}}` (no colon).

The problem seems to be on lines 134 and 520 of the DokuWiki writer: 

https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/DokuWiki.hs#L134
https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/DokuWiki.hs#L520

daamien:
closes #5183

* `` should be converted to `{{foo.png}}` (relative path)
* `

Therefore the ':' prefix is useless and must be removed.
It blocks users from making relative image links.

### DIFF
--- a/src/Text/Pandoc/Writers/DokuWiki.hs
+++ b/src/Text/Pandoc/Writers/DokuWiki.hs
@@ -130,9 +130,7 @@ blockToDokuWiki opts (Para [Image attr txt (src,'f':'i':'g':':':tit)]) = do
   let opt = if null txt
                then ""
                else "|" ++ if null tit then capt else tit ++ capt
-      -- Relative links fail isURI and receive a colon
-      prefix = if isURI src then "" else ":"
-  return $ "{{" ++ prefix ++ src ++ imageDims opts attr ++ opt ++ "}}\n"
+  return $ "{{" ++ src ++ imageDims opts attr ++ opt ++ "}}\n"
 
 blockToDokuWiki opts (Para inlines) = do
   indent <- asks stIndent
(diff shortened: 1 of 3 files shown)


mb21:
You will need to change the failing tests in `/tests` etc..

daamien:
The same problem is happening in the Zimwiki Writer

https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/ZimWiki.hs#L112

Should I fix it too within this PR or do you prefer two separate PR ?

mb21:
I don't really know ZimWiki, but if you're very confident (or can point to documentation saying) that the image namespaces behave the same way as in dokuwiki, then that's probably good enough for us.

It should certainly be a separate commit. About putting it in a  separate pull request, I'm not sure, probably both is fine... @jgm?

daamien:
@mb21 : I don't know ZimWiki at all. But the doc says that most of the syntax is inspired by dokuwiki and the image syntax does not mention the ':' character....

http://zim-wiki.org/manual/Help/Wiki_Syntax.html

Most of the code of the zimwiki writer seems to be copy-pasted from the dokuwiki writer

jgm:
Thanks!

PR #5185, opened 2018-12-27
Fix typo in MANUAL.txt

gely:
The example to create a div with id `refs` instead creates a div with class `#refs`.  The fix is to add curly brackets.


mb21:
Thanks!

PR #5186, opened 2018-12-27
Replace read with safeRead

### LINKED ISSUE #5180: Remove some more reads
Replace with safeRead.

- [ ] ```
      src/Text/Pandoc/Readers/TikiWiki.hs
      505:  return $B.str [toEnum (read inner :: Int) :: Char]
      ```

- [ ] ```
      src/Text/Pandoc/Writers/Docx.hs
      237:  let pgContentWidth = (-) <$> (read <$> mbAttrSzWidth ::Maybe Integer)
      239:                         (+) <$> (read <$> mbAttrMarRight ::Maybe Integer)
      240:                         <*> (read <$> mbAttrMarLeft ::Maybe Integer)
      ```

- [ ] ```
      src/Text/Pandoc/Writers/LaTeX.hs
      558:                                       '%':ds -> showFl (read (reverse ds) / 100 :: Double)
      ```

mb21:
closes #5180


mb21:
Seems like the build failure has nothing to do with my changes...

PR #5190, opened 2018-12-30
Org reader: handle `minlevel` option differently.

### LINKED ISSUE #5188: Failure to convert included files from .org to .rst
# Problem
I am writing blogs using Emacs org-mode. Specifically, I first create contents and get `ml.org` file and then convert it into `ml.rst` using pandoc command `pandoc -f org -t rst ml.org`.

In order to modulize the contents I wrire. I use the
following code in org-mode

    * Supervised Learning
    ** Regression
    #+INCLUDE: "ml/linear.org" :minlevel 3
    ** Classification
    #+INCLUDE: "ml/logistic.org" :minlevel 3
    
    * Unsupervised Learning
    ** Clustering
    #+INCLUDE: "ml/kmeans.org" :minlevel 3
    #+INCLUDE: "ml/spectral.org" :minlevel 3
    
    ** Dimensionality Reduction
    #+INCLUDE: "ml/pca.org" :minlevel 3
    #+INCLUDE: "ml/tsne.org" :minlevel 3

When converting this file into pdf using LaTex, the structure is correct and is shown [here][1]. However, after conversion to `ml.rst` file, **none** of these files are included.

# What I Have Done

It seems that this could be resolved using template feature in pandoc, but I am not sure what to do.

Could anyone help me, thank you in advance.

  [1]: https://i.stack.imgur.com/G3x5v.png

leungbk:
When `minlevel` exceeds the original minimum level observed in the
file to be included, every heading should be shifted rightward.

Closes #5188.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/Blocks.hs
+++ b/src/Text/Pandoc/Readers/Org/Blocks.hs
@@ -547,9 +547,7 @@ include = try $ do
 
   shiftHeader :: Int -> Block -> Block
   shiftHeader shift blk =
-    if shift <= 0
-    then blk
-    else case blk of
+    case blk of
       (Header lvl attr content) -> Header (lvl - shift) attr content
       _ -> blk
 
--- a/test/Tests/Readers/Org/Directive.hs
+++ b/test/Tests/Readers/Org/Directive.hs
@@ -192,9 +192,14 @@ tests =
        headerWith ("level3", [], []) 3 "Level3")
 
     , testWithFiles [("./level3.org", "*** Level3\n\n")]
-      "Minlevel shifts level"
+      "Minlevel shifts level leftward"
       (T.unlines [ "#+include: \"level3.org\" :minlevel 1" ] =?>
        headerWith ("level3", [], []) 1 "Level3")
+
+    , testWithFiles [("./level1.org", "* Level1\n\n")]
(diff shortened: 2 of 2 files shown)


tarleb:
Thank you, this is looking good. I love the fact that deleting code seems to add more features.
I think we currently run into problems if someone uses `:minlevel 0` (which is fine for org-mode, it demotes first-level headers to top-level paragraphs). Would that be easy to add? If not, simply preventing pandoc from behaving weirdly would be enough for now, I think.

leungbk:
Ugh, I'm really sorry for the delay--I've been busier than expected and had some computer problems as well.

Yeah, the `== 0` for `lvl - shift` needs to be there, but I think the `< 0` check belongs outside, in the `blockFilter` function. Right now a negative value would never get passed into `shiftHeader`. I wasn't familiar with `safeRead`, and never got around to learning some of the relevant types.

jgm:
No worries @leungbk!  Thanks for the PR.  I just didn't want it to languish and be forgotten.

jgm:
@leungbk @tarleb have a look at commit 710a22e.
This seems to agree with the org exporter when minlevel == 0. (h1 becomes a regular paragraph, h2 becomes h1 -- that's not very useful behavior, but that's how it seems to work.)
It also guarantees we won't get header level < 1.

PR #5193, opened 2019-01-01
LaTeX Template: Improve package selections

### LINKED ISSUE #4861: footnote and xcolor do not commute
MWE:
```
---
header-includes: \usepackage{xcolor}
...
----
C^[F]
----
```

LaTeX complains of a forgotten `\endgroup`. A bit of digging [shows](https://tex.stackexchange.com/a/15836) that this is due to `footnote` defining `\fn@endnote` with a dependency on `\color@endgroup`, which is subsequently redefined by `xcolor`, breaking the macro.

This in turn breaks any document that uses footnotes and `xcolor`. Since `tikz` depends on `xcolor`, adding it to `header-includes` also triggers this problem.

A quick-and-dirty workaround until `footnote` fixes this is to include `xcolor` before `footnote`.
Subsequent inclusions will note it's already included and not run.
Note that `xcolor` is in `texlive-latex-recommended`, so this shouldn't be an issue.

P.S. I will note, however, that `footnote` was last updated in 1996, so I'm not holding my breath for it to be fixed. Haven't contacted upstream, but seeing as this is a well-known bug that's at least 7 years old, they're probably not responding.

adunning:
Changes:

- Use Babel instead of Polyglossia for LuaLaTeX: there are a number of bugs in Polyglossia under this engine with common languages, e.g. <https://github.com/reutenauer/polyglossia/issues/182>. (Most development appears to have moved from Polyglossia to Babel; it might make sense to make it the default for both LuaTeX and XeTeX before long.)
- Use [`xurl`](https://ctan.org/pkg/xurl) if available. Introduced with TeX Live 2018, this greatly improves the appearance of URLs by allowing them to break at additional points. (Run `tlmgr update --all` if you are seeing problems with this, as there were some bugs in the initial release.)
- Use [`bookmark`](https://ctan.org/pkg/bookmark) if available, which can correct the levels of headings where `hyperref` cannot: see the [KOMA-Script 3.26 release notes](https://komascript.de/release3.26).
- Load `xcolor` if available. The `xcolor` package must be loaded before the `footnote` package, which we load to fix foonotes in tables. Closes #4861.
- Remove obsolete `fixltx2e` package (has no functionality with TeX Live 2015 or later).
- Reindent file.

This test demonstrates the changes:

```shell
pandoc -o test.pdf --pdf-engine=lualatex << EOT

Some text <http://dsf.tqsdatmdtdls.ctm/cudfsdvqfqll-ocd/bdcbcfqlf-lcclldh/cbsntwgqke-esbptb-vvylhy/zkhqnqidf-obedbacpf-lzlal-pxaccqa-gwki.lfof/> and another url: <http://osb.ccdngagkkg.raa/qrkxzvi/dxfsiaa/xidf-lata-dgdqmhp-uoxdl-vst-vcsbhl-aisdsasih-skl-aezb-fhixvyy-qqlachd-achhfc-koe-xgfqp-iyplcu-1.696884>.

[Bonjour!]{lang=fr}

Test!

EOT
```

With pandoc 2.5 template:

screen shot 2019-01-01 at 18 02 21

With changes:

screen shot 2019-01-01 at 18 01 55

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -174,6 +174,7 @@ $if(verbatim-in-note)$
 \usepackage{fancyvrb}
 $endif$
 \IfFileExists{xcolor.sty}{\usepackage{xcolor}}{}
+\IfFileExists{xurl.sty}{\usepackage{xurl}}{} % add URL line breaks if available
 \usepackage{hyperref}
 \hypersetup{
 $if(title-meta)$
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -286,12 +286,12 @@ $for(header-includes)$
 $header-includes$
 $endfor$
 $if(lang)$
-\ifnum 0\ifxetex 1\fi\ifluatex 1\fi=0 % if pdftex
+\ifnum 0\ifxetex 1\fi=0 % if pdftex or luatex
   \usepackage[shorthands=off,$for(babel-otherlangs)$$babel-otherlangs$,$endfor$main=$babel-lang$]{babel}
 $if(babel-newcommands)$
   $babel-newcommands$
 $endif$
-\else
+\else % if xetex
   % load polyglossia as late as possible as it *could* call bidi if RTL lang (e.g. Hebrew or Arabic)
   \usepackage{polyglossia}
   \setmainlanguage[$polyglossia-lang.options$]{$polyglossia-lang.name$}
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -199,7 +199,7 @@ $if(verbatim-in-note)$
 $endif$
 \IfFileExists{xcolor.sty}{\usepackage{xcolor}}{}
 \IfFileExists{xurl.sty}{\usepackage{xurl}}{} % add URL line breaks if available
-\usepackage{hyperref}
+\IfFileExists{bookmark.sty}{\usepackage{bookmark}}{\usepackage{hyperref}}
 \hypersetup{
 $if(title-meta)$
   pdftitle={$title-meta$},
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -2,14 +2,40 @@
 \PassOptionsToPackage{hyphens}{url}
 $if(colorlinks)$
 \PassOptionsToPackage{dvipsnames,svgnames*,x11names*}{xcolor}
-$endif$$if(dir)$$if(latex-dir-rtl)$
+$endif$
+$if(dir)$
+$if(latex-dir-rtl)$
 \PassOptionsToPackage{RTLdocument}{bidi}
-$endif$$endif$%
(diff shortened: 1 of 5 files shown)


mb21:
> Most development appears to have moved from Polyglossia to Babel; it might make sense to make it the default for both LuaTeX and XeTeX before long

I remember polyglossia being "marketed" as a replacement to babel. And when using XeTeX, the recommendation was to use polyglossia. You're saying babel has caught up with polyglossia and/or tackled the whatever problems polyglossia was designed to solve?

jgm:
For one recent issue with polyglossia/xelatex, see
https://github.com/pandoc/lua-filters/pull/36
Special typography for French colons is applied in code blocks, and there seems to be no way to disable this short of switching to English for code blocks. I don't know if babel is any better in this respect.

data/templates/default.latex
```diff
@@ -152,51 +177,49 @@ $endif$
...
 \makeatother
 $endif$
 $if(verbatim-in-note)$
 \usepackage{fancyvrb}
 $endif$
-$if(colorlinks)$
-\usepackage{xcolor}
-$endif$
-\usepackage{hyperref}
+\IfFileExists{xcolor.sty}{\usepackage{xcolor}}{}
```
jgm: What happens if they specify `colorlinks` but `xcolor` doesn't exist?
Previously we would have gotten a latex compilation failure, pointing to the need for xcolor.
What do we get in this case now?

adunning: Good point: if there are no visible links in the document, it will work even if `colorlinks` is specified.

It fails when the document tries to render a colour using the `xcolor` definitions: for example, in a document with a URL and `colorlinks` enabled:

```latex
! LaTeX Error: Undefined color `Blue'.

 ...                                              
                                                  
l.69 \href{https://pandoc.org}{test url}
```

Is it worse, then, to require `xcolor` even in situations when it isn't needed, or to have an error message that doesn't explain the package is missing?

adunning:
That verbatim issue does appear to be a bug specific to Polyglossia. I wonder if it would be worth making Polyglossia an option similar to how `mathspec` is handled right now. (Babel's support for RTL languages is even improving; here's [a Hebrew example](https://tex.stackexchange.com/questions/434549/minimal-babel-and-lualatex-hebrew-bidi-example).) I've looked for a survey of which has broader language support at the moment, but haven't found anything.

For background, Babel is now part of the LaTeX core, supporting both XeTeX and LuaTeX, and has been improving steadily since it received a new maintainer in 2012. Meantime, the creator of Polyglossia is now inactive, and the current maintainer quite understandably has his hands full with managing the TeX hyphenation patterns. It would appear from all this that the eventual aim is make Babel a single interface for all languages and engines, though I'm not aware of a conscious effort to merge the two or any public statements on the matter.

jgm:
Andrew Dunning <notifications@github.com> writes:

> Is it worse, then, to require `xcolor` even in situations when it isn't needed, or to have an error message that doesn't explain the package is missing?

I like the basic principle that, if the latex code we
generate *requires* a package, then the package should
be unconditionally loaded in the preamble.  If the
package would merely improve output, but isn't
required, then it's okay to load it if if exists.
That will ensure that things fail in a way that gives
the user the best clue as to what is going on.

I'm not really concerned about documents with
`colorlinks: true` failing even if they contain no
links.

adunning:
That does seem best; I've simplified the line to require `xcolor`. There are few situations in which it will be missing in any case.

PR #5207, opened 2019-01-09
Change treatment of `number-lines` directives.

### LINKED ISSUE #5182: RST reader giving `startFrom` a blank value when no line number specified
The RST reader presently has

``` haskell
 let kvs = maybe [] (\n -> [("startFrom", trimr n)]) numberLines
```

where `n` can be an empty string. I would have expected the lambda here to return the alist only for non-empty  `n`,  and to return an empty list for empty `n`. The problem is that round-tripping 

```
.. code:: haskell
   :number-lines:

   main = putStrLn "Hello World!"
   unsafePerformIO main
```

from RST to native output back to RST adds an extra space after the `:number-lines:` that makes sense only for non-empty `n` in the above reader snippet; see the following snippet from the RST writer.

``` haskell
let startnum = maybe "" (\x -> " " <> text x) $ lookup "startFrom" kvs
```

Changing the reader snippet to have something like

``` haskell
\n -> let trimmed = trimr n in
                                     if trimmed == ""
                                       then []
                                       else [("startFrom", trimmed)]
```

in the lambda is what I would do, but I also noticed that the MediaWiki reader has something similar that also seems to admit a blank value corresponding to `startFrom`:

``` haskell
let kvs = maybe [] (\x -> [("startFrom",x)]) mbstart
```

An alternative way of preventing the aforementioned round-trip blank space would be to modify the RST writer, but changing the reader instead makes more sense to me. I don't know MediaWiki (or RST, for that matter) very well, so I'm not sure if there's some other reason to keep the reader snippets looking like this.

leungbk:
Directives of this type without numeric inputs should not have a
`startFrom` attribute; with a blank value, the writers can produce
extra whitespace.

Also, the stuff around line 1000 is modified because previously, it failed to catch newlines.

Closes #5182.

### DIFF
--- a/src/Text/Pandoc/Readers/RST.hs
+++ b/src/Text/Pandoc/Readers/RST.hs
@@ -41,7 +41,7 @@ import Data.Char (isHexDigit, isSpace, toLower, toUpper, isAlphaNum)
 import Data.List (deleteFirstsBy, elemIndex, intercalate, isInfixOf, isSuffixOf,
                   nub, sort, transpose)
 import qualified Data.Map as M
-import Data.Maybe (fromMaybe, isJust)
+import Data.Maybe (fromMaybe)
 import Data.Sequence (ViewR (..), viewr)
 import Data.Text (Text)
 import qualified Data.Text as T
@@ -503,12 +503,9 @@ includeDirective top fields body = do
   case lookup "code" fields of
        Just lang -> do
          let numberLines = lookup "number-lines" fields
-         let classes = trimr lang : ["numberLines" | isJust numberLines] ++
-                          maybe [] words (lookup "class" fields)
-         let kvs = maybe [] (\n -> [("startFrom", trimr n)]) numberLines
+         let classes =  maybe [] words (lookup "class" fields)
          let ident = maybe "" trimr $ lookup "name" fields
-         let attribs = (ident, classes, kvs)
-         return $ B.codeBlockWith attribs contents'
+         codeblock ident classes numberLines (trimr lang) contents' False
        Nothing   -> case lookup "literal" fields of
                          Just _  -> return $ B.rawBlock "rst" contents'
                          Nothing -> do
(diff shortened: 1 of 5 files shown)


leungbk:
Hmm, I'm not sure why the CI can't load my file. I'm following the example of `3880.md`.

jgm:
> Hmm, I'm not sure why the CI can't load my file. I'm following the example of 3880.md.

You probably need to add it expliclitly to extra-source-files in pandoc.cabal.

src/Text/Pandoc/Readers/RST.hs
```diff
@@ -505,7 +505,11 @@ includeDirective top fields body = do
          let numberLines = lookup "number-lines" fields
          let classes = trimr lang : ["numberLines" | isJust numberLines] ++
                           maybe [] words (lookup "class" fields)
-         let kvs = maybe [] (\n -> [("startFrom", trimr n)]) numberLines
+         let kvs = maybe [] (\n -> let tn = trimr n
```
jgm: `null tn` would be better than `tn == ""`. But as a matter of style, I think this is cleaner:
```haskell
case trimr n of
    [] -> []
    xs -> [("startFrom", xs)]
```
Indeed, you could use a list comprehension:
```haskell
let kvs = [("startFrom", xs) | x <- maybeToList (trimr <$> tn)]
```
but this probably isn't as clear to the reader.

src/Text/Pandoc/Readers/RST.hs
```diff
@@ -999,9 +1003,11 @@ codeblock ident classes numberLines lang body =
                     : maybe [] (const ["numberLines"]) numberLines
                     ++ classes
           kvs     = case numberLines of
-                          Just "" -> []
                           Nothing -> []
-                          Just n  -> [("startFrom",trim n)]
+                          Just n
```
jgm: Similar remarks to the above.

leungbk:
Done. I also tried to dedupe the code a bit.

jgm:
Looks good, thanks!

PR #5212, opened 2019-01-10
LaTeX template: Prevent scaling of main font

adunning:
When `Scale=MatchLowercase` is set as a default font option, this scales `mainfont` against the old default, meaning that it resizes whatever is set as the main font to match the metrics of Latin Modern. This can result, for example, in a document set to 12pt appearing in 11pt or 13pt. Setting this option for individual families allows everything to scale against the main font, and permits the user to override the setting if desired. Note that it is not necessary to specify `Ligatures=TeX`, as this is already set by default for the appropriate families. See the `fontspec` manual: <https://ctan.org/pkg/fontspec>.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -110,31 +110,30 @@ $if(mathspec)$
 $else$
   \usepackage{unicode-math}
 $endif$
-  \defaultfontfeatures{Ligatures=TeX,Scale=MatchLowercase}
-$for(fontfamilies)$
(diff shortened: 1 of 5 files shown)


adunning:
The manual suggests (¶6.2 Scale) that one could instead do this:

```latex
\defaultfontfeatures[\rmfamily]{Scale=1}
```

This would theoretically be a touch cleaner, and avoid the potential issue of a user unintentionally disabling scaling by setting other options. In my testing, however, I found that setting the defaults this way actually loads `mainfont` with both the `Scale=1` and `Scale=MatchLowercase` options, meaning that it doesn't actually work.

jgm:
OK, great!

Issue #4675, opened 2018-05-26
\subtitle undefined with LaTeX default template


MAccadia:
Hi,

When I override `\maketitle` using header-include with this code :

```yaml
---
title:  "De la possibilité et de la réalité de l'introspection"
subtitle: "Une exploration intérieure"
author: "Archibald Dulapin"
header-includes:
  - |
    ```{=latex}
    \makeatletter
    \renewcommand{\maketitle}{%
        \@title

        \@author
        
        \@subtitle
    }
    \makeatother
    ```
...
```

I get the following error with _pandoc 2.2.1_:

```
$ pandoc test.yml -o test.pdf
Error producing PDF.
! Undefined control sequence.
\maketitle ...title \par \@author \par \@subtitle 
                                                  
l.73 \maketitle
```

It works if I use a modified default.latex template with :
```latex
\makeatletter
\providecommand\subtitle[1]{\renewcommand\@subtitle{#1}}
\providecommand\@subtitle{\@latex@error{No \noexpand\subtitle given}\@ehc}
\makeatother
\subtitle{$subtitle$}
```

I don't know if this code could have any side effect ; I don't understand it well enough. I got it from [here](https://stackoverflow.com/questions/2833954/how-do-i-create-a-command-like-title-in-latex).

As for a workaround, putting the `\providecommand`s in header-include works.

mb21:
Please ask questions this on pandoc-discuss. This tracker is for issues only.

(btw, tot sure what you're asking anyway...)

MAccadia:
Hi, sorry if I wasn't clear. 

My point is that the [default.latex](https://github.com/jgm/pandoc-templates/blob/master/default.latex) template could be improved with :
```latex
$if(subtitle)$
\makeatletter
\providecommand\subtitle[1]{\renewcommand\@subtitle{#1}}
\providecommand\@subtitle{\@latex@error{No \noexpand\subtitle given}\@ehc}
\makeatother
\subtitle{$subtitle$}
$endif$
```

instead of

```latex
$if(subtitle)$
\providecommand{\subtitle}[1]{}
\subtitle{$subtitle$}
$endif$
```

Issue #3896, opened 2017-09-06
Subtitle (from YAML header) not appearing in PDF when converting from markdow


dylan-chong:
Pandoc: 1.19.2.1, Mac OS X 10.11.6

##  markdown file (f.md)

```
---
title: Title
subtitle: Subtitle
author:
    - I am an author
date: \today{}
---

I am text
```

## Output command

`pandoc f.md -o f.pdf`

## Result

The subtitle is not shown in the PDF

<img width="637" alt="screen shot 2017-09-06 at 7 10 27 pm" src="https://user-images.githubusercontent.com/17482349/30098683-118b9b78-9337-11e7-8040-f2e98e015fd4.png">

mb21:
From the [MANUAL](http://pandoc.org/MANUAL.html#variables-set-by-pandoc):

> document subtitle, included in HTML, EPUB, LaTeX, ConTeXt, and Word docx; renders in LaTeX only when using a document class that supports \subtitle, such as beamer or the KOMA-Script series (scrartcl, scrreprt, scrbook)

stevegt:
Thank You!  Saved my life -- several hours of it, anyway.

You see, I'm one of them thar UNIX sysadmin types with the requisite 30 years of crotchety experience hacking around in arcane voodoo.  I, too, saw that same paragraph in the manual yesterday, and I've spent the intervening hours trying to figure out what the heck a user is supposed to do.  Latex docs, stack exchange, and the intertubes at large aren't much help with something this simple -- most latex conversation is between people writing masters theses.  

This is the only example I've been able to find that shows how to do this.  It might be worth adding it to that paragraph in the manual.

Issue #1327, opened 2014-06-09
\subtitle undefined when converting reStructuredText to PDF


kwohlfahrt:
The following reStructuredText document causes an error (note addition of the subtitle):

```
===================
Pandoc User's Guide
===================
---------------
Pandoc Subtitle
---------------

:Author: John MacFarlane
:Date:   January 19, 2013

.. contents::
   :depth: 3.0
..

Synopsis
========

Some text.
```

```
kwohlahrt@laptop> pandoc pandoc_test.rst -o test.pdf
! Undefined control sequence.
l.46 \subtitle

pandoc: Error producing PDF from TeX source
```

The `rst2latex.py` provided by docutils works as expected, the `Pandoc User's Guide` is generated as a large title, with `Pandoc Subtitle` below that as a subtitle. `Synopsis` is then a top-level header.

The issue occurs both with pandoc 1.12 and 1.12.4, installed from gentoo repository.

```
pandoc --version
pandoc 1.12.4
Compiled with texmath 0.6.6.1, highlighting-kate 0.5.7.
Syntax highlighting is supported for the following languages:
    actionscript, ada, apache, asn1, asp, awk, bash, bibtex, boo, c, changelog,
    clojure, cmake, coffee, coldfusion, commonlisp, cpp, cs, css, curry, d,
    diff, djangotemplate, doxygen, doxygenlua, dtd, eiffel, email, erlang,
    fortran, fsharp, gcc, gnuassembler, go, haskell, haxe, html, ini, isocpp,
    java, javadoc, javascript, json, jsp, julia, latex, lex, literatecurry,
    literatehaskell, lua, makefile, mandoc, markdown, matlab, maxima, metafont,
    mips, modelines, modula2, modula3, monobasic, nasm, noweb, objectivec,
    objectivecpp, ocaml, octave, pascal, perl, php, pike, postscript, prolog,
    python, r, relaxngcompact, restructuredtext, rhtml, roff, ruby, rust, scala,
    scheme, sci, sed, sgml, sql, sqlmysql, sqlpostgresql, tcl, texinfo, verilog,
    vhdl, xml, xorg, xslt, xul, yacc, yaml
```

tomleo:
pandoc 1.12.2.1

```
pandoc -f rst document.rst --smart -o document.pdf
pandoc: Error producing PDF from TeX source.
! Undefined control sequence.
l.67 \subtitle
```

adunning:
[...]

PR #5213, opened 2019-01-10
 LaTeX template: Render \subtitle

adunning:
This renders `\subtitle` without requiring a different document class such as `scrartcl`, ~~using the `titling` package. This integrates the solution in the manual, which I believe was based on an example from Enrico Gregorio, <https://tex.stackexchange.com/a/50186> but is simplified.~~ The matter has come up several times (#4675, #3896, #1327 and on pandoc-discuss), so it seems worthwhile to include it by default even if requiring `titling` isn't ideal.

The blank command was added in <https://github.com/jgm/pandoc/commit/4ac59af2429d97c1bb1f54088453a97205c0fcdc> to avoid problems with packages that define `\subtitle`. I've verified that this solution will not load with Beamer or KOMA (I was initially using a separate test for KOMA, but `\providecommand` is sufficient to stop `titling` from loading).

~~I'm not using a test for the presence of `titling` on the principle of providing an error rather than silently changing key aspects of the document's appearance, but there are some distributions that omit it (which was why I didn't add this code to the template earlier), so this could be added.~~

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -166,8 +166,7 @@ pandoc does not require them to be present:
 [`parskip`] (for better inter-paragraph spaces),
 [`xurl`] (for better line breaks in URLs),
 [`bookmark`] (for better PDF bookmarks),
-[`titling`] (if `subtitle` is specified), and
-[`footnote`] (to fix footnotes in tables).
+and [`footnote`] (to fix footnotes in tables).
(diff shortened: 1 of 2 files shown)


jgm:
I like this solution!

adunning:
Thank you!

PR #5214, opened 2019-01-11
Make Math produced by Asciidoc writer work with both Asciidoc and Asciidoctor

tarikgraba:
Feature request discussed here:
https://groups.google.com/forum/#!topic/pandoc-discuss/Hz-XBuNhMA8

Adds a tex_math_bare extension for which the latexmath is written without delimiters (`$` or `\[`)

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -279,7 +279,7 @@ General options {.options}
 :   Specify output format.  *FORMAT* can be:
 
     ::: {#output-formats}
-    - `asciidoc` ([AsciiDoc])
+    - `asciidoc` ([AsciiDoc]) or `asciidoctor` ([AsciiDoctor])
     - `beamer` ([LaTeX beamer][`beamer`] slide show)
     - `commonmark` ([CommonMark] Markdown)
     - `context` ([ConTeXt])
@@ -465,6 +465,7 @@ General options {.options}
 [GNU Texinfo]: http://www.gnu.org/software/texinfo/
 [Emacs Org mode]: http://orgmode.org
 [AsciiDoc]: http://www.methods.co.nz/asciidoc/
+[AsciiDoctor]: https://asciidoctor.org/
 [DZSlides]: http://paulrouget.com/dzslides/
 [Word docx]: https://en.wikipedia.org/wiki/Office_Open_XML
 [PDF]: https://www.adobe.com/pdf/
@@ -3513,7 +3514,11 @@ reStructuredText
   ~ It will be rendered using an [interpreted text role `:math:`].
 
 AsciiDoc
-  ~ It will be rendered as `latexmath:[...]` when inline and `[latexmath]++++....++++` when in display mode.
+  ~ For AsciiDoc output format (`-t asciidoc`) it will appear verbatim
+    surrounded by `latexmath:[$...$]` (for inline math) or
+    `[latexmath]++++\[...\]+++` (for display math).
+    For AsciiDoctor output format (`-t asciidoctor`) the LaTex delimiters
+    (`$..$` and `\[..\]`) are omitted.
 
 Texinfo
   ~ It will be rendered inside a `@math` command.
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3513,7 +3513,7 @@ reStructuredText
   ~ It will be rendered using an [interpreted text role `:math:`].
 
 AsciiDoc
-  ~ It will be rendered as `latexmath:[...]`.
+  ~ It will be rendered as `latexmath:[...]` when inline and `[latexmath]++++....++++` when in display mode.
 
 Texinfo
   ~ It will be rendered inside a `@math` command.
--- a/data/templates/default.asciidoctor
+++ b/data/templates/default.asciidoctor
@@ -1,0 +1,38 @@
+$if(titleblock)$
+= $title$
+$if(author)$
+$for(author)$$author$$sep$; $endfor$
+$endif$
+$if(date)$
+$date$
+$endif$
+$if(keywords)$
+:keywords: $for(keywords)$$keywords$$sep$, $endfor$
+$endif$
+$if(lang)$
+:lang: $lang$
+$endif$
+$if(toc)$
+:toc:
+$endif$
+
+$endif$
+$if(abstract)$
+[abstract]
+== Abstract
+$abstract$
+
+$endif$
+$for(header-includes)$
+$header-includes$
+
+$endfor$
+$for(include-before)$
+$include-before$
+
+$endfor$
+$body$
+$for(include-after)$
+
+$include-after$
+$endfor$
--- a/data/templates/default.asciidoctor
+++ b/data/templates/default.asciidoctor
@@ -15,6 +15,9 @@ $endif$
 $if(toc)$
 :toc:
 $endif$
+$if(math)$
+:stem: latexmath
+$endif$
 
 $endif$
 $if(abstract)$
--- a/src/Text/Pandoc/Writers/AsciiDoc.hs
+++ b/src/Text/Pandoc/Writers/AsciiDoc.hs
@@ -414,7 +414,10 @@ inlineToAsciiDoc _ (Str str) = return $ text $ escapeString str
 inlineToAsciiDoc _ (Math InlineMath str) =
   return $ "latexmath:[$" <> text str <> "$]"
 inlineToAsciiDoc _ (Math DisplayMath str) =
-  return $ "latexmath:[\\[" <> text str <> "\\]]"
+  return $
+      blankline <> "[latexmath]" $$ "++++" $$
+      "\\[" <> text str <> "\\]"
+      $$ "++++" $$ blankline
 inlineToAsciiDoc _ il@(RawInline f s)
   | f == "asciidoc" = return $ text s
   | otherwise         = do
--- a/src/Text/Pandoc/Writers/AsciiDoc.hs
+++ b/src/Text/Pandoc/Writers/AsciiDoc.hs
@@ -37,7 +37,7 @@ that it has omitted the construct.
 
 AsciiDoc:  <http://www.methods.co.nz/asciidoc/>
 -}
-module Text.Pandoc.Writers.AsciiDoc (writeAsciiDoc) where
+module Text.Pandoc.Writers.AsciiDoc (writeAsciiDoc, writeAsciiDoctor) where
 import Prelude
 import Control.Monad.State.Strict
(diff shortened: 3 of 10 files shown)


mb21:
Instead of adding a new extension, we could also enable `tex_math_dollars` by default for the AsciiDoc Writer and do:

```haskell
let content = if isEnabled Ext_tex_math_dollars opts
                then "$" <> text str <> "$"
                else text str
```

tarikgraba:
How do we enable an extension for a particular writer? is it the `getDefaultExtensions` function?

mb21:
Yes, I think extending the [`getDefaultExtensions`](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Extensions.hs#L328) function with a `getDefaultExtensions "asciidoc" case.

But let's first hear whether @jgm agrees. Maybe there was a reason he proposed a new extension.

mb21:
No, it's also here: http://pandoc.org/MANUAL.html#extension-tex_math_dollars

And it's actually used by the markdown writer as well.

jgm:
Does original asciidoc allow `$$...$$` for display math, or just `\[...\]`?
One reason not to go with `tex_math_dollars` is that it isn't appropriate for the latter form.

mb21:
From above pandoc-discuss thread:

> With the original asciidoc, the author needs to add Latex math delimiters (i.e. `$$` and `\[\]`) while in asciidoctor they are added by the tool when printed back.

jgm:
@mb21 Yes - I was wondering, though, whether `$$..$$` also works (in addition to `\[..\]` in original asciidoc.  If not, then I think using `tex_math_dollars` is a nonstarter.

mb21:
@tarikgraba  what do you think? Which name is more appropriate?

tarikgraba:
> Does original asciidoc allow `$$...$$` for display math, or just `\[...\]`?
> One reason not to go with `tex_math_dollars` is that it isn't appropriate for the latter form.

As I understand the way Asciidoc handles latexmath macro/bloc are "passthrough" blocks, and their content is outputted without modification.

tarikgraba:
> @tarikgraba what do you think? Which name is more appropriate?

If some people already use pandoc to generate asciidoc and except having bloc latexmath surrounded by `\[` `\]`, replacing them by `$$` could break their workflow.

tarikgraba:
I think that it is better to have an explicit extension to produce bare math instead of forcing the usage of `$$` for math blocks.
When writing math in Latex, we are supposed to use `$` for inline math and `\[` for display math.
`$$` is a plain Tex command which we are not suppose to use with modern Latex.

jgm:
pandoc's LaTeX writer already uses \(..\) and \[..\].
This issue is just about asciidoc output, right?
I don't see much use for a bare_math extension outside
of asciidoc, do you?  And it's not really "bare",
since you need  the surrounding brackets and prefix.
That's part of my hesitation about the extension
idea.

One idea would be to introduce an `asciidoctor`
extension that only affects asciidoc output and turns
on all special asciidoctor behaviors.

tarikgraba:
>One idea would be to introduce an `asciidoctor` extension that only affects asciidoc output and turns on all special asciidoctor behaviors.

I think it is reasonable. 

Any example of who to add a writer specific extension?

jgm:
Tarik Graba <notifications@github.com> writes:

>>One idea would be to introduce an `asciidoctor` extension that only affects asciidoc output and turns on all special asciidoctor behaviors.
>
> I think it is reasonable. 
>
> Any example of who to add a writer specific extension?

There is no real concept of a writer-specific
extension. There are  just extensions.  They started
out as "markdown extensions," but I think we may now
have one or two that affect other formats:
empty_paragraphs, styles, amuse, ntb.

Another option would be to just have two different
output formats, asciidoc and asciidoctor.  (This could
then be handled the way we handle plain and markdown;
see the markdown writer -- an extension would not
be needed.)

jgm:
> Another option would be to just have two different output formats, asciidoc and asciidoctor.

Indeed, this would be my preferred way of dealing with this, I think.
However, I'm not involved with asciidoc at all.  To what extent do people
think of asciidoctor as a different dialect?

jgm:
@tarkigraba do you have thoughts on the proposal to have two formats, `asciidoc` and `asciidoctor`?

tarikgraba:
> @tarkigraba do you have thoughts on the proposal to have two formats, `asciidoc` and `asciidoctor`?

The two writers will have a lot of code in common.  Asccidoctor tries to be as compatible as possible with the original Asciidoc format and there are very few exceptions.

If there are two format the writer should be unique.

mb21:
@tarikgraba The question is more what the user would expect to write, which one of the following:

    -t asciidoctor
    -t asciidoc+asciidoctor
    -t asciidoc+tex_math_dollars
    -tasciidoc+tex_math_bare

`-t asciidoctor` could be implemented like `-t plain`, as an extension on top of the existing reader.

tarikgraba:
`-t asciidoctor` seems the better choice. It is the most understandable and would also allow to adapt the behavior of the writer if other tweaks are needed.

jgm:
If we distinguished `asciidoc` and `asciidoctor`, we would not have two separate writers.

We'd just have a field in the writer state that keeps track of which has been requested.  Two separate front-end functions `writeAsciidoc` and `writeAsciidoctor` would be provided; these would set this field and then call the common writer code.  This is exactly how we currently distinguish `epub2/epub3`, `docbook4/docbook5`, `html4/html5`, and also `markdown/plain`.

tarikgraba:
Ok I"ll take a look at the code.

tarikgraba:
Seems ok.

MANUAL.txt
```diff
@@ -3513,7 +3514,11 @@ reStructuredText
   ~ It will be rendered using an [interpreted text role `:math:`].
 
 AsciiDoc
-  ~ It will be rendered as `latexmath:[...]`.
+  ~ For AsciiDoc output format (`-t asciidoc`) it will appear verbatim
+    surrounded by `latexmath:[$...$]` (for inline math) or
+    `[latexmath]++++\[...\]+++` (for display math).
+    For AsciiDoctor output format (`-t asciidoctor`) the Latex delimiters
```
jgm: Latex -> LaTeX

src/Text/Pandoc/Writers/AsciiDoc.hs
```diff
@@ -61,18 +61,28 @@ data WriterState = WriterState { defListMarker    :: String
                                , bulletListLevel  :: Int
                                , intraword        :: Bool
                                , autoIds          :: Set.Set String
+                               , asciidoctorDlct  :: Bool
```
jgm: Dlct is a bit obscure.  `asciidoctorVariant`?
or just `asciidoctor`?

src/Text/Pandoc/Writers/AsciiDoc.hs
```diff
@@ -61,18 +61,28 @@ data WriterState = WriterState { defListMarker    :: String
...
+defaultWriterState :: WriterState
+defaultWriterState = WriterState{
+                                 defListMarker = "::"
+                               , orderedListLevel = 1
+                               , bulletListLevel = 1
+                               , intraword = False
+                               , autoIds = Set.empty
+                               , asciidoctorDlct = False
+                                 }
+
```
jgm: Just as a matter of formatting style, I think it looks nicer to line up the bottom `}` with the comma. (And perhaps also the top `{`)

data/templates/default.asciidoctor
```diff
@@ -0,0 +1,38 @@
...
+$if(keywords)$
+:keywords: $for(keywords)$$keywords$$sep$, $endfor$
+$endif$
+$if(lang)$
+:lang: $lang$
+$endif$
+$if(toc)$
+:toc:
+$endif$
+
```
jgm: asciidoctor manual says that when latex math is used,
```
:stem: latexmath
```
is needed in the title block.  We could have a variable in WriterState that keeps track of whether math has been used and sets a variable in the template that causes this to be inserted.

I don't know if something similar is also needed for asciidoc.

src/Text/Pandoc/Writers/AsciiDoc.hs
tarikgraba: done

src/Text/Pandoc/Writers/AsciiDoc.hs
tarikgraba: done

data/templates/default.asciidoctor
tarikgraba: done

MANUAL.txt
tarikgraba: done

MANUAL.txt
```diff
@@ -3513,7 +3513,7 @@ reStructuredText
   ~ It will be rendered using an [interpreted text role `:math:`].
 
 AsciiDoc
-  ~ It will be rendered as `latexmath:[...]`.
+  ~ It will be rendered as `latexmath:[...]` when inline and `[latexmath]++++....++++` when in display mode.
```
jgm: Line longer than 80 characters, should be wrapped.

test/writer.asciidoc
```diff
@@ -466,7 +466,11 @@ Ellipses…and…and….
 * latexmath:[$223$]
 * latexmath:[$p$]-Tree
 * Here’s some display math:
-latexmath:[\[\frac{d}{dx}f(x)=\lim_{h\to 0}\frac{f(x+h)-f(x)}{h}\]]
+
```
jgm: I'm curious about this example.  Does it actually work properly when formatted using asciidoc?  Do you really get one big list with the math in one list item?

jgm: After experimenting, I can confirm that the math in the list item doesn't work. The blank line causes the list to be closed, then the display math occurs, then a new list.  In order to get the right result you need
```
* Here's some display math:
+
[latexmath]
++++
etc
```
But note: we *only* want the single `+` connector if we're in a list item.

jgm: Also, asciidoctor seems to choke on the empty list item.
We might fix that somehow (maybe inserting  a nonbreaking space?), but that doesn't need to go in this PR.

jgm: I think you can check to see if you're in a list by looking at orderedListLevel and bulletListLevel in WriterState.
If they're both == 1 you're not in a list.

jgm:
I'm going to merge this as is and make the minor change to the 'in-list math'.
Thanks for this excellent contribution!

PR #5215, opened 2019-01-11
docx writer -- support custom properties

### LINKED ISSUE #3024: Textile reader: bold and inline code interfere with each other
It is not possible to read correctly a formatting made to look like GitHub mentions. The presence of `@` and `*` interlaced breaks the output. This syntax should probably be supported as it is supported on https://txstyle.org/ and in Redmine.

Actual:
```
$ echo '*@John* not bold *@Jane*' | pandoc -f textile -t markdown
**`John* not bold *`Jane**
```

Expected:
```
$ echo '*@John* not bold *@Jane*' | pandoc -f textile -t markdown
**@John** not bold **@Jane**
```

### LINKED ISSUE #3034: [FEATURE REQUEST] Ability to convert yaml front matter into custom properties in docx
I think this is pretty self-explanatory, but the ability to set custom properties in a Word document would be a total game-changer, especially for those of us working for companies so entrenched in the MS garden that we use SharePoint as a DMS. I have hundreds of md files that I'd love to convert to Word and have them retain the custom key in the key-value pair. Then it's just a drag and drop into a SharePoint library, which should pick up those custom properties and map them correctly for list views, etc.

I appreciate this is a very big ask. Cheers.

agusmba:
partially fixes #3034 (only the docx writer part)

I have only done some preliminary tests, LibreOffice seems to work fine.
Word also works, but has a small problem with not refreshing the document properties inserted in the header/footer (via reference-doc) when opening the document. That was a previously existing problem (workaround: update references from Word)

Note that I'm learning Haskell just for the sake of fixing this, so I may not have used the best style or taken into consideration important things, although I tried to do things as close as what the ODT writer does it.

UPDATE:

Things I would like to do:

* [x] Implement basic support for converting pandoc properties into core and custom docx properties
* [x] Improve the population of core.xml properties, including them only when they are actually present in the original document metadata
* [x] Add golden test(s) for this feature
* [x] Fix incorrectly equating *cd:subject* and *subtitle*
* [x] ~decide on the proper~ pandoc mapping for *dc:description*: \[ ~abstract~ | description | ~comments~ \]
* [x] do not remove "date" from the custom properties
* [x] review and update MANUAL if needed
* [x] update golden tests for this

If we had docx property reader support (so maybe for a later date):

* Improve/implement automatic roundtrip tests for docx properties support
* Take the document properties defined in the `reference-doc` as base values that can be overridden by the original document metadata. This way, we need to include the properties in the original document metadata only when they differ from the defaults in the reference document.


src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -512,7 +513,8 @@ writeDocx opts doc@(Pandoc meta _) = do
   let docPropsEntry = toEntry docPropsPath epochtime $ renderXml docProps
 
   let customProperties :: [(String, String)]
-      customProperties = [] -- FIXME
+      customProperties = [(k, escapeStringForXML $ lookupMetaString k meta) | k <- M.keys (unMeta meta)
+                         , k `notElem` ["title", "lang", "author", "date"]]  -- ^ from ODT.hs
```
agusmba: Apparently this is not a proper comment, my bad...

src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -512,7 +513,8 @@ writeDocx opts doc@(Pandoc meta _) = do
   let docPropsEntry = toEntry docPropsPath epochtime $ renderXml docProps
 
   let customProperties :: [(String, String)]
-      customProperties = [] -- FIXME
+      customProperties = [(k, escapeStringForXML $ lookupMetaString k meta) | k <- M.keys (unMeta meta)
```
jgm: You shouldn't need escapeStringForXML here; I believe mknode will handle that automatically.
In fact, my guess is that with this you'll get double escaping.  Either way, you should test to make sure the escaping is being done correctly.

agusmba: Ok, I'll check that out, thanks for the review!

Besides manual testing, if I wanted to automate some tests, I guess I'd need to implement also reading the docx properties back into pandoc, right? (or is that already supported?)

agusmba:
@jgm you were of course right, the escapeStringForXML was not helpful here.

I've manually tested it with:

```yaml
title: "Pandoc Issue 3034"
custom1: "First custom property value"
"Second C.P.": "Second custom property value"
custom3: "Escaping amp & ."
custom4: "Escaping LT,GT < asdf > <"
custom5: "Escaping html <i>asdf</i>"
custom6: "Escaping *MD* á a"
nested-custom:
  - custom 7: "Nested Custom value 7"
    custom 8: "Nested Custom value 8"
custom9: "Extended chars: € á é í ó ú $"
```

And I get:

pandoc-issue-3034

It seems to work. The formating is lost (custom5 and 6), but everything else is passed correctly. The nested properties are not copied, except for the empty parent (nested-custom). I guess we could improve this last point doing something akin to the *keywords* or at least ignoring non-string root-level properties.

Regarding the dirty flag for docProperties included in a docx reference-doc header, it seems to be enough having one dirty flag in the document, thus if we use `--toc` in the command line, the properties will be updated correctly (because the TOC makes Word refresh all fields at opening time). Otherwise, the reference doc can be manually edited (at the xml level) to add the dirty flag to the inserted docProperty field. I think we can live with this for the time being, it's not always a problem, and when it is we have a workaround.

Regarding the docx reader, I'm not quite sure how to tackle it at my level of understanding of both Haskell and pandoc's codebase, so automatic roundtrip testing of this feature is a bit difficult. Do you have any tips on what kind of automatic tests I could do for this?

agusmba:
I need to review the documentation to see if I should add anything to it, but appart from that, I think this could be it for this PR. I'm not ready to tackle the reader yet.
@jgm do you see anything else I should do?

src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -497,6 +497,9 @@ writeDocx opts doc@(Pandoc meta _) = do
                        _                  -> []
 
   let docPropsPath = "docProps/core.xml"
+  let extraCorePropsMap = M.fromList 
+          [("subtitle","dc:subject"),("abstract","dc:description")
+          ,("lang","dc:language"),("category","cp:category")]
```
agusmba: Better use two lists and zip them for the Map constructor, so that one of the lists can be reused in the filter below (~515) to avoid hard-coded duplicates of pandoc metadata names.

agusmba:
Is it ok if I rebase on master and clean the history of this branch a bit? (either now or when it is deemed ready for merge).

jgm:
Sure feel free to rebase

HeirOfNorton:
One thing with the core properties: Subject is not really equivalent to Subtitle, at least not the way Subtitle is used in the other templates. A Subtitle is like a secondary title. According to Dublin Core, Subject is supposed to be keywords or codes describing the topic, preferably using a controlled vocabulary. ([See here](http://dublincore.org/documents/dces/#subject))Totally different. Dublin Core doesn't really have an equivalent to Subtitle, except to use an alternative or refined Title entry.

MANUAL.txt
```diff
@@ -3141,6 +3141,20 @@ equivalent of the Markdown in the `abstract` field:
     <p>This is the abstract.</p>
     <p>It consists of two paragraphs.</p>
 
+When writing DOCX or ODT, first-level metadata with a String value will be
+included in the resulting document as Document Properties or Custom
+Properties accordingly. The name of the metadata will be the same
+as the Document/Custom Property, except in the following cases:
+
+`subtitle`
```
agusmba: "subtitle" <-> "subject" is no longer an exception, this association has been removed.

agusmba:
I've made a different PR for the odt writer, but both that one and this one should go hand in hand, as they provide similar functionality to both docx and odt writers.

agusmba:
Further discussion affecting this PR is happening [here](https://github.com/jgm/pandoc/pull/5238#issuecomment-456467088)

agusmba:
Closing in favor of https://github.com/jgm/pandoc/pull/5252

PR #5218, opened 2019-01-13
LaTeX template: Simplify fontspec usage

adunning:
Simplify the approach of #5212, ensuring that `mainfont` is used as the base font for scaling and that LuaLaTeX does not display the mono font with TeX ligatures (as it does not use the `Mapping=tex-ansi` option).

With this modified version of `\defaultfontfeatures`, fontspec erroneously reports scaling against the old default font in the log (this had confused me earlier, hence my previous approach), but it nonetheless displays the main font at the specified size. Using this rather than setting `Scale=MatchLowercase` for each family individually means that users will not lose scaling when upgrading to the new template if they were using other font options. Scaling can be disabled for an individual family by adding the option `Scale=1` to `sansfontoptions`, `monofontoptions`, etc.

Remove the `\setromanfont` command added in #4665, as this is not documented in the fontspec manual and appears to be a deprecated alias for `\setmainfont`.

As I neglected to include it in #5212, this illustrates the differences between this version of the template and the one in pandoc 2.5, using the [Adobe Source](https://github.com/adobe-fonts) series of fonts:

```latex
pandoc -o fonttest.pdf --pdf-engine=lualatex << EOT

---
fontsize: 12pt
mainfont: Source Serif Pro
sansfont: Source Sans Pro
monofont: Source Code Pro
---

"The quick fox --- \the\fontdimen6\font\relax"

\sffamily "The quick fox --- \the\fontdimen6\font\relax"

\ttfamily "The quick fox --- \the\fontdimen6\font\relax"

EOT
````

Pandoc 2.5 template:

screen shot 2019-01-12 at 19 51 22

Compare the results of the Pandoc 2.5 template with XeLaTeX:

screen shot 2019-01-12 at 19 57 23

Updated template with either XeLaTeX or LuaLaTeX:

screen shot 2019-01-12 at 19 51 50

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1527,13 +1527,12 @@ LaTeX variables are used when [creating a PDF].
     `fontfamily` set to [`mathpazo`] provides Palatino with old-style
     figures and true small caps; may be repeated for multiple options
 
-`mainfont`, `romanfont`, `sansfont`, `monofont`, `mathfont`, `CJKmainfont`
+`mainfont`, `sansfont`, `monofont`, `mathfont`, `CJKmainfont`
 :   font families for use with `xelatex` or
     `lualatex`: take the name of any system font, using the
-    [`fontspec`] package.  Note that if `CJKmainfont` is used,
-    the [`xecjk`] package must be available.
+    [`fontspec`] package.  `CJKmainfont` uses the [`xecjk`] package.
 
-`mainfontoptions`, `romanfontoptions`, `sansfontoptions`, `monofontoptions`, `mathfontoptions`, `CJKoptions`
+`mainfontoptions`, `sansfontoptions`, `monofontoptions`, `mathfontoptions`, `CJKoptions`
 :   options to use with `mainfont`, `sansfont`, `monofont`, `mathfont`,
     `CJKmainfont` in `xelatex` and `lualatex`.  Allow for any choices
     available through [`fontspec`], such as the OpenType features
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -110,30 +110,29 @@ $if(mathspec)$
 $else$
   \usepackage{unicode-math}
 $endif$
+  \defaultfontfeatures{Scale=MatchLowercase}
+  \defaultfontfeatures[\rmfamily]{Ligatures=TeX,Scale=1}
 $if(mainfont)$
   \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$]{$mainfont$}
 $endif$
-$if(romanfont)$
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by adunning** [comment]
For the release notes, I should also add that one can imitate the previous appearance with `-V mainfontoptions="Scale=MatchLowercase"`:

```latex
pandoc -o fonttest.pdf --pdf-engine=lualatex << EOT

---
fontsize: 12pt
mainfont: Source Serif Pro
mainfontoptions: Scale=MatchLowercase
sansfont: Source Sans Pro
monofont: Source Code Pro
---

"The quick fox --- \the\fontdimen6\font\relax"

\sffamily "The quick fox --- \the\fontdimen6\font\relax"

\ttfamily "The quick fox --- \the\fontdimen6\font\relax"

EOT
```

jgm:
Excellent, thanks!

adunning:
I feel a bit badly about the roundabout series of commits; here is a summary for the release notes:

```markdown
* LaTeX template (Andrew Dunning)

  + Prevent fontspec from scaling `mainfont` to match the default font, Latin Modern. For example, a main font set to 12pt could previously appear between 11pt to 13pt depending on its design. To return to the earlier rendering, use `-V mainfontoptions="Scale=MatchLowercase"` (#5212, #5218).
  + Display monospaced fonts without TeX ligatures when using `--pdf-engine=lualatex`. It now matches the behaviour of other engines (#5212, #5218).
  + Remove the deprecated `romanfont` variable. The functionality of `mainfont` is identical (#5218).
  + Render `\subtitle` with the standard document classes. Previously, `subtitle` only appeared when using the KOMA-Script classes or Beamer (#5213).
  + Use Babel instead of Polyglossia for LuaLaTeX. This avoids several language selection problems, notably with retaining French spacing conventions when switching to a verbatim environment or another language; and in printing Greek text without hyphenation (#5193).
  + Use the [`xurl`](https://ctan.org/pkg/xurl) package if available, improving the appearance of URLs by allowing them to break at additional points. Requires TeX Live 2018 or later; run `tlmgr update --all` if you experience problems with URLs appearing after link text (#5193).
  + Use [`bookmark`](https://ctan.org/pkg/bookmark) if available, which can correct the levels of headings where `hyperref` cannot: see the [KOMA-Script 3.26 release notes](https://komascript.de/release3.26) (#5193).
  + Require the [`xcolor`](https://ctan.org/pkg/xcolor) package to avoid a possible error when using additional packages alongside footnotes in tables (#5193, closes #4861).
  + Remove obsolete `fixltx2e` package, which has no functionality with TeX Live 2015 or later (#5193).
  + Allow multiple `fontfamilies.options` (#5193, closes #5194).
  + Reindent file (#5193).
```

jgm:
Much appreciated!

PR #5219, opened 2019-01-13
LaTeX template: Restrict `institute` to Beamer

### LINKED ISSUE #4209: "institute" metadata doesn't do anything
I tried to compile the following document

```markdown
---
title: test
institute: institute
---
```

But the `institute` field just doesn't work.

I checked my default.latex, and it contains
```latex
$if(institute)$
\providecommand{\institute}[1]{}
\institute{$for(institute)$$institute$$sep$ \and $endfor$}
$endif$
```

I tried several variants, but it just doesn't work.

adunning:
Use the `etoolbox` package (same technique as #5213) to add the `institute` variable to `\maketitle`. Verified to work with `article`, `book`, `report`, `memoir`, `scrartcl`. Closes #4209.

It's not the prettiest thing ever: the tabular environment of `\author` and `\and` is rather inflexible. I'm wondering if the `institute` variable was intended only to support Beamer users (in which case it could simply be excluded from normal LaTeX output), but the manual indicates that it's supposed to work in both LaTeX and Beamer.

Example:

```shell
pandoc -o institute.pdf << EOT

---
title: Test
subtitle: More tests
author:
- Tester 1
- Tester 2
institute:
- University of Testing
- University of Diagnostics
date:
- 2019
---

EOT
```

screen shot 2019-01-12 at 23 09 27

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1362,10 +1362,6 @@ depending on the output format, but include the following:
 `subtitle`
 :   document subtitle, included in HTML, EPUB, LaTeX, ConTeXt, and Word docx
 
-`institute`
-:   author affiliations (in LaTeX and Beamer only).  Can be a
-    list, when there are multiple authors.
-
 `abstract`
 :   document summary, included in LaTeX, ConTeXt, AsciiDoc, and Word docx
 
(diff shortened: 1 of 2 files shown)


adunning:
It looks slightly better with reduced spacing:

screen shot 2019-01-12 at 23 45 17

jgm:
`institute` was indeed intended only for beamer.  This could be clarified in the docs.
And maybe it should be made conditional on beamer in the template?

adunning:
That's certainly cleaner; I've added another commit for this.

adunning:
Looks like the testing system is having a bad day; this works locally.

PR #5225, opened 2019-01-14
sample.lua: Fix some minor issues

MichaWiedenmann:
* The closing tag ' </table' was missing a '>'.
* The content of list items was not escaped.

### DIFF
--- a/data/sample.lua
+++ b/data/sample.lua
@@ -332,7 +332,7 @@ function Table(caption, aligns, widths, headers, rows)
     end
     add('</tr>')
   end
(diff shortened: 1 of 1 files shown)


tarleb:
Good catch on the missing `>`!

I don't understand the additional uses of `escape`, what is the rational for that change? From what I see, this leads to content being double-escaped.

MichaWiedenmann:
I agree, the second commit was a mistake. I have removed it.

tarleb:
Great, thanks!

PR #5234, opened 2019-01-20
LaTeX template: Use `footnotehyper` if available

adunning:
Use the [`footnotehyper`](https://ctan.org/pkg/footnotehyper/) package if available, a rewrite of `footnote` that is compatible with `hyperref` and `babel-frenchb`. This also addresses the incompatibility with `xcolor` noted in #4861, but the new package has only been available since 2016, so this template still loads `xcolor` earlier for compatibility with older distributions. Note that the `footnote` package is part of `collection-latexrecommended` in TeX Live, so there shouldn't be any situations where it isn't available.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -247,8 +247,9 @@ $if(beamer)$
 \def\fnum@table{\tablename~\thetable}
 \makeatother
 $else$
-% Allow footnotes in tables
-\IfFileExists{footnotehyper.sty}{\usepackage{footnotehyper}}{\usepackage{footnote}}\makesavenoteenv{longtable}
+% Allow footnotes in longtable head/foot
+\IfFileExists{footnotehyper.sty}{\usepackage{footnotehyper}}{\usepackage{footnote}}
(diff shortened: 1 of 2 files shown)


data/templates/default.latex
```diff
@@ -247,8 +247,8 @@ $if(beamer)$
 \def\fnum@table{\tablename~\thetable}
 \makeatother
 $else$
-% Fix footnotes in tables (requires footnote package)
-\IfFileExists{footnote.sty}{\usepackage{footnote}\makesavenoteenv{longtable}}{}
+% Allow footnotes in tables
+\IfFileExists{footnotehyper.sty}{\usepackage{footnotehyper}}{\usepackage{footnote}}\makesavenoteenv{longtable}
```
jgm: Maybe put `\makesavenoteenv{longtable}` on a new line, since it is no longer part of the `IfFileExists`??

adunning: Done; thanks!

jgm:
thanks!

PR #5235, opened 2019-01-20
LaTeX template: Respect `numbersections` for books

adunning:
Ensure that `\part` and `\chapter` are only numbered if `numbersections` is set. To return to the previous behaviour, use `-V numbersections -V secnumdepth=0`.

With the Pandoc 2.5 template, one has to set `numbersections` to remove part/chapter numbering:

```markdown
pandoc -o chnum.pdf << EOT

---
documentclass: book
numbersections: true
secnumdepth: -2
---

# Test

# Test

EOT
```

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -277,7 +277,7 @@ $endif$
 $if(numbersections)$
 \setcounter{secnumdepth}{$if(secnumdepth)$$secnumdepth$$else$5$endif$}
 $else$
-\setcounter{secnumdepth}{0}
+\setcounter{secnumdepth}{-2}
 $endif$
 $if(beamer)$
 $else$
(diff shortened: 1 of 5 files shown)


jgm:
Can you explain how this works?  I don't understand what setting this to -2 does...

jgm:
Can you explain how this works?  I don't understand what setting this to -2 does...

adunning:
Setting `secnumdepth` is a bit counterintuitive:

`1` = Number `\section`
`0` = Number `\chapter`
`-1` = Number `\part`
`-2` = No numbering

jgm:
> Setting secnumdepth is a bit counterintuitive:

like so much in LaTeX...

PR #5238, opened 2019-01-20
odt writer: improve standard properties

agusmba:
Including the following properties:

* [x] generator : ~ "Pandoc/x.y.z"
* [x] ~description: use "abstract" metadata content, as I'm proposing for docx in [another PR](https://github.com/jgm/pandoc/pull/5215)~
* [x] subject: use "subject" metadata
* [x] keyword: standard keywords, as in docx
* [x] initial-creator: from authors
* [x] creation-date: actual creation date

Also fixes date (modified date)

Things missing

* [x] ~decide on the proper~ pandoc mapping for *dc:description*: \[ ~abstract~ | description | ~comments~ \]
* [x] do not remove "date" from the custom properties
* [x] review and update MANUAL if needed
* [ ] implement some tests (maybe out of scope if there is no support for golden testing odt yet)

### DIFF
--- a/src/Text/Pandoc/Writers/ODT.hs
+++ b/src/Text/Pandoc/Writers/ODT.hs
@@ -40,6 +40,7 @@ import Data.List (isPrefixOf, intercalate)
 import Data.Maybe (fromMaybe)
 import qualified Data.Map as Map
 import qualified Data.Text.Lazy as TL
+import Data.Time
 import System.FilePath (takeDirectory, takeExtension, (<.>))
 import Text.Pandoc.BCP47 (Lang (..), getLang, renderLang)
 import Text.Pandoc.Class (PandocMonad, report, toLang)
@@ -50,7 +51,7 @@ import Text.Pandoc.Logging
 import Text.Pandoc.MIME (extensionFromMimeType, getMimeType)
 import Text.Pandoc.Options (WrapOption (..), WriterOptions (..))
 import Text.Pandoc.Pretty
-import Text.Pandoc.Shared (stringify, normalizeDate)
+import Text.Pandoc.Shared (stringify, pandocVersion)
 import Text.Pandoc.Writers.Shared (lookupMetaString, fixDisplayMath)
 import Text.Pandoc.UTF8 (fromStringLazy, fromTextLazy, toStringLazy)
 import Text.Pandoc.Walk
@@ -83,6 +84,7 @@ pandocToODT :: PandocMonad m
 pandocToODT opts doc@(Pandoc meta _) = do
   let title = docTitle meta
   let authors = docAuthors meta
+  utctime <- P.getCurrentTime
   lang <- toLang (getLang opts meta)
   refArchive <-
        case writerReferenceDoc opts of
@@ -125,9 +127,14 @@ pandocToODT opts doc@(Pandoc meta _) = do
               )
          )
   let archive' = addEntryToArchive manifestEntry archive
+  -- create meta.xml
   let userDefinedMetaFields = [k | k <- Map.keys (unMeta meta)
-                              , k `notElem` ["title", "lang", "author", "date"]]
+                              , k `notElem` ["title", "lang", "author", "date"
+                                           , "abstract", "subject", "keywords"]]
   let escapedText = text . escapeStringForXML
+  let keywords = case lookupMeta "keywords" meta of
+                      Just (MetaList xs) -> map stringify xs
+                      _                  -> []
   let userDefinedMeta =
         map (\k -> inTags False "meta:user-defined"
               [ ("meta:name", escapeStringForXML k)
@@ -146,18 +153,25 @@ pandocToODT opts doc@(Pandoc meta _) = do
            ,("xmlns:ooo","http://openoffice.org/2004/office")
            ,("xmlns:grddl","http://www.w3.org/2003/g/data-view#")
            ,("office:version","1.2")] ( inTags True "office:meta" [] $
-                 ( metaTag "dc:title" (stringify title)
+                 ( metaTag "meta:generator" ("Pandoc/" ++ pandocVersion)
+                   $$
+                   metaTag "dc:title" (stringify title)
+                   $$  -- FIXME: abstract multiline loses newlines
+                   metaTag "dc:description" (lookupMetaString "abstract" meta)
+                   $$
+                   metaTag "dc:subject" (lookupMetaString "subject" meta)
+                   $$
+                   metaTag "meta:keyword" (intercalate ", " keywords)
                    $$
                    case lang of
                         Just l  -> metaTag "dc:language" (renderLang l)
                         Nothing -> empty
                    $$
-                   metaTag "dc:creator"
-                     (intercalate "; " (map stringify authors))
-                   $$
-                   maybe mempty
-                     (metaTag "dc:date")
-                       (normalizeDate (lookupMetaString "date" meta))
+                   (\d a -> metaTag "meta:initial-creator" a
+                         $$ metaTag "dc:creator" a
+                         $$ metaTag "meta:creation-date" d
+                         $$ metaTag "dc:date" d
+                   ) (formatTime defaultTimeLocale "%FT%XZ" utctime) (intercalate "; " (map stringify authors))
                    $$
                    vcat userDefinedMeta
                  )


agusmba:
Mmmm, am I missing something regarding tests for the odt writer? I see tests for the reader, but not for the writer.
Since the reader does not parse the metadata (yet), I guess we'd need golden testing?

insanetesterftw:
To allow a one-to-many conversion workflow (from markdown to pdf|epub|docx|html) please consider not hard code these values. Source properties should map to output's format specific standard.

It seems that Pandoc uses Dublic Core (http://dublincore.org/documents/dces/#subject) as its format for the metadata block and **Subject** is described as:

> Typically, the subject will be represented using keywords, key phrases, or classification codes. Recommended best practice is to use a controlled vocabulary.

docx also uses Dublin Core. So **Subject** cannot map to a **Description** or **Abstract**. I don't know why docx has both **Subject** and **Keywords** properties.

Consider this metadata block example from markdown file to be converted to epub:

---
title:
- type: main
  text: "Example Title"
creator:
- role: author
  text: "John Smith"
- role: author
  text: "Joan Smith"
lang: "en-US"
subject:
- "Keyword1"
- "Keyword2"
publisher: "Example Publisher"
source: http://www.example.com
rights: © 2019.
---

For epub the **subject** will map to keywords while from your example **subject** has other meaning in docx properties.

I tried several small word processors that support docx (Atlantis Word Processor, SoftMaker FreeOffice TextMaker and WPS Writer) and none of them shows **Abstract** as a default property.

Example from WPS Writer:
image

Perhaps **Description**  should map to **Comments** which is present in all of them.

It's clear that docx has a incorrect implementation of the Dublin Core standard. We should try to limited the damage and only allow a subset of properties to be transferred.

jgm:
> It seems that Pandoc uses Dublic Core

for EPUB only, really.  We have not made any efforts to conform to this for other formats.

insanetesterftw:
```
<cp:coreProperties xmlns:cp="http://schemas.openxmlformats.org/package/2006/metadata/core-properties" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:dcmitype="http://purl.org/dc/dcmitype/" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <cp:category>Examples Category</cp:category>
   <dcterms:created xsi:type="dcterms:W3CDTF">2019-01-21T15:42:00Z</dcterms:created>
   <dc:creator>John Smith, Joan Smith</dc:creator>
   <dc:description>Comments</dc:description>
   <cp:keywords>Keyword1, Keyword2</cp:keywords>
   <cp:lastModifiedBy>User</cp:lastModifiedBy>
   <dcterms:modified xsi:type="dcterms:W3CDTF">2019-01-21T17:08:11Z</dcterms:modified>
   <dc:subject>Examples Subject</dc:subject>
   <dc:title>Example Title</dc:title>
</cp:coreProperties>
```

This is an example of metadata from inside a docx and **Description** really maps to **Comments**.

I don't know if this is a mistake from these specific implementers of the docx standard or if it's the docx standard itself. Someone with Microsoft Office Word should report back.

My recommendation is that for docx we should limit the transfer of metadata to:
- **Title** mapped to dc:title
- **Creator** mapped to dc:creator
- **Description** mapped to dc:description (this will appear in the **Comments** field in all the GUIs that I tested)
- **Subject** mapped to cp:keywords.

We have to be realistic with the users expectations.

Pandoc conversion from docx to odt doesn't transfer metadata. Therefore I can't test how odt stores this information. I suspect that it has the same limitations as docx.

agusmba:
> So Subject cannot map to a Description or Abstract

Maybe I didn't express correctly what has been done.
Mappings from pandoc metadata to odt properties are one to one, using the same names, except in the case of  lang(pandoc)<->language(odt), and abstract<->description. In this latter case we could argue to keep them separate, however it seemed like a good decision to map them. I do not have a strong opinion either way (as long as both odt and docx behave the same way).

Regarding subject, I made a wrong initial proposal in the docx writer to map pandoc's subtitle metadata to docx's subject. It was pointed out that they were different things, so I kept them separate, both in the docx and odt writers.

If I understand what you're proposing is to define a set of "standard" pandoc metadata properties, that would be mapped to different output properties depending on the output format. I guess this would make for an interesting discussion.

insanetesterftw:
> > It seems that Pandoc uses Dublic Core
> 
> for EPUB only, really. We have not made any efforts to conform to this for other formats.

But docx and probably odt also have, at least, some basic Dublin Core support. I have added some recommendations of what we can try to support.

agusmba:
> But docx and probably odt also have, at least, some basic Dublin Core support. I have added some recommendations of what we can try to support.

Yes, on this PR and the one I have for docx, I have tried to make use of the Dublin Core properties present in the output format. This way when you open the docx/odt summary you'll see it filled out where pandoc has available metadata.

agusmba:
> Subject mapped to cp:keywords.

As you can see in your own example, keywords and subject are treated separately both in odt and docx, and are part of their standard properties. It would make sense to keep them separate in pandoc also.

insanetesterftw:
> > So Subject cannot map to a Description or Abstract
> 
> Maybe I didn't express correctly what has been done.
> Mappings from pandoc metadata to odt properties are one to one, using the same names, except in the case of lang(pandoc)<->language(odt), and abstract<->description. In this latter case we could argue to keep them separate, however it seemed like a good decision to map them. I do not have a strong opinion either way (as long as both odt and docx behave the same way).

But epub, docx, pdf and probably odt all support the same basic Dublin Core standard which is also what Pandoc supports for its metadata. It seems wise to correct this from the beginning.

Whatever the decision about metadata transfer between these formats in the case of _docx_ **Description** maps to what Word Processors call **Comments** as you can see from the example above.

insanetesterftw:
> > Subject mapped to cp:keywords.
> 
> As you can see in your own example, keywords and subject are treated separately both in odt and docx, and are part of their standard properties. It would make sense to keep them separate in pandoc also.

I don't know why both docx and odt choose to add another separate field called **Keywords** but **Subject** has the same semantic meaning according to Dublin Core standard.

As I said we should be realistic about user's expectations when transferring metadata from one format to another with same semantic meaning. As I recommend we should ignore **Subject** in both docx and odt and map **Subject** in Pandoc metadata to **Keywords** in both docx and odt metadata.

agusmba:
First of all, thank you for taking the time to comment on these PRs. You raise some interesting points like staying close to the output format internal properties with like-named pandoc metadata vs creating a set of "standard pandoc document" metadata that would be mapped to different properties depending on the output format. And all the range between these two options.

The current PR leans more towards the first option, with the main exception of abstract<->description (the lang case is a minor detail).

That said, 

> I don't know why both docx and odt choose to add another separate field called Keywords but Subject has the same semantic meaning according to Dublin Core standard.

In this case, though I might tend to agree with you on the lack of necessity for creating these two separate fields, they are actually not the same conceptually. While subject can be described with keywords, it can also include key phrases, classification codes, ... Even [Word documentation](https://support.office.com/en-us/article/view-or-change-the-properties-for-an-office-file-21d604c2-481e-4379-8e54-1dd4622c6b75#ID0EAABAAA=Newer_Versions_) treats them as separate things. For me, they are close, but not the same.

Of course I'm no document metadata expert, so I welcome all input from more knowledgeable users/developers.

mb21:
We've already started defining the semantics of some metadata fields. For example `lang` is (as in HTML5`), a string formatted in BCP 47 format. So expecting other metadata fields to conform to Dublin Core doesn't seem such a stretch. You already need to do that for EPUB export, so what if you want to use the same markdown file to also produce a docx?

agusmba:
> So expecting other metadata fields to conform to Dublin Core doesn't seem such a stretch. 

As far as using the standard document properties for docx/odt, that's part of what I've tried to do in these PRs. I didn't identify any additional Dublin Core property that imposed additional format restrictions (but I may have missed them)

> You already need to do that for EPUB export, so what if you want to use the same markdown file to also produce a docx?

I'd say we have to support this as best we can, of course. One thing I didn't do was to look at the EPUB writer for inspiration regarding the pandoc-metadata <-> docx/odt properties mapping (which I ended up simplifying as much as possible). If the EPUB implementation is supposed to be the *canon*, I could look at it to check for inconsistencies with what I'm proposing here.

agusmba:
From a quick look at the EPUB writer and some [online docs](https://www.oreilly.com/library/view/epub-3-best/9781449329129/ch01.html) it looks like the metadata management in EPUB is more complex/extensive than in docx/odt. Even though they share some dublin core properties, they are not directly compatible, as in EPUB most can be multivalued, and have extra metadata modifying said properties.

mb21:
> in EPUB most can be multivalued

I suppose you would just concatenate them in the docx writer, interspersed with newlines or space.

But of course it's fine to ignore properties that don't make sense for docx.

agusmba:
do we have a documented mapping between pandoc metadata and EPUB properties (besides looking at the sourcecode).

It would be nice to have this in the documentation.

mb21:
Like http://pandoc.org/MANUAL.html#epub-metadata ?

P.S. Again, I'm not familiar with docx metadata. Just think it's good to keep the same-markdown-file-for-epub-and-docx-use-case in mind ;)

agusmba:
> Like http://pandoc.org/MANUAL.html#epub-metadata ?
>
> P.S. Again, I'm not familiar with docx metadata. Just think it's good to keep the same-markdown-file-for-epub-and-docx-use-case in mind ;)

That's a start, thank you!

Seeing it brings to light the problem I was talking about.
Forget about adding support for any additional property, just taking into account pandoc's current state, for most writers, the pandoc title and author are completely different, and simpler. For instance, in the same manual page:

> The author variable in the default templates expects a simple list or string

In most cases, docx/odt properties are simpler and closer to pandoc current defaults (string, or stringified simple list), than to EPUB's complex structure.

I understand the desirability for having the same md able to write to different formats: epub, docx, odt, why not also html, latex, pdf?

I'm actually in favor of having this, not limited to docx/odt but expanded to most writers (the code for converting the complex EPUB structure to a String would be common to most of them). However I believe this would go a bit beyond an initial support of custom and core docx/odt properties. Could we split it and improve the metadata management pandoc-wide in a separate improvement issue?

agusmba:
Ok, I'll open another issue for that, linking to this discussion.

Regarding this and the docx PRs, would it help if I listed the current + proposed mappings between pandoc metadata and writer's properties so that we can agree or change said mappings? insanetesterftw and myself have different views on some of them, I agree with some of his recommendations.

agusmba:
The following table shows the mapping between different pandoc metadata and different writers properties. ***Emphasis*** on the new proposed additions

“-“ means it is treated as a custom property with the same name as pandoc metadata; any other root-level pandoc metadata is converted as a custom property of string value (or empty if it’s a complex metadata) to docx/odt output.

| Pandoc metadata       | DOCX property (GUI)             | ODT property (GUI)                               | EPUB property (incomplete list)         |
| --------------------- | ------------------------------- | ------------------------------------------------ | --------------------------------------- |
| title                 | dc:title (Title)                | dc:title (Title)                                 | title \[complex\], subtitle…            |
| author \[1..n\]       | dc:creator (Author)             | dc:creator (Author) ***+ meta:initial-creator*** |                                         |
| creator \[complex\]   | \-                              | \-                                               | dc:creator \[complex\], author, editor… |
| subtitle              | \-                              | \-                                               |                                         |
| keywords \[1..n\]     | cp:keywords (Keyword)           | ***cp:keywords (Keywords)***                     |                                         |
| subject               | ***dc:subject (Subject)***      | ***dc:subject (Subject)***                       | dc:subject                              |
| abstract              | ***dc:description (Comments)*** | ***dc:description (Comments)***                  |                                         |
| description           | *-*                             | *-*                                              | dc:description                          |
| lang                  | ***dc:language***               | dc:language                                      | dc:language                             |
| date                  | ***\****                        | ***\****                                         | dc:date (date of EPUB publication)      |
| *category*            | ***cp:category (Category)***    | *-*                                              |                                         |
| \[calc. at run time\] | dcterms:created (Created)       | ***meta:creation-date (Created)***               | dcterms:modified (Created)              |
| \[calc. at run time\] | dcterms:modified (Modified)     | ***dc:date (Modified)***                         |                                         |
| \[calc. at run time\] |                                 | ***generator***                                  |                                         |

Bear in mind that the EPUB column is there only to provide a comparison since it was mentioned in this thread, but it shouldn’t be treated as the golden rule for metadata mappings. Besides I’m not sure I understood the code correctly so the mapping here might not be real.

Points to consider:

  - We could debate whether choosing “**abstract**” for the cd:description was a good choice. Pandoc documentation states that abstract is a “document summary, included in LaTeX, ConTeXt, AsciiDoc, and Word docx”, but maybe it will be better to use either “description” (as EPUB does) or “comments” (as the field is shown in the GUI) as source metadata for this field (thanks @insanetesterftw \!).

  - The metadata ***(\*)*** “date” is currently purposefully removed from docx/odt properties. It’s not really necessary, since we calculate at runtime the creation/modification dates, and the “date” contents are entered directly in the body of the docx by pandoc, but it might be better to leave it as a simple custom property (we don’t really need to treat it as a specific case, and it could be useful to have it as a custom property to be used in headers/footers).

  - The metadata “**date**” is no longer used to reflect modification time of the odt document (it is calculated at pandoc’s run time).

  - I have removed “**references**” as a custom property in docx/odt, as this seems to be something internal to pandoc (it exists even if it’s not present in the yaml block)

Not for discussion but, interestingly, there seems to be some overlap between pandoc’s standard author and creator. I guess it was separated due to the complexities needed for EPUB, which could impact the author field for other writers (but there wasn’t such distinction in the title case).

agusmba:
The more I think about it the more I regret overloading "abstract" and mapping it to "dc:description". So the choice should be between mapping pandoc's "description" (as in EPUB) or "comments" (what you see in Word or OpenOffice GUI). I'm liking "description" better.

agusmba:
Closing in favor of https://github.com/jgm/pandoc/pull/5252

PR #5245, opened 2019-01-23
Man reader: fix typo.

leungbk:


jgm:
thanks!  always good to fix these broken windows.

PR #5249, opened 2019-01-23
Manual: Reorganize template variables

adunning:
Add additional headings to categorize variables, and alphabetize when there is large number; add more examples.


jgm:
Great

PR #5255, opened 2019-01-28
Org reader: add support for #+SELECT_TAGS.

leungbk:


src/Text/Pandoc/Readers/Org/DocumentTree.hs
```diff
@@ -161,17 +166,40 @@ headlineToBlocks hdln = do
...
+              childHdln <- headlineChildren hdln,
+              not $ any (`Set.member` orgStateExcludedTags st)
+                        (headlineTags childHdln)] }
+
+includeRootAndSelected :: Headline -> OrgParserState -> Headline
+includeRootAndSelected hdln st =
+  hdln { headlineChildren = mapMaybe (`includeAncestorsAndSelected` st)
+                                     (headlineChildren hdln)}
+
+checkForSelectedTags :: Headline -> OrgParserState -> Bool
```
tarleb: Not sure about this name, it doesn't seem clear to me if the result is True or False if a select tag is found. What do you think about `containsSelectTags`?

src/Text/Pandoc/Readers/Org/DocumentTree.hs
```diff
@@ -161,17 +166,40 @@ headlineToBlocks hdln = do
...
+checkForSelectedTags hdln st =
+  if any (`Set.member` orgStateSelectedTags st) (headlineTags hdln)
```
tarleb: Would `||` work here?

src/Text/Pandoc/Readers/Org/Blocks.hs
```diff
@@ -64,11 +68,22 @@ import qualified Text.Pandoc.Walk as Walk
 -- | Get a list of blocks.
 blockList :: PandocMonad m => OrgParser m [Block]
 blockList = do
-  headlines      <- documentTree blocks inline
+  fHeadlineTree  <- documentTree blocks inline
   st             <- getState
-  headlineBlocks <- headlineToBlocks $ runF headlines st
-  -- ignore first headline, it's the document's title
-  return . drop 1 . B.toList $ headlineBlocks
+  let headlineTree = runF fHeadlineTree st
```
tarleb: I haven't thought this through, but would it be possible to move this into `headlineToBlocks`? It would be nice if we could keep all tree-walking/modifications in that file.

leungbk:
@tarleb OK, I've addressed your comments.

PR #5276, opened 2019-02-05
Markdown Reader: yamlToMeta respects extensions

### LINKED ISSUE #5272: --metadata-file does not read LineBlock content (|)
```
lines: |
  | line1
  | line2
```

If this appears inside a standard YAML metadata frontmatter block, the content is (correctly) interpreted as:

```
[LineBlock [[Str "line1"],[Str "line2"]]]
```

However, if `lines` appears in a separate yaml file imported via `--metadata-file`, it is interpreted as:

```
[Plain [Str "|",Space,Str "line1",SoftBreak,Str "|",Space,Str "line2"]]
```

`--metadata-file` does not seem to be reading `LineBlock` content

mb21:
fixes #5272

Unfortunately, that's an API change. Meanwhile, we could use the default pandoc-markdown extensions as a stop-gap-measure instead:

    let opts = def{ readerExtensions = pandocExtensions }

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -211,10 +211,6 @@ convertWithOpts opts = do
                                          ("application/xml", jatsCSL)
                      return $ ("csl", jatsEncoded) : optMetadata opts
                    else return $ optMetadata opts
-    metadataFromFile <-
-      case optMetadataFile opts of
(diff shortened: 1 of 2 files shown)


jgm:
Yeah, why don't you submit another PR with the stopgap; I might want to do a point release before going to 2.7.

PR #5283, opened 2019-02-08
Revert "Docx reader: Tests for alternate document.xml"

jkr:
Don't merge: Just seeing if this builds on appveyor (or, at least, if it doesn't fail with the same error)


jkr:
Okay -- this passes. So the question is why *this* particular file doesn't work on windows.

I just looked at the `_rels/.rels` file and notice that the "/word/document2.xml" is offered as an absolute path (leading slash), while it isn't in other docs. I'll try dealing with that -- and doing another PR.

PR #5287, opened 2019-02-09
Use init.lua only for customizations

tarleb:
The file `init.lua` in pandoc's data directory is run as part of
pandoc's Lua initialization process. Previously, the `pandoc` module was
loaded in `init.lua`, and the structure for marshaling was set-up after.
This allowed simple patching of element marshaling, but made using
`init.lua` more difficult:

  - it encouraged mixing essential initialization with user-defined
    customization;

  - upstream changes to init.lua had to be merged manually;

  - accidentally breaking marshaling by removing required modules was
    possible;

Instead, all required modules are now loaded before calling `init.lua`.
The file can be used entirely for user customization. Patching
marshaling functions, while discouraged, is still possible via the
`debug` module.

### DIFF
--- a/data/init.lua
+++ b/data/init.lua
@@ -1,5 +1,3 @@
 -- This Lua script is run every time the Lua interpreter is started when running
 -- a Lua filter. It can be customized to load additional modules or to alter the
 -- default modules.
-
-pandoc = require 'pandoc'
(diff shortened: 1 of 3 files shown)


jgm:
Seems like a good idea to me!

PR #5312, opened 2019-02-16
Pandoc mediabag extension

tarleb:
This adds three functions to the `pandoc.mediabag` Lua module:

  - `delete`, which removes a single entry;
  - `empty`, which empties the whole bag;
  - `iter`, which simplifies iterating over the bag using Lua's generic `for` statement.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -2450,6 +2450,18 @@ e.g.:
 
     local mb = require 'pandoc.mediabag'
 
+### delete {#mediabag-delete}
+
+`delete (filepath)`
+
+Removes a single entry from the media bag.
+
+Parameters:
+
+`filepath`:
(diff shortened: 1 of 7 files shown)


tarleb:
The `iter` function may seem like a lot of code for little gain, but it should come in handy again when we allow (read-only) access to the media bag via `PANDOC_STATE`. The iterator function could then be re-used as the `__pairs` metamethod for media bags.

jgm:
I think I like `items` best, but I don't feel strongly about it.

tarleb:
Thanks, I tend to agree that `items` is a better name.

I'm delaying work on this until I found a better way of testing Lua functions.

tarleb:
I changed the function name from `iter` to `items` and added a few tests. Should be mergable now.

jgm:
great, merged!

Issue #5229, opened 2019-01-17
FB2 output: support for explicit header identifiers


KetzerX:
It would be good to have a possibility to set explicit identifier for header (and produce section element with that id), as now it is ignored. Although footnote produces correct section element with id.

## the pandoc version

pandoc 2.5

## the exact command line used

pandoc -s -o test.fb2 test.md

## the exact input used

```
# Section 1

Footnote[^f1]

[^f1]: my footnote

My link [link](#l1)

# Section 2 {#l1}

Section text
```

## the output received

(added newlines for readability)

```
<?xml version="1.0" encoding="UTF-8"?>
<FictionBook xmlns="http://www.gribuser.ru/xml/fictionbook/2.0" xmlns:l="http://www.w3.org/1999/xlink">
<description><title-info><genre>unrecognised</genre></title-info>
<document-info><program-used>pandoc</program-used></document-info>
</description>
<body><title><p /></title>
<section><title><p>Section 1</p></title>
<p>Footnote<a l:href="#n1" type="note"><sup>[1]</sup></a></p>
<p>My link <a l:href="#l1">link</a></p></section>
<section><title><p>Section 2</p></title><p>Section text</p></section></body>
<body name="notes">
<section id="n1"><title><p>1</p></title><p>my footnote</p></section>
</body></FictionBook>
```

As we can see id is set only to footnotes section: `<section id="n1">`

## the output you expected instead

```
<?xml version="1.0" encoding="UTF-8"?>
<FictionBook xmlns="http://www.gribuser.ru/xml/fictionbook/2.0" xmlns:l="http://www.w3.org/1999/xlink">
<description><title-info><genre>unrecognised</genre></title-info>
<document-info><program-used>pandoc</program-used></document-info>
</description>
<body><title><p /></title>
<section><title><p>Section 1</p></title>
<p>Footnote<a l:href="#n1" type="note"><sup>[1]</sup></a></p>
<p>My link <a l:href="#l1">link</a></p></section>
<section id="l1"><title><p>Section 2</p></title><p>Section text</p></section></body>
<body name="notes">
<section id="n1"><title><p>1</p></title><p>my footnote</p></section>
</body></FictionBook>
```
Id for section 2 should be set also `<section id="l1"><title><p>Section 2</p>...`

As we can see footnote produces correct section with id and link works. Explicit link creates only correct link element, but there are no target as section id is not set.

[...]

PR #5315, opened 2019-02-17
Add section identifiers support for FB2 writer

KetzerX:
This PR fixes a problem described in: https://github.com/jgm/pandoc/issues/5229

@jgm could you review the PR, please.

PR is ready for review.


src/Text/Pandoc/Writers/FB2.hs
```diff
@@ -177,20 +179,19 @@ docdate meta' = do
...
-
-renderSection :: PandocMonad m => Int -> ([Inline], [Block]) -> FBM m Content
-renderSection level (ttl, body) = do
-    title <- if null ttl
+    let elements = hierarchicalize blocks
+    cMapM (renderSection level) elements
+
+renderSection :: PandocMonad m =>  Int -> Sh.Element -> FBM m [Content]
+renderSection lvl   (Sh.Blk block) 
+  | lvl == 1 = fmap (list . el "section") (blockToXml block)
```
KetzerX: Wraps each 1-level element in section, but we should not have such elements in input file according to FB2 scheme. Original behaviour was to wrap several consequent elements in one section, but I don't know how to implement this.

jgm: Let me see if I understand the problem.  Suppose hierarchicalize gives you something like
```
Blk, Blk, Blk, Section, Section,...
```
You want to convert this to
```
Section, Section, Section,...
```
where the first Section includes the three initial blocks as content?

Try using the `break` function from Prelude.
```
let isSection Sec{} = True
    isSection _ = False
let (initialBlocks, secs) = break isSection elements
let elements' = Sec 1 [] nullAttr mempty initialBlocks
```
Would something like that work?

jgm: Small correction:  if `initialBlocks` is null, you wouldn't want to add the extra section.

KetzerX: > Let me see if I understand the problem. Suppose hierarchicalize gives you something like
> 
> ```
> Blk, Blk, Blk, Section, Section,...
> ```
> You want to convert this to
> 
> ```
> Section, Section, Section,...

 @jgm Yes, it is now how it works in current Pandoc version before hierarchilize. The question is: should we do such thing at all? According to FB2 specification we can't have `Blk, Blk, Blk, Section, Section,...` on first level, only sections.

jgm: I'm not sure you understood my proposal.  I understand that the output of hierarchicalize can be
Blk, Blk, Blk, Section, Section, and I also understand that FB2 wants everything in sections.  My proposal was to modify the list of Elements you get back from hierarchicalize, gathering the initial orphan Blks and inserting them into a new Section.

KetzerX: @jgm thanks, that's the answer I was waiting for! (which behaviour should be implemented). 
I looked at `break`, but it does the trick only for the first orphaned blocks, like `Blk Blk Section Blk Blk` -> `Secion Secion Blk Blk`. 
Maybe we should use folding to wrap every sequence of blocks? Or maybe there is more powerful function?

KetzerX: Thank you for help. All tests are passed now.

KetzerX:
PR is ready for review.

src/Text/Pandoc/Writers/FB2.hs
```diff
@@ -325,9 +317,10 @@ blockToXml (Para [Image atr alt (src,'f':'i':'g':':':tit)]) =
 blockToXml (Para ss) = (list . el "p") <$> cMapM toXml ss
 blockToXml (CodeBlock _ s) = return . spaceBeforeAfter .
                              map (el "p" . el "code") . lines $ s
-blockToXml b@(RawBlock _ _) = do
-  report $ BlockNotRendered b
-  return []
+blockToXml (RawBlock f str) = do
+  if f == Format "xml"
```
jgm: This should say `Format "fb2"`

jgm: also, indicate in your commit message that this change is being made (since it's independent of the section identifiers one).  Ideally this would be a separate commit, since it's a logically separate change, but if it's hard to separate out, don't worry about it.

KetzerX: Ok, I will make a separate commit.
Why it should be "fb2"? For example in epub with raw block we specify "html" format.

mb21: @KetzerX the `Format` should be one of the strings you can use in pandoc in the `--to` flag. The EPUB writer is somewhat special in that it just runs the HTML writer and then zips the output, that's why it's `html` there sometimes instead of `epub`.

KetzerX: Thanks. Done.

jgm:
Thanks @KetzerX for this improvement to pandoc!

PR #5324, opened 2019-02-20
MANUAL: reorder custom-styles section

mb21:
There is frequent confusion as to whether the `styles` extension needs to be enabled when converting from md to docx (and if so, whether on the markdown reader or docx writer), while actually neither the md reader nor the docx writer make use of that extension.

That's why I propose this patch which consists of:

- a small change in the `Extension: styles` section
- switching the "Input" and "Output" sections, as outputting docx is the far more common case.
- adding this sentence:

    > For docx output, you don't need to enable any extensions for custom styles to work.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2216,10 +2216,11 @@ output formats
 
 #### Extension: `styles` #### {#ext-styles}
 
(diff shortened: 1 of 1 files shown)


jgm:
Good changes!

PR #5346, opened 2019-03-03
Add tectonic as an option for --pdf-engine.

### LINKED ISSUE #5345: Add tectonic as a pdf engine
I was reminded by #3195 that I wanted this, hope it's not too late for 2.7.

https://tectonic-typesetting.github.io/en-US/

Tectonic is a fairly new engine that mirrors XeTeX, and although it's beta software in active development, it's pretty stable. It also doesn't write any support files by default, so no need to fluff around with .aux. It does its own steady state detection, so also only needs one run from Pandoc's point of view.

You might find if you try it out that it takes a very long time the first time -- it's auto-downloading support files. You should be able to see that with `--verbose`.

cormacrelf:
Closes #5345

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1200,9 +1200,9 @@ Options affecting specific writers {.options}
 
 :   Use the specified engine when producing PDF output.
     Valid values are `pdflatex`, `lualatex`, `xelatex`, `latexmk`,
-    `wkhtmltopdf`, `weasyprint`, `prince`, `context`, and `pdfroff`.
-    The default is `pdflatex`.  If the engine is not in your PATH,
-    the full path of the engine may be specified here.
+    `tectonic`, `wkhtmltopdf`, `weasyprint`, `prince`, `context`,
+    and `pdfroff`. The default is `pdflatex`.  If the engine is
+    not in your PATH, the full path of the engine may be specified here.
 
 `--pdf-engine-opt=`*STRING*
 
--- a/data/bash_completion.tpl
+++ b/data/bash_completion.tpl
@@ -34,7 +34,7 @@ _pandoc()
              return 0
              ;;
          --pdf-engine)
-             COMPREPLY=( $(compgen -W "pdflatex lualatex xelatex latexmk wkhtmltopdf weasyprint prince context pdfroff" -- ${cur}) )
+             COMPREPLY=( $(compgen -W "pdflatex lualatex xelatex latexmk tectonic wkhtmltopdf weasyprint prince context pdfroff" -- ${cur}) )
              return 0
              ;;
          --print-default-data-file)
--- a/src/Text/Pandoc/PDF.hs
+++ b/src/Text/Pandoc/PDF.hs
@@ -70,7 +70,7 @@ makePDF :: String              -- ^ pdf creator (pdflatex, lualatex, xelatex,
         -> WriterOptions       -- ^ options
         -> Pandoc              -- ^ document
         -> PandocIO (Either ByteString ByteString)
-makePDF program pdfargs writer opts doc = do
+makePDF program pdfargs writer opts doc =
   case takeBaseName program of
     "wkhtmltopdf" -> makeWithWkhtmltopdf program pdfargs writer opts doc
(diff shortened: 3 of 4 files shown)


src/Text/Pandoc/PDF.hs
```diff
@@ -70,8 +70,8 @@ makePDF :: String              -- ^ pdf creator (pdflatex, lualatex, xelatex,
         -> WriterOptions       -- ^ options
         -> Pandoc              -- ^ document
         -> PandocIO (Either ByteString ByteString)
-makePDF program pdfargs writer opts doc = do
-  case takeBaseName program of
+makePDF program pdfargs writer opts doc
+  = case takeBaseName program of
```
mb21: agreed that the `do` can be removed, but the `=` should stay on the upper line...

MANUAL.txt
```diff
@@ -1200,9 +1200,9 @@ Options affecting specific writers {.options}
 
 :   Use the specified engine when producing PDF output.
     Valid values are `pdflatex`, `lualatex`, `xelatex`, `latexmk`,
-    `wkhtmltopdf`, `weasyprint`, `prince`, `context`, and `pdfroff`.
-    The default is `pdflatex`.  If the engine is not in your PATH,
-    the full path of the engine may be specified here.
+    'tectonic', `wkhtmltopdf`, `weasyprint`, `prince`, `context`,
```
jgm: ought to be backticks around tectonic

changelog
```diff
@@ -33,6 +33,8 @@ pandoc (2.7)
     Note that you can use `--pdf-engine-opt=-outdir=bar` to specify
     a persistent temp directory.
 
+  * Add `tectonic` as an option for `--pdf-engine` (#5345).
```
jgm: Better not to change the changelog in the PR, since you don't know whether it will be included in this release.

data/bash_completion.tpl
```diff
@@ -30,7 +30,7 @@ _pandoc()
              return 0
              ;;
          --pdf-engine)
-             COMPREPLY=( $(compgen -W "pdflatex lualatex xelatex wkhtmltopdf weasyprint prince context pdfroff" -- ${cur}) )
+             COMPREPLY=( $(compgen -W "pdflatex lualatex xelatex latexmk tectonic wkhtmltopdf weasyprint prince context pdfroff" -- ${cur}) )
```
jgm: We really should have a separate PR or commit just adding `latexmk` to this, since that's an unrelated (but needed) change.

src/Text/Pandoc/PDF.hs
```diff
@@ -276,12 +274,11 @@ runTeXProgram verbosity program args runNumber numRuns tmpDir' source = do
...
     unless exists $ liftIO $ BS.writeFile file $ UTF8.fromText source
-    let programArgs =
-          if takeBaseName program == "latexmk"
-             then ["-interaction=batchmode", "-halt-on-error", "-pdf",
-                   "-quiet", "-outdir=" ++ tmpDir] ++ args ++ [file]
-             else ["-halt-on-error", "-interaction", "nonstopmode",
-                   "-output-directory", tmpDir] ++ args ++ [file]
+    let programArgs | takeBaseName program == "latexmk"  = ["-interaction=batchmode", "-halt-on-error", "-pdf",
+                                                            "-quiet", "-outdir=" ++ tmpDir] ++ args ++ [file]
+                    | takeBaseName program == "tectonic" = args ++ [file] -- outdir defaults to dir of input file
```
jgm: One question is what tectonic will put in the output dir.  The pandoc policy is not to leave any files on the file system that the user didn't explicitly request (e.g. with `-o myfile.pdf`).

cormacrelf: The answer to that one is nothing.

It has extra flags that optionally enable writing aux files into the output dir (which is, as the comment says, the temp directory in this case), but it never needs them, that’s only so you can speed up reprocessing with the Makefiles it also knows how to generate.

jgm:
Excellent, thanks!

PR #5359, opened 2019-03-10
LaTeX reader: `\underline` now outputs interior text directly rather …

pixlark:
…than ignoring it

Previously, any text in an `\underline` command in LaTeX would be ignored when being read by the LaTeX reader. Now, it simply pushes the text through without doing anything else to it. This is likely preferable to losing the text completely, even though it won't have the same formatting.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -866,7 +866,7 @@ inlineCommands = M.union inlineLanguageCommands $ M.fromList
   , ("slash", lit "/")
   , ("textbf", extractSpaces strong <$> tok)
   , ("textnormal", extractSpaces (spanWith ("",["nodecor"],[])) <$> tok)
-  , ("textunderline", inlines)
+  , ("underline", underlineSpan <$> tok)
   , ("ldots", lit "…")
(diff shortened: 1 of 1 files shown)


jgm:
Thanks!  I think we can go a bit farther  Some of the readers actually do parse underline and emit a span with class `underline`.  See the function `underlineSpan` in Text.Pandoc.Shared.  Until we have a  native underline element, that is our way of representing underlined text in pandoc.  So, why don't you emit that instead of just the inline contents?

Second point: https://alexwlchan.net/2017/10/latex-underlines/ lists four different ways to underline in LaTeX.  Your commit affects `\textunderline` -- which isn't one of these ways.  (What package is that from?)  But it would make sense to handle all of them, I think.

jgm:
The build failures aren't related to your change and should be fixed soon.

pixlark:
This new commit fixes the textunderline typo (long story as to how I typoed on such a tiny change) and implements two other underline modes from the article you linked, via the soul package and the ulem package. The fourth thing mentioned is more of a conglomerate of other LaTeX stuff and so seems a little out-of-scope for pandoc. At least, if it was to be implemented, it would have to be be someone who has more of an idea what they're doing :p

jgm:
Perfect, thanks!

PR #5363, opened 2019-03-13
 LaTeX template: Improve readability

adunning:
This improves the readability of the template in the following ways:

- Use `hidelinks` option for hyperref, which has the same effect as `pdfborder={0 0 0}`, but its purpose is clearer. (The template didn't use it before because the `pdfborder` option was previously used alongside link colouring, before the `colorlinks` option existed.)
- Remove hyperref `breaklinks` option: according to the documentation, hyperref sets this automatically as appropriate to the driver.
- Use a simpler conditional for Polyglossia/Babel.
- Format comments consistently.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -1,4 +1,5 @@
-\PassOptionsToPackage{unicode=true$for(hyperrefoptions)$,$hyperrefoptions$$endfor$}{hyperref} % options for packages loaded elsewhere
+% Options for packages loaded elsewhere
+\PassOptionsToPackage{unicode=true$for(hyperrefoptions)$,$hyperrefoptions$$endfor$}{hyperref}
 \PassOptionsToPackage{hyphens}{url}
 $if(colorlinks)$
 \PassOptionsToPackage{dvipsnames,svgnames*,x11names*}{xcolor}
@@ -46,7 +47,7 @@ $endif$
 $for(beameroption)$
 \setbeameroption{$beameroption$}
 $endfor$
-% Prevent slide breaks in the middle of a paragraph:
+% Prevent slide breaks in the middle of a paragraph
 \widowpenalties 1 10000
 \raggedbottom
 $if(section-titles)$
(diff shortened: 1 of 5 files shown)


jgm:
> According to the documentation, hyperref sets this automatically as appropriate to the driver.

Have you tested with very long links, to see if they still break?  I do remember that we added this for a reason, and it seemed to change behavior at the time...

jgm:
Aside from (possibly) that, these look like great changes!

adunning:
I did look into it: you added the `breaklinks` option in <https://github.com/jgm/pandoc/commit/6f76111e925bf918d6b444f358f421e368ca05b9>. Possibly hyperref has changed its behaviour since 2007, but I can't find evidence of this.

Using the following test file, I cannot find any difference with the `breaklinks` option enabled or disabled in any of pdfLaTeX, XeLaTeX, or LuaLaTeX (which is what hyperref documentation implies should happen):

```latex
\PassOptionsToPackage{hyphens}{url} % URLs will not break at hyphens without this or xurl
\documentclass{article}
\usepackage{hyperref}
% \hypersetup{breaklinks=true} % should be set automatically by hyperref
% \usepackage{xurl} % uncomment for correct output
\urlstyle{same}

\begin{document}

\url{https://pandoc.org/jshdlfkjshlkdjfhsalkdfjcnlskjfnhlsdkjfnclskjfchsalkjcfalnskdjcfhdklnsjcfhlnsdkjfcalnskfjlnskfjhnslkjdfadklnsjfchdlnskjfchlnskjfhdlnskfjchsaldknfjchdslcfknjashlkcfjnslafnksjdhcfaklnsjcfhlsadknjcfh}

\url{https://pandoc.org/jshdlfkjs/hlkdjfhsalkdfjcnlskjfnhlsdkjfnclskjfc/hsalkjcf-alnskdjcfhdklns/jcfhlnsdkjfcalnskfjlnskfjhnslkjdfad/klnsjfchdlnsk-jfchlnskj/fhdlnskfjchsaldknfjchdslcfknjashlkc-fjnslafnksjdhcfaklnsj/cfhlsadknjcfh}

\end{document}
```

jgm:
Great, thanks for looking into it.

PR #5370, opened 2019-03-16
default.latex: thanks should be appended to author not to title

karland:


jgm:
I've seen thanks footnotes on both title and author, so this may be a stylistic preference.
@adunning do you have any thoughts on this?

adunning:
If there is only one `\thanks`, I think it should stay on the title, because we would have to provide a way to add `\thanks` after *each* author, and not just the first or last. (It tends to be used in that position to provide affiliations or contact information.) But one can already place a footnote anywhere one likes in the title block:

```shell
pandoc -o thanks.pdf << EOT

---
title: Thanks[^1]
author:
- Plato[^2]
- Aristotle[^3]
date: 2019[^4]
---

[^1]: test1
[^2]: test2
[^3]: test3
[^4]: test4

EOT
```

It renders in the same way whether one uses `\footnote` or `\thanks`. It appears there are some obsolete document classes that will only accept `\thanks` within `\maketitle`, but the information I could find indicates that these have not been updated since LaTeX 2.09.

karland:
+ In my opinion, one of the big achievements of pandoc is that it works most of the cases without much configuration.

+ LaTeX and KOMA classes both put the thanks behind the author(s), so I think pandoc should not change this. For me this seems also more logical because the thank you note belongs to the people saying thank you. Personally I had not seen thanks on the title before, this is why I tripped over it, but you may be right and it is a matter of taste / preference. My suggestion was just to bring pandoc closer to the standard.

+ I suppose the default case is that the author(s) want(s) to thank somebody for a contribution. A very special case is that each author wants to thank somebody individually. For this case, @adunning has suggested a very nice solution of which I was not aware.

adunning:
The `\thanks` command doesn't have a semantic meaning, in spite of its name – it's just the normal method for adding footnotes within `\maketitle` elements, from before `\footnote` worked there directly. It really shouldn't have been added as a variable to the template in the first place, given that no other writer supports it.

The LaTeX writer could be configured to use `\thanks` instead of `\footnote` within the title commands – I believe that using it can still avoid problems if one is using `\maketitle` more than once in a document. But that's not possible with Pandoc in any event, so probably not worth bothering with it.

karland:
@adunning I am not sure what you are suggesting and I was not aware that `\thanks` is supposed to not have a semantic meaning.

This PR suggests just to change the place of the footnote symbol and append it to the author(s) which is probably the standard. See also:

[http://texdoc.net/texmf-dist/doc/latex/latex2e-help-texinfo/latex2e.pdf](http://texdoc.net/texmf-dist/doc/latex/latex2e-help-texinfo/latex2e.pdf) on p.37.

@jgm If you do not want to change the default place of the footnote symbol, then I am happy to close this PR. This PR now documents an alternative to setting thank-you-notes in the title, which for me is a workable and actually pretty flexible solution which also does not require a change of  `default.latex`.

adunning:
@karland I am suggesting that the `thanks` variable could be removed entirely from the template without losing any functionality (except that it has been around since Pandoc 1.16, and I do not have a sense of how widely used it might be).

karland:
@adunning, thanks for the clarification. I am not sure how to move on from here. Removing `\thanks` alltogether sounds like quite a big change to me. If this is the direction @jgm wants to go, I can close this PR because it becomes obsolete then.

the-solipsist:
My 2 cents: If functionality A (`\thanks`) is equivalent to functionality B (footnotes or `\footnote`), but A will work only with one export type (LaTeX) while B will work with many, then A should be dropped and people should be encouraged to use B.  Generally, people will expect to have broadly similar-looking documents regardless of the format they choose to export into.  This would be the case if they use B, but not if they use A.  Allowing for the functionality to be expressed through A will only lead to people getting confused as to why the acknowledgement footnote shows up in the PDF but not in the HTML file or the ODT file.  This can be avoided if A (i.e., `\thanks`) is removed.

Issue #5179, opened 2018-12-26
linestretch option ignored when lang=ko is set


esnahn:
![snipaste_2018-12-26_22-02-50](https://user-images.githubusercontent.com/1122053/50446764-09f0e680-095a-11e9-9b61-4d99253b7252.png)

Pandoc version:

```
pandoc.exe 2.5
Compiled with pandoc-types 1.17.5.4, texmath 0.11.1.2, skylighting 0.7.4
```

With the command:

```
pandoc "$(STYLEDIR)/template_test.yaml" "$(INPUTDIR)"/14*.md \
	-o "$(OUTPUTDIR)/$(STDNO)-$(FULLNAME)-Thesis-test.pdf" \
	--pdf-engine=xelatex
```

with file template_test.yaml as below:

```
---
lang: ko # BCP 47 tag for Korean
# this option changes everything back to default...

linestretch: 3
# indent: true

fontfamilies: # for xetex and polyglossia
  - name: \hangulfont
    font: NanumMyeongjo
    options:
      - Script = Hangul
      - Path = style/fonts/
      - Extension = .otf
      - BoldFont = *Bold 
  - name: \hangulfontsf
    font: NanumBarunGothic
    options:
      - Script = Hangul
      - Path = style/fonts/
      - Extension = .otf
      - BoldFont = *Bold 
  - name: \hangulfonttt
    font: D2Coding
    options:
      - Path = style/fonts/
      - Extension = .ttf
      - BoldFont = *Bold 
---

```

`linestretch` option is respected only when `lang` is not set `ko`.
lang=en and lang=it have no problem.

jgm:
I suspect that if you look at the intermediate tex
file that pandoc generates (which you can see by
adding --verbose), you'll see that linestretch
is not ignored by pandoc; an appropriate command
like `\setstretch{3}` is added.  Probably this doesn't
work with the Korean font for some reason.  You may
need to ask on a LaTeX forum to clear this up, posting
a minimal tex example.

In any case I don't think it's a problem with pandoc,
specifically.

esnahn:
It is definitely not the font because it happens too with korean font and text.

![snipaste_2018-12-27_14-13-45](https://user-images.githubusercontent.com/1122053/50466593-d5c10880-09e1-11e9-8b29-040cd9d60c4e.png)

with addition of [main/sans/mono]font of course, otherwise it wouldn't run because of missing characters:

```
---
lang: ko # BCP 47 tag for Korean
[...]

PR #5371, opened 2019-03-16
default.latex: Do not apply linestretch to title

karland:
See: Kohm, M. (2018). KOMA-Script. p45.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -391,7 +391,13 @@ $if(title)$
 $if(beamer)$
 \frame{\titlepage}
 $else$
+$if(linestretch)$
+\begin{spacing}{1}
 \maketitle
+\end{spacing}
+$else$
+\maketitle
+$endif$
 $endif$
 $if(abstract)$
 \begin{abstract}


jgm:
This is a fairly verbose change in the generated latex.

See #5179, which proposes putting the setspace command after `\begin{document}` in order to handle a mysterious bug that arises with asian languages.  If we put the setspace command after `\maketitle`, then I believe it would handle this issue too, in a much less verbose way.

karland:
Well, I was working through the KOMA-Script book and came across this issue. Kohm suggests the solution that is in this PR, so I just took it over. I was not so concerned about verbosity, but fair enough. 

I understand the issue because the generation of the title page is quite some piece of code and it seems to be negatively impacted by a line spacing larger than 1.

adunning:
In testing this further, I've realized that not only `\maketitle` but also `\tableofcontents`, `\listoffigures`, and so forth are also affected negatively by changing the line spacing. I've added a new PR at <https://github.com/jgm/pandoc/pull/5378> to allow for testing this.

adunning:
I need to test this theory, but logically, using `\setstretch` outside the preamble will probably cause the height of the text block to come out unrounded to the number of lines, which can leave extra space between paragraphs. This isn't terrifically noticeable unless one is using indented paragraphs, and I doubt whether many people are using the `heightrounded` option in geometry in any case. But it might be better to manually set areas to single spacing, as this proposes. We need to balance the tradeoffs between partially dealing with #5179 and simplicity of code on the one hand, and adding an extra environment on the other hand.

Another consideration: should the abstract be single-spaced, or should single spacing be applied only to `\maketitle` and the various lists in the front matter?

karland:
@adunning Thanks for working on this.

jgm:
The issue is fixed by #5378.
If the fix in #5378 turns out to be a bad idea, we can revisit.

PR #5372, opened 2019-03-16
default.latex: Offer additional title page options from KOMA classes

karland:
[http://texdoc.net/texmf-dist/doc/latex/koma-script/scrguien.pdf](http://texdoc.net/texmf-dist/doc/latex/koma-script/scrguien.pdf)

See p.66ff

This is important for all types of documents that need more information in the title, eg. Dissertations, Bachelor Thesis, Books, etc.

e.g.

```LATEX
\documentclass{scrartcl}
\usepackage[english]{babel}
\begin{document}
\titlehead{{\LargeUnseen University\hfillSS~2019\\}
Higher Analytical Institute\\
Mythological Rd\\
34567 Etherworld}
\subject{Dissertation}
\title{Digital space simulation with the DSP\,56004}
\subtitle{Short but sweet?}
\author{Fuzzy George}
\date{30. February 2019}
\publishers{Adviser Prof. John Eccentric Doe}
\dedication{To my friend}
\maketitle
\end{document}
```

This PR is generalizing PR [#4536](https://github.com/jgm/pandoc/pull/4536)

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -945,7 +945,7 @@ Options affecting specific writers {.options}
 
 :   Use the [`listings`] package for LaTeX code blocks. The package
     does not support multi-byte encoding for source code. To handle UTF-8
-    you would need to use a custom template. This issue is fully 
+    you would need to use a custom template. This issue is fully
     documented here: [Encoding issue with the listings package].
 
 `-i`, `--incremental`
@@ -1533,7 +1533,7 @@ These variables change the appearance of PDF slides using [`beamer`].
 :   enables "title pages" for new sections (default is true)
 
 `theme`, `colortheme`, `fonttheme`, `innertheme`, `outertheme`
-:   beamer themes: 
+:   beamer themes
 
 `themeoptions`
 :   options for LaTeX beamer themes (a list).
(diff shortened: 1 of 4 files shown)


adunning:
Variables in Pandoc that have content (as opposed to just formatting) are meant to work across all supported formats as much as possible. Adding all these as variables across all the writers would be a lot of work, especially since it's not immediately obvious how they would translate in many major formats.

If this were to be implemented, a better way to do so would be to add the `scrextend` package rather than make these commands specific to KOMA-Script classes. This would have the added benefit of allowing arbitrary font sizes (which is a periodic complaint from users) – but `scrextend` has compatibility problems with some other packages, especially if one isn't using a recent TeX Live distribution, and this has always stopped me from proposing its use in the past.

You can already achieve what you're looking for by using `header-includes`:

```shell
pandoc -o test.pdf << EOT

---
documentclass: scrbook
title: Digital space simulation with the DSP 56004
subtitle: Short but sweet?
author: Fuzzy George
date: 30 February 2019
header-includes:
- \titlehead{Unseen University}
- \subject{Dissertation}
- \publishers{Adviser Prof. John Eccentric Doe}
- \dedication{To my friend}
---

# A test

Test

EOT
```

karland:
@adunning  Yes indeed, this is a nice solution. I was aware of `header-includes` but your solution did not occur to me. We now even have it documented by this PR. Thanks.

PR #5373, opened 2019-03-16
default.latex: Define subtitle only for non-KOMA classes

karland:
KOMA has already a definition of subtitle.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -358,11 +358,12 @@ $endif$
 $if(subtitle)$
 $if(beamer)$
 $else$
-\usepackage{etoolbox}
 \makeatletter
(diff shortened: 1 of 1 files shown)


jgm:
Given that `\providecommand` is used, it's harmless to include this code for KOMA classes, right?
It would be one thing if we could exclude it from the generated latex code -- but that would require testing for KOMA classes in the writer and passing a variable to the template, and I'm not sure it's worth the effort.  This solution doesn't reduce code size of generated latex -- in fact, it increases it -- and doesn't affect the output.

adunning:
I actually had this originally when adding the `subtitle` section, but confirmed in testing that there's no point – KOMA-Script already ignores it due to `\providecommand`.

karland:
Good point so I close this issue.

PR #5374, opened 2019-03-16
latex.default: add fontfamilydefault option.

karland:
This allows the font to be changed for the whole document.

+ [https://www.latex-project.org/help/documentation/fntguide.pdf](https://www.latex-project.org/help/documentation/fntguide.pdf)
    See: p8.

+ [https://tex.stackexchange.com/questions/58833/effect-of-changing-fontfamily-and-familydefault](https://tex.stackexchange.com/questions/58833/effect-of-changing-fontfamily-and-familydefault)

+ [https://texwelt.de/wissen/fragen/785/wie-stelle-ich-alle-schrift-in-meinem-dokument-auf-serifenlos](https://texwelt.de/wissen/fragen/785/wie-stelle-ich-alle-schrift-in-meinem-dokument-auf-serifenlos)
    (in German)

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -945,7 +945,7 @@ Options affecting specific writers {.options}
 
 :   Use the [`listings`] package for LaTeX code blocks. The package
     does not support multi-byte encoding for source code. To handle UTF-8
-    you would need to use a custom template. This issue is fully 
+    you would need to use a custom template. This issue is fully
     documented here: [Encoding issue with the listings package].
 
 `-i`, `--incremental`
@@ -1533,7 +1533,7 @@ These variables change the appearance of PDF slides using [`beamer`].
 :   enables "title pages" for new sections (default is true)
 
 `theme`, `colortheme`, `fonttheme`, `innertheme`, `outertheme`
-:   beamer themes: 
+:   beamer themes:
 
 `themeoptions`
 :   options for LaTeX beamer themes (a list).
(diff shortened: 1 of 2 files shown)


jgm:
@adunning - could I get your feedback on this proposal?

adunning:
@karland Would you mind illustrating the functionality that this would add? This is more easily specified through fontspec and/or KOMA-Script, though you can always add a line such as this through the `header-includes` variable.

karland:
pandoc is great! In the context of LaTeX is amongst other things so great because people can go from Markdown to PDF (via LaTeX) without knowing very little about LaTeX. 

In addition, KOMA classes are probably the standard in the meantime. However, other then standard LaTeX, titles and section headers are in sans serif whereas text is in roman. Even though there are good typographical reasons for it as described by the author of KOMA-script sometimes articles or books need to use the same font type. When I was researching this issue, I found that the **easy option** to achieve this was missing in pandoc. Of course there might be other possibilities to get this done, however, I feel, fonts in LaTeX already require some deeper knowledge, so I felt this option would make it easier for the user and provide some guidance.

adunning:
If your aim is to use a serifed font for the KOMA-Script headings, it's much more reliable to add its own command for taking care of this using `header-includes` (plus the same mechanism allows you to achieve many other things):

```shell
pandoc -o test.pdf << EOT

---
documentclass: scrartcl
title: Test
header-includes:
- \setkomafont{disposition}{\normalfont}
---

# A test

Test

EOT
```

karland:
@adunning Thanks.

Issue #4322, opened 2018-01-29
Wrong image size for pdf images in docx


timtroendle:
PDF images in a markdown -> docx workflow using `pandoc 1.19.2` end up with the wrong image size; unfortunately even with wrong aspect ratio. This issue seems closely related to #2720 and #3798.

Assuming there is a PNG and a PDF version of the same image, this simple Markdown file is enough to reproduce the issue:

```
![PNG for reference.](test.png)

![PDF version of the image.](test.pdf)
```

Running this with `pandoc test.md -t docx -o test.docx` shows a Word document, in which the PNG version has a correct aspect ratio and scale, but the PDF version is distorted. (Actually the PDF version has the default image size of 300/200 defined in [ImageSize.hs](https://github.com/jgm/pandoc/blob/b7d8930dc097defcae754120a4fa0d6727c9b265/src/Text/Pandoc/ImageSize.hs#L103).)

I assume this is due to the fact that the [image size detection is not yet implemented for PDF images](https://github.com/jgm/pandoc/blob/b7d8930dc097defcae754120a4fa0d6727c9b265/src/Text/Pandoc/ImageSize.hs#L141).

As I couldn't find a workaround so far, I'd be happy for any hint in improving the current situation which requires manual correction of all image sizes.

jgm:
>   I assume this is due to the fact that the [4]image size detection is
>   not yet implemented for PDF images.

No doubt that's true.  See #2350.

I tried converting lalune.jpg to pdf using ImageMagick, and
looking at the pdf in a text editor.  It begins as follows:

```
%PDF-1.3 
1 0 obj
<<
/Pages 2 0 R
/Type /Catalog
>>
endobj
2 0 obj
<<
/Type /Pages
/Kids [ 3 0 R ]
/Count 1
>>
endobj
3 0 obj
<<
/Type /Page
/Parent 2 0 R
/Resources <<
/XObject << /Im0 8 0 R >>
/ProcSet 6 0 R >>
/MediaBox [0 0 150 150]
/CropBox [0 0 150 150]
/Contents 4 0 R
/Thumb 11 0 R
>>
endobj
4 0 obj
<<
/Length 5 0 R
>>
stream
q
150 0 0 150 0 0 cm
/Im0 Do
Q

endstream
endobj
5 0 obj
31
endobj
6 0 obj
[ /PDF /Text /ImageC ]
endobj
7 0 obj
<<
>>
endobj
8 0 obj
<<
/Type /XObject
[...]

PR #5383, opened 2019-03-21
Improve pdfSize in ImageSize by ignoring all whitespace in /MediaBox command

richarddavis:
This is a small improvement to commit [9573141](https://github.com/jgm/pandoc/commit/957314143faec08b4687822557dac6ac32216cb9). That commit fixed PDF image sizes for PDFs created with Illustrator, but it was not working with PDFs created using Figma. The fix for this was to ignore all of the whitespace in the /MediaBox line. 

This fix ignores all whitespace in the PDF /MediaBox line so that a wider range of PDF sizes can be read. This improves fix to #4322.

### DIFF
--- a/src/Text/Pandoc/ImageSize.hs
+++ b/src/Text/Pandoc/ImageSize.hs
@@ -278,16 +278,16 @@ pPdfSize = do
   A.skipWhile (/='/')
   A.char8 '/'
   (do A.string "MediaBox"
-      A.skipWhile (==' ')
+      A.skipSpace
       A.char8 '['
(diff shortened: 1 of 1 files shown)


jgm:
Excellent.  To cover all the bases, maybe `skipWhile isSpace`?  I don't know if tabs and newlines are allowed in this context in PDFs, but it couldn't hurt.

richarddavis:
Good thinking! I pushed a commit with `skipSpace` and that seems to be working correctly.

jgm:
Great!

Issue #1658, opened 2014-09-30
[Markdown -> PDF] - Header level 4+ is misplaced when followed by a table


TiBeN:
Using pandoc from official ubuntu repos, v1.12.2.1,

When a header level 4 or higher is followed by a table, it is placed after the table. 
Unless i misunderstund a feature, i suspect a bug.

Here's an example and it's pdf output:

```markdown

# Title 1

  Right     Left     Center     Default 
-------     ------ ----------   -------
     12     12        12            12
                                                                    
## title 2

  Right     Left     Center     Default 
-------     ------ ----------   ------- 
     12     12        12            12 

### Title 3

  Right     Left     Center     Default
-------     ------ ----------   ------- 
     12     12        12       12

#### Title 4

  Right     Left     Center     Default
-------     ------ ----------   -------
     12     12        12            12

#### Title 5
       
  Right     Left     Center     Default 
-------     ------ ----------   -------
     12     12        12            12 

```

![pandoc-bug](https://cloud.githubusercontent.com/assets/910115/4457062/3bd3c024-4892-11e4-9f9b-75303c8f7c53.jpg)

Forgot to mention the command line used: 
```bash
pandoc test.md -o test.pdf
```

lierdakil:
I'm thinking on how this should be addressed, and apparently there are three options:

1. Insert empty paragraph before all tables
2. Insert empty paragraphs after all lv. 4-5 headings
3. Find where lv. 4-5 headings are followed by a table and insert `RawBlock (Format "latex") "\\mbox{}"` there.

Thoughts?

jgm:
@lierdakil
Could this be fixed by redefining `\paragraph` in the default
template, so it works more like `\subsection`?

There's reason to do that anyway: people are always complaining
about how the level-4 and level-5 headings don't have line breaks
after them in PDF output, and this behavior isn't really consistent
with how other writers work.

lierdakil:
@jgm, that would be a variation of (2). Something like this could work:

```latex
\let\oldparagraph\paragraph
\renewcommand{\paragraph}[1]{\oldparagraph{#1}\mbox{}\\}
```

lierdakil:
Note that for lv.5 headings, we need to give `\subparagraph` the same treatment.

lierdakil:
I'm a little concerned about labels however, since they would point to paragraph *after* line break.

lierdakil:
[...]

PR #5421, opened 2019-04-02
LaTeX template: Ensure correct heading/table order

### LINKED ISSUE #5365: Improving appearance of subparagraph with indentation
The `\paragraph` rewrite in the LaTeX template gives an odd appearance when indentation is enabled. Would you mind if I moved it so that these lines are only run when the `indent` option is false? Or would you prefer that I come up with a way of making this look right with indentation?

Here's an example showing an indented document, before and after the (sub)paragraph lines:

Screen Shot 2019-03-13 at 21 21 52

```latex
\documentclass{article}
\usepackage{mwe}

\begin{document}

\paragraph{test}

\blindtext

\subparagraph{test}

\blindtext

\ifx\paragraph\undefined\else
  \let\oldparagraph\paragraph
  \renewcommand{\paragraph}[1]{\oldparagraph{#1}\mbox{}}
\fi
\ifx\subparagraph\undefined\else
  \let\oldsubparagraph\subparagraph
  \renewcommand{\subparagraph}[1]{\oldsubparagraph{#1}\mbox{}}
\fi

\paragraph{test}

\blindtext

\subparagraph{test}

\blindtext

\end{document}
```

adunning:
Improve the workaround for #1658, adapting a solution by @u-fischer in <https://github.com/latex3/latex2e/issues/131> that works whether or not the `indent` variable is enabled. Rename the previous approach under the `subparagraph` variable as `block-headings`. This is no longer enabled by default, as it gives headings 3–5 an identical appearance. Provide an example in the manual of reformatting the appearance of headings more thoroughly using KOMA-Script. Closes #5365.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -436,7 +436,7 @@ General options {.options}
 [XHTML]: http://www.w3.org/TR/xhtml1/
 [LaTeX]: http://latex-project.org
 [`beamer`]: https://ctan.org/pkg/beamer
-[Beamer User's Guide]: http://ctan.math.utah.edu/ctan/tex-archive/macros/latex/contrib/beamer/doc/beameruserguide.pdf
+[Beamer User's Guide]: http://mirrors.ctan.org/macros/latex/contrib/beamer/doc/beameruserguide.pdf
 [ConTeXt]: http://www.contextgarden.net/
 [Rich Text Format]: http://en.wikipedia.org/wiki/Rich_Text_Format
 [DocBook]: http://docbook.org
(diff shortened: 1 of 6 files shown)


adunning:
Note that it is not necessary to disable the workaround when `block-headings` is enabled; these two examples provide identical results:

```shell
pandoc -o test.pdf << EOT

#### paragraph

-----
aaa
-----

##### subparagraph

-----
bbb
-----

EOT
```

and

```shell
pandoc -o test2.pdf -V block-headings << EOT

#### paragraph

-----
aaa
-----

##### subparagraph

-----
bbb
-----

EOT
```

PR #5426, opened 2019-04-03
Manual: Add syntax highlighting

adunning:
Use fenced code blocks to mark language names for code examples.

(Is there a highlighting language for the templates?)

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -368,7 +368,9 @@ General options {.options}
 :   Generate a bash completion script.  To enable bash completion
     with pandoc, add this to your `.bashrc`:
 
-        eval "$(pandoc --bash-completion)"
+    ~~~bash
+    eval "$(pandoc --bash-completion)"
+    ~~~
 
 `--verbose`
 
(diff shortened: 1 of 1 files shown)


mb21:
Nice! But aren't we usually using ` ``` ` instead of ` ~~~ ` in the MANUAL?

adunning:
I was doing that on the basis that the tildes are listed as the primary method of fenced code blocks, but I've no preference in the matter.

mb21:
me neither :)

jgm:
I have mixed feelings about this.  I don't really think syntax highlighting is is so important for these short samples, and I think adding the fences detracts from the plain-source readability of the manual.

If we did do it, I think we should probably use backticks, because that's compatible with more implementations.

adunning:
Possibly use syntax highlighting only for extended non-Markdown examples? I think it improves readability most with the examples in YAML, CSS, Lua, etc.

jgm:
I think I'm going to pass on this one.  The vast majority of code samples are markdown.  And I think the aesthetics look funny if some of the code is highlighted and other not.  We already do use highlighting for the lua filters documentation, where I think it makes sense.

PR #5445, opened 2019-04-13
Improved lua tikz filter

xcthulhu:
Recently this has been the version of this lua filter I have been using for blogging.  There are three changes:

1. It only processes elements which begin with `\begin{tikzpicture}`
2. It uses pdf2svg instead of imagemagick to preserve fidelity
3. The images produced have transparent backgrounds


jgm:
Thanks, these look like good improvements.

PR #5448, opened 2019-04-15
Remove reference to head.min.js

gitlab-winnie:
This file has been removed in 3.8.0: https://github.com/hakimel/reveal.js/commit/29b0e86089eb3ec0d4bb5811c9b723dfcf36703c

### DIFF
--- a/data/templates/default.revealjs
+++ b/data/templates/default.revealjs
@@ -89,7 +89,6 @@ $body$
     </div>
   </div>
 
-  <script src="$revealjs-url$/lib/js/head.min.js"></script>
   <script src="$revealjs-url$/js/reveal.js"></script>
 
   <script>


jgm:
I guess one question to ask is whether some people might be using pandoc's reveal.js output with older versions of reveal.js.  This might break things for them.  We could just target the most recent version, but my guess is that the missing link doesn't break things with the most recent version, but just causes an unnecessary request?

gitlab-winnie:
> but just causes an unnecessary request?

@jgm Yes, and therefore unnecessary and potentially confusing output in the browser console:

```
The resource from “.../reveal.js/lib/js/head.min.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
Loading failed for the <script> with source “.../reveal.js/lib/js/head.min.js”.
```

(you can seee it happen on https://winnie-slides.gitlab.io/2019/wie-macht-gitlab-gitlab-de/)

Could we introduce a variable and then include the script conditionally? (defaulting to include, potentially changing the default in a future version) Something like:

```html
$if(include-headjs)$
<script src="$revealjs-url$/lib/js/head.min.js"></script>
$endif$
```

jgm:
I wonder if we can query the reveal.js version programatically and load head.min.js depending on the version?

gitlab-winnie:
> query the reveal.js version programatically and load head.min.js depending on the version

@jgm Yes, that is possible but we would also need to wait for `head.min.js` to be loaded before we can do `Reveal.initialize()` on older versions.

jgm:
OK, I guess the easiest thing  is just to tell people that they need an up-to-date reveal.js.

Issue #5146, opened 2018-12-13
HTML5 validation warning on "style"


crystalfp:
Validating Pandoc generated HTML5 page elicit warning during validation.

The source (bug.md):
```
---
pagetitle: Support
author: Mario Valle
lang: en
...

# Support
Support for bla bla.
```
The build command (bug.sh):
```
pandoc -s -c style.css -o bug.html -f markdown+smart -t html bug.md
```
The validation generates a warning: "The type attribute for the style element is not needed and should be omitted." Also not sure that defining the namespace and xml:lang in the header is needed for HTML5
```
<html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
```

Is generating the template `pandoc -D html5`, modifying and using it the only workaround?
Thanks for clarifying!
mario

pandoc.exe 2.4 (32 bits) on windows10 64bits

mb21:
I think I agree. The `type="text/css"` is not recommended for HTML5 and [apparently not needed for XHTML Polyglot either](https://www.w3.org/TR/html-polyglot/#h5_safe-text-content). @JohnLukeBentley?

The namespace and xml:lang stuff makes the file also a valid XHTML document, which is nice for some use-cases. See https://github.com/jgm/pandoc/pull/3473

JohnLukeBentley:
@mb21 you've done well to catch the relevant history of the project here...

@crystalfp, as @mb21 points to, I managed to persuade @jgm to use a slightly stricter standard than HTML (5). That is, the use polyglot markup. 

Essentially that entails using HTML (5) with a XHTML conforming syntax (a syntax that is XML valid) and a small number of additional restrictions. That results in a markup that is consistent with the HTML (5) spec thereby facilitating an easy switching between serving a web page as "application/xhtml+xml" or "text/html" MIME type and having it render, in the browser, identically. 

So on the assumption that polyglot markup continues to be our standard...

On the html element, default namespace attribute: this is required under polyglot

https://www.w3.org/TR/html-polyglot/#element-level-namespaces
> Polyglot markup declares the default namespaces on the root HTML element, html,"  ...

That is also exemplified in the minimal example, https://www.w3.org/TR/html-polyglot/#minimal-polyglot-html-document

On the html element, language attributes there are two relevant rules.

Firstly, If you specify the language you MUST use lang and xml:lang with identical values.  http://www.w3.org/TR/html-polyglot/#language-attributes 

[...]

PR #5466, opened 2019-04-24
Epub3 stylesheet link compatible with kindlegen

codingisacopingstrategy:
It’s seems that the commit https://github.com/daamien/pandoc/commit/411119b603509d9c9c42e48a661a61bc33eb6a24 removes the `type="text/css"` from both `<style>` and `<rel="stylesheet">` elements in all templates. The rationale for the change is described in https://github.com/jgm/pandoc/issues/5146 — the elements are optional and can cause HTML validators to complain.

However, this change causes Amazon’s [Kindlegen](https://www.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211]) (MAC OSX V2.9 build 1028-0897292) to not pick up the stylesheet when converting a Pandoc generated ePub3 to Mobi. No styles! This is probably a bug on Amazon’s end 😳—Epub3 uses XHTML5, so I think the type attribute should indeed be optional. But it is probably much easier reverting the change at Pandoc’s end, at least for the epub3 template.

One can verify the issue by running the [Toy example](https://pandoc.org/epub.html#a-toy-example) from Pandoc’s documentation, then running `kindlegen mybook.epub` and then use [`kindleunpack.py mybook.mobi`](https://github.com/kevinhendricks/KindleUnpack). You will find the default pandoc stylesheet not to be included. I could work around the issue by copying pandoc’s epub3 template and adding the attribute to the link tag. I suggest doing this change in the default template.

Whether or not I added the type attribute had no effect on the inline `<style>` tag which Pandoc places in the header by default—these get added regardless, as separate stylesheets for each page. Since the validators mentioned by @crystalfp in https://github.com/jgm/pandoc/issues/5146 complained only about the `<style>` tag, and only in the context of standalone HTML5, it seems save to suggest that adding the `type` attribute only to the link tag of the epub3 template is not very intrusive.

BTW I wonder why is there an inline style tag in the epub template, these could be part of the default css right?

### DIFF
--- a/data/templates/default.epub3
+++ b/data/templates/default.epub3
@@ -20,7 +20,7 @@ $highlighting-css$
   </style>
 $endif$
 $for(css)$
-  <link rel="stylesheet" href="$css$" />
+  <link rel="stylesheet" type="text/css" href="$css$" />
 $endfor$
 $for(header-includes)$
   $header-includes$


jgm:
Thanks.  I think it would make sense to move the inline style to the default stylesheet, too.

PR #5468, opened 2019-04-28
Expose more system functionalities to Lua

tarleb:
The `system` Lua module provides utility functions to interact with the
operating- and file system. Like the `text` module, the `system` module
must be loaded before use. E.g.

    local system = require 'system'
    print(system.currentdir())


tarleb:
This adds about 40kB to the executable.

jgm:
What happens if you `cd` to change directory in a lua filter?
Does this run in a subprocess, affecting only the working directory for the subprocess?
Or does this mean that when pandoc exits, you'll be in a different directory (which would be bad to allow, I think).

tarleb:
Indeed. With `chdir.lua` containing `(require 'system').chdir '/tmp'` and `filter.sh` being any filter in the current directory:

    $ echo test | pandoc --lua-filter=chdir.lua --filter=filter.sh
    Error running filter filter.sh:
    Could not find executable filter.sh

I'll have to add [`withCurrentDirectory`](https://hackage.haskell.org/package/directory-1.3.3.2/docs/System-Directory.html#v:withCurrentDirectory) somewhere in the Lua code.

tarleb:
The current working directory will now be restored after each Lua invocation.

Should we be doing the same with environment variables? It would be a little more work.

tarleb:
A bit more context for this PR.

- The main motivation was to get access to Haskell's functions for temporary files and directories. These would be of great help with tempfile-heavy filters like [`diagram-generator`](https://github.com/pandoc/lua-filters/tree/master/diagram-generator).
- Environment modifying functions where added to allow configuration of external programs. Filter authors may want to set or override environment variables before calling out to another executable. Example: temporarily overriding `TEXMF*` variables before calling pdflatex.
- The rest is merely for completeness.

Other features that I think could be helpful, but aren't implemented (yet):

- Platform-dependent creation and canonicalization of file paths;
- turning relative paths into absolute paths and vice versa.

jgm:
I am still a bit worried about these environment-polluting things; would it make more sense to add functions like `with_environment(env, function)`and `with_working_directory(dir, function)` instead of `setenv` and `cd`?

tarleb:
That makes sense and seems like the cleaner solution.

My approach will then be to remove the `system` module and replace it with a `pandoc.sys` module which contains just the three functions `with_environment`, `with_working_directory`, and `with_temp_directory`. That should strike a good balance between useful features and unnecessary bloat.

jgm:
I think it should be harmless to keep the "getter" functions exposed: e.g. `getenv` and `get_current_directory`.

tarleb:
Should be ok now.

doc/lua-filters.md
```diff
@@ -2618,3 +2618,111 @@ Parameters:
...
+:   Action to execute in the custom environment (function)
+
+Returns:
+
+-   The result(s) of the call to `callback`
+
+### with\_temporary\_directory {#system-with_temporary_directory}
+
+`with_tmpdir ([parent_dir,] templ, callback)`
+
```
jgm: Note mismatch in the name: is it `with_tmpdir` or `with_temporary_directory`?

doc/lua-filters.md
```diff
@@ -2618,3 +2618,111 @@ Parameters:
...
+:   Function which takes the name of the temporary directory as its
+    first argument (function).
+
+Returns:
+
+-   The result of the call to `callback`.
+
+### with\_working\_directory {#system-with_working_directory}
+
+`with_wd (directory, callback)`
```
jgm: ditto for `with_wd`

doc/lua-filters.md
tarleb: Argh, sorry. Copy-paste error.

jgm:
> Should be ok now. I left out getenv, as Lua already comes with os.getenv().

Confused about this, since there is still `environment` -- is that different from `os.getenv`?

jkr:
> Confused about this, since there is still `environment` -- is that different from `os.getenv`?

`os.getenv` takes an argument (`os.getenv("HOME")`) , while according to the docs, `environment` returns a table of all env vars.

Issue #2649, opened 2016-01-13
Mediawiki parser errors on the nl.wikipedia.org source


kerokeroloid:
As per #2648, I tried to make the smallest example that made the parser die, and in the process I found out that my files were generated incorrectly :) Anyway, here is the test case

```
pandoc testcase -f mediawiki -t plain > testcase.out
```

```
{| class="wikitable" style="line-height: 1.0"

|- bgcolor="#efefef"
|}
```

```
pandoc: 
Error at "source" (line 3, column 1):
unexpected "|"
expecting lf new-line or "|}"
|- bgcolor="#efefef"
^
```

If I remove that newline:

```
{| class="wikitable" style="line-height: 1.0"
|- bgcolor="#efefef"
|}
```

```
$ pandoc testcase -f mediawiki -t plain > testcase.out
(no output)
$ 
```

So that works. It seems then that it dies if there are empty newlines in tables. Is that invalid mediawiki code, or is it a pandoc bug?

kerokeroloid:
This:

```
{| class="wikitable"
! Eigenschap || Doorsnede || Vereniging
|-----
| [[Commutatief]]
| A ∩ B = B ∩ A
| A ∪ B = B ∪ A
|-----
| [[Associativiteit (wiskunde)|Associatief]]
| A ∩ (B ∩ C) = (A ∩ B) ∩ C
| A ∪ (B ∪ C) = (A ∪ B) ∪ C
|-----
| Er is een 0/1 element
| A ∩ ∅ = ∅
| A ∪ U = U
|-----
| [[Distributief]]
| A ∪ (B ∩ C) = (A ∪ B) ∩ (A ∪ C)
| A ∩ (B ∪ C) = (A ∩ B) ∪ (A ∩ C)
|}
```

Doesn't give the prettiest output:

```
  Eigenschap   Doorsnede   Vereniging |-----   Commutatief   A ∩ B = B ∩ A   A ∪ B = B ∪ A |-----   Associatief   A ∩ (B ∩ C) = (A ∩ B) ∩ C   A ∪ (B ∪ C) = (A ∪ B) ∪ C |-----   Er is een 0/1 element   A ∩ ∅ = ∅   A ∪ U = U |-----   Distributief   A ∪ (B ∩ C) = (A ∪ B) ∩ (A ∪ C)   A ∩ (B ∪ C) = (A ∩ B) ∪ (A ∩ C)
[...]

PR #5471, opened 2019-04-29
Replace parseAttr with parseAttrs

chinapedia:
Got similiar error with #2649

> Error at "source" (line 37, column 26):
> unexpected "c"
> expecting lf new-line
> |----- bgcolor="#DDDDFF" class="bottom"
>                          ^
> Pandoc could not convert successfully, error code: 65. Tried to run the following command: /usr/bin/pandoc --from=mediawiki --to=markdown_github /tmp/pandoc5cc73e2708042


mb21:
What input is not working for you?

Could you add a test to `test/commands/` with `pandoc -f mediawiki -t native` for the case you're fixing?

jgm:
This is clearly a good chance, so I'll merge it.  In general though, as @mb21 says, it is good to have regression tests (in `test/command/...`) and to specify in your comment the input that raised a problem.

PR #5473, opened 2019-04-30
Fix broken links in documents

violet716:
"User's Guide" links are pointing to MANUAL.html which gives 404 page.
This replaces them with http://pandoc.org/MANUAL.html.

### DIFF
--- a/doc/customizing-pandoc.md
+++ b/doc/customizing-pandoc.md
@@ -111,9 +111,9 @@ use an interface built directly into pandoc, and must be written
 in the Lua language).  If you are writing your own filters, it
 is best to use Lua filters, which are more portable (they
 require only pandoc itself) and more efficient.  See [Lua
-filters](lua-filters.html) for documentation and examples.  If
+filters](https://pandoc.org/lua-filters.html) for documentation and examples.  If
 you would prefer to write your filter in another language, see
-[Filters](filters.html) for a gentle introduction to JSON
+[Filters](https://pandoc.org/filters.html) for a gentle introduction to JSON
 filters.
 
 There's a repository of lua filters at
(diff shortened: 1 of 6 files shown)


jgm:
Thanks!  It would be better to use a relative URL: `../MANUAL.html`.

violet716:
Thanks for comment!

I replaced them with http://pandoc.org/MANUAL.html#...
because the "User's Guide" links of other documents such as [CONTRIBUTE.md](https://github.com/jgm/pandoc/blob/master/CONTRIBUTING.md#found-a-bug), [README.md](https://github.com/jgm/pandoc/blob/master/README.md#documentation) and [doc/customizing-pandoc.md](https://github.com/jgm/pandoc/blob/master/doc/customizing-pandoc.md#template-variables) have already been doing so.

By the way, I found that there are broken links in other documents as well.
Please take a look at new commits! Thanks.

jgm:
Thank you. I would still prefer the relative links. That way they will still work even if the whole site is moved or served locally.

violet716:
Oh, I appologize for the insufficient information in PR.

I thought those links are broken because I was reading the docs on github directly, but now noticed that it works fine on the pandoc.org website.

Anyway, some links pointing to http URL works fine in both github(+local repo) and pandoc.org while others, relative links, does not.
I think it's a bit confusing. What do you think?

jgm:
I see, I hadn't considered the case of people reading this with GitHub's rendering on the GitHub site.
Well, I guess we could use the absolute URL, but I'd prefer `https:` for all the links, because we now require https.

violet716:
I checked that all links work fine, except heading anchors in lua-filters.md.
I couldn't fix it because I don't know well about heading anchor of Markdown now. 😢 

I'd like to leave it as a new issue so that I (or someone) fix it later.

jgm:
Great, thanks!

PR #5475, opened 2019-05-01
Remove unused head.min.js link from revealjs

groovytron:
As [lib/js/head.min.js has been removed in the latest version of reveal.js](https://github.com/hakimel/reveal.js/commit/29b0e86089eb3ec0d4bb5811c9b723dfcf36703c#diff-d6453d319d952303b6752f08260dcb0d), the inclusion of that script needs to be removed.

Otherwise, an error occurs when you try to output `revelajs` presentation (**File ./reveal.js/lib/js/head.min.js not found in resource path**).

### DIFF
--- a/data/templates/default.revealjs
+++ b/data/templates/default.revealjs
@@ -89,7 +89,6 @@ $body$
     </div>
   </div>
 
-  <script src="$revealjs-url$/lib/js/head.min.js"></script>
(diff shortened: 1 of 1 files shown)


groovytron:
Just saw there was already a pull request for this fix. (https://github.com/jgm/pandoc/pull/5448)

PR #5478, opened 2019-05-02
FB2 writer: use genre metadata field

KetzerX:
In current implementation FB2 genre tag contains hardcoded value "unrecognised".
This PR makes use of genre field specified in metadata.

@jgm, could you review this please?

### DIFF
--- a/src/Text/Pandoc/Writers/FB2.hs
+++ b/src/Text/Pandoc/Writers/FB2.hs
@@ -40,6 +40,7 @@ import Text.Pandoc.Definition
 import Text.Pandoc.Logging
 import Text.Pandoc.Options (HTMLMathMethod (..), WriterOptions (..), def)
 import Text.Pandoc.Shared (capitalize, isURI, orderedListMarkers, hierarchicalize)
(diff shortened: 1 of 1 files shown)


mb21:
Maybe you could rebase from master so the CI can run?

KetzerX:
> Thanks! Maybe you could rebase from master so the CI can run?

Rebased.

src/Text/Pandoc/Writers/FB2.hs
```diff
@@ -96,7 +96,10 @@ pandocToFB2 opts (Pandoc meta blocks) = do
 
 description :: PandocMonad m => Meta -> FBM m Content
 description meta' = do
-  let genre = el "genre" "unrecognised"
+  let genre = case lookupMeta "genre" meta' of
```
jgm: simplify this by using `lookupMetaString` from T.P.Writers.Shared.
Otherwise looks good!

KetzerX: Done!

jgm:
thanks!

PR #5486, opened 2019-05-05
add documentation of title-meta for latex

p3732:


MANUAL.txt
```diff
@@ -1783,6 +1783,10 @@ Pandoc uses these variables when [creating a PDF] with ConTeXt.
     from [ConTeXt ICC Profiles].  See also [ConTeXt PDFA] for more
     details.
 
+`title-meta`
+:   sets a title for the document. Can be used to skip generating a titlepage
+    while still setting a document name.
+
```
jgm: Probably a bit too cryptic.  Maybe

> sets the title in the PDF metadata.  Normally this is automatically generated from `title`.  However, it may be set manually when `title` is omitted.

PR #5489, opened 2019-05-06
man: removes double inside

wookietreiber:


mb21:
Thanks! However, that file is generated. You'd need to change it here instead: https://github.com/jgm/pandoc/blob/master/MANUAL.txt#L3781

wookietreiber:
@mb21 Thanks for the update. When I change it in `MANUAL.txt`, should I run some command to regenerate the man page?

wookietreiber:
I've just figured out how to regenerate the man page. It seems someone else didn't do that as well, so these changes to the man page are in here now, too.

jgm:
As you probably discovered, there's a Makefile target
for that.  Normally we only regenerate it right before
a release, so you don't need to do it now; just fix
the bug in MANUAL.txt.

PR #5492, opened 2019-05-10
FB2 writer: do not wrap note references into <sup> and brackets

link2xt:
Existing FB2 readers, such as FBReader, already display links with type="note" as a superscript.

As for brackets, "official" FB2 converters add their own brackets, so they would be duplicated:
https://github.com/gribuser/fb2any/blob/c4a8b96d616f901d737f457e6d06daaad19836d6/FB_2_txt/fb2txt_engine.pas#L311

### DIFF
--- a/src/Text/Pandoc/Writers/FB2.hs
+++ b/src/Text/Pandoc/Writers/FB2.hs
@@ -444,7 +444,7 @@ toXml (Note bs) = do
   let fn_id = footnoteID n
   fn_desc <- cMapM blockToXml bs
   modify (\s -> s { footnotes = (n, fn_id, fn_desc) : fns })
-  let fn_ref = el "sup" . txt $ "[" ++ show n ++ "]"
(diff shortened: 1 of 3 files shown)


link2xt:
Here is how it looks in FBReader:

link2xt:
@KetzerX You seem to use FB2 writer (#5478), maybe you have some comments on this.

KetzerX:
@link2xt sorry for the late answer. Checked this kind of behavior in Fiction Book Editor, seems it uses square brackets as well:

image

image

KetzerX:
As far as I know some fb2 readers make such links in sup automatically, but some does not, so leaving square brackets is preferable in such case, otherwise it will look strange on readers without "supping" like: "Footnote2".

link2xt:
@KetzerX Ok, I will restore the square brackets, only remove the sup.

Issue #264, opened 2011-06-13
Embedded PDF images in HTML


candlerb:
The syntax:

```
![whee](foo.pdf)
```

works when generating LaTeX/PDF output, but not for HTML. The reason is that instead of the `<IMG>` tag you need to use the `<EMBED>` tag. See http://blogs.adobe.com/pdfdevjunkie/2007/08/using_the_html_embed_tag_to_di.html

Perhaps there should be a whitelist of filename extensions which get IMG tags, and the rest get EMBED (or vice versa). Or you could look at `mime.types` file to see if it maps to `image/*` or not. This could also set the `type` attribute of the element.

PR #5497, opened 2019-05-11
Output HTML5 video and audio elements

mb21:
As requested on [pandoc-discuss](https://groups.google.com/forum/#!topic/pandoc-discuss/BkBPXQ8pb3A) and also discussed in a [commonmark thread](https://talk.commonmark.org/t/embedded-audio-and-video/441/25), so I gave it a try.

Some tests needed to be changed slightly because the `alt` attribute comes last now.

---

I was also thinking of adding a way to disable the media player controls (which I enabled by default), like:

    {controls=false}

But the equivalent HTML would actually show the controls, since it doesn't take the value of the attribute into account, only the key. We could of course still implement it this way, but it sort of sets a precedent that the string `"false"` is somehow special in pandoc markdown attribute values. Or alternatively something like `{.no-controls}`.

---

About the list of image file extensions: we're loosing a few obscure ones in this pull. I was about to put the following in, but no browser can read them anyway. And `.raw` is not even guaranteed to be a raw image, could also be some other form of raw data used in science...

    -- following from https://stackoverflow.com/a/47612661/214446
    ,("arw","image/x-sony-arw")
    ,("cr2","image/x-canon-cr2")
    ,("crw","image/x-canon-crw")
    ,("dng","image/x-adobe-dng")
    ,("erf","image/x-epson-erf")
    ,("k25","image/x-kodak-k25")
    ,("KDC","image/x-kodak-kdc")
    ,("MRW","image/x-minolta-mrw")
    ,("NEF","image/x-nikon-nef")
    ,("ORF","image/x-olympus-orf")
    ,("PEF","image/x-pentax-pef")
    ,("RAF","image/x-fuji-raf")
    ,("RAW","image/x-panasonic-raw")
    ,("SR2","image/x-sony-sr2")
    ,("SRF","image/x-sony-srf")
    ,("X3F","image/x-sigma-x3f")

### DIFF
--- a/src/Text/Pandoc/MIME.hs
+++ b/src/Text/Pandoc/MIME.hs
@@ -55,9 +55,13 @@ reverseMimeTypes = M.fromList $ map (\(k,v) -> (v,k)) mimeTypesList
 mimeTypes :: M.Map String MimeType
 mimeTypes = M.fromList mimeTypesList
 
+-- | Collection of common mime types.
+-- Except for first entry, list borrowed from
+-- <https://github.com/Happstack/happstack-server/blob/master/src/Happstack/Server/FileServe/BuildingBlocks.hs happstack-server>
 mimeTypesList :: [(String, MimeType)]
-mimeTypesList = -- List borrowed from happstack-server.
-           [("gz","application/x-gzip")
+mimeTypesList =
+           [("cpt","image/x-corelphotopaint")
+           ,("gz","application/x-gzip")
            ,("cabal","application/x-cabal")
            ,("%","application/x-trash")
            ,("323","text/h323")
--- a/src/Text/Pandoc/MIME.hs
+++ b/src/Text/Pandoc/MIME.hs
@@ -11,12 +11,13 @@
 Mime type lookup for ODT writer.
 -}
 module Text.Pandoc.MIME ( MimeType, getMimeType, getMimeTypeDef,
-                          extensionFromMimeType )where
+                          extensionFromMimeType, mediaCategory ) where
 import Prelude
 import Data.Char (toLower)
 import Data.List (isPrefixOf, isSuffixOf)
+import Data.List.Split (splitOn)
 import qualified Data.Map as M
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, listToMaybe)
 import System.FilePath
 
 type MimeType = String
@@ -41,6 +42,12 @@ extensionFromMimeType :: MimeType -> Maybe String
 extensionFromMimeType mimetype =
   M.lookup (takeWhile (/=';') mimetype) reverseMimeTypes
   -- note:  we just look up the basic mime type, dropping the content-encoding etc.
+
+-- | Determine general media category for file path, e.g.
+--
+-- prop> mediaCategory "foo.jpg" = Just "image"
+mediaCategory :: FilePath -> Maybe String
+mediaCategory fp = getMimeType fp >>= listToMaybe . splitOn "/"
 
 reverseMimeTypes :: M.Map MimeType String
 reverseMimeTypes = M.fromList $ map (\(k,v) -> (v,k)) mimeTypesList
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -65,18 +65,21 @@ import Text.Blaze.Internal (preEscapedString, preEscapedText)
 #endif
 #if MIN_VERSION_blaze_html(0,5,1)
 import qualified Text.Blaze.XHtml5 as H5
+import qualified Text.Blaze.XHtml5.Attributes as A5
 #else
 import qualified Text.Blaze.Html5 as H5
+import qualified Text.Blaze.Html5.Attributes as A5
 #endif
 import Control.Monad.Except (throwError)
 import Data.Aeson (Value)
-import System.FilePath (takeBaseName, takeExtension)
+import System.FilePath (takeBaseName)
 import Text.Blaze.Html.Renderer.Text (renderHtml)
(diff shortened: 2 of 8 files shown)


src/Text/Pandoc/MIME.hs
```diff
@@ -52,6 +59,7 @@ mimeTypesList :: [(String, MimeType)]
 mimeTypesList = -- List borrowed from happstack-server.
            [("gz","application/x-gzip")
            ,("cabal","application/x-cabal")
+           ,("cpt","image/x-corelphotopaint")
```
link2xt: What is that? It's not even in happstack-server as the comment says. Maybe move it to a separate commit?

src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -1135,33 +1126,31 @@ inlineToHtml opts inline = do
...
                               (if isReveal
                                   then customAttribute "data-src" $ toValue s
                                   else A.src $ toValue s) :
                               [A.title $ toValue tit | not (null tit)] ++
                               attrs
-                        return $ foldl (!) H5.embed attributes
+                            imageTag = (if html5 then H5.img else H.img
+                              , [A.alt $ toValue alternate' | not (null txt)] )
+                            mediaTag t = (
+                                t "Sorry, your browser doesn't support embedded video or audio."
```
link2xt: I am not sure it is a good idea to use hardcoded non-localized text. Maybe just put a link to a file instead, with the text equal to the "alt" attribute?

mb21: Good point! I did just that, thanks!

src/Text/Pandoc/MIME.hs
mb21: Yeah, moved it to the second commit... the file extension was in the markdown writer code I removed in that second commit.

test/command/video-audio.md
```diff
@@ -0,0 +1,19 @@
+```
+% pandoc -f markdown-implicit_figures -t html
+![](./test.mp4)
+
+![Your browser does not support video.](foo/test.webm){width=300}
```
link2xt: Seems like you have not updated the test file

mb21: No, that's the alt text... see a few lines above.

src/Text/Pandoc/MIME.hs
```diff
@@ -11,12 +11,13 @@
 Mime type lookup for ODT writer.
 -}
 module Text.Pandoc.MIME ( MimeType, getMimeType, getMimeTypeDef,
-                          extensionFromMimeType )where
+                          extensionFromMimeType, mediaCategory ) where
```
jgm: Be sure to document API changes like this in the commit message.

test/command/2662.md
```diff
@@ -7,5 +7,5 @@
    :scale: 300 %
    :alt: alternate text
 ^D
-<p><img src="http://url.to.image/foo.png" alt="alternate text" class="align-left" width="600" height="300" /></p>
```
jgm: Some change seems to have switched the position of the alt attribute relative to the others.  I think it would be desirable not to introduce this change, if possible, since it has no essential connection to the issue being solved here and might conceivably create problems for others (e.g. in test suites).

src/Text/Pandoc/MIME.hs
jgm: Oh, I see that you did. :)

jgm:
Generally this looks promising to me, but I'm still a bit undecided.  I guess I don't see any real harm in overloading image syntax this way.  However, a user might expect the results to work in any output format, and currently it will only work in HTML-based formats.  Maybe it doesn't make a lot of sense in most other formats.  On the other hand, if it's only for HTML formats, one could just use raw HTML, which would give full control over the details.  This would also lead to less confusion, since users will expect a markdown form to work in all formats...

If we do overload markdown images in this way, we should decide whether to put it under an extension.  And we should document this behavior in the manual.

test/command/2662.md
mb21: Yes, it would be possible by making the haskell source code more complicated and I didn't think it was worth the trade-off. But I could change it of course...

mb21:
Mainly, I thought it was an improvement over the current `<embed>` output and such is an easy win. Basically, I cannot think of any scenario where you would prefer the current output over the new one.

But yes, if you think of it as a somehow "sanctioned" markdown syntax for video (or embedding arbitrary files), then I see your concerns. But I wouldn't think of this as such, because of all the output formats of a markdown-converter I can think of, video only works for HTML-based output formats anyway.

Arguably, pandoc's current output with the `<embed>` tag for non-image-extensions is already unique among markdown converters. This was introduced for embedding PDFs in https://github.com/jgm/pandoc/issues/264. So if we were to introduce an extension, the `<embed>` output should also go behind that extension. Then again, the chances that you have an image file you want to embed that contains an extension we detect wrongly as pdf or video are probably approaching zero...

jgm:
OK, you've convinced me!

mb21:
Haha, great :-)

lukesmurray:
> I was also thinking of adding a way to disable the media player controls (which I enabled by default)

Is there a way to disable controls? Setting controls to nil in a lua filter does not work.

```lua
if FORMAT:match("html") then
    function Image (elem)
        if isVideo(elem.src) then
            elem.attr.attributes.autoplay = ""
            elem.attr.attributes.loop = ""
            elem.attr.attributes.muted = ""
            elem.attr.attributes.playsinline = ""
            -- controls are still rendered even though the attribute is removed
            elem.attr.attributes["controls"] = nil
        end

        return elem
    end
end
```

tarleb:
Closed pull requests are not a good place to ask these questions. Please
reach out on the pandoc-discuss mailing list.

PR #5499, opened 2019-05-12
Improve output of Lua tests

tarleb:
This makes use of tasty-lua, a package to write tests in Lua and integrate the results into Tasty output. Test output becomes more informative: individual tests and test groups become visible in test output. Failures are reported with helpful error messages.

The downside of tasty-lua is its way of running tests, which is due to the need to create tests dynamically, and because of problems with strictness/lazyness and which I was unable to resolve quickly. The current method will run all tasty-lua tests as the very first tests when executing the test suite, regardless of their position in the output. It seems acceptable right now, as tests are fast (< 0.05 sec).


tarleb:
I guess we could go the route of `Tests.Command` and use `unsafePerformIO` to fix these shortcomings. Would that be better?

jgm:
Can you describe these problems with laziness/strictness more fully?

tarleb:
My original idea was to convert Lua functions into IO actions, and then to test these. That approach failed, i.e., led to crashes, probably because the Lua state had changed in an unexpected way by the time the function was executed.

I didn't spend much time debugging, as my original use-case was just to write better tests for the hslua-module-* packages. The current way of handling things was easier to write and is sufficient for what I had in mind.

Of course, pandoc has a much larger test-suite, so the downsides are more relevant here. I can give my original idea another try if you deem it important.

jgm:
This feels a bit kludgy to me, running the lua tests first as a special case.
I can't actually remember why we had to resort to unsafePerformIO for the tasty tests. That doesn't seem ideal.  But if we're already doing it, trying that for the lua tests might make sense.

tarleb:
I switched to a different approach which conforms better to tasty's design priciples: a Lua file is now a single test case, but information about successes or failures is collected and included in the test result.

A successful test will be presented like this:

    pandoc.util:                                         OK
      +++ Success: 9 Lua tests passed

Failures contain the relevant information:

    pandoc.util:                                                FAIL
      sanity check:
      lua/module/pandoc.utils.lua:52: 
      expected values to be equal, got 'Header' and 'Sec'

      09 Nov 1989:
      lua/module/pandoc.utils.lua:61: 
      expected values to be equal, got '1989-11-9' and '1989-11-09'

All other approaches that I tried didn't amount to much. Either the Lua interpreter's state was closed before the tests could run, leading to crashes, or the problems outlined in the previous comments persisted. I believe the new approach strikes a good balance between being elegant and informative.

jgm:
Sounds like a good solution. I'll merge now!

PR #5505, opened 2019-05-14
Fix using Beamer with geometry

### LINKED ISSUE #5504: Cannot define geomtry when using beamer
Pandoc v2.7.2

When I run `pandoc -t beamer -o slides.pdf slides.md`, I get an error:
```
Error producing PDF.
! LaTeX Error: Option clash for package geometry.

 ...                                              
                                                  
l.82 \newif
```

When using Beamer, instead of pulling in the geometry package, one should use `\geometry{}` instead according to https://tex.stackexchange.com/questions/400807/how-to-change-geometry-with-beamer-class?noredirect=1&lq=1.

orangecms:
This should fix #5504 :)

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -229,7 +229,11 @@ $if(verbatim-in-note)$
 \VerbatimFootnotes % allow verbatim text in footnotes
 $endif$
 $if(geometry)$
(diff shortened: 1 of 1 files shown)


jgm:
Thanks!

Issue #4737, opened 2018-06-26
Image alt tag and figcaption should be differentiated


matthewlehew:
[As discussed in this thread on the discussion group,](https://groups.google.com/d/msg/pandoc-discuss/BefkPmp0x84/OYbnQcR9AAAJ) the behavior of captioning images is less than ideal for maintaining accessibility.

Currently, the bracketed text becomes both the alt tag and the figcaption:

``` 
![alt text](lalune.jpg) 

  -> 

<figure> 
<img src="lalune.jpg" alt="alt text" /><figcaption>alt text</figcaption> 
</figure> 
```

When wanting to create accessible documents, it is very often necessary to provide an alt tag that describes the image itself with a caption that provides extra explanatory content. Having them be the same doesn't align with their intended use.

Adding an additional descriptor creates a title tag, which is unhelpful in this instance:

``` 
![alt text](lalune.jpg "title") 

  -> 

<figure> 
<img src="lalune.jpg" title="title" alt="alt text" /><figcaption>alt text</figcaption> 
</figure> 
``` 

I propose amending the behavior so that the additional descriptor becomes the alt text and the bracketed text becomes the figure caption, with some other method used to derive the image title.

``` 
![caption text](lalune.jpg "alt text") 

  -> 

<figure> 
<img src="lalune.jpg" alt="alt text" /><figcaption>caption text</figcaption> 
</figure> 
``` 

**The additional benefit is that this would no longer require the unofficial method of escaping a space to suppress a caption and have it persist as solely an alt tag.** Instead, you would simply:

``` 
![](lalune.jpg "alt text") 

  -> 

<figure> 
<img src="lalune.jpg" alt="alt text" /> 
</figure> 
```

mb21:
Should this be closed in favour of https://github.com/jgm/pandoc/issues/3177 ?

reagle:
@mb21 I don't *think* so, I think you'd have to say more as how it would fit...?

mb21:
It’s just that the current figure handling is somewhat of a hack and a proper figure elemtn int the AST, i.e. #3177, would handle caption and alt separately.

matthewlehew:
@mb21 I agree. I missed #3177 when searching on this topic. It appears my issue is a subset of the issue raised there. 

My only additional comment is that #3177 should be considered high priority. The current way Pandoc handles figures is a serious hindrance for accessibility in academic publishing.

reagle:
[...]

PR #5518, opened 2019-05-22
HTML writer: emit empty alt tag in figures

mb21:
The same alt text is already in the `<figcaption>` and screen-readers would read it twice, see #4737

Note: `Image attr [] (s,tit)` wouldn't work, since this would emit no alt tag which is invalid HTML.


jgm:
My only concern about this is that some accessibility validation software might complain about the missing alt tag.  (I'm not sure about this, though.)  I agree that this would improve usability with screen readers.  I also agree that ultimately we need to separate alt and caption.  (That's the issue for adding a proper figure type.)

mb21:
> My only concern about this is that some accessibility validation software might complain about the missing alt tag.

That's why it emits an empty alt tag `alt=""`... see e.g. https://davidwalsh.name/accessibility-tip-empty-alt-attributes

jgm:
OK, thanks.

PR #5522, opened 2019-05-24
fix handling of `file:` URL scheme in `downloadOrRead`

### LINKED ISSUE #5517: [mac] org to pdf image absolute path not work
- pandoc veriosn: pandoc 2.7.2
- os: mac 10.14.3 (18D109)
- org version: 9.1.9

when i run command
`pandoc --self-contained --pdf-engine=xelatex --template=/Users/username/soft/plugin/latex/pm-template.latex -f org -t latex apollo-config-swim.org -o test.pdf`

I got warn info:
`[WARNING] Could not fetch resource 'file:///Users/username/github/blog/images/apollo-home-screenshot.png': PandocHttpError "file:///Users/username/github/blog/images/apollo-home-screenshot.png" (InvalidUrlException "file:///Users/username/github/blog/images/apollo-home-screenshot.png" "Invalid scheme")`

org file image content:
[[file:/Users/username/github/blog/images/apollo-home-screenshot.png]]
[[/Users/username/github/blog/images/apollo-home-screenshot.png]]

these all not work!

What should i do ?

mb21:
move up the pattern match to be reachable, closes #5517

seems this goes all the way back to https://github.com/jgm/pandoc/commit/4cb124d147790814cf2055afdfd17e500cece559

### DIFF
--- a/src/Text/Pandoc/Class.hs
+++ b/src/Text/Pandoc/Class.hs
@@ -572,10 +572,10 @@ downloadOrRead s = do
             Nothing -> openURL s' -- will throw error
     (Nothing, s') ->
        case parseURI s' of  -- requires absolute URI
-            -- We don't want to treat C:/ as a scheme:
-            Just u' | length (uriScheme u') > 2 -> openURL (show u')
             Just u' | uriScheme u' == "file:" ->
                  readLocalFile $ uriPathToPath (uriPath u')
+            -- We don't want to treat C:/ as a scheme:
+            Just u' | length (uriScheme u') > 2 -> openURL (show u')
             _ -> readLocalFile fp -- get from local file system
    where readLocalFile f = do
              resourcePath <- getResourcePath
--- a/test/command/5517.md
+++ b/test/command/5517.md
@@ -1,0 +1,26 @@
+Use epub output to trigger `downloadOrRead` in `Text.Pandoc.Class`
+in order to test `file:` URL-scheme handling.
+
+There are no relative `file:` URLs, so we cannot
+test with an actual file, since we don't know the
+current working directory. Instead, we use `/dev/null`
+as a file that certainly exists, redirect stderr
+to stdout and check that there is no warning.
+
+```
+% pandoc -M title=test -f native -t epub -o /dev/null 2>&1
+[Para [Image ("",[],[]) [] ("file:/dev/null","")]]
+^D
+```
+
+```
+% pandoc -M title=test -f native -t epub -o /dev/null 2>&1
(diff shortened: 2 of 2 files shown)


src/Text/Pandoc/Class.hs
```diff
@@ -573,9 +573,9 @@ downloadOrRead s = do
     (Nothing, s') ->
        case parseURI s' of  -- requires absolute URI
             -- We don't want to treat C:/ as a scheme:
```
tarleb: That comment should probably be moved, too.

mb21: of course, thanks!

test/command/5517.md
```diff
@@ -0,0 +1,28 @@
...
+test with an actual file, since we don't know the
+current working directory. Instead, we redirect
+stderr to stdout and check for the correct warning.
+
+```
+% pandoc -M title=test -f native -t epub -o /dev/null 2>&1
+[Para [Image ("",[],[]) [] ("file:/testpath/foo.jpg","")]]
+^D
+[WARNING] Could not fetch resource 'file:/testpath/foo.jpg': PandocIOError "/testpath/foo.jpg" /testpath/foo.jpg: openBinaryFile: does not exist (No such file or directory)
+```
```
jgm: I don't like the idea of triggering a read to an unknown location on the file system, even though a read should generally be harmless.

A cleaner way to test this feature (though more work to set up, since it would have to be done in Haskell) would be to use `runPure` and set up a fake file system.  (This is done using `modifyPureState` to update `stFiles` to a `FileTree` containing the files you want.  See also `insertInFileTree`.)

jgm:
Good catch!  Can't believe that has been there so long...

test/command/5517.md
mb21: Agreed that this would be a tad cleaner. Although it would open up the possibility of the fake fs behaving differently than the real one. Then again, we have Windows-specific preprocessor changes in the filepath handling anyway, which makes it difficult to test behaviour on different platforms...

I may have the time to look into this at some later point... but what do you think of reading from `/dev/null` for now? I updated the pull and it seems to work.

jgm: OK, I guess that's good for now.

jgm:
Looks like you can have relative file: URIs of the form `file:test/lalune.jpg`, so I'll add a test like that too.

jgm:
Whoops, look like Windows CI didn't like these tests.
Not sure why, but we need something that will work on Windows.

jgm:
Added new issue #5528 to track this.

PR #5532, opened 2019-05-29
adds optional draft to latex template

wookietreiber:


jgm:
I think this is getting too specialized to include in the default template.
Easy enough to put `\usepackage{draftwatermark}` in header includes if you want to use this package.

wookietreiber:
I was hoping this is a general enough use case to warrant an inclusion for the simplicity of just using `-V draft` instead of defining the header include. I can understand your hesitance, though, to not include every other package.

jgm:
Yes, we try to limit the packages we rely on in the default template.

Issue #2497, opened 2015-11-05
Support Jira markup (a form of textile)?


hisplan:
Jira Markdown is slightly different from GitHub-flavored Markdown, and it's massively used by lots of software development teams (eBay, Salesforce, LinkedIn, Square, ...)

Any plan to support Jira Markdown?

Here's Jira's markdown spec:
https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=all

tarleb:
AFAIK Jira's markup language is Textile, not Markdown.  Textile is fairly well supported by Pandoc, just use <kbd>pandoc -f textile</kbd>  when converting Jira notes.

hisplan:
I do not think JIRA markup is Textile. Just to be sure, I tried http://pandoc.org/try with some JIRA markdown syntax and confirmed that `{code}...{code}` or `{{monospaced}}` were not correctly converted to HTML or any other.

tarleb:
You are right, I missed those `{{mono}}` and `{code}` issues.  I'm not really sure *what* language JIRA is using. It seems closer to Textile than to Markdown (and wikipedia [lists](https://en.wikipedia.org/wiki/Textile_%28markup_language%29) JIRA as a project using textile). However, apparently it isn't standard textile either.

I'm not aware of any plan to support that dialect.  From what I see, it should be possible to add it as a textile variant.  It might actually be a nice task for someone trying to get their feet wet with pandoc/haskell.

StevenACoffman:
From [this Atlassian support source](https://answers.atlassian.com/questions/31681034/how-different-is-jira-markup-from-textile):
> Jira Markdown and Textile are the same for the most used markups like:
> Bullets, enumerable lists, headers, italic, bold, tables, html, quote ...
> 
> And different for other functions like:
> Preformated text, links, identation, code block ...

StevenACoffman:
For reference, [J2M](https://github.com/FokkeZB/J2M/) has some good regex's for the bits of Jira markdown that are not in Textile. Preformated text, links, code block at least. They are [considering using pandoc](https://github.com/FokkeZB/J2M/issues/12) for the much more comprehensive Textile support.

retronym:
Here's a related effort I came across, but haven't tried:

https://github.com/HanStolpo/JiraStructureToDocx/tree/master/PandocReaderJira
[...]

PR #5548, opened 2019-06-04
Add jira writer

tarleb:
This adds support for Atlassian's jira markup.

Closes: #2497

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -53,6 +53,7 @@ data-files:
                  data/templates/default.latex
                  data/templates/default.context
                  data/templates/default.texinfo
+                 data/templates/default.jira
                  data/templates/default.man
                  data/templates/default.ms
                  data/templates/default.markdown
@@ -242,6 +243,7 @@ extra-source-files:
                  test/tables.docbook4
                  test/tables.docbook5
                  test/tables.jats
+                 test/tables.jira
                  test/tables.dokuwiki
                  test/tables.zimwiki
                  test/tables.icml
@@ -273,6 +275,7 @@ extra-source-files:
                  test/writer.docbook4
                  test/writer.docbook5
                  test/writer.jats
+                 test/writer.jira
                  test/writer.html4
                  test/writer.html5
                  test/writer.man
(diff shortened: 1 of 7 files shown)


tarleb:
Jira's markup is a less-powerful variant of Textile. It is different enough to warrant a separate writer, but large parts were copy-pasted from the Textile writer.

There is no native math support in Jira, but plugins exist to make it work with LaTeX formulas. I'll publish a Lua filter for that somewhere.

Motivation for all this: not going crazy at work.

pandoc.cabal
```diff
@@ -53,6 +53,7 @@ data-files:
                  data/templates/default.latex
                  data/templates/default.context
                  data/templates/default.texinfo
+                 data/templates/default.jira
```
tarleb: I forgot to add `tables.jira`; will do tonight.

tarleb:
Found some problems, will reopen once those are resolved.

src/Text/Pandoc/Writers/Jira.hs
```diff
@@ -0,0 +1,322 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{- |
+   Module      : Text.Pandoc.Writers.Jira
+   Copyright   : Copyright (C) 2010-2019 John MacFarlane
```
jgm: Put yourself here?

jgm:
Could you briefly describe the differences from textile, and say why it wouldn't work better to handle this by parameterizing the code in the textile writer?

tarleb:
The main differences from Textile seem to be:

- no support for embedded HTML;
- different syntax for inline code (`@code@` vs `{{code}}`);
- different syntax for code blocks (`bc.` vs `{code}`);
- different sytnax for links (`"link example:https:://example.com` vs `{link example|https://example.com}`);
- different syntax for subscript and superscript (`[^sub^]` vs `{^sub^}`);
- different syntax for table header rows (`_.` vs `||`);
- syntax support for multi-paragraph quotes via `{quote}`;
- support for anchors (`{anchor:identifier}`);
- no way to add classes to elements;
- no support for table column alignment;
- no support for footnotes;
- no support for ordered lists starting at numbers other than 1;
- no support for definition lists.

It seems that jira syntax is only loosly based on textile; handling both by the same writer would mean littering all but the most simple element handlers with `if` statements. Despite being similar at the surface, there is not much code that can be shared by both writers.

jgm:
OK, great. I agree it makes sense to have a separate writer.

jgm:
@tarleb can you also update

- MANUAL.txt (list of output formats)
- the cabal description

tarleb:
done

Issue #5568, opened 2019-06-10
Lua filters: provide clone methods for pandoc.Attr and pandoc.Attributes objects


bpj:
I have a problem with `pandoc.Attr` objects and in particular `pandoc.Attributes` objects: since the latter are quite complex objects which have metatable magic to act as arrays and as maps at the same time they are non-trivial, not to say impossible to copy, at least without major encapsulation breaches. I'm trying to transfer the attributes of one element to another element, but also need to get a new Attr object — a copy or clone — so that I can modify some attributes — notably the id if any — in the second element without modifying the first element.  I have succeeded in writing a function which clones an Attr object by first looping through the pairs of the Attributes object copying them to a plain table, and then passing the id and a clone of the classes List of the original Attr object and that plain table to the Attr constructor.

````lua
local function copy_attrs (attrs, x)
    local keyvals = attrs.attributes
    local kv_copy = {}
    for k,v in pairs(keyvals) do
        kv_copy[k] = v or nil
    end
    x = x or {}
    for k,v in pairs(x) do
          kv_copy[k] = v or nil
    end
    return pandoc.Attr(
        attrs.identifier,
        attrs.classes:clone(),
        kv_copy
    )
end
````

Edit: for some reason the `or nil` was lost from the `kv_copy[k] = v or nil` line above

The issue is simply that I think that it would be a great improvement if both Attr objects and Attributes objects had clone methods to make this kind of thing easier and safer.

I have seen and read #4071

See also #5569

tarleb:
Good idea. If we do this, we should probably add `clone` method
to all AST elements for consistency.

---

The pandoc.Attributes type is weird in that it is really a list
of pairs. We can use this to shorten the clone function a little:

    local function copy_attrs (attr)
        return pandoc.Attr(
          attr.identifier,
          attr.classes:clone(),
          pandoc.List:new(attr.attributes):clone()
        )
    end

Adding

    pandoc.Attr.behavior.clone = copy_attrs

allows to use colon syntax like this:

    function Div (div)
      local attr = div.attr:clone()
      ...
    end

bpj:
I was actually going to suggest adding clone methods to all AST elements but decided to keep this focused on the immediate problem

bpj:
Great! Thanks! I hope it wasn't too much work!

tarleb:
No worries. I do this for fun, after all :smile:

PR #5572, opened 2019-06-11
Lua: add a `clone()` method to all AST elements

tarleb:
Closes: #5568


jgm:
I noticed that the `clone` method is repeated in the docs for all elements.  Would it perhaps be better to list it once under "Methods applicable to all elements", or something like that?

tarleb:
I was torn between the two options and Emacs keyboard-macros had made this repetitive editing like quite easy.

I have changed the docs, so now there is a "Shared Properties" section and described `clone` there.

Issue #4999, opened 2018-10-22
Powerpoint fenced code blocks and verbatim uses hard coded formatting


mainframed:
I reported these two issues on the google group and was told to submit them here. 

Basically fenced code blocks ` ``` ` and verbatim ` `` ` do not match the font size of the slide reference doc. Using custom `--reference-doc` I have a theme but unfortunately the size of code blocks are hard coded to 18pt [as seen here](https://github.com/jgm/pandoc/blob/11e681edac3cbdc94af90bee8370e8aa4fddf955/src/Text/Pandoc/Writers/Powerpoint/Output.hs#L738): ` then [("sz", "1800")]`

Here's what that looks like in my presentation:

<img width="868" alt="screen shot 2018-10-21 at 9 35 12 pm" src="https://user-images.githubusercontent.com/1973519/47277923-b89f3200-d579-11e8-8b8b-5401f38af559.png">

The second issue can be found [here](https://github.com/jgm/pandoc/blob/11e681edac3cbdc94af90bee8370e8aa4fddf955/src/Text/Pandoc/Writers/Powerpoint/Output.hs#L834). This code `Just px -> [("marL", show $ 12700 * px), ("indent", "0")]` indents the code block about 1.4 inches from the start of the text area. It doesn't appear there's a way to set this distance without modifying the source as it too is hard coded.

jeroendehaas:
@mainframed For my own work, I tried to address the issues you reported a couple of months ago. My changes are now merged. You might appreciate them. If you do get a chance to try a new version, I'd be interested in your thoughts on the new behavior.

mb21:
So this issue can be closed?

PR #5573, opened 2019-06-11
Make Powerpoint code formatting configurable

jeroendehaas:
This commit alters the way in which the Powerpoint writer treats inline code and code blocks. Previously all code was formatted in Courier 18pt. As a result inline code looked out of place in most circumstances (e.g. headers, main content). Furthermore, code blocks were given a left margin that leaves very little room for actual content in two-column slides. These observations were previously reported in issue #4999.

This commit formats inline code in the same font size as the surrounding text. The font in which code is typeset may be set via the `code-font` variable. The font size and left margin of code blocks are controlled via the `code-font-size` and `code-margin` variables respectively. 

If these variables are unset, code blocks will not appear any differently from before. However, the size of inline code is now context dependent.


mb21:
Not sure why the tests are failing on some platforms... can you maybe rebase from current master and force-push to the branch again?

jeroendehaas:
Rebased changes, all tests pass on my machine

jgm:
@jkr I'd like to get your thoughts before merging this.

jkr:
This looks good in principle (and it was thought through a bit on the list). I hope that at some point in the future we will provide a better default based on base body text size, but this will be a good default.

I'd like to see (a) some tests, and (b) how it looks with a template. The latter would come automatically with the former, since we test with a template as well.

@jeroendehaas: the procedure is to add the input (native) file to `test/pptx`. Don't add an output file. Add it to `Tests/Writers/Powerpoint.hs`. When you run the tests, it will produce two output files (`{$WHATEVER}.pptx` and `{$WHATEVER}_templated.pptx`). Open those up in your local PowerPoint and inspect them. If they work as expected, and without corruption, commit them, along with the input file and the changed `.hs` file, and note in your commit message what version of PowerPoint you checked them in (in case we run into problems later).

Let me know if you're not able to get to this, and I could try to add them later today.

jeroendehaas:
@jkr I added two tests: one where code uses the default formatting (i.e. the old behavior for code blocks and a flexible font size for inline code) and one where the code formatting is controlled via the newly introduced variables.

Tested on PowerPoint for Mac version 16.25 (19051201)

jeroendehaas:
I had a quick glance at the xml-files a pptx-archive and some OOXML documentation. I might be able to base the font size of code blocks on the reference archive (for instance by looking at `p:defaultTextStyle` in `presentation.xml`). I'll try to experiment a bit tonight.

Maybe it's worth postponing the merge until I've had a look at that?

jkr:
Sure -- that would be great. It seems like the margin issue wouldn't be touched by that, though, so if this has been a sticking point for you, something like this might still be useful.

It might be helpful to take a look at how images are sized, since that follows a similar flow (get data at the document level from the reference and then use it in presentation).

One other point: could you put the testing platform in the commit message with the tests? That will make it easier to track down if (when) we have a problem that comes down to version/platform.

src/Text/Pandoc/Writers/Powerpoint/Presentation.hs
```diff
@@ -29,6 +29,7 @@ module Text.Pandoc.Writers.Powerpoint.Presentation ( documentToPresentation
                                                    , Paragraph(..)
                                                    , ParaElem(..)
                                                    , ParaProps(..)
+                                                   , CodePlacement(..)
```
jkr: One condern here: on the whole, I want the `Presentation` data type to model the actual powerpoint format as closely as possible, and keep formatting things that aren't in pptx in `Output.hs`. `CodePlacement` isn't something that really has an analogue in PowerPoint,  right?

Would there be any way to keep these changes confined to `Output`?

jkr:
See also my comment on the code re: trying to keep these changes in `Output.hs` as much as possible.

src/Text/Pandoc/Writers/Powerpoint/Presentation.hs
jeroendehaas: Thank you for getting back on this. I had been pondering about this as PowerPoint has no real concept of code

jeroendehaas:
As per @jkr request, I have altered the way in which inline code and code blocks are treated once more. 

The font size of both inline code and code blocks is now determined by the level of nesting as is the margin. 

The font itself can still be changed via a variable, which I've renamed back to `monofont` in order to be consistent with other writers that offer such this option.

jgm:
> Maybe it's worth postponing the merge until I've had a look at that?

Do you want me to hold off some more, or is it ready to go?

jeroendehaas:
> > Maybe it's worth postponing the merge until I've had a look at that?
> 
> Do you want me to hold off some more, or is it ready to go?

This is ready to go. Late last night I noticed a small issue with syntax highlighted code blocks. That is now also fixed.

jgm:
Great, many thanks!

jgm:
You can't set options in YAML metadata with pandoc. You must be using a third-party wrapper or something.  If you look at the above commit, you'll see what it changes.  It adds monofont and some contextual changes in code formatting.  Not code-font or code-font-size.

PR #5578, opened 2019-06-13
Add option to include source files in ConTeXt PDFs

tstenner:
Most conversions can be reversed (with some formattings lost, though).
Generally, this isn't possible with conversions to PDFs.

This PR adds a metadata variable to include the source files in the resulting PDF so they can be extracted from the PDF afterwards (e.g. with `pdfdetach -saveall`).

Test case:

`test.md`:

```
---
title: Attachment test
includesource: true
...

Your PDF should include this source document as attachment to the file.
```

```
$ echo 'And this one (`test2.md`) too.' > test2.md
$ pandoc -t context test.md test2.md -o test.pdf
$ pdfdetach -list test.pdf
2 embedded files
1: test.md
2: test2.md
```

visual result

### DIFF
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -105,6 +105,11 @@ $endif$
 \setupxtable[body][]
 \setupxtable[foot][bottomframe=on]
 
+$if(includesource)$
+$for(sourcefile)$
+\attachment[file=$curdir$/$sourcefile$,method=hidden]
+$endfor$
+$endif$
 $for(header-includes)$
 $header-includes$
 $endfor$
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1828,6 +1828,9 @@ Pandoc uses these variables when [creating a PDF] with ConTeXt.
 `whitespace`
 :   spacing between paragraphs, e.g. `none`, `small` (using [`setupwhitespace`])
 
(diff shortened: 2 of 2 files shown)


mb21:
Nice, looks like a useful addition to me. Can you add something to https://github.com/jgm/pandoc/blob/master/MANUAL.txt in the `Variables for ConTeXt` section?

data/templates/default.context
```diff
@@ -104,6 +104,12 @@ $endif$
 \setupxtable[body][]
 \setupxtable[foot][bottomframe=on]
 
+$if(includesource)$
+$for(sourcefile)$
+\attachment[file=$curdir$/$sourcefile$,method=hidden]
+$endfor$
+$endif$
+
```
jgm: Remove line 112, as it creates an extra blank line in context documents even if the new variable is not set (that's why the tests fail).  If you want the blank line in the generated context, put it before the endif, and then it will only be included if the attachment command is.

tstenner: Done.

mb21:
Could you try rebasing from master? I think the CI/tests should pass then..

tstenner:
Done, the CI is happy now.

PR #5586, opened 2019-06-15
Fix #4499: mbox and hbox

### LINKED ISSUE #4499: LaTeX parser should parse text in boxes
I've been using Pandoc quite a bit for LaTex -> docx conversions. It works really well; thanks for working on this!

I have one little bug to report. When running Pandoc on Windows 10 to convert `.tex` to `.docx`, Pandoc misses `\mbox{}` blocks that are used to keep text from line breaking (useful for abbreviations and such).

Example:
```
\documentclass{article}
\begin{document}
text \mbox{missing text} more text
\end{document}
```

Pandoc script example:
`Pandoc.exe -o test.docx test.tex`

Expected result:
`text missing text more text`

Pandoc produces:
`text more text`

Version info:
```
Pandoc.exe 1.19.2.1
Compiled with pandoc-types 1.17.0.4, texmath 0.9, skylighting 0.1.1.4
```

vasalf:
I tried to fix behaviour for mbox and hbox as my first issue in pandoc.


src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -893,6 +894,8 @@ inlineCommands = M.union inlineLanguageCommands $ M.fromList
   , ("cref", rawInlineOr "cref" $ doref "ref")       -- from cleveref.sty
   , ("vref", rawInlineOr "vref" $ doref "ref+page")  -- from varioref.sty
   , ("eqref", rawInlineOr "eqref" $ doref "eqref")   -- from amsmath.sty
+  , ("mbox", rawInlineOr "mbox" $ spanWith ("",["mbox"],[]) . dropLineBreaks <$> tok)
+  , ("hbox", rawInlineOr "hbox" $ spanWith ("",["hbox"],[]) . dropLineBreaks <$> tok)
```
jgm: What's the rationale for `dropLineBreaks` here?

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -1287,6 +1290,12 @@ rawInlineOr name' fallback = do
      then rawInline "latex" <$> getRawCommand name' ("\\" <> name')
      else fallback
 
+dropLineBreaks :: Inlines -> Inlines
+dropLineBreaks = Many . Data.Sequence.filter isNotLineBreak . unMany
+    where
```
jgm: Do you want to remove only line breaks at the top level, or nested ones too? (e.g. a LineBreak inside an Emph)

src/Text/Pandoc/Readers/LaTeX.hs
vasalf: `hbox` stands for «Horizontal box» in TeX. They do not only drop the line breaks here but they also prohibit any line wraps inside the box. `mbox` is some kind of LaTeX analog for `hbox`.

I made a few examples in this screenshot.
The first two lines stand for explicit line breaks inside `mbox` and `hbox`, respectively.
The third contains an explicit line break inside an `Emph`, this is for another comment.
The last is a very long paragraph completely inside `mbox`, so one is not able to see its end because it doesn't fit into one line.

Screenshot from 2019-06-16 09-16-46

src/Text/Pandoc/Readers/LaTeX.hs
vasalf: Thanks for pointing this out. I missed this case. I **do** want to remove the nested line breaks. For example, see the third line in the screenshot I sent to the previous conversation.

vasalf:
@jgm thanks for the review! I sent some comments on the line breaks inside the boxes and I'm definitely going to fix the nested line breaks issue.

Speaking of `span` outside the box. At least, those boxes have special effect of ignoring any explicit line breaks and preventing line wraps. It seems that this is not a TeX-specific feature. For example, it is possible in HTML+CSS:

```html
<p><span style="white-space: nowrap">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum</span></p>
```

This results to a very long line without line breaks at least in Google Chrome. Maybe we could add this feature to some of the writers that support this?

vasalf:
@jgm I implemented an alternative solution proposed to be by @tarleb in person. Now there are no spans. It only converts all spaces and soft breaks to a no-break space character. This might not be an elegant solution but it seems to work perfectly well.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -1287,6 +1289,14 @@ rawInlineOr name' fallback = do
      then rawInline "latex" <$> getRawCommand name' ("\\" <> name')
      else fallback
 
+processHBox :: Inlines -> Inlines
+processHBox = walk convert
+  where
+    convert Space     = Str [chr 160] -- non-breakable space
```
jgm: I think it's somewhat nicer to just write

    Str "\160"

jgm:
Thanks for this contribution!
I hope it won't be the first.

PR #5587, opened 2019-06-15
Support epigraph command in LaTeX Reader. Fix #3523

### LINKED ISSUE #3523: Pandoc does not recognise epigraphs from epigraph-package
I can successfully convert from .tex to .docx, but pandoc does not recognise or convert my epigraphs. They are simply lost.

MWE:

```
\documentclass{book}
\usepackage{epigraph}
\begin{document}

\epigraph{Das Recht hat kein Dasein f{\"u}r sich, sein Wesen vielmehr ist das Leben des Menschen selbst, von einer besonderen Seite angesehen.}{Friedrich Carl von Savigny}

\end{document}
```

And this for conversion:

`pandoc -f latex -t docx test-epigraph-conversion-pandoc.tex -o test-epigraph-conversion-pandoc.doc`

Version is: pandoc 1.19.2.1

(I have also asked about this here: http://tex.stackexchange.com/questions/357779/usepackageepigraph-with-pandoc-conversion-to-docx)

oquechy:


src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -1640,6 +1640,12 @@ looseItem = do
   skipopts
   return mempty
 
+epigraph :: PandocMonad m => LP m Blocks
+epigraph = do
+  p1 <- grouped blocks
+  p2 <- grouped blocks
+  return $ divWith ("", ["epigraph"], []) (fromList $ concat $ map toList [p1, p2])
```
jgm: Can't `fromList $ concat $ map toList [p1, p2]` be replaced with `p1 <> p2`?

oquechy: Definitely. Fixed!

jgm:
Looks good, thanks!

PR #5589, opened 2019-06-16
Add support for EPUB2 covers (fix #3992)

### LINKED ISSUE #3992: Pandoc does not extract cover from epub
pandoc.EXE 1.19.2.1, windows 10

command line:
pandoc.exe" "C:\tmp\1632.epub" --extract-media="1632_files" -o "C:\tmp\1632.markdown"
pandoc: extracting 1632_files\067157849901.jpg
pandoc: extracting 1632_files\067157849902.jpg

Result: images and text are extracted, but cover image is not extracted.
Expected result: extraction of the cover image.

I downloaded book from http://www.baen.com/1632.html, in epub format.

When I open epub file using 7zip, there is 0671578499_Cover.jpg, and 0671578499_top.htm inside.

This happens with some other epub files too, and also when converting epub to other formats, like docx.
(When using kindlegen, cover image is supported, and added to destination format, although cover.htm is not supported.)

Thank you.

blmage:


mb21:
Thanks! Unfortunately, seems like somebody tackled the same issue at almost the same time: https://github.com/jgm/pandoc/pull/5592

That pull seems a bit shorter and includes a test, or is it missing some other case?

blmage:
From what I could gather it seems to be missing those two cases:
* EPUB2 covers may not be present in the manifest ([see example here](https://fr.flossmanuals.net/creer-un-epub/epub-2/))
* the ID of the manifest item may begin with something else than `cover` (and can/should rather be inferred from the content of the `cover` meta)

Also, the new test does not seem to be specifically aimed towards EPUB2 documents (`wasteland.epub` uses EPUB3 format)

martinfrancois:
Hi @blmage and @mb21
That's indeed a bit unfortunate...
Thanks for pointing out wastelands being EPUB3, I overlooked that it also contained a `cover-image` property, I only saw the old meta tag being present.
I think while your implementation looks to be a bit more complex, it could be covering more cases than mine.
If you agree, I would offer adding some test cases to this PR with a few different cases so we could collaborate and get a broader picture overall, what do you think?

blmage:
@martinfrancois Sorry for any inconvenience (I sure was not expecting this when I started working on it), and thanks for your proposition!

I've just added you as a collaborator on [my fork](https://github.com/blmage/pandoc/), so feel free to push some tests as you see fit!

martinfrancois:
@blmage nevermind, I wasn't expecting that as well as you could imagine 😄.

Thanks for taking up my offer, I just added some tests and your implementation is solid, well done! Now we have the best of both worlds in one PR 😃. I'll be closing mine in favor of this one.

jgm:
Thanks, all!

PR #5593, opened 2019-06-17
docbook: richer parse for admonitions

### LINKED ISSUE #1234: DocBook: admonition-like environments
Hello,

Docbook has a handy way to create text blocks. The current way Pandoc translates them makes it very difficult to recover them through XML post-processing.
# Text Flow
## Asciidoc

``` adoc
Regular paragraph.

[TIP]
====
.Title of my tip
A paragraph inside a tip.
====

Another regular paragraph.
```
## DocBook

``` docbook
<simpara>Regular paragraph.</simpara>
<tip>
  <formalpara>
    <title>Title of my tip</title>
    <para>A paragraph inside a tip.</para>
  </formalpara>
</tip>
<simpara>Another regular paragraph.</simpara>
```
## OpenDocument (from pandoc)

``` opendocument
<text:p text:style-name="Text_20_body">Regular paragraph.</text:p>

<text:p text:style-name="P11"><text:span text:style-name="T279">Tip</text:span></text:p>
<text:p text:style-name="P11"><text:span text:style-name="T282">Title of my tip</text:span>.<text:line-break/>A paragraph inside a tip.</text:p>

<text:p text:style-name="Text_20_body">Another regular paragraph.</text:p>
```
# Ideal solution

``` opendocument
<text:p text:style-name="Text_20_body">Regular paragraph.</text:p>

<text:h text:style-name="ADMONITION11 ADMONITITION-TIP">Title of my tip</text:h>
<text:p text:style-name="ADMONITION11 ADMONITITION-TIP">A paragraph inside a tip.</text:p>

<text:p text:style-name="Text_20_body">Another regular paragraph.</text:p>
```

Something like that.

How feasible is it? It would help a lot for the code blocks as well, as they share a very similar logic (which is going to be a next issue).

Thanks very much :-)

michaelpj:
Fixes #1234.

This parses admonitions not as a blockquote, but rather as a div with an
appropriate class. We also handle titles for admonitions as a nested div
with the "title" class.

(I followed the behaviour of other docbook-to-html converters in this -
there are clearly other ways you could encode it.)

In general, the handling of elements with nested title elements is very
inconsistent. I think we should make it consistent, but I'm leaivng that
for later to make this a small change.

Example:
```docbook
<warning xml:id="someId">
<title>My title</title>
<simpara>An admonition block</simpara>
</warning>
```

goes to

```html
<div id="someId" class="warning">
<div class="title">My title</div>
<p>An admonition block</p>
</div>
```

### DIFF
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -619,19 +619,23 @@ instance HasMeta DBState where
   deleteMeta field s = s {dbMeta = deleteMeta field (dbMeta s)}
 
 isBlockElement :: Content -> Bool
-isBlockElement (Elem e) = qName (elName e) `elem` blocktags
-  where blocktags = ["toc","index","para","formalpara","simpara",
+isBlockElement (Elem e) = qName (elName e) `elem` blockTags
+isBlockElement _ = False
+
+blockTags :: [String]
+blockTags = ["toc","index","para","formalpara","simpara",
            "ackno","epigraph","blockquote","bibliography","bibliodiv",
            "biblioentry","glossee","glosseealso","glossary",
            "glossdiv","glosslist","chapter","appendix","preface",
            "bridgehead","sect1","sect2","sect3","sect4","sect5","section",
-           "refsect1","refsect2","refsect3","refsection",
-           "important","caution","note","tip","warning","qandadiv",
+           "refsect1","refsect2","refsect3","refsection", "qandadiv",
            "question","answer","abstract","itemizedlist","orderedlist",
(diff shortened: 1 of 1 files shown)


mb21:
Thanks! Maybe you could add a test case to `tests/commands/1234.md` ?

jgm:
For comparison, pandoc's current output for this is
```html
<blockquote>
<p><strong>Warning</strong></p>
<p>An admonition block</p>
</blockquote>
```

This has the advantage of giving a better render out of the box, without special CSS, but in other respects it's inferior.

jgm:
Thanks!

Issue #5570, opened 2019-06-10
JATS validation fails if footnotes include block quotes


coryschires:
## Background
The JATS spec does not allow you to have block quotes (i.e. `<disp-quote>`) inside footnotes (i.e. `<fn>`).

Frankly, I think this is an odd and unreasonable restriction. Some authors – and some disciplines, such as legal scholarship – make heavy use of footnotes. As far as I know, there's nothing fundamentally wrong with placing a block quote inside a footnote.

## Problem
I am encountering real-world examples of this problem, so I need some sort of workaround.

To be clear, here's an example of invalid JATS:

    <fn>
      <p> ... </p>
      <disp-quote> ... </disp-quote>
    </fn>

This JATS XML would fail validation with the error: `Element fn content does not follow the DTD, expecting (label? , p+), got (p disp-quote)`

### Steps to recreate
**Pandoc version:** 2.7.2
**Files:** [jats_example.zip](https://github.com/jgm/pandoc/files/3273824/jats_example.zip)

#### To reproduce the issue described above
1. `pandoc -s --metadata-file metadata.json --to jats example.md -o output.xml`
2. Validate `output.xml` using the PMC online validation tool: https://www.ncbi.nlm.nih.gov/pmc/tools/xmlchecker
3. The validation tool will display the errors described above

## Solution
After examining the JATS spec, I think I have a solid workaround. I want to wrap the `<disp-quote>` element in a `<p>` element. This will ensure the JATS is valid while while only minimally changing the semantic meaning. Unless it's too much work, it would be nice if we could also include a [`specific-use` attribute](https://jats.nlm.nih.gov/publishing/tag-library/1.1/attribute/specific-use.html).

So in practice, we would convert this:

    <fn>
      <p> ... </p>
      <disp-quote> ... </disp-quote>
    </fn>

Into this:

    <fn>
      <p> ... </p>
      <p specific-use="wrapper">
        <disp-quote> ... </disp-quote>
      </p>
    </fn>

[...]

PR #5595, opened 2019-06-18
JATS Writer: do not emit empty <fn-group>

mb21:
see https://github.com/jgm/pandoc/issues/5570#issuecomment-502861382

introduced by e87b54dcad5e37133bc0f4cfc8039e9fd0dd1b4e


jgm:
Sorry about that one!

mb21:
That's why we have users that report bugs ;)

PR #5605, opened 2019-06-20
Add font-size attribute to customize font size of code blocks in PowerPoint slide shows

jeroendehaas:
So I have been tinkering some more with Pandoc's PowerPoint writer and added the possibility to customize the font size of individual code blocks in PowerPoint slide shows. To this end, I've added a font-size attribute.

I'm offering this as a pull request as I feel it might be useful to other people. This makes Pandoc suitable for developing lectures on computer programming at my university. 

Tested on PowerPoint 365 for Windows 1808 (Build 10730.20344 Click-To-Run)


jeroendehaas:
I was curious if there's any feedback on this pull request. We have been using this patch for over three months internally. It produces presentations that render fine under recent versions of PowerPoint for both Windows and macOS as well as LibreOffice Impress.

If any changes are required to have this merged, I am willing to work on them.

jgm:
The PR looks fine; it's really just a question of whether the feature is one we should add.
@jkr I'd be interested in your feedback on this.
It's slightly inelegant to support a `font-size` field in code blocks that only works in one output format, and I worry that people will have misguided expectations that it will work more generally.

andy-5:
I just found this PR because I have the exact same problem/use case. I understand that adding a attribute just for pptx output is inelegant, but the need to change font size is quite common, I think. 

For other output formats like html or pdf, the workaround seems to be to use raw html, css classes or raw latex code snippets to modify the font size. But I think such a workaround is not currently possible for pptx (e.g. embed raw xml)?

An alternative solution could be to specify a class or custom style in markdown and then provide pptx specific options/metadata to define the font used for this style/class. This way, there are no pptx specific fields in the markdown source. For example, something like
```markdown
---
pptx-style:
  - small-code: Courier, 14pt
---

~~~{.small-code}
// some long code snippet
~~~
```

Or maybe some common flags to change the font-size relative to the current size, e.g. `smaller`, `larger`, could be added which could eventually also be used by other output formats?

Either way, I would appreciate some way to change the font size locally.

jgm:
Well, I guess since the previous commit already added ppt-specific `monofont`, this isn't doing anything much different.

mb21:
Seems like custom styles would be more the semantic pandoc way... but for some reason that wasn't working for powerpoint?

Issue #5608, opened 2019-06-22
ConTeXt writer: improve PDF/A output


mb21:
Moved from #5409:

> mb21: @klpn Any downsides when using your gist? If not, would you like to make a pull request? For context: the `pdfa` tempalte variable was added in https://github.com/jgm/pandoc/commit/46f4238a2a40b5542612bc745e63ce503ce12a32

> klpn: I have not discovered any problems, but I should perhaps test with some more documents. However, the Pandoc manual explicitly states that the `pdfa` variable "adds to the preamble the setup necessary to generate PDF/A-1b:2005", so this should then be changed as well, if we want to always use 2a (i.e. version 2, level A conformance). When using PDF/A for documents born digital, it is best to use level A (which includes Unicode mapping and tagging) if possible, rather than B, but some older preservation guidelines still require version 1. Perhaps, the `pdfa` variable should be changed so that the user can choose which version of PDF/A to use ([different PDF:s supported by ConTeXt](https://source.contextgarden.net/tex/context/base/mkiv/lpdf-fmt.lua?search=pdf-fmt))?

mb21:
@klpn sounds good! The template could be sensitive to two variables: `pdfa1b` and `pdfa2a`. For backwards compatibility, the ConTeXt writer would set `pdfa1b` to true if `pdfa` was set.

@hmenke what do you think?

klpn:
Or perhaps, we could leave the boolean `pdfa` variable as it is (for compatibility), and add a new `pdfatype`, whose value could be specified as `1a:2005`, `1b:2005`, `2a` and so on (if `pdfa` is not specified). Something like this in the template:
```
$if(pdfa)$
% attempt to generate PDF/A
\setupbackend
  [format=PDF/A-1b:2005,
   intent=sRGB IEC61966-2.1,
   profile=sRGB.icc]
$else$
$if(pdfatype)$
\setupbackend
  [format=PDF/A-$pdfatype$,
   intent=sRGB IEC61966-2.1,
   profile=sRGB.icc]
$endif$
$endif$
```

mb21:
Or the Context writer could set the `pdfa` variable to `1b:2005` if it was to `true`, then:

    $if(pdfa)$
    % attempt to generate PDF/A
    \setupbackend
      [format=PDF/A-$pdfa$,
       intent=sRGB IEC61966-2.1,
       profile=sRGB.icc]
    $endif$

klpn:
[...]

Issue #5409, opened 2019-03-29
Accessibility mode for LaTeX


jgm:
PDFs produced using latex are not accessible.
We could introduce a command-line option that causes the latex writer to include annotations for math (perhaps using the unicode fallback or even raw tex), image alt text, and more:
http://ctan.math.washington.edu/tex-archive/macros/latex/contrib/oberdiek/accsupp.pdf

This package also includes an option that makes spaces visible to copy and paste (often when you copy from a latex-compiled PDF, spaces disappear).

Structural elements (paragraphs, lists, etc.) need to be tagged, and reading order indicated.

See also:
https://www.tug.org/twg/accessibility/
http://web.science.mq.edu.au/~ross/TaggedPDF/
https://tex.stackexchange.com/questions/124291/revisiting-producing-structured-pdfs-from-latex (with information about using ConTeXT)

frastlin:
Note that tagged PDFs are starting to be required at confrances such as [SIG Access](https://assets19.sigaccess.org/creating_accessible_pdfs.html) and [ICAD](https://icad2019.icad.org/call-for-participation/)
Governments around the world, such as the [United States,](https://www.hhs.gov/web/section-508/making-files-accessible/checklist/pdf/index.html) [Ontario,](https://www.aoda.ca/the-four-levels-of-pdf-accessibility/) [Australia,](https://www.visionaustralia.org/services/digital-access/blog/10-12-2014/pdf-accessibility-update-from-australian-government) [European Union,](https://accessible-digital-documents.com/blog/the-eu-accessibility-directive-and-pdf/) and many other governments all require at minimum, all government PDFs to be properly tagged. This means any university receiving government money in the U.S. needs to have [all their content](https://www.essentialaccessibility.com/blog/section-508-higher-education/) be [WCAG](https://www.w3.org/WAI/standards-guidelines/wcag/glance/) compliant.
This means that if Pandoc has no way to produce properly tagged PDFs, it will not be legally usable by any institution that falls under the above mandates.
I would rate this as an extremely high priority as the U.S. started requiring accessible PDFs from all government and entities receiving government money in January 2018 and EU started requiring any government sector website to have only accessible PDFs produced starting on September 23 2018. So millions of PDFs are effected by these requirements.

jgm:
Agreed, it's an important issue.  It also comes up for
materials distributed in connection with courses.  I'm
motivated to make it easier to produce accessible PDFs
using pandoc, but I need some guidance on the LaTeX side.

[...]

PR #5610, opened 2019-06-22
Customizable type of PDF/A for the ConTeXt writer (issue #5608)

klpn:
The type of PDF/A produced by the ConTeXt writer can be changed so that e.g. a PDF/A-2A can be produced by giving `pdfa=2a`, as discussed under [#5608](https://github.com/jgm/pandoc/issues/5608). If `pdfa` is given without value, a PDF/A-1B is produced, in accordance with the old behavior of the variable. For a list of supported types of PDF/A, see the formats beginning with `pdf/a` in [the ConTeXt source](https://source.contextgarden.net/tex/context/base/mkiv/lpdf-fmt.lua?search=pdf-fmt).

To  fix issues with tagging (which is a requirement for level A PDF/A conformance), the ConTeXt template is changed according to [discussion under #5409](https://github.com/jgm/pandoc/issues/5409#issuecomment-504637400).

### DIFF
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -27,7 +27,6 @@ $endif$
 % make chapter, section bookmarks visible when opening document
 \placebookmarks[chapter, section, subsection, subsubsection, subsubsubsection, subsubsubsubsection][chapter, section]
 \setupinteractionscreen[option=bookmark]
-\setuptagging[state=start]
 
 $if(papersize)$
 \setuppapersize[$for(papersize)$$papersize$$sep$,$endfor$]
@@ -41,10 +40,13 @@ $endif$
 $if(pdfa)$
 % attempt to generate PDF/A
 \setupbackend
-  [format=PDF/A-1b:2005,
+  [format=PDF/A-$pdfa$,
    intent=sRGB IEC61966-2.1,
    profile=sRGB.icc]
 $endif$
+
+\setupbackend[export=yes]
+\setupstructure[state=start,method=auto]
 
 % use microtypography
 \definefontfeature[default][default][script=latn, protrusion=quality, expansion=quality, itlc=yes, textitalics=yes, onum=yes, pnum=yes]
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -44,7 +44,6 @@ $if(pdfa)$
    intent=sRGB IEC61966-2.1,
    profile=sRGB.icc]
 $endif$
-
 \setupbackend[export=yes]
 \setupstructure[state=start,method=auto]
 
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -41,8 +41,8 @@ $if(pdfa)$
 % attempt to generate PDF/A
 \setupbackend
   [format=PDF/A-$pdfa$,
-   intent=sRGB IEC61966-2.1,
-   profile=sRGB.icc]
+   profile={default_cmyk.icc,default_rgb.icc,default_gray.icc},
+   intent=ISO coated v2 300\letterpercent\space (ECI)]
 $endif$
 \setupbackend[export=yes]
 \setupstructure[state=start,method=auto]
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -41,8 +41,8 @@ $if(pdfa)$
 % attempt to generate PDF/A
 \setupbackend
   [format=PDF/A-$pdfa$,
-   profile={default_cmyk.icc,default_rgb.icc,default_gray.icc},
-   intent=ISO coated v2 300\letterpercent\space (ECI)]
+   profile={$if(pdfaiccprofile)$$for(pdfaiccprofile)$$pdfaiccprofile$$sep$,$endfor$$else$sRGB.icc$endif$},
+   intent=$if(pdfaintent)$$pdfaintent$$else$sRGB IEC61966-2.1$endif$]
 $endif$
 \setupbackend[export=yes]
 \setupstructure[state=start,method=auto]
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -15,7 +15,7 @@ Conversion of 'Pandoc' format into ConTeXt.
 module Text.Pandoc.Writers.ConTeXt ( writeConTeXt ) where
 import Prelude
 import Control.Monad.State.Strict
-import Data.Char (ord, isDigit)
+import Data.Char (ord, isDigit, toLower)
 import Data.List (intercalate, intersperse)
 import Data.Maybe (mapMaybe)
 import Data.Text (Text)
@@ -94,8 +94,8 @@ pandocToConTeXt options (Pandoc meta blocks) = do
                           | all isDigit (d:ds) -> resetField "papersize"
                                                      (('A':d:ds) :: String)
                         _                     -> id)
-                $ (case getField "pdfa" metadata of
-                        Just ("true" :: String) -> resetField "pdfa" ("1b:2005" :: String)
+                $ (case toLower <$> lookupMetaString "pdfa" meta of
+                        "true" -> resetField "pdfa" ("1b:2005" :: String)
                         _                     -> id) metadata
   let context' = defField "context-dir" (toContextDir
                                          $ getField "dir" context) context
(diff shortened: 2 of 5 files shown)


data/templates/default.context
```diff
@@ -41,10 +40,12 @@ $endif$
 $if(pdfa)$
 % attempt to generate PDF/A
 \setupbackend
-  [format=PDF/A-1b:2005,
+  [format=PDF/A-$pdfa$,
    intent=sRGB IEC61966-2.1,
    profile=sRGB.icc]
 $endif$
+\setupbackend[export=yes]
```
hmenke: This just exports the document to XML.  I'm pretty sure that this is not needed to get tagging right, unless you want to embed the XML into the PDF.

klpn: I tried removing it from the template, but then the Poppler `pdfinfo --struct-text` yields empty output.

klpn:
Should not the color profile settings in the template also be updated in accordance with [the ConTeXt wiki on PDF/A](https://wiki.contextgarden.net/PDF/A), which has not been changed since 2014, and to which the Pandoc documentation refers?
```
\setupbackend[
format={pdf/a-1b:2005}, % or pdf/a-1a:2005
profile={default_cmyk.icc,default_rgb.icc,default_gray.icc},
intent=ISO coated v2 300\letterpercent\space (ECI)]
```
If you install just the ICC profiles referred to in that wiki, the default template, which uses `sRGB.icc` will not produce valid PDF/A with colored content.

hmenke:
That implies that you are printing on paper of type ISO coated v2 300% (ECI).  The sRGB profile is for displaying the PDF on screen which I think is the most common nowadays.

klpn:
Perhaps the color profiles and output intent should be configurable with variables too (with sRGB as fallback)?

data/templates/default.context
mb21: I also tested this on macOS using `context 2018.04.04`. Without `-M pdfa=true`, the output of `pdfinfo -struct-text` seems always correct:

```
Div
  Sect "section"
    Div
      H (block)
        "hi there"
    Div
      "how are things going today?"
```

But when setting the `pdfa` variable, the `\setupbackend[export=yes]` seems to be needed, otherwise I get:

```
Div
  Sect "section"
    Div
      H (block)
        "hithere"
    Div
      "howarethingsgoingtoday?"
```

btw. I'm curious, what is the syntax of that output? is it part of the pdf standard and implemented by context, or is it just how `pdfinfo` displays it?

data/templates/default.context
```diff
@@ -27,7 +27,6 @@ $endif$
 % make chapter, section bookmarks visible when opening document
 \placebookmarks[chapter, section, subsection, subsubsection, subsubsubsection, subsubsubsubsection][chapter, section]
 \setupinteractionscreen[option=bookmark]
-\setuptagging[state=start]
```
mb21: this is not needed anymore?

src/Text/Pandoc/Writers/ConTeXt.hs
```diff
@@ -93,6 +93,9 @@ pandocToConTeXt options (Pandoc meta blocks) = do
                         Just (('a':d:ds) :: String)
                           | all isDigit (d:ds) -> resetField "papersize"
                                                      (('A':d:ds) :: String)
+                        _                     -> id)
+                $ (case getField "pdfa" metadata of
+                        Just ("true" :: String) -> resetField "pdfa" ("1b:2005" :: String)
                         _                     -> id) metadata
   let context' = defField "context-dir" (toContextDir
```
mb21: This should probably be written as:

```haskell
$ (if lookupMetaBool "pdfa" metadata
   then resetField "pdfa" "1b:2005"
   else id) metadata
```

data/templates/default.context
klpn: No, the tagging seems to be controlled by the `setupstructure` setting, regardless of whether or not the output is PDF/A.

src/Text/Pandoc/Writers/ConTeXt.hs
```diff
@@ -93,6 +93,9 @@ pandocToConTeXt options (Pandoc meta blocks) = do
                         Just (('a':d:ds) :: String)
                           | all isDigit (d:ds) -> resetField "papersize"
                                                      (('A':d:ds) :: String)
+                        _                     -> id)
+                $ (case getField "pdfa" metadata of
+                        Just ("true" :: String) -> resetField "pdfa" ("1b:2005" :: String)
                         _                     -> id) metadata
```
mb21: Pandoc metadata can be set using a confusing number of ways... this only checks for the string "true", but we also have actual booleans... but this should do the trick:

```haskell
import Data.Char (ord, isDigit, toLower)
-- rest of the file...

                $ (case toLower <$> (lookupMetaString "pdfa" meta) of
                        "true" -> resetField "pdfa" ("1b:2005" :: String)
```

jgm: You could use `lookupMetaBool`

mb21: That's what I thought at first, but the `pdfa` variable in this pull might contain a string now. (like `"1b:2005"`).

klpn: When I try this, the `resetField` condition is never triggered, so that `-V pdfa` gives a literal `PDF/A-true` in the context document.  Note that `lookupMetaString` has type `String -> Meta -> String`, while `getField` is `FromJSON a => String -> Value -> Maybe a`.

mb21: I believe `-V` didn't work with your version either? It should be `-M` see e.g. https://github.com/jgm/pandoc/blob/master/doc/customizing-pandoc.md#template-variables

mb21: uh, yes... the `metadata` has the template variables (`-V`) already merged in, while the `meta :: Meta` only has the metadata variables (`-M`), see e.g. https://github.com/jgm/pandoc/blob/master/doc/customizing-pandoc.md#template-variables

I think it's okay if `-V pdfa` doesn't work, as long as `-M pdfa` and the following work:

```
---
pdfa: true
---

my markdown doc...
```

This is a bit confusing for the users, but we already have lots of places where `-V` doesn't work.

klpn: Yes, both `-M pdfa` and `pdfa: true` in a YAML metadata block work.

mb21: The following does it all, but it's really not very nice code...

```haskell
  let resetPDFA = resetField "pdfa" ("1b:2005"::String)
  let context =   ...

                $ (case getField "pdfa" metadata of
                        Just True             -> resetPDFA
                        _                     -> id)
                $ (case getField "pdfa" metadata of
                        Just ("true" :: String) -> resetPDFA
                        _                     -> id) metadata
```

I don't know... I'll let @jgm make a judgement call ;-)

mb21: Apart from this, I think this pull is good to merge, as hmenke's concerns seem to have been addressed if I understand correctly. Thanks a lot!

klpn: `-V pdfa=2a` will work regardless of what option is chosen here, and this should be the normal use case. The support for boolean `pdfa` (which necessitates the changes in the Context writer), is for backwards compatibility. Perhaps I should clarify the documentation on that.

mb21: @jgm what do you think? apart from this minor issue, I think this is good to merge!

klpn: I used the [first suggestion](https://github.com/jgm/pandoc/pull/5610#discussion_r300665428) made by @mb21 in the [last commit](https://github.com/jgm/pandoc/pull/5610/commits/e5c5bf12b11302a7d2cf5856a447a5fd861330fb). I have not found a solution that requires less code than [the later suggestion](https://github.com/jgm/pandoc/pull/5610#discussion_r300712630), and also works with `-V pdfa`, but I updated the manual to clarify on that.

jgm:
Thanks, all!

PR #5625, opened 2019-07-01
Fix redundant constraint warnings.

peteryland:
Removes compiler warnings about redundant constraints (-Wredundant-constraints).


jgm:
If I recall, I left these redundant constraints in because they were needed for some older ghc versions.
It may be that we no longer officially support these versions, so if the CI passed, it's probably okay!

peteryland:
I'm also intending to submit a PR with hlint fixes.  Would that be welcome too?  My main intention is actually to get familiar with the code and then work on a reader for AsciiDoc.

jgm:
In general I like most, but not all, of hlint's suggestions.  So in many cases we have intentionally disregarded hlint's suggestions (for example, keeping redundant parentheses when these make grouping more perspicuous to the reader).

Some hlint-suggested changes may be a good idea, but they'd have to be assessed case by case.

jgm:
Thanks!

PR #5628, opened 2019-07-04
Markdown reader: handle inline code more eagerly within lists.

### LINKED ISSUE #5627: Fenced code blocks not escaped when converting Markdown to HTML5
Hey,

I'm trying to convert example.md from Markdown to HTML using pandoc. File is attached.
[example.md.txt](https://github.com/jgm/pandoc/files/3347158/example.md.txt)

The command I'm using is:
`pandoc --from markdown -t html5 -o 'out.html' 'example.md'`

The version I'm using is:
```
pandoc --version
pandoc 2.7.3
Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.8.1
Default user data directory: /Users/user/.local/share/pandoc or /Users/user/.pandoc
Copyright (C) 2006-2019 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose
```

The problem I'm running into is that the fenced code is not escaped as HTML, which makes the fenced JavaScript code execute when viewing the document.

I guess this is because I'm a horrible person using HTML comment tags within the escaped sections, though I would expect pandoc to handle this since they are inside the fenced areas.

Could you please confirm if this is a bug, and if there is any workaround? I can't escape the characters within the fences because that would cause pandoc to double-encode them :(.

leungbk:
Closes #5627.

### DIFF
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -857,7 +857,8 @@ listLine continuationIndent = try $ do
 
 listLineCommon :: PandocMonad m => MarkdownParser m String
 listLineCommon = concat <$> manyTill
-              (  many1 (satisfy $ \c -> c /= '\n' && c /= '<')
+              (  many1 (satisfy $ \c -> c `notElem` ['\n', '<', '`'])
+             <|> fmap snd (withRaw code)
              <|> fmap snd (htmlTag isCommentTag)
              <|> count 1 anyChar
               ) newline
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -857,7 +857,8 @@ listLine continuationIndent = try $ do
 
 listLineCommon :: PandocMonad m => MarkdownParser m String
 listLineCommon = concat <$> manyTill
-              (  many1 (satisfy $ \c -> c /= '\n' && c /= '<')
+              (  many1 (satisfy $ \c -> c `notElem` ['\n', '<', '`'])
```
leungbk: I don't understand why the `many1` doesn't need to be wrapped in a `try`.

leungbk: Never mind, I see it.

leungbk:
Found some oddball cases that weren't covered, I'll try to fix them up right now.

leungbk:
OK, this should be ready for review.

jgm:
Great, thanks.

PR #5637, opened 2019-07-08
RST reader: keep `name` property in `imgAttr`.

### LINKED ISSUE #5619: Readers.RST should keep the figure `name` option to allow using pandoc-crossref and pandoc-fignos with rst input
[pandoc-crossref](https://github.com/lierdakil/pandoc-crossref) and [pandoc-fignos](https://github.com/tomduck/pandoc-fignos) are great tools that enable numbering figures and tables and referencing them. Unfortunately, they only work for markdown input.

However, I've become convinced that it would be quite easy to allow pandoc RST input users to allow numbering figures and tables. All is needed for this is for the RST reader to keep the id option for these directives.

Let's take a small md fragment from the fignos demo:

```
Reference to @fig:1.

![The number one.](img1.jpg){#fig:1 width=1in}
```

We can look at the equivalent native code with `pandoc fignos_demo.md -o fignos_demo_md.native`:

```
[Para [Str "Reference",Space,Str "to",Space,Cite [Citation {citationId = "fig:1", citationPrefix = [], citationSuffix = [], citationMode = AuthorInText, citationNoteNum = 0, citationHash = 0}] [Str "@fig:1"],Str "."]
,Para [Image ("fig:1",[],[("width","1in")]) [Str "The",Space,Str "number",Space,Str "one."] ("img1.jpg","fig:")]
]
```

And this works like expected (the figure is numbered and the reference becomes a link) with fignum and crossref, like we can test with `pandoc -F pandoc-fignos fignos_demo_md.native -o fignos_demo.html`

The roughly equivalent rst would be (without the citation, because it seems Readers.RST does not have support for that):

```
.. figure:: img1.jpg
  :name: The number one.
  :fig: 1
  :width: 1in

  The number one.
```

Which with pandoc 2.7 becomes the following native code:

```
[Para [Image ("",[],[("width","1in")]) [Str "The",Space,Str "number",Space,Str "one."] ("img1.jpg","fig:")]]
```

Note that there is no Image id (no more "fig:1" after Image), which stops fignos or crossref from working with rst input. Simply adding it to the native format enables figure numbering!

This is obviously related to the great discussion about internal links in https://github.com/jgm/pandoc/issues/813

It would be great if the RST reader keeps the "fig:1" after Image (not 100% sure if id is the right name for this attribute), so that the native output from parsing this rst would include it. Then, fignos would be capable of numbering rst figures too. References to those figures still would not work, I think, without more work on rst citations, but one fix at a time.

I would love to propose a patch, but I don't know enough Haskell.

The obvious place to patch is https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Readers/RST.hs#L727

There are several places in Readers.Markdown that use imageWith, but I could not be sure which was the right one to use for inspiration. Don't hesitate to point it out please.

The equivalent work could be done for tables too.

leungbk:
Closes #5619.

### DIFF
--- a/src/Text/Pandoc/Readers/RST.hs
+++ b/src/Text/Pandoc/Readers/RST.hs
@@ -645,7 +645,7 @@ directive' = do
       name = trim $ fromMaybe "" (lookup "name" fields)
       classes = words $ maybe "" trim (lookup "class" fields)
       keyvals = [(k, trim v) | (k, v) <- fields, k /= "name", k /= "class"]
-      imgAttr cl = ("", classes ++ alignClasses, widthAttr ++ heightAttr)
+      imgAttr cl = (name, classes ++ alignClasses, widthAttr ++ heightAttr)
         where
           alignClasses = words $ maybe "" trim (lookup cl fields) ++
                           maybe "" (\x -> "align-" ++ trim x)
--- a/test/command/5619.md
+++ b/test/command/5619.md
@@ -1,0 +1,10 @@
+```
+% pandoc -f rst -t native
+.. figure:: img1.jpg
+  :width: 1in
+  :name: test
+
+  The caption. Here's what piggybacking on caption would look like {#fig:1}
+^D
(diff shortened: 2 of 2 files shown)


jgm:
Looks good, thanks!

PR #5639, opened 2019-07-09
Markdown sequential table

lspitzner:
This PR adds a new extension, named "sequential tables". See the included description in `Manual.txt`.

This implementation seems to do what it is supposed to, but I have not run the test-suite because `new-test` does not work. I'd have to re-compile the world, and I'd rather trust on your CI.

No test code was added either, because I could not find any for existing tables. Please point me in the direction if I have overlooked something. I think this also lacks some parsing code for the new extension. I hope you can point out necessary changes that I have overlooked.

I realize that I have somewhat skipped a step in your contributing guidelines, namely discussing this on pandoc-discuss. I do now know how to subscribe to that mailing list. But feel free to send me any sort of feedback to `hexagoxel@hexagoxel.de`.

I, the only copyright holder of the additions in this PR, release these additions under the GPL v2 license.

Thanks for creating and maintaining this great tool.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *~
 deb/.vagrant
 dist/*
+dist-newstyle/
 MANUAL.*
 !MANUAL.txt
 .configure-stamp
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3244,6 +3244,104 @@ you'll need to add colons as above.
 
 [PHP Markdown Extra tables]: https://michelf.ca/projects/php-markdown/extra/#table
 
+#### Extension: `sequential_tables` ####
+
+Sequential tables look like this:
+
+    ----sequentialtable
+    header
+    - One
+    - Two
+    align
+    - left
+    - right
+    width
+    - 10
+    - 10
+    row
+    - my
+    - table
+    row
+    - is
+    - nice
+    ----
+
+producing
+
+    One    Two
+    ----------
+    my   table
+    is    nice
+
+The most important advantage of this format over other markdown table
+parsers is that it makes handling complex or large table contents easy,
+without needing to resort to special editor functionality such as
+[Emacs table mode].
+
+----sequentialtable
+width
+- 10
+- 1
+- 10
+row
+- ~~~~.hs
+  test1 = evalContT $ do
+    resetT $ do
+      moveToNewThread
+      k -- must be `m ()` due to types
+    o   -- of resetT and moveToNewThread
+  ~~~~
+- ~
+- ~~~~.hs
+  test1 = do
+    void $ forkIO $ evalContT k
+    evalContT o
+  ~~~~
+----
+
+It might also be the only table format that allows nesting without much
+of a hassle:
+
+    ----sequentialtable
+    row
+    - ----sequentialtable
+      row
+      - a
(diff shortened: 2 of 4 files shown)


mb21:
Thanks for contributing to pandoc!

> I realize that I have somewhat skipped a step in your contributing guidelines, namely discussing this

Yeah, I've a feeling @jgm will not want to add another table extension, especially one that seems to conflict [markdown's philosophy](https://pandoc.org/MANUAL.html#philosophy):

> A Markdown-formatted document should be publishable as-is, as plain text, without looking like it’s been marked up with tags or formatting instructions.

But seems like a great use-case for a pandoc filter... there are already a few table-related ones in https://github.com/jgm/pandoc/wiki/Pandoc-Filters

lspitzner:
Do filters have access to pandoc's markdown parser?

mb21:
> Do filters have access to pandoc's markdown parser?

Good question! Well... Haskell filters kind of do. An example of mine: https://github.com/mb21/pandoc-placetable/#the-inlinemarkdown-flag (and [the code](https://github.com/mb21/pandoc-placetable/blob/master/pandoc-placetable.hs#L247)).

lspitzner:
But those disregard any reader options set in the "outer" parsing pass. And the option of nesting presumably also gets lost, not that it is very important.

You omit the very first sentence of the markdown philosophy:

> Markdown is designed to be easy to write, and, even more importantly, easy to read:

I argue that sequential tables are an improvement for both writing and reading tables in certain scenarios.

For writing, take the case of wanting to add a single line to one cell in an existing table. This essentially requires some special support from the editor to be somewhat automatic currently, while it is trivial with sequential tables (add a new line, indented by two spaces).

For reading, take the case of wanting to grab the contents of one cell to copy-paste it into some other document (a source file perhaps). With any sort of grid tables, you again end up having to resort to some sort of visual block mode or manually clip away the surrounding cells. With sequential tables, you select the relevant lines, remove the indentation (shift-tab in any editor suitable for general-purpose programming) and be good to go.

jgm:
I do think this goes against the markdown philosophy of prioritizing ease of reading, in original source form, over ease of writing.  (I think it's a stretch to consider ease of copy-pasting a kind of ease of reading.)

It also uses English keywords, which we've tried to avoid in extensions to pandoc's markdown.

Restructured Text has "list tables," which have similar advantages: http://docutils.sourceforge.net/docs/ref/rst/directives.html#list-table
List tables avoid the use of the "row" keyword and just use a two-level list.  That seems a bit better to me, especially because without the extension it would render as a list.

I'd be more inclined to add something closer to rst list tables. But even for those, I think it goes against some of the ideals we've been aiming at.

If you wanted to write a filter for something like list tables, you could simply have it turn certain lists into tables.  The parser wouldn't need to be modified, because the things would initially be parsed as lists.  (Of course, you'd have to ensure that the syntax is that of a list, unlike your current syntax.)  You'd have to answer two questions:

- how to identify that a given list is really a table?
- how to pass information about alignments and widths?

There are a variety of ways you might do this, including using an enclosing div, or having the first list element lay out this information.

mb21:
I'm closing this in an attempt to unclutter our open pull requests...

PR #5641, opened 2019-07-12
Update diff-zip to support windows

agusmba:
Trying to use this script on windows (git-bash) gave several errors:

- *mktemp* complained about missing XXXs in the template (fixed by adding XXX)
- *find* didn't understand the -E switch (fixed by using the equivalent arguments)
- *terminal codes* were not being recognized by the console (-e added to echo)


mb21:
nice! I can confirm this still works on macOS.

PR #5647, opened 2019-07-18
Change order of ilvl and numId in document.xml

### LINKED ISSUE #5645: Docx outputs in "compatibility mode" format so formating lost in Office365
Hello and thank you for the great tool.

I created a simple markdown file with bulleted lists called [conferences.md](https://www.dropbox.com/s/d8ntuq2zl55aqqf/conferences.md?dl=0). This converts nicely to [conferences.pdf](https://www.dropbox.com/s/sfzcdho005tfnjq/conferences.pdf?dl=0) using the command `pandoc conferences.md -o conferences.pdf`

I convert to docx using `pandoc conferences.md -o conferences.docx` which yields [conferences.docx](https://www.dropbox.com/s/ebcov62o8fc1a2v/conferences.docx?dl=0). When opened in Office 365, it loses the formating for bullet points and instead each bulleted line prints out non-indented with no bullet points. However when opened in Microsoft Word Professional Plus 16, conferences.docx is formated correctly (with bullet points and indendations saved). 

I noticed that in the header it states it is in "Compatibility mode." I then "convert" it to upgrade to the most recent docx format and save it as [conferencesUpgraded.docx](https://www.dropbox.com/s/su450gkanx5lpnc/conferencesUpgraded.docx?dl=0). This then opens with preserved bullets and indentations in Office 365.

I additionally tried making a custom `--reference-doc` by printing the default one that comes with pandoc, opening in Office 365, creating a few bullet points and changing then font, then saving then using that as the style reference for pandoc with no success (other than changing the font which was different in the two documents).

What I presume is happening based off this [Microsoft support link](https://docs.microsoft.com/en-us/deployoffice/compat/manage-compatibility-mode-for-office) is that pandoc is outputing in an older variant of docx that Office 365 is not recogning but Office 2016 is recognizing. Do you think this is the source of the bug, and if so, is that a tough fix to make? If not and you need more data, please let me know, happy to provide.

As far as my environement, I am running Gentoo linux. The pandoc version in portage is old (1.19.2.1) but I also installed the latest one (2.7.3) which is running as demonstrated below.
```
thomas@athena $ pandoc -v
pandoc 2.7.3
Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.8.1
Default user data directory: /home/thomas/.local/share/pandoc or /home/thomas/.pandoc
Copyright (C) 2006-2019 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
```

agusmba:
Workaround for Word Online shortcomming. Fixes #5645


agusmba:
Since the fix looked easy enough, I haven't tested it locally yet

agusmba:
Ok, Docx tests related to lists fail:

```
    Docx
      inlines
        font formatting:                                             OK (0.04s)
        hyperlinks:                                                  OK (0.19s)
        inline image:                                                OK (0.05s)
        inline images:                                               OK (0.05s)
        handling unicode input:                                      OK (0.04s)
        inline code:                                                 OK (0.04s)
        inline code in subscript and superscript:                    OK (0.04s)
      blocks
        headers:                                                     OK (0.05s)
        nested anchor spans in header:                               OK (0.05s)
        lists:                                                       FAIL (0.05s)
          Non-matching xml in word/document.xml:
          +  16           <w:ilvl w:val="0" />
          -  17           <w:ilvl w:val="0" />
          +  27           <w:ilvl w:val="0" />
          -  28           <w:ilvl w:val="0" />
          +  38           <w:ilvl w:val="1" />
          -  39           <w:ilvl w:val="1" />
```

I suppose these errors are normal, and I should modify the reference docs for these tests, shouldn't I?

mb21:
> I should modify the reference docs for these tests

Exactly. But please try to open them in desktop word and online word to make sure they still work.

agusmba:
Ok, in the tests, there are 4 docx affected by this:

* lists.docx
* lists_continuing.docx
* lists_restarting.docx
* table_with_list_cell.docx

The first three work fine in local and Online Word. However the last one still fails in Word Online.

I need to investigate more...

agusmba:
See #5645 for a blow by blow account of my findings.

agusmba:
All should be set, please review https://github.com/jgm/pandoc/issues/5645#issuecomment-513138913 to make sure older Word versions don't have problems opening the new docs with lists. There shouldn't be any problems, but it never harms to check.

mb21:
Nice work!

jgm:
Excellent!

PR #5653, opened 2019-07-19
MANUAL: reveal.js flags

mb21:
Unless anybody has a better idea?

The thing is, the revealjs template contains things like:

```
$if(controls)$
        // Display controls in the bottom right corner
        controls: $controls$,
$endif$
```

Since `0` is truthy in pandoc templates but falsy in JavaScript, this allows us to override [revealjs' default](https://github.com/hakimel/reveal.js#configuration).


jgm:
Clever!

PR #5656, opened 2019-07-23
HTML writer: render inline formulas correctly

### LINKED ISSUE #5655: --webtex uses display style regardless of delimiters
The web service used by Pandoc to convert TeX formulas to images uses display style by default. I don't know if the service has a parameter for specifying the style. A workaround is to prefix inline formulas with `\textstyle` to force it to render using inline style.

**Version:**
```sh
$ pandoc --version
pandoc 2.7.4
```

**Reproduce:**
```sh
$ pandoc --webtex
$\frac ab$
^D
```

**Expected:** URL to image of inline style formula.

**Actual:** URL to image of display style formula.

pesca:
Closes #5655.

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -1062,8 +1062,11 @@ inlineToHtml opts inline = do
       case writerHTMLMathMethod opts of
            WebTeX url -> do
               let imtag = if html5 then H5.img else H.img
+              let s = case t of
+                           InlineMath  -> "\\textstyle "
+                           DisplayMath -> ""
               let m = imtag ! A.style "vertical-align:middle"
-                            ! A.src (toValue $ url ++ urlEncode str)
+                            ! A.src (toValue $ url ++ urlEncode (s ++ str))
                             ! A.alt (toValue str)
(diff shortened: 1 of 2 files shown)


jgm:
Good idea!

pesca:
@jgm I feel the assumptions about the web service needs to be documented in the manpage. Otherwise a service which renders using inline style by default will not correctly render display formulas. Another alternative is to make no assumptions and prefix display formulas with `\displaystyle`. I'd be glad to make whichever change you feel is better.

src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -1062,8 +1062,11 @@ inlineToHtml opts inline = do
       case writerHTMLMathMethod opts of
            WebTeX url -> do
               let imtag = if html5 then H5.img else H.img
+              let s = case t of
+                           InlineMath  -> "\\textstyle "
+                           DisplayMath -> ""
```
pesca: Prefix display formulas in case an alternative service uses inline style by default?

jgm: This probably makes sense.

PR #5670, opened 2019-08-02
Docx writer: preserve heading and blockquote styles

bensteinberg:
This PR allows the user to change the behavior of custom styles when the output format is DOCX. The default behavior is for a custom style to propagate from a containing div down into its child elements, including headings and blockquotes -- the custom style trumps the
"native" styling of the child elements, so that they are not styled as headings or blockquotes.

In our application, we need those elements to retain their built-in styles, so I've added the command-line option `--docx-preserve-style` to prevent the custom style from overriding the built-in style in these two cases.

I was unable to find GitHub issues or posts in pandoc-discuss that addressed this exactly.


mb21:
Sounds reasonable I guess. Why only headings and blockquotes though?

I guess you could also define a new style in your reference.docx for those that simply looks the same as the default, or that wouldn't work?

bensteinberg:
Thanks!

The reason for only handling headings and blockquotes is that all other elements already behave like you'd expect; tables are tables, codeblocks are codeblocks, and so on.

Defining a new style in reference.docx would fail for nested elements: for instance, headings contained in blockquotes, or headings contained in divs with custom styles.

rebeccacremona:
I would really love to see this merged in. @mb21 it sounds like your concerns were allayed... Is there anything else in the way?

jgm:
I'm always reluctant to multiply command-line arguments beyond necessity.
So, my question is: why put this under a command-line flag?  Why not just make it the default?
It's hard for me to imagine a case where you'd want a container style to replace the Header or BlockQuote style.

MANUAL.txt
```diff
@@ -5223,8 +5223,10 @@ styles, pandoc allows you to define custom styles for blocks and text
 using `div`s and `span`s, respectively.
 
 If you define a `div` or `span` with the attribute `custom-style`,
-pandoc will apply your specified style to the contained elements. So,
-for example using the `bracketed_spans` syntax,
+pandoc will apply your specified style to the contained elements (with
+the exception of elements whose function depends on a style, like
+headers, code blocks, block quotes, or links). So, for example, using
```
jgm: We're now using "headings" instead of "headers" in the docs.

jgm:
I find the name `withoutTextProp` a bit confusing:  `withoutTextProp X` actually applies property X; it just omits other properties.

And this makes me think:  what if, instead of adding `withoutTextProp`, we simply modified `withTextProp` as follows:  if the text property being added is a style, first filter out any existing styles, so that there's always just one style in the list?

A more Haskellish way to do this would be to change the type for the list of text environments so that it has one slot for a style and a list for other things (which can be combined).

Something like this could be done for withParaProp too, and then we'd have a nice general solution that didn't require figuring out piecemeal which elements need special treatment.

MANUAL.txt
bensteinberg: Thanks, will fix this.

bensteinberg:
Thanks, @jgm. That's extremely helpful. I'll take a crack at implementing it this way.

bensteinberg:
@jgm, let me know if this is not close enough to what you had in mind, or if there are any infelicities I can smooth out.

src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -1119,6 +1150,9 @@ formattedRun els = do
 setFirstPara :: PandocMonad m => WS m ()
 setFirstPara =  modify $ \s -> s { stFirstPara = True }
 
+setInTable :: PandocMonad m => WS m ()
+setInTable =  modify $ \s -> s { stInTable = True }
+
```
jgm: A bit awkward to have `setInTable` but then have to use `modify` directly to reset.
Either have `set` and `reset`, or just use `modify`, I'd say.

src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -1063,16 +1087,22 @@ withNumId numid = local $ \env -> env{ envListNumId = numid }
...
 getTextProps :: (PandocMonad m) => WS m [Element]
 getTextProps = do
   props <- asks envTextProperties
-  return $ if null props
+  let squashed = squashProps props
+  return $ if null squashed
               then []
-              else [mknode "w:rPr" [] props]
+              else [mknode "w:rPr" [] $ squashed]
```
jgm: redundant `$`?

jgm:
One question is whether it's okay to squash all these changes into a single commit, or whether there are distinct atomic changes you'd rather preserve.  If so, you should do the rebasing yourself.

src/Text/Pandoc/Writers/Docx.hs
bensteinberg: Thanks. I'll switch to `modify` -- I was following the example of `setFirstPara`, which in one case has `modify $ \s -> s { stFirstPara = False }` afterwards.

jgm: Ah, I see. I guess i'd have the same reaction to that!  (It may be that I'm responsible for it.)

jgm:
Excellent!  Thank you for this excellent contribution.

Issue #4725, opened 2018-06-22
\include within LaTeX environment scrambles conversion


uliska:
Pandoc 2.0.6, installed from the `.deb` package

I am using Pandoc for a Markdown->LaTeX->PDF chain and use custom LaTeX in the Markdown file. I have a case where the *content* of a LaTeX environment scrambles the generated LaTeX code:

```tex
\begin{lilypond}[nofragment]
\include "oll-core/package.ily"
\loadModule stylesheets.span
{
  cis'4 \span blurred { d'8 d' } eis'4
}
\end{lilypond}
```
(making use of functionality from the [lyluatex](https://github.com/jperon/lyluatex) package, the environment content is passed to LilyPond, an external program).

This will be rendered to the following `.tex`:

```tex
\textbackslash{}begin\{lilypond\}{[}nofragment{]}
\include ``oll-core/package.ily'' \loadModule stylesheets.span \{ cis'4
\span blurred \{ d'8 d' \} eis'4 \} \textbackslash{}end\{lilypond\}
```
where nearly everything is escaped, which is of course wrong and makes the .tex code fail.

---

It can be made to work by one out of:

* commenting the `\include` line
* insert at least two tokens *before* the `\include`

```tex
\begin{lilypond}[nofragment]
\noop something
\include "oll-core/package.ily"
\loadModule stylesheets.span
{
  cis'4 \span blurred { d'8 d' } eis'4
}
\end{lilypond}
```
However, it doesn't work with a simple `\noop` (regardless of the fact that this is a fake command anyway)

---

Shouldn't Pandoc, when it encounters a `\begin{something}` pipe its content through regardless of what it contains?

mb21:
Can you try again with the latest pandoc release?

> the environment content is passed to LilyPond, an external program

btw. this sounds like a usecase for a pandoc filter http://pandoc.org/filters.html

uliska:
> Can you try again with the latest pandoc release?

This didn't change anything.

I have found a workaround by inserting `noop = {}` which is a consequence-less assignment in LilyPond. But it still is a hack.

>> the environment content is passed to LilyPond, an external program

> btw. this sounds like a usecase for a pandoc filter http://pandoc.org/filters.html

I know, and there even *is* a filter for LilyPond that makes use of `lyluatex` (https://github.com/jgm/pandocfilters/blob/master/examples/lilypond.py). But that uses a totally outdated version of the package, which we pushed from something like 0.1 to 1.0b since January.

[...]

PR #5671, opened 2019-08-04
Treat `ly` as verbatim too

uliska:
According to https://github.com/jgm/pandoc/issues/4725#issuecomment-399772217 not only the `lilypond` environment but also `ly` should be included in the verbatim list.

@jperon
https://github.com/jperon/lyluatex/issues/203


jgm:
thanks!

uliska:
> thanks!

You could consider it an act of self-interest ;-) So there's reason for me as well to thank you for merging without discussion ...

PR #5679, opened 2019-08-08
Remove blank line

leifmetcalf:
This blank line was breaking the code block into two. Visible at http://hackage.haskell.org/package/pandoc-2.7.3/docs/Text-Pandoc.html


jgm:
thanks!

PR #5688, opened 2019-08-14
Fix lua filter traversal order

tarleb:
I failed at keeping Circle CI happy, and I'd rather not break master. It works locally, so I am putting this up as a PR to be merged at a later time. Maybe I'm using `cabal.project` incorrectly?


jgm:
Hm, what you're doing with cabal.project looks right. Usually Exit 1 means it has run out of time -- the build has gone over CircleCI's time limit.
Try rerunning the workflow.

tarleb:
Figured it out: `cabal.project` is missing from the list of `extra-source-files` and is thus exluded from the tar archive built by `cabal new-sdist`. I can do a proper fix later today.

Issue #5691, opened 2019-08-16
Ensure GHC options are passed as intended


tarleb:
GHC options are listed in various places:

- `pandoc.cabal`
- `cabal.project`
- `stack.yaml`
- `.circleci/config.yml` (`-Werror`, stack build only)

Some options, which were not supported by older GHC versions, had to be removed from `cabal.project` in order to allow the project to build while respecting that file. The removed options should be reintroduced.

See also 6166f9499.

PR #5693, opened 2019-08-18
Cabal cleanup

tarleb:
This consolidates handling of GHC options in the build. The ghc-options specified in `pandoc.cabal` are no longer repeated in `cabal.project`. Some options ­– those changing only console output – are moved to `cabal.project.local`, which can be generated via `make cabal.project.local`. The make target respects the `GHC_VERSION` variable, so only valid options are added.

Closes: #5691


cabal.project
```diff
@@ -2,11 +2,11 @@ packages: pandoc.cabal
 
 package pandoc
   flags: +embed_data_files -trypandoc
-  ghc-options: -Wall -fno-warn-unused-do-bind -Wincomplete-record-updates -Wnoncanonical-monad-instances -Wnoncanonical-monadfail-instances -Wincomplete-uni-patterns -Widentities -j +RTS -A32M -RTS
+  ghc-options: -fhide-source-paths -j +RTS -A32M -RTS
 
 package pandoc-citeproc
   flags: +embed_data_files +bibutils -unicode_collation -test_citeproc -debug
-  ghc-options: -Wall -fno-warn-unused-do-bind -Wincomplete-record-updates -Wnoncanonical-monad-instances -Wnoncanonical-monadfail-instances -Wincomplete-uni-patterns -Widentities -j +RTS -A32M -RTS
+  ghc-options: -fhide-source-paths -j +RTS -A32M -RTS
```
jgm: Are you sure that options here get added to the ghc-options in the cabal file, rather than replacing them altogether?
I think I believed the latter, perhaps as a result of experience.
But this should be easy to test.

pandoc.cabal
```diff
@@ -182,6 +182,8 @@ extra-source-files:
                  INSTALL.md, AUTHORS.md, README.md,
                  CONTRIBUTING.md, BUGS, changelog,
                  man/pandoc.1
+                 -- Makefile
+                 Makefile
                  -- cabal and stack build plans
```
jgm: The argument against adding this is that the Makefile isn't necessary to build pandoc.
It's really just a convenience for developers, like some of the things in `tools/`.

cabal.project
```diff
@@ -2,11 +2,11 @@ packages: pandoc.cabal
 
 package pandoc
   flags: +embed_data_files -trypandoc
-  ghc-options: -fhide-source-paths -j +RTS -A32M -RTS
+  ghc-options: -j +RTS -A32M -RTS
 
 package pandoc-citeproc
   flags: +embed_data_files +bibutils -unicode_collation -test_citeproc -debug
-  ghc-options: -fhide-source-paths -j +RTS -A32M -RTS
+  ghc-options: -j +RTS -A32M -RTS
```
jgm: Why not just add `-fhide-source-paths` in pandoc.cabal, conditionally on ghc version?
That would eliminate the need for an extra makefile target and a cabal.project.local.
And (I think) people could still override this default by adding `-fno-hide-source-paths` in their cabal.project.local (though I haven't tested this).

cabal.project
tarleb: Options were combined when I tested it (cabal-install 2.4.0.0).

cabal.project
tarleb: I was afraid that this would be too opinionated and wouldn't quite belong in a cabal file. I'll happily change it.

I realized only yesterday that `cabal new-configure` overwrites cabal.project.local, which makes my approach even worse.

jgm: I understand your thinking here -- but I think being a bit opinionated is okay if it lets us avoid some complexity.

cabal.project
jgm: OK, great.

jgm: And if we put `-fhide-source-paths` in pandoc-citeproc.cabal, we're just left with `-j +RTS -A32M -RTS`.  I wonder if those should also go in pandoc-citeproc.cabal?

Issue #1209, opened 2014-03-28
Spell check not run on output .docx files 


pjking07:
## Problem

When converting a `.md` file to `.docx`, Word does not automatically run spell check on the document.
## Environment

`Mac OSX 10.9.2`
`pandoc 1.11.1`
`Word for Mac 2011`
## Reproduction

```
echo "I don't need no spell chek" > check.md && pandoc -S check.md -o check.docx && open check.docx
```

In the created document, the obvious spelling and grammar mistakes are not identified.
## Workaround

One can force Word to recheck the spelling by (on a Mac) going to `Preferences` -> `Spelling and Grammar` and clicking the button labeled `Recheck Document`.

According the the option, this button:

> Checks the spelling and grammar again after you change spelling and grammar options or open a custom or special dictionary. When you click this button, Word resets the internal Ignore All list so that Word will check all words for which you previously clicked Ignore All.
## Question

Does Pandoc already deal with this `Ignore All` flag somehow? Is it even possible to fix this from pandoc's end or is this merely an undesirable side effect of importing documents into Word?

krystofbe:
Sorry, to respond five years later. But even after following the instructions, I'm still not getting Word to perform a spell check. Although my reference docx is contains `dirty` for spell check and grammar check the created docx file still has 
```
  <w:proofState w:spelling="clean" w:grammar="clean" />
```

Any ideas?

jgm:
We've changed things a bit since that last comment, so that not everything in settings.xml gets copied over (to avoid corruption).
I think we could safely add w:proofState to the list of things copied, though.
Why don't you add a new issue requesting this, so we don't lose track of it.
This relates to `settingsList` in Text.Pandoc.Writers.Docx.hs.

krystofbe:
Hi @jgm! Thanks for the quick reply. 

Although I know nothing about Haskell, is this the change you would be doing?
https://github.com/jgm/pandoc/pull/5703

detoxhby:
Using v2.13 it is still not possible to set dirty values in reference doc. The element is copied over but the property values pick up default openxml enum value (which is clean). The process of forcing word to properly re-analyize the document by default is cumbersome and verbose, which is almost unmanagable after the 20th open a day.

Any workaround is appreciated!

Thank you.

detoxhby:
> OK, I'll reopen this.... It looks like we're getting _two_ proofState elements in settings.xml.

[...]

PR #5703, opened 2019-08-23
add proofState to settingsList

krystofbe:
Relates to https://github.com/jgm/pandoc/issues/1209#issuecomment-523961452


jgm:
Thanks! Let me know if this works in your testing.

PR #5713, opened 2019-08-27
[LaTeX] Enable user-override for hyphenations.

ascherer:
User can specify `hyphenation: every-where` in YAML section.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -390,6 +390,9 @@ $if(logo)$
 \logo{\includegraphics{$logo$}}
 $endif$
 $endif$
+$if(hyphenation)$
+\hyphenation{$hyphenation$}
+$endif§
 
 \begin{document}
 $if(has-frontmatter)$


mb21:
I don't know LaTeX that well, but did you check that setting the `lang: en` or similar not already has the desired effect?

ascherer:
`\hyphenation{}` is [quite standard](https://tex.stackexchange.com/questions/182569/how-to-manually-set-where-a-word-is-split), hence the example `every-where`.

Although it is used in the late stages of document creation, in rare cases it is necessary to give TeX a few extra hints where to /not hyphenate words.

Adding keyword `hyphenation` in the YAML preamble seems quite natural, compared to `header-includes` or other means to add stuff in the LaTeX preamble.

adunning:
This is an interesting idea, although it can lead to some unexpected results, since this overrides all languages. My feeling is that this is too specialized to justify inclusion in the main template, especially as many would want to use `\babelhyphenation` instead. Possibly we should consider adding `\input{ushyphex}` when the main language is US English.

ascherer:
Good point. I wasn't aware of `\babelhyphenation`.

Issue #5615, opened 2019-06-23
Extension to treat first heading level as title?


DanielSWolf:
I'm using Pandoc 2.7.3 to convert Markdown to AsciiDoc. In the output file, the level of every heading is 1 deeper than expected.

*README.md:*

```markdown
# Title

## Level 2
```

*Invocation:*

```
pandoc -o README.adoc README.md
```

*Output (README.adoc):*

```adoc
== Title

=== Level 2
```

*Expected output:*

```adoc
= Title

== Level 2
```

jgm:
The `=` is reserved for the document title, provided in metadata.
This is standard in asciidoc; see e.g. https://raw.githubusercontent.com/asciidoctor/asciidoctor.org/master/docs/asciidoc-syntax-quick-reference.adoc as an example.

jgm:
Use this form:
```
---
title: Title
...

# First section
 
```

DanielSWolf:
I'm not sure that I got my problem across. I realize that in AsciiDoc, `=` is the document title. The thing is that Markdown has no comparable syntax for defining the document title. As far as I know, the syntax you mention in your previous comment isn't [standard Markdown](https://daringfireball.net/projects/markdown/syntax). So in practice, `#` is often used in Markdown documents for the document title, then `##`, `###`, etc. are used for headings -- just like in AsciiDoc.

I have a large number of Markdown files that use this approach and that I'd like to convert to AsciiDoc. So I need to convert `#` to `=`, `##` to `==`, and so on. Is there a way to achieve this?

mb21:
Well, pandoc uses pandoc flavoured markdown...

I was thinking you can could write a lua-filter to decrement all header levels by one, similar to this example: https://pandoc.org/lua-filters.html#modifying-pandocs-manual.txt-for-man-pages, but this causes the current pandoc version to crash on `Prelude.init: empty list` somewhere in the asciidoc writer.

You could change [this line](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/AsciiDoc.hs#L177) and compile pandoc yourself...

mb21:
Specifically, the `hierarchicalizeWithIds` function in `Pandoc.Shared` breaks...

mb21:
Or write a filter that decrements all heading levels by one, except for the heading 1, which should be set as the document meta data title.

tarleb:
[...]

PR #5716, opened 2019-08-29
New markdown to asciidoc tests

mb21:
I've written a script (not included) and transformed the tests from https://github.com/bodiam/markdown-to-asciidoc/tree/master/src/test/resources/nl/jworks/markdown_to_asciidoc to our test format.

There are also another two files in that repo we could look at named `testsuite.[adoc|md]`: https://github.com/bodiam/markdown-to-asciidoc/tree/master/src/test/resources

To execute only the new tests:

    stack test  --test-arguments='-p markdown-to-asciidoc'

After fixing quite a few tests manually (insignificant changes like word-wrapping), we are at:

    23 out of 61 tests failed (2.45s)

Those are somewhat interesting cases where I'm not sure whether pandoc is wrong or the test is wrong. 

Also a few places where it should be possible to improve pandoc to output nicer asciidoc. I'll see whether I get to that as well...

### DIFF
--- a/test/command/markdown-to-asciidoc-headings.md
+++ b/test/command/markdown-to-asciidoc-headings.md
@@ -1,0 +1,94 @@
+converted from: <https://github.com/bodiam/markdown-to-asciidoc/tree/9bd90bc405a7d25b03822cc91154bcb315ab39bf/src/test/resources/nl/jworks/markdown_to_asciidoc/headings.feature>
+
+Render a level 1 heading
+
+```
+% pandoc -t asciidoc
+# Title #
+^D
+= Title
+```
+
+
+Render a level 1 underscored heading
+
+```
+% pandoc -t asciidoc
+Title
+=====
+^D
+= Title
+```
+
+
+Render a level 2 heading
+
+```
+% pandoc -t asciidoc
+## Title
+^D
+== Title
+```
+
+
+Render a level 2 underscored heading
+
+```
+% pandoc -t asciidoc
+Title
+-----
+^D
+== Title
+```
+
+
+Render a level 3 heading
+
+```
+% pandoc -t asciidoc
+## Title
+^D
+== Title
+```
+
+
+Render a level 4 heading
+
+```
+% pandoc -t asciidoc
+#### Title
+^D
+==== Title
+```
+
+
+Render a level 5 heading
+
+```
+% pandoc -t asciidoc
+##### Title
+^D
+===== Title
+```
+
+
+Render a level 6 heading
+
+```
+% pandoc -t asciidoc
+###### Title
+^D
+====== Title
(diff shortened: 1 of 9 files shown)


mb21:
Hm.. [asciidoc lists are interesting](https://www.methods.co.nz/asciidoc/chunked/ch17.html#_numbered_lists)... btw. what's the difference between `OrderedList (1,Decimal,Period)` and `[OrderedList (1,DefaultStyle,DefaultDelim)` in the AST?

test/command/markdown-to-asciidoc-headings.md
```diff
@@ -0,0 +1,94 @@
...
+^D
+== Title
+```
+
+
+Render a level 3 heading
+
+```
+% pandoc -t asciidoc
+## Title
```
cagix: shouldn't this read `### Title`?

test/command/markdown-to-asciidoc-headings.md
```diff
@@ -0,0 +1,94 @@
...
+^D
+== Title
```
cagix: ... and accordingly "`=== Title`" here?

jgm:
> what's the difference between OrderedList (1,Decimal,Period) and OrderedList (1,DefaultStyle,DefaultDelim) in the AST?

Decimal says "I definitely want a decimal list number." DefaultStyle says "use whatever is the default for this list level in this output format" (e.g. in LaTeX, just give me an unadorned `\begin{enumerate}`).  Similarly for Period vs DefaultDelim.

test/command/markdown-to-asciidoc-headings.md
mb21: Maybe :) We should probably implement https://github.com/jgm/pandoc/issues/5615

mb21:
About the heading levels... I think I let myself be confused by #5615 (which seems more about how we should handle heading levels in the markdown input).

From [asciidoctor docs](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#section-titles):

>     = Document Title (Level 0)
>
>     == Level 1 Section Title
>
>     === Level 2 Section Title
>
> When using the article doctype (the default), you can only have one level-0 section title (i.e., the document title) and it must be in the document header.

@cagix I guess you were right and I assume that's how it's always worked in asciidoc (not just asciidoctor) and [these test](https://github.com/bodiam/markdown-to-asciidoc/tree/master/src/test/resources/nl/jworks/markdown_to_asciidoc) by @bodiam are just not correct? The correct equivalences would be:

- Markdown: `# foo`
- HTML: `<h1>foo</h1>`
- AsciiDoc: `== foo`

tremain74:
I would like to contribute to this code

PR #5717, opened 2019-08-30
Org reader: modify handling of example blocks.

### LINKED ISSUE #4186: org-mode: respect -i switch on src/example blocks
(Using pandoc v2.0.5 on Arch Linux).

In org-mode, passing an `-i` switch to an example or src block tells it to respect leading whitespace on render:

```
#+BEGIN_EXAMPLE -i
    This should retain the four leading spaces
#+END_EXAMPLE

#+BEGIN_EXAMPLE
    This should not retain the four leading spaces
#+END_EXAMPLE
```

pandoc currently does not respect the switch, at least not when converting to HTML; it always strips leading whitespace.  This is a problem if you're citing examples where whitespace is significant (e.g. Python code).

leungbk:
Closes #4186.

I'm pretty sure Emacs' built-in Org-to-Markdown conversion is wrong; it doesn't respect the `-i` switch at all, for example. For this reason, I chose to use HTML instead for the test. It looks to me that Org's native HTML conversion (`org-html-export-to-html`) is consistent with what this PR suggests. I also tested by exporting to a PDF (`org-latex-export-to-pdf`) and eyeballing.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/Blocks.hs
+++ b/src/Text/Pandoc/Readers/Org/Blocks.hs
@@ -821,11 +821,22 @@ listItem parseIndentedMarker = try . withContext ListItemState $ do
 
 -- continuation of a list item - indented and separated by blankline or endline.
 -- Note: nested lists are parsed as continuations.
-listContinuation :: Monad m => Int
-                 -> OrgParser m String
+listContinuation :: PandocMonad m => Int -> OrgParser m String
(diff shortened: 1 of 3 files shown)


tarleb:
I just skimmed this, looks nice! It might take me a while to review in detail, not sure when I'll find the time. Thanks for being patient.

jgm:
Thanks for this!

PR #5728, opened 2019-09-04
Add to other-extensions field for use by cabal solver

vmchale:
As of `cabal-install` 3.0, the solver can use the `other-extensions` field to set flags - with this commit, pandoc should cross-compile out of the box :)


jgm:
Does this cabal file also work with cabal-install < 3.0, or will it cause an error?

jgm:
I guess CI answered my question.

Issue #5074, opened 2018-11-14
'-f docx+style' drops first space in name of style


K4zuki:
### Environment
- `pandoc 2.3.1` on Ubuntu16.04 on WSL

### How to reproduce

- create a docx file from general markdown(`input.md`). 
    - The file contains paragraphs in `custom-style="Source Code"` and `custom-style="Bullet List 1"` divs. 
    - **Important thing is the style name contains spaces**

```markdown
::: {custom-style="Source Code"}
Source Code
:::

::: {custom-style="Bullet List 1"}
Bullet List 1
:::
```
`pandoc -t docx -o docx.docx input.md`

- backward creation from the docx file to markdown with `-f docx+styles` option

`pandoc -f docx+styles -t markdown -o output.md  docx.docx`

### Problem
- the first space is dropped in style in `output.md`
```markdown
::: {custom-style="SourceCode"}
Source Code
:::

::: {custom-style="BulletList 1"}
Bullet List 1
:::
```

`Source Code` vs. `SourceCode`/ `Bullet List 1` vs. `BulletList 1` styles are treated as different thing in Word. There would be problem when running `pandoc -t docx -o docx2.docx output.md`

jgm:
This is what we get in docx.docx, style.xml:
```xml
<w:style w:type="paragraph" w:customStyle="1"                               
  w:styleId="BulletList1"><w:name w:val="Bullet List 1" /><w:basedOn          
  w:val="BodyText" /><w:qFormat /></w:style><w:style w:type="paragraph"       
  w:customStyle="1" w:styleId="SourceCode"><w:name w:val="Source Code"        
  /><w:basedOn w:val="BodyText" /><w:qFormat /></w:style><w:style             
  w:type="paragraph" w:customStyle="1" w:styleId="SourceCode"><w:name         
  w:val="Source Code" /><w:basedOn w:val="Normal" />
```
As you can see, the `w:name` has the spaces, and the `w:styleId` does not.
I'm not sure how custom styles work - maybe @jkr can chime in here.

jkr:
We filter out the spaces in `newParaPropToXML` to make a valid styleId:

```haskell

newParaPropToOpenXml :: String -> Element
newParaPropToOpenXml s =
  let styleId = filter (not . isSpace) s
...
```

I *believe* this was implemented because word complains if `styleId`s  have spaces. To the best of my undersanding, and recollection, the `document.xml` only makes reference to styleId so to reference custom styles we need to deal with these.

[...]

Issue #5523, opened 2019-05-24
'-f docx+style' misinterprets metadata styles (Title, Author, Date...) if docx file is modified in Word


agusmba:
Using pandoc v2.7.2

If we create a simple docx file:

```bash
$ pandoc -f markdown -t docx -o test.docx  << EOF
---
title: AMB Título
author: AMB
date: 24/05/19
...

# My first level title

Blah blah blah
EOF
```

We can convert back to markdown quite well:

```bash
$ pandoc -f docx+styles -t markdown -s test.docx
---
author: AMB
date: '24/05/19'
title: AMB Título
---

My first level title
====================

::: {custom-style="FirstParagraph"}
Blah blah blah
:::
```

However if I open the docx, modify some text and save, pandoc doesn't understand the metadata styles anymore:

```
$ pandoc -f docx+styles -t markdown -s testM.docx
::: {custom-style="Ttulo"}
AMB Título
:::

::: {custom-style="Author"}
AMB
:::

::: {custom-style="Fecha"}
24/05/19
:::

My first level title modified
=============================

::: {custom-style="FirstParagraph"}
Modified
:::
```

Here are the docx files:

[test.docx](https://github.com/jgm/pandoc/files/3216556/test.docx)

[testM.docx](https://github.com/jgm/pandoc/files/3216558/testM.docx)

agusmba:
Some initial investigation:

The original docx generated by pandoc uses english style names:

```xml
		<w:p>
			<w:pPr><w:pStyle w:val="Title" /></w:pPr>
			<w:r><w:t xml:space="preserve">AMB</w:t></w:r>
			<w:r><w:t xml:space="preserve"> </w:t></w:r>
			<w:r><w:t xml:space="preserve">Título</w:t></w:r>
		</w:p>
		<w:p>
			<w:pPr><w:pStyle w:val="Author" /></w:pPr>
			<w:r><w:t xml:space="preserve">AMB</w:t></w:r>
		</w:p>
		<w:p>
			<w:pPr><w:pStyle w:val="Date" /></w:pPr>
			<w:r><w:t xml:space="preserve">24/05/19</w:t></w:r>
		</w:p>
		<w:p>
			<w:pPr><w:pStyle w:val="Heading1" /></w:pPr>
			<w:bookmarkStart w:id="20" w:name="my-first-level-title" />
			<w:r><w:t xml:space="preserve">My first level title</w:t></w:r>
			<w:bookmarkEnd w:id="20" />
[...]

Issue #5052, opened 2018-11-07
docx to html hyperlink bug


redstoneleo:
`C:\Users\i\Downloads\pandoc-2.3.1-windows-i386\pandoc.exe -s a.docx -o example35.html`
test file [a.docx](https://github.com/jgm/pandoc/files/2557807/a.docx)

The text with a hyperlink in docx file is always followed with a ` <span class="underline">` in the  converted HTML file , which is of no usage and instead induce problems .
BTW, the bug also occurred when convert docx to Markdown .

mb21:
Well yes...

    $ pandoc -t native /Users/maurobieg/Downloads/a.docx 
    [Para [Link ("",[],[]) [Span ("",["underline"],[]) [Str "pandoc"]] ("https://github.com/jgm/pandoc",""),Space,Str "is",Space,Str "good"]]

The question is how did the underline get into the word file? How did you create the link in the file?

jgm:
The text is underlined in the Word file, so in that sense pandoc's conversion is correct.
But it seems to be Word's default to add underlining when you insert a hyperlink, so a case could be made for ignoring this for links.  @jkr any thoughts?

agusmba:
Just my 2c.

I know it is the default to have links underlined in Word, but since Word can change styles (in fact my pandoc-word template has links without underlines), I would try not being too smart about this.

If the original is underlined we could assume the intention is to keep the underline (otherwise, shouldn't it be removed from the original?)

jkr:
This is something that, for the most part, we already do. We make an effort not to pick up the underlining if it's the result of a hyperlink style (as opposed to explicitly being underlined). Our general approach is specifically to blacklist the style named "Hyperlink" when we're figuring out style formatting. (See `blacklistedCharStyles` in Docx.hs). 

Unfortunately, the docx in the bug report has a custom hyperlink style (`a7`). Now in the styles.xml file, this is listed as having the  "Hyperlink" name, so it should be something we should be able to figure out through checking the style inheritance, in order to figure out the blacklisting a bit better. It *should* be a fairly straightforward fix. I'll take a look.

jkr:
Actually, this is sort of a judgment call. The problem is that @redstoneleo used a custom hyperlink style (`a7`) but it has a `<w:name val="Hyperlink"/>` in `word/styles.xml`. According to [this page](http://officeopenxml.com/WPstyleGenProps.php), the name element 

```
[...]

Issue #4321, opened 2018-01-29
Custom bullet symbols in reference docx are ignored


kugelblitz:
Hello,

I would like to know what is the recommended way to set my own bullet symbols in lists for docx output, or is it possible at all.

I tried a number of things and none worked. Pandoc puts the filled circles as bullets and nothing else. The most obvious that I tried (also not working) is:

- pandoc.exe bullet-test.md -o bullet-test.docx --reference-docx bullet-test-ref.docx

Open bullet-test.docx, place the cursor on the list, click "Define New Bullet" in the ribbon menu, chose e.g. "+" character, save as bullet-test-ref.docx with "keep compatibility" checkbox checked.

- pandoc.exe bullet-test.md -o bullet-test.docx --reference-docx bullet-test-ref.docx

Open bullet-test.docx, see circles and not pluses.

If it matters, I have Word 2010 Russian edition.
Pandoc version is 2.1.1.

Regards,

Dmitry

[bullet-test.docx](https://github.com/jgm/pandoc/files/1674471/bullet-test.docx)
[bullet-test-ref.docx](https://github.com/jgm/pandoc/files/1674472/bullet-test-ref.docx)
[bullet-test.md.txt](https://github.com/jgm/pandoc/files/1674474/bullet-test.md.txt)

jkr:
It's a difficult problem. In your `bullet-test-ref.docx` file, you have *numerous* different bullet styles defined. The one with the `+` is actually bullet style 11. So it's hard to know which ones you wanted to use.

The problem here is that bullets are not part of the document's styles, and that's what we bring over from the reference.docx.  

We could *possibly* try to do something where the bullets that you've used in your reference document get priority in the output file. But that would be tricky to get right (in figuring out priority, do we care more about order, or how often it's used?), and might provide unwanted consequences as often as not. But I'll poke around and see if I can come up with anything that makes sense.

kugelblitz:
Jesse,

Thank you for a quick answer.
The bullet-test-ref.docx was obtained from Pandoc itself, with just one manual additional operation of chaning the bullet symbol. (I misquoted the pandoc.exe call in my initial message, that has been corrected now.) So whatever number of styles are there in the bullet-test-ref, they should be Pandoc's styles, plus maybe one new that Word created for my "+" bullet? I am not sure if I it is possible to change an existing style directly, without creating a new one? I will be happy to do anything with the template to get my bullets right.

[...]

PR #5732, opened 2019-09-07
Docx reader styles

lierdakil:
This is a quick stab at resolving #5523, #5052, #5074 and probably others. This more or less follows the option (1) outlined in https://github.com/jgm/pandoc/issues/5523#issuecomment-528813371

Seems to work fine, and it's not as much work as I initially feared.

I might have missed some corner case or another, so it's probably better if someone with a good understanding of the reader code will review this.

One important change is related to #5074, which is reflected in the tests, is that Docx Reader now outputs style _display name_ in as the "custom style". This should be generally more consistent with the Writer code, but I did not verify this thoroughly (the question is, does Writer expect the style name consistently, or does it use styleId in some cases)

I might have made more changes to the Docx Reader architecture than strictly necessary, for instance, I changed `ParStyle` and `CharStyle` to be records instead of synonyms for 2-tuples. I find this makes the code a little cleaner, but YMMV. I have also removed strictly duplicate fields from the `ParagraphStyle`, and made it carry complete information about associated styles instead of only the `styleId`s. Let me know if you disagree with any of those, so that we can discuss this further.

Some things that still might need to be done:

- [x] Perhaps block quote detection code should be moved from parser to reader, and the code detecting those based on `styleId` can be removed.
- [x] ~~The same applies to heading level detection.~~ Parser needs this information in one place, no reason to duplicate the code. The code detecting header styles based on `styleId` has been removed in d490df4
- [x] Documentation might need updating, wrt. #5074 
-  ~~I'm not entirely clear on the use of `List Paragraph` style (mostly used in `Readers.Docx.Lists`). At the first glance, it seems like a bit of a hack. This code could perhaps use a bit of a revision.~~ Won't be doing that here. Changing this requires sizable changes to the code, so it's going to be a separate PR if I ever get to it anyway.
- [x] Styles missing the `w:name` property are currently handled very, ahem, ungracefully. Technically, one could argue that styles missing `w:name` aren't proper OOXML, but it's probably not a good idea to fail on those altogether. On the other hand, returning "Missing style name" as the style name is not a good idea either. The choice is between two options, either (a) styles missing `w:name` will be ignored, or (b) styles missing `w:name` will get assigned a name matching their `w:styleId`. Let me know what is preferable, I should be able to implement either rather quickly. UPD: Since there was no input, I've gone with (b) in https://github.com/jgm/pandoc/pull/5732/commits/892c5b61f0e963f5b00b46ab46c021b1cbb9ae0a as probably the more robust option. It can potentially lead to unintuitive behaviours on malformed docx files though.
- [x] Remove obsolete (?) `list-item` style from `divsToKeep`?
- [x] ~~Remove obsolete (?) `Block Quotation` style from `isBlockQuote`?~~ Going to leave this in. Not much overhead, and removing it can potentially break some workflows.
- [x] Filter `Compact` style from output?
- [x] ~~Fix `Definition`/`Definition Term` classes bleeding to output in compact definition lists? (auto-fixed if `Compact` is filtered from output at a certain point)~~ This is a much more involved thing that I thought initially, so that will go as a separate pull request

Hopefully this proves useful in some way.

#### Docx Reader
1. Reader now consistently uses docx style names instead of ids for recovering semantic information. Practically, this means that built-in Word styles should be recognized regardless of Word locale.
2. With `docx+styles` output, `custom-style` will contain the docx style display name instead of id. For Word built-in styles, this means consistent style names regardless of Word locale. For custom styles, this means the `custom-style` content will be identical to style's name displayed in Word interface.
3. `Verbatim Char`, `Source Code` and various block quote styles (notably, `Block Text`) now honour inheritance. That is, if you create a custom docx character/paragraph style inherited from respectively `Verbatim Char` or `Source Code` (defined in `reference.docx`), characters/paragraphs using that custom style will be respectively recognized as inline code or code block by pandoc.
4. Reader now honours `Compact` style (defined in `reference.docx`). Any paragraphs having the `Compact` paragraph style, or any custom style inherited from `Compact` will be interpreted as `Plain` instead of `Para` by pandoc. This is mostly relevant for tight lists. A side effect of this is that `Compact` style will _not_ show up in `docx+styles` output. Notice that styles _inherited_ from `Compact` _will_ show up in `docx+styles` output.
5. Custom header styles will gain a class matching the style display name (previously, internal docx style identifier). One caveat is that pandoc classes, unlike docx style display names, can't have spaces in them, so spaces are replaced with dashes (`-`).

#### Docx Writer
1. Writer now consistently uses docx style names instead of ids. In particular, this means that any custom styles used by the writer will be detected by their name as seen in Word, not the internal id. Styles that this change affects: paragraph styles: `Author`, `Abstract`, `Compact`, `Figure`, `Captioned Figure`, `Image Caption`, `First Paragraph`, `Source Code`, `Table Caption`, `Definition`, `Definition Term`; character styles: `Verbatim Char`, token styles (those with names ending in `Tok`)


jgm:
Great!  @jkr it would be nice to have your thoughts on this.

jkr:
Looks good to me -- this has always been the area that I was the most unsure of, and that was built by other folks the most in the past. Everything certainly scans.

@lierdakil : The `ListParagraph` style is indeed a hack, as the whole list-processing step is. It all comes from the weirdness of ooxml's non-nesting lists, where each item and sub-paragraph of an item is its own `<w:p>`. To keep things confined, I repeated the docx structure in the docx type, and parsed it in `T.P.R.Docx`, which requires handing stuff off to the list module. This was the way to overload the information channels we had available (since it was already being translated into Pandoc-native) to communicate that "this is a further paragraph within the same item."

Frankly, the whole list-processing module, which only processes this stuff as a second pass at the end of conversion,is an ugly hack. I've abandoned a couple of more elegant approaches because I couldn't iron out all the issues.

jkr:
See the `bodyToOutput` function, which runs `blocksToBullets` and `blocksToDefinitions` after all the other work is done.

lierdakil:
@jkr, what is the purpose of `"list-item"` in `divsToKeep`? As far as I can tell, `divsToKeep` is only ever compared against `w:styleId` (in `master` branch that is), and I don't think there is such a style? More than that, I don't think Word allows dash `-` in `styleId` in the first place.

lierdakil:
For context, this is what I'm referring to:
https://github.com/jgm/pandoc/blob/f0577fe3609e81f653570e3e8e164be1d3ddbcef/src/Text/Pandoc/Readers/Docx.hs#L138-L139

And here's where it's used:
https://github.com/jgm/pandoc/blob/f0577fe3609e81f653570e3e8e164be1d3ddbcef/src/Text/Pandoc/Readers/Docx.hs#L528-L535

lierdakil:
Also, while at it, not entirely sure on `Block Quotation` style name. I've managed to track the original `BlockQuotation` thingy back to https://github.com/jgm/pandoc/pull/1529 and subsequently [this mailing list thread](https://groups.google.com/forum/#!searchin/pandoc-discuss/before$3A2014$2F08$2F13$20docx|sort:date/pandoc-discuss/3d9e0gtg4Nk/_xDrNCREJSgJ). The original message was, regrettably, deleted since then, so I can't check what's in there exactly, but the following quote makes me think that it was originally a user-defined style (it's a quote of a quote, yes; emphasis mine):

> > \- 191–205, 568–70, 576–79: A block quotation is not picked up, but
 *that’s my fault for using a non-standard style name*. I only bring it
 up because it seems odd that the one block quotation that was picked
 up was the one that didn’t use ***my** ‘Block Quotation’ style.*

lierdakil:
@jgm, I've made a small change in 7da6a21 to honor `Compact` style, which signals that a given paragraph was originally a `Plain` in pandoc-generated Docx. But now we have a bit of a snag: with  `+styles`, this style is passed through; this, however, is not the case with other "semantic" styles: like headings, block quotes, definition lists, etc. Should we perhaps remove the `Compact` style from the output as well?

As a side note, I've found a curious bug: `Definition` and `DefinitionTerm` classes don't get stripped from compact definition lists. I'll try to figure out what happens and fix it here as well.

jgm:
>  But now we have a bit of a snag: with +styles, this style is passed through; this, however, is not the case with other "semantic" styles: like headings, block quotes, definition lists, etc. Should we perhaps remove the Compact style from the output as well?

@lierdakil I'm not sure I follow, can you explain a bit more?

lierdakil:
@jgm, well, the issue's easy to demonstrate. Run:
`echo "1. One" | pandoc -t docx -o- | pandoc -f docx+styles -t markdown`
You will get
```markdown
1.  ::: {custom-style="Compact"}
    One
    :::
```
As you can see, the list of one item is treated like a tight list, hence docx writer adds `Compact` style to list items. Docx writer then sees it and interprets it as a "custom style".

Also of note, if you run `echo "1. One" | pandoc -t docx -o- | pandoc -f docx+styles -t native` on pandoc 2.7 (or master), you'll see
```markdown
[OrderedList (1,Decimal,Period)
 [[Div ("",[],[("custom-style","Compact")])
   [Para [Str "One"]]]]]
```
Notice how the list item is in `Para` instead of `Plain`. This behaviour is actually changed by https://github.com/jgm/pandoc/pull/5732/commits/7da6a214b7ce642ea85796ef67336b9b24f02098 (currently the last commit on this PR). `Compact` is still passed through, so with https://github.com/jgm/pandoc/pull/5732/commits/7da6a214b7ce642ea85796ef67336b9b24f02098 the output would be
```markdown
[OrderedList (1,Decimal,Period)
 [[Div ("",[],[("custom-style","Compact")])
   [Plain [Str "One"]]]]]
```

The question is, should we also remove the `custom-style=Compact` div from the output in this case? I'm leaning to "yes", after all `Definition`, `Definition Term` and `Source Code` styles that also serve as semantic markers _are_ removed from `docx+styles` output. But I thought I might as well ask.

agusmba:
Thank you for all the work on this! It looks great!

As long as you're discussing also lists, might I ask if https://github.com/jgm/pandoc/issues/4321 is relevant here? or is it a completely different beast best left undisturbed?

jkr:
@lierdakil not sure if you already go past this point, but the `list-paragraph` class was to pass along the fact that this particular paragraph came out of a paragraph (`<w:p>`) with list info (which is translated in the parser into a `ListItem`, which is essentially a paragraph with the necessary list info pulled from the numbering xml file). In other words, this is again temporarily holding onto info, *in pandoc format*, for the list parser, since that comes at the end, and works on pandoc format. It doesn't make it out of this temporary usage into the output.

Why use lowercase and a `Div` while the paragraph uses camelcase and a `pPr`? Seems lost to time

jgm:
We shouldn't omit a div with custom-style="Compact".  Compact is used to distinguish between Para and Plain, if I recall.  The reader should presumably do some kind of Para -> Plain transformation and not render the Compact style as a custom style, right?

lierdakil:
@jkr, you've missed the point of the question. I wasn't asking what's the purpose of `list-item`. I was asking why is `list-item` in `divsToKeep`? That's a list of style names. Those values are compared against Word style names/ids. `list-item` doesn't seem like it's can ever appear in a Word document.

@jgm, I'm not sure if I failed to communicate on some level, but could you please carefully re-read my previous comment, including the last few paragraphs? Because it feels like you didn't really read it. Perhaps that's because I've used misleading phrasing? By "remove div" I meant "drop the div, but leave its contents", as in
```haskell
stripCompactDiv :: Block -> Block
stripCompactDiv (Div ("", [], [("custom-style", "Compact")]) content) = content
stripCompactDiv x = x
```
(not actual code, just something to help communicate my point)

@agusmba, #4321 would be related to the Writer. I'm primarily concerned with the Reader here.

jgm:
Sorry @lierdakil - somehow my fingers wrote something different from what I meant to say.
I meant to say "we *should* omit a div," not *shouldn't*.  I meant to be agreeing with you: we should use something like the `stripCompactDiv` as you suggest.

jkr:
@lierdakil because div (style) names get dropped at the `parStyleToTransform` step, which is in `bodyPartToBlocks`, while that needs to survive until `blocksToDefinitions/Bullets`, which comes later:

~~~haskell
bodyToOutput :: PandocMonad m => Body -> DocxContext m (Meta, [Block])
bodyToOutput (Body bps) = do
  let (metabps, blkbps) = sepBodyParts bps
  meta <- bodyPartsToMeta metabps
  blks <- smushBlocks <$> mapM bodyPartToBlocks blkbps
  blks' <- rewriteLinks $ blocksToDefinitions $ blocksToBullets $ toList blks
  blks'' <- removeOrphanAnchors blks'
  return (meta, blks'')
~~~

jkr:
So it does get dropped, but in the list module.

lierdakil:
@jkr. Okay. Let's try this one more time, I feel like I'm failing to communicate my point here. 

`divsToKeep`'s name is a bit misleading. _At no point in time an actual `Div` is compared against `divsToKeep`._ More accurately, this should perhaps be named `paragraphStylesToKeep`. When a paragraph style is in that list, a `Div ("", [styleName], [])` gets created at some step of output transformation (`parStyleToTransform` does this actually). That's currently the _only_ purpose `divsToKeep` serves. Still with me? Cool. Now the issue is, at no point there is a paragraph style named `"list-item"`. The `Div ("", ["list-item"], kvs)` is created in `bodyPartToBlocks`, sure. But that is not predicated on any style names in the source docx, and that `Div` will never be compared against `divsToKeep`, since `divsToKeep` is only ever compared against paragraph style names _that are present in the original docx_. Does that make sense? So, based on this, it would seem to me that `"list-item"` in `divsToKeep` is completely extraneous and should probably be removed to avoid further confusion.

So the question is, am I being an idiot and missing something really obvious, or is this `"list-item"` an artefact from some old version of the Reader code? FWIW, removing it from `divsToKeep` does not affect tests at all (that said, some docx reader tests seem a wee bit outdated themselves). I'm asking because the code is a wee bit too complicated to run it in my head entirely, so I can't be 100% sure I'm not missing some arcane piece of code that does something underhanded somewhere (and that is not covered by tests)

For reference, here's where `Div ("", ["list-item"], kvs)` is created:
https://github.com/jgm/pandoc/blob/f0577fe3609e81f653570e3e8e164be1d3ddbcef/src/Text/Pandoc/Readers/Docx.hs#L642-L659
`parStyleToTransform` is not (can not) be applied to that for obvious technical reason, types don't match: `parStyleToTransform` accepts `ParagraphStyle` as argument, while `bodyPartToBlocks` returns `Blocks`.
https://github.com/jgm/pandoc/blob/f0577fe3609e81f653570e3e8e164be1d3ddbcef/src/Text/Pandoc/Readers/Docx.hs#L528
https://github.com/jgm/pandoc/blob/f0577fe3609e81f653570e3e8e164be1d3ddbcef/src/Text/Pandoc/Readers/Docx.hs#L574

jgm:
@lierdakil is this ready to merge now?

lierdakil:
@jgm, I'm going to go over the code once more, then it'll be good to go I think.

lierdakil:
Okay, I'm done with this one for the time being. Feel free to review/merge/etc whenever. Another PR or a couple related to this will be inbound at some point after this is merged.

I've added the primary changes which happened in this PR to the opening post.

I might've got carried away slightly with refactoring, sorry about that. All non-trivial changes are useful in one way or another, but the diff ended up being rather noisy.

The build failure is not my fault, apparently something broke in `statistics-0.15.1.0`

jgm:
This looks good to me!  @jkr , @mb21 do you agree this is okay to merge?

One question is whether these separate commits should be squashed, or is there value in keeping them separate in the history?  It would be good to have the motivating issues listed in the commit messages, if they aren't already, and it seems to me that at least some squashing would be a good idea.

lierdakil:
@jgm I tried to clean up the commit history as far as possible. Hopefully that's good enough.

mb21:
@jgm Unfortunately, I don't know the docx reader that well...

conklech:
I've confirmed that this branch resolves #5074; in particular, I can now successfully round-trip docx files that contain source text in styles with spaces in the name, using the original docx as the reference-doc. (In reality I am doing intermediary processing; docx->docx roundtripping is not a crazy use-case.) Furthermore, pandoc no longer creates duplicate styles in the output document's styles.xml, which was likely caused by the inconsistent styleId/styleName treatment.

Thanks very much—I was just about to undertake this same project myself.

jgm:
I've merged these commits manually (some conflicts had to be resolved), so closing this PR!
Thank you, this seems like a big improvement.

jgm:
It would be helpful if someone could check the motivating issues, to make sure the issues are resolved, and if so close them.

Issue #5730, opened 2019-09-06
Pandoc cannot find installed latex packages


chrisothal:
I'm running into a weird situation where pandoc cannot load installed packages (`tcolorbox` and `lipsum` in my case). I've included a MWE.

mwe.tex
```
\documentclass{article}
\usepackage{tcolorbox}
\usepackage{lipsum}
\begin{document}
\begin{tcolorbox}
  \lipsum[1]
\end{tcolorbox}
\end{document}
```
When I run `$ pdflatex mwe.tex -o mwe.pdf` I get the desired output:
![image](https://user-images.githubusercontent.com/29053059/64459305-9052f000-d0c5-11e9-9856-4bc7cc655201.png)
When I run `$ pandoc mwe.tex -o mwe.pdf --verbose` I see this in the first few lines from my terminal output before pandoc tries to compile a PDF:
```
[INFO] Could not load include file 'tcolorbox.sty' at line 2 column 23
[INFO] Could not load include file 'lipsum.sty' at line 3 column 20
[INFO] Skipped '\lipsum[1]
  ' at line 7 column 1
[INFO] Skipped '\lipsum[1]
  ' at line 7 column 1
[INFO] Skipped '\begin{tcolorbox}' at line 5 column 18
[INFO] Skipped '\end{tcolorbox}' at line 7 column 16
```

jgm:
Why do you use pandoc to turn a tex into a pdf?  Just use pdflatex.

When you use pandoc with a .tex input and .pdf output, pandoc tries to parse all the latex into a native Pandoc representation of the document.  It even tries to find included styles (in this case it doesn't find them, because they're not in the working directory).  Then pandoc will convert this Pandoc representation to .tex and then run pdflatex.  This conversion is lossy, as the manual says!  If you want to turn .tex into .pdf, just use pdflatex directly.

Note that the sty files would be found if you moved them to the working directory or used `--resource-path` to specify the directory that contains them.  In many cases this is not helpful, though, because pandoc can't really understand the lower-level tex that many .sty files contain.

mb21:
Maybe we should add a warning when people try to do `-f latex -o pdf`. Seems to be a fairly common mistake.

jgm:
Agreed, that might be useful!

philippemiron:
I'm having the same problem on Macos trying to convert a TeX file to Microsoft Docx format. Note that I have a full Latex installation and the file gets compile properly to PDF using pdflatex.

[...]

PR #5736, opened 2019-09-08
Emit warning on `-f latex -o out.pdf`

mb21:
add UnusualConversion to LogMessage [API change]

see https://github.com/jgm/pandoc/issues/5730#issuecomment-529086055


jgm:
thanks!

Issue #5744, opened 2019-09-12
Setting attrs in Lua filters from a table


shreevatsa:
(Extracted from comments on #5743 into a separate bug at request of @tarleb.)

When writing a Lua filter for pandoc, many of the functions for creating Elements are documented to take an “attr” argument for “additional attributes”, e.g. [pandoc.Image](https://pandoc.org/lua-filters.html#Image).

The currently supported interface for setting this `attr` argument is not obvious: something like

```
    pandoc.Image({}, png, "title", {width=36.51})
```
does not work; instead it appears that one must write
```
   attrs = pandoc.Attr("", {}, {{"width", 36.51}})
   pandoc.Image({)}, png, "title", attrs)
```

(This is based on the answer to [this StackOverflow question](https://stackoverflow.com/questions/47290793/pandoc-lua-filters-how-to-specify-attributes-for-span-element/), which seems to have led to #4071 which is closed now so maybe there's a slightly better way, but definitely the first code snippet above doesn't work.)

It would be nice to be able to write the former and/or for the syntax to be documented on the Lua filters page.

PR #5745, opened 2019-09-12
Lua filters: accept HTML-like attribute tables as Attr

tarleb:
Attr values can now be given as normal Lua tables; this can be used as
an alternative to constructing Attr values with `pandoc.Attr`.
Identifiers are taken from the `id` field, classes must be given as
space separated words in the `class` field. All remaining fields are
included as misc attributes.

With this change, the following lines now create equal elements:

    pandoc.Span('test', {id = 'test', class = 'a b', check = 1})
    pandoc.Span('test', pandoc.Attr('test', {'a','b'}, {check = 1}))

Note that the conversion from Lua table to Attr value is performed
during unmarshaling, _not_ during object construction:

    pandoc.Span('test', {id = 'test'}).identifier == nil
    pandoc.Span('test', pandoc.Attr('test')).identifier == 'test'
    -- equality is tested after unmarshaling
    pandoc.Span('x', {id = 'y'}) == pandoc.Span('x', pandoc.Attr('y')

Closes: #5744

### DIFF
--- a/src/Text/Pandoc/Lua/Marshaling/AST.hs
+++ b/src/Text/Pandoc/Lua/Marshaling/AST.hs
@@ -21,11 +21,13 @@ module Text.Pandoc.Lua.Marshaling.AST
 
 import Prelude
 import Control.Applicative ((<|>))
+import Data.Maybe (fromMaybe)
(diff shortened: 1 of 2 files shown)


tarleb:
Closing in favor of #5750.

PR #5746, opened 2019-09-12
permit loading custom format file to speed up tex compilation

Konfekt:
See https://web.archive.org/web/20160712215709/http://www.howtotex.com:80/tips-tricks/faster-latex-part-iv-use-a-precompiled-preamble/

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -1,3 +1,5 @@
+%&preamble
+% the first line above loads a custom format file preamble.fmt if existent
 % Options for packages loaded elsewhere
 \PassOptionsToPackage{unicode$for(hyperrefoptions)$,$hyperrefoptions$$endfor$}{hyperref}
 \PassOptionsToPackage{hyphens}{url}


Konfekt:
> I think this is probably too fancy for the default template. Power users may want to use their own templates to enable this.

A comment line is plain simple and does little harm. 
There are already fancy packages in the default template, such as 
```tex
% Allow footnotes in longtable head/foot
\IfFileExists{footnotehyper.sty}{\usepackage{footnotehyper}}{\usepackage{footnote}}
\makesavenoteenv{longtable}
```
which throw mysterious warnings such as
```
Package footnotehyper Warning: 
  The footnote environment will not be fully functional, sorry.
  You may try to email the author with the following information:.
\long macro:->\cref@constructprefix {footnote}{\cref@result }\protected@edef \c
ref@currentlabel {[footnote][\arabic {footnote}][\cref@result ]\p@footnote \@th
efnmark }\cref@old@makefntext 
```

> Note that you can pass in -fmt whatever using --pdf-engine-opt, and use a precompiled format without any changes to the default template.

Yes, but this quits with an error message if `whatever` does not exist, whereas the comment line lets it compile as usual. Less work and fewer errors.

> On the other hand @Konfekt I'm having a hard time visualizing the workflow for this. If it works I would think it should be enabled by default, which I think means passing the engine some arguments too no? Given Pandoc's habit of running LaTeX builds with everything tidied out of the way and intermediate artifacts destroyed at the end, what exactly would one have to do to take advantage of this?

One compiles, once the preamble is stable, a format file by
```sh
	pdftex -ini -interaction=nonstopmode -jobname=preamble "&pdflatex" mylatexformat.ltx main.tex
```
For example, in my makefile, there is an entry
```
preamble: $(NAME).tex
	pdftex -ini -interaction=nonstopmode -jobname=preamble "&pdflatex" mylatexformat.ltx $(NAME).tex
```
where the variable `NAME` is set to the name of the main file (without its file extension).

Thus, with this pull request, usually a single `make preamble` suffices.  For something fully automatic, [here](https://tex.stackexchange.com/questions/49388/custom-format-file-how-to-automate-the-generation-of-the-precompiled-preamble/50416#50416) is a `makefile` that automatically recompiles the format file in case the preamble changes.

jgm:
> > Note that you can pass in -fmt whatever using --pdf-engine-opt, and use a precompiled format > without any changes to the default template.
>
> Yes, but this quits with an error message if whatever does not exist, whereas the comment line lets 

Suppose you've generated a preamble at some point in the past (or someone else has in the working directory), but you've forgotten about it.  Now you make some changes to your markdown file, which lead to new packages being included.   What happens?  (This is not a rhetorical question: I really don't understand how this feature works.)  Does the new stuff get ignored because the precompiled preamble is being used?  If so, this might be a source of confusion and difficult to diagnose problems.

The explicit method on the command line makes the user specify when they want a preamble to be used, and I see some advantages to that...

Konfekt:
> Suppose you've generated a preamble at some point in the past (or someone else has in the working directory), but you've forgotten about it. Now you make some changes to your markdown file, which lead to new packages being included. What happens? (This is not a rhetorical question: I really don't understand how this feature works.) Does the new stuff get ignored because the precompiled preamble is being used?

Yes, the new stuff is ignored, and the preamble has to be recompiled manually, if the makefile such as proposed [here](https://tex.stackexchange.com/questions/49388/custom-format-file-how-to-automate-the-generation-of-the-precompiled-preamble/50416#50416) does not automatically take account of it.

A precompiled preamble is something to be aware of, but the user consciously compiled it and should know about its static nature. If years later something is off, such as an error being thrown about a missing package that just has been explicitly added, she might remember about his practice of precompiling preambles.

Perhaps other could chime in; this could be discussed in the mailing list, but perhaps few will comment over there. I am not sure how big the benefit in compile time is; but as the `default.latex` preamble is considerably large, especially in the `beamer` document class, it might be worth it.

jgm:
>  but the user consciously compiled it and should know about its static nature

Or maybe some other user working on the same document did.  Or maybe it was a few weeks ago.  It's just one more thing that can cause subtle and hard to diagnose problems.  That's why the explicit approach, where you're forced to say "I want to use this precompiled preamble", seems better to me.

lierdakil:
I'm with John on this. For one, only users who could take advantage of the feature are users who know about this already and would take explicit actions to make it work. For two, it's very easy to configure your pandoc installation locally to get the desired effect (just drop the modified tex template into `~/.pandoc` or pass `-fmt` option in the Makefile as John suggested [above](https://github.com/jgm/pandoc/pull/5746#issuecomment-531099350)). And it can potentially lead to unintuitive behaviour for users who _don't_ know about the feature.

So including this by default doesn't make any sense to me whatsoever. Those who do know about this precompiled prelude thing can make it work in less than a minute with minimal effort. Those who don't, don't need this in the default template.

PR #5750, opened 2019-09-15
Lua filters: allow passing of HTML-like tables instead of Attr

tarleb:
Attr values can now be given as normal Lua tables; this can be used as a
convenient alternative to define Attr values, instead of constructing
values with `pandoc.Attr`. Identifiers are taken from the *id* field,
classes must be given as space separated words in the *class* field. All
remaining fields are included as misc attributes.

With this change, the following lines now create equal elements:

    pandoc.Span('test', {id = 'test', class = 'a b', check = 1})
    pandoc.Span('test', pandoc.Attr('test', {'a','b'}, {check = 1}))

This also works when using the *attr* setter:

    local span = pandoc.Span 'text'
    span.attr = {id = 'test', class = 'a b', check = 1}

Furthermore, the *attributes* field of AST elements can now be a plain
key-value table even when using the `attributes` accessor:

    local span = pandoc.Span 'test'
    span.attributes = {check = 1}   -- works as expected now

Closes: #5744

### DIFF
--- a/data/pandoc.lua
+++ b/data/pandoc.lua
@@ -50,11 +50,15 @@ local utils = M.utils
 -- @param indices list of indices, starting with the most deeply nested
 -- @return newly created function
 -- @local
-function make_indexing_function(template, indices)
+function make_indexing_function(template, ...)
+  local indices = {...}
   local loadstring = loadstring or load
   local bracketed = {}
   for i = 1, #indices do
-    bracketed[i] = string.format('[%d]', indices[#indices - i + 1])
+    local idx = indices[#indices - i + 1]
+    bracketed[i] = type(idx) == 'number'
+      and string.format('[%d]', idx)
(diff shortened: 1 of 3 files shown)


tarleb:
Alternative to #5745. Rough performance tests show a ~3% performance degradation when running filter `function Pandoc(d) return pandoc.Pandoc(d.blocks, d.meta) end` on MANUAL.txt. I am planning to offset this by separating user-facing and internal element constructors.

jgm:
Looks good!  Did you want to merge this now or wait for the other changes you mentioned?

tarleb:
Refactoring constructors did not yield any performance gains. In fact, I can no longer measure *any* performance difference when using the code in this PR. This is probably due to my low-effort benchmarking.

jgm:
Great!

PR #5765, opened 2019-09-21
odt: Add external option for native numbering

pyssling:
This adds an external options +native_numbering to the
ODT writer enabling enumeration of figures and tables in
ODT output.

### DIFF
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -132,6 +132,7 @@ data Extension =
     | Ext_multiline_tables    -- ^ Pandoc-style multiline tables
     | Ext_native_divs             -- ^ Use Div blocks for contents of <div> tags
     | Ext_native_spans            -- ^ Use Span inlines for contents of <span>
+    | Ext_native_numbering    -- ^ Use output format's native numbering for figures and tables
     | Ext_ntb                 -- ^ ConTeXt Natural Tables
     | Ext_old_dashes          -- ^ -- = em, - before number = en
     | Ext_pandoc_title_block  -- ^ Pandoc title block
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2361,6 +2361,16 @@ input formats
 output formats
 :  `docx`, `odt`, `opendocument`, `html`
 
+#### Extension: `native_numbering` ####
+
+Enables native numbering of figures and tables. Enumeration
+starts at 1.
+
+This extension can be enabled/disabled for the following formats:
+
+output formats
+:  `odt`, `opendocument`
+
 #### Extension: `styles` #### {#ext-styles}
 
 When converting from docx, read all docx styles as divs (for
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -397,9 +397,9 @@ blockToOpenDocument o bs
         captionDoc <- if null c
                       then return empty
                       else inlinesToOpenDocument o c >>=
-                             if True -- temporary: see #5474
-                                then unNumberedCaption "TableCaption"
-                                else numberedTableCaption
+                             if isEnabled Ext_native_numbering o
+                                then numberedTableCaption
+                                else unNumberedCaption "TableCaption"
         th <- if all null h
                  then return empty
                  else colHeadsToOpenDocument o (map fst paraHStyles) h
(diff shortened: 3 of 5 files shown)


pyssling:
I'd like to add unit tests too, but I've no clue about doing so for options. Any pointers much appreciated.

src/Text/Pandoc/Extensions.hs
```diff
@@ -368,6 +369,8 @@ getDefaultExtensions "beamer"          = extensionsFromList
 getDefaultExtensions "context"         = extensionsFromList
                                           [Ext_smart,
                                            Ext_auto_identifiers]
+getDefaultExtensions "odt"             = extensionsFromList
```
pyssling: Ok, this was wrong. Removing.

src/Text/Pandoc/Extensions.hs
```diff
@@ -132,6 +132,7 @@ data Extension =
     | Ext_multiline_tables    -- ^ Pandoc-style multiline tables
     | Ext_native_divs             -- ^ Use Div blocks for contents of <div> tags
     | Ext_native_spans            -- ^ Use Span inlines for contents of <span>
+    | Ext_native_numbering    -- ^ Use pandoc native numbering for figures and tables
```
lierdakil: The doc-comment looks a little misleading. It sounds like pandoc has some kind of "native numbering", which it doesn't (yet). I suggest rephrasing this as "Use output format's native numbering for figures and tables".

jgm: Probably also a good idea to indicate that it's only currently supported for odt output

lierdakil: @jgm, we don't do that for `Ext_styles` though, so I don't think it's necessary here. It could potentially be useful for docx too down the line, and it doesn't seem like a good idea to write documentation that someone might (and probably will) forget to update later which will then lead to confusion, chaos, dogs and cats living together, mass hysteria.

pyssling: Ok, updated. No mention of odt though. I guess the correct thing to do (to keep things consistent) would be to specify which options apply to which formats, but that's out of scope for this work at least.

jgm: At least it needs to be documented clearly in the manual that this extension only affects ODT.  One could make a case for documenting in the Haddocks as well -- otherwise someone using pandoc as a Haskell library might make incorrect assumptions.  Not sure how to balance this against the possibility of not updating the comment in the future.  I don't feel too strongly about it, as long as it's clear in the manual.

jgm: Thinking ahead:  If we had an enumerated type for formats (which is somewhere in the issue tracker), we could add a typeclass `ImplementedFor Format Extension`, and then it would be clear from the types themselves what extensions worked for what formats.

jgm:
Unit tests:  writer unit tests are in test/Tests/Writers/FORMATNAME.hs.
Currently there's nothing for OpenDocument.  You could add one.  Then it's a matter of looking at some of the other unit tests to see how to proceed.  You can run writeOpenDocument with any options you like, including extension settings. Let me know if you run into problems.

jgm:
An easy way to add a test would be to add a command test in `test/command/`. See the examples there.  You'd just test `pandoc -t opendocument+native_numbering`.

src/Text/Pandoc/Extensions.hs
lierdakil: @jgm, uhh... what would `class ImplementedFor Format Extension` accomplish exactly? I mean, both `Format` and `Extension` would be enum types, I guess, so the only instance could in fact be `instance ImplementedFor Format Extension`, which does not seem very useful... Unless we get into dark type magic and start lifting data constructors into type algebra, but in that case I think type families will be more appropriate than typeclasses.

As a side note, there is support for type-level strings in the typechecker, so one could conceivably get away without using a enum type for `Format`. That is going to be somewhat tricky since it's impossible to bring a runtime string to the type level, for obvious reasons, but I think it's possible.

pyssling:
I've added unit tests. Let me know if there's anything else needed.

lierdakil:
Looking at the tests, having `Table` style on table captions doesn't seem right.

I suppose the first `String` argument on line 426 should be `"TableCaption"` instead of `"Table"`?
https://github.com/jgm/pandoc/blob/022a7f59fca6459a4f728f76998405a2ec816135/src/Text/Pandoc/Writers/OpenDocument.hs#L421-L433

At least for unnumbered captions we're using `"TableCaption"`, so it probably should be the same here.
https://github.com/jgm/pandoc/blob/022a7f59fca6459a4f728f76998405a2ec816135/src/Text/Pandoc/Writers/OpenDocument.hs#L399-L402

Or am I missing something?

src/Text/Pandoc/Extensions.hs
jgm: yes, you're right, that wasn't well thought out.

pyssling:
No, nothing missed, this was a bug, or inconsistency at least. I've fixed it now. Hope it all looks good.

lierdakil:
Looks good to me. The only thing left to do is to document this in the manual, I think. Could you perhaps add a short description to `MANUAL.txt` under "Other extensions" section? Remember to indicate that the extension is currently only implemented for `odt` output format (take a look at ``Extension: `empty_paragraphs` `` for an example)

pyssling:
Ok, hopefully that's it then.

jgm:
Great, thanks!

PR #5766, opened 2019-09-21
[Docx Writer] Re-use Readers.Docx.Parse for StyleMap

lierdakil:
The primary motivation for this PR is reducing code duplication.

Basically, at this point, Reader and Writer do the same exact work when parsing styles (Writer has to parse `reference.docx` styles), but use two completely separate code paths to do that.

This PR completely rewrites `Readers.Docx.StyleMap` to be a thin wrapper around `Readers.Docx.Parse`. To facilitate that, `Readers.Docx.Parse` is separated into two modules, see below for explanation. I've also moved `Readers.Docx.StyleMap` to `Writers.Docx.StyleMap`, also see below.

Finally, I've added my copyrights to `Readers.Docx.Parse` and `Readers.Docx.Parse.Styles`. The rationale is mostly to unambiguously indicate the person to yell at, just in case.

Rationale behind individual commits:
- 53a1317ea39bbc24870853f828d5ac918f0c007e is done to keep `Readers.Docx.Parse` internals somewhat isolated. Writer needs access to different internals than Reader, so the alternative is exporting more internals than needed by either of them, which doesn't sound good to me.
- f9072ae39abac7fee2c2a999e11cbc292fefd626 shouldn't need any additional explanation, that's the primary goal of this PR.
- a339ab4e82bb9387c82d6bdf9c9500e2a065fd87 is there to avoid confusion in the future. `StyleMap` is only used in Writer code, while `Parse` is only used in Reader code. Having `StyleMap` under `Readers` was not a good decision in the first place, so I thought this is a good opportunity to correct that.


jgm:
@jkr does this change seem okay to you?

jkr:
Sure -- seems good to me, on all fronts.

Issue #5692, opened 2019-08-16
Update CI pipeline to use latest fpco/stack-build:lts


tarleb:
The docker image fpco/stack-build:lts was updated to a new version on Docker Hub. It seems that it includes changes which break the current CI pipeline. The docker image used during Circle CI builds has been fixed at fpco/stack-build:lts-13 for the time being. The build-pipeline should be updated.

tarleb:
The sources for the new release have not been published yet. We may have to manually probe the image to find out why lts-14 is not working for us.

jgm:
Can you clarify what motivated this?  I don't see any recent build failures for pandoc master in Circle CI.

tarleb:
All my builds were suddenly failing while working on the Lua fix:

```
#!/bin/bash -eo pipefail
apt-get install -y cabal-install-2.4

Reading package lists... Done


Building dependency tree       


Reading state information... Done

E: Unable to locate package cabal-install-2.4
E: Couldn't find any package by glob 'cabal-install-2.4'
E: Couldn't find any package by regex 'cabal-install-2.4'
Exited with code 100
```

Took me a while to realize that the docker image had changed: https://hub.docker.com/r/fpco/stack-build/tags The latest image had been uploaded just a few minutes prior.

jgm:
@tarleb is this still an issue?

tarleb:
Yes, still an issue:

``` sh
docker run --rm -it fpco/stack-build:latest sh -c 'apt-get update && apt-get install cabal-install-2.4'
```

failes, while this succeeds:

``` sh
docker run --rm -it fpco/stack-build:lts-13 sh -c 'apt-get update && apt-get install cabal-install-2.4'
```

tarleb:
Or quicker:

```
$ docker run --rm -it fpco/stack-build:lts-13 grep -r hvr '/etc/apt/'
/etc/apt/sources.list.d/hvr-ubuntu-ghc-xenial.list:deb http://ppa.launchpad.net/hvr/ghc/ubuntu xenial main
/etc/apt/sources.list.d/hvr-ubuntu-ghc-xenial.list:# deb-src http://ppa.launchpad.net/hvr/ghc/ubuntu xenial main
/etc/apt/sources.list.d/hvr-ubuntu-ghc-xenial.list.save:deb http://ppa.launchpad.net/hvr/ghc/ubuntu xenial main
[...]

PR #5771, opened 2019-09-24
Circle CI: use latest stack LTS

tarleb:
Fixes: #5692


jgm:
I guess this is not relevant any more, since we aren't building on CircleCI.

PR #5778, opened 2019-09-27
Test with updated regex-pcre-builtin

lierdakil:
:warning: :rotating_light: ***do not merge*** :rotating_light: :warning: 

Sorry for the noise, I want to test if Pandoc builds with updated regex-pcre-builtin on all platforms, and it's the easiest/fastest way I can think of.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,4 +1,5 @@
 packages: pandoc.cabal
+allow-newer: regex-base
 
 package pandoc
   flags: +embed_data_files -trypandoc
@@ -12,3 +13,8 @@ source-repository-package
     type: git
     location: https://github.com/jgm/pandoc-citeproc
     tag: 0.16.3
+
+source-repository-package
+    type: git
+    location: https://github.com/lierdakil/regex-pcre-builtin
+    tag: v0.95.0.0-beta


alerque:
@lierdakil When you create PR's that aren't to be merged (yet) Github allows you to create them as Drafts and the merge options will be disabled. Unfortunately you can't convert them to to drafts after creation yet, but for next time...

lierdakil:
Ah, that's a relatively new feature apparently. And also not very obvious one. Thanks for the heads-up.

jgm:
Are you sure this test tested what you think it did?
Currently the released version of skylighting (the part of pandoc that uses regex-pcre-builtin) is set up to require an older version.  So I imagine that this build still used the older version (unless your branch has an older version number)?

You could try forking skylighting.  Remove the constraint in skylighting-core/skylighting-core.cabal, and add your fork of regex-pcre-builtin to cabal.project.  There are travis tests with a bunch of ghc versions.

lierdakil:
@jgm, pretty sure this did what I think it did -- at least where the build system uses cabal new-build (which is I think Windows and most Linux builds, but not macOS one apparently). Package overrides in `cabal.project` essentially substitute the package in the dependency resolution tree, so that affects also all upstream dependencies. Notice I had to specify `allow-newer: regex-base` to work around the constraint in `skylighting-core`.

jgm:
Ah, right. `allow-newer` would do it, I guess.
In any case you should be able to tell from the build logs which version was used.

PR #5782, opened 2019-10-02
Test forked regex-pcre-builtin

lierdakil:
Sorry for more noise, github doesn't allow me to reopen my old PR for some reason (probably a careless rebase broke something)

Anyway, this also adds lierdakil/regex-pcre-builtin to stack builds.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -12,3 +12,8 @@ source-repository-package
     type: git
     location: https://github.com/jgm/pandoc-citeproc
     tag: 0.16.3
+
+source-repository-package
+    type: git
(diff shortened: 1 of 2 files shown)


lierdakil:
Side note: apparently GitHub Actions pulls latest GHC since build failures there point to incompatibility with `base-4.13.0.0` (which is with GHC 8.8) with haddock.

jgm:
I think that since I set it up, they must have changed things so that all cabal/ghc versions are pre-installed?  So `ghc` now gets the most recent.  I changed it to use an explicit version number in the path; hopefully that will fix.

PR #5785, opened 2019-10-03
Add additional listings languages.

Wandmalfarbe:
I added the additional languages `Eiffel`, `elisp`, `OORexx` and `Swift` (see the [listings documentation](http://mirrors.ctan.org/macros/latex/contrib/listings/listings.pdf)). I also added `cs` and `objectivec` as an "alias" for `C`. Listings has a dialect definition for Objective C but pandoc currently has no support for the listings dialects.


jgm:
thanks!

PR #5808, opened 2019-10-09
Updates MathJax URL to work with version 3

zorkow:
This updates the URL and filenames to work with MathJax version 3.
The only problem is the  `--standalone` option, where loading SRE and the initial fonts fail. That should change in the near future, once v3 can webpack SRE properly. 

On the other hand for v2.7.2 the option did also not work as expected, as loading of fonts,  Menu and Zoom fails. So I feel there is not that much change there.


jgm:
Standalone never worked with mathjax.  Glad to hear it might eventually.

jgm:
and thanks!

PR #5818, opened 2019-10-13
KaTeX: fleqn option

### LINKED ISSUE #5815: Feature request: allow left-aligned display math in HTML output (with Katex)
If Pandoc is invoked with `--metadata=classoption:fleqn` then it will left-align display equations in latex output. I would like to propose that this feature be extended to HTML output, in particular if the `--katex` option is used because Katex has an [option](https://katex.org/docs/options.html) for it.

I think that what it would take to implement this feature is changing [this string](https://github.com/jgm/pandoc/blob/3351dcfc45cf2053ae6cb6fb811d88cb6ebbc969/src/Text/Pandoc/Writers/HTML.hs#L285) depending on the document's metadata. (I'm not the best at Haskell though so I'm not sure how that would work :S)

This would be useful because it would allow for left-aligned display equations in HTML output. Currently I don't think there is any way to have Pandoc pass the `fleqn` option to Katex.

mb21:
closes #5815

Is there a better way to write this? Do we have an analogue of `lookupMetaBool` for `Context` somewhere?

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1988,6 +1988,12 @@ ODT or pptx.
 [Unicode Bidirectional Algorithm]: http://www.w3.org/International/articles/inline-bidi-markup/uba-basics
 [Language subtag lookup]: https://r12a.github.io/app-subtags/
 
+### Variables for HTML math
+
+`classoption`
+:   when using [KaTeX](#option--katex), you can render display math equations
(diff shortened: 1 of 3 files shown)


mb21:
Sure, managed to squeeze it in below 80chars..

jgm:
Great idea to use classoption.

PR #5822, opened 2019-10-15
[LaTeX Writer] Don't convert unicode quotes to TeX ligatures

seifferth:
Converting unicode ligatures to TeX ligatures may kick in unexpectedly,
such as when specifying pdf metadata. Consider the following YAML:

    title-meta: Lorem „ipsum“

Running pandoc and later `pdfinfo *.pdf|grep Title` will produce the
following output:

    Title:          Lorem „ipsum``

Rather than the expected:

    Title:          Lorem „ipsum“

Since LaTeX handles unicode quotes quite well by itself already (either
through `\usepackage[utf8]{inputenc}` or through native support in Xe-
and LuaLaTeX), there is no need to convert unicode quotes. I would thus
suggest to leave unicode quotes the way they were entered in markdown.

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -380,10 +380,6 @@ stringToLaTeX context zs = do
          '\x200B' -> emits "\\hspace{0pt}"  -- zero-width space
          '\x202F' -> emits "\\,"
          '\x2026' -> emitcseq "\\ldots"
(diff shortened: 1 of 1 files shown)


jgm:
You can achieve this by using `-t latex-smart`.  However, I think the default makes sense, since the default pdf-engine is still pdflatex, and we aim to produce old-fashioned standard LaTeX as much as possible.

seifferth:
Thanks for pointing me to `-smart`, I hadn't thought of that.

Issue #5796, opened 2019-10-06
Support <kbd> in HTML


jgm:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd
```
Please press <kbd>Ctrl</kbd>
```
would better be parsed as a Span with a special class `kbd`.
The writer should be sensitive to this as well, rendering such spans using `<kbd>`.

salim-b:
While switching to Pandoc 2.8.x I've noticed that

```sh
echo "This is my [key]{.kbd} example" | pandoc --from=markdown --to=markdown-raw_attribute-inline_code_attributes
```

outputs

```
This is my [key]{.kbd} example
```

while

```sh
echo "This is my <kbd>key</kbd> example" | pandoc --from=markdown --to=markdown-raw_attribute-inline_code_attributes
```
outputs

```
This is my <kbd>key</kbd> example
```

I think it would be awesome if the former command would produce the same output as the latter since

> [Markdown allows you to insert raw HTML (or DocBook) anywhere in a document (...)](https://pandoc.org/MANUAL.html#extension-raw_html)

Currently [inline code attributes](https://pandoc.org/MANUAL.html#extension-inline_code_attributes) seem like a one-way road to me...

jgm:
Yes, but raw HTML is just passed through as raw HTML; it isn't parsed.

salim-b:
> Yes, but raw HTML is just passed through as raw HTML; it isn't parsed. (That's by design.)

Well, since Pandoc 2.8, `<kbd>` isn't just "passed through" anymore as long as I don't explicitly disable the `raw_attribute` extension:

```sh
echo "This is my <kbd>key</kbd> example" | pandoc --from=markdown --to=markdown
```

yields:

```
This is my `<kbd>`{=html}key`</kbd>`{=html} example
```

jgm:
That is just a pandoc-markdown syntax for explicitly marking up raw text in some format.
If you don't like it, disable the `raw_attribute` extension:
```
 % echo "This is my <kbd>key</kbd> example" | pandoc --from=markdown --to=markdown-raw_attribute
This is my <kbd>key</kbd> example
```

salim-b:
> That is just a pandoc-markdown syntax for explicitly marking up raw text in some format.

OK, I got that.

But somehow I still feel that

```sh
echo "This is my [key]{.kbd} example" | pandoc --from=markdown --to=markdown-inline_code_attributes
```

should then yield

```
[...]

PR #5824, opened 2019-10-15
Add support for <mark>

cideM:
It is a bit hard to believe that this is all it takes and since this is my first PR on any meaningful project I'm still somewhat skeptical :thinking: but the tests pass and manual testing in the shell also seemed fine. Thanks to the groundwork from #5796 

I don't quite understand the meaning of the numbered `*.md` files for command and couldn't find any info on whether they're generated based on some source. Anyway, let me know if anything's wrong :pray: and thanks for tagging with good first issue and for `pandoc` in general <3

### DIFF
--- a/src/Text/Pandoc/Shared.hs
+++ b/src/Text/Pandoc/Shared.hs
@@ -698,7 +698,7 @@ underlineSpan = B.spanWith ("", ["underline"], [])
 -- | Set of HTML elements that are represented as Span with a class equal as
 -- the element tag itself.
 htmlSpanLikeElements :: Set.Set T.Text
-htmlSpanLikeElements = Set.fromList [T.pack "kbd"]
+htmlSpanLikeElements = Set.fromList [T.pack "kbd", T.pack "mark"]
 
 -- | Returns the first sentence in a list of inlines, and the rest.
 breakSentence :: [Inline] -> ([Inline], [Inline])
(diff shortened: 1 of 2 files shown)


jgm:
Yep, it's that simple!

I would only suggest renaming the command test.  `kbd.md` doesn't make sense; use the issue number that the commit closes, `5797.md`.

My other suggestion would be to revise the commit message so it gives a clearer idea of what the change does (including which issue it closes and how the behavior will change).  Say that it causes `<mark>` elements to be parsed from HTML as a Span with class `mark`, and renders such spans into HTML using the `<mark>` element.  This is important because when I construct the changelog for a new release, I need to describe all the important changes that have been made.

cideM:
Ha, I must've spent so much time looking at the kbd PR that I eventually confused the name. I mean to call it `mark.md`. Anyway thanks for the quick and thorough feedback :bowing_man: I changed the commit message and renamed the file.

One thing: maybe if I hadn't looked at it late in the evening I might have made the connection between file name and issue for the `test/command`s but it might not hurt to add one sentence about this to the `CONTRIBUTING.md`. Is this something you'd want and if so would you want me to do it in this PR in another commit?

mb21:
Indeed, it only says:

> Alternatively, you may add a "command test" to the /test/command/ hierarchy, following the pattern of the tests there.

a new pull request for that would be great.

cideM:
Added in #5829

PR #5837, opened 2019-10-19
HTML reader: Handle cite attribute for quotes

### LINKED ISSUE #5798: Handle cite attribute of <q> in HTML
If `cite` attribute is present, we should parse a `<q>` as a Link around a Quoted.
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/q
```
<q cite="https://www.imdb.com/title/tt0062622/quotes/qt0396921">I'm sorry, Dave. I'm afraid I can't do that.</q>
```

barskern:
If a `<q>` tag has a `cite` attribute, we interpret it as a link with an
inner quote.

Closes #5798

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -67,6 +67,11 @@
 
     + Add RTL support (Alexander Krotov, #5551).
 
+  * HTML reader:
+
+    + If `cite` attribute is present on a `<q>` tag, represent it as a
+      link around a quote.
+
   * JIRA writer:
 
     + Remove escapeStringForJira for code blocks (Jan-Otto Kröpke).
--- a/changelog.md
+++ b/changelog.md
@@ -67,11 +67,6 @@
 
     + Add RTL support (Alexander Krotov, #5551).
 
-  * HTML reader:
-
-    + If `cite` attribute is present on a `<q>` tag, represent it as a
-      link around a quote.
-
   * JIRA writer:
 
     + Remove escapeStringForJira for code blocks (Jan-Otto Kröpke).
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -703,13 +703,12 @@ pQ = try $ do
   case maybeFromAttrib "cite" tag of
        Nothing   -> quote
        Just url' -> do
-         let title = T.unpack $ fromAttrib "title" tag
          -- take id from id attribute if present, otherwise name
          let uid = fromMaybe (T.unpack $ fromAttrib "name" tag) $
                       maybeFromAttrib "id" tag
          let cls = words $ T.unpack $ fromAttrib "class" tag
          url <- canonicalizeUrl url'
-         extractSpaces (B.linkWith (uid, cls, []) (escapeURI url) title) <$> quote
+         extractSpaces (B.spanWith (uid, cls, [("cite", escapeURI url)])) <$> quote
 
 pEmph :: PandocMonad m => TagParser m Inlines
 pEmph = pInlinesInTags "em" B.emph <|> pInlinesInTags "i" B.emph
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -708,11 +708,7 @@ pQ = try $ do
          let uid = fromMaybe (T.unpack $ fromAttrib "name" tag) $
                       maybeFromAttrib "id" tag
          let cls = words $ T.unpack $ fromAttrib "class" tag
-         mbBaseHref <- baseHref <$> getState
-         let url = case (parseURIReference url', mbBaseHref) of
-                        (Just rel, Just bs) ->
-                          show (rel `nonStrictRelativeTo` bs)
-                        _                   -> url'
+         url <- canonicalizeUrl url'
          extractSpaces (B.linkWith (uid, cls, []) (escapeURI url) title) <$> quote
 
 pEmph :: PandocMonad m => TagParser m Inlines
@@ -774,21 +770,13 @@ pLink = try $ do
        Nothing   ->
          return $ extractSpaces (B.spanWith (uid, cls, [])) lab
        Just url' -> do
-         mbBaseHref <- baseHref <$> getState
-         let url = case (parseURIReference url', mbBaseHref) of
-                        (Just rel, Just bs) ->
-                          show (rel `nonStrictRelativeTo` bs)
-                        _                   -> url'
+         url <- canonicalizeUrl url'
          return $ extractSpaces (B.linkWith (uid, cls, []) (escapeURI url) title) lab
 
 pImage :: PandocMonad m => TagParser m Inlines
 pImage = do
   tag <- pSelfClosing (=="img") (isJust . lookup "src")
-  mbBaseHref <- baseHref <$> getState
-  let url' = T.unpack $ fromAttrib "src" tag
-  let url = case (parseURIReference url', mbBaseHref) of
-                 (Just rel, Just bs) -> show (rel `nonStrictRelativeTo` bs)
-                 _                   -> url'
+  url <- canonicalizeUrl $ T.unpack $ fromAttrib "src" tag
   let title = T.unpack $ fromAttrib "title" tag
   let alt = T.unpack $ fromAttrib "alt" tag
   let uid = T.unpack $ fromAttrib "id" tag
@@ -1300,6 +1288,17 @@ isSpace '\t' = True
 isSpace '\n' = True
 isSpace '\r' = True
 isSpace _    = False
+
+-- Utilities
+
+-- | Adjusts a url according to the document's base URL.
+canonicalizeUrl :: PandocMonad m => String -> TagParser m String
+canonicalizeUrl url = do
+  mbBaseHref <- baseHref <$> getState
+  return $ case (parseURIReference url, mbBaseHref) of
+                (Just rel, Just bs) -> show (rel `nonStrictRelativeTo` bs)
+                _                   -> url
+
 
 -- Instances
 
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -685,7 +685,6 @@ pSelfClosing f g = do
 pQ :: PandocMonad m => TagParser m Inlines
 pQ = try $ do
   tag <- pSatisfy $ tagOpenLit "q" (const True)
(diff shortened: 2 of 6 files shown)


mb21:
Nice! Thanks for your pull!

Two suggestions:

1. Maybe we should add a class `cite` or something to the link, so the info is not lost? @jgm?

2. This seems like a good time to extract something like the following as a helper function, and use it in `pLink`, `pImage` and `pQ`:

   ```haskell
   canonicalizeUrl :: PandocMonad m => String -> m String
   canonicalizeUrl url = do
     mbBaseHref <- baseHref <$> getState
     return $ case (parseURIReference url, mbBaseHref) of
                   (Just rel, Just bs) -> show (rel `nonStrictRelativeTo` bs)
                   _                   -> url
   ```

jgm:
I agree about the helper function.

I'm having second thoughts, a little, about the idea of making the whole quote a link.  It seems to me that this will often be quite ugly (imagine a several-lines quote).  An alternative would be creating a Span with `cite` attribute around the Quoted -- which wouldn't have any default behavior, but could be handled by filters and the HTML writer could be trained to render it back as `<q cite=`.

barskern:
> I'm having second thoughts, a little, about the idea of making the whole quote a link. It seems to me that this will often be quite ugly (imagine a several-lines quote). An alternative would be creating a Span with `cite` attribute around the Quoted -- which wouldn't have any default behavior, but could be handled by filters and the HTML writer could be trained to render it back as `<q cite=`.

I agree regarding this. And the specification does not render this as a link in the browser, hence doing that in an PDF document would perhaps be suprising.

barskern:
I've look at the error and I'm not really sure how to solve it. Do any of you have a suggestion?

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -1301,6 +1289,17 @@ isSpace '\n' = True
 isSpace '\r' = True
 isSpace _    = False
 
+-- Utilities
+
+-- | Adjusts a url according to the document's base URL.
+canonicalizeUrl :: PandocMonad m => String -> m String
+canonicalizeUrl url = do
+  mbBaseHref <- baseHref <$> getState
```
jgm: One problem here is that you're using `getState`, which is defined in Parsec monads, but the type sig says this works in any instance of `PandocMonad`.  Does that help?

jgm: You probably want
```
PandocMonad m => String -> TagParser m String
```

barskern: Thank you for the tip :smile:

changelog.md
```diff
@@ -67,6 +67,11 @@
 
     + Add RTL support (Alexander Krotov, #5551).
 
+  * HTML reader:
```
jgm: Leave off the changelog change for now.  I will gather these together from the git log eventually, but since I look at the log since the last modification of the changelog, modifications now will confuse the process...

jgm: This explanation should go in the git commit message for the actual change.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -694,8 +697,19 @@ pQ = do
...
+  case maybeFromAttrib "cite" tag of
+       Nothing   -> quote
+       Just url' -> do
+         let title = T.unpack $ fromAttrib "title" tag
+         -- take id from id attribute if present, otherwise name
+         let uid = fromMaybe (T.unpack $ fromAttrib "name" tag) $
+                      maybeFromAttrib "id" tag
+         let cls = words $ T.unpack $ fromAttrib "class" tag
+         url <- canonicalizeUrl url'
+         extractSpaces (B.linkWith (uid, cls, []) (escapeURI url) title) <$> quote
```
jgm: Let's go with the span with attribute `cite` rather than the link, as discussed.

changelog.md
barskern: Okay ill add it to the commit when I squash

barskern:
When writing HTML, should we turn any `Span` with a `cite` attribute into a `<q>` (reguardless if there is a `Quoted` inside) or should we ensure that there is both a `cite` and a `Quoted`?

jgm:
I'd say we should require this exact structure: Quoted inside Span with cite.

jgm:
Well actually I'm not so sure. Maybe using a 'q' element when it's just a Span with a cite attribute would also make sense.  What do you think?

barskern:
I think we should do the strict version (attribute and inner tag) because there might be some other use of the same attribute name which then would lead to some unexpected behavior. `<q>` is a quoted tag so it makes sense that the content are a quote.

mb21:
Unfortunately, the `Quoted` document AST element doesn't have `Attr`... but we could do:

    Quoted SingleQuote [Span (... cite="http://....") ...]

barskern:
@mb21 Currently the code makes a `Span` with the `cite` attribute and a quoted `Quoted` inside.

mb21:
@barskern I think that's fine as well.

barskern:
After experimenting a bit with a solution, I think it's preferable to rather have a `Span` inside of the `Quoted` so that the relevant code can be under the case when we have a `Quoted` inline.

```haskell
inlineToHtml :: PandocMonad m
             => WriterOptions -> Inline -> StateT WriterState m Html
    -- ...
    -- instead of checking if `ils` has a `Quoted` here
    (Span (id',classes,kvs) ils) ->
    -- ...
    -- we can check if `lst` has a `Span` with `cite` here
    (Quoted quoteType lst) ->
```

barskern:
Okay now I implemented it for both reader and writer. I made a choice that it will only make `<q>` tags if the writer option is enabled. This means that if that option is not enabled the span will simply be stripped away as if it wasn't there. This could of course be changed, but it seemed like the most obvious solution. However this does mean that one looses the information about the `cite` attribute though.

src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -1047,12 +1048,20 @@ inlineToHtml opts inline = do
...
+                        in if writerHtmlQTags opts
                                then do
                                  modify $ \st -> st{ stQuotes = True }
-                                 H.q `fmap` inlineListToHtml opts lst
+                                 H.q `fmap` inlineListToHtml opts lst' >>= withAttr
                                else (\x -> leftQuote >> x >> rightQuote)
-                                    `fmap` inlineListToHtml opts lst
+                                    `fmap` inlineListToHtml opts lst'
+                          where
+                            lst' = maybe lst (\cs -> cs ++ tail lst) (snd <$> citeSpan)
```
jgm: We should generally avoid using partial functions like `tail` -- consider using an alternative, perhaps one of those defined in Safe.

src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -1047,12 +1048,20 @@ inlineToHtml opts inline = do
...
+                            withAttr = maybe return (addAttrs opts) (fst <$> citeSpan)
+                            citeSpan = listToMaybe lst >>= (\case
```
jgm: I'm not so sure about the use of `listToMaybe` here.
I think we should just pattern match for the exact structure `Quoted _ [Span attr xs]` where attr includes cite.
If they have `Quoted _ [Span attr xs, Str "hello"]`, for example, it shouldn't have special handling.  That will allow you to simplify your code here.

Also: why not just use the orginal inline list when `writerHtmlQTags` is not set?  We'll get regular quotes around span tags, that's fine.

test/Tests/Writers/HTML.hs
```diff
@@ -30,6 +33,11 @@ infix 4 =:
      => String -> (a, String) -> TestTree
 (=:) = test html
 
+infix 4 =::
+(=::) :: (ToString a, ToPandoc a)
+     => String -> (a, String) -> TestTree
+(=::) = test htmlQTags
+
```
jgm: I think that for code clarity it would be better not to define a special-purpose operator for this; just use `test htmlQTags` to avoid confusion down the line...

test/Tests/Writers/HTML.hs
```diff
@@ -13,6 +13,9 @@ import Text.Pandoc.Builder
 html :: (ToPandoc a) => a -> String
 html = unpack . purely (writeHtml4String def{ writerWrapText = WrapNone }) . toPandoc
 
+htmlQTags :: (ToPandoc a) => a -> String
+htmlQTags = unpack . purely (writeHtml4String def{ writerWrapText = WrapNone, writerHtmlQTags = True }) . toPandoc
```
jgm: Watch for over-long lines -- our guidelines say to try to fit into 80 columns.

src/Text/Pandoc/Writers/HTML.hs
barskern: `tail` is safe to use here because the code will only run if `listToMaybe` returned `Just` which means there are at least one element in `lst`. Nonetheless I see your point. Would you like me to change it still?

src/Text/Pandoc/Writers/HTML.hs
barskern: Hmm yeah that would simplify the code. Thank you for the tip! I will change it.

test/Tests/Writers/HTML.hs
barskern: Yeah I did that originally but I got an error saying the type was to generic. Hence I needed to add a bound in some way to limit the possible types. I figured I would do the same as done before. I could try to see if I get it working another way, but I need some way to limit the types hence I think I will need a function declaration no matter what.

jgm: Probably you just need to add a type declaration somewhere to the argument of `test htmlQTags` (using `ScopedTypeVariables` extension).
Or use a function declaration, as you are, but I think it's cleaner if we reserve the operators for really generic stuff.

src/Text/Pandoc/Writers/HTML.hs
barskern: > yeah, we've had cases when a later refactoring change the thing around the partial function...

Yeah that is a good point. This will be changed when I rather pattern match directly (as mentioned below).

jgm: Yes, it's good practice just to avoid using these partial functions; then you know you won't get in trouble.

test/Tests/Writers/HTML.hs
barskern: Ended up using a simple function declaration because I didn't figure out how to solve it using `ScopedTypeVariables`.

jgm:
Excellent, thanks for your contribution!

PR #5855, opened 2019-10-26
Default latex template: patch for fixing spacing problem on CJK language.

jeongminkim-islab:
Issue and solution is from below.
https://tex.stackexchange.com/questions/17292/how-to-write-spaces-between-korean-words-with-xecjk

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -137,7 +137,7 @@ $endif$
 $endif$
 $if(CJKmainfont)$
   \ifxetex
-    \usepackage{xeCJK}
+    \usepackage[space]{xeCJK}
     \setCJKmainfont[$for(CJKoptions)$$CJKoptions$$sep$,$endfor$]{$CJKmainfont$}
   \fi
 $endif$


jgm:
Thanks. I'd like to get feedback from other CJK users about this proposed change in the default.
Could you post something on pandoc-discuss?
I'm curious why it's the default to ignore spaces.

jgm:
well, nobody has objected, so I'm merging it!  Thanks.

annProg:
This may be problem when documentclass already load xeCJK with no option.  
```
! LaTeX Error: Option clash for package xeCJK.

 ...

l.19     \setCJKmainfont
                        [BoldFont=Source Han Serif SC Heavy,ItalicFont=Fando...

? H
The package xeCJK has already been loaded with options:
  []
There has now been an attempt to load it with options
  [space]
Adding the global options:
  ,space
to your \documentclass declaration may fix this.
Try typing  <return>  to proceed.
```

May use `\xeCJKsetup { CJKspace = true }` to do this. Code from `ctex-engine-xetex.def`

```
\keys_define:nn { ctex }
  {
    space .choice: ,
    space / true  .code:n =
      { \xeCJKsetup { CJKspace = true } } ,
    space / auto  .code:n =
      { \xeCJKsetup { CJKspace = false } } ,
    space / false .code:n =
      { \xeCJKsetup { CJKspace = false } } ,
    space .default:n = { true } ,
    space .initial:n = { auto }
  }
```

jgm:
@annProg thanks for this; could you create a new issue so we don't lose track?
You can link back here.

Issue #5799, opened 2019-10-06
Handle <var> in HTML


jgm:
Should be parsed as Code with a class `var`.
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/var
This could be supported bidirectionally, in both reader and writer.

Mauricio-123:
Primeramente el var para que quieres usarlo

Wandmalfarbe:
The elements [`<samp>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/samp) and [`<tt>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt) could be treated similarly.

DavSanchez:
Hello!

I'd like to know more about this issue. For example, when talking about an HTML reader, do you just want `<var>` to have the same behaviour as `<code>`?

Maybe I can try and contribute to it. I'll read the code a little bit to see if I'm capable of doing it.

jgm:
No, not the same behavior.  In pandoc Code elements can have attributes; the suggestion is to add the `var` class to these attributes in parsing `var`.

jgm:
@Wandmalfarbe  on samp, see #5792.  tt is already parsed as generic Code, which seems right.

amogh09:
For Writers/HTML, do we want to check if `var` or `samp` is a member of `Code`'s attribute classes and if so write `<var>` or `<samp>`, respectively, instead of `<code>`?

jgm:
> For Writers/HTML, do we want to check if var or samp is a member of Code's attribute classes and if so write <var> or <samp>, respectively, instead of <code>?

Yes, I think so.

Question: should we use `var` and `samp` as the classes, or `variable` and `sample`?

amogh09:
I think `variable` and `sample` makes more sense. I will add this feature if @DavSanchez is not working (or planning to work) on it.

DavSanchez:
@amogh09 please feel free to do it! At the moment I’m focusing my free time on HIE, so I don’t think I’ll be able to work in this, at least for a time.

amogh09:
@jgm Added a PR for this feature. Please take a look.

PR #5861, opened 2019-10-28
HTML Reader/Writer - Add support for <var> and <samp>

amogh09:
Change for #5799. Adds Reader/Writer support for `<var>` HTML element and Writer support for `<samp>` HTML element.

Before - 
```
➜  pandoc git:(issue5799) ✗ pandoc -t markdown -f html
<var>result</var>
result
➜  pandoc git:(issue5799) ✗ pandoc -f markdown -t html
`result`{.sample}
<p><code class="sample">result</code></p>
➜  pandoc git:(issue5799) ✗ pandoc -f markdown -t html
`result`{.variable}
<p><code class="variable">result</code></p>
```

After - 
```
➜  pandoc git:(issue5799) ✗ pandoc -t markdown -f html
<var>result</var>
`result`{.variable}
➜  pandoc git:(issue5799) ✗ pandoc -f markdown -t html
`result`{.sample}
<p><samp>result</samp></p>
➜  pandoc git:(issue5799) ✗ pandoc -f markdown -t html
`result`{.variable}
<p><var>result</var></p>
```

### DIFF
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -788,7 +788,7 @@ pCodeWithClass elemToClass = try $ do
   TagOpen open attr' <- pSatisfy $ tagOpen tagTest (const True)
   result <- manyTill pAny (pCloses open)
   let (ids,cs,kvs) = mkAttr . toStringAttr $ attr'
-      cs'          = maybe cs id . fmap (:cs) . lookup open $ elemToClass
+      cs'          = maybe cs (:cs) . lookup open $ elemToClass
   return . B.codeWith (ids,cs',kvs) .
     unwords . lines . T.unpack . innerText $ result
 
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -38,7 +38,7 @@ import Data.Foldable (for_)
 import Data.List (isPrefixOf)
 import Data.List.Split (wordsBy, splitWhen)
 import qualified Data.Map as M
-import Data.Maybe (fromMaybe, isJust, isNothing, fromJust)
+import Data.Maybe (fromMaybe, isJust, isNothing)
 import Data.Monoid (First (..))
 import qualified Data.Set as Set
 import Data.Text (Text)
@@ -784,10 +784,12 @@ pImage = do
 
 pCodeWithClass :: PandocMonad m => [(T.Text,String)] -> TagParser m Inlines 
 pCodeWithClass elemToClass = try $ do 
-  TagOpen open attr' <- pSatisfy $ tagOpen (flip elem . fmap fst $ elemToClass) (const True)
+  let tagTest = flip elem . fmap fst $ elemToClass
+  TagOpen open attr' <- pSatisfy $ tagOpen tagTest (const True)
   result <- manyTill pAny (pCloses open)
   let (ids,cs,kvs) = mkAttr . toStringAttr $ attr'
-  return . B.codeWith (ids,fromJust (lookup open elemToClass) : cs,kvs) .
+      cs'          = maybe cs id . fmap (:cs) . lookup open $ elemToClass
+  return . B.codeWith (ids,cs',kvs) .
     unwords . lines . T.unpack . innerText $ result
 
 pCode :: PandocMonad m => TagParser m Inlines
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -38,7 +38,7 @@ import Data.Foldable (for_)
 import Data.List (isPrefixOf)
 import Data.List.Split (wordsBy, splitWhen)
 import qualified Data.Map as M
-import Data.Maybe (fromMaybe, isJust, isNothing)
+import Data.Maybe (fromMaybe, isJust, isNothing, fromJust)
 import Data.Monoid (First (..))
 import qualified Data.Set as Set
 import Data.Text (Text)
@@ -652,7 +652,7 @@ inline = choice
            , pLink
            , pImage
            , pCode
-           , pSamp
+           , pCodeWithClass [(T.pack "samp","sample"),(T.pack "var","variable")]
            , pSpan
            , pMath False
            , pScriptMath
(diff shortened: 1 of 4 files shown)


jgm:
Can you fix the CI build failure?

amogh09:
I think a rerun should fix it given that macos and windwos have passed. I cannot figure out how to retry the build.

jgm:
Looks like you just have a redundant import.

src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -1022,11 +1022,13 @@ inlineToHtml opts inline = do
...
+                     -> case hlCode of
                              Left msg -> do
                                unless (null msg) $
                                  report $ CouldNotHighlight msg
-                               addAttrs opts attr $ H.code $ strToHtml str
+                               addAttrs opts (ids,classes',kvs) $ 
+                                 elemType $ strToHtml str
                              Right h -> do
                                modify $ \st -> st{ stHighlighting = True }
                                addAttrs opts (id',[],keyvals) h
```
amogh09: I don't understand what needs to be done for this case.

amogh09:
Sorry, was a little busy. Removed unused import now.

src/Text/Pandoc/Writers/HTML.hs
jgm: Sorry, I'm not sure I understand your question - can you elaborate?

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -782,12 +782,12 @@ pImage = do
...
-pSamp :: PandocMonad m => TagParser m Inlines 
-pSamp = try $ do 
-  TagOpen open attr' <- pSatisfy $ tagOpen (=="samp") (const True)
+pCodeWithClass :: PandocMonad m => [(T.Text,String)] -> TagParser m Inlines 
+pCodeWithClass elemToClass = try $ do 
+  TagOpen open attr' <- pSatisfy $ tagOpen (flip elem . fmap fst $ elemToClass) (const True)
   result <- manyTill pAny (pCloses open)
   let (ids,cs,kvs) = mkAttr . toStringAttr $ attr'
-  return . B.codeWith (ids,"sample":cs,kvs) .
+  return . B.codeWith (ids,fromJust (lookup open elemToClass) : cs,kvs) .
```
jgm: Avoid partial functions like `fromJust`.  This will crash if the argument is Nothing.
Even if you're sure the argument won't be Nothing, it's good to code defensively.

amogh09: This `do` block is under a `try` so I think it's okay to throw exception on some failure? I have changed this to `maybe` nonetheless.

src/Text/Pandoc/Writers/HTML.hs
amogh09: When `hlCode` is a `Left` value, it was straightforward to choose from `H.code`, `H.var`, or `H.samp`. But when `isJust (writerHighlightStyle opts)` is `True` (i.e. `hlCode` is a `Right` value), the `Html` is generated from ``` highlight (writerSyntaxMap opts) formatHtmlInline attr str ``` is output. Do we need to support highlighted `<samp>` or `<var>`?

src/Text/Pandoc/Readers/HTML.hs
jgm: Parsec `try` isn't what you might expect from `try` in other languages.
If an exception is raised there, the `try` doesn't catch it.

src/Text/Pandoc/Writers/HTML.hs
jgm: Ah, I see.  The highlighted code returned by skylighting already includes a code tag, right?
Well, maybe we could just include the whole thing under `samp`.  As far as I can see, all phrasing content, including `<code>`, is legal inside a `<samp>` element. Same for `<var>`.

src/Text/Pandoc/Readers/HTML.hs
amogh09: Ah I thought it's `try :: Exception e => IO a -> IO (Either e a)`. Thank you.

src/Text/Pandoc/Writers/HTML.hs
amogh09: Added `<samp>` and `<var>` support for highlighted code.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -782,12 +782,14 @@ pImage = do
...
+pCodeWithClass :: PandocMonad m => [(T.Text,String)] -> TagParser m Inlines 
+pCodeWithClass elemToClass = try $ do 
+  let tagTest = flip elem . fmap fst $ elemToClass
+  TagOpen open attr' <- pSatisfy $ tagOpen tagTest (const True)
   result <- manyTill pAny (pCloses open)
   let (ids,cs,kvs) = mkAttr . toStringAttr $ attr'
-  return . B.codeWith (ids,"sample":cs,kvs) .
+      cs'          = maybe cs (:cs) . lookup open $ elemToClass
+  return . B.codeWith (ids,cs',kvs) .
     unwords . lines . T.unpack . innerText $ result
```
amogh09: I think this needs to be changed as it won't parse tags inside `<samp>` and `<var>`? Currently it does the following. 

```
➜  pandoc git:(issue5799) ✗pandoc -f html -t markdown
<var><code class=\"sourceCode haskell\"><span class=\"op\">&gt;&gt;=</span></code></var>
`>>=`{.variable}
```

jgm: That's on purpose.  The `Code` constructor takes a plain String.  So we have to strip out (ignore) all the tags inside it when we're reading.  There's no way to represent them.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -652,7 +652,7 @@ inline = choice
            , pLink
            , pImage
            , pCode
-           , pSamp
+           , pCodeWithClass [(T.pack "samp","sample"),(T.pack "var","variable")]
```
jgm: Because you have the OverloadedStrings pragma set in this module, you can get rid of the `T.pack`.

jgm:
Thanks for the PR!

src/Text/Pandoc/Readers/HTML.hs
amogh09: Thanks for the tip. Sorry, I was on vacation for two weeks so couldn't fix this.

PR #5864, opened 2019-10-30
docbook reader: fix nesting of chapters and sections

flokli:
Currently, both a `<title/>` in a chapter and in a `<section>` below that
chapter have the same level if they're not inside a `<book/>`.

This can be fixed by setting `dbBook` to true when traversing a chapter.

This can happen in a multi-file book project. Also see the example at
https://tdg.docbook.org/tdg/4.5/chapter.html

We add a new test for that case. The existing `docbook-xref` and `docbook-reader` tests already cover the "with `<book/>`" and "`<article/>`" case.


jgm:
Great, thanks!

PR #5875, opened 2019-11-05
Update default.context

denismaier:
Define an start-stop-pair `cslreferences` to allow for hanging indents in the bibliography. Analogous to the cslreferences-environment in the default latex template. See here: https://github.com/jgm/pandoc-citeproc/issues/410
For this to work the context writer must be adapted.

### DIFF
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -105,6 +105,24 @@ $endif$
 \setupxtable[body][]
 \setupxtable[foot][bottomframe=on]
 
+% Wrap CSL references in start-stop-environment
+$if(csl-refs)$
+\newdimen\cslhangindent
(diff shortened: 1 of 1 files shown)


denismaier:
Sorry, I don't understand what is going wrong here. Any hints?

denismaier:
Well, but that's only a comment and an additional newline, both only there for more readable code. Should I remove those?

jgm:
It makes sense to put them under the conditional.  Otherwise they appear in all context documents, even when pandoc-citeproc isn't used.  The comment will certainly seem odd without the stuff following it.  Alternatively, you can use the syntax `$-- ` to create a template comment which won't appear in the fgenerated source.

PR #5882, opened 2019-11-06
Change the implementation of `htmlSpanLikeElements` and implement `<dfn>`

### LINKED ISSUE #5795: Support <dfn> in HTML
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dfn

It would be good to parse
```
<dfn id="def-validator">validator</dfn>
```
as (pandoc markdown)
```
[validator]{#def-validator .definition}
```
and if we did that we could also render this construction to HTML using a `<dfn>` element.

cideM:
This PR fixes #5795 

As outlined in [my comment](https://github.com/jgm/pandoc/issues/5795#issuecomment-547640600) implementing `<dfn>` in the way that was suggested in the issue required making some changes to `htmlSpanLikeElements`. `<dfn>` can take [title and id attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dfn), so `htmlSpanLikeElements` could no longer omit attributes. 

I then went ahead and also changed the implementation from taking a single CSS class that represents the element to a pandoc-specific data attribute. I think this implementation is more robust and has less chances of interfering with user CSS classes. For example, on the current master, manually adding `<span class="kbd">foo</span>` will compile to `<kbd>foo</kbd>` when doing `-f html -t html`. Then again, I'm very new to contributing to `pandoc` so if there are things I'm overlooking here please let me know :bowing_man:

Let me know what you think, looking forward to feedback!

### DIFF
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -62,7 +62,7 @@ import Text.Pandoc.Options (
     extensionEnabled)
 import Text.Pandoc.Parsing hiding ((<|>))
 import Text.Pandoc.Shared (addMetaField, blocksToInlines', crFilter, escapeURI,
-                           extractSpaces, htmlSpanLikeDataAttrName, htmlSpanLikeElements,
+                           extractSpaces, htmlSpanLikeElements,
                            onlySimpleTableCells, safeRead, underlineSpan)
 import Text.Pandoc.Walk
 import Text.Parsec.Error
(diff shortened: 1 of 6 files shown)


jgm:
I don't like the `data-is-pandoc-spanlike` approach (though I do see the point).  I think using a simple class is more consistent with what we're already doing in a number of places elsewhere.

cideM:
Alrighty then I'll revert that part

cideM:
I pushed some changes that revert back to the old implementation while still supporting user classes on `htmlSpanLikeElement`s @jgm

jgm:
Looks good, thanks!

PR #5884, opened 2019-11-07
Switch to Text

despresc:
This is the start of the effort to switch `pandoc` to `Text` and bring it in line with the changes proposed in this [pull request](https://github.com/jgm/pandoc-types/pull/61) for `pandoc-types` and [this request](https://github.com/jgm/texmath/pull/146) for `texmath`. As of now:

- All of the common modules used in the readers have had their external `String` types changed to `Text`.
- All of the common modules except `Text.Pandoc.Parsing` now have a `Text.Pandoc.Legacy.*` version that re-exports the contents of the non-legacy version, suitable packed or unpacked. These allow the switching of most of the readers and writers to be more incremental, allowing testing throughout the process, and will all be removed at the end.
- Switching the interface of `Text.Pandoc.Parsing` to `Text` required changing the previous `[Char]` stream type to `Text` and having the existing combinators return `Text`. A few new combinators like `manyChar` were also added that pack the string that the corresponding `parsec` combinator would normally return, so that the readers depending on `Parsing` could incorporate the changes with minimal disruption to their logic. It may be desirable for efficiency to have the parsers operate differently, say by returning lazy `Text` or `[Char]`. The current decisions can still be changed or reversed without much difficulty.
- All of the readers that depend on `Parsing` were switched to `Text`. Most of the other readers have been switched as well.
- Most modules need varying degrees of internal refactoring (as the many `TODO text` comments sprinkled through the source will attest), and all of the writers still need to be switched.
- The current code passes its tests on my computer, but it still emits a large number of incomplete pattern (and similar) warnings because of the temporary `PatternSynonym` usage in the package. Rather than add `COMPLETE` pragmas here and in `pandoc-types` I have just silenced them while building and testing.

Incidentally, I noticed a reasonable amount of `count 1 anyChar` and similar constructions in some readers that could be rewritten to take advantage of existing (or new) combinators in `Parsing`.


jgm:
Excellent news.

> Incidentally, I noticed a reasonable amount of count 1 anyChar and similar constructions in some readers that could be rewritten to take advantage of existing (or new) combinators in Parsing.

Feel free to add new combinators if you see places where that would help.  Or this could be saved for later.

despresc:
The switch is complete, more or less. There will probably be areas where efficiency can be increased, but the public types now use `Text` as much as possible. Some other notes:

- `glob` in `Text.Pandoc.Class` still takes `String` since it seems to be intended as an interface to `Glob`, which uses strings. It seems to be used only once in the package, in the EPUB writer, so that is not hard to change.
- some internal types and functions still use `String`, especially the ones that relate to `FilePath` or need to interface with some external libraries that use that type. It is possible that I was too aggressive in switching some types to `Text` (resulting in excess packing and unpacking) but I tried to be careful.
- the tests often use `String` internally, probably more than what `tasty` requires, but I decided not to change those in this request

despresc:
The command-line arguments given to the filters in `Text.Pandoc.Filter` are also still `String`, I should mention, as are some of the types in `Text.Pandoc.App` (the epub subdirectory, pdf engine name, and pdf engine arguments). I can switch those to `Text` as well, though internally there will be some amount of packing and unpacking required.

jgm:
Great! We'll still need the changes to pandoc-citeproc before this can be merged into master.  I don't think it's necessary at this point to change pandoc-citeproc so it uses text throughout -- though you're welcome to do that if you like.  All that is needed are the changes necessary to make pandoc-citeproc work with this text-ified version of pandoc.  We could do that by expanding the use of the compatibility module (in `compat/`), or just require the latest pandoc and get rid of `compat` (I'm not sure why we need to ensure that the latest version of pandoc-citeproc compiles with earlier versions of pandoc).

jgm:
I'll fix the merge conflicts now.

despresc:
I think it would be wise to drop support for older `pandoc` versions. The `compat` module would have to become fairly intricate to support anything lower than the next release, given the extensive type changes.

As for switching `pandoc-citeproc` to `Text`, I will look through it and see what is easiest. It may be simpler to change types to `Text` as I incorporate the changes to the other packages, but I will not feel obligated to switch the whole thing.

jgm:
I've created a new branch `text` that merges these changes with current master.

jgm:
Feel free to remove `compat` in pandoc-citeproc, then, and depend on the latest version.

jgm:
I ran some benchmarks on current `master` and `text` branches. I was expecting to see a significant performance boost and memory reduction from the switch from String to Text.  The results were surprising.

Here's a simple comparsion converting the manual to HTML:
```
 % time ./pandoc-master-bin MANUAL.txt -o m.html +RTS -s
   1,660,240,344 bytes allocated in the heap
     187,821,408 bytes copied during GC
      14,683,576 bytes maximum residency (15 sample(s))
          89,024 bytes maximum slop
              14 MB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0      1583 colls,     0 par    0.173s   0.176s     0.0001s    0.0008s
  Gen  1        15 colls,     0 par    0.148s   0.166s     0.0111s    0.0249s

  TASKS: 4 (1 bound, 3 peak workers (3 total), using -N1)

  SPARKS: 0(0 converted, 0 overflowed, 0 dud, 0 GC'd, 0 fizzled)

  INIT    time    0.000s  (  0.003s elapsed)
  MUT     time    0.352s  (  0.359s elapsed)
  GC      time    0.321s  (  0.343s elapsed)
  EXIT    time    0.000s  (  0.012s elapsed)
  Total   time    0.673s  (  0.715s elapsed)

  Alloc rate    4,716,953,698 bytes per MUT second

  Productivity  52.3% of total user, 50.1% of total elapsed


real    0m0.748s
user    0m0.679s
sys     0m0.055s


% time ./pandoc-text-bin MANUAL.txt -o m.html +RTS -s
   1,752,195,424 bytes allocated in the heap
     161,574,640 bytes copied during GC
       9,750,176 bytes maximum residency (17 sample(s))
          94,896 bytes maximum slop
               9 MB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0      1677 colls,     0 par    0.164s   0.169s     0.0001s    0.0005s
  Gen  1        17 colls,     0 par    0.149s   0.163s     0.0096s    0.0176s

  TASKS: 4 (1 bound, 3 peak workers (3 total), using -N1)

  SPARKS: 0(0 converted, 0 overflowed, 0 dud, 0 GC'd, 0 fizzled)

  INIT    time    0.000s  (  0.003s elapsed)
  MUT     time    0.377s  (  0.384s elapsed)
  GC      time    0.314s  (  0.331s elapsed)
  EXIT    time    0.000s  (  0.005s elapsed)
  Total   time    0.691s  (  0.723s elapsed)

  Alloc rate    4,650,348,270 bytes per MUT second

  Productivity  54.5% of total user, 53.1% of total elapsed


real    0m0.756s
user    0m0.697s
sys     0m0.049s
```
(In doing this comparison, be sure to do a regular `stack install` since `make` compiles without optimizations.)

As you can see, pandoc isn't much faster in the `text` branch; it's actually consistently a bit slower.  Memory usage is lower (at least maximum residency is), but this may be balanced by more GC activity.  At any rate, the difference is less than I'd imagined it would be.

I also ran the benchmark suite for both branches.  Timings on this do tend to fluctuate a bit from one run to another, so we can only make coarse comparisons.  But `text` is often about the same speed as `master`, most often a bit slower, sometimes a bit faster.  There are a few striking data points that call out for explanation, though, and might be interesting clues:

- The RTF writer is significantly faster in `text` (11.4 ms vs 15.2)
- The commonmark writer is significantly faster in `text` (2.7 vs 4.3) - same with `gfm`
- The `markdown_phpextra` reader is much faster in `text` (40 ms vs 60!).  Interestingly, in the `text` branch the `markdown_phpextra` reader is in the same ballpark as `markdown_strict` and `markdown_mmd`, while in `master` it is much slower (and more like the full extended `markdown` reader).  So, apparently one of the extensions `markdown_phphextra` enables is quite slow with String but speeds up considerably with Text.
- The RST reader is significantly faster in `text` (76 vs 108 ms)
- The DocBook reader is significantly slower in `text` (37 vs 29 ms)
- The HTML reader is signifciantly faster in `text` (47 vs 67 ms)
- The LaTeX reader is significantly faster in `text` (38 vs 61 ms)
- The FB2 reader is MUCH faster in `text` (12 vs 52 ms)

The substantial gains in FB2 and LaTeX show that we can get real gains out of the move to `text`; I am assuming that some further code changes may be needed to get these gains in other parts of the code base, which may be doing too many conversions or something?  Anyway, this would be good to investigate, though it can wait until after these changes are merged.

despresc:
The DocBook reader relies heavily on `xml`, which uses `String` internally, so there is more packing in that reader than before. That may account for some of the difference there.

jgm:
Merged into master.  Thank you very much for taking this on: it's a long overdue change.

PR #5908, opened 2019-11-14
Add functions liftPandocPure and liftPandocIO, with a HasPandocError class

ddssff:
I'm not sure whether these are generally useful, or they just fit into a design pattern I've been using.

### DIFF
--- a/src/Text/Pandoc/Class.hs
+++ b/src/Text/Pandoc/Class.hs
@@ -73,6 +73,9 @@ module Text.Pandoc.Class ( PandocMonad(..)
                          , setTranslations
                          , translateTerm
                          , Translations
+                         , HasPandocError(fromPandocError)
+                         , liftPandocPure
+                         , liftPandocIO
                          ) where
 
 import Prelude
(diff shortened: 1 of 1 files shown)


jgm:
Could you give a little explanation of how you use these?

ddssff:
Generally speaking, if you are building a system of any size you start to accumulate a lot of error types, and these might turn into a heirarchy, where a `FileError` type might have one constructor to hold an `IOException`, another to hold a `PandocError`, and so on.  To keep the code in different parts of the system uncoupled, you give each subsystem's error type a class such as

```class HasPandocError e where fromPandocError :: PandocError -> e```

Now your pandoc code can throw an error of some type it doesn't know about, as long as that error type has a `HasPandocError` instance.

The idea of `liftPandocPure` and `liftPandocIO` is to use the `fromPandocError` function to transform the error type, which in the pandoc library is always `PandocError`.  So if I want to call pandoc from inside a LaTeXT monad I can say
```
latexify :: AppAbbrevs s e m => Markup -> LaTeXT m ()
latexify x = latexFromMarkup (liftPandocPure . pandocFromMarkdown) x

pandocFromMarkdown :: PandocMonad m => Text -> m Pandoc
pandocFromMarkdown t = softBreakToLineBreak <$> P.readMarkdown readOptsForMarkdown (removeCarriageReturns t)

latexFromMarkup :: AppAbbrevs s e m => (Text -> LaTeXT m Pandoc) -> Markup -> LaTeXT m ()
...
```

So `pandocFromMarkdown` gets lifted into `LaTeXT` m.

Issue #5556, opened 2019-06-06
Add Jira reader


mb21:
Since https://github.com/jgm/pandoc/issues/2497 we have a Jira writer to output [Jira markup](https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=all).

If there's demand, we could also add a Jira markup reader to convert from it.

phuffman:
At our company, Jira issues/tickets form the basis of our bi-weekly release notes. I'm able to access the Jira issues and convert them to Asciidoctor, but my conversion is not great and requires quite a bit of cleanup. (Jira as the basis for nicely formatted release notes would seem to be a common workflow.) A Jira reader would be terrific.

tarleb:
@phuffman (and all others with a similar setup) please help us by sharing some more details of your current and your intended workflow. The following points would be interesting to know:

- How do you currently fetch the tickets? Do you use Jira's "Export" functionality?
- How would you like pandoc do deal with the tickets? Should pandoc read the wiki markup, or would you rather is parses XML produced by the exporter?
- Jira's exporter creates a table for each issue. Should pandoc do the same?

phuffman:
@tarleb, thanks for your consideration.
- Our dev team works exclusively in Jira.
- The dev team lead identifies tickets of interest, populates two key fields -- a high level summary and noteworthy implementation details
- He sends me a Jira query string with the tickets of interest.
- I use a Powershell module called "JiraPS" which allows me to take the Jira query as given and iterate over the returned tickets.
- Each ticket is a Powershell object with text fields formatted as Jira wiki markup.

- I grab the few fields I need, perform a rudimentary conversion (Jira wiki to Asciidoctor), and save each ticket as an Asciidoc file.
- The result is a folder of individual Asciidoc files, one for each ticket. I create an Asciidoctor "container" document where I can sort/organize the tickets, add headers, etc.)
- As for the format of the individual Asciidoctor files ... I have a Powershell "here-string" (multi-line text string) formatted as an Asciidoc table. I do variable substitution to drop in the values.

In summary:
* I'm given a Jira query
* I use Powershell "JiraPS" to pull the tickets via api which returns Jira wiki markup.
* I do some rudimentary conversionsion on the fields I care about then drop the converted text in an Asciidoctor file, preformatted as an Asciidoctor table.
* The result is a folder of Asciidoctor files which get consolidated in an Asciidoctor container file, sorted and organized manually.

[...]

PR #5913, opened 2019-11-17
Add jira reader

tarleb:
Closes: #5556

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -240,6 +240,7 @@ header when requesting a document from a URL:
     - `html` ([HTML])
     - `ipynb` ([Jupyter notebook])
     - `jats` ([JATS] XML)
+    - `jira` ([Jira] wiki markup)
     - `json` (JSON version of native AST)
     - `latex` ([LaTeX])
     - `markdown` ([Pandoc's Markdown])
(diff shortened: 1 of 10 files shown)


tarleb:
It's not perfect yet, as some corner-cases are handled incorrectly, but I believe it to be usable.

tarleb:
Missing:

- Block quotes
- HTML-entities

tarleb:
Found more issues, closing till those are fixed.

jgm:
Is this ready to go?

tarleb:
This gives good results when parsing `test/writer.jira`, so I'm confident that it's ready.

jgm:
Would it be worth adding some "old-style" tests like we have for the other readers?
`jira-reader.jira` -> `jira-reader.native` -- perhaps just using `writer.jira`, and carefully checking the output?  This gives more complete test coverage than the unit tests.

tarleb:
True, I'll do that. I assume this won't be part of 2.8, so we could also wait till the printer part of jira-wiki-markup is done. This will allow me to do roundtrip tests, which should uncover problems and inconsistencies.

jgm:
OK, I'll wait til after 2.8 then.

tarleb:
Should be good now.

jgm:
Great, merged!

PR #5917, opened 2019-11-18
DokuWiki reader: parse markup inside monospace ('')

### LINKED ISSUE #5916: Convert dokuwiki with inline code
Hello,

I use pandoc version **2.7.3.**

There is a weird behaviour when dealing with inline code in dokuwiki format, when we use '%%' to prevent further interpretion.

**For example :** 
- you have this text in markdown format (1) : here is some `inline code`
- you convert (1) to dokuwiki format (2) : here is some ''%%inline code%%'' (it's ok, the "%%" part is not displayed when interpreting dokuwiki)
- you convert (2) back to markdown format : here is some `%%inline code%%`
In the end you have some extra "%%". 

**The same problem occurs with this example :** 
- you have this text in dokuwiki format : here is some ''%%inline code%%''
- you convert to HTML : %%inline code%% 
You have extra "%%" that are displayed in the HTML page.

See there for another example : https://groups.google.com/forum/#!topic/pandoc-discuss/rClN82ofG9o

link2xt:
Fixes #5916

### DIFF
--- a/src/Text/Pandoc/Readers/DokuWiki.hs
+++ b/src/Text/Pandoc/Readers/DokuWiki.hs
@@ -31,7 +31,7 @@ import Text.Pandoc.Definition
 import Text.Pandoc.Error (PandocError (PandocParsecError))
 import Text.Pandoc.Options
 import Text.Pandoc.Parsing hiding (enclosed, nested)
-import Text.Pandoc.Shared (crFilter, trim, underlineSpan, tshow)
+import Text.Pandoc.Shared (crFilter, trim, underlineSpan, stringify, tshow)
 
 -- | Read DokuWiki from an input string and return a Pandoc document.
 readDokuWiki :: PandocMonad m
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Shared.hs
```diff
@@ -543,6 +544,27 @@ inlineListToIdentifier exts =
       | otherwise = c == '_' || c == '-' || c == '.'
     spaceToDash = T.map (\c -> if isSpace c then '-' else c)
 
+-- | Convert an 'Inline' to plaintext.
+inlineToText :: Inline -> String
```
tarleb: This looks a lot like `stringify`, but with some differences (e.g., "\n" instead of " " for LineBreak). Maybe we can unify these functions somehow?

link2xt: I have replaced it with a call to `stringify` in this PR, thanks.

Replacing FB2 `plain` with `stringify` might be a good idea too, but it is out of scope of this PR.

jgm:
Great, thanks

PR #5921, opened 2019-11-21
LaTeX reader: parse \micro siunitx unit command

jlduran:
This was somehow missed in 884aef31c55e375cd62fcb55a71829d005087cae.


jgm:
Thanks!

Issue #5858, opened 2019-10-28
Minor formatting issues with Jira output


jgneff:
Thank you @tarleb for adding the Jira notation writer! That feature let me [automate](https://github.com/jgneff/netbeans-doc-src/blob/master/Makefile) the conversion of a long [README.md](https://github.com/jgneff/netbeans-doc-src) file in Markdown to an [issue description](https://issues.apache.org/jira/browse/NETBEANS-3296) in Jira notation. I found a few issues with the Jira output, summarized below. I'm using Pandoc version 2.7.3.

## Backslash

An extraneous backslash is inserted before an opening brace in a code block:

```ShellSession
$ printf '```\ntest {\n  one, two, three\n}\n```\n' \
    | ~/opt/pandoc-2.7.3/bin/pandoc --from=gfm --to=jira
{code}
test \{
  one, two, three
}

{code}

$
```

## Newlines

An additional newline character is added before the closing `{code}` of a code block:

```ShellSession
$ printf '```\nthis is some code\n```\n' \
    | ~/opt/pandoc-2.7.3/bin/pandoc --from=gfm --to=jira
{code}
this is some code

{code}

$
```

An additional newline character is added after a block quote:

```ShellSession
$ printf '>Block quote.\n\nNext paragraph.\n' \
    | ~/opt/pandoc-2.7.3/bin/pandoc --from=gfm --to=jira
bq. Block quote.


Next paragraph.

$
```

## Entities

The output contains entity references even for ASCII characters:

```ShellSession
$ printf '"-_\n' | ~/opt/pandoc-2.7.3/bin/pandoc --from=gfm --to=jira
&quot;&hyphen;&lowbar;

$
```

I was unable to find in the documentation the reason for converting ASCII characters to entities, nor any way to disable them.

## Wrap

It seems the option ```--wrap=preserve``` has no effect:

```ShellSession
$ printf 'Sentence one.\nSentence two.\n' \
    | ~/opt/pandoc-2.7.3/bin/pandoc --from=gfm --to=jira --wrap=preserve
Sentence one. Sentence two.

$
```

tarleb:
Sorry for the late reply.

- The backslash issue has already been fixed in the development version.
- Newlines: already fixed for code; I'm looking at fixing the block quote issue.
[...]

PR #5925, opened 2019-11-22
Jira writer: improve escaping of special chars

tarleb:
Backslash-escaping is used instead of HTML entities, as escaped
characters are easier to read this way. Furthermore, Confluence, which
seems to use a subset of Jira markup, seems to get confused by HTML
entities.

Closes: #5858


jgm:
Since I had to fix something else, I'll squeeze it in!

PR #5930, opened 2019-11-23
Add support for $toc-title$ to HTML (4 and 5)

afranke:
Inspired by [this trick](https://github.com/jgm/pandoc/wiki/How-to-add-a-%22Table-of-Contents%22-title-in-the-HTML-template). I didn’t see a valid reason for this to remain a trick and not be included by default and that also makes it more consistent with other output formats.


jgm:
thanks, it's a good idea

PR #5949, opened 2019-11-29
ConTeXt template: Adjustments to title formatting

denismaier:
Added `\setupinterlinespace` to `title`, `subtitle` and `author` elements => otherwise longer titles that run over multiple lines will look squashed as `\tfd` etc. won't adapt the line spacing to the font size.

### DIFF
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -135,14 +135,14 @@ $endfor$
 \starttext
 $if(title)$
 \startalignment[middle]
-  {\tfd $title$}
+  {\tfd\setupinterlinespace $title$}
 $if(subtitle)$
   \smallskip
-  {\tfa $subtitle$}
+  {\tfa\setupinterlinespace $subtitle$}
 $endif$
 $if(author)$
   \smallskip
-  {\tfa $for(author)$$author$$sep$\crlf $endfor$}
+  {\tfa\setupinterlinespace $for(author)$$author$$sep$\crlf $endfor$}
 $endif$
 $if(date)$
   \smallskip
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -146,7 +146,7 @@ $if(author)$
 $endif$
 $if(date)$
   \smallskip
-  {\tfa $date$}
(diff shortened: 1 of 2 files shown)


denismaier:
Hmmm, of course that fails. That's whole idea of changing the template, right? What's to do about this?

jgm:
Quoting from CONTRIBUTING.md:

> You can rebuild the golden tests in `tests/` by passing
`--accept` to the test script. (If you're using stack, `stack
test --test-arguments "--accept"`; or `make TESTARGS=--accept`).
Then check the changed golden files for accuracy, and
commit the changes.  For docx or pptx tests, open the files in Word
or Powerpoint to ensure that they weren't corrupted and that
they had the expected result, and mention the Word/Powerpoint
version and OS in your commit comment.

Or you can simply manually modify `test/writer.context`.

denismaier:
Ok. Tests manually modified. But things still fail. Any hints? Sorry to bother you here again. It must be rather trivial.

jgm:
As build logs show, this test is failing:
```
          +  68   {\tfa July 17, 2006}
1351
          -  68   {\tfa\setupinterlinespace July 17, 2006}
```
Looks like you didn't modify the template for date.  I don't know if you intended to, but the test file and template need to match in this respect.

denismaier:
Oh, I made three changes in both files, but obviously not the right ones... 
The template has `title`, `subtitle`, `author` and `date`, while the tests have only `title`, `author` and `date`.Is that mismatch intentional?

denismaier:
Concerning the latest fail: Is there a problem with the tests on windows? The PR seems now to pass all tests on Linux and Mac, but the Windows test runner seems to choke here:

```
pandoc> test (suite: test-pandoc, args: '--hide-successes')

Invalid argument `'--hide-successes''```

jgm:
Only a few formats support subtitle, and the writer tests are based on a generic document (which actually goes back to very early pandoc), so subtitle isn't tested by these.

jgm:
Let me try to fix that windows issue (which is due to a change I made recently to the ci tests).

PR #5956, opened 2019-12-02
Add title-background-image to revealjs template

### LINKED ISSUE #4498: revealjs: cannot set background in title slide
Following [Background in reveal.js](http://pandoc.org/MANUAL.html#background-in-reveal.js), the following syntax can be used to set a background image (etc.) on a particular slide:

```mdhe
# Slide {data-background-image='/images/bg.jpg'}
```

Is there an equivalent for the title slide (i.e. the slide derived from the [metadata block](http://pandoc.org/MANUAL.html#metadata-blocks))? I expected the following to work, to no avail.

```yaml
---
title: My talk
author: \@tlvince
date: 2018-03-26
data-background-image: /images/bg.jpg
---
```

```
❯ pandoc -v
pandoc 2.1.3
```

frederik-elwert:
This pull request adds a new variable `title-background-image` to the default reveal.js template. When set, it adds a `data-background-image` attribute to the title slide which is generated from the metadata.

Closes #4498

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2104,6 +2104,9 @@ These affect HTML output when [producing slide shows with pandoc].
 All [reveal.js configuration options] are available as variables.
 To turn off boolean flags that default to true in reveal.js, use `0`.
 
+`background-image`
+:   background image for all slides
+
 `revealjs-url`
 :   base URL for reveal.js documents (defaults to `reveal.js`)
 
@@ -2117,6 +2120,9 @@ To turn off boolean flags that default to true in reveal.js, use `0`.
 `slideous-url`
 :   base URL for Slideous documents (defaults to `slideous`)
 
+`title-background-image`
+:   background image for the title slide
+
 [reveal.js configuration options]: https://github.com/hakimel/reveal.js#configuration
 
 ### Variables for Beamer slides
@@ -2127,6 +2133,9 @@ These variables change the appearance of PDF slides using [`beamer`].
 :   slide aspect ratio (`43` for 4:3 [default], `169` for 16:9,
     `1610` for 16:10, `149` for 14:9, `141` for 1.41:1, `54` for 5:4,
     `32` for 3:2)
+
+`background-image`
+:   background image for all slides
 
 `beamerarticle`
 :   produce an article from Beamer slides
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -5349,6 +5349,9 @@ To set an image for a particular reveal.js slide, add
 `{data-background-image="/path/to/image"}`
 to the first slide-level heading on the slide (which may even be empty).
 
+To set an image for the title slide generated from the metadata,
+use the configuration variable `title-background-image`.
+
 In reveal.js's overview mode, the parallaxBackgroundImage will show up
 only on the first slide.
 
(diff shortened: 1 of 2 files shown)


mb21:
Nice!

perhaps the new variable should also be listed in the section of the MANUAL.txt: https://pandoc.org/MANUAL.html#variables-for-html-slides

frederik-elwert:
Thanks for the feedback. I thought about that, too. But since `background-image` is not mentioned there, I thought it might be confusing. Should I add both to that section? Or leave it as is, since it’s now in the section dedicated to background images?

jgm:
I think it makes sense to add both to the section on variables.

frederik-elwert:
Okay, I added the variables. I added also `background-image` to the list of beamer variables, as it is also supported by the beamer template.

lassepe:
how would one set other options like `data-background-size` for the title slide in this case? Can we maybe just add something like:

```html
<section id="$idprefix$title-slide" $for(title-slide-attributes)$ $title-slide-attributes$ $endfor$>
```

And allow the user to add arbitrary attributes to the title slides like so:

```yaml
title-slide-attributes:
    - data-background-image=./media/title-logos.svg
    - data-background-size=contain
```

jgm:
@lassepe makes a good point.  Would it make more sense to do it this way (more general)?

frederik-elwert:
Yes, agreed, thanks for the suggestion! I will prepare a new pull request. I am currently a bit busy, so it might take me a couple of days.

frederik-elwert:
I experimented a bit with this. I like the general approach, but it seems not to be as easy as I thought. The problem is:

Take a yaml entry like this:

```yaml
title-slide-attributes:
    - data-background-image="img/title.jpg"
    - data-background-size="contain"
```

The strings will get interpreted as Markdown, resulting in the output

```html
<section id="title-slide" data-background-image=“img/title.jpg”  data-background-size=“contain” >
```

or, worse, with `--html-q-tags`

```html
<section id="title-slide" data-background-image=<q>Bilder/samuel-zeller-VK284NKoAVU-unsplash.jpg</q>  data-background-size=<q>contain</q> >
```

I guess this is the reason why @lassepe omitted the quotes in his examples. That somehow works in HTML5, but it would break for multi value entries like `data-background-position=bottom right`.

I see two solutions for this:

Either escape the quotes (which would be explained in the documentation):

```yaml
title-slide-attributes:
    - data-background-image=\"img/title.jpg\"
    - data-background-size=\"30%\"
    - data-background-position=\"bottom right\"
```

Or use a more complex data structure:

```yaml
title-slide-attributes:
    - attribute: data-background-image
      value: Bilder/samuel-zeller-VK284NKoAVU-unsplash.jpg
    - attribute: data-background-size
      value: 30%
    - attribute: data-background-position
      value: bottom right
```

This seems to be the cleanest approach, but might be a bit too heavy.

I’m happy to proceed either way. Opinions?

jgm:
What about
```
title-slide-attributes:
  background-image: foo.jpg
  background-size: 30%
  background-position: bottom right
```

frederik-elwert:
@jgm That’s perfect, thanks. I didn’t know how to access arbitrary key/value maps in the template syntax, but now discovered the `pairs` filter. I think this is a clean solution. I created a new clean request #5981.

Issue #5963, opened 2019-12-05
Support all `Reveal.initialize` options.


lassepe:
Currently, not all `Reveal.initialize` options are supported. E.g. there is now `fragementInURL` and others that are not part of the `default.revealjs` template (at least not as of pandoc 2.7.3). These should either be added or we should replace the entire block of manually forwarded initialization options with a single for-loop over `initoptions`:

```javascript
      // Full list of configuration options available at:
      // https://github.com/hakimel/reveal.js#configuration
      Reveal.initialize({
$for(initoptions)$
        $initoptions$,
$endfor$

$if(zoomKey)$
        // Modifier key used to click-zoom to part of the slide
        zoomKey: '$zoomKey$',
$endif$
...
```
At least this would avoid having to update the template every time reveal.js adds or changes an option. However, specifying string-options is a bit annoying in this case as the use would be required to escape the quotes (like for the `transition` option below):

```yaml
initoptions:
    - "width: 1400"
    - "height: 788"
    - "slideNumber: true"
    - "transition: \\'slide\\'"
```
Maybe there is a better way to do this? Maybe there is also a way to avoid the quoting of these options in the first place by some form of reflection? In any case I feel like it would be desirable to forward arbitrary initialization options.

**Note:**

Note that this will change some of the behavior as currently the `default.revealjs` template sets some slightly different defaults (e.g. `history: true` by default while it is `false` by default in reveal.js). IMHO it is less confusing if defaults are close to the documentation of reveal.js but this should be debated.

jgm:
If there are specific options you want to add, feel free to propose them.
It doesn't make sense to add all of them, probably.
I would like to keep supporting them in the way we do currently.

lassepe:
May I ask what is the benefit of manually forwarding the options to `Reveal.initialize`? Is it to provide the user with a simpler syntax? Or is the motivation to allow for pandoc-specific default values?

mb21:
The advantage is that you can set them in the YAML metadata inside the markdown doc, and don't need to have a custom template..

lassepe:
[...]

PR #5981, opened 2019-12-10
Add title-slide-attributes variable to reveal.js template.

### LINKED ISSUE #4498: revealjs: cannot set background in title slide

frederik-elwert:
Second attempt to add support for background images in reveal.js title slides.

Fixes #4498.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2127,6 +2127,10 @@ To turn off boolean flags that default to true in reveal.js, use `0`.
 `slideous-url`
 :   base URL for Slideous documents (defaults to `slideous`)
 
+`title-slide-attributes`
+:   additional attributes for the title slide of reveal.js slide shows.
+    See [background in reveal.js and beamer] for an example.
+
 [reveal.js configuration options]: https://github.com/hakimel/reveal.js#configuration
 
 ### Variables for Beamer slides
@@ -5369,6 +5373,10 @@ Other reveal.js background settings also work on individual slides, including
 `data-background-size`, `data-background-repeat`, `data-background-color`,
 `data-transition`, and `data-transition-speed`.
 
+To add a background image to the automatically generated title slide, use the
+`title-slide-attributes` variable in the YAML metadata block. It must contain
+a map of attribute names and values.
+
 See the [reveal.js
 documentation](https://github.com/hakimel/reveal.js#slide-backgrounds)
 for more details.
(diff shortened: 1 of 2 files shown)


lassepe:
Does this work with boolean values? A quoted "false" is truethy in JavaScript I think.

frederik-elwert:
What would you need boolean values for? Do you have an example?

jgm:
Ah, nice!  A use of the new `/pairs` pipe from doctemplates.

lassepe:
> What would you need boolean values for? Do you have an example?
> 
> The cases regarding slide backgrounds I had on mind are CSS values passed as HTML data attributes. As far as I see, they don't get interpreted as JavaScript anywhere. This does not replace the reveal configuration variables that pandoc passes into the JS constructor.

Sorry I confused this with [another use case I recently thought about](https://github.com/jgm/pandoc/issues/5963). I think here it should not be an issue. Maybe it would still be more intuitive if the template does not add the quotes but they are added by the user (e.g. not hide the conversion to strings)?

frederik-elwert:
> Sorry I confused this with [another use case I recently thought about](https://github.com/jgm/pandoc/issues/5963). I think here it should not be an issue. Maybe it would still be more intuitive if the template does not add the quotes but they are added by the user (e.g. not hide the conversion to strings)?

Ah, right, that’s a different case, since that’s JavaScript. If users want to be explicit, they can wrap the options in quotes, e.g.,

```yaml
data-background-position: "left bottom"
```

However, for the YAML parser these are both strings and passed as such to the template. And in HTML, all attribute values should be quoted, regardless of their type. So I don’t see any need to allow for unquoted strings here or put the burden of quoting to the user.

As I see it, this is just a translation of key-value syntax from YAML to HTML. In YAML, it takes the `key: value` form, in HTML it is `key="value"`.

PR #6000, opened 2019-12-20
Added glsl mime type for WebGL maps

jaredlander:
Added glsl mime type for WebGL maps

### DIFF
--- a/src/Text/Pandoc/MIME.hs
+++ b/src/Text/Pandoc/MIME.hs
@@ -195,6 +195,7 @@ mimeTypesList =
            ,("gjc","chemical/x-gaussian-input")
            ,("gjf","chemical/x-gaussian-input")
            ,("gl","video/gl")
+           ,("glsl","text/plain")
            ,("gnumeric","application/x-gnumeric")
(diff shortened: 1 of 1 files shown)


jgm:
There isn't a special MIME type for GLSL?

jaredlander:
It appears to be just a code file as seen here: https://github.com/robertleeplummerjr/Leaflet.glify/blob/master/src/shader/fragment/dot.glsl

This is the code in that file:

```
precision mediump float;
uniform vec4 color;
uniform float opacity;

void main() {
    float border = 0.05;
    float radius = 0.5;
    vec2 center = vec2(0.5);

    vec4 color0 = vec4(0.0);
    vec4 color1 = vec4(color[0], color[1], color[2], opacity);

    vec2 m = gl_PointCoord.xy - center;
    float dist = radius - sqrt(m.x * m.x + m.y * m.y);

    float t = 0.0;
    if (dist > border) {
        t = 1.0;
    } else if (dist > 0.0) {
        t = dist / border;
    }

    //works for overlapping circles if blending is enabled
    gl_FragColor = mix(color0, color1, t);
}
```

PR #6004, opened 2019-12-22
doc/lua-filters.md: replace metadata example with image centering example

tarleb:
Metadata defaults can be given via the command line `--metadata-file`.
Adding raw format snippets is a common use case for Lua filters, so it
seems sensible to provide an example.

Thanks to @efx for proposing this filter.

Closes: pandoc/lua-filters#70

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -264,32 +264,33 @@ return {
 }
 ```
 
-## Default metadata file
+## Center images in LaTeX and HTML output
 
-This filter causes metadata defined in an external file
-(`metadata-file.yaml`) to be used as default values in a
-document's metadata:
+For LaTeX, wrap an image in LaTeX snippets which cause the image
+to be centered horizontally. In HTML, the image element's style
+attribute is used to achieve centering.
 
 ``` lua
--- read metadata file into string
-local metafile = io.open('metadata-file.yaml', 'r')
-local content = metafile:read("*a")
-metafile:close()
--- get metadata
-local default_meta = pandoc.read(content, "markdown").meta
+-- Filter images with this function if the target format is LaTeX.
+if FORMAT:match 'latex' then
+  function Image (elem)
+    -- Surround all images with image-centering raw LaTeX.
+    return {
+      pandoc.RawInline('latex', '\\hfill\\break{\\centering'),
+      elem,
+      pandoc.RawInline('latex', '\\par}')
+    }
+  end
+end
 
-return {
-  {
-    Meta = function(meta)
-      -- use default metadata field if it hasn't been defined yet.
-      for k, v in pairs(default_meta) do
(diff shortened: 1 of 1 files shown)


cagix:
just out of curiosity: why should the centering example replace the old metadata example? Maybe I've missed something, but the old metadata example demonstrated how the metadata values are accessed in a filter. the new example is very useful in itself, but there are no metadata used (at least according to the diff (https://github.com/jgm/pandoc/pull/6004/files))?

tarleb:
I'd like the examples to be ready-to-use, useful filters. The current metadata example is better replaced by a command line argument, so it doesn't quite fit anymore. I think of it less as replacing an example, and more as "adding a new example and remove an outdated example". The number of examples is in no way fixed.

There are still the examples "Setting the date in the metadata" and "Replacing placeholders with their metadata value", which hold the basics of metadata be modifiable. I'd also rather people look at pandoc/lua-filters for more complex examples.

cagix:
Ah, I see :)  I missed the fact this issue is just combining the adding of a new example and removing of an deprecated example into a single pull request ...

alerque:
The PR title probably didn't help ;-) I read the change a couple times before I saw the logic — but I do agree the existing example should go away even if you _don't_ add another one because it's actually a deprecated and needlessly clumsy way to accomplish the task shown. The incoming example is quite useful. Perhaps it isn't a replacement, but the removal and the addition are both good in their own rights.

jgm:
I'm fine with this.  But I think the new filter should add a conditional `if FORMAT == 'latex'`.  Better yet, provide sensible output for HTML as well?

tarleb:
I left out HTML, as this is what it would look like if done cleanly. It's not as readable as I'd wish:

```lua
-- Use CSS to center images in HTML
if FORMAT:match 'html' then
  return {
    {Meta = function (meta)
       local header_includes = meta['header-includes'] or pandoc.MetaList{}
       print(header_includes.t)
       -- ensure header-includes is a list
       header_includes = header_includes.t == 'MetaList'
         and header_includes
         or pandoc.MetaList(header_includes)
       -- add css to header
       table.insert(header_includes, pandoc.MetaBlocks {
         pandoc.RawBlock('html','<style>img{display:block;margin:auto}</style>')
       })
       meta['header-includes'] = header_includes
       return meta
    end}
  }
end

-- Othereise, don't do anything unless the target format is LaTeX.
if not FORMAT:match 'latex' then
  return
end

function Image(elem)
  -- Surround all images with image-centering raw LaTeX.
  return {
    pandoc.RawInline('latex', '\\hfill\\break{\\centering'),
    elem,
    pandoc.RawInline('latex', '\\par}')
  }
end
```

tarleb:
Seconds after sending this, I realized that I missed the obvious. HTML example is now included.

Issue #6014, opened 2019-12-27
JATS: support articleauthoring tag set


tarleb:
JATS offers three different tag sets: *archiving*, *publishing*, and *articleauthoring*, with the latter ones being a subset of the preceding tag set. Pandoc currently uses the *publishing* set. Since the easy authoring of article is one of pandoc's great strengths, it seems reasonable to also support the most restrictive variant of these tag sets.

As far as I can tell, this would only require different handling of footnotes. The *articleauthoring* tag set does not allow `fn-group` as part of the backmatter, so `<fn>` footnote elements have to be placed inline. All other changes can be handled in a custom template.

The current behavior has been implemented in response to #5511.

jgm:
Can you explain a bit more what you have in mind?  Looks like we already switched from publishing to the archiving DTD.  Are you saying that the response to #5511 (moving from inline footnotes to footnotes in fn-group in the backmatter) was needed only because we were using the publishing DTD, and now we can go back to inline fn elements?

Perhaps it would be clearest if you just prepared a PR with the changes you propose.

PR #6020, opened 2019-12-29
Update JATS dtd

### LINKED ISSUE #5634: @misc bibtex items cause invalid JATS output
I'm trying to produce JATS XML using from Markdown source and validating the output using the the [Pubmed central validator](https://www.ncbi.nlm.nih.gov/pmc/tools/xmlchecker)

## The problem

If the reference metadata (bibtex in this case) includes a `@misc` item, which is citing a GitHub repository then the the output JATS XML is not well-formed. Included in the zip file here is a minimal example that demonstrates the issue

This results in the following error:

```
37:    GitHub</publisher-name></element-citation></ref>
>                           ^
>  Element publisher-name was declared #PCDATA but contains non text nodes
38:  </ref-list>
```

Apologies for not fixing this with a pull request - I'm struggling with the `jats.csl` template and couldn't remedy locally.

## Steps to reproduce

**Pandoc version**

2.7.3

**Files**

Minimal example files demonstrating the problem: [jats-example-files.zip](https://github.com/jgm/pandoc/files/3363795/jats-example-files.zip)

Note the `jats.csl` file here is the same in `master`.

**Commands**

1. `pandoc -s --metadata-file metadata.json --bibliography references.bib --filter pandoc-citeproc  --to jats example.md -o output.xml`
2. Validate the output using the [Pubmed central validator](https://www.ncbi.nlm.nih.gov/pmc/tools/xmlchecker)
3. XML returns the error above.

**Also**

Note, if I updated the DTD to v1.1 (as in #5632) then I get the following error with the validator and [this file](https://github.com/jgm/pandoc/files/3363800/output.xml.zip)

```
37:    GitHub</publisher-name></element-citation></ref>
>                           ^
>  Element ext-link is not declared in publisher-name list of possible children
```

arfon:
The current DTD for the [JATS writer template](https://github.com/jgm/pandoc/blob/6c435a17cd48af40dbf87ebbec8baf6ac6aabe27/data/templates/default.jats#L5) is for Journal Publishing (`JATS-journalpublishing1.dtd`), which does not permit `ext-link` as a valid child (https://jats.nlm.nih.gov/publishing/tag-library/1.1/element/publisher-name.html).

This update modifies the default output template to be the less restrictive JATS archiving and interchange DTD which systems like PubMed use internally to represent their articles. The new dtd is as follows:

```
<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2d1 20190208//EN" "JATS-archivearticle1.dtd">
```
As suggested by @jgm - I proposed this change to the Pandoc mailing list [a few weeks ago](https://groups.google.com/forum/#!searchin/pandoc-discuss/jats%7Csort:date/pandoc-discuss/CPqStRl9qdQ/O5GbnbKrAwAJ) and haven't received any feedback on this suggestion.

Running the following command in [jats-example-files.zip](https://github.com/jgm/pandoc/files/4008669/jats-example-files.zip) 

```
pandoc -s --metadata-file metadata.json --bibliography references.bib --filter pandoc-citeproc --to jats example.md -o output.xml
```

Produces the file `output.xml` which is currently invalid. Updating the `default.jats` template as modified in this pull request produces `updated-output.xml` which is valid JATS according to the [Pubmed central validator](https://www.ncbi.nlm.nih.gov/pmc/tools/xmlchecker)

Fixes #5634

### DIFF
--- a/data/templates/default.jats
+++ b/data/templates/default.jats
@@ -2,7 +2,7 @@
 $if(xml-stylesheet)$
 <?xml-stylesheet type="text/xsl" href="$xml-stylesheet$"?>
 $endif$
-<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2d1 20190208//EN"
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2 20190208//EN"
                   "JATS-archivearticle1.dtd">
 $if(article.type)$
 <article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.2" article-type="$article.type$">
--- a/data/templates/default.jats
+++ b/data/templates/default.jats
@@ -5,9 +5,9 @@ $endif$
 <!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2d1 20190208//EN"
                   "JATS-archivearticle1.dtd">
 $if(article.type)$
-<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.1" article-type="$article.type$">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.2" article-type="$article.type$">
 $else$
-<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.1" article-type="other">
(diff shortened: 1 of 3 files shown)


tarleb:
I'm in very much in favor of this, as it would also remove the requirement to specify journal-metadata (which is required for the "publishing" tag set.)

Maybe we can even make this configurable? See also: #6014.

data/templates/default.jats
```diff
@@ -2,12 +2,12 @@
 $if(xml-stylesheet)$
 <?xml-stylesheet type="text/xsl" href="$xml-stylesheet$"?>
 $endif$
-<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Publishing DTD v1.1 20151215//EN"
-                  "JATS-journalpublishing1.dtd">
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2d1 20190208//EN"
```
tarleb: The draft versions are not recognized by some validators. I recommend to use `v1.2` instead.

PR #6029, opened 2020-01-04
add group biblatex citations if prefix present

UnkDevE:
Works for multiple prefixes, took a while but I'm quite happy with result, may need commits cleaning up and code review before going into main repo


UnkDevE:
I knew I forgot something, String has changed to Text, ok I'll fix that and then give another pull request

mb21:
Thanks for the pull! Is this in regard to a specific issue (link to github issue number?) or something discussed on pandoc-discuss, or..?

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1454,15 +1454,35 @@ citationsToBiblatex
...
                braces (literal (T.intercalate "," (map citationId (c:cs))))
-  | otherwise = do
+  | all (\cit -> null (citationSuffix cit)) (c:cs) 
+    = do  
+      let cmd = case citationMode c of 
+                    SuppressAuthor -> "\\autocites*"
+                    AuthorInText   -> "\\textcites"
+                    NormalCitation -> "\\autocites"
+
+      let pfxs = tail $ scanl (\prev cit -> let pfx = citationPrefix cit in
```
jgm: `tail` is a partial function and should be avoided generally.  It can cause an exception if the argument is an empty list.

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1454,15 +1454,35 @@ citationsToBiblatex
...
+                if pfx == [] then prev else pfx) [] (c:cs)
+
+      let pfxGroups = map (map snd) $ groupBy (\(pre, _) (pre1, _) -> pre == pre1)
+            (zip pfxs (c:cs))  
+      prefixes <- mapM (\group -> 
+                (fmap brackets) $ inlineListToLaTeX $ citationPrefix $ head group)
```
jgm: `head` is also a partial function.
`fmap brackets $ xxx` is more idiomatically written `brackets <$> xxx`.

UnkDevE:
Ah right! I was trying to the request all sorted with Text to string
business first

PR #6040, opened 2020-01-09
Lua: filtering element lists

tarleb:
Allow filtering of lists of inlines and lists of blocks via Lua filter functions "Inlines" and "Blocks", respectively.

E.g.:

    -- remove all spaces before a Cite element
    function Inlines (inlns)
      for i=#inlns,1, -1 do
        if inlns[i] == "Cite" and inlns[i-1] and inlns[i-1].t == "Space" then
          table.remove(inlns, i-1)
        end
      end
    end

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -136,13 +136,52 @@ Elements without matching functions are left untouched.
 See [module documentation](#module-pandoc) for a list of pandoc
 elements.
 
-## Execution order
+## Filters on element sequences
+
+For some filtering tasks, the it is necessary to know the order
+in which elements occur in the document. It is not enough then to
+inspect a single element at a time.
+
+There are two special function names, which can be used to define
+filters on lists of blocks or lists of inlines.
+
+[`Inlines (inlines)`]{#inlines-filter}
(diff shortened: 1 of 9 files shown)


tarleb:
Todo:

- [ ] docs
- [ ] tests
- [ ] make table that's passed to the filter functions an instance of pandoc.List

tarleb:
The Lua functions for this are currently called `ListOfInlines` and `ListOfBlocks`. Suggestions for better names are welcome.

jgm:
Another option would be just `Inlines` and `Blocks`.  Have you considered that?

tarleb:
I figured it might be too similar to the Inline and Block fallback functions, but other than that I like them more.

I don't have a strong opinion, either option seems fine to me.

tarleb:
I updated the PR to use `Blocks` and `Inlines` as function names. The previous version remains available as branch tarleb:ListOfInlines/ListOfBlocks.

jgm:
I think I like these names, so should I go ahead and merge?

tarleb:
Should be ok to merge 👍

Issue #5849, opened 2019-10-24
group biblatex citations if prefix present


rnuske:
Would it be possible to also group biblatex citations if there is only a simple prefix like `[cf. @a1;@a2;@a3]`?

I've seen that revision  a92e435 closing #4960 facilitates grouping of biblatex citations without pre- or suffixes. With current pandoc (2.7.3) I get the following (notation following [test/command/4960.md](https://github.com/jgm/pandoc/blob/a92e43575fb44c47263603b1dc5db9ef4f7421d3/test/command/4960.md))
```
% pandoc -t latex --biblatex
[@a1;@a2;@a3]
^D
\autocite{a1,a2,a3}
```
and
```
% pandoc -t latex --biblatex
[cf. @a1;@a2;@a3]
^D
\autocites[cf.~][]{a1}{a2}{a3}
```

The later prevents citation from being sorted automatically using `\usepackage[sortcites]{biblatex}`

Similar issues: #2065, #2091

jgm:
That makes sense to me.

UnkDevE:
hey, I would like to take this issue on, what should the output be for it to be able to sort automatically?

rnuske:
That's great.
BibLaTeX needs all keys to be in one set of curly braces separated by commas to understand that they belong to one group and to thus can be sorted. So it should be like
```
\autocites[cf.~][]{a1,a2,a3}
\autocites[DAP, e.g.~][]{b1,b2,b3,b4}
```

UnkDevE:
right, what do you do if there are multiple prefixes? do we have something like 
`\autocites[cf.~][]{a1,a2,a3}[cg.~]{b1,b2,b3}`
or could you do something like this:
`\autocites[cf.~, e.g.~][]{a1,a2,a3}{b1,b2,b3}`

UnkDevE:
also what happens when there are prefixed citations and non prefixed citations?

rnuske:
Multiple prefixes might be more complicated. That's why I asked for the simple case of only one prefix.

The behavior of pandoc 2.3.1 with multiple prefixes is
```
% pandoc -t latex --biblatex
[e.g. @a1;@a2;@a3; but also @b1;@b2;@b3]
^D
\autocites[e.g.][]{a1}{a2}{a3}[but also][]{b1}{b2}{b3}
```

I guess the following could be a viable solution at least it gets sorted correctly by biblatex-sortcites
```
[...]

PR #6058, opened 2020-01-10
Fixes: group biblatex citations if prefix present #5849

UnkDevE:
Allows citations to be grouped up into prefixes in LaTeX Writer.
commits need cleaning up, code looks pretty clean, may still need review.


src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1448,6 +1450,8 @@ stripLocatorBraces = walk go
   where go (Str xs) = Str $ T.filter (\c -> c /= '{' && c /= '}') xs
         go x        = x
 
+
```
jgm: Small suggestion: it makes it easier to review patches if irrelevant changes like the addition of blank space here, or the inserted newline at 1471/2, are left out.

test/command/5849-prefix.md
```diff
@@ -0,0 +1,8 @@
+```
+% pandoc -t latex --biblatex
+[e.g. @a1;@a2;@a3; but also @b1;@b2;@b3]
+^D
+\autocites[e.g.~][]{a1}{a2}{a3}[but also][]{b1}{b2}{b3}
+```
```
jgm: Is this an adequate test for this feature?
Because I get exactly the same output with current pandoc on this input.

jgm:
Thanks for the PR!  I made a couple preliminary comments; haven't had a chance to look more carefully at the code yet.

UnkDevE:
Cool I'll add some more tests, and remove that newline there, thanks!

jgm:
```
% pandoc -t latex --biblatex
[e.g. @a1; e.g. @a2;@a3; but also @b1;@b2;but also @b3]
^D
\autocites[e.g.~][]{a1}{a2}{a3}[but also][]{b1}{b2}{b3}
```
This doesn't actually do what was requested in #5849, does it?  I think the expected output would be
```
\autocites[e.g.~][]{a1}[e.g.~][]{a2,a3}[but also][]{b1,b2}[but also][]{b3}
```

UnkDevE:
Ok that makes more sense! thanks for finding it out, let me fix that then

test/command/5849-prefix.md
jgm: This revised test output looks right.
Maybe it would be worth testing suffixes as well as prefixes?  Something like
```
[e.g. @a1, ch. 3 and elsewhere; @a2; @a3; but also @a4; @a5]
```

UnkDevE:
It only works when there are no suffixes, so I don't think that would be a
problem

jgm:
Why shouldn't it work with suffixes as well, though. Even though the original issue didn't mention this, the example I gave seems along the same lines.

The algorithm for grouping could be:

- start the group with a citation with no suffix (and maybe a prefix)
- include zero or more citations with no suffix or prefix
- and optionally one citation with a suffix but no prefix

Repeat until all the citations have been processed.

UnkDevE:
It should, that would be exciting! I'll see what I can do

test/command/5849-prefix.md
UnkDevE: what would be the output for that test? I've writtern some code, but I've made the output like this:
`\autocites[e.g.~][ch. 3 and elsewhere]{@a1}[e.g.~][]{@a2}{@a3}[but also][]{@a4}{@a5}` Is this right?

UnkDevE:
right I've found a bug with the citations being read, when there is a suffix and a prefix present the citations prefixes and suffixes are read as null lists, also when there is a suffix present in the first group a comma appears from how the suffix is read. If you could direct me to the code that reads the latex code then I can fix my new commit.

jgm:
Sorry, I'm not clear what you're saying.  Are you saying you found a bug in some other part of pandoc, unrelated to your change, or a bug in your change?  And what code exactly do you want to find?  Something in the LaTeX reader?

UnkDevE:
I've found a bug in another part of pandoc, It makes the suffixes being read in the citation function comes up with `, ` in the suffix. When there is a suffix and a prefix in a identifier, everything is read as null.

jgm:
I'll understand better if you give an example with expected and actual output.

UnkDevE:
ok, for the prefix and suffix making all of the citations null bug:
```
        --- expected
        +++ pandoc -t latex --biblatex
        +   1 {[}e.g.~\textcite{a1}, ch.3 and elsewhere;\textcite{a2};\textcite{a3};
        +   2 but also \textcite{a4};\textcite{a5};{]}
        -   1 \autocites[e.g.~][ch.3 and elsewhere]{a1}[e.g.~][]{a2,a3}[but also][]{b1,b2}[but
        -   2 also][]{b3}

```
and for the suffix with a comma and a space before it:
```        
        --- expected
        +++ pandoc -t latex --biblatex
        +   1 \autocites[, blah]{a1,a2}[see][]{a3}
        -   1 \autocites[blah]{a1}{a2}[see][]{a3}
```

jgm:
What's the markdown input that produced these outputs?

UnkDevE:
markdown for first:
`[e.g. @a1, ch.3 and elsewhere;@a2;@a3; but also @a4;@a5;]`
second:
[@a1, blah; @a2; see @a3]

jgm:
The first markdown example has a syntax problem: the trailing semicolon. If you remove that the results are correct I think:
```
 % pandoc -t latex --biblatex
[e.g. @a1, ch.3 and elsewhere;@a2;@a3; but also @a4;@a5]
^D
\autocites[e.g.~][ch.3 and elsewhere]{a1}{a2}{a3}[but also][]{a4}{a5}
```
In the second case (with current master) I get
```
% pandoc -t latex --biblatex
[@a1, blah; @a2; see @a3]
^D
\autocites[blah]{a1}{a2}[see][]{a3}
```
which looks right.  So if there's a bug here it probably has to do with your cite-collapsing changes rather than something already in master.  (Or maybe an interaction between your changes and something else that could be fixed.)

Note that if you do `pandoc -t native` you can see how this is parsed;
```
[Para [Cite [Citation {citationId = "a1", citationPrefix = [], citationSuffix = [Str ",",Space,Str "blah"], citationMode = NormalCitation, citationNoteNum = 0, citationHash = 0},Citation {citationId = "a2", citationPrefix = [], citationSuffix = [], citationMode = NormalCitation, citationNoteNum = 0, citationHash = 0},Citation {citationId = "a3", citationPrefix = [Str "see"], citationSuffix = [], citationMode = NormalCitation, citationNoteNum = 0, citationHash = 0}] [Str "[@a1,",Space,Str "blah;",Space,Str "@a2;",Space,Str "see",Space,Str "@a3]"]]]
```
Note here that citationSuffix includes the leading comma -- I think that's the problem you're having? If we just have
```
 % pandoc -t latex --biblatex
[@a, blah]
\autocite[blah]{a}
```
then pandoc removes the comma.  Something in the code you've added has blocked the comma-removing code, I guess.
I believe this happens in the `citeArguments` function.

UnkDevE:
Thanks that's a big help!

UnkDevE:
It works now, however I need to go in and clean some of the code up, it's a bit messy

jgm:
Great, I'll wait to look at it until you've had a chance to clean up.

UnkDevE:
cleaned up, when you have time have a look at it, I wasn't too sure about the ``citeArgumentsList`` function but I think it works well.

test/command/5849-prefix.md
```diff
@@ -0,0 +1,43 @@
...
+also][]{a4,a5}
+```
+```
+% pandoc -t latex --biblatex
+[e.g. @a1, blah;@a2, ch.3 and elsewhere;@a3; but also @b4;@b5]
+^D
+\autocites[e.g.~][blah]{a1}[e.g.~][ch.3 and
+elsewhere]{a2}[e.g.~][]{a3}[but also][]{b4,b5}
+```
+
```
jgm: Why don't you get rid of these blanks at the  end of the test file?

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1472,18 +1483,36 @@ citationsToBiblatex (c:cs)
...
+                                    (fn $ drop 1 $ takefst, [cid]):prev
+                             | otherwise = (\(a, b) -> (fn a, cid:b):drop 1 prev) 
+                                    =<< take 1 prev
+                  grouper' pfx [] cid = ([pfx], [cid]):prev
+                  grouper' pfx sfx cid = ([sfx, pfx], [cid]):prev 
+
+        correctOrder (pfxs, ids) = (reverse pfxs, reverse ids)
+        citeArgsList ((p:s:_), k) = citeArgumentsList p s k
+        citeArgsList ((p:_), k) = citeArgumentsList p [] k
+        citeArgsList ([], k) = citeArgumentsList [] [] k
```
jgm: This looks like it's probably correct, but I think the code could be clearer (and then it would be more obvious whether it's right).   For example, I think it would be clearer if you used pattern matching for `addToGroup` instead of `take 1` and then `drop 1`.  Something like
```
case prev of
  ((a,b):rest
    | null a ->
    | otherwise ->
  [] -> 
```
This would avoid the need for the confusing `takefst` stuff and make it clearer what's going on.

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1423,24 +1423,35 @@ citeCommand c p s k = do
...
-          _ -> s
-        _   -> s
-  pdoc <- inlineListToLaTeX p
-  sdoc <- inlineListToLaTeX s'
-  let optargs = case (isEmpty pdoc, isEmpty sdoc) of
+citeArgumentsList :: PandocMonad m
+              => [Inline] -> [Inline] -> [Text] -> LW m (Doc Text)
+citeArgumentsList _ _ [] = return empty
+citeArgumentsList p s (cid:k) 
+  | k == [] = do
```
jgm: `null k` is better.

jgm: It wouldn't hurt to have more descriptive argument names, or a comment on top of the function indicating what is what.

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1423,24 +1423,35 @@ citeCommand c p s k = do
...
+citeArgumentsList _ _ [] = return empty
+citeArgumentsList p s (cid:k) 
+  | k == [] = do  
+      pdoc <- inlineListToLaTeX p
+      sdoc <- inlineListToLaTeX s' 
+      return $ (optargs pdoc sdoc) <> braces (literal cid)
+  | otherwise = do  
+      pdoc <- inlineListToLaTeX p
+      sdoc <- inlineListToLaTeX s'
+      return $ (optargs pdoc sdoc) <>
```
jgm: Fair amount of code duplication here.  Would it make sense to put the conditional branch further down, since you're processing `pdoc` and `sdoc` the same on both branches?

UnkDevE:
All of the changes you suggested have been made in this commit, I liked the case statment as it made the grouper function very clear. look over it when you ready.

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1423,24 +1423,32 @@ citeCommand c p s k = do
...
+              => [Inline] -> [Inline] -> [Text] -> LW m (Doc Text)
+citeArgumentsList _ _ [] = return empty
+citeArgumentsList pfxs sfxs (cid:ids) = do  
+      pdoc <- inlineListToLaTeX pfxs
+      sdoc <- inlineListToLaTeX sfxs' 
+      if null ids then  
+            return $ (optargs pdoc sdoc) <> braces (literal cid)
+      else return $ (optargs pdoc sdoc) <> 
+              (braces (literal (T.intercalate "," (cid:ids))))
+      where sfxs' = stripLocatorBraces $ case sfxs of
```
jgm: I'm not sure this conditional is needed, since when ids is null, `T.intercalate "," (cid:ids) == cid`, right?
So the else clause by itself should be all you need.

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1472,18 +1480,36 @@ citationsToBiblatex (c:cs)
...
+                                 | otherwise -> (fn $ pfxs, [cid]):prev
+                             (([], ids):rest) -> (fn [[]], cid:ids):rest
+                             [] -> (fn [[]], [cid]):prev
+                  grouper' pfx [] cid = ([pfx], [cid]):prev
+                  grouper' pfx sfx cid = ([sfx, pfx], [cid]):prev 
+
+        correctOrder (pfxs, ids) = (reverse pfxs, reverse ids)
+        citeArgsList ((p:s:_), k) = citeArgumentsList p s k
+        citeArgsList ((p:_), k) = citeArgumentsList p [] k
+        citeArgsList ([], k) = citeArgumentsList [] [] k
```
jgm: This is getting a lot clearer but I still think it could be made more perspicuous.

As I understand it, `foldl' grouper` produces a list of pairs `(affixes, citeIds)` in which `affixes` is a list with either 0, 1, or 2 members.  If 0, we have neither prefix nor suffix. If 1, it's a prefix.  If 2, then the first is a prefix and the second is a suffix.  The function `citeArgsList` takes this list and converts it into two values, for prefix and suffix, which it passes to `citeArgumentsList`.  This all seems overly complicated and a bit fragile.  We can encode this better in the types.  Why not have the fold contsruct a list of triples `(prefix, suffix, ids)` and then change the type of `citeArgumentList` so that it can *directly* consume these triples?  THen you avoid the need for `citeArgsList` and simplify the code.  Or, even more haskelly, you could introduce a data type
```
type Prefix = [Inline]
type Suffix = [Inline]
type CiteId = Text
data CiteGroup = Prefix Suffix [CiteId]
```

I think this change would also allow you to simplify the definition of `grouper` and make it a lot clearer.

UnkDevE: nice I get what you saying and you're understanding is correct, I also found the code fragile while working with it, will do soon.

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1486,30 +1489,25 @@ citationsToBiblatex (c:cs)
...
-        citeArgsList ((p:s:_), k) = citeArgumentsList p s k
-        citeArgsList ((p:_), k) = citeArgumentsList p [] k
-        citeArgsList ([], k) = citeArgumentsList [] [] k
+                             ((CiteGroup pfx oSfx ids):rest) 
+                                 | null oSfx ->
+                                     (fn (CiteGroup pfx oSfx (cid:ids))):rest
+                                 | otherwise -> (fn (CiteGroup pfx [] [cid])):prev
+                             [] -> (fn (CiteGroup [] [] [cid])):prev
+                  grouper' pfx [] cid = (CiteGroup pfx [] [cid]):prev
+                  grouper' pfx sfx cid = (CiteGroup pfx sfx [cid]):prev
```
jgm: This is so much clearer now!  The only improvement I'd suggest now would be flattening out the logic in `grouper`.  It's awfully hard to read the code with three levels of nested `where`s - `grouper`, then `grouper'`, then `addToGroup`.  I think I'd understand better if it were flattened out, ideally into one level.  Maybe there's a reason for not doing this, but it's what I'd suggest for code readability.  E.g. my first reaction to line 1501 was "wait, is the previous suffix just being thrown away?"  But then I discovered that the function is never applied to a CiteGroup with non-null suffix. It would be better if the logic were more perspicuous here.

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1494,20 +1494,14 @@ citationsToBiblatex (c:cs)
...
-                                 | null oSfx ->
-                                     (fn (CiteGroup pfx oSfx (cid:ids))):rest
-                                 | otherwise -> (fn (CiteGroup pfx [] [cid])):prev
-                             [] -> (fn (CiteGroup [] [] [cid])):prev
-                  grouper' pfx [] cid = (CiteGroup pfx [] [cid]):prev
-                  grouper' pfx sfx cid = (CiteGroup pfx sfx [cid]):prev 
+  where grouper prev cit = case prev of  
+         ((CiteGroup oPfx oSfx ids):rest)
+             | null oSfx && null pfx -> (CiteGroup oPfx sfx (cid:ids)):rest
+             | null pfx -> (CiteGroup oPfx sfx [cid]):prev
```
jgm: Much clearer!  I understand line 1499:
```
(prefix1 cite1 [no suffix]) ([no prefix] cite2 suffix2) ==> prefix1 [citet,cite2] suffix2
```
But I'm not sure I get 1500.  Doesn't it amount to this?
```
(prefix1 cite1 suffix1) ([no prefix] cite2 suffix2) ==> prefix1 cite1 suffix1 ; prefix1 cite1 suffix2
```
which would mean prefix1 (oPfx) gets doubled? Maybe I'm missing something.

UnkDevE: that's a new suffix but no new prefix, the new suffix ends the group and starts with that same prefix.

jgm: Hm, still not sure I get it.  Do you have a test case like
```
[pref1 @cite1 suff1; @cite2 suff2]
```
and what does it spit out?

UnkDevE: Yes we have:
 ``% pandoc -t latex --biblatex
[e.g. @a1, blah;@a2, ch.3 and elsewhere;@a3; but also @b4;@b5]``
 which spits out:
``\autocites[e.g.~][blah]{a1}[e.g.~][ch.3 and
elsewhere]{a2}[e.g.~][]{a3}[but also][]{b4,b5}``

jgm: I see: so in this case the `e.g.~` gets repeated for a1, a2, and a3. Is that expected behavior?
I would have expected just one `e.g.~` at the beginning if I were writing this markdown.

UnkDevE: So would you want something like 
``\autocites[e.g.~][blah]{a1}[ch.3 and elsewhere]{a2}[]{a3}[but also][]{b4,b5}``?

jgm: Yes, that's what I'd expect.  I suppose you could check back with the author of the original issue, but I can't really imagine anyone wanting the `e.g.` to be repeated like that.
And imagine something like
```
[in order of importance, @a1, p. 34; @a2, p. 44]
```
Here it would be quite bad if this were rendered "in order of importance, Smith 2000, p. 34, in order of importance, Jones 1999, p. 44".

UnkDevE:
Yeah I agree okay will do that soon

src/Text/Pandoc/Writers/LaTeX.hs
UnkDevE: is this output correct:
\autocites[e.g.~][blah]{a1}[ch.3 and elsewhere]{a2}{a3}[but also][]{b4,b5}
does it need another bracket between {a2} and {a3}?

jgm: I tried it.  With a numerical style it comes out as

> [e.g. 1, blah, 2, ch.3 and elsewhere, 3, but also 4, 5]

With author-year

> (e.g. Doe 2005, blah; Doe 2006, ch.3 and elsewhere; Doe 2007; but also Doe 2008; Doe 2009)

This looks right to me!

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -1497,7 +1497,8 @@ citationsToBiblatex (c:cs)
   where grouper prev cit = case prev of  
          ((CiteGroup oPfx oSfx ids):rest)
              | null oSfx && null pfx -> (CiteGroup oPfx sfx (cid:ids)):rest
-             | null pfx -> (CiteGroup oPfx sfx [cid]):prev 
+             | null pfx && null sfx -> (CiteGroup oPfx [] [cid]):prev 
+             | null oPfx && null sfx && null pfx -> (CiteGroup [] [] [cid]):prev 
          _ -> (CiteGroup pfx sfx [cid]):prev
          where pfx = citationPrefix cit
```
jgm: Line 1501 is superfluous since the same case will be handled the same way by line 1502, correct?
 I'm puzzled about line 1500.  Isn't this again duplicating a prefix, which is what we wanted to avoid?  Do you have a test case that exercises this condition, and what does it look like?

UnkDevE: No it isn't, it passes the test cases because of the ``optargs`` code in ``citeArgumentsList``, however you're right line 1501 is not needed

UnkDevE: checked over it again, it's not because of that it's because of the prefix before was null in the test, looking at it the tests only passes because of that,  this wouldn't work for some thing like:
``[e.g. @a1, ch.3 and elsewhere;@a2; but also @a4;@a5]`` I'll give this another try.

jgm:
Great, looks perfect now!

PR #6059, opened 2020-01-10
docs/lua-filters.md: clarify filter function execution order

tarleb:


tarleb:
This information is currently hidden in the example "Replacing placeholders with their metadata value".

PR #6061, opened 2020-01-11
doc/lua-filters.md: unify, fix anchors and internal links

tarleb:
Links and anchors now follow consistent conventions, like lowercase-only
anchor names.

This breaks some links to specific sections in the document, but will
make it much easier to link documentation in the future.


tarleb:
As far as I can tell by some googling, it seems that most links on the web are to sections whose anchor ids remain unchanged. I don't expect this change to cause problems.

PR #6064, opened 2020-01-14
JATS writer: simplify JATS type

tarleb:
The JATSVersion type appears to be unused.


tarleb:
Will propose a different change in a subsequent PR, thus closing again.

PR #6067, opened 2020-01-14
JATS writer: allow choosing the JATS tag set

tarleb:
The JATS tag set can be set via the `jats-tag-set` meta variable.
Recognized values are

- `archiving` for the "Archiving and Interchange Tag Set",
- `publishing` for the "Journal Publishing Tag Set", and
- `articleauthoring` for the "Article Authoring Tag Set."

Closes: #6014


jgm:
I wonder if it would be better to support this the way we now support epub v2/v3:  export different writer functions, which set the tag set and version behind the scenes.  This would mean there would be different targets -- `jats_archiving`, etc. -- with `jats` defaulting to one of them.  There could also be separate templates (as with epub, html).  One advantage of this approach would be that you could produce fragments.  The current approach is set up for documents with metadata.

tarleb:
Aah yes, that would be much cleaner. I'll go that route.

tarleb:
Missing

- [ ] docs
- [ ] tests

jgm:
Should we capitalize as `writeJATSPublishing` etc.?  Wasn't sure.

jgm:
Another thought.  We could avoid a major version bump in the API by exporting `writeJATS` as a synonym of (whatever the default one is) and using some other name for the common executive function.  I'm not sure there's all that big a reason to avoid a major bump, but it does have the side effect of forcing upgrades to everything that specifies upper bounds on pandoc.

tarleb:
> Should we capitalize as `writeJATSPublishing` etc.? Wasn't sure.

Unfortunately, we are not very consistent in this regard. Examples: `writeHtml5`, `writeEPUB3` `runIOorExplode`, and `getEPUBMetadata` all use different capitalization strategies. I think I'd prefer `writeJatsPublishing`, with `writeJats` being the parameterized function. Then we can keep `writeJATS` for backwards compatibility.

src/Text/Pandoc/Writers.hs
```diff
@@ -148,10 +149,10 @@ writers = [
   ,("docbook"      , TextWriter writeDocbook5)
   ,("docbook4"     , TextWriter writeDocbook4)
   ,("docbook5"     , TextWriter writeDocbook5)
-  ,("jats"         , TextWriter writeJATSarchiving)
-  ,("jats_articleauthoring", TextWriter writeJATSarticleauthoring)
-  ,("jats_publishing" , TextWriter writeJATSpublishing)
-  ,("jats_archiving" , TextWriter writeJATSarchiving)
+  ,("jats"         , TextWriter writeJATS)
```
jgm: I'd just use writeJatsAarchiving here, so it's clear and explicit.
(And we may want to deprecate writeJATS at some point.)

jgm:
If you rebase on master it should fix the CI failure.

tarleb:
Rebased and fixed. This could still use some more tests, but should be mergeable as-is.

src/Text/Pandoc/Templates.hs
```diff
@@ -90,6 +90,7 @@ getDefaultTemplate writer = do
        "docbook" -> getDefaultTemplate "docbook5"
        "epub"    -> getDefaultTemplate "epub3"
        "beamer"  -> getDefaultTemplate "latex"
+       "jats"    -> getDefaultTemplate "jats_archiving"
```
jgm: Don't we need `jats_archiving` and the others here as well?

src/Text/Pandoc/Writers/JATS.hs
```diff
@@ -43,19 +48,44 @@ import Text.Pandoc.XML
...
+-- | Convert a @'Pandoc'@ document to JATS (Journal Publishing Tag Set.)
+writeJatsPublishing :: PandocMonad m => WriterOptions -> Pandoc -> m Text
+writeJatsPublishing = writeJats TagSetPublishing
+
+-- | Convert a @'Pandoc'@ document to JATS (Archiving and Interchange
+-- Tag Set.)
+writeJatsArticleAuthoring :: PandocMonad m => WriterOptions -> Pandoc -> m Text
+writeJatsArticleAuthoring = writeJats TagSetArticleAuthoring
+
+-- | Alias for @'writeJatsArchiving'@.
```
jgm: Maybe worth adding that this is left here for backwards compatibility and may be deprecated in the future. Recommend using one of the explicit writers.

jgm:
I think a few more tests would be worth adding at this point, or we won't get around to it later.

src/Text/Pandoc/Templates.hs
tarleb: The templates for those have the same name as the format specifier. So they are handled by the default case a bit further down.

jgm: Ah yes, I see.

tarleb:
I added old-style writer tests and checked each file against the [JATS4R validator](https://validator.jats4r.org/): The files `writer.jats_archiving` and `writer.jats_articleauthoring` are valid JATS, but `writer.jats_publishing` is missing some necessary journal metadata. I think that is ok, but in the future we may want to emit a warning if required metadata is missing.

PR #6090, opened 2020-01-27
Update URLs and use HTTP**S** wherever possible

salim-b:
I've changed _all_ `http:` URLs of which the respective web servers have properly configured HTTPS to their `https` equivalents.

Plus a bunch of URLs have simply been updated.


jgm:
Terrific, thanks for doing that!

PR #6102, opened 2020-02-02
Clean up a confusing triple negative

josephcsible:
There's currently `unless`, `not`, and `notParaOrPlain` in the same expression, which is a rather confusing triple negative. Replace `notParaOrPlain` with `paraOrPlain` and switch to `all` from `any` to clean this up.


jgm:
Good improvement!

PR #6105, opened 2020-02-03
Clean up overcomplicated maybe logic

josephcsible:
We're using maybe in redundant ways. Remove it altogether in one case and simplify it to fromMaybe in another.


jgm:
I'm not sure the change is an improvement.  The replacement is about as long, and `maybe` has the advantage of being in Prelude, which is why I tend to reach for it.

jgm:
But definitely the redundancy should be eliminated, thanks for noticing that. As for the other, it's a matter of taste and I'm happy to merge it!

PR #6106, opened 2020-02-03
Subdivide Text.Pandoc.Class into small modules

tarleb:
The module `Text.Pandoc.Class` is large and combines different aspects, making it difficult to understand and extend. The module is split and spread over (unexported) submodules `CommonState`, `PandocMonad`, `IO`, and `Pure`.


tarleb:
Motivation: I'm trying to create a new type `PandocLua`, which would be a PandocMonad instance and should simplify our Lua code.

jgm:
Nice idea!  I agree that this is beneficial.

tarleb:
It is only a first step, but I think this would be ok to merge. I have also added haddock comments for all functions in module `PandocMonad`, just in case we choose to expose the module at some point.

jgm:
looks good to me!

PR #6111, opened 2020-02-04
Swap suboptimal uses of maybe and fromMaybe

josephcsible:
Anywhere `maybe` is used with `id` as its second argument, using `fromMaybe` instead will simplify the code. Conversely, anywhere `fromMaybe` is used with the result of `fmap` or `<$>` as its second argument, using `maybe` instead will simplify the code.


jgm:
I agree, thanks!

PR #6113, opened 2020-02-04
Resolve HLint warnings

tarleb:
HLint distinguishes between *errors*, *warnings*, and *suggestions*. These changes ensure that only *suggestions* remain. All *warnings* are either fixed or, if more appropriate, hlint is configured to ignore them.

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -84,7 +84,7 @@ convertWithOpts opts = do
   let needsCiteproc = isJust (lookupMeta "bibliography"
                                 (optMetadata opts)) &&
                       optCiteMethod opts `notElem` [Natbib, Biblatex] &&
-                      all (not . isPandocCiteproc) filters
+                      not (any isPandocCiteproc filters)
   let filters' = filters ++ [ JSONFilter "pandoc-citeproc" | needsCiteproc ]
 
   let sources = case optInputFiles opts of
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -342,7 +342,7 @@ readSource src = case parseURI src of
                    else BS.readFile fp
           E.catch (return $! UTF8.toText bs)
              (\e -> case e of
-                         TSE.DecodeError _ (Just w) -> do
+                         TSE.DecodeError _ (Just w) ->
                            case BS.elemIndex w bs of
                              Just offset -> E.throwIO $
                                   PandocUTF8DecodingError (T.pack fp) offset w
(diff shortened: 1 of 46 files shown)


src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -114,14 +113,14 @@ pandocTitleBlock tit auths dat =
 
 mmdTitleBlock :: Context Text -> Doc Text
 mmdTitleBlock (Context hashmap) =
-  vcat $ map go $ sortBy (comparing fst) $ M.toList hashmap
+  vcat $ map go $ sortOn fst $ M.toList hashmap
```
josephcsible: In this particular case, `sortBy` and `comparing` are better than `sortOn`, since `fst` is basically free. If you check the source of `sortOn`, you'll see that it does a bunch of mapping to save the result, but in this case, the mapping is more expensive than just recomputing the result is. (Now in src/Text/Pandoc/Shared.hs, for `rank`, this was a good change.)

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -143,7 +142,7 @@ yamlMetadataBlock v = "---" $$ (contextToYaml v) $$ "---"
 
 contextToYaml :: Context Text -> Doc Text
 contextToYaml (Context o) =
-  vcat $ map keyvalToYaml $ sortBy (comparing fst) $ M.toList o
+  vcat $ map keyvalToYaml $ sortOn fst $ M.toList o
```
josephcsible: Same as above; I don't think this makes sense for `fst`.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -163,7 +163,7 @@ inTextStyle d = do
                        [("style:name", styleName)
                        ,("style:family", "text")]
                        $ selfClosingTag "style:text-properties"
-                          (sortBy (comparing fst) . Map.toList
+                          (sortOn fst . Map.toList
```
josephcsible: Again with `fst`.

src/Text/Pandoc/Writers/Markdown.hs
josephcsible: By the way, I just opened ndmitchell/hlint#860 about this.

src/Text/Pandoc/App.hs
```diff
@@ -84,7 +84,7 @@ convertWithOpts opts = do
   let needsCiteproc = isJust (lookupMeta "bibliography"
                                 (optMetadata opts)) &&
                       optCiteMethod opts `notElem` [Natbib, Biblatex] &&
-                      all (not . isPandocCiteproc) filters
+                      not (all isPandocCiteproc filters)
```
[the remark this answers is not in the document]
tarleb: Ouch. Good catch, thanks.

src/Text/Pandoc/Writers/Markdown.hs
tarleb: Interesting, thanks. "Better" is a very ambiguous measure, and I think this is an example where the improved readability might be worth it.

tarleb:
Closing in favor of #6117.

PR #6115, opened 2020-02-05
Simplify an overcomplicated filtering function

josephcsible:
There's no need to use `catMaybes`, `uncurry`, `bool`, etc., just to get elements where the second element of a tuple is True.


tarleb:
Honestly took me a while to verify that this is correct. So much more readable! :+1:

tarleb:
Could you include this in your other PR? Then I can rebase my hlint fixes against yours.

tarleb:
Actually, never mind. I'm going to open a PR on your fork against the `cleanup` branch. Then we can have all changes in one PR.

josephcsible:
Okay. I think this can be merged separately either way, since there's no HLint changes around this area of the code once this change is made, so it doesn't conflict with my other PR.

PR #6116, opened 2020-02-05
Remove our bool function

josephcsible:
Data.Bool already provides a bool function identical to this one.


jgm:
It was added in base 4.7 - probably after this module was written!

josephcsible:
Actually it wasn't. Version 4.7 of base was released with GHC 7.8.1 in April 2014, and our `bool` was added in f068093 in July 2015.

jgm:
I'm guessing the author of that module wanted to retain compatibility with earlier versions of GHC.  But at this point we don't support < 7.8 anyway.

PR #6117, opened 2020-02-05
Various minor cleanups and refactoring

josephcsible:


josephcsible:
Let me know if there's any particular refactorings here you don't like, or if you'd rather have separate PRs for separate kinds of refactoring.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -443,7 +443,7 @@ elementToStr x = x
 parseInline :: PandocMonad m => Content -> JATS m Inlines
 parseInline (Text (CData _ s _)) = return $ text $ T.pack s
 parseInline (CRef ref) =
-  return $ maybe (text $ T.toUpper $ T.pack ref) text $ T.pack <$> lookupEntity ref
+  return $ text $ fromMaybe (T.toUpper $ T.pack ref) $ T.pack <$> lookupEntity ref
```
tarleb: How about

```haskell
return . text . maybe (T.toUpper $ T.pack ref) T.pack $ lookupEntity ref
```

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -121,7 +121,7 @@ mmdTitleBlock (Context hashmap) =
                  | null xs        -> empty
                  | otherwise      -> k' <> ":" <> space <>
                                       hcat (intersperse "; " $
-                                          catMaybes $ map fromVal xs)
+                                          mapMaybe fromVal xs)
```
tarleb: Indenting this line by 2 more spaces would make the code clearer.

PR #6125, opened 2020-02-07
Factor out a findM function

josephcsible:


jgm:
Good change.  Can you rewrite the commit message to include a note about the API change (addition of findM in exported functions in Shared)?  That way I'll be sure to include it in the changelog when doing the next releaes.

josephcsible:
Done.

PR #6140, opened 2020-02-13
Add highlight directive to the rST reader

flupe:
This PR adds the [highlight directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-highlight) from Sphinx rST to the pandoc rST reader.
This enables one to declare a default language used for highlighting in literal blocks.

```rst
.. highlight:: haskell

From now on code will be highlighted as Haskell code::

  data Tree = Leaf | Node Tree Tree

Another exemple to show how terse using literal blocks becomes::

  main :: IO ()
  main = putStrLn "Hello World!"

We can change the language at any point in the document.

.. highlight:: python

So this block contains python code::

  def main():
    print("OK")

An we can still disable having a default language:

.. highlight::

::
  This literal block will not be highlighted

```

As explained on [pandoc-discuss], this directive is not in the docutils rST reference, however Sphinx is arguably one of the most popular target for rST, and this directive has proven to be very useful.


jgm:
Looks good to me.  I'll merge if the CI tests don't reveal any problems.

PR #6141, opened 2020-02-13
A bit more cleanup

josephcsible:


src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -1405,8 +1405,8 @@ treatAsInline = Set.fromList
 label :: PandocMonad m => LP m ()
 label = do
   controlSeq "label"
-  t <- untokenize <$> braced
-  updateState $ \st -> st{ sLastLabel = Just t }
+  t <- braced
+  updateState $ \st -> st{ sLastLabel = Just $ untokenize t }
```
jgm: I don't have a problem with this, but I'm just curious whether there's a non-aesthetic reason for removing the fmap here?

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -2070,12 +2070,11 @@ addImageCaption :: PandocMonad m => Blocks -> LP m Blocks
 addImageCaption = walkM go
   where go (Image attr@(_, cls, kvs) alt (src,tit))
             | not ("fig:" `T.isPrefixOf` tit) = do
-          mbcapt <- sCaption <$> getState
-          mblabel <- sLastLabel <$> getState
-          let (alt', tit') = case mbcapt of
+          mbs <- getState
+          let (alt', tit') = case sCaption mbs of
```
jgm: mbs seems an odd namefor this.  Why not st ?
And is there any advantage to doing it this way?  I would be surprised if there's a measurable performance difference.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -2395,9 +2394,9 @@ simpTable envname hasWidthParameter = try $ do
 addTableCaption :: PandocMonad m => Blocks -> LP m Blocks
 addTableCaption = walkM go
   where go (Table c als ws hs rs) = do
-          mbcapt <- sCaption <$> getState
-          mblabel <- sLastLabel <$> getState
-          capt <- case (mbcapt, mblabel) of
+          mbs <- getState
+          let mblabel = sLastLabel mbs
```
jgm: See above on mbs.

src/Text/Pandoc/Readers/LaTeX.hs
josephcsible: Depending on the optimizer, it may be slightly faster (no promises, but it definitely won't be slower), for the same reason that `map (x . y)` is better than `map x . map y`.

src/Text/Pandoc/Readers/LaTeX.hs
josephcsible: I picked `mbs` because the style here looked like `mb` followed by an abbreviation of what it was, so I tried to follow it with `s` for "state". I'm fine with `st` instead if you want. As for speed, it will be slightly faster, since you're basically replacing `m >>= return . f >>= g` with `m >>= g . f`. It's definitely not a big difference, though. My main reason for the change was to reduce cognitive load.

jgm: I would have thought the compiler would have optimized this difference away, but maybe not.
Let's use `st` instead of `mbs` - that's usually what I use for the state variable.

josephcsible: I don't think the compiler does optimizations that require that instances be lawful to be correct (though I could be wrong, maybe if enough stuff gets specialized). Anyway, I just changed it to `st`.

jgm:
OK, thanks!

PR #6146, opened 2020-02-17
Remove redundant `otherwise` guard in inlineToAsciiDoc

RyanGlScott:
The `RawInline` case in `inlineToAsciiDoc` currenty looks like this:

https://github.com/jgm/pandoc/blob/0808c2a03f0dec1cd442e5e3cddfb70419724874/src/Text/Pandoc/Writers/AsciiDoc.hs#L490-L495

Notice how there are there are two overlapping `otherwise` guards. The second `otherwise` guard is completely unreachable, so this patch removes it.


jgm:
Thanks!

PR #6148, opened 2020-02-18
readability improvements to  default.latex:

harriott:
blank lines and a couple of comments added to make clauses clearer
     spurious blank and empty comment lines removed

   some indentation, following existing style
     spurious indentations removed

   one closing parentheses position corrected

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -1,18 +1,23 @@
+
 % Options for packages loaded elsewhere
 \PassOptionsToPackage{unicode$for(hyperrefoptions)$,$hyperrefoptions$$endfor$}{hyperref}
 \PassOptionsToPackage{hyphens}{url}
+
 $if(colorlinks)$
 \PassOptionsToPackage{dvipsnames,svgnames*,x11names*}{xcolor}
 $endif$
+
+% script direction
 $if(dir)$
 $if(latex-dir-rtl)$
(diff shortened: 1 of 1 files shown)


jgm:
I'll pass on these. I don't want to introduce spurious indentation and blank lines to the generated latex.

jgm:
Note that adding all these blank lines means that in many cases you'll get sequences of >1 blank in the output, because the conditionals between them don't always render to anything. That's definitely undesirable.

In general, I'd say the best approach is to look at the standalone latex output first.  If it's not optimal, we could consider modifying the template.  But we've accepted the cost of making the template less readable in order to generate better latex output.

harriott:
Ah, now I see what you mean.
As my priority had been to make sense of the options in `default.latex`, I'd not much bothered to look into the `--verbose` output.
It would be nice if both `default.latex` and the output from `--verbose` were clearly readable. So my fix is now not to add blank lines into my copy of `default.latex`, for the reason you have given, but instead explanatory comments, so that what I then see for unused clauses in the `--verbose` output is like this:

    %
    % links-as-notes
    %
    % prevent overfull lines
    %

\- not to everyone's liking, I suppose, but a helpful compromise for me.

jgm:
Note that you can add comments in the template like this:
```
$-- a comment which will not appear in the generated tex
```

PR #6156, opened 2020-02-26
Add an option to disable certificate validation

### LINKED ISSUE #5636: "--no-check-certificate" option, or equivalent
It's sometimes useful to fetch content from a TLS-enabled website, despite the fact it has an invalid certificate. When using a standalone HTTP client like `curl` or `wget`, there are options for disabling the certificate check (using `--insecure` and `--no-check-certificate`, respectively), but Pandoc has no such option.

I'd suggest it would be a useful enhancement would be to add such an option. Besides the option-parsing code, this would require only a couple of lines additional code in src/Text/Pandoc/Class.hs: instead of `tlsManagerSettings` being used as the settings for `Network.HTTP.Client.TLS`, when the option was enabled, code like:

```
mkManagerSettings (def  { settingDisableCertificateValidation = True }) Nothing
```
would be used instead.

Alternatively, if that's considered out of scope, it might be useful to improve the error-reporting when a TLS-related exception is thrown, suggesting to the user that they use `curl` or `wget` instead. Currently, understanding the exceptions requires some knowledge of how TLS works -- 

```
$ pandoc -f html https://self-signed.badssl.com/
Could not fetch https://self-signed.badssl.com/
HttpExceptionRequest Request {
  host                 = "self-signed.badssl.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
  path                 = "/"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (InternalException (HandshakeFailed (Error_Protocol ("certificate rejected: [SelfSigned]",True,CertificateUnknown))))
```
which users who simply want to use Pandoc to convert website content might not be familiar with.

micedre:
Hello,

This adds a `--no-check-certificate` option to the command line which instructs pandoc to disable certificate verification when fetching resources (like images) from https url with non trusted certificate (self signed for exemple).

We had to add a dependency to connection, not sure if we could have done otherwise.

It fixes #5636

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -426,6 +426,7 @@ library
                  doctemplates >= 0.8 && < 0.9,
                  network-uri >= 2.6 && < 2.7,
                  network >= 2.6,
+                 connection >= 0.3.1,
                  http-client >= 0.4.30 && < 0.7,
                  http-client-tls >= 0.2.4 && < 0.4,
                  http-types >= 0.8 && < 0.13,
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -903,6 +903,12 @@ header when requesting a document from a URL:
     downloaded). If you're behind a proxy, you also need to set
     the environment variable `http_proxy` to `http://...`.
 
+`--no-check-certificate
+
+:   Disable the certificate verification to allow access to 
+    unsecure HTTP resources (for example when the certificate
+    is no longer valid or self signed).
+
 ## Options affecting specific writers {.options}
 
 `--self-contained`
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -270,6 +270,8 @@ convertWithOpts opts = do
 
     mapM_ (uncurry setRequestHeader) (optRequestHeaders opts)
 
+    setNoCheckCertificate (optNoCheckCertificate opts)
+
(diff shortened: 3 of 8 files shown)


pandoc.cabal
```diff
@@ -426,6 +426,7 @@ library
                  doctemplates >= 0.8 && < 0.9,
                  network-uri >= 2.6 && < 2.7,
                  network >= 2.6,
+                 connection >= 0.3.1,
                  http-client >= 0.4.30 && < 0.7,
                  http-client-tls >= 0.2.4 && < 0.4,
```
jgm, line 431: Is it now possible to remove the dependency on `http-client-tls`, if we're using connection now?
At first I was not too happy about bringing in a new dependency, but since `http-client-tls` depends on `connection` this would be a net reduction!

jgm: I also wonder whether `connection` would allow simplifying the code for `openURL`?

micedre: I think we still need http-client-tls to define the TLSSettings in Http client manager. 
And I don't see how to improve the existing code with the connection package (which is very low level , I think). But I am really new to Haskell so I may be missing something.

jgm:
OK, thanks a lot for the PR!

Issue #4464, opened 2018-03-18
`stack ghci` fails


oltolm:
`stack ghci` fails with the following error message. `intero` does not work either because of the same problem.

```
pandoc-2.1.3: configure (lib + exe)
Configuring pandoc-2.1.3...
pandoc-2.1.3: initial-build-steps (lib + exe)
The following GHC options are incompatible with GHCi and have not been passed to it: -threaded
Configuring GHCi with the following packages: pandoc
Using main module: 1. Package `pandoc' component exe:pandoc with main-is file: C:\src\pandoc\pandoc.hs
GHCi, version 8.2.2: http://www.haskell.org/ghc/  :? for help

<interactive>:1:1: error:
    attempting to use module `main:Prelude' (C:\src\pandoc\prelude\Prelude.hs) which is not loaded
```

jgm:
This is because of the custom Prelude we're now using for compatibility across base versions.
The problem is a thorny one:
https://ghc.haskell.org/trac/ghc/ticket/10920
https://www.reddit.com/r/haskell/comments/3ryf2p/how_can_you_use_a_custom_prelude_with_ghci/

jgm:
The most recent commit should allow you to use `stack ghci` or `cabal repl` as long as you use ghc 8.4.1.

jgm:
Better fix coming soon.

PR #6187, opened 2020-03-14
Use implicit Prelude

tarleb:
The previous behavior was introduced as a fix for #4464. It seems that
this change alone did not fix the issue, and `stack ghci` and `cabal
repl` only work with GHC 8.4.1 or newer, as no custom Prelude is loaded
for these versions. Given this, it seems cleaner to revert to the
implicit Prelude.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -377,7 +377,8 @@ flag trypandoc
   Default:       False
 
 library
-  build-depends: base >= 4.9 && < 5,
+  build-depends: base-noprelude >= 4.9 && < 5,
+                 base-compat >= 0.9,
                  syb >= 0.1 && < 0.8,
                  containers >= 0.4.2.1 && < 0.7,
                  unordered-containers >= 0.2 && < 0.3,
@@ -440,10 +441,6 @@ library
      build-depends: basement >= 0.0.10,
                     foundation >= 0.0.23
                     -- basement 0.0.9 won't build on 32-bit windows.
-  if impl(ghc < 8.4)
-     hs-source-dirs: prelude
-     other-modules:  Prelude
-     build-depends:  base-compat >= 0.9
   if os(windows)
     cpp-options:      -D_WINDOWS
   else
@@ -471,7 +468,8 @@ library
 
   default-language: Haskell2010
   other-extensions: NoImplicitPrelude
-  hs-source-dirs:  src
+  hs-source-dirs:  src,
+                   prelude
 
   exposed-modules: Text.Pandoc,
                    Text.Pandoc.App,
@@ -565,7 +563,8 @@ library
                    Text.Pandoc.ImageSize,
                    Text.Pandoc.BCP47,
                    Text.Pandoc.Class
-  other-modules:   Text.Pandoc.App.CommandLineOptions,
+  other-modules:   Prelude,
+                   Text.Pandoc.App.CommandLineOptions,
                    Text.Pandoc.App.FormatHeuristics,
                    Text.Pandoc.App.Opt,
                    Text.Pandoc.App.OutputSettings,
@@ -640,19 +639,19 @@ library
   buildable:       True
 
 executable pandoc
-  build-depends: pandoc, base >= 4.8 && < 5
-  if impl(ghc < 8.4)
-     hs-source-dirs: prelude
-     other-modules:  Prelude
-     build-depends:  base-compat >= 0.9
+  build-depends:   pandoc,
+                   base-noprelude >= 4.8 && < 5,
+                   base-compat >= 0.9
   if flag(static)
     ld-options:     -static
   default-language: Haskell2010
   other-extensions: NoImplicitPrelude
   hs-source-dirs:  .
+                   prelude
   main-is:         pandoc.hs
   buildable:       True
-  other-modules:   Paths_pandoc
+  other-modules:   Paths_pandoc,
+                   Prelude
   ghc-options:     -rtsopts -with-rtsopts=-K16m -threaded
                    -Wall -fno-warn-unused-do-bind
                    -Wincomplete-record-updates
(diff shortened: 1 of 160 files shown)


tarleb:
Another solution to #4464, which I believe should work:

1. Define a separate package `pandoc-prelude` (or choose an existing custom Prelude).
2. Use package `base-noprelude` instead of `base`.
3. Re-export the custom prelude from `prelude/Prelude.hs`.

This worked ok for me when testing in a different package.

jgm:
Just so I understand:  does the change in this PR still have us using the custom prelude, only without an explicit import?  IF so, that seems fine to me.

jgm:
Also, could you clarify what the issue is?
`stack ghci --flag 'pandoc:embed_data_files'` seems to work fine for me already.

tarleb:
> Also, could you clarify what the issue is?
> stack ghci --flag 'pandoc:embed_data_files' seems to work fine for me already.

This currently works because we use the custom prelude *only* when compiling with older GHC (changed in dd53545154c5149dc720cc34c9990b92aec78c1a):

```
  if impl(ghc < 8.4)
     hs-source-dirs: prelude
     other-modules:  Prelude
     build-depends:  base-compat >= 0.9
```

The [cause (GHC ticket)](https://ghc.haskell.org/trac/ghc/ticket/10920) for the ghci failure still persists, and if we change the above to `impl(ghc < 8.10)`, then `stack ghci` still fails.

The use of `{-# LANGUAGE NoImplicitPrelude #-}`, added in 7e389cb3dbdc11126b9bdb6a7741a65e5a94a43e, was suggested as a solution in [a reddit thread](https://www.reddit.com/r/haskell/comments/3ryf2p/).  It doesn't seem to have the intended effect.

I now added another commit, which ensures that we *always* use the custom prelude, even with newer GHC, and with which `ghci` still works.

jgm:
ok, great, thanks!

jgm:
Now we're running into a problem, but I'm not sure I understand it.  Do you?
https://github.com/jgm/pandoc/runs/580246898?check_suite_focus=true
```
cabal: Could not resolve dependencies:
15
[__0] trying: pandoc-2.9.2.1 (user goal)
16
[__1] trying: zlib-0.6.2.1 (dependency of pandoc)
17
[__2] trying: base-4.14.0.0/installed-4.14.0.0 (dependency of zlib)
18
[__3] next goal: base-noprelude (dependency of pandoc)
19
[__3] rejecting: base-noprelude-4.13.0.0 (conflict:
20
base==4.14.0.0/installed-4.14.0.0, base-noprelude => base==4.13.0.0)
21
[__3] skipping: base-noprelude-4.12.0.0, base-noprelude-4.11.1.0,
22
base-noprelude-4.11.0.0, base-noprelude-4.10.1.0, base-noprelude-4.10.0.0,
23
base-noprelude-4.9.1.0, base-noprelude-4.9.0.0, base-noprelude-4.8.2.0,
24
base-noprelude-4.8.1.0, base-noprelude-4.8.0.0, base-noprelude-4.7.0.2,
25
base-noprelude-4.7.0.1, base-noprelude-4.7.0.0, base-noprelude-4.6.0.1,
26
base-noprelude-4.6.0.0, base-noprelude-4.5.1.0, base-noprelude-4.5.0.0,
27
base-noprelude-4.3.1.0, base-noprelude-4.3.0.0 (has the same characteristics
28
that caused the previous version to fail: excludes 'base' version 4.14.0.0)
```
I didn't know there WAS a base 4.14; it doesn't show up on Hackage.

tarleb:
I'm confused as well. According to the [GHC version history page](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history), base 4.14 is part of GHC 8.10, but the nightly build should use 8.6.5, unless I'm mistaken.

It seems that not even base-4.13 has been pushed to hackage ([issue](https://github.com/haskell-infra/hackage-trustees/issues/240)). Not sure what to make of all this, but I'll look into it some more.

jgm:
OK, what was happening was this: the GitHub runners actually have ghc 8.10.1 pre-installed and it was in the path ahead of 8.6.5.  I've fixed this, but the larger issue is that pandoc won't compile with ghc 8.10.1 due to this new base-noprelude dependency.  I've submitted an issue here https://github.com/haskell-hvr/base-noprelude/issues/11 but if this isn't updated soon we may need to find a different solution that doesn't use this package.  Can you think of one?

tarleb:
Huh. That's really unfortunate. I guess we'll have to revert 0e48a02bdd4fa3c605b79233ef2018859012c7f4.

jgm:
Can you remind me what the cost of reverting that would be?  Is there an alternative solution to the problem it solves?

tarleb:
The problem was with the custom prelude getting in the way of ghci. It used to be that `stack ghci` would fail for pandoc < 8.4, apparently due to the way that the Prelude module is loaded. Using `base-noprelude` makes sure that there is no other prelude but the one defined in `prelude/Prelude.hs`, preventing ghci from getting confused.

I also like the consistency of always using a custom Prelude, regardless of which GHC version is used.

One alternative approach could be to get rid of the custom prelude. The custom prelude could be made redundant by replacing `base-compat` with `base-compat-batteries`, enabling the `NoImplicitPrelude` extension in the cabal file, and adding `import Prelude (); import Prelude.Compat` in all modules. I think I'd prefer the conditional use of a custom Prelude for older GHCs, just as it was before.

jgm:
Sounds good.  Can you create a PR to revert that?  Unfortunately I squashed your commits when I merged your PR, so I can't revert that single commit, but you probably still have it in a branch?

jgm:
Never mind, @tarleb, I figured out how to download the raw diff from the commit, so I will revert it myself.

PR #6188, opened 2020-03-15
Improvements to trypandoc

Chocobo1:
* Use "strict mode" for javascript code
* Fetch resources over https
  Otherwise it will depend on the protocol of the webpage served which is a bad idea.
* Remove unnecessary attributes
  Those attributes are not required in HTML5.


jgm:
look good, thanks!

PR #6189, opened 2020-03-16
trypandoc: Add checkbox for standalone option

Chocobo1:
I think it is really useful to generate the full document instead of just a fragment.

screenshot: screen


jgm:
Good addition!

PR #6191, opened 2020-03-16
update MANUAL.txt - Multi-line block quote

jaybe-jekyll:
Multi-line block quotes with pandoc markdown requires escaping with backslashes at end of lines to produced the expected/documented behavior.

This does not work per the current documentation:

```markdown
> This is a block quote. This
paragraph has two lines.
```

Results in: [^1]

```This is a block quote. This paragraph has two lines.```

[^1]: Refers to *pandoc markdown* and not *gfm* which behaves as described.

This will/does work as expected for *pandoc markdown*:

```markdown
> This is a block quote. This\
paragraph has three lines.\
The back backslash escapes are required.
```

Results in: [^2]

```
This is a block quote. This
paragraph has three lines.
The backslash escape is required.
```

[^2]: Refers to *pandoc markdown*.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3096,8 +3096,9 @@ it should not be indented more than three spaces.)
 A "lazy" form, which requires the `>` character only on the first
 line of each block, is also allowed:
 
-    > This is a block quote. This
-    paragraph has two lines.
+    > This is a block quote. This\
+    paragraph has three lines.\
+    The back backslash escape is required.
 
     > 1. This is a list inside a block quote.
     2. Second item.


jgm:
Pandoc's behavior is correct.  Text flows across soft line breaks in block quotes, just like in regular paragraphs.  I think the construct you want is not a block quote, but a *line block* (see the manual).  Note that you can also turn on the `hard_line_breaks` extension if you want all newline characters to cause hard breaks.

jaybe-jekyll:
Ok; makes sense. Thanks for the explanation. I think the wording is somewhat confusing and misleads. But with the explanation becomes clearer.

So this behavior is correct for *pandoc*, however GFM *does* behave as I've described. Interesting.

jgm:
> GFM does behave as I've described. Interesting.

IN comments, yes, but not in READMEs and longer docs. gfm has two modes, one of which makes newlines create hard breaks.

PR #6201, opened 2020-03-20
Make MANUAL more explicit about NBSP handling by all_symbols_escapable

### LINKED ISSUE #6154: Non-breaking spaces in HTML ignored
The user manual states:

> A backslash-escaped space is parsed as a nonbreaking space. It will appear in TeX output as ~ and in HTML and XML as \&#160; or \&nbsp;.

While it works for TeX, it doesn't seem to work for HTML in my case.

```bash
$ echo "jo\ what " | pandoc -t html
<p>jo what</p>

$ echo "jo\ what " | pandoc -t latex
jo~what

$ echo "jo\ what " | pandoc -t json
{"blocks":[{"t":"Para","c":[{"t":"Str","c":"jo what"}]}],"pandoc-api-version":[1,20],"meta":{}}
```

pandoc: 2.9.2

fabschurt:
Fixes #6154.

I’m not sure if I should update [man/pandoc.1](https://github.com/jgm/pandoc/blob/master/man/pandoc.1) too? The syntax is quite cryptic, and it looks like it’s actually auto-generated (with Pandoc itself maybe?). If that’s the case, I could auto-generate it (and include the result in this PR), but I’m not sure what command with what set of options I should run to achieve that :man_shrugging:.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -4019,9 +4019,12 @@ which allows only the following characters to be backslash-escaped:
 (However, if the `markdown_strict` format is used, the standard Markdown rule
 will be used.)
 
-A backslash-escaped space is parsed as a nonbreaking space.  It will
-appear in TeX output as `~` and in HTML and XML as `\&#160;` or
-`\&nbsp;`.
+A backslash-escaped space is parsed as a nonbreaking space. In TeX output,
+it will appear as `~`. In HTML and XML output, it will appear as a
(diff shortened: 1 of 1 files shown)


tarleb:
Thanks for this!

Not updating man/pandoc.1 is the correct thing to do. The file will be
updated as part of the release process. See file `RELEASE-CHECKLIST`.

PR #6203, opened 2020-03-22
Finer grained imports of Text.Pandoc.Class submodules

tarleb:
This should speed-up recompilation after changes in `Text.Pandoc.Class`,
as the number of modules affected by a change will be smaller in
general. It also offers faster insights into the parts of `T.P.Class`
used within a module.


jgm:
Looks good!

PR #6209, opened 2020-03-25
More cleanup

josephcsible:


test/Tests/Readers/Docx.hs
```diff
@@ -90,9 +90,7 @@ testForWarningsWithOpts opts name docxFile expected =
 -- testForWarnings = testForWarningsWithOpts defopts
 
 getMedia :: FilePath -> FilePath -> IO (Maybe B.ByteString)
-getMedia archivePath mediaPath = do
-  zf <- toArchive <$> B.readFile archivePath
-  return $ fromEntry <$> findEntryByPath ("word/" ++ mediaPath) zf
+getMedia archivePath mediaPath = fmap fromEntry . findEntryByPath ("word/" ++ mediaPath) . toArchive <$> B.readFile archivePath
```
jgm: This should be wrapped to stay within 80 columns, thanks!

josephcsible: Done

PR #6210, opened 2020-03-25
trypandoc: Add standalone option to the command text

Chocobo1:
* trypandoc: Add standalone option to the command text
  I somehow missed it in PR #6189.
* trypandoc: Update 3rd party libraries
  Those links are taken as-is from the library website.


jgm:
Thanks!

PR #6211, opened 2020-03-25
API change: create PandocLua type, use PandocError for exceptions

tarleb:
The PandocError type is used throughout the Lua subsystem, all Lua
functions throw an exception of this type if an error occurs. The
`LuaException` type is removed and no longer exported from
`Text.Pandoc.Lua`. In its place, a new constructor `PandocLuaError` is
added to PandocError.


tarleb:
Required ito make a future `PandocLua` type an instance of `PandocMonad`.

jgm:
Will require a bump to 2.10 I guess.  Holding off merging for a short time in case we need to do fixes on 2.9.2.1.

tarleb:
I won't mind holding this off till shortly before the next release. Changes to the Lua system don't seem to be prone to merge conflicts, so we can just collect everything in one big PR. (I just like to collect feedback early on).

tarleb:
I've added more commits which show the bigger picture. This adds a new type `PandocLua` which is an instance of `PandocMonad`. This means we can use `PandocLua` exactly like `PandocIO`. The next step would be to modularize and generalize `Text.Pandoc.App` in a way which allows us to expose the functions to Lua.

tarleb:
I have rebased to master and removed non-essential commits. Ready to merge on my part.

jgm:
looks good to me, merged!

PR #6224, opened 2020-03-29
Better tables

despresc:
The start of implementing [the changes to Table](https://github.com/jgm/pandoc-types/pull/66).

Most readers use the old builders in `pandoc-types`, and the rest that deal with tables use a similar procedure: a non-empty caption is wrapped in a `Para` (should this be `Plain` instead?) and simple row padding and truncation is used to keep the table rectangular.

The writers use a new `toLegacyTable` in `Text.Pandoc.Shared`. At the moment it does the easiest thing, which is to forget about cell dimensions then pad or truncate the resulting table like `Builder.table` does. This isn't a problem for now, since none of the readers use the new `Table` functionality at all. It might instead be better to cut cells up so that cells of dimension `(h, w)` become `h * w` cells of dimension `(1, 1)`, with the content in the upper-left cell and the rest empty. I can make that change, if desired, when I change the signature of `Builder.table`, since the code involved is fairly similar.

Note that three Lua tests fail, the ones that test that tables round-trip correctly. I modified `Lua.Marshaling.AST` so that it uses `toLegacyTable`, which is of course lossy. Properly supporting the new tables requires some changes to the Lua code, and I am not too familiar with Lua, so I have left that alone. I can take a stab at changing the Lua code, but it might be easier if its current maintainer adapts it instead.

I had to change the native writer to fix the other tests. A sample of its output is this:
```haskell
Table ("",[],[]) (Caption Nothing
 []) [(AlignDefault,Nothing),(AlignDefault,Nothing)] 0
 [Row ("",[],[])
  [Cell ("",[],[]) Nothing 1 1 
   [Plain [Str ""]]
  ,Cell ("",[],[]) Nothing 1 1 
   [Plain [Str ""]]]]
 [Row ("",[],[])
  [Cell ("",[],[]) Nothing 1 1 
   [Plain [Str "Orange"]]
  ,Cell ("",[],[]) Nothing 1 1 
   [Plain [Str "Apple"]]]
 ,Row ("",[],[])
  [Cell ("",[],[]) Nothing 1 1 
   [Plain [Str "Bread"]]
  ,Cell ("",[],[]) Nothing 1 1 
   [Plain [Str "Pie"]]]
 ,Row ("",[],[])
  [Cell ("",[],[]) Nothing 1 1 
   [Plain [Strong [Str "Butter"]]]
  ,Cell ("",[],[]) Nothing 1 1 
   [Plain [Str "Ice",Space,Str "cream"]]]]
 []
```
It should be fairly simple to change.


jgm:
> a non-empty caption is wrapped in a Para (should this be Plain instead?) 

Yes, I think a `Plain` is better.

jgm:
> The writers use a new toLegacyTable in Text.Pandoc.Shared. 

If this is only used by writers, let's put it in T.P.Writers.Shared.

jgm:
>  It might instead be better to cut cells up so that cells of dimension (h, w) become h * w cells of dimension (1, 1), with the content in the upper-left cell and the rest empty. 

This would probably be better, yes, if it's not too difficult to do.

jgm:
> Note that three Lua tests fail, the ones that test that tables round-trip correctly. I modified Lua.Marshaling.AST so that it uses toLegacyTable, which is of course lossy. Properly supporting the new tables requires some changes to the Lua code, and I am not too familiar with Lua, so I have left that alone. I can take a stab at changing the Lua code, but it might be easier if its current maintainer adapts it instead.

@tarleb is the one to consult with about this.

jgm:
```
Table ("",[],[]) (Caption Nothing
 []) [(AlignDefault,Nothing),(AlignDefault,Nothing)] 0
```
Just curious why the line break in the middle of the Caption constructor?

Otherwise, this looks good.

One thing that strikes me, though, looking at this output, is how unperspicuous the Maybe and numerical values are.  `Nothing` means what?  `1` means what? We might think about creating custom types and newtypes that are clearer, like
```
data ShortCaption = ShortCaption [Inline] | NoShortCaption
newtype RowSpan = RowSpan Int
newtype ColSpan = ColSpan Int
data AlignCell = AlignCell Alignment | AlignCellDefault
```
Then we'd have
```
(Caption NoShortCaption [])
```
and
```
Cell ("",[],[]) AlignCellDefault (RowSpan 1) (ColSpan 1)
```
Alternatively, instead of RowSpan and ColSpan, we could have
```
CellDimensions Int Int
```
I'm curious what you think. (Also @tarleb.)  Of course these are things we could leave as is for now, but while we're messing with types it might make sense to think it through.

jgm:
Excellent work, by the way!  I think this is a great approach --- we can integrate the new types into pandoc without changing any reader or writer behavior, and then do that gradually.

mb21:
> One thing that strikes me, though, looking at this output, is how unperspicuous the Maybe and numerical values are. Nothing means what? 1 means what? We might think about creating custom types and newtypes that are clearer

true... I'm not sure about performance impact though. Maybe in the future we'll have a more human readable JSON format, though...?

jgm:
@mb21 `newtype`s have no performance impact.  `data` does introduce additional indirection and thus has a performance impact.  But the places where these will be used are not hot spots.

tarleb:
Exciting! I'll happily take a close look later this week.

despresc:
I had the line break there because I thought the block content of the caption might look better on its own line. It does look a little odd when the caption is empty. Maybe a conditional line break? Or it could be removed entirely.

The `toLegacyTable` seems to be used only in `Writers/` and `Lua/Marshaling/AST.hs`. The latter use should go away once the Lua code is adapted to new type, so `toLegacyTable` can be moved to `Writers.Shared` as long as the temporary dependency in `Lua.Marshaling.AST` is okay.

Having `ShortCaption` and `AlignCell` and the newtypes would make the writer output clearer, certainly. The `RowSpan` and `ColSpan` types could also have a `Num` instance, unless writing the constructors explicitly in the source is important. If that were our approach, then the `Maybe Double` in `ColSpec` could be made into its own type (`data ColWidth = ColWidth Double | ColWidthDefault`, or something like that).

despresc:
Perhaps the `Caption` should also have an `Attr`? At least for filters and HTML.

despresc:
Having the type be `ShortCaption` seems to imply that the short caption will always be available for a `Caption`. I think that either the `NoShortCaption` should be renamed if that is the case (to `ShortCaptionDefault`, or `UnspecifiedShortCaption`, or something better), or it should be `Maybe ShortCaption`, with `type ShortCaption = [Inline]`.

The argument for keeping the various types as `Maybe` is that that is easier to work with, since functions like `maybe` and `fromMaybe` would work on all of them. The `ShortCaption` discussion above suggests that this could be a little more flexible in design, too; the `Nothing :: Maybe a` would simply mean "unspecified `a`", and the users of the type could decide how to deal with such a thing.

jgm:
> The argument for keeping the various types as Maybe is that that is easier to work with, since functions like maybe and fromMaybe would work on all of them

This is true, and the readability of Native output isn't a particularly strong desideratum.

On the other hand, some of the newtypes might be worth adding; e.g. RowSpan and ColSpan -- this would help the type checker prevent accidentally mixing these up, for example.

As for `Maybe Alignment` - `Alignment` already has `AlignDefault` as one of its constructors; why not just use this instead of `Nothing`?

I'm fine with leaving the short caption as a Maybe.

I like the idea of replacing the `Maybe Double` with `data ColWidth = ColWidth Double | ColWidthDefault`.  But I could be persuaded otherwise.

despresc:
Having the various new dimensions be newtypes is good. I don't think that there is too much difference having the `Maybe` as opposed to new data types, so I can change things so that the short caption is `Maybe`, the cell alignment is just `Alignment`, and the column width is `ColWidth = ColWidth Double | ColWidthDefault`. That's reasonable.

despresc:
A sample of the new `Writers.Native` output:

```haskell
Table ("",[],[]) (Caption Nothing
[Plain [Str "Food",Space,Str "complements"]])
[(AlignDefault,ColWidthDefault)
,(AlignDefault,ColWidthDefault)]
(TableHead ("",[],[])
[Row ("",[],[])
 [Cell ("",[],[]) AlignDefault (RowSpan {getRowSpan = 1}) (ColSpan {getColSpan = 1})
  [Para [Str "Orange"]]
 ,Cell ("",[],[]) AlignDefault (RowSpan {getRowSpan = 1}) (ColSpan {getColSpan = 1})
  [Para [Str "Apple"]]]])
[(TableBody ("",[],[]) (RowHeadColumns {getRowHeadColumns = 0})
 []
 [Row ("",[],[])
  [Cell ("",[],[]) AlignDefault (RowSpan {getRowSpan = 1}) (ColSpan {getColSpan = 1})
   [Para [Str "Bread"]]
  ,Cell ("",[],[]) AlignDefault (RowSpan {getRowSpan = 1}) (ColSpan {getColSpan = 1})
   [Para [Str "Pie"]]]
 ,Row ("",[],[])
  [Cell ("",[],[]) AlignDefault (RowSpan {getRowSpan = 1}) (ColSpan {getColSpan = 1})
   [Para [Str "Butter"]]
  ,Cell ("",[],[]) AlignDefault (RowSpan {getRowSpan = 1}) (ColSpan {getColSpan = 1})
   [Para [Str "Ice",Space,Str "cream"]]]])]
(TableFoot ("",[],[])
[])
```

jgm:
```
(RowSpan {getRowSpan = 1}) (ColSpan {getColSpan = 1})
```
The output is much more compact without these accessors.  Maybe we should leave them off the `newtype`?  (that is, `newtype Rowspan = RowSpan Int`)

Seems like we can just use `coerce` to get the value for both of them, without a special constructor.  (or use pattern matching)

Note that in general we don't have these accessors defined in pandoc-types.  Partly this is just to keep the native output from being so verbose.

despresc:
Sure, I will change that.

despresc:
When a table of width `w` with no header is encountered, quite a few of the readers (`html`, `textile`, `txt2tags`, possible a few others) currently output `replicate w emptyCell` as the header, and not `[]`. This is reflected in the current tests. Is this desired behaviour? If not, I'll fix the offending tests as I adapt `pandoc` to the new `Builder.table`.

jgm:
re `replicate w emptyCell`: I'm pretty sure there WAS a reason for this, but I've forgotten it.
On the other hand, it doesn't make a whole lot of sense, and we could change it, but this might require some changes in the writers.

despresc:
As it turns out, only the XWiki writer seemed to depend on the header row being non-empty. Many of the others that might have cared pad the header row with empty cells, then do something to the effect of `if all null headerRow then outputNoHead ...`, so they were fine. I believe all the readers now don't emit an empty header row on a head that isn't present.

The new `toLegacyTable` (that cuts up cells) normalizes the input table components before emitting the legacy table components. I thought that was safer.

despresc:
I think these are all the changes, unless there's something more that you would like me to do. The `Builder.table` adaptation was mostly shallow, only touching code near uses of `table`. Later refactoring can change that.

despresc:
I forgot about `Lua.Marshaling.AST` pretending to lua, via `toLegacyTable`, that the table constructor hasn't changed. That still remains.

tarleb:
I'll have a look at Lua marshaling soon.

tarleb:
I made a PR to your "better-tables" branch, @despresc. It is rather bare-bones, but works well enough to let the tests pass. I guess we can fine-tune the Lua structure later.

despresc:
Thanks @tarleb. That should be everything.

jgm:
Excellent!  One thing I noticed is the need for explicit qualifications in `Pandoc.Row`.  Why is that needed?  Is `Row` defined somewhere else?  I wonder whether using more table-specific types like `TableRow`, `TableColumn`, `TableCell` would be a good idea?

jgm:
Oh, maybe `Row` is defined in tagsoup?  On second thought, I'm not inclined to change the type names; `Row` is pretty unambiguous in a pandoc context.

despresc:
Yes, it was. I must have thought that `Readers.Markdown` was using the tag soup `Row`, since it's easier just to hide it. The only internal type that conflicts with `Row` is a `Row` defined in the `Docx` reader, I believe.

jgm:
This all looks great to me -- we can fine tune as we go, but I'm merging this now!

Next steps:

- support row/colspans in readers and writers where that makes sense
- provide a mechanism for "downshifting" tables with row/colspans to tables without them, for writers where these can't be supported (e.g. markdown) (@despresc is there already such a function in the changes you've added?)

Longer term:

- think about how to support row/colspans in markdown table formats

despresc:
Excellent news.

The `Writers.Shared.toLegacyTable` function does that, except that it aims for strict compatibility with the old signature of `Table`, so it also does things like convert the `Caption` to an `[Inline]`.

despresc:
A function like `toSimpleTable :: TableHead -> [TableBody] -> TableFoot -> ([[Block]], [[[Block]]])` could be taken out of it, for instance.

jgm:
I think `toLegacyTable` is pretty much what I had in mind, thanks.

mb21:
> think about how to support row/colspans in markdown table formats

I suppose it would be a good start if that is at least one markdown table style that can support that: grid tables seem the obvious candidate.

despresc:
There are also simple tables. It would be nice to write the booktabs version of the traditional LaTeX table:

```
        Item
--------------------------  ---------
Animal    Description           Price
--------- ----------------  ---------
Gnat      per-gram              13.65
          each                   0.01
Gnu       stuffed               92.50
Emu       stuffed               33.33
Armadillo frozen                 8.99
```

jgm:
Would it make sense to create separate issues for reader and writer support for some of the most important formats, to help people start working on these things? I would think the highest-priority formats to which to add row/colspan support are

- LaTeX
- Docx
- HTML

There could also be an issue for Markdown, which would require syntax decisions (but I think there already is an issue with extensive discussion on this).

Other formats that I think allow row/colspans, to some degree:

- RST
- Org
- ODT
- DocBook
- JATS
- Ms
- Man

and no doubt others.

despresc:
I think that's a good idea. It would at least alert people that the table handling can now be improved, and indicate what still remains to be done so that they can contribute. Maybe the original issue and the "list of projects" issue can be updated as well.

PR #6229, opened 2020-03-30
Clean up and simplify Text.Pandoc.Writers.Docx

josephcsible:


jgm:
Good improvements, thanks

PR #6243, opened 2020-03-31
Allow hslua 1.1

felixonmars:
Builds fine and all tests pass.


jgm:
Checking with @tarleb

tarleb:
Fine by me. HsLua 1.1 is somewhat experimental, but should be mostly backwards compatible. Moreover, we'd need it for the `PandocLua` type, so might as well start using it now.

PR #6245, opened 2020-04-01
Fix description of BulletList Lua type

lggruspe:
Change description of BulletList parameter from 'List of Blocks' to 'List of List of Blocks'.


tarleb:
Thanks!

PR #6252, opened 2020-04-06
Custom writer: support pretty-printing with DocLayout

tarleb:
This would allow users to use the full power the DocLayout library in their custom writers.

Unfortunately, this is not fully backwards compatible. We could probably hide it behind some kind of toggle.

### DIFF
--- a/src/Text/Pandoc/Writers/Custom.hs
+++ b/src/Text/Pandoc/Writers/Custom.hs
@@ -20,15 +20,16 @@ import Control.Applicative (optional)
 import Control.Arrow ((***))
 import Control.Exception
 import Control.Monad (when)
+import Control.Monad.IO.Class (MonadIO)
 import Data.List (intersperse)
 import Data.Maybe (fromMaybe)
 import qualified Data.Text as T
-import Data.Text (Text, pack)
+import Data.Text (Text)
 import HsLua as Lua hiding (Operation (Div))
 import HsLua.Aeson (peekViaJSON)
-import Text.DocLayout (render, literal)
+import qualified HsLua.Module.DocLayout as Lua.DocLayout
+import Text.DocLayout (Doc, render)
 import Text.DocTemplates (Context)
-import Control.Monad.IO.Class (MonadIO)
 import Text.Pandoc.Definition
 import Text.Pandoc.Lua (Global (..), runLua, setGlobals)
 import Text.Pandoc.Lua.Marshal.Attr (pushAttributeList)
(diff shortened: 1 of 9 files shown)


tarleb:
It's backwards-compatible now. Strings are still pushed as return values for custom writer functions unless a global variable `PANDOC_USE_DOCLAYOUT` is defined and non-nil.

src/Text/Pandoc/Writers/Custom.hs
```diff
@@ -99,31 +115,39 @@ writeCustom luaFile opts doc@(Pandoc meta _) = do
...
+docToCustom :: WriterOptions -> Pandoc -> Lua (Doc Text)
 docToCustom opts (Pandoc (Meta metamap) blocks) = do
   body <- blockListToCustom blocks
-  Lua.callFunc "Doc" body (fmap Stringify metamap) (writerVariables opts)
+  useDoc <- useDocLayout
+  let metamap' = fmap Stringify metamap
+  let variables = writerVariables opts
+  if useDoc
+    then Lua.callFunc "Doc"  body                 metamap' variables
+    else Lua.callFunc "Doc" (render Nothing body) metamap' variables
```
josephcsible, line 144:
```suggestion
  let body' = if useDoc then body else render Nothing body
  Lua.callFunc "Doc" body' metamap' variables
```

src/Text/Pandoc/Writers/Custom.hs
```diff
@@ -35,22 +35,38 @@ import Text.Pandoc.Writers.Shared
...
+  result <- Lua.toboolean Lua.stackTop
+  Lua.pop 2
+  return result
+
+pushDoc :: Doc Text -> Lua ()
+pushDoc d = do
+  useDoc  <- useDocLayout
+  if useDoc
+    then Lua.push d
+    else Lua.push (render Nothing d)
```
josephcsible, line 58:
```suggestion
  Lua.push $ if useDoc
    then d
    else render Nothing d
```

josephcsible:
Whoops, never mind, those are different types. (I assume that the only reason you wrote it the way you did was that what I suggested doesn't actually work.)

tarleb:
Indeed, that's the reason. Maybe I should add some explicit type to make it more obvious?
Thanks for taking the time to review!

jgm:
A couple issues that occurred to me in looking at this:

First, a small question:  Is it possible to eliminate the need for the global variable?  If the functions return doclayout Doc objects, then use doclayout; if they return Strings, not?

Bigger issue:  I'm getting a bit worried about the ongoing maintenance cost of shadowing large parts of the pandoc (and related) API with Lua.  In the present case: if I make further changes to doclayout, then this custom Lua code needs to be updated, and so does the documentation.  However, I'm not sure anything would automatically alert us to the need to do these things, which raises a danger of breakage as things get out of sync (especially if you ever stop being such an active contributor to the project). 

Is there a way to generate both the Haskell bridge code and the Lua API documentation automatically from doclayouts itself?

tarleb:
>  I'm getting a bit worried about the ongoing maintenance cost of shadowing large parts of the pandoc (and related) API with Lua.

I completely understand, and I agree. Maybe a separate `pandoclua` package, which just ships Lua interpreter, pandoc library, and experimental packages, would be a better choice for now. We can still merge things back if they prove useful. Right now, I'm honestly not sure if anybody but me would like to use the things I'm building. I'm closing this and instead will make proposals like #6263, meaning only for cases where the API doesn't support my experiments.

I keep thinking about ways to automate the generation of library glue code, but keep putting it off. Producing a better Lua bindings system might take me a while, but would hopefully be worth the maintenance improvements.

> Is it possible to eliminate the need for the global variable? If the functions return doclayout Doc objects, then use doclayout; if they return Strings, not?

The problem is that we can't know beforehand if a writer functions expects a string or a Doc as input. The only thing I came up with was to let the writer try and run the functions with Doc objects, then fall back on Strings if an error occurs. It would always give the expected result, but might have to redo some computations – probably not a good idea if IO is involved.

jgm:
> The problem is that we can't know beforehand if a writer functions expects a string or a Doc as input. The only thing I came up with was to let the writer try and run the functions with Doc objects, then fall back on Strings if an error occurs. It would always give the expected result, but might have to redo some computations – probably not a good idea if IO is involved.

This could just be done once, and the result cached -- the assumption being that either all the functions return doc objects or all return strings?  But perhaps the global variable solution is cleaner.

tarleb:
With the custom reader on the way and the cleaner hslua design, it might be worth discussing this again.

> This could just be done once, and the result cached -- the assumption being that either all the functions return doc objects or all return strings? But perhaps the global variable solution is cleaner.

That would make for a nicer interface, but the code to make that work might turn out a bit messy. I think I'd prefer using the global.

tarleb:
I've added you to the list of hslua-module-doclayout maintainers on Hackage and as a maintainer of the repo. This should prevent situations in which that package becomes a problem due to version constraints.

tarleb:
Closing in favor of #7899 and #7897.

PR #6257, opened 2020-04-06
Vimwiki reader: Add nested syntax highlighting

### LINKED ISSUE #6256: Vimwiki nested syntaxes aren't highlighted
Vimwiki reader doesn't highlight the code in preformatted blocks that have nested syntax. Vimwiki's nested syntax code blocks look like this:
```
{{{python
print "hello world!"
}}}
```
Currently the reader only looks for `key=value` attributes and the internal code block has always an empty class list and therefore it's never highlighted.

wladh:
Nested syntaxes are specified like this:
{{{sql
SELECT * FROM table
}}}

The preformatted code block parser has been extended to check if the
first attribute of the block is not a `key=value` pair, and in that case
it will be considered as a class.

Closes #6256.


jgm:
Thanks!

Issue #5794, opened 2019-10-06
Support <bdo> in HTML


jgm:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdo

We can parse `<bdo dir="ltr">foo</bdo>` as
```
Span ("",[],[("dir","ltr")]) [Str "foo"]
```

PR #6271, opened 2020-04-10
HTML reader: support <bdo>

TristanCacqueray:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdo

Closes: #5794

### DIFF
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -803,13 +803,17 @@ pCode = try $ do
   result <- manyTill pAny (pCloses open)
   return $ B.codeWith (mkAttr attr) $ T.unwords $ T.lines $ innerText result
 
+-- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdo
+-- Bidirectional Text Override
 pBdo :: PandocMonad m => TagParser m Inlines
 pBdo = try $ do
   TagOpen _ attr' <- lookAhead $ pSatisfy $ tagOpen (=="bdo") (const True)
   let attr = toStringAttr attr'
   contents <- pInTags "bdo" inline
-  let dir = fromMaybe "" $ lookup "dir" attr
-  return $ B.spanWith ("", [], [("dir",dir)]) contents
+  return $ case T.toLower $ fromMaybe "" $ lookup "dir" attr of
+    -- Only right-to-left direction actually matters
+    "rtl" -> B.spanWith ("", [], [("dir","rtl")]) contents
+    _     -> contents
 
 pSpan :: PandocMonad m => TagParser m Inlines
 pSpan = try $ do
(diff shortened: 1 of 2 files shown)


test/command/1592.md
```diff
@@ -47,3 +47,9 @@ pandoc -f native -t markdown
 [hi]{.smallcaps}
 ```
 
+```
+% pandoc -f html -t native
+<bdo dir="ltr">foo</bdo>
+^D
+[Plain [Span ("",[],[("dir","ltr")]) [Str "foo"]]]
```
TristanCacqueray, line 54: Note that I can't explain why this is wrapped in a `[Plain ...]` block.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -802,6 +803,14 @@ pCode = try $ do
   result <- manyTill pAny (pCloses open)
   return $ B.codeWith (mkAttr attr) $ T.unwords $ T.lines $ innerText result
 
+pBdo :: PandocMonad m => TagParser m Inlines
+pBdo = try $ do
+  TagOpen _ attr' <- lookAhead $ pSatisfy $ tagOpen (=="bdo") (const True)
+  let attr = toStringAttr attr'
+  contents <- pInTags "bdo" inline
+  let dir = fromMaybe "" $ lookup "dir" attr
+  return $ B.spanWith ("", [], [("dir",dir)]) contents
```
mb21, line 812: nit-pick... without the `"dir"` attribute, this is uselels, so maybe do a `case` on the `Maybe` returned from `lookup` ?

TristanCacqueray: Alright. Then I guess when dir is different than rtl, bdo is also useless. Next commit fixes that.

mb21: well... if the whole doc is rtl, then a ltr span is useful :-)

TristanCacqueray: Ah, good point! Ok let me not skip rtl then :)  What about validating that dir must be either rtl or ltr? should that be done in the `pBdo` ?

mb21: hm.. good question... I guess it wouldn't hurt.. although we're not doing much validation on most attributes...

test/command/1592.md
despresc: The readers all return a full `Pandoc` document, so a bare inline expression has to be wrapped in some kind of block. This ends up being `Plain` for the html reader.

TristanCacqueray: If I understand correctly I don't need to update the `pBdo` and this output is expected?

despresc: It's expected, yes. You'd have to modify `readHtml` or `writeNative` themselves not to have it.

jgm:
Great, thanks!

TristanCacqueray:
Thank you for putting the `good first issue` label!

Issue #6255, opened 2020-04-06
ignoring underscored text 


nedvedikd:
Hi,

I'm using pandoc (**version 2.9.2.1**) to convert documents to dokuwiki format.
I ran into a problem using this command:

```bash
pandoc input.html -f html -t dokuwiki -o output_dokuwiki.txt --preserve-tabs
```

**Input:**
```html
<strong>bold</strong>
<em>italic</em>
<u>underline</u>
```

**Output:**
```
**bold** //italic// underline
```

**Expected output:**
```
**bold** //italic// __underline__
```

As you can see, pandoc is clearly ignoring underscored text.

Thank you so much for your help!

jgm:
Currently pandoc parses HTML u as
```
Span ("",["underline"],[]) inlines
```
so it would be easy to modify the docuwiki writer to recognize this form and emit `__...__`.

3v0k4:
Would it make sense to introduce an additional `Underline` value constructor in [`Inline`](https://hackage.haskell.org/package/pandoc-types-1.20/docs/Text-Pandoc-Definition.html#t:Inline)?

I'm asking because I see there is a `Emph` and `Strong` but not a `Underline`. Instead `Span` is used as @jgm mentioned.

jgm:
Probably yes. https://github.com/jgm/pandoc-types/pull/53
I've resisted this for a long time, on grounds that Underline is presentational not structural.
But doing all of this via Spans is not very robust.

mb21:
> I've resisted this for a long time, on grounds that Underline is presentational not structural.
> But doing all of this via Spans is not very robust.

True, there is also the point that `<em` and `<strong` are encouraged in HTML5 but, `<u` is deprecated...

julianespinel:
Hello! I just ran the commands suggested by @nedvedikd (against master), and I got the expected result:
```
**bold** //italic// __underline__
```

Could we close this issue?

jgm:
yes, thanks

Issue #5825, opened 2019-10-15
HTML Writer: Output underline as <u> instead of <span class="underline">


hftf:
Opening an issue per https://github.com/jgm/pandoc/pull/5805#issuecomment-542306918. As of that PR, a keyboard element is now represented as a `Span` with a `kbd` class, and the HTML Writer outputs using the native `<kbd>` element.

Pandoc partially implements Underline as a `Span` with an `underline` class. Currently, the HTML Writer outputs `<span class="underline">`. Shouldn’t it output the native `<u>` element instead?

### What are some advantages of outputting native HTML elements?

* Semantics: Native HTML elements have defined and well-known semantics.
* Idempotence: Pandoc can convert more HTML to HTML in a lossless round trip.
* Simpler templates: Currently, templates must give special treatment to classes by adding rules in internal or external stylesheets (see [default HTML template](https://github.com/jgm/pandoc-templates/blob/master/default.html5#L17-L25), [EPUB stylesheet](https://github.com/jgm/pandoc/blob/5d790db5cf3ca97f901b3d64adc67da43d290ac0/data/epub.css)). Moreover, those styles are only included for standalone output. Custom templates must be updated more frequently, depending on upstream changes in Pandoc.
* Still works outside of standalone output.
* Still works when CSS is disabled.
* Interoperability: Documents with native HTML elements are more likely to be preserved in other applications. For example, consider rich text copying/pasting: Chrome copies inline styles (and even inlinifies applicable internal/external styles), while Firefox doesn’t. While both browsers copy class attributes, it doesn’t matter because the destination may not define the same styles. [Demo](http://ophir.li/misc/pandoc-underline.html)

---

Previously:

* `<kbd>`: #5796 #5805 
* `<samp>`, `<abbr>`, `<dfn>`, `<mark>`, `<var>`: #5792 #5793 #5795 #5797 #5799. Most of these elements can presumably be implemented easily by adding to the map added in #5805.
* Underline: #2270 #2264 (#4633 #5044 #5135)

jgm:
I don't see much of a downside.  I'd say let's do it, unless someone sees a problem with this?

mb21:
Well... from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/u

> This element used to be called the "Underline" element in older versions of HTML, and is still sometimes misused in this way. To underline text, you should instead apply a style that includes the CSS text-decoration property set to underline.

[...]

PR #6277, opened 2020-04-12
Support Underline

vaibhavsagar:
This PR depends on https://github.com/jgm/pandoc-types/pull/68.


vaibhavsagar:
References https://github.com/jgm/pandoc/issues/6255.

mb21:
Hey Vaibhav, how are you?

Cool that you're contributing to pandoc! Some time off in Singapore? ;-)

Just wanted to make sure you had seen my comment here: https://github.com/jgm/pandoc/issues/5825#issuecomment-544111741

So while we all agree that the `Underline` element in the pandoc AST is a good thing, I guess the discussion about how to render it in HTML is still open....?

Cheers,
Mauro

vaibhavsagar:
Hi Mauro! It's great to hear from you :smile:, you're right that I have some downtime and thought that this might be a good opportunity to contribute. It seems like the latest guidance is that `<u>` is okay? I'm happy to leave this open until that discussion is concluded.

vaibhavsagar:
I'm encountering a test failure that isn't making any sense to me:

```
    Docx
      inlines
        font formatting:                                             FAIL (0.07s)
          Non-matching xml in word/document.xml:
          + 101           <w:u w:val="single" />
          - 102           <w:u w:val="single" />
```

Any idea what is going on here?

mb21:
don't know about the test failure... can't see how it could be whitespace...? but maybe have a look yourself, along the lines of:

    pandoc -f native input.native -o output.docx
    unzip -d outdir output.docx
    diff outdir/word/document.xml testdir/word/document.xml

alerque:
Github has a useful feature for better handling of WIP pull requests called Draft PR's. You can create them as drafts when you open them or [can now convert existing ones to drafts](https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/). This does a number of useful things and I suggest converting this to a Draft until it at least you as the submitter consider it ready for merging. PR reviews can happen during both phases.

vaibhavsagar:
Thanks, I've made this a draft PR. @mb21 I tried to follow your instructions but wasn't sure what `input` and `testdir` are meant to be in this case. I converted `inline_formatting.docx` and even regenerated it, but the regenerated version caused more tests to fail (for the reader this time), which seems to indicate that the parsing logic for `.docx` files needs to be updated as well.

mb21:
hm.. I cannot reproduce? I added this to stack.yaml

```
extra-deps:
- github: vaibhavsagar/pandoc-types
  commit: 89bf303f11af19177948d9b6f24ef3e6a09a8009
```

then `stack test` gives me: `2 out of 2325 tests failed` with:

```
    Docx
      document
        allow different document.xml file as defined in _rel...:     FAIL
          test/Tests/Helpers.hs:51:
          
          [Header 1 ("test",[],[]) [Str "Test"]
          ,Para [Str "This",Space,Str "is",Space,Emph [Str "italic"],Str ",",Space,Strong [Str "bold"],Str ",",Space,Underline [Str "underlined"],Str ",",Space,Emph [Underline [Str "italic",Space,Str "underlined"]],Str ",",Space,Strong [Underline [Str "bold",Space,Str "underlined"]],Str ",",Space,Emph [Strong [Underline [Str "bold",Space,Str "italic",Space,Str "underlined"]]],Str "."]]
            [Header 1 ("test",[],[]) [Str "Test"]
          - ,Para [Str "This",Space,Str "is",Space,Emph [Str "italic"],Str ",",Space,Strong [Str "bold"],Str ",",Space,Span ("",["underline"],[]) [Str "underlined"],Str ",",Space,Emph [Span ("",["underline"],[]) [Str "italic",Space,Str "underlined"]],Str ",",Space,Strong [Span ("",["underline"],[]) [Str "bold",Space,Str "underlined"]],Str ",",Space,Emph [Strong [Span ("",["underline"],[]) [Str "bold",Space,Str "italic",Space,Str "underlined"]]],Str "."]]
          + ,Para [Str "This",Space,Str "is",Space,Emph [Str "italic"],Str ",",Space,Strong [Str "bold"],Str ",",Space,Underline [Str "underlined"],Str ",",Space,Emph [Underline [Str "italic",Space,Str "underlined"]],Str ",",Space,Strong [Underline [Str "bold",Space,Str "underlined"]],Str ",",Space,Emph [Strong [Underline [Str "bold",Space,Str "italic",Space,Str "underlined"]]],Str "."]]
      inlines
        font formatting:                                             FAIL
          test/Tests/Helpers.hs:51:
          
          [Para [Str "Regular",Space,Str "text",Space,Emph [Str "italics"],Space,Strong [Str "bold",Space,Emph [Str "bold",Space,Str "italics"]],Str "."]
          ,Para [Str "This",Space,Str "is",Space,SmallCaps [Str "Small",Space,Str "Caps"],Str ",",Space,Str "and",Space,Str "this",Space,Str "is",Space,Strikeout [Str "strikethrough"],Str "."]
          ,Para [Str "Some",Space,Str "people",Space,Str "use",Space,Underline [Str "single",Space,Str "underlines",Space,Str "for",Space],Emph [Underline [Str "emphasis"]],Str "."]
          ,Para [Str "Above",Space,Str "the",Space,Str "line",Space,Str "is",Space,Superscript [Str "superscript"],Space,Str "and",Space,Str "below",Space,Str "the",Space,Str "line",Space,Str "is",Space,Subscript [Str "subscript"],Str "."]
          ,Para [Str "A",Space,Str "line",LineBreak,Str "break."]]
            [Para [Str "Regular",Space,Str "text",Space,Emph [Str "italics"],Space,Strong [Str "bold",Space,Emph [Str "bold",Space,Str "italics"]],Str "."]
            ,Para [Str "This",Space,Str "is",Space,SmallCaps [Str "Small",Space,Str "Caps"],Str ",",Space,Str "and",Space,Str "this",Space,Str "is",Space,Strikeout [Str "strikethrough"],Str "."]
          - ,Para [Str "Some",Space,Str "people",Space,Str "use",Space,Span ("",["underline"],[]) [Str "single",Space,Str "underlines",Space,Str "for",Space,Emph [Str "emphasis"]],Str "."]
          + ,Para [Str "Some",Space,Str "people",Space,Str "use",Space,Underline [Str "single",Space,Str "underlines",Space,Str "for",Space],Emph [Underline [Str "emphasis"]],Str "."]
            ,Para [Str "Above",Space,Str "the",Space,Str "line",Space,Str "is",Space,Superscript [Str "superscript"],Space,Str "and",Space,Str "below",Space,Str "the",Space,Str "line",Space,Str "is",Space,Subscript [Str "subscript"],Str "."]
            ,Para [Str "A",Space,Str "line",LineBreak,Str "break."]]
```

vaibhavsagar:
Oops, I forgot to push my latest changes. Can you try again?

mb21:
Okay, so the failing test is test/Tests/Writers/Docx.hs -> search for `"font formatting"`

to see what's going on:

    stack exec pandoc -- -f native test/docx/inline_formatting.native -o test.docx
    unzip -d test test.docx 
    unzip -d golden test/docx/golden/inline_formatting.docx
    xmllint --format test/word/document.xml > test.xml
    xmllint --format golden/word/document.xml > golden.xml
    diff test.xml golden.xml

ha, indeed, your writer produces:

        <w:rPr>
          <w:u w:val="single"/>
          <w:i/>
        </w:rPr>

while the test says it should be:

        <w:rPr>
          <w:i/>
          <w:u w:val="single"/>
        </w:rPr>

Not sure why and which one is right, but that's why...

vaibhavsagar:
Thanks for finding that! I think my changes to `inline_formatting.native` are to blame, I modified it slightly and now I'm getting this error in the reader:

```
      inlines
        font formatting:                                             FAIL
          test/Tests/Helpers.hs:51:
          
          [Para [Str "Regular",Space,Str "text",Space,Emph [Str "italics"],Space,Strong [Str "bold",Space,Emph [Str "bold",Space,Str "italics"]],Str "."]
          ,Para [Str "This",Space,Str "is",Space,SmallCaps [Str "Small",Space,Str "Caps"],Str ",",Space,Str "and",Space,Str "this",Space,Str "is",Space,Strikeout [Str "strikethrough"],Str "."]
          ,Para [Str "Some",Space,Str "people",Space,Str "use",Space,Underline [Str "single",Space,Str "underlines",Space,Str "for",Space],Emph [Underline [Str "emphasis"]],Str "."]
          ,Para [Str "Above",Space,Str "the",Space,Str "line",Space,Str "is",Space,Superscript [Str "superscript"],Space,Str "and",Space,Str "below",Space,Str "the",Space,Str "line",Space,Str "is",Space,Subscript [Str "subscript"],Str "."]
          ,Para [Str "A",Space,Str "line",LineBreak,Str "break."]]
            [Para [Str "Regular",Space,Str "text",Space,Emph [Str "italics"],Space,Strong [Str "bold",Space,Emph [Str "bold",Space,Str "italics"]],Str "."]
            ,Para [Str "This",Space,Str "is",Space,SmallCaps [Str "Small",Space,Str "Caps"],Str ",",Space,Str "and",Space,Str "this",Space,Str "is",Space,Strikeout [Str "strikethrough"],Str "."]
          - ,Para [Str "Some",Space,Str "people",Space,Str "use",Space,Underline [Str "single",Space,Str "underlines",Space,Str "for",Space,Emph [Str "emphasis"]],Str "."]
          + ,Para [Str "Some",Space,Str "people",Space,Str "use",Space,Underline [Str "single",Space,Str "underlines",Space,Str "for",Space],Emph [Underline [Str "emphasis"]],Str "."]
            ,Para [Str "Above",Space,Str "the",Space,Str "line",Space,Str "is",Space,Superscript [Str "superscript"],Space,Str "and",Space,Str "below",Space,Str "the",Space,Str "line",Space,Str "is",Space,Subscript [Str "subscript"],Str "."]
            ,Para [Str "A",Space,Str "line",LineBreak,Str "break."]]
```

For some reason part of the document is being parsed as `Emph [Underline [Str "emphasis"]]` even though it's already contained in an `Underline` element. I think this might be a change that needs to be made to the reader, what do you think?

mb21:
well... I don't think pandoc does any kind of normalization on nestings of inline elements like strong/emph etc. but just preserves that, which is not perfect, but keeps the code simple:

    echo '_**_foo_**_' | pandoc
    <p><em><strong><em>foo</em></strong></em></p>

so guess just do the same with Underline?

vaibhavsagar:
Ah, thanks for clarifying that, I assumed that this was happening for the old `Span` that was being used.

vaibhavsagar:
I've regenerated `inline_formatting.docx` and now I'm running into an issue where `w:val="single"` isn't being output correctly, even though I think I've accounted for it in the writer:

```
  Writers
    Docx
      inlines
        font formatting:                       FAIL (0.05s)
          Non-matching xml in word/document.xml:
          +  89           <w:u w:val="single" />
          -  89           <w:u />
          +  95           <w:u w:val="single" />
          -  95           <w:u />
          + 101           <w:u w:val="single" />
          - 101           <w:u />
```

I think that if I'm able to fix this the PR should be ready to review.

vaibhavsagar:
Seems to work now!

jgm:
The CI tests are still failing...

vaibhavsagar:
@jgm that's because this PR relies on https://github.com/jgm/pandoc-types/pull/68

jgm:
> @jgm that's because this PR relies on jgm/pandoc-types#68

You can add something to stack.yaml and cabal.project telling it to use a particular commit from your PR.  Indeed, pandoc currently has such a stanza because we're depending on some table changes.  (You'll need to rebase your pandoc-types changes on top of the commit pandoc is currently depending on, and then change stack.yaml and cabal.project to point to your version.)

vaibhavsagar:
Thanks, I did that and now all tests are passing except on macOS, where there is a GHC panic that I don't think I'm responsible for.

vaibhavsagar:
My guess for why the error is occurring is that you are caching `.stack-work`, which I don't think is a good idea.

jgm:
Can you say more about why caching `.stack-work` is a bad idea?
It has dramatically cut down compile times on CI.

jgm:
I've just pushed a commit removing the caching of `.stack-work`; we can see if that helps.
(You'd have to rebase on this.)

vaibhavsagar:
It might be partly superstition, but in the past I remember encountering weirdness in the way `stack` tries to incrementally rebuild, because caching intermediate build artifacts is a hard problem that I don't expect them to have necessarily solved in full generality. I would expect the same issues if we were caching `dist-newstyle`. The error I saw would be symptomatic of this issue, because it seems like it's looking for a symbol that doesn't exist yet as it's using a stale intermediate build artifact instead of compiling from scratch (which is slower but always correct).

This is only a guess on my part, though, and I do understand that caching `.stack-work` makes CI significantly faster. For the project I maintain (https://github.com/gibiansky/IHaskell) I've settled on caching `~/.stack` which I think is more likely to be correct since it only contains finished build outputs.

vaibhavsagar:
Hmm, this still seems to be happening without `.stack-work` so it appears my guess was wrong. I have no idea what the real cause of the error is.

jgm:
I tried building your branch locally on my macOS computer, and got the same error we get in CI.

jgm:
Unfortunately, this is a blocker.  It's probably worth reporting it as a bug to ghc, as the message requests, since it's reproducible.

jgm:
For reference, here's the error I get on my machine:
```
<no location info>: error:
    ghc: panic! (the 'impossible' happened)
  (GHC version 8.6.5 for x86_64-apple-darwin):
        Loading temp shared object failed: dlopen(/var/folders/l0/2t_cldbj26j_vsd9_q2tsf400000gn/T/ghc54764_0/libghc_107.dylib, 5): Symbol not found: _pandoczmtypeszm1zi20zmBUPf2Tq5atP8Lb94Ld23Jj_TextziPandocziBuilder_underline_closure
  Referenced from: /var/folders/l0/2t_cldbj26j_vsd9_q2tsf400000gn/T/ghc54764_0/libghc_107.dylib
  Expected in: flat namespace
 in /var/folders/l0/2t_cldbj26j_vsd9_q2tsf400000gn/T/ghc54764_0/libghc_107.dylib
                        
Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug                    
```

One more idea:  try incrementing the version number in your pandoc-types.  Eventually the version will need to go to 1.21 because of API changes (underline, table), but if you change that you'd have to change texmath and other dependencies, so to keep it simple you could try 1.20.1.

jgm:
I'm going to bump the pandoc-types version and release a new texmath that allows this, if you want to go ahead and rebase on pandoc-types HEAD.

jgm:
Same thing again, huh.  I don't see any problems in your code, but this is reproducible.

mb21:
I remember having that error as well on this branch when I first checked it out a week ago or something, and I'm on macOS Catalina. I just nuked `~/.stack/` and then it worked again :S

now pulled this branch and `stack build` still works.

jgm:
I'm going to try changing to a more conservative caching approach: only use cache if stack.yaml hasn't changed.

jgm:
@vaibhavsagar try rebasing again on master and maybe the caching change will make the problem disappear.

vaibhavsagar:
Thank you for your support with this PR @jgm :smile:.

vaibhavsagar:
That caching change seems to have made the macOS build pass! The new failure is because the `Muse` writer doesn't support underlines, so the roundtrip is failing. I think the solution there would be to modify the `Arbitrary` instance for all writers that don't support underlines so that they aren't generated, but I can't see how to do this. What do you think?

jgm:
See the approach in Tests.Readers.HTML.  Before running rt tests, we purify the AST by removing elements that can't round-trip.  You could add something similar to the Muse tests.

vaibhavsagar:
I took a closer look and learned that underlines are actually supported in Emacs Muse, so I updated the writer accordingly and this test run seems to be going much better.

vaibhavsagar:
Is there anything further you need from me on this PR?

jgm:
It looks good on first glance; I just need to make time for a more careful review.

mb21:
For the record: switching back to master branch from this branch, couldn't build anymore either, got:

```
ghc: panic! (the 'impossible' happened)
  (GHC version 8.6.5 for x86_64-apple-darwin):
	Loading temp shared object failed: dlopen(/var/folders/55/rrvbjfdn5fd0wf7p8r8spc_m0000gn/T/ghc53115_0/libghc_214.dylib, 5): Symbol not found: _pandoczmtypeszm1zi21zmLb7evYKD6jv46JnhbWrkOK_TextziPandocziDefinition_zdfDataInline60_closure
  Referenced from: /var/folders/55/rrvbjfdn5fd0wf7p8r8spc_m0000gn/T/ghc53115_0/libghc_214.dylib
  Expected in: flat namespace
 in /var/folders/55/rrvbjfdn5fd0wf7p8r8spc_m0000gn/T/ghc53115_0/libghc_214.dylib

Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug


--  While building package pandoc-2.9.2.1 using:
      /Users/maurobieg/.stack/setup-exe-cache/x86_64-osx/Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.5 --builddir=.stack-work/dist/x86_64-osx/Cabal-2.4.0.1 build lib:pandoc exe:pandoc --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
```

jgm:
Probably worth reporting this as a bug.
Deleting the stack caches seems to fix it, but that's not pleasant.

src/Text/Pandoc/Writers/CommonMark.hs
```diff
@@ -237,6 +237,10 @@ inlineToNodes opts SoftBreak
   | writerWrapText opts == WrapNone     = (node (TEXT " ") [] :)
   | otherwise                           = (node SOFTBREAK [] :)
 inlineToNodes opts (Emph xs) = (node EMPH (inlinesToNodes opts xs) :)
+inlineToNodes opts (Underline xs)
+  | isEnabled Ext_raw_html opts = ((node (HTML_INLINE (T.pack "<u>")) [] : inlinesToNodes opts xs ++
```
jgm, line 241: Looks like this line is > 80 characters long.

jgm: I wonder if it would make more sense to fall back to regular Emph behavior when no more specific rendering of underlined text is available -- both here and in other writers?

src/Text/Pandoc/Writers/Haddock.hs
```diff
@@ -209,6 +209,10 @@ inlineToHaddock opts (Span (ident,_,_) ils) = do
 inlineToHaddock opts (Emph lst) = do
   contents <- inlineListToHaddock opts lst
   return $ "/" <> contents <> "/"
+-- Underline is not supported
+inlineToHaddock opts (Underline lst) = do
+  contents <- inlineListToHaddock opts lst
+  return contents
```
jgm, line 215: Would it make sense to render as Emphasis here?  I'm not sure.

src/Text/Pandoc/Writers/Jira.hs
```diff
@@ -193,6 +193,7 @@ toJiraInlines inlines = do
         Code _ cs          -> return . singleton $
                               Jira.Monospaced (escapeSpecialChars cs)
         Emph xs            -> styled Jira.Emphasis xs
+        Underline xs       -> styled Jira.Insert xs
```
jgm, line 196: Semantically Insert seems different than underline, even if underline is used to indicate it.  But this is probably okay for now.  (At some point we might add Inserted and Deleted elements.)

src/Text/Pandoc/Writers/Man.hs
```diff
@@ -270,6 +270,9 @@ inlineToMan :: PandocMonad m => WriterOptions -> Inline -> StateT WriterState m
 inlineToMan opts (Span _ ils) = inlineListToMan opts ils
 inlineToMan opts (Emph lst) =
   withFontFeature 'I' (inlineListToMan opts lst)
+-- Underline is not supported, so treat the same as Emph
+inlineToMan opts (Underline lst) =
+  withFontFeature 'I' (inlineListToMan opts lst)
```
jgm, line 275: Here I notice you do render as Emph; would it make sense to do that in other writers where underline is not supported?

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -1045,6 +1045,12 @@ inlineToMarkdown opts (Emph lst) = do
                       then "_" <> contents <> "_"
                       else contents
               else "*" <> contents <> "*"
+inlineToMarkdown _ (Underline []) = return empty
+inlineToMarkdown opts (Underline lst) = do
+  contents <- inlineListToMarkdown opts lst
+  return $ if isEnabled Ext_raw_html opts
+              then "<u>" <> contents <> "</u>"
+              else contents
```
jgm, line 1053: Raw HTML is not ideal.  I'm thinking we should render underline in Markdown the way we've been doing already, as a bracketed Span with a class (falling back to raw HTML only when bracketed Spans aren't enabled).  (And the Markdown reader should recognize that syntax as Underline.)  That will give us round-trip behavior without the mess of raw HTML.  What do you think?  Maybe also, when bracketed spans aren't enabled but native spans are, we should use `<span class="">` syntax.

src/Text/Pandoc/Writers/MediaWiki.hs
```diff
@@ -372,6 +372,10 @@ inlineToMediaWiki (Emph lst) = do
   contents <- inlineListToMediaWiki lst
   return $ "''" <> contents <> "''"
 
+inlineToMediaWiki (Underline lst) = do
+  contents <- inlineListToMediaWiki lst
+  return $ "<u>" <> contents <> "</su>"
```
jgm, line 377: Is `</su>` correct here? Should it be `</u>`?

src/Text/Pandoc/Writers/Ms.hs
```diff
@@ -351,6 +351,9 @@ inlineToMs :: PandocMonad m => WriterOptions -> Inline -> MS m (Doc Text)
 inlineToMs opts (Span _ ils) = inlineListToMs opts ils
 inlineToMs opts (Emph lst) =
   withFontFeature 'I' (inlineListToMs opts lst)
+inlineToMs opts (Underline lst) = do
+  contents <- inlineListToMs opts lst
+  return $ literal ".UL " <> contents <> " "
```
jgm, line 356: This isn't what you want here.  Note that the `.UL` can only go at the beginning of a line, and this might be rendered anywhere.  If it's not possible to use `withFontFeature` as we do for Emph, then at the very least we need to force a `cr` before and after here.  But that could cause problems when the inline sequence is itself the argument of something like `.SE`.  Maybe the best approach would just be treating underline exactly like emphasis in this format?

src/Text/Pandoc/Writers/Muse.hs
```diff
@@ -594,6 +594,13 @@ inlineToMuse (Strong [Emph lst]) = do
     else if null lst' || startsWithSpace lst' || endsWithSpace lst'
            then emphasis "**<em>" "</em>**" lst'
            else emphasis "***" "***" lst'
+-- | Underline is only supported in Emacs Muse mode.
+inlineToMuse (Underline lst) = do
+  opts <- asks envOptions
+  contents <- inlineListToMuse lst
+  if isEnabled Ext_amuse opts
+     then return $ "_" <> contents <> "_"
+     else return contents
```
jgm, line 603: fall back as emph?

src/Text/Pandoc/Writers/RST.hs
```diff
@@ -581,6 +585,10 @@ inlineToRST (Span (_,_,kvs) ils) = do
 inlineToRST (Emph lst) = do
   contents <- writeInlines lst
   return $ "*" <> contents <> "*"
+-- Underlining is not supported
+inlineToRST (Underline lst) = do
+  contents <- writeInlines lst
+  return contents
```
jgm, line 591: fall back to emph?

src/Text/Pandoc/Writers/Texinfo.hs
```diff
@@ -392,6 +392,10 @@ inlineToTexinfo (Span _ lst) =
 inlineToTexinfo (Emph lst) =
   inCmd "emph" <$> inlineListToTexinfo lst
 
+-- Underlining does not appear to be supported
+inlineToTexinfo (Underline lst) =
+  inlineListToTexinfo lst
```
jgm, line 397: fall back to emph?

src/Text/Pandoc/Writers/MediaWiki.hs
vaibhavsagar: Oops, thanks for pointing this out!

src/Text/Pandoc/Writers/Ms.hs
vaibhavsagar: Updated to behave the same as `Emph`.

src/Text/Pandoc/Writers/RST.hs
vaibhavsagar: Done.

src/Text/Pandoc/Writers/Texinfo.hs
vaibhavsagar: Done.

src/Text/Pandoc/Writers/Muse.hs
vaibhavsagar: Done.

src/Text/Pandoc/Writers/Haddock.hs
vaibhavsagar: Done.

src/Text/Pandoc/Writers/CommonMark.hs
vaibhavsagar: Done.

src/Text/Pandoc/Writers/Man.hs
vaibhavsagar: Done.

src/Text/Pandoc/Writers/Markdown.hs
vaibhavsagar: I implemented this change in both the reader and writer. I wasn't sure if the class should be `ul` as indicated above or `underline`, so my current approach supports both.

vaibhavsagar:
Most of the readers were already generating `underlineSpan` in case of an underline, so I changed that to output `Underline` instead of a span as part of my earlier changes. I think that takes care of most of the readers, but I'm not sure if I got them all.

jgm:
Oh right, you handle the readers already via the change to underlineSpan. I'd missed that.
I think it would be cleaner to deprecate underlineSpan and just use B.underline directly in the readers instead of underlineSpan.

jgm:
Otherwise this is looking good.  I'll go ahead and merge the pandoc-types changes, and then you can also change to depend on master in pandoc-types.

vaibhavsagar:
Updated, thanks!

src/Text/Pandoc/Shared.hs
```diff
@@ -750,7 +750,7 @@ eastAsianLineBreakFilter = bottomUp go
         go xs
           = xs
 
--- | Builder for underline.
+-- | Builder for underline (deprecated).
```
jgm, line 753: Before the comment let's add
```
{-# DEPRECATED underlineSpan "Use Text.Pandoc.Builder.underline instead" #-}
```

vaibhavsagar: Done.

jgm:
Thanks!

nackd:
I'm looking into converting HTML with underlines to RTF and got here. It doesn't seem to work or at least nothing is underlined when I open the output files in LibreOffice. This is using the `\pnul` control word. I don't know much about the RTF format, but `\pn*` control words seem to be related with paragraph numbering. When I replace `\pnul` with `\ul`, the files look as expected on LibreOffice. Can someone confirm it should be `\ul`?

jgm:
Note that this change has been merged into master but it's not yet in any released version.

However, in all likelihood we just haven't supported underline yet in RTF.  If you can confirm that (using a nightly) then please submit a new issue for that.

jgm:
Ah, I see -- `\pnul` is for paragraph number styling.  Yes, it should be `\ul`. I can make that chaneg.

nackd:
Great, thank you!

jgm:
Done

PR #6289, opened 2020-04-15
[Markdown Reader] Fix inline code in lists (#6284)

### LINKED ISSUE #6284: Markdown parsing bug: inline code in list item
Seems like this is as of yet unreported, sorry for the noise if it is. Anywho, given the following Markdown:

```markdown
1. If `(1) x`, then `2`
```
Pandoc produces:
```haskell
[OrderedList (1,Decimal,Period)
 [[Plain [Str "If",Space,Str "`(1)",Space,Str "x",Code ("",[],[]) ", then",Str "2`"]]]]
```

This is unexpected, I would expect something along the lines of

```haskell
[OrderedList (1,Decimal,Period)
 [[Plain [Str "If",Space,Code ("",[],[]) "(1) x",Str ",",Space,Str "then",Space,Code ("",[],[]) "2"]]]]
```

Parser seems to be confused specifically by the `(1) ` part, without it, it works as expected. Also works as expected outside of list items.

Also breaks in the same way with things like

```markdown
1. `#. x`
```

Seems like a precedence conflict between inline code and list markers?

lierdakil:
This should fix #6284, as far as I can tell. All this does is moves the check for `notFollowedBy (inList >> listStart)` to after newline. I took the liberty of formatting the relevant code to hopefully be a little more readable.

Sorry for the mess I've arguably made in the tests, I've developed a splitting headache halfway through making those, and basically resorted to throwing semi-random strings at the compiler until it did what I wanted it to. The second part of the tests ("lists with X in backticks") are there basically to make sure I didn't break anything with the change. Those should pass both before and after the patch. The first part ("code with list marker ...") contains some examples that motivated the change in the first place, i.e. those fail on the current pandoc version.

### DIFF
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -1517,12 +1517,17 @@ code :: PandocMonad m => MarkdownParser m (F Inlines)
 code = try $ do
   starts <- many1 (char '`')
   skipSpaces
-  result <- (trim . T.concat) <$>
-            manyTill (notFollowedBy (inList >> listStart) >>
-                      (many1Char (noneOf "`\n") <|> many1Char (char '`') <|>
-                       (char '\n' >> notFollowedBy' blankline >> return " ")))
-                      (try (skipSpaces >> count (length starts) (char '`') >>
-                      notFollowedBy (char '`')))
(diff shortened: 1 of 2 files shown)


jgm:
Great, thanks!

PR #6307, opened 2020-04-22
Manual: separate adjacent verbatim code blocks

tom-audm:
These should appear as separate blocks (since they can't all exist within the same document) but currently appear combined in single blocks.

Visible here: https://pandoc.org/MANUAL.html#metadata-blocks


jgm:
The change makes sense.
However, for consistency with the rest of the manual, use ` ``` ` instead of ` ~~~ ` for the fenced code blocks, and remove the indentation inside the code blocks.

jgm:
Great, thanks!

PR #6310, opened 2020-04-23
Fix Apple Books bug w/overflow in source highlighting

### LINKED ISSUE #6242: Syntax highlighting broken in Apple Books v2.4 (1923)
Hi, recently I'm recompiled my markdown ebook and found that is all blocks of code are invisible in Apple Books.
I don’t know which update broke this OSX or Pandoc itself, but digging deeper I found a simple example that breaks it.

Info:
```
OS: MacOS Catalina 10.15.4 (19E266)
Apple Books v2.4 (1923)
pandoc 2.9.2.1
Compiled with pandoc-types 1.20, texmath 0.12.0.1, skylighting 0.8.3.2
```
File:

[test_case.md](https://github.com/jgm/pandoc/files/4410109/test_case.txt)

Runs:

`pandoc -t epub3 -o test_case.epub test_case.md` - second block of code is invisible
broken

`pandoc -t epub3 -o test_case.epub test_case.md` - test_case.md without 16 line, everything is fine 
without_16line

`pandoc --no-highlight -t epub3 -o test_case.epub test_case.md` - original test_case.md with 16 line, everythins is fine
no-highlight

nateberkopec:
Fixes #6242


data/epub.css
```diff
@@ -23,3 +23,8 @@ span.underline{ text-decoration: underline; }
 q { quotes: "“" "”" "‘" "’"; }
 div.column{ display: inline-block; vertical-align: top; width: 50%; }
 div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
+/* Apple Books 2.4+ doesn't like overflow:auto on syntax highlighting generated by skylighting */
```
nateberkopec, line 26: I noticed a lot of similar CSS rules in the final, generated epub CSS with hacks for kindle, etc. Is that skylighting generating that? Should we, instead, fix this there?

jgm: Yes.  skylighting adds this in the CSS:
``` css
@media screen {
 div.sourceCode { overflow: auto; }
}
```
Code is in skylighting-core, Skylighting.Format.HTML
The problem is that we *want* the behavior of overflow: auto, not overflow:visible, which causes long code lines to overflow the box.  Changing this would not be desirable for many users (and note that the same code is used for regular HTML).

jgm: If you can override it in your own epub.css, that's good to know.
Have you tried?

nateberkopec: Yup! Works fine.

jgm:
Now fixed in another way (adding wrapping instead of just letting the lines be cut off).

PR #6328, opened 2020-05-01
Unify defaults metadata and markdown metadata parsers

lierdakil:
This is prompted by https://github.com/lierdakil/pandoc-crossref/issues/259.

Apparently, pandoc doesn't parse metatdata fields in the defaults file the way it does it for Markdown files, particularly boolean fields are parsed as strings. I took a gander, and to my surprise, found that pandoc uses completely different code path for parsing defaults file metadata field compared to yaml metadata block in Markdown.

Long story short, this is a first attempt at merging those code paths. This should hopefully behave _a little_ more consistently (although I'll admit I quickly gave up on figuring out how to parse fields as Markdown while parsing defaults file -- I believe it's feasible, but likely more than a bit tricky)

P.S. This is more a request for comments than a bona fide pull request. The code perhaps needs a little polish. And I keep forgetting about the draft PR feature.

P.P.S. Sorry if this text is a little rambly or a bit incoherent, it's 4 AM and I'm slightly sleep-deprived.

### DIFF
--- a/src/Text/Pandoc/App/Opt.hs
+++ b/src/Text/Pandoc/App/Opt.hs
@@ -34,9 +34,13 @@ import Text.Pandoc.Options (TopLevelDivision (TopLevelDefault),
                             ObfuscationMethod (NoObfuscation),
                             CiteMethod (Citeproc))
 import Text.Pandoc.Shared (camelCaseStrToHyphenated)
-import Text.DocLayout (render)
-import Text.DocTemplates (Context(..), Val(..))
+import qualified Text.Pandoc.Parsing as P
+import Text.Pandoc.Readers.Metadata (yamlMap)
+import Text.Pandoc.Class.PandocPure
+import Text.DocTemplates (Context(..))
 import Data.Text (Text, unpack)
+import Data.Default (def)
+import Control.Monad (join)
 import qualified Data.Text as T
 import qualified Data.Map as M
 import Text.Pandoc.Definition (Meta(..), MetaValue(..), lookupMeta)
@@ -185,8 +189,7 @@ doOpt (k',v) = do
       -- Note: x comes first because <> for Context is left-biased union
       -- and we want to favor later default files. See #5988.
     "metadata" ->
-      parseYAML v >>= \x -> return (\o -> o{ optMetadata = optMetadata o <>
-                                               contextToMeta x })
+      return (\o -> o{ optMetadata = optMetadata o <> yamlToMeta v })
     "metadata-files" ->
       parseYAML v >>= \x ->
                         return (\o -> o{ optMetadataFiles =
@@ -475,16 +478,14 @@ defaultOpts = Opt
     , optStripComments         = False
     }
 
-contextToMeta :: Context Text -> Meta
-contextToMeta (Context m) =
-  Meta . M.map valToMetaVal $ m
-
-valToMetaVal :: Val Text -> MetaValue
-valToMetaVal (MapVal (Context m)) =
-  MetaMap . M.map valToMetaVal $ m
-valToMetaVal (ListVal xs) = MetaList $ map valToMetaVal xs
-valToMetaVal (SimpleVal d) = MetaString $ render Nothing d
-valToMetaVal NullVal = MetaString ""
+yamlToMeta :: Node Pos -> Meta
+yamlToMeta (Mapping _ _ m) = runEverything (yamlMap pMetaString m)
+  where
+    pMetaString = pure . MetaString <$> P.manyChar P.anyChar
+    runEverything :: P.ParserT Text P.ParserState PandocPure (P.F (M.Map Text MetaValue)) -> Meta
+    runEverything p =
+      either (const mempty) (Meta . flip P.runF def) . join . runPure $ P.readWithM p def ""
+yamlToMeta _ = mempty
 
 addMeta :: String -> String -> Meta -> Meta
 addMeta k v meta =
--- a/src/Text/Pandoc/App/Opt.hs
+++ b/src/Text/Pandoc/App/Opt.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE TemplateHaskell     #-}
 {-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE FlexibleContexts    #-}
 {- |
    Module      : Text.Pandoc.App.Opt
    Copyright   : Copyright (C) 2006-2020 John MacFarlane
@@ -40,7 +41,6 @@ import Text.Pandoc.Class.PandocPure
 import Text.DocTemplates (Context(..))
 import Data.Text (Text, unpack)
 import Data.Default (def)
-import Control.Monad (join)
 import qualified Data.Text as T
 import qualified Data.Map as M
 import Text.Pandoc.Definition (Meta(..), MetaValue(..), lookupMeta)
@@ -189,7 +189,7 @@ doOpt (k',v) = do
       -- Note: x comes first because <> for Context is left-biased union
       -- and we want to favor later default files. See #5988.
     "metadata" ->
-      return (\o -> o{ optMetadata = optMetadata o <> yamlToMeta v })
+      yamlToMeta v >>= \x -> return (\o -> o{ optMetadata = optMetadata o <> x })
     "metadata-files" ->
       parseYAML v >>= \x ->
                         return (\o -> o{ optMetadataFiles =
@@ -478,14 +478,14 @@ defaultOpts = Opt
     , optStripComments         = False
     }
 
-yamlToMeta :: Node Pos -> Meta
-yamlToMeta (Mapping _ _ m) = runEverything (yamlMap pMetaString m)
+yamlToMeta :: Node Pos -> Parser Meta
+yamlToMeta (Mapping _ _ m) =
+    either (fail . show) return $ runEverything (yamlMap pMetaString m)
   where
     pMetaString = pure . MetaString <$> P.manyChar P.anyChar
-    runEverything :: P.ParserT Text P.ParserState PandocPure (P.F (M.Map Text MetaValue)) -> Meta
-    runEverything p =
-      either (const mempty) (Meta . flip P.runF def) . join . runPure $ P.readWithM p def ""
-yamlToMeta _ = mempty
+    runEverything p = runPure (P.readWithM p def "")
+      >>= fmap (Meta . flip P.runF def)
+yamlToMeta _ = return mempty
 
 addMeta :: String -> String -> Meta -> Meta
 addMeta k v meta =
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -67,14 +67,17 @@ yamlToMeta :: PandocMonad m
            -> m Meta
 yamlToMeta opts bstr = do
   let parser = do
-        meta <- yamlBsToMeta parseBlocks bstr
+        meta <- yamlBsToMeta (fmap asBlocks parseBlocks) bstr
         return $ runF meta defaultParserState
   parsed <- readWithM parser def{ stateOptions = opts } ""
   case parsed of
     Right result -> return result
     Left e       -> throwError e
+  where
 
 
+asBlocks :: Functor f => f (B.Many Block) -> f MetaValue
+asBlocks p = MetaBlocks . B.toList <$> p
 
 --
 -- Constants and data structure definitions
@@ -240,7 +243,7 @@ yamlMetaBlock = try $ do
   -- by including --- and ..., we allow yaml blocks with just comments:
   let rawYaml = T.unlines ("---" : (rawYamlLines ++ ["..."]))
   optional blanklines
-  newMetaF <- yamlBsToMeta parseBlocks
+  newMetaF <- yamlBsToMeta (asBlocks <$> parseBlocks)
               $ UTF8.fromTextLazy $ TL.fromStrict rawYaml
   -- Since `<>` is left-biased, existing values are not touched:
   updateState $ \st -> st{ stateMeta' = stateMeta' st <> newMetaF }
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -67,15 +67,12 @@ yamlToMeta :: PandocMonad m
            -> m Meta
 yamlToMeta opts bstr = do
   let parser = do
-        meta <- yamlBsToMeta (asBlocks <$> parseBlocks) bstr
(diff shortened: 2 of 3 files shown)


jgm:
This is intentional.  The defaults file is supposed to be more or less equivalent to specifying things on the command line. When you specify `--metadata`, it gets treated as a string, not parsed as Markdown. It wouldn't in general make sense to parse it as Markdown -- maybe you're not converting from Markdown!  Same reasoning for the different treatment in the defaults file.

lierdakil:
Okay, ignore Markdown parsing part then, this patch doesn't do it anyway. The main issue here is booleans don't get parsed as booleans. And it's entirely possible to specify booleans in metadata on the CLI.

lierdakil:
I guess the question is: do you want to have the same semantics in the Markdown reader metadata parser and defaults metadata parser? (modulo parsing strings as Markdown)

If not, I would argue you need a good reason for that, because different semantics will lead to much confusion.

jgm:
If booleans can be specified using `--metadata` on the command line, then yes, it should be the same in the defaults file.

alerque:
@lierdakil You can now convert existing PRs to Draft mode. The link to do so us under the "Reviewers" section in the sidebar.

brainchild0:
For boolean values given in a YAML file, I find no reason not to follow YAML semantics for evaluating the value and type. The fact of the fields being the same as those that might be given on the command line gives no reason that the way the text is processed should be the  same.

Or am I missing the issue?

jgm:
@lierdakil - can you clarify the status of this draft?  What work is still needed before it's ready to review?

brainchild0:
> @lierdakil - can you clarify the status of this draft? What work is still needed before it's ready to review?

Further to the above, identical handling of the same verbatim YAML block in both contexts seems to be a very important ideal. Ultimately this objective would seem to ask for fully merging the functionality into a single path, not necessarily respecting execution sequence of overall processing but certainly respecting which code is executed.  I'm curious to understand how closely  the current work reaches such an ideal, that is how much non-trivial divergence remains comparing the two paths.

lierdakil:
So, okay, a bit of an explanation of how I'm going about this.

There's a function `yamlMap` in `Text.Pandoc.Readers.Metadata`, which takes a block-parser and a YAML object and parses it into a `MetaMap`. This is specifically the parser that is used in parsing Markdown metadata blocks (albeit in the "normal" code path this happens through the proxy of `yamlBsToMeta`). The slight issue with this function is that it expects block parser to return `Blocks`, which is less flexible than we would want here, so I have to amend it to instead expect a parser that returns a `MetaValue` straight away (whatever that `MetaValue` happens to be). This is relatively trivial, because `yamlMap` parser wraps the result of block-parser in `MetaBlocks` anyway.

Given this modification, I'm re-using `yamlMap` to parse metadata YAML object in the defaults file using this parser. However, instead of the "blocks parser" that the function originally expects, I'm passing a dumb "parser" that just returns the input string wrapped as a `MetaString`.

The original question I was implicitly asking with this PR is whether or not this all seems like a sound approach. On the positive side, this guarantees consistent handling of metadata in the defaults file and in-line as metadata blocks, with the notable exception of parsing string values as Markdown (in the current implementation, defaults file strings are treated as unformatted strings). On the negative side, this doesn't at all guarantee that anything would be consistent with the CLI (which uses a different code path altogether). Another possible negative is that the code ends up being somewhat convoluted, although I'm not sure if there's a better way to handle deep-ish monad stacks anyway.

As for what still needs to be done, the current implementation silently ignores any parsing errors reported by `runPure`. Frankly, with the current implementation, there isn't much that could go wrong in the first place, but to future-proof, I would argue it'd be good to handle possible errors anyway. The modification required is somewhat trivial as far as I can tell, but I didn't find the time to make it yet. Furthermore, the code needs a little bit of a stylistic clean-up, there are some useless hanging `where`s in places, and variable names could perhaps use some work (for instance `pBlocks` isn't really a blocks-parser at this point, but a metadata-parser)

I'd like to encourage a quick look over the code as it is now (while the changeset is relatively small and contains little noise), to figure out if I should pursue this further, or if this approach should be abandoned in favour of another perhaps.

src/Text/Pandoc/App/Opt.hs
```diff
@@ -475,16 +478,14 @@ defaultOpts = Opt
...
-valToMetaVal (SimpleVal d) = MetaString $ render Nothing d
-valToMetaVal NullVal = MetaString ""
+yamlToMeta :: Node Pos -> Meta
+yamlToMeta (Mapping _ _ m) = runEverything (yamlMap pMetaString m)
+  where
+    pMetaString = pure . MetaString <$> P.manyChar P.anyChar
+    runEverything :: P.ParserT Text P.ParserState PandocPure (P.F (M.Map Text MetaValue)) -> Meta
+    runEverything p =
+      either (const mempty) (Meta . flip P.runF def) . join . runPure $ P.readWithM p def ""
+yamlToMeta _ = mempty
```
jgm, line 488: I'm confused about this code. What's the point of `runEverything` here if we're just passing in `pMetaString`?

jgm:
Quick review:

I am okay with the change in T.P.Readers.Metadata.

I don't quite understand some of the added complexity in T.P.App.Opt, as noted in the comment on the code.

Also, I thought that one of the aims here was to ensure that CLI metadata and (leafs nodes for) metadata in defaults files are parsed the same (particularly as regards boolean values), and I don't see that in this PR.

src/Text/Pandoc/App/Opt.hs
lierdakil: Basically: to make the type checker happy. `yamlMap` is several monad stacks deep before we get to the actualy `MetaMap`, `runEverything` has to unwrap those.

FWIW, some of those layers need to be joined with the `Parser` in `doOpt` for the purposes of error handling, but I was being a little lazy with the proof-of-concept.

lierdakil:
> Also, I thought that one of the aims here was to ensure that CLI metadata and (leafs nodes for) metadata in defaults files are parsed the same (particularly as regards boolean values), and I don't see that in this PR.

As I said above, this PR does not touch the **CLI** code path at all. The idea here is to handle metadata specified from the defaults file the same as metadata specified via the YAML metadata block (because both are YAML metadata, and discrepancies are confusing). The way I achieve the intended goal is to pass the whole YAML object containing the metadata in the defaults file to the parser in `Text.Pandoc.Readers.Metadata`. Booleans are handled by the `yamlMap` parser directly.

lierdakil:
Basically, this is what I'm trying to point out: we don't have to reinvent the wheel with parsing metadata in the defaults file, we already have a parser for YAML metadata, which works in the context of the Markdown parser. We can use that one with minimal modification (at the cost of some arguably ugly monad stack unwrapping). That reduces effort duplication and gives us some UX consistency for free. If this sounds reasonable then I can tidy this PR up, and we can move on to review/merge and whatnot. Otherwise, we can abandon this angle and I can quickly slap together some ad-hoc measure to fix boolean parsing in defaults metadata and be done with that (this will lead to some code duplication, which irks me because DRY, but oh well)

Sorry if I'm failing to express my thoughts clearly.

brainchild0:
If I might try  to summarize, as well to embellish, as seen helpful, comments given up to now by @lierdakil:

1. Previously, Pandoc compiled metadata  from  two sources, the command line and the metadata block. The processing of these inputs has been different, as might be expected and desired, due to the different constraints and capabilities relating to the structure of  data from the distinct sources.

1. Currently, Pandoc compiles metadata from  three sources, adding to the above the defaults file.

1. The metadata block and the defaults file are both structured as YAML data and formatted as  YAML blocks. Thus, overall, the two sources carry the same constraints and capabilities.

1. Following from the above, numerous advantages follow from common processing,   except whatever differences minimally might be  required,  of the metadata given in the metadata block versus the metadata given in the defaults file.

    These advantages include:
   
    1.  The user may enjoy the convenience of copying  a block of YAML source from one file to another in an editor, without  editing the contents, and yet having a predictable effect.

    1.  The application needs to include only a single  implementation for processing any metadata given through YAML.

    1.  Documentation needs to represent only a single set of rules for such processing,  while yet preserving clarity, accuracy, or completeness.

    1.  Users need to learn only a single set of rules.

    1.  Users cannot confuse one set of rules while attempting to apply another.



1. Currently, due to historical reasons, the processing of metadata in the metadata block takes better advantage of the structure and format of YAML  than processing of metadata in the defaults file.

1. Following from the above, the earlier advantageous would be best realized by a change to the application in which the logic that currently processes data in the metadata block  also processes metadata in the defaults file.

1. If starting from scratch, the common logic, for processing both the metadata block and the defaults file,  might  be provided most wisely in  a separate common dependency.


1. From the current situation, attempting this particular  restructuring is unfavorably drastic,  at least until the concept, separate from any actual implementation, is accepted.

1. Following from the above,  the current PR has attempted, toward the overall objective,  a minimal set of modifications, in which  the logic that processes the defaults file depends on the logic that processes the metadata block, through revision of the former and some minimal necessary changes to the latter.

1. Although the PR may need more work before  merging, attempting this further work seems most befitting following  an indication of administrative approval of the overall efforts.

brainchild0:
@jgm: The author has referenced an issue in an external project with a user base that is adversely affected by current application design. He has further explained the intrinsic  utility  for unifying processing of the same data fields from the two  sources, and has demonstrated a working case. Supporting his contribution would allow  him to close the issue in `pandoc-crossref`,  for which he otherwise has no clean solution,  and also to provide a valuable improvement to Pandoc. Putting aside any potential objections from code structure or code quality related to the current state of work, does any hesitation remain, and if so, for what reasons, against pursuing the overall objectives?

jgm:
@brainchild0 - just a note that comments like the last one are motivation-killers in open-source projects.  I haven't forgotten about this PR, but believe it or not, I have other work to do besides pandoc.  I've indicated general agreement with the proposed change -- my last message asks for clarification only, which I happily received.  The walls of text you have provided make it harder to review the PR; please keep your comments as succinct as possible.

brainchild0:
@jgm: Although the overall intention was different from what may have been construed, I understand the view you express. I removed the preceding comment, which I am interpreting as the target of your remark, being the only one you indicated explicitly. (Please be aware that my reading of the conversation, though perhaps inaccurate, was not one of agreement. It may be beside the point.)

lierdakil:
@jgm, so, okay, sorry it took a while, I got busy with day job and didn't find much time to work on this.

Anyway, I've updated the PR. A few notes:

-  Error handling is somewhat anaemic, since `PandocError` ideally should be handled in IO, but we're not in IO -- and the only other option is to basically `fail . show`. Note that at present there are no errors to handle -- there is no code path in `yamlToMeta` that would throw a pure exception as far as I can tell, so this is more future-proofing than anything.
- I took the liberty to clean up some code in `T.P.R.Metadata`. That includes renaming some bindings to hopefully avoid confusion in the future and getting rid of some do-notation, one menacingly looking `case` tower, and unused language extensions.
- While at it, I also removed references to `RelaxedPolyRec` language extension everywhere I found it. Since I believe  the current pandoc source can't be built with GHC 7 and below anyway, and `RelaxedPolyRec` is literally impossible to turn off since at least GHC 7.6, and almost all mention of this extension is removed from GHC documentation, I'd say it's a good idea to remove this extension as obsolete.

lierdakil:
This is a friendly reminder that this PR is ready for review. I'm afraid this will go out of sync with master if it marinates for much longer, and it does fix a real issue (report linked in the OP). No pressure, just in case it slipped through the cracks.

jgm:
Sorry, I just forgot about this. I'll take a look now.

src/Text/Pandoc/Readers/MediaWiki.hs
```diff
@@ -1,6 +1,4 @@
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RelaxedPolyRec    #-}
```
jgm, line 2: Can you split these deletions of RelaxedPolyRec into a separate PR, since they aren't related?

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -68,14 +67,15 @@ yamlToMeta :: PandocMonad m
...
+        meta <- yamlBsToMeta (asBlocks <$> parseBlocks) bstr
         return $ runF meta defaultParserState
   parsed <- readWithM parser def{ stateOptions = opts } ""
   case parsed of
     Right result -> return result
     Left e       -> throwError e
 
-
+asBlocks :: Functor f => f (B.Many Block) -> f MetaValue
+asBlocks p = MetaBlocks . B.toList <$> p
```
jgm, line 78: `asMetaBlocks` might be a better name?

jgm:
I think I'm fine with this change, though I have a nagging worry that there's something I'm overlooking.  What more would be involved in making boolean values work in metadata in defaults files?

I'd like to make a pandoc release very soon, so we can support ghc 8.10, and it would be convenient to include this (since we'd be going to 2.10 anyway and this is an API change).

jgm:
Btw it's helpful if all API changes are clearly indicated as such in the commit messages, so I don't forget to include them in the changelog.

src/Text/Pandoc/Readers/Markdown.hs
tarleb: Sidenote: this is the same as `fmap toMetaValue`, where the latter is from Text.Pandoc.Builder.

lierdakil: @tarleb, good point.

lierdakil:
Okay, I've removed RelaxedPolyRec stuff and `asBlocks` (in favour of `fmap toMetaValue` as per @tarleb's suggestion).

> What more would be involved in making boolean values work in metadata in defaults files?

As long as we're using the same code path in Markdown and defaults file metadata parsing, nothing. That's mostly the point: avoid duplicating https://github.com/jgm/pandoc/blob/f2b337768e95c8903f65e597f7d8cd81938dacc3/src/Text/Pandoc/Readers/Metadata.hs#L84-L88 and associated code.

Note to self, `checkBoolean` itself is already duplicated in https://github.com/jgm/pandoc/blob/34775b4128de2801e4d127064f012501ca18d208/src/Text/Pandoc/App/Opt.hs#L500-L508. Might want to take a closer look.

> Btw it's helpful if all API changes are clearly indicated as such in the commit messages, so I don't forget to include them in the changelog.

Uh... about that. Are there any public-facing API changes? Now that I look at it, `Text.Pandoc.Readers.Metadata` isn't exposed, and `yamlBsToMeta` isn't reexported directly. Other changes are local to modules, and I didn't touch any instances. So I guess there aren't technically any?

jgm:
Sorry, I had thought the module was exposed.  It isn't, you're right.

jgm:
And sorry about my misunderstanding; I now see that Booleans are handled.

The branch has some conflicts that need to be resolved before merging.

lierdakil:
Rebased.

jgm:
Thanks!

PR #6332, opened 2020-05-02
Lua filters: improve error messages for failing filters

tarleb:
Print the Lua error properly instead of displaying their `show` string.


tarleb:
This makes error output better, but issues remain. E.g.:

``` lua
-- test.lua
function Pandoc (doc)
  pandoc.mediabag.fetch('does-not-exist.txt')
end
```

This triggers a `PandocResourceNotFound` error, which is still just `show`n:

```
Error running filter ./test.lua:
PandocResourceNotFound "does-not-exist.txt"
stack traceback:
        [C]: in function 'pandoc.mediabag.fetch'
        ./test.lua:10: in function 'Pandoc'
```

One solution could be to add a new PandocError constructor `PandocErrorStack (NonEmpty PandocError)`.

jgm:
This is good -- the other case you mention doesn't seem bad to me.

PR #6342, opened 2020-05-05
Opting out of a Reveal.js theme with no-theme

jez:
I've written my own Reveal.js theme, and I'm using `revealjs-url=https://revealjs.com`, which means that the only way to include my theme is with Pandoc's existing `css` variable in the `revealjs` template.

The problem is that the `revealjs` template defaults to `black.css` when no `theme` variable is given, which means my custom theme has to go to extra effort to reset the styles introduced by `black.css`. I'd rather have a way to opt out of that.

I think the current default (no `theme` implies `black.css`) is a good default for people, so I've added another variable to opt in to no theme: `no-theme`.

```yaml
---
no-theme: true
---
```

### DIFF
--- a/data/templates/default.revealjs
+++ b/data/templates/default.revealjs
@@ -21,10 +21,13 @@ $endif$
   <style>
     $styles.html()$
   </style>
+$if(no-theme)$
+$else$
 $if(theme)$
   <link rel="stylesheet" href="$revealjs-url$/css/theme/$theme$.css" id="theme">
 $else$
   <link rel="stylesheet" href="$revealjs-url$/css/theme/black.css" id="theme">
+$endif$
 $endif$
 $for(css)$
   <link rel="stylesheet" href="$css$"/>


jgm:
Not sure what I think.  It's always a cost to make the templates more complex (and documentation would have to be added to MANUAL.txt).  Since it's not too hard to override things in your custom css, and it's not too hard to use a custom template, I'm not sure there's a compelling case.

jez:
That's fair, I can definitely understand wanting hold the API surface area low.

While it's definitely not hard to use a custom template, I like sticking to the default template as much as possible because the default templates are continually improved, especially when there are new features of pandoc that can be harnessed in a template. It the same problem with forking an open source project: when the open source project moves forward, it takes effort to backport those changes to the fork.

So I think this is a useful feature, but if you disagree, feel free to close it.

jez:
Actually, going to go ahead and close this. I got around this by writing a Makefile rule that fetches the latest `default.revealjs` and `sed`'s out the `black.css` line:

https://github.com/jez/pandoc-starter/blob/07ebd2b0f792b5192347c7bcbe7e80954a809748/revealjs-solarized/src/Makefile#L30

jgm:
If enough other people have the same need, I'd reconsider this one.

PR #6347, opened 2020-05-07
Update revealjs template with newly available options

jez:
I wanted to use an option and realized that the default pandoc template
didn't pass it through, so I went through and found all the options that
Reveal.js advertises but that Pandoc's template doesn't support.

<https://github.com/hakimel/reveal.js#configuration>

I also noticed that rollingLinks isn't a setting anymore, but it's fine
to keep it in the template because people might be using an old version
of Reveal.js (via `revealjs-url`) that still has this option.

### DIFF
--- a/data/templates/default.revealjs
+++ b/data/templates/default.revealjs
@@ -89,6 +89,20 @@ $if(controls)$
         // Display controls in the bottom right corner
         controls: $controls$,
 $endif$
+$if(controlsTutorial)$
+        // Help the user learn the controls by providing hints, for example by
+        // bouncing the down arrow when they first encounter a vertical slide
+        controlsTutorial: $controlsTutorial$,
+$endif$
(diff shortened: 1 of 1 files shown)


jgm:
Thanks. In retrospect, I should probably have put all these variables under `revealjs` namespaces.  Maybe a bad idea to change that now, though.

jez:
> Thanks. In retrospect, I should probably have put all these variables under `revealjs` namespaces. Maybe a bad idea to change that now, though.

Yeah, it might pose potential problems; there's no real versioning built into the templates. Though it's probably something that can be put off until it manifests concretely as a problem (easier to solve a problem when you can see it!)

By the way, if you have a second I'd love your input on the other PR I have open: #6342 

Thanks again! Huge fan of pandoc.

PR #6356, opened 2020-05-10
Add option to not surrounding display math in <br>

Aver1y:
Motivation is that I would rather handle the spacing with CSS. Also when using pandoc-crossref with the tableEqns options it formats equations with references inside a table and the <br>s mess up spacing.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1558,6 +1558,7 @@ reference-doc: myref.docx
 html-math-method:
   method: mathjax
   url: "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"
+html-math-surround-br: true
 # none, references, or javascript
 email-obfuscation: javascript
 
--- a/src/Text/Pandoc/App/CommandLineOptions.hs
+++ b/src/Text/Pandoc/App/CommandLineOptions.hs
@@ -720,6 +720,12 @@ options =
                       return opt { optHTMLMathMethod = GladTeX }))
                  "" -- "Use gladtex for HTML math"
 
+    , Option "" ["no-html-math-surround-br"]
+                 (NoArg
+                  (\opt ->
+                      return opt { optHTMLMathSurroundBr = False }))
+                 "" -- "Don't surround display math with br in html output."
+
     , Option "" ["trace"]
                  (NoArg
                   (\opt -> return opt { optTrace = True }))
--- a/src/Text/Pandoc/App/Opt.hs
+++ b/src/Text/Pandoc/App/Opt.hs
@@ -96,6 +96,7 @@ data Opt = Opt
     , optSyntaxDefinitions     :: [FilePath]  -- ^ xml syntax defs to load
     , optTopLevelDivision      :: TopLevelDivision -- ^ Type of the top-level divisions
     , optHTMLMathMethod        :: HTMLMathMethod -- ^ Method to print HTML math
+    , optHTMLMathSurroundBr    :: Bool -- ^ Whether to put <br> around display math elements
     , optAbbreviations         :: Maybe FilePath -- ^ Path to abbrevs file
     , optReferenceDoc          :: Maybe FilePath -- ^ Path of reference doc
     , optEpubSubdirectory      :: String -- ^ EPUB subdir in OCF container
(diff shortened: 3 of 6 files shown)


jgm:
I'm not too keen on the idea of adding an option for this.  It's an API change and each addition like this increases the maintenance surface. There are ways to work around this.

However, it might be interesting to consider a shift in our default display math rendering. Instead of the br we could try to do everything with CSS.

jgm:
I'd be interested in the CSS you use for formatting display math.

Aver1y:
Re CSS: Just something along these lines:

```css
.display.math {
  display: block;
  text-align: center;
  margin: 0.5rem 0;
}
```

Actually we could just use `div` or `p` instead of `span` to get the `display: block`. That way we could get it on it's own line without any CSS or `br`. While it doesn't seem to do anything, nesting `p`s seems kind of ugly to me, so I would be more in favor of `div`.

I also noticed that if you put display math in it's own paragraph (i.e. putting empty lines around it) it will result in uneven spacing as the leading `br` in the `p` will be rendered, but the trailing one will be ignored. So this results in uneven spacing:

```
Test

$$ f(x) = x +1 $$

Test
```

but this doesn't:

```
Test
$$ f(x) = x +1 $$
Test
```

jgm:
> Actually we could just use div or p instead of span to get the display: block.

No, I'm pretty sure it's not valid HTML5 to put a div or another p inside a p element.
But maybe using the CSS you give would be a better solution than adding the brs.

Aver1y:
Oh, yes you're right. However, for the same reason I think that display math shouldn't be inside a <p> with other text at all. <p> is supposed to contain phrasing content, i.e. inline content not block content. Using an html tag that is supposed to be phrasing content and then turning it into a block using CSS is not much better than just using a <div> in my opinion.

So I think display math should always be parsed as it's own block even without empty lines around it. Fenced code blocks already do this. It may be ideal if this would also be reflected in the AST by adding a constructor for display math to `Block`, however if that is not worth the breaking change we can just put it in it's own `p`. Though, with the breaking change in the table representation coming up this may be a good time to throw such a change in as well?

jgm:
Display math can occur inside paragraphs in LaTeX; this is the main reason we do it that way.

Aver1y:
OK, so with the CSS option we would just add the proposed CSS to `styles.html` and remove the `br`s?

jgm:
Yes, that would be the idea.

Aver1y:
A problem with doing it this way is that CSS isn't meant to describe the semantics and changing dsiplay to block is kind of doing that. A real problem that comes from this is that Firefox's reader view completely ignores CSS.

Also I found this:

> The top-level element in MathML is <math>. Every valid MathML instance must be wrapped in <math> tags. In addition you must not nest a second <math> element in another, but you can have an arbitrary number of other child elements in it.

https://developer.mozilla.org/en-US/docs/Web/MathML/Element/math

So shouldn't we put math inside `<math>` instead of inside `<span>`? `<math>` also has an attribute `display` which can be `block` or `inline`. It used to have an attribute `mode` which can be `display` or `inline`, but that was deprecated in favor of `display`. We still have to style it though, with the CSS I proposed above, because that isn't done by default :/. Also while reader view displays `<math display="block">` very nicely `<math display="inline">` still get's the huge top and bottom margin of `display="block"`. That seems to be a bug in Firefox though.

Also interestingly [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content), i.e. the content allowed inside of `<p>`, includes `<math>` and it doesn't say anything about this depending on the `display` attribute. So maybe math block content is allowed inside of `<p>` as an exception or maybe it's just an oversight.

jgm:
> So shouldn't we put math inside `<math>` instead of inside `<span>`? 

Yes. And we do, when `--mathml` is used to produce mathml output.  (That's not the default.)

jgm:
> So maybe math block content is allowed inside of `<p>` as an exception or maybe it's just an oversight.

I think it's intentional. As I noted, display math is considered part of a paragraph in LaTeX and several other formats of which I'm aware.

jgm:
> A real problem that comes from this is that Firefox's reader view completely ignores CSS.

Not sure how serious this is: it would mean that in reader view, display math is rendered inline, but that's not a huge problem.

Aver1y:
> Yes. And we do, when --mathml is used to produce mathml output. (That's not the default.)

Oh, I was so sure, I saw `mo`s and `mi`s in the pandoc output, but I must've just seen them in the HTML reference and then my brain confused things.

Anyway, yes after some thought I now also feel it's reasonable to allow display math inside of paragraphs.

> Not sure how serious this is: it would mean that in reader view, display math is rendered inline, but that's not a huge problem.

Yeah, if your math is basic enough that you don't need proper math markup it's probably not a big deal...

Also the reason for the weird line height of <math display="inline"> was actually an outdated version of the latinmodern-math font on my computer.

Aver1y:
I'll make a PR for the CSS solution.

jgm:
Are you still planning to make a PR @Aver1y

torresjrjr:
I'd like to say thank you. I only recently had a couple thoughts to share, but I think I'll just wait for that next release. I've had them same trouble with uneven spacing and aligning math SVGs too.

PR #6364, opened 2020-05-13
Correct license listed in cabal

Aver1y:
`GPL-2` is interpreted as `GPL-2.0-only` only and is displayed as such on Hackage, however according to all the headers in the actual Haskell files the license is GPLv2 or later, i.e. `GPL-2.0-or-later` (https://spdx.org/licenses/).

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -2,7 +2,7 @@ name:            pandoc
 version:         2.9.2.1
 cabal-version:   2.0
 build-type:      Simple
-license:         GPL-2
+license:         GPL-2.0-or-later
 license-file:    COPYING.md
 copyright:       (c) 2006-2020 John MacFarlane
 author:          John MacFarlane <jgm@berkeley.edu>


tarleb:
The [GPL-2.0](https://spdx.org/licenses/GPL-2.0.html) identifier is also deprecated as a SPDX identifier. This should probably be changed in all module headers as well.

Aver1y:
In the module headers it doesn't use SPDX identifier it says "GNU GPL, version 2 or above".

tarleb:
Oh, right.

Thanks for bringing this up.

jgm:
SPDX identifiers are only recognized by Cabal 3, I think.
If that's right, then this change may cause problems with earlier versions.  But I may be mistaken.

jgm:
(Earlier cabal versions had a fixed list of valid identifiers.)

ickc:
Just trying to clarify the situation: is it always GPLv2+ in this repo, although it might has been mislabeled somewhere?

jgm:
I think it always is GPLv2+ -- it's just that earlier versions of cabal didn't have anything to indicate this, so we used GPL-2 in the cabal file. My worry is that with this change pandoc will break with earlier cabal versions, though maybe this has been tested?

jgm:
I guess it worked on all the versions of cabal we do CI with, so no worries after all.

PR #6370, opened 2020-05-16
LaTeX Writer: use addchap instead of chapter* for bookscr

mmiszy:
Using chapter* in bookscr makes the running head incorrect in the chapter (the previous running head is used). KomaScript provides a special function for this usecase: `addchap`. It generates a new chapter, adds proper running head and also generates a toc item, so `addcontentsline` can be skipped in this case.

# Todo
- [ ] Add tests? (Sorry, I tried but I couldn't find any tests which would modify documentclass or state)

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -72,6 +72,7 @@ data WriterState =
               , stEmptyLine     :: Bool          -- true if no content on line
               , stHasCslRefs    :: Bool          -- has a Div with class refs
               , stCslHangingIndent :: Bool       -- use hanging indent for bib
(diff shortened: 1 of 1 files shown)


jgm:
I'm unsure about going down the path of customizing output for various different classes -- this could add a lot of complexity (which is always a cost).  Have you considered the option of simply redefining the `chapter*` macro?

mmiszy:
@jgm thank you for the idea, I didn't think of that 🤦‍♀️ 

This override works:
```latex
\RenewDocumentCommand{\chapter}{s m}{
  \IfBooleanTF#1
    {\oldchapter*{#2}\markboth{#2}{#2}}
    {\oldchapter{#2}}
}
```
The missing piece was actually `\markboth` so maybe you should consider adding it for unnumbered chapters?

PR #6372, opened 2020-05-17
Use CSS in favor of <br> for display math

Aver1y:
Sorry for the delay

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -318,6 +318,10 @@ pandocToHtml opts (Pandoc meta blocks) = do
                                      defField "mathjaxurl"
                                        (T.takeWhile (/='?') u)
                         _         -> defField "mathjax" False) $
+                  (case writerHTMLMathMethod opts of
+                        PlainMath -> defField "displaymath-css" True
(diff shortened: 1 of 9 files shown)


src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -1108,14 +1112,11 @@ inlineToHtml opts inline = do
...
-                            ! A.title (toValue str)
-              let brtag = if html5 then H5.br else H.br
-              return $ case t of
-                        InlineMath  -> m
-                        DisplayMath -> brtag >> m >> brtag
+              return $ imtag ! A.style "vertical-align:middle"
+                             ! A.src (toValue $ url <> T.pack (urlEncode (T.unpack $ s <> str)))
+                             ! A.alt (toValue str)
+                             ! A.title (toValue str)
+                             ! A.class_ mathClass
```
Aver1y, line 1119: This may be somewhat easy to miss in this diff, but the `A.class_ mathClass` here is new.

jgm:
This looks good to me!  But it says there are conflicts with the base branch. Can you look into this?

Aver1y:
Oh, yes some of the tests need fixing.

jgm:
Thank you!

Issue #6026, opened 2020-01-03
More flexible citation key syntax


dhimmel:
From the Pandoc manual:

> Citations go inside square brackets and are separated by semicolons. Each citation must have a key, composed of '@' + the citation identifier from the database, and may optionally have a prefix, a locator, and a suffix. The citation key must begin with a letter, digit, or `_`, and may contain alphanumerics, `_`, and internal punctuation characters (`:.#$%&-+?<>~/`)

The citation key syntax is limited ([see as a regex](https://regex101.com/r/mXZyY2/latest)), preventing use of a variety of types of citekeys that various users would like:

- https://github.com/jgm/pandoc-citeproc/issues/308
- https://github.com/jgm/pandoc/issues/5350
- https://github.com/jgm/pandoc/issues/1161
- https://github.com/manubot/rootstock/issues/2

One user case that I'm interested in [for Manubot](https://github.com/manubot/rootstock/blob/37b33788b3456cf8ddcd61aea1689070c97012fc/USAGE.md#citations) is citation-by-persistent-identifier where the citekey is an actual identifier. Oftentimes however, identifiers contain characters forbidden from Pandoc's citation key syntax. For example, we'd like to be able to include citekeys like:

```
Citekey with parentheses @doi:10.1016/S0022-2836(05)80360-2
Citekey with closing slash @https://www.google.com/
Citekey with equal sign @https://openreview.net/forum?id=HkwoSDPgg
```

My intended use case would just require more flexible citekeys for markdown input. We would likely  use a custom filter to generate new citekeys for the output. However, it also seems from the issues above that some users would like more flexibility for citekeys across the board.

In https://github.com/jgm/pandoc-citeproc/issues/308#issuecomment-451230632, @jgm proposed a syntax like:

```
@{whatever you want here}
```

mb21:
Not sure about your usecase, but potentially you could also just use a markdown link or span with custom attributes and use a filter to go from there...

btw. #813 might be related?

dhimmel:
> potentially you could also just use a markdown link

The [`pandoc-url2cite`](https://github.com/phiresky/pandoc-url2cite) filter by @phiresky has a nice syntax for defining citekey aliases that can contain forbidden characters:

```markdown
Citekey with closing slash @google

[@google]: https://www.google.com/
```

I think this is a good workaround, but I still see a benefit in being able to skip the link reference / alias step altogether and do something like:

```
Citekey with closing slash @{https://www.google.com/}
```

[...]

PR #6373, opened 2020-05-17
More flexible citation key syntax #6026

Aver1y:
This implements the `{citekey}` syntax and also allows escaping using `\` both `}` and `\` inside `{}` ~~and any illegal character in the existing syntax~~.

This does not address all of #6026. I still want to implement the `[@citekey]: citekey'` syntax.

#### TODO:

- [ ] Update documentation to reflect new behaviour of citekey parser regarding newlines.
- [ ] Add tests?

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -4825,7 +4825,7 @@ Each citation must have a key, composed of '@' + the citation
 identifier from the database, and may optionally have a prefix,
 a locator, and a suffix.  The citation key must begin with a letter, digit,
 or `_`, and may contain alphanumerics, `_`, and internal punctuation
-characters (`:.#$%&-+?<>~/`).  If you want to use a citation key that do not
+characters (`:.#$%&-+?<>~/`).  If you want to use a citation key that does not
 conform to these constraints, enclose the key in curly braces (`{` and `}`).
 Inside the curly braces, the characters `}` and `\\` must be backslash-escaped.
 Here are some examples:
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -4825,13 +4825,20 @@ Each citation must have a key, composed of '@' + the citation
 identifier from the database, and may optionally have a prefix,
 a locator, and a suffix.  The citation key must begin with a letter, digit,
 or `_`, and may contain alphanumerics, `_`, and internal punctuation
-characters (`:.#$%&-+?<>~/`).  Here are some examples:
+characters (`:.#$%&-+?<>~/`).  Alternatively arbitrary citation keys
+can be given inbetween `{` and `}`. Inside them `}` and `\` can be escaped
+using `\`. Here are some examples:
 
     Blah blah [see @doe99, pp. 33-35; also @smith04, chap. 1].
 
     Blah blah [@doe99, pp. 33-35, 38-39 and *passim*].
 
     Blah blah [@smith04; @doe99].
+
+You can also use citation keys directly without brackets:
+
+    Blah blah @smith04.
+    @{A more complex cite key}
 
 `pandoc-citeproc` detects locator terms in the [CSL locale files].
 Either abbreviated or unabbreviated forms are accepted. In the `en-US`
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -4825,9 +4825,10 @@ Each citation must have a key, composed of '@' + the citation
 identifier from the database, and may optionally have a prefix,
 a locator, and a suffix.  The citation key must begin with a letter, digit,
 or `_`, and may contain alphanumerics, `_`, and internal punctuation
-characters (`:.#$%&-+?<>~/`).  Alternatively arbitrary citation keys
-can be given inbetween `{` and `}`. Inside them `}` and `\` can be escaped
-using `\`. Here are some examples:
+characters (`:.#$%&-+?<>~/`).  If you want to use a citation key that do not
+conform to these constraints, enclose the key in curly braces (`{` and `}`).
+Inside the curly braces, the characters `}` and `\\` must be backslash-escaped.
+Here are some examples:
 
     Blah blah [see @doe99, pp. 33-35; also @smith04, chap. 1].
 
@@ -4835,10 +4836,7 @@ using `\`. Here are some examples:
 
     Blah blah [@smith04; @doe99].
 
-You can also use citation keys directly without brackets:
-
-    Blah blah @smith04.
-    @{A more complex cite key}
+    Blah blah [@{more complex citekey}]
 
 `pandoc-citeproc` detects locator terms in the [CSL locale files].
 Either abbreviated or unabbreviated forms are accepted. In the `en-US`
--- a/doc/org.md
+++ b/doc/org.md
@@ -127,9 +127,10 @@ the citation identifier from the database, and may optionally
 have a prefix, a locator, and a suffix. The citation key must
 begin with a letter, digit, or `_`, and may contain
 alphanumerics, `_`, and internal punctuation characters
-(`:.#$%&-+?<>~/`). Alternatively arbitrary citation keys can be
-given inbetween `{` and `}`. Inside them `}` and `\` can be
-escaped using `\`.
+(`:.#$%&-+?<>~/`). If you want to use a citation key that do not
+conform to these constraints, enclose the key in curly braces
+(`{` and `}`). Inside the curly braces, the characters `}` and
+`\\` must be backslash-escaped.
 
 ### Simple citation
 
@@ -139,11 +140,10 @@ ID prefixed by '@'.
 
 Example:
 
-    @citekey
-    @{this is a citekey}
     [prefix @citekey suffix]
     [see @doe2000 pp. 23-42]
     [@doe2000 p. 5; to a lesser extend @doe2005]
+    [@{more complex citekey}]
 
 
 LaTeX-Syntax
--- a/doc/org.md
+++ b/doc/org.md
@@ -127,7 +127,9 @@ the citation identifier from the database, and may optionally
 have a prefix, a locator, and a suffix. The citation key must
 begin with a letter, digit, or `_`, and may contain
 alphanumerics, `_`, and internal punctuation characters
-(`:.#$%&-+?<>~/`). Here are some examples:
+(`:.#$%&-+?<>~/`). Alternatively arbitrary citation keys can be
+given inbetween `{` and `}`. Inside them `}` and `\` can be
+escaped using `\`.
 
 ### Simple citation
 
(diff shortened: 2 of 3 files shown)


jgm:
I'm not sure about the idea of allowing escapes in regular citations (without the `{...}`).
This changes current behavior in a way that might have some bad consequences.
For example, currently `@doe\-1999` will create a citation to `doe` followed by the string `-1999`, whereas `@doe-1999` will create a citation to `doe-1999`.  I think that if we have the braces syntax, we don't need the escapes. And in braces, only braces and `\\` need to be escapable.  That keeps things simple.

Aver1y:
Changed.

MANUAL.txt
```diff
@@ -4825,14 +4825,21 @@ Each citation must have a key, composed of '@' + the citation
...
 
     Blah blah [@doe99, pp. 33-35, 38-39 and *passim*].
 
     Blah blah [@smith04; @doe99].
 
+You can also use citation keys directly without brackets:
+
+    Blah blah @smith04.
+    @{A more complex cite key}
+
```
jgm, line 4842: This part is a bit misleading.  Without brackets you get an "author-in-text" citation form, so it's not simply an alternative syntax for the same thing.  And that is already covered in the manual a bit below this point, under "in-text" citations. So I think you can just remove these lines.

doc/org.md
```diff
@@ -127,7 +127,9 @@ the citation identifier from the database, and may optionally
 have a prefix, a locator, and a suffix. The citation key must
 begin with a letter, digit, or `_`, and may contain
 alphanumerics, `_`, and internal punctuation characters
-(`:.#$%&-+?<>~/`). Here are some examples:
+(`:.#$%&-+?<>~/`). Alternatively arbitrary citation keys can be
+given inbetween `{` and `}`. Inside them `}` and `\` can be
```
jgm, line 131: inbetween should be two words "in between" -- or better, just "between"

jgm: This doesn't really bring out the message that the braces are the thing to use if you have special characters not otherwise allowed.  I'd suggest something like:

```
If you want to use a citation key that do not conform to these constraints,
enclose the key in curly braces (`{` and `}`).  Inside the curly braces,
the characters `}` and `\\` must be backslash-escaped.
```

Should `{` also be escapable in braces?

src/Text/Pandoc/Parsing.hs
```diff
@@ -1468,12 +1468,14 @@ citeKey = try $ do
...
-                 try (oneOf ":/" <* lookAhead (char '/'))
-  let key = firstChar:rest
+  key <-
+    do firstChar <- alphaNum <|> char '_' <|> char '*' -- @* for wildcard in nocite
+       let regchar = satisfy (\c -> isAlphaNum c || c == '_')
+       let internal p = try $ p <* lookAhead regchar
+       rest <- many $ regchar <|> internal (oneOf ":.#$%&-+?<>~/") <|>
+                      try (oneOf ":/" <* lookAhead (char '/'))
+       pure $ firstChar:rest
+    <|> char '{' *> many (noneOf "}\\" <|> char '\\' *> anyChar) <* char '}'
```
jgm, line 1478: I wonder if this is too permissive. What about something like
```
@{hi there

- a list
- and another list item

another paragraph.}
```
Perhaps disallow newlines?
Do we want to allow spaces?

doc/org.md
Aver1y: > Should { also be escapable in braces?

Everything is escapable. It's just that you only have to do it for `\` and `{`.

src/Text/Pandoc/Parsing.hs
Aver1y: Yes, that's a good point, could be confusing if one forgets the `}`. I think spaces are fine though.

Aver1y: Another point here is that, if we disallow newlines they can still be included by escaping them, however I think it may be more useful to ignore escaped newlines, so that long identifiers can be split up over multiple lines. Kind of like Haskell Strings. In that case I would also want to skip (unescaped) whitespace after the escaped newline, to allow lining things up nicely.

```
- Item 1, @{a long \
  citekey}
- Item 2
```

Aver1y: Re do we want to allow spaces: I think citekeys like `@{Hirshburg, 2016}` are useful.

Aver1y: Also a related question, because I don't have the big overview of the parser. Is parsing ever supposed to fail? If we encounter an unescaped newline while parsing a citekey in braces, does parsing fail or should it just backtrack and treat the citekey as normal text and do we have to backtrack inside of the citekey parser or does the calling parser do that for us? Also right the citekey parser results in a committed failure if `\` is followed by eof. Is that correct behaviour?

jgm: the whole citekey parser is under a `try`, so if anything fails inside it the whole thing will fail without consuming input. (and then, yes, the enclosing parser will likely parse this as regular text.)

jgm: As for newlines, I think it would make sense to allow newlines only if not followed by a blankline (there's a parser for this).  And not to allow newlines to be backslash-escaped; that would create confusion with the existing markdown syntax for hard breaks.

Aver1y: And these newlines should be interpreted as newlines or ignored? And if they should be ignored should leading whitespace after newlines also be ignored?

jgm: The main consideration is that, if spaces are allowed, then in a hard-wrapped document one of the spacse might be turned into a newline + leading spaces. So, yes, I think it would be reasonable to parse a newline + leading spaces as a space for these purposes.  Not sure about backslash newlines.  Maybe it's cleanest just to allow them to be parsed as newlines.

Aver1y: Ah, yes, that's a good consideration. I agree that parsing escaped newlines as newlines is the cleanest, however I'm not sure if it is that useful. On the other hand if we parse an unescaped newline as a space we still don't have a way to break up long citekeys without spaces into over multiple lines. We could instead ignore escaped newlines. I suppose the question is whether it's better to have useless behavior or non-intuitive, but also probably not harmful, rarely useful behavior.

Aver1y: Also another thing would be to parse `\n` as a newline. That way it would actually be possible to use arbitrary citation keys. I don't really think anyone has citation keys with newlines inside them, but who knows.

jgm: > we still don't have a way to break up long citekeys without spaces into over multiple lines

I think that's okay.  It's an issue that arises in many other contexts too (e.g. there's no way to split a long URL over multiple lines in a link).

As for newlines in citation keys: I'd prefer not to make that possible.

MANUAL.txt
```diff
@@ -4825,14 +4825,19 @@ Each citation must have a key, composed of '@' + the citation
 identifier from the database, and may optionally have a prefix,
 a locator, and a suffix.  The citation key must begin with a letter, digit,
 or `_`, and may contain alphanumerics, `_`, and internal punctuation
-characters (`:.#$%&-+?<>~/`).  Here are some examples:
+characters (`:.#$%&-+?<>~/`).  If you want to use a citation key that do not
```
jgm, line 4828: "do not" should be "does not"

src/Text/Pandoc/Parsing.hs
```diff
@@ -1468,12 +1468,20 @@ citeKey = try $ do
...
+       rest <- many $ regchar <|> internal (oneOf ":.#$%&-+?<>~/") <|>
+                      try (oneOf ":/" <* lookAhead (char '/'))
+       pure $ firstChar:rest
+    <|> char '{' *>
+        many `id` choice
+          [ noneOf "}\n\\"
+          , ' ' <$ newline <* skipSpaces <* notFollowedBy newline
+          , char '\\' *> anyChar
+          ]
+        <* char '}'
```
Aver1y, line 1484: > As for newlines, I think it would make sense to allow newlines only if not followed by a blankline (there's a parser for this

I think you mean `spnl`? That one is in `Readers/Markdown.hs`, so not available here unless we move it.

Aver1y: I wondered if it would be more consistent to parse any occurrence of multiple spaces as a single space. The way it is now I would think it may be especially confusing with trailing spaces? If one wants multiple spaces they can still escape them.

jgm: Yes, we could collapse consecutive spaces, that might make sense.

brainchild0:
I suggest an alternative approach for handling squiggly braces within a key, which is to parse nested brace or parenthesis pairs. Thus `@{a{b}c(d)e[f]g}` would correspond to key `a{b}c(d)e[f]g`, because only the final brace matches the first. Such rules would handle all cases but a few pathological ones, which I believe are currently only hypothetical. It also would avoid user difficulty related to escaping sequences, which may be long and often would be inserted by cut-and-paste.

If a later need arises to support the remaining cases, then I would suggest doing so through a slightly more explicit form, such as `@{"pathological\}case"}`. Perhaps this support might be postponed.

Aver1y:
@brainchild0 I see the point for matching braces. It is probably more common for braces to appear paired than on their own. For other parenthesis I don't see the benefit.

brainchild0:
Yes, you may be right that in this case matching other braces is unnecessary, as long as they are protected from the processing that would occur in a regular context. Thus the first `]` in `[@{protect]me}]` should probably not close the `[`. However, fully applying the same rules for all brace types creates a  convention that robustly generalizes to other scenarios, which later may be useful considering the many separate meanings given to only a few brace types during processing depending on context.

dhimmel:
An additional benefit of this pull request might be that it allows a way to terminate a citekey without space or punctuation. Specifically, `what will @{happen}here?` Am I correct that there prior to this PR there is no way to have a citekey `happen` immediately followed by `here`? And that this PR enables this?

Where this would come in handy is for filter like `pandoc-fignos` by @tomduck. Currently, to have a reference like "Figure 1A" with pandoc-fignos, the user has to write `Figure {@fig:id}A`. I think this PR would enable `Figure @{fig:id}A` without any special support by the filter.

dhimmel:
Looks like https://github.com/jgm/pandoc/commit/3f09f53459b877f53072efbf57dec21fa37280b5 added this feature.

jgm:
Oh, I'm sorry -- I'd completely forgotten about this PR!
Note that in my implementation, I don't allow backslash escapes at all.  Braces are allowed, but only if matched.

PR #6389, opened 2020-05-21
Use <kbd> tag

salim-b:
Native support for `<kbd>` was added in [Pandoc 2.8](https://github.com/jgm/pandoc/releases/tag/2.8).


jgm:
I think I'll pass on this -- the manual needs to render well in a variety of formats, including PDF and man page, and we'll get better results with the regular code span.

PR #6390, opened 2020-05-21
update template for reveal.js 4.0.0

salim-b:
cf.
- https://github.com/hakimel/reveal.js/releases/tag/4.0.0
- https://revealjs.com/upgrading/

see also https://github.com/jgm/pandoc-templates/pull/13


jgm:
Thanks!

PR #6403, opened 2020-05-26
Break up large module.

fisx:
Hi there,

I have no knowledge of the code base, and did not spend more than 10 minutes on this, so I'm not offended if you close it without comment, but @tarleb tells me that large modules are an issue in pandoc, and I wanted to know how hard it is to break them up.

In this PR, i make a module that collects all Latex reader top-level definitions that look like tables.  :)

If you like, I'll fix the unused-import errors, and whatever else the CI will complain about.  And perhaps there is a less misleading name for the new module.

cheers!


fisx:
hum, this doesn't look like it's going to go anywhere within my attention span.  sorry for the noise!

Issue #6417, opened 2020-06-02
LaTeX: ranges of SI units disappear during conversion


efharkin:
# Description of bug and steps to reproduce 

As of the current master (7c6dbd37eb3e785f875e0030d723e422db72d453) ranges of SI units provided by `\SIRange{firstvalue}{secondvalue}{unit}` of siunitx are dropped during parsing of LaTeX files. This leads to funny conversions like the following:

Original LaTeX `test_sirange.tex`
```latex
Something important is happening \SIRange{100}{200}{\ms} after the event.
```

Expected output in markdown
```markdown
Something important is happening 100 ms to 200 ms after the event.
```

Actual output of `pandoc -so test_sirange.md test_sirange.tex`
```markdown
Something important is happening after the event.
```

## Why is this a bug?

The manual states that pandoc doesn't aim for perfect document conversion, and that in particular formatting details may not be preserved across formats. However, by dropping `\SIRange`, pandoc's output can become semantically different from the original LaTeX (ie phrases might not make sense or even take on a different meaning after conversion), so this isn't a formatting detail.

# Proposed fix

siunitx's `\SIRange` is highly configurable. Depending on the options used, `\SIRange{100}{200}{\ms}` might be rendered as 
- 100 ms to 200 ms
- 100 ms &ndash; 200 ms
- 100 &ndash; 200 ms
- etc

In the spirit of prioritizing semantics over formatting details, I think it would be good enough for pandoc to support only the first option. I've implemented this approach and will submit a PR shortly.

# Related issues and PRs

How pandoc should parse siunitx commands has also been discussed in issue https://github.com/jgm/pandoc/issues/3587. I'm opening a new issue because that issue (and the associated PR https://github.com/jgm/pandoc/pull/3823) have been focused on the `\num` command and are a bit stale.

efharkin:
Implemented a fix here https://github.com/jgm/pandoc/pull/6418.

tarleb:
Closing, as the fix was merged last year. (Thanks!)

PR #6418, opened 2020-06-02
Minimal support for SIRange in LaTeX reader

efharkin:
This patch adds minimal support for `\SIRange{firstnumber}{secondnumber}{unit}` (provided by siunitx) consistent with existing support for `\SI{value}{unit}`. Implementation is an extension of the existing `dosiunitx` function in the LaTeX reader. Added unit tests for SIRange in `test/command/3587.md`. 

Code passes all new and existing unit tests and compiles without warnings*.

This PR implements a fix for issue https://github.com/jgm/pandoc/issues/6417.

# Example

LaTeX input
```latex
Something important happens \SIRange{100}{200}{\ms} after the event.
```

Markdown output
```markdown
Something important happens 100 ms to 200 ms after the event.
```

----------

*As far as I can tell... I don't see any warning messages when I compile normally. I'm completely new to haskell and `cabal build -Wall` and `cabal test -Wall` both raise errors about unrecognized options, so it's possible there's something I've missed. If there's something different I should be doing please let me know.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -258,6 +258,29 @@ dosiunitx = do
                       emptyOr160 unit,
                       unit]
 
+-- converts e.g. \SIRange{100}{200}{\ms} to "100 ms to 200 ms"
+doSIRange :: PandocMonad m => LP m Inlines
+doSIRange = do
+  skipopts
+  startvalue <- tok
+  startvalueprefix <- option "" $ bracketed tok
+  stopvalue <- tok
+  stopvalueprefix <- option "" $ bracketed tok
+  unit <- grouped (mconcat <$> many1 siUnit) <|> siUnit <|> tok
+  let emptyOr160 "" = ""
+      emptyOr160 _  = "\160"
+  return . mconcat $ [startvalueprefix,
+                      emptyOr160 startvalueprefix,
+                      startvalue,
+                      emptyOr160 unit,
+                      unit,
+                      " to ",
+                      stopvalueprefix,
+                      emptyOr160 stopvalueprefix,
+                      stopvalue,
+                      emptyOr160 unit,
+                      unit]
+
 siUnit :: PandocMonad m => LP m Inlines
 siUnit = do
   Tok _ (CtrlSeq name) _ <- anyControlSeq
@@ -1140,6 +1163,7 @@ inlineCommands = M.union inlineLanguageCommands $ M.fromList
   , ("acsp", doAcronymPlural "abbrv")
   -- siuntix
   , ("SI", dosiunitx)
+  , ("SIRange", doSIRange)
   -- hyphenat
   , ("bshyp", lit "\\\173")
   , ("fshyp", lit "/\173")
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -274,7 +274,7 @@ doSIRange = do
                       startvalue,
                       emptyOr160 unit,
                       unit,
-                      " to ",
+                      " \8211 ", -- An en-dash
                       stopvalueprefix,
                       emptyOr160 stopvalueprefix,
                       stopvalue,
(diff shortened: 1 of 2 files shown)


jgm:
Thanks, this generally looks good!
One comment: people use pandoc in many different languages, so we avoid English words like "to".
Two options here:

- use an en-dash instead
- use our relatively new "Translations" to fetch a localized version of "to" (see `translateTerm` in `T.P.Class.PandocMonad`; `translateTerm To` should do it).  Perhaps fall back to en-dash if not found?

efharkin:
Thanks for the input @jgm! I tried to implement the Translations fix but got a bit stuck. I tried inserting `translateTerm Translations.To` into my code, but it looks like I have to convert the type from `Text` to `Inlines`. I tried using `text $ translateTerm Translations.To` for the conversion, but my compiler is complaining that the argument to `text` is `m0 Text` rather than the expected `Text`. Any chance you could point me in the right direction to fixing this? 

If it's too much trouble, I might just fall back to the en-dash fix myself.

mb21:
> it looks like I have to convert the type from Text to Inlines

Hoogle is great: https://www.stackage.org/lts-15.15/hoogle?q=Text+-%3E+Inlines

ah, I see you already found `text`.... so the `m0` refers to some monad, in this case the `PandocMonad`. To get the value "temporarily out of the monad" when in a `do` block you can use `<-`, like:

    do
      myTxt <- translateTerm Translations.To
      let myInline = text myTxt

efharkin:
Thanks everyone for your feedback! I ended up implementing both versions in separate commits. 

- 116ec3fc00c930c660f39b7334cc99a7492af82c uses an en-dash and passes all unit tests.
- bdc370c3f6b1a00dd6409878b313e7d6a35ed686 uses a lower-case locale-based "to" and falls back to an en-dash if one can't be found.

The locale-based version might be a bit more brittle for a couple of reasons

1. The result of `translateTerm Translations.To` is "To", which I had to coerce to lower case. This is correct in English, but might not be in other languages.
2. Right now I only have unit tests for English because I couldn't figure out how to set the locale in the test file. (A quick search for tests mentioning `translateTerm` came up empty.)
3. Fallback to an en-dash is triggered by the result of `translateTerm` being an empty string when translation fails. I'm not sure how to mock a failed translation, so right now the fallback behaviour of `doSIRange` is untested. Unfortunately that means that if the return value of `translateTerm` for a failed translation changes, `doSIRange` fallback will break and there won't be any unit tests to detect the regression.

Let me know if you have any ideas for how to address these issues, or whether you think it's ok as-is. Otherwise, I can always revert to the en-dash version.

efharkin:
@jgm just wanted to flag that I implemented your suggestion in bdc370c3f6b1a00dd6409878b313e7d6a35ed686:

```haskell
  localTo <- translateTerm Translations.To
  let separator = (if localTo /= ""
                     then text $ T.toLower localTo 
                     else "\8211")
```

This means that in an English locale we get the following translation, which is covered by unit tests.

```latex
Something important happens \SIRange{100}{200}{\ms} after the event.
```
```markdown
Something important happens 100 ms to 200 ms after the event.
```

If the translation can't be fetched, we get `\8211`, which is an en-dash. This branch isn't covered by unit tests.

jgm:
So I guess you're asking which branch should be preferred?
I'm not too confident that the "To" translations will always be appropriate in this context.  I think I'd feel more solid just using en-dash all the time, and that would give acceptable results.

efharkin:
Thanks for the comments @jgm! I think you're right that the en-dash is more robust, so I reverted to that version. Good catch about the spaces around the separator. I've removed them to be consistent with the native behaviour of SIRange. Let me know if there's anything else I can do to improve this PR!

jgm:
Excellent, thanks for contributing to pandoc!

pauliacomi:
Hi @efharkin. Thanks a lot for going through the trouble of implementing this addition.

I have had some trouble making it work for my conversions since the `siunitx` syntax is `\SIrange`, rather than `\SIRange`. As such, the latest pandoc version is still ignoring the command. Am I missing something?

efharkin:
Oops, thanks for catching that @pauliacomi! That's what I get for implementing that feature on a different machine than my usual latex setup. Will submit a fix later today, just triple-checking that everything works first.

PR #6422, opened 2020-06-03
doc: fix FreeBSD port link

mo-pyy:
The FreeBSD port was renamed from pandoc to hs-pandoc in 2010.
The old pandoc port is still at version 1.5.1.1


mo-pyy:
Hi,
so I just discovered that on the website there is the same issue (see [here](https://github.com/jgm/pandoc-website/blob/master/INSTALL.md))
Should I open another PR there?

jgm:
Thanks. This file is used to generate the web page, so no need for a separate PR.

PR #6439, opened 2020-06-07
Make it possible to compile using Stack on NixOS

mboes:
On NixOS, it is necessary to compile with `stack --nix`. It is
furthermore necessary to provide zlib headers when `--nix` is enabled.


jgm:
Thanks

PR #6442, opened 2020-06-07
Docbook reader: implement <procedure>

mboes:
A `<procedure>` contains a sequence of `<step>`'s, or `<substeps>`
that themselves contain `<step>`'s. Procedures are mapped to bullet
lists.


jgm:
Thanks for all these improvements!

Issue #6443, opened 2020-06-09
Make polyglossia options list-aware in LaTeX template


frederik-elwert:
In the latex template, most package options are list-aware, i.e., when a package supports multiple options, they can be expressed as a list in the YAML metadata. However, for some reasons, this is not true for the polyglossia(-lang|-otherlangs) options. Thus, instead of writing

```yaml
lang: de
polyglossia-lang:
    name: german
    options:
        - spelling=new
        - babelshorthands=true
```

one currently has to write

```yaml
lang: de
polyglossia-lang:
    name: german
    options: spelling=new,babelshorthands=true
```

Since this makes the interface for LaTeX package options incoherent, I think this should be aligned.

cderv:
FYI Just passing by here looking for some information, and it seems this issue has been fixed by #6444 so probably can be closed.

tarleb:
Thanks! What's more is that pandoc now uses babel instead of polyglossia for all LaTeX engines.

PR #6444, opened 2020-06-09
Make polyglossia package options list-aware in the LaTeX template.

frederik-elwert:
Suggested fix for #6443.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -328,9 +328,9 @@ $if(lang)$
 \ifxetex
   % Load polyglossia as late as possible: uses bidi with RTL langages (e.g. Hebrew, Arabic)
   \usepackage{polyglossia}
-  \setmainlanguage[$polyglossia-lang.options$]{$polyglossia-lang.name$}
+  \setmainlanguage[$for(polyglossia-lang.options)$$polyglossia-lang.options$$sep$,$endfor$]{$polyglossia-lang.name$}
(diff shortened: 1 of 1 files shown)


data/templates/default.latex
```diff
@@ -328,9 +328,9 @@ $if(lang)$
 \ifxetex
   % Load polyglossia as late as possible: uses bidi with RTL langages (e.g. Hebrew, Arabic)
   \usepackage{polyglossia}
-  \setmainlanguage[$polyglossia-lang.options$]{$polyglossia-lang.name$}
+  \setmainlanguage[$for(polyglossia-lang.options)$$polyglossia-lang.options$$sep$,$endfor$]{$polyglossia-lang.name$}
 $for(polyglossia-otherlangs)$
```
jgm, line 332: Note: a nicer syntax that now works is
```
$for(ployglossia-lang.options)$$it$$sep$,$endfor$
```

frederik-elwert: Thanks for the hint!

Since part of the rationale for this was aligning it with the rest of the template in handling LaTeX package options: Should this also be changed in all other places in the template where a similar syntax is used? Or just in this instance?

jgm: You're probably right that it's best to keep it uniform.
Maybe we should just merge it as it is, and perhaps think about updating the template more generally in another commit.

PR #6449, opened 2020-06-09
Allow multicols env in LaTeX reader

ip1981:
The multicols environment has a mandatory argument - number of columns.
It is incorrectly parsed as plain string. Wrap multicols content into nested divs.

Signed-off-by: Igor Pashev <pashev.igor@gmail.com>

### DIFF
--- a/test/command/latex-multicols.md
+++ b/test/command/latex-multicols.md
@@ -1,0 +1,14 @@
+# `\begin{multicols}`
+
+```
+% pandoc -f latex -t native
+\begin{multicols}{2}
+Hello
+\end{multicols}
+^D
+[Div ("",["columns"],[])
+ [Div ("",["column"],[])
+  [Div ("",["column"],[])
+   [Para [Str "Hello"]]]]]
+```
+
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -1913,6 +1913,7 @@ environments = M.fromList
    , ("abstract", mempty <$ (env "abstract" blocks >>= addMeta "abstract"))
    , ("sloppypar", env "sloppypar" $ blocks)
    , ("letter", env "letter" letterContents)
+   , ("multicols", env "multicols" multicols)
    , ("minipage", env "minipage" $
           skipopts *> spaces *> optional braced *> spaces *> blocks)
    , ("figure", env "figure" $ skipopts *> figure)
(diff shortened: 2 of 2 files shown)


jgm:
Maybe even better, we could parse it into a structure with Divs.
This would give people a chance to do something with the structured data.
````
Div [("",["columns"],[])
  [ Div [("",["column"],[])
    [...]
````
We actually support this in beamer output.

**Comment by ip1981** [comment]
> Maybe even better, we could parse it into a structure with Divs.
> This would give people a chance to do something with the structured data.
> 
> ```
> Div [("",["columns"],[])
>   [ Div [("",["column"],[])
>     [...]
> ```

Agree! I just discovered that HTML/CSS support multi-columns docs nowadays.
I think we then should keep the number of columns somehow, probably as a key-value pair?
And maybe use just one `Div` - I fail to see how the external Div might be helpful.

**Comment by jgm** [comment]
There is one external div with class "columns", and N internal divs (where N = the number of columns) each with class "column".

File: test/command/latex-multicols.md [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+% pandoc -f latex -t native
+\begin{multicols}{2}
+Hello
+\end{multicols}
+^D
+[Div ("",["columns"],[])
+ [Div ("",["column"],[])
+  [Div ("",["column"],[])
+   [Para [Str "Hello"]]]]]
+```
```
jgm, line 13: I think you may have misunderstood what I meant about two `column` Divs: the structure should be:
```
Div columns
[ Div column
, Div column ]
```
and not
```
Div columns
[ Div column
  [ Div column ...
```
Does that make more sense?

ip1981: Even simpler :) I will update

ip1981: Oh, what would be the content of the first div for example? Empty?

```
Div columns
[ Div column ""
, Div column "text foo hello"]
```
?

jgm: Maybe I need to understand better what you're trying to do.

The latex multicols environment creates columned text, but it computes where the column breaks go.
In order to duplicate this in pandoc we'd need pandoc to compute where a new column starts.
Is this something you're trying to do?

ip1981: No, I am basically trying to ignore the number of columns in latex :)

```latex
\begin{multicols}{2} % <- here
% ...
\end{multicols}
```

There is nothing more we can do, I think.

jgm: OK, then maybe I led you astray in my advice to include the nested column divs.
Maybe it makes most sense simply to include an outer "columns" div, and then the content inside.

Issue #2923, opened 2016-05-17
Markdown reader: add extension for GitHub wiki's internal link syntax [[page name]]


ferventcoder:
Forgive me if this has already been noted as an issue but I did a bit of searching through issues and online and didn't turn up with anything.

I'm on the latest version of Pandoc (1.17.0.2) and I'm attempting to use markdown_github to convert to HTML. I'm noticing that Pandoc does not convert links that look like the following:
- `[[example link|MarkdownPage]]`
- `[[MarkdownPage]]`

It's possible we are using outdated constructs, but GitHub has supported this for years. I noticed that Sublime supports the second but not the first with it's regular markdown preview tool.

The command I'm running:
`pandoc --from markdown_github+simple_tables+native_spans+native_divs+multiline_tables --to html5 [..snip..]`

Is this something that could be added for conversion? Or is there an extension that already does this that I may have missed?

jgm:
The syntax is documented [here](https://help.github.com/articles/adding-links-to-wikis/).
However, I believe it only works for GitHub wikis, and not in other contexts (e.g. bug reports).  Proof:
[[example link|MarkdownPage]]

It would be worth adding an extension for this.

ferventcoder:
Thanks @jgm! Pandoc is pretty amazing!  I'll admit I haven't read the source, but this is the regex I'm using to do this conversion now:
https://github.com/ferventcoder/chocolatey.org/blob/45765dd4cc2f9d2675d0f48fd4dd60669afb8bf8/UpdateDocumentation.ps1#L12-L24

jgm:
Re-opening, on the realization that we really need the parser in a commonmark extension (so it can be used with`gfm`).  Once that is available, we could *also* add a parser to pandoc's markdown reader, but I don't want to just add it to the latter; it will cause too much confusion.

jgm:
The tests and extension module work in #6458 will be useful when the commonmark-hs extension is available.

edrex:
[Emanote](https://note.ema.srid.ca/) (a publishing tool compatible with Obsidian and Foam) has its own implementation of the `[[Page|Label]]` syntax used by most wikis but it would be nice to switch to an upstream one.

https://github.com/srid/emanote/blob/master/src/Emanote/Pandoc/Markdown/Syntax/WikiLink.hs

anutator:
Please add mediawiki images conversion:
`![[image1.png]]`

PR #6458, opened 2020-06-14
Markdown reader: support GitHub wiki's internal links (#2923)

blackheaven:
closes https://github.com/jgm/pandoc/issues/2923

Changes overview:

 * Add a `Ext_markdown_github_wikilink` extension
   * Added in the `githubMarkdownExtensions` list
   * Added in the `getAllExtensions` list
 * Add the parser `githubWikiLink` in `Text.Pandoc.Readers.Markdown`
   * Added it high in the `inline` parsers list
 * Add two tests, one for each form

Notes:

 * It has been done as part as ZuriHac 2020
 * It is my first contribution
 * I am available to pair on Discord if needed

Thanks by advance for your feedback.

### DIFF
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -118,6 +118,7 @@ data Extension =
     | Ext_literate_haskell    -- ^ Enable literate Haskell conventions
     | Ext_markdown_attribute      -- ^ Interpret text inside HTML as markdown iff
                                   --   container has attribute 'markdown'
+    | Ext_wikilinks -- ^ Interpret a markdown wiki link
     | Ext_markdown_in_html_blocks -- ^ Interpret as markdown inside HTML blocks
     | Ext_mmd_header_identifiers -- ^ Multimarkdown style header identifiers [myid]
     | Ext_mmd_link_attributes     -- ^ MMD style reference link attributes
@@ -258,6 +259,7 @@ githubMarkdownExtensions = extensionsFromList
   , Ext_emoji
   , Ext_fenced_code_blocks
   , Ext_backtick_code_blocks
+  , Ext_wikilinks
   ]
 
 -- | Extensions to be used with multimarkdown.
@@ -444,6 +446,7 @@ getAllExtensions f = universalExtensions <> getAll f
        , Ext_tex_math_single_backslash
        , Ext_tex_math_double_backslash
        , Ext_markdown_attribute
+       , Ext_wikilinks
        , Ext_mmd_title_block
        , Ext_abbreviations
        , Ext_autolink_bare_uris
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -1485,6 +1485,7 @@ inline = choice [ whitespace
                 , note
                 , cite
                 , bracketedSpan
+                , githubWikiLink
                 , link
                 , image
                 , math
@@ -1779,6 +1780,42 @@ source = do
 
 linkTitle :: PandocMonad m => MarkdownParser m Text
 linkTitle = quotedTitle '"' <|> quotedTitle '\''
+
+-- Github wiki style link, with optional title
+-- syntax documented under https://help.github.com/en/github/building-a-strong-community/editing-wiki-content
+githubWikiLink :: PandocMonad m => MarkdownParser m (F Inlines)
+githubWikiLink = try $ do
+  guardEnabled Ext_wikilinks
+  string "[["
+  inlinedContents <- option (return $ B.fromList [])
+                      $ lookAhead $ try $ mconcat <$> manyTill inline (try $ string "]]")
+  rawContents <- manyTillChar anyChar (try $ string "]]")
+  let allowedInline x = case x of
+                          B.Str _   -> True
+                          B.Space   -> True
+                          B.Link {} -> True
+                          _         -> False
+  let hasNotInlines = all allowedInline . B.toList
+  let ilcs = runF inlinedContents defaultParserState
+  let valid = hasNotInlines ilcs
+  let formatUri l = if isURI l
+                      then l
+                      else T.replace " " "-" l
+  let (label, url) = case T.splitOn "|" rawContents of
+                       (l:u:_) -> (l, u)
+                       [l]     -> (l, l)
+                       []      -> error "T.splitOn changed behavior"
+  let dropLastLink' is = case is of
+                           []                           -> []
+                           -- Merge strings, special case
+                           [B.Str p, B.Link _ _ (l, _)] -> [B.Str $ p <> l]
+                           [B.Link _ _ (l, _)]          -> [B.Str l]
+                           [x]                          -> [x]
+                           (x:xs)                       -> x:dropLastLink' xs
+  let dropLastLink is = B.text "[[" <> B.fromList (dropLastLink' $ B.toList is) <> B.text "]]"
+  return $ if valid
+             then B.link (formatUri url)  "wikilink"  (B.text label) <$ inlinedContents
+             else dropLastLink <$> inlinedContents
 
 link :: PandocMonad m => MarkdownParser m (F Inlines)
 link = try $ do
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -21,6 +21,7 @@ module Text.Pandoc.Readers.Markdown (
 import Control.Monad
 import Control.Monad.Except (throwError)
 import Data.Char (isAlphaNum, isPunctuation, isSpace)
+import Data.Functor (($>))
 import Data.List (transpose, elemIndex, sortOn)
 import qualified Data.Map as M
 import Data.Maybe
@@ -1790,10 +1791,10 @@ githubWikiLink = try $ guardEnabled Ext_wikilinks >> wikilink
       string "[["
       firstPart <- fmap mconcat . sequence <$> wikiText
       (char '|' *> complexWikilink firstPart)
-        <|> (string "]]" *> return (B.link
-                                      <$> (stringify <$> firstPart)
-                                      <*> return "wikilink"
-                                      <*> firstPart))
+        <|> (string "]]" $> (B.link
+                               <$> (stringify <$> firstPart)
+                               <*> return "wikilink"
+                               <*> firstPart))
 
     complexWikilink firstPart = do
       url <- fmap stringify . sequence <$> wikiText
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -1787,7 +1787,7 @@ githubWikiLink :: PandocMonad m => MarkdownParser m (F Inlines)
 githubWikiLink = try $ do
   guardEnabled Ext_wikilinks
   string "[["
-  inlinedContents <- option (return $ B.fromList [])
+  inlinedContents <- option (return mempty)
                       $ lookAhead $ try $ mconcat <$> manyTill inline (try $ string "]]")
   rawContents <- manyTillChar anyChar (try $ string "]]")
   let allowedInline x = case x of
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -1784,38 +1784,25 @@ linkTitle = quotedTitle '"' <|> quotedTitle '\''
 -- Github wiki style link, with optional title
 -- syntax documented under https://help.github.com/en/github/building-a-strong-community/editing-wiki-content
 githubWikiLink :: PandocMonad m => MarkdownParser m (F Inlines)
-githubWikiLink = try $ do
-  guardEnabled Ext_wikilinks
-  string "[["
-  inlinedContents <- option (return mempty)
-                      $ lookAhead $ try $ mconcat <$> manyTill inline (try $ string "]]")
-  rawContents <- manyTillChar anyChar (try $ string "]]")
-  let allowedInline x = case x of
-                          B.Str _   -> True
-                          B.Space   -> True
-                          B.Link {} -> True
-                          _         -> False
-  let hasNotInlines = all allowedInline . B.toList
-  let ilcs = runF inlinedContents defaultParserState
-  let valid = hasNotInlines ilcs
-  let formatUri l = if isURI l
-                      then l
-                      else T.replace " " "-" l
-  let (label, url) = case T.splitOn "|" rawContents of
-                       (l:u:_) -> (l, u)
-                       [l]     -> (l, l)
-                       []      -> error "T.splitOn changed behavior"
-  let dropLastLink' is = case is of
-                           []                           -> []
-                           -- Merge strings, special case
-                           [B.Str p, B.Link _ _ (l, _)] -> [B.Str $ p <> l]
-                           [B.Link _ _ (l, _)]          -> [B.Str l]
-                           [x]                          -> [x]
-                           (x:xs)                       -> x:dropLastLink' xs
-  let dropLastLink is = B.text "[[" <> B.fromList (dropLastLink' $ B.toList is) <> B.text "]]"
-  return $ if valid
-             then B.link (formatUri url)  "wikilink"  (B.text label) <$ inlinedContents
-             else dropLastLink <$> inlinedContents
+githubWikiLink = try $ guardEnabled Ext_wikilinks >> wikilink
+  where
+    wikilink = try $ do
+      string "[["
+      firstPart <- fmap mconcat . sequence <$> wikiText
+      (char '|' *> complexWikilink firstPart)
+        <|> (string "]]" *> return (B.link
+                                      <$> (stringify <$> firstPart)
+                                      <*> return "wikilink"
+                                      <*> firstPart))
+
+    complexWikilink firstPart = do
+      url <- fmap stringify . sequence <$> wikiText
+      string "]]"
+      return $ B.link <$> url
+                      <*> return "wikilink"
+                      <*> firstPart
+
+    wikiText = many (whitespace <|> bareURL <|> str <|> endline <|> escapedChar)
 
 link :: PandocMonad m => MarkdownParser m (F Inlines)
 link = try $ do
--- a/test/Tests/Readers/Markdown.hs
+++ b/test/Tests/Readers/Markdown.hs
@@ -307,6 +307,36 @@ tests = [ testGroup "inline code"
             "[https://example.org(](url)" =?>
             para (link "url" "" (text "https://example.org("))
           ]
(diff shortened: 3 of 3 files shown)


src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1735,6 +1736,17 @@ source = do
 linkTitle :: PandocMonad m => MarkdownParser m Text
 linkTitle = quotedTitle '"' <|> quotedTitle '\''
 
+-- Github wiki style link, with optional title
```
mb21, line 1739: maybe add a comment like:

-- syntax documented under https://help.github.com/en/github/building-a-strong-community/editing-wiki-content

test/Tests/Readers/Markdown.hs
```diff
@@ -307,6 +307,14 @@ tests = [ testGroup "inline code"
             "[https://example.org(](url)" =?>
             para (link "url" "" (text "https://example.org("))
           ]
+        , testGroup "Github wiki links"
+          [ test markdownGH "autolink" $
+            "[[https://example.org]]" =?>
+            para (link "https://example.org" "" (text "https://example.org"))
+          , test markdownGH "link with title" $
+            "[[title|https://example.org]]" =?>
+            para (link "title" "" (text "https://example.org"))
```
mb21, line 316: maybe add a test for e.g.

    [[Link Text|nameofwikipage]]

and one for:

    [[Name of Page]]

(assuming that's the syntax for pages with spaces in them)

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1735,6 +1736,17 @@ source = do
...
+githubWikiLink :: PandocMonad m => MarkdownParser m (F Inlines)
+githubWikiLink = try $ do
+  guardEnabled Ext_markdown_github_wikilink
+  string "[["
+  skipSpaces
+  let allUntil = manyTillChar (satisfy $ const True)
```
jgm, line 1745: `satisfy $ const True` is just `anyChar`, right?

jgm: Also: do we really want anyChar here?  Suppose you have `[[` followed by sixteen paragraphs of text, each separated by a blank line.

jgm: In pandoc's links, the link description can contain inline formatting.  It would be natural to extend that feature here, though I don't know how GitHub's parsing actually works.

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1735,6 +1736,17 @@ source = do
...
+  part1 <- allUntil (oneOf "|]")
+  part2 <- (char ']' >> return part1) <|> allUntil (string "]]")
```
jgm, line 1747: `allUntil (string "]]")`: note that you generally want something like `(try (string "]]"))` in this context, because otherwise your parser will fail when it encounters a single `]` with something else after it.  (Maybe that's what you want in this case, though?)

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1735,6 +1736,17 @@ source = do
...
+  returnF $ B.link part1 "" (B.text $ escapeURI $ trimr part2)
```
jgm, line 1748: In the MediaWiki reader, we insert a special title "wikilink" (if I recall) for wikilinks. This allows them to be distinguished in the AST, so people can do things like transform their URLs (adding a prefix for example).

test/Tests/Readers/Markdown.hs
jgm: Another good test (as noted above) wolud be `[[Name of page] and]]`

src/Text/Pandoc/Extensions.hs
```diff
@@ -118,6 +118,7 @@ data Extension =
     | Ext_literate_haskell    -- ^ Enable literate Haskell conventions
     | Ext_markdown_attribute      -- ^ Interpret text inside HTML as markdown iff
                                   --   container has attribute 'markdown'
+    | Ext_markdown_github_wikilink -- ^ Interpret a markdown wiki link
```
jgm, line 121: I think I'd favor the name `ext_wikilink`.  This is a pretty generic wikilink format, certainly not GitHub specific.

mb21: apparently it's inspired by mediawiki.... and [apparently](https://help.github.com/en/github/building-a-strong-community/editing-wiki-content) you can also use mediawiki image syntax:

    [[https://github.com/username/repository/blob/master/img/octocat.png|alt=octocat]]

but yeah, in markdown the link and image syntaxes are also closely related, so why not `ext_wikilink` :)

src/Text/Pandoc/Readers/Markdown.hs
mb21: I just tried in https://github.com/jgm/pandoc/wiki/Sandbox/_edit so yeah, bold and italics appear supported... but not links of course.

blackheaven:
I have tried to take care of comments:

 * Change the extension name
 * Added documentation
 * Imported the MediaWiki's way to read wikilinks (drop 'Category' part and limit support to italics and strongs)
 * Added tests

Sadly, my tests are failing and I do not see what I am doing wrong, if someone can have a look and see something obvious, I would be grateful.

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1737,15 +1737,15 @@ linkTitle :: PandocMonad m => MarkdownParser m Text
...
 githubWikiLink = try $ do
-  guardEnabled Ext_markdown_github_wikilink
-  string "[["
-  skipSpaces
-  let allUntil = manyTillChar (satisfy $ const True)
-  part1 <- allUntil (oneOf "|]")
-  part2 <- (char ']' >> return part1) <|> allUntil (string "]]")
-  returnF $ B.link part1 "" (B.text $ escapeURI $ trimr part2)
+  guardEnabled Ext_wikilink
+  try $ string "[["
```
jgm, line 1744: You don't need this `try` or the one on l 1747 because you've got a `try` over the whole `githubWikiLink` parser (which maybe should be named `wikiLinkParser`?)  Having the unnecessary `try`s just hurts performance.

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1737,15 +1737,15 @@ linkTitle :: PandocMonad m => MarkdownParser m Text
...
+  pagename <- T.unwords . T.words <$> manyChar (noneOf "|]")
+  label <- option (return $ B.text pagename) $ char '|' *> (mconcat <$> many1 (notFollowedBy (string "]]") *> strongOrEmph))
```
jgm, line 1746: Here I think you *do* need a `try`:

`notFollowedBy (try (string "]]"))`

To test this, use something like `[[a|b]and more]]`

blackheaven: good catch, great, thank you!

blackheaven:
Any blocking points?

src/Text/Pandoc/Extensions.hs
```diff
@@ -118,6 +118,7 @@ data Extension =
     | Ext_literate_haskell    -- ^ Enable literate Haskell conventions
     | Ext_markdown_attribute      -- ^ Interpret text inside HTML as markdown iff
                                   --   container has attribute 'markdown'
+    | Ext_wikilink -- ^ Interpret a markdown wiki link
```
jgm, line 121: Sorry, but on reflection I'm thinking `Ext_wikilinks` would be more consistent with the extension naming style.

test/Tests/Readers/Markdown.hs
```diff
@@ -307,6 +307,26 @@ tests = [ testGroup "inline code"
...
+            "[[Name of *page*]]" =?>
+            para (link "Name of *page*" "wikilink" (text "Name of " <> emph "page"))
+          , test markdownGH "link with title" $
+            "[[title|https://example.org]]" =?>
+            para (link "https://example.org" "wikilink" (text "title"))
+          , test markdownGH "bad link with title" $
+            "[[title|random string]]" =?>
+            para (link "random string" "wikilink" (text "title"))
+          ]
         , testGroup "Headers"
```
jgm, line 330: Does GitHub (or the markdown parser they use in wikis, which I think may be kramdown) have a suite of tests for this syntax?  If so, it may be worth finding it and seeing if any relevant tests can be added.

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1735,6 +1736,35 @@ source = do
...
+  contents <- lookAhead $ manyTillChar anyChar (try $ string "]]")
+  let title = "wikilink"
+  if T.any (== '|') contents
+  then do
+    st <- getState
+    setState $ st{ stateAllowLinks = False }
+    label <- mconcat <$> manyTill inline (char '|')
+    setState $ st{ stateAllowLinks = True }
+    url <- manyTillChar anyChar $ try $ string "]]"
+    return $ B.link url title <$> label
```
jgm, line 1754: This approach seems potentially unreliable.  Example: 
```
[[hello `|` there]]
```
IN this case `contents` will be `` hello `|` there ``, but when we do `manyTill inline (char '|')` we will keep parsing past the `]]`, because the `|` character occurs *within* an inline (a code span).

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1442,6 +1442,7 @@ inlines1 = mconcat <$> many1 inline
 
 inline :: PandocMonad m => MarkdownParser m (F Inlines)
 inline = choice [ whitespace
+                , githubWikiLink
```
jgm, line 1445: Is there a reason this is here and not further down in the list?
Probably best to put it right before the regular link and image parsers; that should improve performance a bit.

src/Text/Pandoc/Readers/Markdown.hs
jgm: This is tricky, and I'm not sure how it should behave, because we don't know whether to treat the content as plain text or inlines until we know whether there's a `|` character, but that `|` character might be part of regular inline content.  I'm really not sure what to do here.

blackheaven: Good point, I have actually borrowed it from the VimWiki parser, we may have the same issue.

jgm:
Worth figuring out whether the GitHub implementation of this allows inline formatting in the description part.  If it's treated as a plain string, maybe we could do the same, and it would solve the problem noted above.  Finding their test suite for this would be nice!

blackheaven:
If there is a test suite, it is well hidden.

I have looked for hours, tried babelmark2, no other implementation supports it, no documentation... I begin to doubt of it.

I have tried to make it recursive but it causes issues with autolinks.

blackheaven:
Great news! wikilinks are actually way simpler than we expected: https://github.com/blackheaven/pandoc/wiki

blackheaven:
it should be good now

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1732,6 +1733,42 @@ source = do
 linkTitle :: PandocMonad m => MarkdownParser m Text
 linkTitle = quotedTitle '"' <|> quotedTitle '\''
 
+-- Github wiki style link, with optional title
+-- syntax documented under https://help.github.com/en/github/building-a-strong-community/editing-wiki-content
+githubWikiLink :: PandocMonad m => MarkdownParser m (F Inlines)
+githubWikiLink = try $ do
+  guardEnabled Ext_wikilinks
+  string "[["
+  inlinedContents <- option (return $ B.fromList [])
```
jgm, line 1742: Note that `B.fromList []` is just `mempty` which would be clearer and faster.

jgm: If you want to get both the inline contents and the raw contents, you can use the `withRaw` combinator (T.P.Parsing).

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1732,6 +1733,42 @@ source = do
...
+                      $ lookAhead $ try $ mconcat <$> manyTill inline (try $ string "]]")
+  rawContents <- manyTillChar anyChar (try $ string "]]")
+  let allowedInline x = case x of
+                          B.Str _   -> True
+                          B.Space   -> True
+                          B.Link {} -> True
```
jgm, line 1748: Are all markdown links allowed?  Or just bare URIs?
If just bare URIs, then maybe a better approach would be to use a parser that just targets what we need.

jgm: GitHub's documentation says: `[[Link Text|nameofwikipage]]`

So there are two questions.  1) What can go in "Link Text"?  Your experiments appear to reveal that only plain text is allowed, no special formatting.  2) What can go in "nameofwikipage"?  Your experiments show  that a fully qualified URL is allowed, but clearly that's not all -- names of wiki pages are also allowed.  I imagine that just about any plain text could be the name of a wiki page, so presumably we should allow e.g. `[[Link Text|This is my page]]`.  For this we should emit a link with URL `This is my page` but also a special attribute indicating that it's a wikilink, so that users can do transformations on the path.

jgm: (OK, looks like that's what you're doing, from the tests!)

hftf: This is not pertinent to GitHub-flavored Markdown, but it may be informative to note that wikilinks have been growing in popularity due to the explosion of note-taking and (personal) knowledge management tools, which are often based on Markdown or Markdown-ish syntaxes. Some (nonstandard) implementations allow link text to contain inline formatting, and some even allow page titles (and thus link targets) to contain both formatting and nested links.

jgm: Anyway, seems to me we should try to simplify the parser.  Instead of parsing inlines and raw, and then trying to piece things together, why not something like

``` haskell
wikilink = try $ do
  string "[["
  firstPart <- wikiText
  (char '|' *> complexWikilink firstPart) <|> (string "]]" *> return (wikilink with firstPart as link text and (stringify firstPart) as url)

complexWikilink firstPart = do
  url <- stringify <$> wikiText
  string "]]"
  return $ wikilink with firstPart as link text and url as URL

wikiText = many ( whitespace <|> bareURL <|> str <|> endline <|> escapedChar )
```

jgm:
Sorry I left this sit for so long!  I've made some comments in the code.

src/Text/Pandoc/Readers/Markdown.hs
blackheaven: Actually, I do not know if I did correctly, but it broke the tests

src/Text/Pandoc/Readers/Markdown.hs
blackheaven: I have tried you suggestion, but it seems that I miss the point

blackheaven:
> Sorry I left this sit for so long! I've made some comments in the code.

No worries, my sparse time has drastically decreased in the mean time.

src/Text/Pandoc/Readers/Markdown.hs
jgm: It's okay, let me merge your commit, and then I'll see if there's a way to simplify it.

jgm:
Thanks for this contribution!

jgm:
Sorry!  After merging this I realized something that has caused me to revert the commit.

Since this is designed to be used with `gfm`, we should implement the parser as a commonmark extension (commonmark-extensions), rather than here.  The code in pandoc's markdown reader is NOT used when parsing `gfm`. I presume that most users who will want this feature will want to use it with `gfm`, so I think we should wait.

I'm sorry I didn't think of this earlier, before you put the time into this PR.

Writing a commonmark-hs extension is an entirely different thing, and I need to think more about how it would be done with this syntax.

jgm:
The work you've done in constructing the tests and in the extensions module can be re-used once we have the new commonmark-hs module in place.

jgm:
I've started to add a `wikilinks` extension in commonmark-hs.

blackheaven:
Sure, no problem

marcelotto:
This is a really exciting extension. I was already about to submit a feature request when I came across this PR. I just want to undermine that this extension isn't only useful for Github Wiki pages. Using Wikilinks in Markdown is getting fairly popular recently and presumably even more in the future with the recent raise of local-first, Markdown-based knowledge management tools like [Obsidian](https://obsidian.md/), [Logseq](https://github.com/logseq/logseq), [Dendron](https://github.com/dendronhq/dendron) to name just a few.

jgm:
I've run into a problem implementing this.
There are two incompatible syntaxes (different orders) out there: see
https://github.com/jgm/commonmark-hs/issues/69
Not quite sure what to do about it.

hftf:
There are now many incompatible implementations and interpretations of "wikilinks" in existence, which is partly due to the recent growth of note-taking tools that [I](https://github.com/jgm/pandoc/pull/6458#discussion_r552125324) and others have mentioned above.

I foresee that many users of pandoc are going to want to operate on documents containing wikilinks of all kinds. This is a challenge.

Perhaps an encompassing solution would be "DIY." Instead of favoring one implementation out of many for a single extension called `wikilinks`, Pandoc could implement various small pieces of functionality commonly found in wikilink implementations, such as a pair of mutually exclusive options to handle either `[[target|link text]]` or `[[link text|target]]` (`wikilink-piped-link-text-on-right` and `-left`), or to allow or disallow nested inline formatting. Particularly well-known implementations could be aliased to combinations of options (e.g. `wikilinks-gfm` is an alias of `wikilinks-option1+wikilinks-option3` etc.).

marcelotto:
Sorry @hftf, didn't notice your similar remarks. I guess you're right, it's a challenging problem which ideally would be solved in the most flexible way you've outlined, but I can image that this can lead to a maintainability nightmare. From what I can see two alternative extensions should be sufficient if the extensions are limited on capturing the pure links: One for classical Wikilinks (`[[link]]` and `[[link|label]]`) and one for Github-flavoured Wikilinks (`[[link]]` and `[[label|link]]`). Regarding the additional system-specific features on top of that (inline formatting etc.), either people with portability in mind are hesitant on their usage in the first place or have  with these extensions at least the option to capture the links with a filter and do the additional interpretation on their own.

One remark regarding the understandably frustrated comment

> This is how it's done in GitHub wikis. I don't know why everyone has to reinvent the wheel in incompatible ways!

by @jgm in the [commonmark-hs issue](https://github.com/jgm/commonmark-hs/issues/69). It should be noted that GitHub were the ones breaking with the established way. The other tools are using the classical Wikilink-style introduced by UseModWiki and used in almost all Wiki systems [1] and in all the new systems from the new generation of knowledge management tools (as far as I can see).

jgm:
Fair point: GitHub was never great about looking at prior art in making markdown extensions.

PR #6462, opened 2020-06-15
Remove unused dependency `vector`

LaurentRDC:
Hello,

I noticed that the `vector` package is actually never used explicitly in Pandoc. This PR removes the dependency. Most importantly, there won't be a need to track the version upper bound anymore (e.g. dce6a7388a5f546ebbfd192febcb1753b743b377 and c56669ec39b741f8b14f16603294e27415576510).


jgm:
Good, thanks!

Issue #1761, opened 2014-11-14
EPUB3 otf fonts - incorrect media-type


andrecastro0o:
EPUB3 media-type for a otf font in item manifest should be application/vnd.ms-opentype instead of application/x-font-opentype

When converting a markdown to epub3 using `--epub-embed-font` and otf fonts the conversion results in EPUB where the manifest item for the font has an incorrect media-type. It uses:
 `<item id="VAGRoundedStd-Thin_otf" href="VAGRoundedStd-Thin.otf" media-type="application/x-font-opentype" />` 

The recommendation is to use `media-type="application/vnd.ms-opentype"`: 

> "The media type application/vnd.ms-opentype should be specified as the media-type value in the item of the manifest that references an OpenType font." http://www.idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section2.3.1

When checking the EPUB with epubcheck it trows an Warning cause by the media-type media-type="application/x-font-opentype" :  `WARN: /stylesheet.css(5): Font-face reference VAGRoundedStd-Light.otf to non-standard font type application/x-font-opentype`

PR #6464, opened 2020-06-15
Fix MIME type for TrueType fonts in EPUBs

mreed20:
Heyo. [Per the EPUB 3.2 spec][1], `application/x-font-truetype` is no longer a valid identifier for TrueType fonts. This fixes warnings when validating pandoc-generated EPUBs using [epubcheck][2].

References plfa/plfa.github.io#479.
References #1761.

[1]: https://www.w3.org/publishing/epub3/epub-spec.html#sec-core-media-types
[2]: https://github.com/w3c/epubcheck


jgm:
thanks!

PR #6487, opened 2020-06-28
Remove obsolete RelaxedPolyRec extension

lierdakil:
Moved from #6328 :

I removed references to RelaxedPolyRec language extension everywhere I found it. Since I believe the current pandoc source can't be built with GHC 7 and below anyway, and RelaxedPolyRec is literally impossible to turn off since at least GHC 7.6, and almost all mention of this extension is removed from GHC documentation, I'd say it's a good idea to remove this extension as obsolete.


jgm:
thanks!

PR #6492, opened 2020-06-30
docbook reader: Implement column span support for tables

pyssling:
Implement column span support for tables in the DocBook reader.


pyssling:
@jgm this is hopefully good for merging. I added a lot of function signatures - they made the refactoring a lot easier as it narrowed the scope of error messages.

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -900,15 +905,35 @@ parseBlock (Elem e) =
...
+                 if (isJust si) && (isJust ei)
+                   then ColSpan $ (fromJust ei) - (fromJust si) + 1
+                   else 1
+           let toColSpan :: Element -> ColSpan
+               toColSpan en = do
+                 let sa = findAttr (unqual "namest") en
+                 let ea = findAttr (unqual "nameend") en
+                 if (isJust sa) && (isJust ea)
+                   then colDistance (fromJust sa) (fromJust ea)
+                   else 1
```
mb21, line 933: you might want to rewrite this to something like:

```haskell
let mStrt = findAttr (unqual "namest") en
let mEnd = findAttr (unqual "nameend") en
case (mStrt, mEnd) of
  (Just start, Just end) -> colDistance start end
  _ -> 1
```

mb21:
nice!

>  I added a lot of function signatures - they made the refactoring a lot easier as it narrowed the scope of error messages.

that makes sense... however, we usually don't have signatures for inline functions.... but I'll leave it up to @jgm to decide whether they should be removed... or you could move those functions out to the top level (where e.g. `getBlocks` is)...?

src/Text/Pandoc/Readers/DocBook.hs
pyssling: done - thanks - much cleaner!

pyssling:
@mb21 ok, I removed the signatures that were trivial and moved the functions where I wanted the signatures out to the top level. I agree - much cleaner.

pyssling:
@jgm any chance you could have a look at this?

pyssling:
@jgm sorry to ping you again, but any luck looking at this?

pyssling:
@jgm @mb21 any chance this could be merged, or that you could let me know what's missing?

jgm:
Sorry for the delay!  This looks good to me, thanks!  (Row spans next?)

Issue #6491, opened 2020-06-30
Pandoc 2.x renders images' alternative texts in an inaccessible fashion


jmuheim:
As stated on StackOverflow (https://stackoverflow.com/questions/62639927/pandoc-2-x-renders-images-alternative-texts-in-an-inaccessible-fashion?noredirect=1#comment110781365_62639927), Pandoc 2.x renders images' alternative texts in an inaccessible fashion. I was told there to ask for a bugfix here.

---

Here's the original post:

Since I upgraded from Pandoc v1.19 to 2.9, decorative images are not exported as expected anymore.

First of all, when generating HTML from `![](test.jpg)`, in v1.19 a `<p class="figure">` structure was wrapped around the image, but now it's only a `<p>`:

    <p>
      <img src="test.jpg">
    </p>

This makes it harder to style in line with other images that have an alternative text.

But what's really a problem here: there's no `alt=""` attribute produced anymore! This means that e.g. screen readers will not recognise this as a decorative image anymore.

So let's see what happens to an image with an actual alternative text, e.g. when generating HTML from `![Hello](test.jpg)`:

    <div class="figure">
      <img src="test.jpg" alt="">
      <p class="caption">Hello</p>
    </div>

Here we get a `class="figure"` in the surrounding element, but now it's a `<div>` instead of a `<p>` (I don't bother too much about this, but again, it makes it harder to style everything the same).

What again is a big problem though is the fact that the `alt` attribute is now set empty: this prevents screen readers from perceiving them at all, which is horribly wrong! I guess that Pandoc concludes that having alternative text and caption would be redundant, which is correct, and that the caption below would be the right thing to show - which it is not.

The right structure would look something like this:

    <div class="figure">
      <img src="test.jpg" alt="Hello">
      <p class="caption" aria-hidden="true">Hello</p>
    </div>

Any reason why this behaviour would make sense? Can it be changed somehow? Otherwise I will have to fiddle around with some post-processing JavaScript...

tarleb:
I started to implement this, but was given pause by the fact that this would cause pandoc to produce invalid xhtml when targeting HTML4. @jmuheim, do you know of a good workaround for HTML4?

On the other hand, we *already* produce invalid xhtml for any document which includes code blocks, as line numbers contain the `aria-hidden="true"` attribute.

jmuheim:
Interesting. You mean because `aria-hidden` has a dash in the attribute name, right?

[...]

PR #6495, opened 2020-07-01
HTML writer: improve alt-text/caption handling for HTML5

tarleb:
Screen readers read an image's `alt` attribute and the figure caption,
both of which come from the same source in pandoc. The figure caption is
hidden from screen readers with the `aria-hidden` attribute. This
improves accessibility.

For HTML4, where `aria-hidden` is not allowed, pandoc still uses an
empty `alt` attribute to avoid duplicate contents.

Closes: #6491


jmuheim:
Nice, thank you! ❤️

PR #6500, opened 2020-07-02
CONTRIBUTING.md: add notes on getting nightly directly from GitHub Actions

ickc:
Note: There isn't a direct link to the latest nightly from GitHub Actions, and it requires people to log in with a GitHub account (public access to that link has artifacts greyed out.)


jgm:
Thanks. I wanted to word this differently so I made a separate commit. I also think it's best in this document just to point people to the official nightlies.  It's okay if they have to be logged in with a GH account, since they'll need that to report a bug anyway.

Issue #6501, opened 2020-07-03
Crazy conversions


SarahSiani-IT98:
Hello, I have a word document which is strangely converted into html.

My text is this:
[cit_lit.docx](https://github.com/jgm/pandoc/files/4869380/cit_lit.docx)
> **_“We are all in the gutter, but some of us are looking at the stars”_** – Oscar Wilde, Lady Windermere’s Fan (1892)
> **_“We cast a shadow on something wherever we stand, and it is no good moving from place to place to save things; because the shadow always follows. Choose a place where you won’t do harm – yes, choose a place where you won’t do very much harm, and stand in it for all you are worth,  facing the sunshine.”_** – E. M. Forster, A Room with a View (1908)
> **_“Life appears to me too short to be spent in nursing animosity or registering wrongs.”_** – Charlotte Brontë, Jane Eyre (1847)
> **_“It was a bright cold day in April, and the clocks were striking thirteen.”_** – George Orwell, 1984 (1949)
> **_“No one is useless in this world who lightens the burdens of another.”_** – Charles Dickens, Doctor Marigold (1874)

And it is converted this:
`<div data-custom-style="Testo citato"><p><span data-custom-style="Enfasi"><em>“We are all in the gutter, but some of us are looking at the stars”</em> – Oscar Wilde, Lady Windermere’s Fan (1892)</span></p></div>`
tag strong lost

`<div data-custom-style="Testo citato"><p><span data-custom-style="Enfasi">“We cast a shadow on something wherever we stand, and it is no good moving from place to place to save things; because the shadow always follows. Choose a place where you won’t do harm – yes, choose a place where you won’t do very much harm, and stand in it for all you are worth, facing the sunshine.” – E. M. Forster, A Room with a View (1908)</span></p></div>`
both tags strong & em lost

`<div data-custom-style="Testo citato"><p><span data-custom-style="Enfasi">“Life appears to me <em>too short</em> to be spent in nursing animosity or registering wrongs.” – Charlotte Brontë, Jane Eyre (1847)</span></p></div>`
both tags strong & em lost but "too short" sentence is tagged em (**???**)

`<div data-custom-style="Testo citato"><p><span data-custom-style="Enfasi"><em><strong>“It was a bright cold day in April, and the clocks were striking thirteen.”</strong></em> – George Orwell, 1984 (1949)</span></p></div>`
All correct!

[...]

PR #6504, opened 2020-07-05
[Docx Reader] paragraph run styles

### LINKED ISSUE #6452: pandoc does not parse italic text from docx properly
Using pandoc 2.9.2.1. When convert this attached file to any formats, italic attribute is ignored.

[raw.docx](https://github.com/jgm/pandoc/files/4762552/raw.docx)

### LINKED ISSUE #6499: Problems identifying the correct character formatting from docx document
I am having many problems recovering the correct format of the text.
Problems arise when document styles (character styles and/or paragraph styles) contain information about character formatting.
(_This is a common practice in publishing to be able to "tag" documents with styles that can then be used in XML databases._)

First case (paragraph styles):
As you can see in the attached example file, paragraphs 1 to 4 are visually the same as those from 5 to 8. 
"`pandoc -f docx -t html test_styles.docx -o test.html`": the html/markdown is **different**.
"`pandoc -f docx+style -t html test_styles.docx -o test.html`": the html/markdown is **different**.

Second case (character styles)
Paragraphs 9 to 12 are the same as paragraphs 13 to 16.
"`pandoc -f docx`": the html/markdown is the **same**.
"`pandoc -f docx+style -t html test_styles.docx -o test.html`": the html/markdown is **different**.

**Is it a bug?**
Because it wouldn't make sense as "desired behavior".
Is there a **patch** or **workaround**?

Thank you very much

[test_styles.docx](https://github.com/jgm/pandoc/files/4861281/test_styles.docx)

lierdakil:
This is a request for comments related to #6501. 

Some design considerations (in no particular order):

- Generally speaking, we don't want to pull in run styles (bold, italic, etc) from semantically-meaningful paragraph styles. That is, we don't want to add explicit `Strong` to each heading or explicit `Emph` to each block quote, etc. Hence, those are filtered out by name and inheritance (see `semanticStyleNames` and `isSemanticParaStyle`) 

- As I outlined in https://github.com/jgm/pandoc/issues/6501#issuecomment-653809325, Word behaves differently from any other software that understands docx. This particular PR implements something mimicking Word's behaviour, although I can't guarantee it's exactly the same. Doing it the way everyone else does is possibly much easier. I can't really decide which option is the better one, but Word is generally considered _the_ reference docx implementation, so there's that. Anyway, this mess is all contained in `resolveDependentPsRs` and `mergeParaRunStyle`. The idea of the algorithm is as follows:

    1. Inline run style (i.e. one defined on the text itself) takes precedence over everything else (if defined).
    2. Character run style takes precedence over paragraph run style, except if both are bold/italic
    3. If both character and paragraph run styles are bold/italic, bold/italic is instead disabled.

    There is _some_ logic behind this behaviour: if a character style is used for emphasis, then in a paragraph where such style is the default, that style should be switched off for emphasis instead, f.ex. consider:

    > Within a larger body of text, a single piece of _emphasis_ does not stand out much.

    > _Within a larger body of text, a single piece of_ emphasis _does not stand out much._

    This is, however, somewhat counter-intuitive and I don't think any software besides Word handles it this way (?)

- We might want to hide this new behaviour behind an option, if we decide to ship it, since it might break some workflows. But, as far as I can tell, this is unlikely, since particularly tricky workflows probably rely on `docx+styles`, which is unaffected.

- Implementation-wise, I've found two relatively straightforward options for threading the paragraph style data to the place where inlines are constructed:

    1. The simpler one is implemented in the first commit -- it just adds a `Maybe ParagraphStyle` field to `DEnv` (a `Reader` environment passed around). It is simpler, but it has an unfortunate quirk -- `ParagraphStyle` passed this way is always defined at the place of consumption, hence an unnecessary check for `Maybe` is preformed. This approach is also somewhat brittle and unprincipled.
    2. The slightly more complex one is implemented in the second commit. It adds an additional `ReaderT ParagraphStyle` layer to the monad transformer stack for functions that depend on `ParagraphStyle` being known. The code is rather clean, but it does add a layer of the transformer stack, which by itself is fine, but overuse of deep transformer stacks will generally lead to horrors untold. It would possibly be a bit cleaner with extensible records, but Haskell doesn't have first-class labels, and I'm not going to drag in a big dependency just for this.

    Let me know what you think is better: unprincipled extraneous `Maybe` or an additional `Reader` layer. I'm personally strongly leaning to (2), because I like letting compiler prove that the code works. Maybe I'm missing an obvious great solution that uses neither, if so, please do tell.

- I've updated precisely one test. Didn't add new ones, although I probably should. In the one case I've updated, Docx Reader's output arguably became more consistent.

- There is at least one issue with the current implementation: `smushInlines`, which in theory should simplify the docx messiness, doesn't handle cases like `Strong [...], Space, Strong [...]` (which I reckon should be "smushed" into `Strong [..., Space, ...]`). This I believe to be a separate issue, and I can't claim a complete understanding of how `smushInlines` works, so I didn't attempt a fix for that here.

This is in theory working code and can be merged, so I'm not marking it as a draft. However, I would prefer to at least get some opinions on whether to use Word's behaviour for "double negation" of emph/strong and if there's a simpler way to thread `ParagraphStyle` around.

P.S. Apparently ~~fixes. #6501,~~ fixes #6452, fixes #6499; fix for #6501 is only partial, see also #6511


lierdakil:
After thinking about this for a bit, I've opted to add an extension to control Word's emphasis quirk. Also rebased on `master`, added a couple tests, made some cosmetic changes.

lierdakil:
Ping. Sorry to nag, but I'm afraid this will go stale. Are there any blockers? What can I do to help?

PR #6508, opened 2020-07-07
[Bench] Benchmark bytestring readers

lierdakil:
This is a quick hack to enable benchmarking on ByteString readers (e.g. docx). Not the safest code, due to sum-records, but I trust it's not that big of an issue in relatively short benchmark code.

### DIFF
--- a/benchmark/benchmark-pandoc.hs
+++ b/benchmark/benchmark-pandoc.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TupleSections     #-}
 {-
 Copyright (C) 2012-2019 John MacFarlane <jgm@berkeley.edu>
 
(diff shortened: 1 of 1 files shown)


benchmark/benchmark-pandoc.hs
```diff
@@ -21,39 +20,49 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
...
-          (rdr, rexts) <- getReader name
+            -> IO (Maybe Benchmark)
+readerBench doc name = do
+  let (rdr, rexts) = either (error . show) id . runPure $ getReader name
+  res <- runIO $ do
           (wtr, wexts) <- getWriter name
           case (rdr, wtr) of
             (TextReader r, TextWriter w) -> do
-                     setResourcePath ["../test"]
+                     setResourcePath ["./test"]
```
jgm, line 46: Was this just a mistake in the original, or did you change something that makes this necessary?

lierdakil: This was apparently a mistake in the original, since various `cabal bench` run benchmarks with the root directory as `cwd`. For pure writers it doesn't really matter though, since those don't access the FS anyway.

PR #6509, opened 2020-07-07
[Docx Reader] Refactor/update Text.Pandoc.Readers.Docx.Combine.smushInlines

lierdakil:
I did some minor refactoring in `Text.Pandoc.Readers.Docx.Combine`

### What I did

Reimplemented `spaceOutInlines`, merged `ilModifier` and `ilInlines` into a single function, cleaned up `unstackInlines`.

UPD: I also removed `NullModifier` which as it turns out is never used outside of recursion termination, and it makes much more sense to use `Maybe` for that.

UPD2: `spaceOutInlines` wrapped its result using `stackInlines` which `spaceOutInlinesL`/`spaceOutInlinesR` immediately unwrapped back with `unstackInlines`. Removed this as well.

Notable changes:

- `spaceOutInlines` didn't handle `SoftBreak` correctly -- which isn't necessarily an issue, docx reader doesn't produce `SoftBreak`s anyway. But I figured it's better to future-proof it just in case.

- `unstackInlines` didn't handle `Underline` correctly. Now it does. That is the reason tests had to be updated.

### Miscellaneous notes

There are a lot of functions a-la
```haskell
isSpace Space = True
isSpace SoftBreak = True
isSpace _ = False
```
around the code base. Perhaps it would be better to export this from pandoc-types instead of duplicating it everywhere?

### Performance considerations

Could not detect any change in performance (via #6508)

/cc @jkr -- pretty sure most of it is your code, so it would be great if you could take a look.


jkr:
This all looks good to me -- and if the tests like it, then so do I. I especially appreciate you cutting down the duplication in Modifier and Innards (which always hurt my eyes a bit, but which I never did anything about).

jgm:
Thanks!

Issue #6315, opened 2020-04-24
Docx writer - support new table features


jgm:
Add support for new table features introduced in
jgm/pandoc-types#66
including table attributes (including identifier), rowspan, colspan, table head and foot, multiple header lines, row headers, captions that allow block-level content and include an optional short caption.

danlobo02:
Is it worth me raising a ticket for open office writer?

tarleb:
I'm starting to work on this now.

tarleb:
I'm wondering how important it is to ensure support for Word 2007? I'm currently writing against the spec and testing with LibreOffice, so I'd miss any incompatibilities. Is this worth considering?

mb21:
From what I hear no version of Word follows their own specs very well? While I'm not sure about Word 2007, maybe we should test with a relatively recent M$ Word?

jgm:
In our current reference.docx we use the 2006 versions of the schemas (that's Word 2007) I guess.
So is your question is whether we should update to a more recent one, or use a more recent one in addition?
A blank document from recent Word has
``` xml
<w:document
xmlns:wpc="http://schemas.microsoft.com/office/word/2010/wordprocessingCanvas"
xmlns:cx="http://schemas.microsoft.com/office/drawing/2014/chartex"
xmlns:cx1="http://schemas.microsoft.com/office/drawing/2015/9/8/chartex"
xmlns:cx2="http://schemas.microsoft.com/office/drawing/2015/10/21/chartex"
xmlns:cx3="http://schemas.microsoft.com/office/drawing/2016/5/9/chartex"
xmlns:cx4="http://schemas.microsoft.com/office/drawing/2016/5/10/chartex"
xmlns:cx5="http://schemas.microsoft.com/office/drawing/2016/5/11/chartex"
xmlns:cx6="http://schemas.microsoft.com/office/drawing/2016/5/12/chartex"
xmlns:cx7="http://schemas.microsoft.com/office/drawing/2016/5/13/chartex"
xmlns:cx8="http://schemas.microsoft.com/office/drawing/2016/5/14/chartex"
xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
xmlns:aink="http://schemas.microsoft.com/office/drawing/2016/ink"
xmlns:am3d="http://schemas.microsoft.com/office/drawing/2017/model3d"
xmlns:o="urn:schemas-microsoft-com:office:office"
[...]

PR #6512, opened 2020-07-08
Docx reader - support new table features

### LINKED ISSUE #6316: Docx reader - support new table features
Add support for new table features introduced in
jgm/pandoc-types#66
including table attributes (including identifier), rowspan, colspan, table head and foot, multiple header lines, row headers, captions that allow block-level content and include an optional short caption.

undergroundquizscene:
I haven't done many of the features yet, but I figure it's good to get something up early to show what I'm doing (and in case I'm heading in wildly the wrong direction).

Will close #6316


src/Text/Pandoc/Readers/Docx/Parse.hs
```diff
@@ -253,7 +253,15 @@ defaultTblLook = TblLook{firstRowFormatting = False}
...
-newtype Cell = Cell [BodyPart]
+data Cell = Cell GridSpan VMerge [BodyPart]
+            deriving Show
+
+type GridSpan = Integer
+
+data VMerge = Continue
+            -- ^ This cell should be merged with the one above it
+            | Restart
+            -- ^ This cell should not be merged with the one above it
```
mb21, line 264: I have no clue about the docx format's way or representing table spans, but I assume you checked whether you really need these intermediate data structures and cannot convert directly to the [pandoc AST table type](https://github.com/jgm/pandoc-types/blob/master/Text/Pandoc/Definition.hs#L214)?

undergroundquizscene: I could probably do without the intermediate data structures, if that'd be preferred – I had decided that it made sense to have an intermediate representation that captures the structure of the docx format, to separate reading the XML from converting from the format into the pandoc AST types.

For example, for the `VMerge` type here, the xml may contain:
- no `vMerge` child, in which case the cell shouldn't be merged with the one above
- a `vMerge` child with no `val` attribute, in which case the cell should be merged
- a `vMerge` child with a `val` attribute of either `continue` or `restart`, meaning merged and not merged respectively

I figured it made sense to separate these XML rules from the conversion of the vMerge data into the pandoc AST type (which I haven't done yet, but I'm assuming will require looking first at the `GridSpan` values to group the cells into columns, followed by an iteration down each column to work out the row spans).

I can see that it could be adding unnecessary code though, so I'm happy to change it. What do you think?

(I'm following [the 5th edition of ECMA-376](http://www.ecma-international.org/publications/standards/Ecma-376.htm) for the format. Let me know if there's another reference I should be using)

mb21: I haven't really looked at the vMerge attributes.. but yes, if it's sufficiently different from rowspan/colspan semantics (which the pandoc AST uses), then perhaps it makes sense to use an intermediate data structure... maybe @jkr has an opinion?

about the word format: we try to be [quite backwards-compatible](https://github.com/jgm/pandoc/issues/5645#issuecomment-511941239)... and usually we rely as much on testing files that were written by Word as the spec, as the two unfortunately are not always in sync :P

undergroundquizscene: Ok, I'll see how it looks as I go and keep both approaches in mind :)

Ah, that makes sense! I'll try and check that I'm keeping things backwards-compatible. Will I need to add any new tests, or should existing tests cover it?

mb21: I don't think we have any docx reader tests that cover row- or col-spans currently... so adding a few would certainly be welcome!

undergroundquizscene: Excellent, I will add some.

undergroundquizscene:
@mb21 Can I ask some advice?

For table captions inserted using Word, there doesn't seem to be an explicit link from each caption to the table it describes – a caption is just a paragraph with a table number field, and references to the table are actually references to this field. The paragraph is automatically inserted next to the table (either before or after it), but can be moved around afterwards. It's also possible to insert multiple captions against the same table, which get sequential numbers – here's a screenshot to show what I mean:

Screenshot 2020-08-31 at 11 59 32

I'm wondering what the best way to link up the captions and tables is. I think the best option would be to link them up sequentially, so the first table caption in the document with the first table in the document and so on, though this would involve dropping captions which reference tables that don't exist (e.g. for the document in the screenshot we'd keep "Table 1, which has stuff in it" and drop the other two captions).

What do you think?

mb21:
we usually try to do whatever makes most sense in regards to most documents found "in the wild", but I'm not sure what's that in this case... do you have a few sample docs lying around, maybe from coworkers?

Another consideration is that round-tripping should work: i.e. from pandoc native to word and back, but since the #6315 is still open, we cannot test this.

undergroundquizscene:
That sounds good, I'll have a look and see what I can find in general use.

I'll keep round-tripping in mind too. I think the approach I'm leaning towards should be fine, since it isn't possible to have a table caption without a table in the pandoc native format.

lrosenthol:
Looking forward to this PR @undergroundquizscene - thanks!

Two other things to keep in mind:
- making sure that internal document links to tables continue to work if you remove a caption. 
- not breaking any of the auto-numbering filters.

undergroundquizscene:
> Looking forward to this PR @undergroundquizscene - thanks!

No problem! I'll get there eventually 😅 

>     * making sure that internal document links to tables continue to work if you remove a caption.

Does this functionality exist in Word? From my experimentation, it seems like the internal links are actually fundamentally links to the captions, so I have a feeling if there's no caption there's no link in the source Word document. I'll test this and see.
 
>     * not breaking any of the auto-numbering filters.

I'm not certain what you mean by "filters" here – do you just mean the auto-numbering of tables/table captions in general?

tshort:
@undergroundquizscene, I tried this branch out. It's nice to be able to convert from MS-Word to HTML now and preserve colspans! I'm really looking forward to this.

One issue I ran into is that for some tables, I lost some of contents of the cells on the bottom of the table. I can send an example if that helps.

undergroundquizscene:
> @undergroundquizscene, I tried this branch out. It's nice to be able to convert from MS-Word to HTML now and preserve colspans! I'm really looking forward to this.
> 
> One issue I ran into is that for some tables, I lost some of contents of the cells on the bottom of the table. I can send an example if that helps.

Thanks for letting me know! I’ve been busy recently, but I’m hoping to make some more progress on this shortly. I _think_ I had noticed what you’re describing at some point, but please do send on an example, it’ll be a big help.

src/Text/Pandoc/Readers/Docx.hs
```diff
@@ -494,15 +497,30 @@ singleParaToPlain blks
...
+rowToRow :: PandocMonad m => Docx.Row -> DocxContext m Pandoc.Row
+rowToRow (Docx.Row _ cells) = do
+  cells' <- traverse (cellToCell (RowSpan 1)) cells
+  return (Pandoc.Row nullAttr (cells'))
+
+rowsToRows :: PandocMonad m => [Docx.Row] -> DocxContext m [Pandoc.Row]
+rowsToRows rows = do
+  let rowspans = (fmap . fmap) RowSpan (Docx.rowsToRowspans rows)
+  let rowCells = fmap (\(Docx.Row _ cs) -> cs) rows
+  cells <- zipWithM (zipWithM cellToCell) rowspans rowCells
```
tshort, line 515: I added a `trace` function here, and I think `rowspans` and `rowCells` have different lengths, so that's why cells at the end are coming up empty.

src/Text/Pandoc/Readers/Docx/Parse.hs
```diff
@@ -250,12 +253,37 @@ newtype TblLook = TblLook {firstRowFormatting::Bool}
...
+
+    colspans :: [Cell] -> [Int]
+    colspans = fmap snd . fst . foldr calculateColspan ([], 0)
+
+    -- TODO: Should the contents of merged cells be combined?
+    calculateColspan :: Cell -> ([(Cell, Int)], Int) -> ([(Cell, Int)], Int)
+    calculateColspan cell@(Cell _ vMerge _) (column, continuesBelow) =
+      case vMerge of
+        Restart -> ((cell, 1 + continuesBelow) : column, 0)
+        Continue -> (column, 1 + continuesBelow)
```
tshort, line 285: I'm a little confused on the naming here. When you have `Restart` and `Continue`, it seems like those are for `rowspan`s, but this is named `calculateColspan`.

src/Text/Pandoc/Readers/Docx.hs
undergroundquizscene: Yup, absolutely! Sorry, I didn't see this, just came back here to point that out. The problem is my `rowsToRowspans` function, which uses `transpose` to look at the columns and then `transposes` again to get back to rows: unfortunately that only works when every row has the same number of elements, which is not true here. I'll write a `transpose` that takes the column spans into account I reckon, that should sort that.

src/Text/Pandoc/Readers/Docx/Parse.hs
undergroundquizscene: Yes, the name is incorrect. When I wrote it I had colspans and rowspans backwards in my head, and when I renamed everything I missed the names inside this function.

src/Text/Pandoc/Readers/Docx.hs
```diff
@@ -660,18 +686,14 @@ bodyPartToBlocks (Tbl cap _ look parts@(r:rs)) = do
...
 
-  let toRow = Pandoc.Row nullAttr . map simpleCell
-      toHeaderRow l = if null l then [] else [toRow l]
+  let toHeaderRow :: Pandoc.Row -> [Pandoc.Row]
+      toHeaderRow row@(Pandoc.Row _ cs) = if null cs then [] else [row]
 
-  -- pad cells.  New Text.Pandoc.Builder will do that for us,
-  -- so this is for compatibility while we switch over.
-  let cells' = map (\row -> toRow $ take width (row ++ repeat mempty)) cells
+  hdrCells <- fmap (>>= toHeaderRow) (traverse rowToRow hdr)
```
undergroundquizscene, line 694: @tshort Also, since the header now allows multiple rows, this should be using `rowsToRows` instead of `rowToRow`, which is no longer needed.

tshort:
@undergroundquizscene, after your recent update, it works great! It properly handled all the rowspans and colspans I've tried it with.

src/Text/Pandoc/Readers/Docx/Parse.hs
```diff
@@ -236,12 +240,82 @@ newtype TblLook = TblLook {firstRowFormatting::Bool}
...
+
+--     -- TODO: Should the contents of merged cells be combined?
+--     calculateRowspan :: Cell -> ([(Cell, Int)], Int) -> ([(Cell, Int)], Int)
+--     calculateRowspan cell@(Cell _ vMerge _) (column, continuesBelow) =
+--       case vMerge of
+--         Restart -> ((cell, 1 + continuesBelow) : column, 0)
+--         Continue -> ((cell, 0) : column, 1 + continuesBelow)
+
+rowsToRowspans :: [Row] -> [[(Int, Cell)]]
+rowsToRowspans rows = let
```
undergroundquizscene, line 280: @mb21 I want to unit test this function, but I can't figure out where to put the tests. I've tried `test/Tests/Readers/Docx.hs`, but I can't import the function in that file. Any idea how I can make that work?

ebahdc:
A status update:

I've completed the following new table features listed in #6316:

- rowspans
- colspans
- multiple header lines
- optional short caption
- captions that allow block-level content
  - this may not currently handle fields (auto-numbered captions) correctly, I need to double-check that (and add a test for it)

I have not yet done:

- row headers
  - I'm not sure what this means, I'll check
- table head and foot
- table attributes
  - as far as I can tell, tables in the docx format don't have identifiers, but I'll check what other attributes there are and see if they're represented in the docx format

undergroundquizscene:
> A status update:
> 
> I've completed the following new table features listed in #6316:
> 
>     * rowspans
> 
>     * colspans
> 
>     * multiple header lines
> 
>     * optional short caption
> 
>     * captions that allow block-level content
>       
>       * this may not currently handle fields (auto-numbered captions) correctly, I need to double-check that (and add a test for it)
> 
> 
> I have not yet done:
> 
>     * row headers
>       
>       * I'm not sure what this means, I'll check
> 
>     * table head and foot
> 
>     * table attributes
>       
>       * as far as I can tell, tables in the docx format don't have identifiers, but I'll check what other attributes there are and see if they're represented in the docx format

(Oops, this is my other account! Was signed in with the wrong one)

tarleb:
This is great! @undergroundquizscene, do you have time to zoom about this some time? I'm working on the writer and would be curious about your input.

undergroundquizscene:
> This is great! @undergroundquizscene, do you have time to zoom about this some time? I'm working on the writer and would be curious about your input.

Sure, let me email you directly and we’ll work out a time

undergroundquizscene:
Ok, I believe this is ready to be merged. I've done what I listed in my previous message, with the exception that there's only partial support for block-level content within captions: putting an ordered list in the caption in Word creates a new paragraph, for example, which is then not included in the caption. However, I think it's an improvement over the current situation, so I think it's worth merging as-is while I work on the other parts.

tarleb (approved):
Wonderful work!

jgm:
We're still getting some CI failures here.

src/Text/Pandoc/Readers/Docx.hs
```diff
@@ -637,49 +662,43 @@ bodyPartToBlocks (ListItem pPr _ _ _ parparts) =
...
-  cells <- mapM rowToBlocksList rows
+bodyPartToBlocks (Tbl cap grid look parts) = do
+  captions <- gets docxTableCaptions
+  fullCaption <- case captions of
+    c : cs -> do
+      modify (\s -> s { docxTableCaptions = cs })
+      return (Just c)
+    [] -> return Nothing
+  let shortCaption = if T.null cap then Nothing else Just (toList (text cap))
+      cap' = caption shortCaption (fromMaybe mempty fullCaption)
```
tarleb, line 677: Would it make sense to default to `plain (text cap)` instead of `mempty` for the full caption?

tarleb:
Any objections if I cherry-pick and amend the commits, so we can get this into the next release?

undergroundquizscene:
> We're still getting some CI failures here.

I'll fix these now: I think it just requires updating the golden docx documents to match the test output.

src/Text/Pandoc/Readers/Docx.hs
undergroundquizscene: Would that not duplicate the caption text (as it's going in the short caption as well)? I don't have the clearest idea of when the full caption or short caption is used.

tarleb:
Could you check that this is ok? https://github.com/jgm/pandoc/compare/master...tarleb:docx-reader-new-table-features?expand=1
I fixed the tests and renamed another such that it's used only by the writer tests. I also took the freedom to use the `caption` field for the long caption, simply because that's how this field was used before AFAIK.

undergroundquizscene:
> Could you check that this is ok? https://github.com/jgm/pandoc/compare/master...tarleb:docx-reader-new-table-features?expand=1
> I fixed the tests and renamed another such that it's used only by the writer tests. I also took the freedom to use the `caption` field for the long caption, simply because that's how this field was used before AFAIK. The commit message is altered slightly as to be more in line with the default style, but I left the explanation in one of them.

Sure, I'll take a look now.

undergroundquizscene:
> Could you check that this is ok? https://github.com/jgm/pandoc/compare/master...tarleb:docx-reader-new-table-features?expand=1
> I fixed the tests and renamed another such that it's used only by the writer tests. I also took the freedom to use the `caption` field for the long caption, simply because that's how this field was used before AFAIK. The commit message is altered slightly as to be more in line with the default style, but I left the explanation in one of them.
> The idea about using `Rational` for column widths should probably be raised here: [jgm/pandoc-types#86](https://github.com/jgm/pandoc-types/issues/86)

Looks good! I actually can't see what you're saying about the caption field though: mind pointing me to the specific bit?

tarleb:
Sorry, I forgot to push that change. It's here now: https://github.com/jgm/pandoc/compare/master...tarleb:docx-reader-new-table-features?expand=1#diff-22df782fded6b5f41521820cbf8de809ecd9835ce95f74fa2ca4eb49e0b0831dR675

undergroundquizscene:
> Sorry, I forgot to push that change. It's here now: https://github.com/jgm/pandoc/compare/master...tarleb:docx-reader-new-table-features?expand=1#diff-22df782fded6b5f41521820cbf8de809ecd9835ce95f74fa2ca4eb49e0b0831dR675

Looks good, thanks :)

undergroundquizscene:
@tarleb Will I close this PR then? I plan to work on more of the docx table features over the next while, but it's probably easier to put them in a new PR rather than this one.

tarleb:
I think so, too. For reference: the commits have been pushed as 44484d0dee1bd095240b9faf26f8d1dad8e560ea and 56b211120c62a01f8aba1c4512acfe4677d8c7d0.

undergroundquizscene:
I saw it automatically closed #6316: we'll probably want to re-open that as some of the table features are not yet supported.

PR #6517, opened 2020-07-09
Lua filters typos

tajmone:
This PR fixes some minor typos in the Lua Filters document.

It also contains an additional commit that fixes the `.gitattributes` file by setting `*.* text eol=lf` to allow Windows users working with an EditorConfig-compatible editor to commit to the repository.

The current `.editorconfig` settings are enforicing `LF` EOLs on all files, which conflicts with Git defaults (`core.eol = native`) which normalizes all text files to the native EOL, which under Windows means that all text files should be checked out using `CRLF`. If end users have Git configured to `core.autocrlf = true` (reccomended) they won't be able to commit edited files, due to EditorConfig settings enforcing `LF` at savetime.

This was just a quick fix to allow me (and possibly others) to contribute to text documents, but the repository settings for EditorConfig and Git attributes should be revisited to ensure an optimal cross-platform collaboration — i.e. by improving either the `.editorconfig` file or the `.gitattributes` settings (possibly both).

I use EditorConfig settings in most of my repositories to enfoce code-consistency validation of every commit and PR (via Travis CI); and I can see that the current `.editorconfig`/`.gitattributes` settings are far from ideal and need some extra attention.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 test/fb2/reader/* -text
+
+# Enforce .editorconfig EOL nomralization rules for Windows OS:
+*.* text eol=lf
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -138,7 +138,7 @@ elements.
 
 ## Filters on element sequences
 
-For some filtering tasks, the it is necessary to know the order
+For some filtering tasks, it is necessary to know the order
 in which elements occur in the document. It is not enough then to
 inspect a single element at a time.
 
@@ -1127,7 +1127,7 @@ A [table cell]{#type-table-cell} is a list of blocks.
 
 *[Alignment]{#type-alignment}* is a string value indicating the
 horizontal alignment of a table column. `AlignLeft`,
-`AlignRight`, and `AlignCenter` leads cell content tob be
+`AlignRight`, and `AlignCenter` leads cell content to be
 left-aligned, right-aligned, and centered, respectively. The
 default alignment is `AlignDefault` (often equivalent to
 centered).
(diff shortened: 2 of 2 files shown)


tarleb:
Thank you. I cherry-picked and pushed the typo-fixing commit.
I'm not sure about the `.gitattributes` changes. If I understand correctly, then we should probably unset the line-endings setting in .editorconfig. I'm also worried that we accidentally force git to treat binary files as text. Could you open a separate PR?

tajmone:
> Thank you. I cherry-picked and pushed the typo-fixing commit.

Good idea! After all, I just needed those changes to be able to commit, and this was the reason I had kept the two commits separate.

> I'm not sure about the `.gitattributes` changes. If I understand correctly, then we should probably unset the line-endings setting in `.editorconfig`. 

I think that the best approach is to use native EOLs unless strictly required (e.g. `LF` for make files or `CRLF` for batch scripts), because Bash for Windows (and it tools) uses the native EOL when applying end-of-line characters, so enforcing `LF` might require running dos2unix on generated files.

> I'm also worried that we accidentally force git to treat binary files as text. 

Yes, that's a risk in some edge cases involving unknown extensions for files that Git might not be able to autodetect as binary. But the main problem with using patterns like `*` in `.editorconfig` is that it will make EditorConfig validators behave erratcly (and treat binary files like text files).

Although it's time consuming, I always add one file type at the time in `.editorconfig`, and ensure that the rules are matched in `.gitattributes` to, but this allows me to validate all commits for code consistency via Travis CI, which prevents commits with spurious content changes. 

> Could you open a separate PR?

Definitely a good idea.

Issue #5948, opened 2019-11-29
Docx writer: Multiple initial list levels not rendered


martinlester:
I am writing a document with nested lists, like this:

````
1. (a) foo

   (b) bar

---

1. asdf

   (a) foo

   (b) bar
````

When I render to a PDF, it looks as I would like. The first line contains "1. (a) foo".

When I render to HTML, there is a line break after "1.". However, the HTML appears semantically to match what I wrote, so I suppose this is just a quirk of HTML rendering.

When I render to docx, the first line is "(a) foo" and the number has disappeared. I am not sure if the effect can be easily achieved in Word, while still using its list facilities. If it can, that would be my preference. If not, maybe something closer to the HTML rendering, which can be achieved at the cost of a line break, would be better?

I am using pandoc 2.8.0.1:

pandoc -o number.docx number.md

[number.docx](https://github.com/jgm/pandoc/files/3905131/number.docx)
[number.pdf](https://github.com/jgm/pandoc/files/3905132/number.pdf)

**Comment by agusmba** [comment]
I don't think word supports that.

Would

```
1. &nbsp;

   (a) foo

   (b) bar
```

yield an acceptable result?

**Comment by martinlester** [comment]
I think that your example produces an acceptable rendering in Word of
my original example. (Or at least, as good as one might hope for,
given Word's apparent limitation here.)

To clarify:

I'd like to get that rendering from my original input, so that I can
get the "right" output where supported (PDF) and the best fallback
where not supported (HTML, docx).

I'm not sure if that would upset anyone who might be relying on the
existing behaviour, although I'd be surprised if it was what they
really wanted.

**Comment by jgm** [comment]
You could use a filter or lua filter to add the nonbreaking space when output is docx.
```
--- fixlist.lua
function OrderedList(el)
  if FORMAT=='docx' then
    if el.content[1] and el.content[1][1].t == 'OrderedList' then
      table.insert(el.content[1], 1, pandoc.Plain{pandoc.Str""})
    end
    return el
  end
end
```

jgm:
I think it would be reasonable to change the docx writer so it does this automatically.

PR #6522, opened 2020-07-10
Make multiple initial list levels work for docx

brisad:
Hi, is this an okay way to solve #5948 ?
I'm not sure how what the best way to test this would be. The existing docx tests seem to use golden copies, is that the way to go here too?


lierdakil:
> The existing docx tests seem to use golden copies, is that the way to go here too?

Most certainly. How else would one check that the docx produced by pandoc matches the expected? Granted, not very flexible, but neither is docx.

The code looks fine (provided Word renders the document as expected), although I'd suggest using record syntax for matching `OrderedList`, i.e. use `OrderedList{}` instead of `OrderedList _ _`. That way even if the number of argument changes, the code will still work.

Also, why `OrderedList` specifically? Doesn't this also happen with `BulletList`? Pretty sure it does.

brisad:
Sorry for the long delay. I have updated the PR with your suggestions and added tests. The resulting docx-file rendered fine in Word.

brisad:
I realized this PR has been lying here for a while now. It is ready for review, how would I best notify someone that it can be looked at? Should I assign someone myself? I'm pinging you here now, but I'm asking for future contributions I hope to make :-) @lierdakil @jgm

tarleb (approved):
Looks good to me.

One could remove a tiny bit of repetition by defining

```haskell
let isListBlock = \case
      BulletList {}  -> True
      OrderedList {} -> True
      _              -> False
```

and using an `if ... then ... else ...` term instead of a `case`. I think that's a little cleaner, but it's a matter of taste.

jgm:
@brisad - up to you whether you want to implement @tarleb's stylistic suggestion.
Let me know and I can merge.

brisad:
@jgm I'll implement the suggestion. Thanks @tarleb!

brisad:
I've updated the PR :)

jgm:
Perfect, thank you!

PR #6527, opened 2020-07-13
[Docx Reader] Only use bCs/iCs on runs with rtl or cs property

### LINKED ISSUE #6514: <strong> tag not closed correctly or invented 😳
I noticed that sometimes the `<strong>` tag is not closed in the correct place or that it is inserted where it does not exist. 😳
It may also happen for other tags (a.e. `<em>`) but for now I have noticed this behavior only on `<strong>`.

The problem occurs with both version 2.9.2.1 than branch with build 2.10 (#6511 ) of @lierdakil  (which handles "negative inline formatting").

The anomalous behavior occurs with both the "-f docx" and "-f docx+styles" option

I am attaching a docx where the problem arises.
(The file was saved in LibreOffice but the original, coming from Word, also had the same problem)

bug_strong

**_Result_**

> <p><em><strong><u>I. Cantico dei Cantici, di Salomone.</u></strong></em> <strong><u>1</u></strong> Mi <strong>baci con i baci della sua bocca! Sì migliore del vino è il tuo amore. Inebrianti sono i tuoi profumi per la fragranza,</strong> Aroma che si spande è il tuo nome: per questo le ragazze di te si innamorano. Trascinami con te, corriamo!</p>

`<strong>` is closed after "fragranza," instead of "bocca!"

> <p><em><strong><u>The Bride's Dream.</u></strong></em> <strong><u>1</u></strong> On my bed by night <strong>I sought him whom my soul loves;</strong> I sought him, but found him not. <strong>I will rise now and go about the city, in the streets and in the squares; I will seek him whom my soul loves.</strong></p>

Inexistent `<strong>` is created on "I will rise [...] loves."

Thank you

Files:
[bug_strong_not_closed_or_not_exist.docx](https://github.com/jgm/pandoc/files/4895727/bug_strong_not_closed_or_not_exist.docx)

lierdakil:
As discussed in #6514.

Apparently, Word will apply `bCs`/`iCs` to a run iff the run has `cs` or `rtl` property set to `true`. This is confirmed by spec. Experiments show that both `cs` and `rtl` can be inherited from styles.

Fixes #6514

**NOTE:** Without #6504 this will likely have some confusing quirks, since I reckon `cs` and `rtl` are likely to be set on paragraph styles, actually.

### DIFF
--- a/src/Text/Pandoc/Readers/Docx.hs
+++ b/src/Text/Pandoc/Readers/Docx.hs
@@ -261,46 +261,43 @@ resolveDependentRunStyle rPr
   | otherwise = return rPr
 
 runStyleToTransform :: PandocMonad m => RunStyle -> DocxContext m (Inlines -> Inlines)
-runStyleToTransform rPr
-  | Just sn <- getStyleName <$> rParentStyle rPr
-  , sn `elem` spansToKeep = do
-      transform <- runStyleToTransform rPr{rParentStyle = Nothing}
-      return $ spanWith ("", [normalizeToClassName sn], []) . transform
-  | Just s <- rParentStyle rPr = do
-      ei <- extraInfo spanWith s
-      transform <- runStyleToTransform rPr{rParentStyle = Nothing}
-      return $ ei . transform
-  | Just True <- isItalic rPr = do
-      transform <- runStyleToTransform rPr{isItalic = Nothing}
-      return $ emph  . transform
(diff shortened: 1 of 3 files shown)


lierdakil:
Uh. CI is apparently broken. I'll try to do something about it shortly.

jgm:
This is great. Thanks for working through that spec!

PR #6528, opened 2020-07-13
[CI] Use setup-haskell action

lierdakil:
This replaces Haskell setup on Linux by a GitHub-provided action https://github.com/actions/setup-haskell. At the moment, it is a bit more robust. On Linux, this action will try hvr-ghc PPA, and if that fails, install via ghcup. Windows and macOS are also supported. Certain GHC/Cabal versions are preinstalled on GitHub Action runners, the action won't reinstall those.

A transient failure on CI prompted this.

~~Actually, I think I will also fix caching while at it. Converting to draft in the meantime.~~ nevermind, it seems mostly fine as is. I could swear I've noticed something weird going on, but apparently not.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,16 +61,10 @@ jobs:
     # need to install older cabal/ghc versions from ppa repository
 
     - name: Install recent cabal/ghc
-      run: |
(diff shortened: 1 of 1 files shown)


jgm:
Great, I didn't know about this!

PR #6544, opened 2020-07-17
Remove duplicate 'titlepage' from epub docs

Blake-Eryx:


jgm:
Thanks for noticing this.  But the man page is autogenerated from MANUAL.txt, so it's that that must be modified.

PR #6552, opened 2020-07-21
Update using-the-pandoc-api.md

favonia:
I noticed that the introduction was outdated by the type change of the `writerTemplate` field. Please feel free to update the wording.


jgm:
great, thanks!

Issue #6562, opened 2020-07-25
added unexpected `.` at the end of a title


rdmueller:
docbook has the `<formalpara>` tag which is a paragraph with a title.

an example, generated by asciidoctor (and manually shortened) is the following docbook file (`demo.xml`):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE book PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN" "http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd">
<?asciidoc-toc?>
<?asciidoc-numbered?>
<book xmlns="http://docbook.org/ns/docbook" lang="en">
<bookinfo>
<title>Demo</title>
<date>2020-07-25</date>
</bookinfo>

<chapter id="some-id">
<title>chapter title</title>

<formalpara>
<title>just a title</title>
<para>lorem ipsum</para>
</formalpara>

</chapter>
</book>
```

when converted via 

`pandoc -r docbook -t docx -o demo.docx demo.xml`

the following output is created:

![image](https://user-images.githubusercontent.com/1856308/88466731-afba5d80-cecf-11ea-829d-a08896d29817.png)

as you can see, pandoc added a `.` behind the title of the `formalpara`. This is unexpected.

I think I've already identified the line of code which is responsible for this:

https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Readers/DocBook.hs#L711

```
pandoc 2.10
Compiled with pandoc-types 1.21, texmath 0.12.0.2, skylighting 0.8.5
```

PR #6563, opened 2020-07-25
remove unexpected dot after formalpara title

rdmueller:
see #6562

### DIFF
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -708,7 +708,7 @@ parseBlock (Elem e) =
         "para"  -> parseMixed para (elContent e)
         "formalpara" -> do
            tit <- case filterChild (named "title") e of
-                        Just t  -> (para . strong . str) <$>
+                        Just t  -> (para . strong . (<> str "")) <$>
                                      getInlines t
                         Nothing -> return mempty
            (tit <>) <$> parseMixed para (elContent e)
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -708,7 +708,7 @@ parseBlock (Elem e) =
         "para"  -> parseMixed para (elContent e)
         "formalpara" -> do
            tit <- case filterChild (named "title") e of
-                        Just t  -> (para . strong . (<> str ".")) <$>
+                        Just t  -> (para . strong . (<> str " ")) <$>
                                      getInlines t
                         Nothing -> return mempty
            (tit <>) <$> parseMixed para (elContent e)
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -708,7 +708,7 @@ parseBlock (Elem e) =
         "para"  -> parseMixed para (elContent e)
         "formalpara" -> do
            tit <- case filterChild (named "title") e of
-                        Just t  -> (para . strong . (<> str " ")) <$>
+                        Just t  -> (para . strong . str) <$>
                                      getInlines t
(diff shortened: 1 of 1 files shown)


rdmueller:
note: I couldn't test the changed code yet but will do asap

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -708,7 +708,7 @@ parseBlock (Elem e) =
         "para"  -> parseMixed para (elContent e)
         "formalpara" -> do
            tit <- case filterChild (named "title") e of
-                        Just t  -> (para . strong . (<> str ".")) <$>
+                        Just t  -> (para . strong . (<> str " ")) <$>
```
alerque, line 711: Why append a space? This should probably be removed entirely, maybe like:

```suggestion
                        Just t  -> (para . strong . str) <$>
```

rdmueller: perfect - I am not a haskell developer and thought my change was on the safe side.
Your suggestion is perfect. Thanx!

rdmueller: hm. seems not to work - breaks tests

mb21:
thanks for the pull! I'm wondering though, there surely was a reason why the dot was added... aren't we breaking another case by just removing it?

rdmueller:
that's what I wondered too. But I couldn't find a reason in the original issue or commit why the dot was appended and it also doesn't break the tests. In addition, I strongly believe that pandoc - as a converter - should not change the content. But it does in this case - not only for the docx output. I also gave it a try with HTML.

rdmueller:
I've now checked the history of the code and can't find a reason for the dot.
But it is somehow funny that the few rendering examples in the definitive all add the dot. But I still think that this is wrong:

https://tdg.docbook.org/tdg/4.5/formalpara.html

In the current version, the title even vanished completly:

https://tdg.docbook.org/tdg/5.1/formalpara.html

jgm:
The dot is probably there to match the rendering example.

jgm:
Leaving the dot makes sense to me, since (a) it provides a break between the title and the rest of the paragraph, and (b) it is present in the documentation's own example.

rdmueller:
what if we make it configurable?
In my eyes, the dot does not make sense because
- it is not part of the content
- the title is already distinguished by another format
- a colon `:` could also make sense if there is no other visual distinguished style between title and paragraph

The dot is also rendered for HTML output. In HTML, I would expect to add it through a CSS-style, if necessary.

If you could point me to an example where a command line option is used, I could try to update my PR to make the dot configurable with `.` as default. This would not break any other installations and give us the opportunity not to use a workaround.

But I would also be ok for me if you close this issue. I would then try not to use `formalpara` in my documents anymore

jgm:
I think it's too cluttered to make this configurable with a command line option.
The way to make it configurable would be to put the title in a specially marked span, so Lua filters can manipulate it (adding or subtracting a dot).

rdmueller:
so, I do understand that this is a "won't fix". Can you please apply the corresponding label and cose this PR and the issue?
thanx!

jgm:
You can close the PR if you like.  But as for the issue, I think it's not quite a "wont fix" -- the course of action I think makes most sense is to enclose the title in a marked Span so that it can be manipulated by a Lua filter, as noted above.  Probably that suggestion should be added to the issue.

nikolaialex:
I really would like to see this PR merged. The current implementation is really annoying. Especially, if there is no reasoning why the dot is added in the first place.

If you cannot merge it, maybe you can provide a filter example that fixes that. 

Thanks in advance.

jgm:
This issue has been closed by e16df8d2718d3a2901c6ac0d7f35165755cde89f
which handles it a bit differently than the patch proposed here.
Thank you.

Issue #6570, opened 2020-07-28
 _Additional _ user-defined variables created in `Meta` doesn't show up in `Pandoc` function later.  i.e., In ODT context, the meta.xml doesn't contain these _additonal_ user variables


kjambunathan:
## Attachment

[meta.xml-bug.zip](https://github.com/jgm/pandoc/files/4989006/meta.xml-bug.zip)

## start in a tmp dir

~$ cd tmp/

## unzip the attachment and place it in tmp

~/tmp$ unzip meta.xml-bug.zip 

Archive:  meta.xml-bug.zip
   creating: meta.xml-bug/
  inflating: meta.xml-bug/works.lua  
  inflating: meta.xml-bug/Makefile   
  inflating: meta.xml-bug/doesnotwork.lua  

## Do a `make`

~/tmp/meta.xml-bug$ make 

Just look at what is shown on the screen.   I have attached a `make` transcript at the end of this bug.  Here is what the above set of files do (or don't do)

---------

In the `works.lua` case, the 

```
$ cat works.lua        

function Meta(meta)
  meta.AAAAAAAAAAAAAAAA='BBBBBBBBBBBBBBBB'
  return meta
end
```
the resulting `works.odt:meta.xml` has the above LibreOffice user variable

```
    <dc:date>2020-07-28T14:40:33Z</dc:date>
    <meta:user-defined meta:name="AAAAAAAAAAAAAAAA" meta:value-type="string">BBBBBBBBBBBBBBBB</meta:user-defined>
  </office:meta>
```
This is good.

But in the `doesnotwork.lua` case, the 

```
$ cat doesnotwork.lua        

function Meta(meta)
  meta.AAAAAAAAAAAAAAAA='BBBBBBBBBBBBBBBB'
  return meta
end

function Pandoc(doc, meta)

  table.insert(doc.blocks, 1, pandoc.RawBlock('opendocument', "<text:p>Hello World</text:p>"))
  table.insert(doc.blocks, pandoc.RawBlock('opendocument', "<text:p>Good Bye!</text:p>"))
  
  return pandoc.Pandoc(doc.blocks, meta)
end
```

the `doesnotwork.odt:content.xml` has `Hello World` and `Good Bye!` added to body text.  This is good. 

```
$ cat content.xml | tail

<office:text>
<text:p>Hello World</text:p>
<text:p text:style-name="Text_20_body">Body text</text:p>
<text:p>Good Bye!</text:p>
</office:text>

```

[...]

PR #6572, opened 2020-07-29
doc/lua-filters.md:  Illustrate how to augment body and meta

kjambunathan:
See https://github.com/jgm/pandoc/issues/6570

The example on metadata is necessary.  

Most of the example on modifying metadata uses the variables from the 'standard' set (like author etc).  

My example clarifies that the user variable need not be one among the 'standard' set.  

I am a programmer, but not a _lua_ programmer.  So, in the issue I cited above--where meta.xml--was rejecting the newly added variable--I was wondering if I can only modify the 'standard' user variables.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -371,6 +371,56 @@ function Meta(m)
 end
 ```
 
+## Adding additional text
+
+This filter adds 'Hello World!' to the beginning of the document, and
+'Good Bye!' to the end of the document.
+
+
+```lua
+function Doc(doc)
+  table.insert(doc.blocks, 1, pandoc.Para{ pandoc.Str 'Hello World!' })
+  table.insert(doc.blocks, pandoc.Para{ pandoc.Str 'Good Bye!' })
+  return pandoc.Pandoc(doc.blocks, doc.meta)
+end
+```
+
+## Adding additional text to only specific output formats
+
+This filter adds 'Hello World!' to the beginning of the document, and
+'Good Bye!' to the end of the document, _only if_ the output format is
+`HTML`.
+
+```lua
+function Doc(doc)
+  table.insert(doc.blocks, 1, pandoc.RawBlock('html', "<p>Hello World</p>"))
+  table.insert(doc.blocks, pandoc.RawBlock('html', "<p>Good Bye!</p>"))
+  return pandoc.Pandoc(doc.blocks, doc.meta)
+end
+```
+
+## Adding a custom user-variable and additonal text to only specific output formats
+
+This filter adds 'Hello World!' to the beginning of the document, and
+'Good Bye!' to the end of the document, _only if_ the output format is
+`odt`.  
+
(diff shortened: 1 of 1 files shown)


doc/lua-filters.md
```diff
@@ -371,6 +371,56 @@ function Meta(m)
...
+```
+
+## Adding additional text to only specific output formats
+
+This filter adds 'Hello World!' to the beginning of the document, and
+'Good Bye!' to the end of the document, _only if_ the output format is
+`HTML`.
+
+```lua
+function Doc(doc)
```
tarleb, line 395: This doesn't quite work as described above, as there are other formats which can include raw HTML. Markdown and Org are examples. Using `if FORMAT:match 'html' then … end` would be closer.

doc/lua-filters.md
```diff
@@ -371,6 +371,56 @@ function Meta(m)
...
+
+## Adding a custom user-variable and additonal text to only specific output formats
+
+This filter adds 'Hello World!' to the beginning of the document, and
+'Good Bye!' to the end of the document, _only if_ the output format is
+`odt`.  
+
+Furthermore, it adds a custom user-variable called 'producer'.
+
+```lua
```
tarleb, line 410: I like this example best, although I'm not sure whether the use of `RawBlock` is aproriate here. If the intention is do demonstrate how to add text to the start and end of a document, a normal Para as in the first example above should do.

Note that the separate filter on Meta is not needed, as one can modify `doc.meta` in the `Pandoc` filter function.

kjambunathan:
Feel free to do whatever feels right for you.  I wouldn't mind if you pull
the PR locally on your machine and force push whatever changes on top of
that.

tarleb:
I decided against including this example. All other examples demonstrate a real-world use case, with only the first example being an exception. The proposed example demonstrates how to use the `Pandoc` filter and constructor, without a direct real-world use case.

The example **is** helpful, but is doesn't has a good place yet in the document. A section on filter functions would be a good place to include an example like this.

kjambunathan:
> I decided against including this example. All other examples demonstrate a real-world use case, with only the first example being an exception. The proposed example demonstrates how to use the `Pandoc` filter and constructor, without a direct real-world use case.
> 
> The example **is** helpful, but is doesn't has a good place yet in the document. A section on filter functions would be a good place to include an example like this.

 If not, it is possible that you have _forgotten_ to push the changes (you  have accepted).

Your comment indicates that you have accepted _a part_ of my proposal.  I don't see a corresponding commit to the master, and _yet the PR is closed_.  I hope you have the changes queued up in your local machine.

kjambunathan:
> It is possible that you have _forgotten_ to push the changes (you have accepted).
> 
> Your comment indicates that you have accepted _a part_ of my proposal. I don't see a corresponding commit to the master, and _yet the PR is closed_. I hope you have the changes queued up in your local machine.

Now, I see these changes in the master branch.  Thanks.

PR #6575, opened 2020-07-29
Lua filters: allow using SimpleTable instead of Table

tarleb:
Old filters using tables now require minimal changes and can use

    if PANDOC_VERSION > {2,10,1} then
      pandoc.Table = function(c,a,w,h,r)
        return pandoc.utils.from_simple_table(pandoc.SimpleTable(c,a,w,h,r))
      end
    end

and

    function Table (tbl)
      smpl_tbl = pandoc.utils.to_simple_table(tbl)
      return pandoc.utils.from_simple_table(smpl_tbl)
    end

to work with the current pandoc version.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -78,6 +78,7 @@
       - Tests.Writers.Native
       - Text.Pandoc.Extensions
       - Text.Pandoc.Lua.Marshaling.Version
+      - Text.Pandoc.Lua.Module.Utils
       - Text.Pandoc.Readers.Odt.ContentReader
       - Text.Pandoc.Readers.Odt.Namespaces
 
--- a/data/pandoc.lua
+++ b/data/pandoc.lua
@@ -1058,6 +1058,30 @@ M.ListAttributes.behavior.__pairs = function(t)
   return make_next_function(fields), t, nil
 end
 
+--
+-- Legacy and compatibility types
+--
+
+--- Creates a simple (old style) table element.
+-- @function SimpleTable
+-- @tparam      {Inline,...} caption    table caption
+-- @tparam      {AlignDefault|AlignLeft|AlignRight|AlignCenter,...} aligns alignments
+-- @tparam      {int,...}    widths     column widths
+-- @tparam      {Block,...}  headers    header row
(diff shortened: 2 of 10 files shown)


tarleb:
A new block type `SimpleTable` would be introduced, effectively behaving like old `Table` blocks (except for the tag).

This is a draft, as I couldn't fully convince myself that this is the best way forward. It requires some unfortunate special-casing, but otherwise seems to work quite well.

jgm:
A simpler approach might be not to add SimpleTable as a faux block element, but instead add a function that treats the element as a simple table.  That is:

``` lua
function Table(el)
  local el2 = to_simple_table(el)
  -- do stuff with el2
  return from_simple_table(el)
end
```
or maybe this could be encapsulated in one function
``` lua
   with_simple_table(el, function(el2) ... end)
```

tarleb:
If I understand correctly, then this would make SimpleTable a standalone type which does not inherit from type Block. We'd be moving the unmarshaling code for SimpleTable into a separate function, which is then exposed to Lua.  That does seem cleaner.

jgm:
I'm sorry, I sort of lost track of this. Do you want to rebase so we can see if the tests pass?

tarleb:
Rebased, squashed, and updated the commit message.

tarleb:
Cabal-based tests fail because

1. there's a problem with text-conversions, which require us to add `constraints: base16-bytestring < 1` to cabal.project and
2. pandoc-citeproc hasn't been updated yet to allow pandoc-types-1.22.

Stack-based tests succeed.

jgm:
What's the problem with text-conversions?
And where do we depend on that library?

jgm:
I've pushed a commit that uses the dev version of pandoc-citeproc.

tarleb:
Thanks!

> What's the problem with text-conversions?

IIUC, then there has been a new release of base16-bytestring which breaks text-conversions, as that package does not set an upper bound for base16-bytestring. [Relevant discussion](https://github.com/haskell/base16-bytestring/issues/14).

> And where do we depend on that library? 

It is a dependency of *doctemplates*.

PR #6578, opened 2020-07-30
Load package `microtype` after `xeCJK` to avoid conflict

likanzhan:
1. This version works:

```
\documentclass{article}
\usepackage{xeCJK}
\usepackage{microtype}
\begin{document}
Hello World
\end{document}
```

2. This version does not work:

```
\documentclass{article}
\usepackage{microtype}
\usepackage{xeCJK}
\begin{document}
Hello World
\end{document}
```

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -176,12 +176,6 @@ $if(outertheme)$
 \useoutertheme{$outertheme$}
 $endif$
 $endif$
-% Use upquote if available, for straight quotes in verbatim environments
-\IfFileExists{upquote.sty}{\usepackage{upquote}}{}
-\IfFileExists{microtype.sty}{% use microtype if available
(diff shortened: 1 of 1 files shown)


jgm:
Some of the tests are failing.
You can do `make TESTARGS=--accept` to update them.  (But do check the diffs carefully to make sure the new test output is correct.)

jgm:
Is there really a problem here?  For example, I can run
```
/usr/local/bin/pandoc -Vmainfont="PingFang HK" -VCJKmainfont="PingFang HK" -o my.pdf MANUAL.txt --pdf-engine=xelatex
```
and it works fine with no error.  ANd I can compile the version 2 above which is said not to work.  Maybe this is all fixed in recent LaTeX package versions?

PR #6597, opened 2020-08-06
[Latex Reader] Table cell parser not consuming spaces correctly

### LINKED ISSUE #6596: LaTeX reader and multirow-multicol problem
from an example in the code:
```
\begin{tabular}{ |l|l|l| }
\multicolumn{2}{c}{One} & Two \\
\end{tabular}
```
does give the desired effect.
However, this one doesn't span and loses the text in the multicol:
```
\begin{tabular}{ |l|l|l| }
Two & \multicolumn{2}{c}{One} \\
\end{tabular}
```
Similarly,
```
\begin{tabular}{ |l|l| }
\multirow{2}{c}{One} & Two \\
a
\end{tabular}
```
gives the desired result, but not the following one (no row span, text lost):
```
\begin{tabular}{ |l|l| }
Two & \multirow{2}{c}{One} \\
a
\end{tabular}
```
(pandoc 2.10.1 on win10x64, but verifiable on the demo page)

I noticed this when I tried to successive multicols and failed, but it appears from the above there is a problem for a right-most multicolumn/multirow (at least that)

LaurentRDC:
This PR fixes #6596 . It also adds a test so this problem does not come up again.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -2387,9 +2387,11 @@ parseTableRow envname prefsufs = do
 
 parseTableCell :: PandocMonad m => LP m Cell
 parseTableCell = do
+  spaces
   updateState $ \st -> st{ sInTableCell = True }
   cell' <- parseMultiCell <|> parseSimpleCell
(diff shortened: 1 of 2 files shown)


jgm:
great, thanks!

PR #6600, opened 2020-08-07
Options.hs: defaultMathJaxURL: use tex-chtml-full instead of tex-mml-…

### LINKED ISSUE #6599: Bug: inlining mathjax 3.0 in html fails to load properly [2.10.1]
It's probably not a pandoc bug (maybe special character escaping or so?)

**Reproduce**:
```shell
echo '$asd$' | pandoc -f markdown+tex_math_dollars --standalone --self-contained --mathjax -t html5 > Content.html
```
**Output**:
Correct HTML which does not load in Chrome properly because mathjax (I assume) has html tags in the source `,r.write('<html><head><`

Can this somehow be fixed? Escping special characters in the source `.js` ??

ickc:
…chtml

closes #6599

c.f. https://docs.mathjax.org/en/latest/web/components/combined.html

Note that while this use the full variant of the js, this drops the mml support. AFAIK, pandoc don't depends on MathJax to process MML, right?

Edit: by using the full it improves latency at the expense of size (but is compensated by dropping MML support from MathJax.)

### DIFF
--- a/src/Text/Pandoc/Options.hs
+++ b/src/Text/Pandoc/Options.hs
@@ -308,7 +308,7 @@ isEnabled :: HasSyntaxExtensions a => Extension -> a -> Bool
 isEnabled ext opts = ext `extensionEnabled` getExtensions opts
 
 defaultMathJaxURL :: Text
-defaultMathJaxURL = "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"
+defaultMathJaxURL = "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-chtml-full.js"
 
 defaultKaTeXURL :: Text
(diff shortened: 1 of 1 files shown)


jgm:
Definitely worth considering, but I'm not positive it has a significant advantage. I tried with `--self-contained`, and although it works I got very ugly math with the wrong fonts.

ickc:
We've discussed this in #2908 already. Full is better in terms of latency with a small penalty of size, which is cached for reused.

And the side effect of it is that you don't have dynamic loaded js so it will be self-contained.

The fonts are ugly because the web fonts is neither available in the local file nor in local computer. Pandoc can be smarter about this. But I'm not sure if it's worth it.

jgm:
Tests seem to be failing -- some golden test output needs to be updated.
`make TESTARGS=--accept` and check the changes.

jgm:
I'll fix the tests after merging.

PR #6601, opened 2020-08-08
Add CSS to default HTML template

mb21:
Two things that inspired this:

- questions like [this one on StackOverflow](https://stackoverflow.com/questions/63293929/pandoc-change-font-family-to-sans-while-converting-from-markdown-to-html)
- I've recently updated and streamlined the CSS I use to generate simple HTML files (or publish to my blog), and was surprised how minimal it can be, while still looking pretty good (IMHO). The idea is to rely on browser defaults as much as possible (which are fairly consistent nowadays), so you don't have to override/normalize a lot.

Maybe this has been discussed before, but what about including something like this in the default HTML template?

~~I also quickly looked into moving this into the `styles.html` partial, but there we don't have access to all those variables I assume?~~

TODO:

- [x] update the MANUAL, only new variable names are `fontcolor`, `backgroundcolor` (and `style`), although I suppose we could add a few more... considering how many we have for the LaTeX template... to allow even more customizations out of the box.
- [x] add a variable to disable outputting that whole CSS block? Perhaps we could also set that variable automatically when `--css` was used? But I cannot imagine any scenario where you'd use `--standalone`, but wouldn't want any CSS whatsoever.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2142,6 +2142,43 @@ ODT or pptx.
 [Unicode Bidirectional Algorithm]: https://www.w3.org/International/articles/inline-bidi-markup/uba-basics
 [Language subtag lookup]: https://r12a.github.io/app-subtags/
 
+### Variables for HTML
+
+`document-css`
+:   Enables inclusion of most of the CSS in the `styles.html`
+    [partial](#partials) (have a look with
+    `pandoc --print-default-data-file=templates/styles.html`).
+    Unless you use [`--css`](#option--css), this variable
+    is set to `true` by default. You can disable it with
+    e.g. `pandoc -M document-css=false`.
+
+`mainfont`
+:   sets the CSS `font-family` property on the `html` element.
+
+`fontsize`
+:   sets the base CSS `font-size`, which you'd usually set
+    to e.g. `20px`, but it also accepts `pt`
+    (12pt = 16px in most browsers).
+
+`fontcolor`
+:   sets the CSS `color` property on the `html` element.
+
+`linkcolor`
+:   sets the CSS `color` property on all links.
+
+`monofont`
+:   sets the CSS `font-family` property on `code` elements.
+
+`linestretch`
+:   sets the CSS `line-height` property on the `html` element,
+    which is preferred to be unitless.
+
+`backgroundcolor`
+:   sets the CSS `background-color` property on the `html` element.
+
+`margin-left`, `margin-right`, `margin-top`, `margin-bottom`
+:   sets the corresponding CSS `padding` properties on the `body` element.
+
 ### Variables for HTML math
 
 `classoption`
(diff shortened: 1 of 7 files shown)


data/templates/default.html5
```diff
@@ -15,6 +15,113 @@ $if(keywords)$
...
+      color: black;
+    }
+    nav a:not(:hover) {
+      text-decoration: none;
+    }
+    #TOC > ul {
+      margin-top: 1em;
+    }
+    #TOC:before {
+      content: 'Contents';
```
mb21, line 120: TODO: read out [our translation](https://github.com/jgm/pandoc/blob/master/data/translations/en.yaml#L6) into a variable and replace it here.

jgm:
I'd have to think more about this, it's an interesting idea.

Note: the styles partial should have access to the same variables.

data/templates/default.html5
```diff
@@ -16,6 +16,7 @@ $endif$
   <title>$if(title-prefix)$$title-prefix$ – $endif$$pagetitle$</title>
   <style>
     $styles.html()$
+    $style$
```
jgm, line 19: IS this change intended?

mb21: yes, see [the MANUAL diff](https://github.com/jgm/pandoc/pull/6601/files#diff-1187c774ee83820d49b1f14f64e3aa4dR2181)

open to discuss all of it :-)

MANUAL.txt
```diff
@@ -2142,6 +2142,56 @@ ODT or pptx.
...
 [Language subtag lookup]: https://r12a.github.io/app-subtags/
 
+### Variables for HTML
+
+`document-css`
+:   Enables inclusion of most of the CSS in the `styles.html`
+    [partial](#partials) (have a look with
+    `pandoc --print-default-data-file=templates/styles.html`).
+    This variable is set to `true` by default, unless you use
+    [`--css`](#option--css).
```
jgm, line 2152: It occurs to me that this would be a breaking change -- currently, the bits of styling inserted by pandoc through styles partial are included even if you specify `--css`, though you can override them using `--css`.  Is there a reason not to keep this behavior?

mb21: The bits that were in the partial already before this change are not part of the `$if(document-css)$` block.. so they'll still always be included.

data/templates/default.html5
mb21: ah, but of course this doesn't work as intended, as it's interpreted as markdown... I'll remove it again. The README read:
> `style` can be used to provide additional CSS,
> potentially overriding the default. For example:
> 
> ```
> ---
> title: my doc
> style: |
>   blockquote {
>     border-left: none;
>   }
> ---
> ```

jgm:
I decided to merge this; comments were favorable.  It was mentioned that this might give rise to endless requests for tweaks in the styling -- that is a danger, but we'll see!

Issue #5541, opened 2019-06-01
ICML writer: Differentiate hyperlinks and cross-references


nathan-artist:
ICML differentiates between hyperlinks (external links) and cross-references (document-internal links). Pandoc's ICML writer does not correctly write document-internal links. I will summarize below the output changes that are needed to fix this; the full input and output files for the markup below can be found in the related topic on the pandoc-discuss list (see [link-citations in ICML writer](https://groups.google.com/d/topic/pandoc-discuss/3QeFuZsMVLo)). It is likely not a simple one-line change but does not require a massive overhaul of the entire writer either, only the part that writes links. I don't know enough about Haskell to fix the writer myself, though I may try if nobody else wants to fix this.

There are essentially four things to change, enumerated below:

1. A `<CrossReferenceFormat>` element should be present just before the `<Story>` element. The following example worked in my test:

```xml
    <CrossReferenceFormat Self="u1" Name="Text Anchor Name">
      <BuildingBlock Self="u1BuildingBlock0" BlockType="BookmarkNameBuildingBlock" CustomText="$ID/" AppliedDelimiter="$ID/" IncludeDelimiter="false" />
    </CrossReferenceFormat>
```

2. Internal-link source points should use a `<CrossReferenceSource>` element, NOT a `<HyperlinkTextSource>` element (which is for external links only). The `Self` attribute of the `<CrossReferenceFormat>` tag (from number 1 above) should be referenced in the `AppliedFormat` attribute of `<CrossReferenceSource>` tags, and the `Name` attribute of `<CrossReferenceSource>` tags should have a relevant value. For example:

- Current incorrect output of internal-link source point:
```xml
  <HyperlinkTextSource Self="htss-1" Name="" Hidden="false">
    <CharacterStyleRange AppliedCharacterStyle="CharacterStyle/Cite Link">
      <Content>2017</Content>
    </CharacterStyleRange>
  </HyperlinkTextSource>
```

- Desired correct output of internal-link source point:
```xml
  <CrossReferenceSource Self="htss-1" AppliedFormat="u1" Name="2017" Hidden="false">
    <CharacterStyleRange AppliedCharacterStyle="CharacterStyle/Cite Link">
      <Content>2017</Content>
    </CharacterStyleRange>
  </CrossReferenceSource>
```

[...]

Issue #684, opened 2012-12-06
Permit adding attributes to all Markdown elements


phyllisstein:
I'd like to suggest that Pandoc allow attributes to be attached to any Markdown element, not just code blocks. So for example, if it encountered this:

```
#{.main} Header
```

it would generate this:

```
<h1 class="main" id="header">Header</h1>
```

rather than this:

```
<h1 id="main-header">{.main} Header</h1>
```

Ditto for, say:

```
Header
----{.main}
```

...which is probably even legible enough to fit with Markdown's philosophical aversion to looking, well, marked-up.

serge-stroobandt:
This issue is also listed as [item 2 on this list](http://www.third-bit.com/2013/06/13/problems-with-pandoc.html). (Disclaimer: I have no relation to that web site.)

fmatheus:
Would be great if any *header_attributes* was supported in *beamer* output.
Looks like {.allowframebreaks} and .fragile by test made in 1.15.0.6.
But why not support arbitrary attributes? In particular *c*, *noframenumbering* and *plain*.

jgm:
+++ fmatheus [Oct 14 15 15:05 ]:
>   Would be great if any header_attributes was supported in beamer output.
>   Looks like {.allowframebreaks} and .fragile by test made in 1.15.0.6.
>   But why not support arbitrary attributes? In particular c,
>   noframenumbering and plain.

Sounds plausible to me.  This just takes a small change in
the LaTeX writer.

jgm:
+++ fmatheus [Oct 14 15 15:05 ]:
>   Would be great if any header_attributes was supported in beamer output.
>   Looks like {.allowframebreaks} and .fragile by test made in 1.15.0.6.
>   But why not support arbitrary attributes? In particular c,
>   noframenumbering and plain.

I've added support for all frame attributes in commit
504bf3f8e79bd502f406264e2cc2794b129a26c0

serge-stroobandt:
> Sounds plausible to me. This just takes a small change in the LaTeX writer.
Attributes to all Markdown elements is not only useful to LaTeX and Beamer.
I personally can think of many use cases for it as XHTML classes.

jgm:
+++ Serge Y. Stroobandt [Oct 16 15 06:24 ]:
>     Sounds plausible to me. This just takes a small change in the LaTeX
>     writer.
>     Attributes to all Markdown elements is not only useful to LaTeX and
>     Beamer.
[...]

PR #6606, opened 2020-08-10
Implement support for internal document links in ICML (#5541)

lrosenthol:
This PR addresses #5541 by supporting internal links when generating ICML.  I added two tests - one for testing that URL links aren't broken and one for testing the new inter-doc functionality.

This is my first PR for Pandoc and my first time writing Haskell, so all comments welcome!

### DIFF
--- a/src/Text/Pandoc/Writers/ICML.hs
+++ b/src/Text/Pandoc/Writers/ICML.hs
@@ -136,10 +136,11 @@ writeICML opts (Pandoc meta blocks) = do
   let colwidth = if writerWrapText opts == WrapAuto
                     then Just $ writerColumns opts
                     else Nothing
-      renderMeta f s = fst <$> runStateT (f opts [] s) defaultWriterState
+      renderBlockMeta f s = fst <$> runStateT (f opts [] s) defaultWriterState
+      renderInlineMeta f s = fst <$> runStateT (f opts [] "" s) defaultWriterState
   metadata <- metaToContext opts
-             (renderMeta blocksToICML)
-             (renderMeta inlinesToICML)
+             (renderBlockMeta blocksToICML)
+             (renderInlineMeta inlinesToICML)
              meta
   (main, st) <- runStateT (blocksToICML opts [] blocks) defaultWriterState
   let context = defField "body" main
@@ -266,9 +267,17 @@ charStylesToDoc st = vcat $ map makeStyle $ Set.toAscList $ inlineStyles st
 
 -- | Escape colon characters as %3a
 escapeColons :: Text -> Text
-escapeColons = Text.concatMap $ \x -> case x of
-  ':' -> "%3a"
-  _   -> Text.singleton x
+escapeColons txt = Text.replace ":" "%3a" txt
+
+-- | figure out the link destination for a given URL
+-- | HyperlinkURLDestination with more than one colon crashes CS6
+makeDest :: Text -> Doc Text
+makeDest txt = literal $
+  if "#" `Text.isPrefixOf` txt
+  then "HyperlinkTextDestination/" <> escTxt
+  else "HyperlinkURLDestination/" <> escTxt
+  where
+    escTxt = escapeColons $ escapeStringForXML txt
 
 -- | Convert a list of (identifier, url) pairs to the ICML listing of hyperlinks.
 hyperlinksToDoc :: Hyperlink -> Doc Text
@@ -277,13 +286,15 @@ hyperlinksToDoc (x:xs) = hyp x $$ hyperlinksToDoc xs
   where
     hyp (ident, url) = hdest $$ hlink
       where
-        hdest = selfClosingTag "HyperlinkURLDestination"
+        hdest = if "#" `Text.isPrefixOf` url
+                then empty
+                else selfClosingTag "HyperlinkURLDestination"
                   [("Self", "HyperlinkURLDestination/"<>escapeColons url), ("Name","link"), ("DestinationURL",url), ("DestinationUniqueKey","1")] -- HyperlinkURLDestination with more than one colon crashes CS6
         hlink = inTags True "Hyperlink" [("Self","uf-"<>tshow ident),  ("Name",url),
                     ("Source","htss-"<>tshow ident), ("Visible","true"), ("DestinationUniqueKey","1")]
                   $ inTags True "Properties" []
                   $ inTags False "BorderColor" [("type","enumeration")] (text "Black")
-                  $$ inTags False "Destination" [("type","object")] (literal $ "HyperlinkURLDestination/"<>escapeColons (escapeStringForXML url)) -- HyperlinkURLDestination with more than one colon crashes CS6
+                 $$ inTags False "Destination" [("type","object")] (makeDest url)
 
 -- | Key for specifying user-defined styles
 dynamicStyleKey :: Text
@@ -297,16 +308,16 @@ blocksToICML opts style lst = do
 
 -- | Convert a Pandoc block element to ICML.
 blockToICML :: PandocMonad m => WriterOptions -> Style -> Block -> WS m (Doc Text)
-blockToICML opts style (Plain lst) = parStyle opts style lst
+blockToICML opts style (Plain lst) = parStyle opts style "" lst
 -- title beginning with fig: indicates that the image is a figure
 blockToICML opts style (Para img@[Image _ txt (_,Text.stripPrefix "fig:" -> Just _)]) = do
-  figure  <- parStyle opts (figureName:style) img
-  caption <- parStyle opts (imgCaptionName:style) txt
+  figure  <- parStyle opts (figureName:style) "" img
+  caption <- parStyle opts (imgCaptionName:style) "" txt
   return $ intersperseBrs [figure, caption]
-blockToICML opts style (Para lst) = parStyle opts (paragraphName:style) lst
+blockToICML opts style (Para lst) = parStyle opts (paragraphName:style) "" lst
 blockToICML opts style (LineBlock lns) =
   blockToICML opts style $ linesToPara lns
-blockToICML opts style (CodeBlock _ str) = parStyle opts (codeBlockName:style) [Str str]
+blockToICML opts style (CodeBlock _ str) = parStyle opts (codeBlockName:style) "" [Str str]
 blockToICML _ _ b@(RawBlock f str)
   | f == Format "icml" = return $ literal str
   | otherwise          = do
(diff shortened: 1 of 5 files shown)


lrosenthol:
I see that I forgot to update the standard ICML tests.  Will do...

src/Text/Pandoc/Writers/ICML.hs
```diff
@@ -270,20 +270,30 @@ escapeColons = Text.concatMap $ \x -> case x of
   ':' -> "%3a"
   _   -> Text.singleton x
 
+-- | figure out the link destination for a given URL
+-- | HyperlinkURLDestination with more than one colon crashes CS6
+makeDest :: Text -> Doc Text
+makeDest x = case Text.uncons x of
+    Just ('#', _) -> literal $ "HyperlinkTextDestination/"<>escapeColons (escapeStringForXML x)
+    _             -> literal $ "HyperlinkURLDestination/"<>escapeColons (escapeStringForXML x)
```
mb21, line 278: welcome to the world of Haskell! :-)
nit-pick, but I'd probably write it more like this:

```suggestion
makeDest :: Text -> Doc Text
makeDest txt = literal $
  if "#" `Text.isPrefixOf` txt
  then "HyperlinkTextDestination/" <> escTxt
  else "HyperlinkURLDestination/" <> escTxt
  where
    escTxt = escapeColons $ escapeStringForXML txt
```

src/Text/Pandoc/Writers/ICML.hs
```diff
@@ -527,11 +537,33 @@ parStyle opts style lst =
...
 
+-- | Replace space characters as _
+replaceSpaces :: Text -> Text
+replaceSpaces = Text.concatMap $ \x -> case x of
+  ' ' -> "_"
+  _   -> Text.singleton x
+
+-- | Create the destination name
+makeDestName :: Text -> Text
+makeDestName name = "#" <> replaceSpaces name
```
mb21, line 548: could use [replace](https://www.stackage.org/haddock/lts/text/Data-Text.html#v:replace):

```suggestion
makeDestName name = "#" <> Text.replace " " "_" name
```

src/Text/Pandoc/Writers/ICML.hs
lrosenthol: I like that syntax - thanks!  Much cleaner (and it explains nicely how `where` works)

src/Text/Pandoc/Writers/ICML.hs
lrosenthol: I was copying a method from later on in the code, so I'll trying replacing that too with replace - perfect!

src/Text/Pandoc/Writers/ICML.hs
```diff
@@ -527,11 +537,27 @@ parStyle opts style lst =
...
+-- | Create the markup for the content (incl. named destinations)
+-- |  NOTE: since we have no easy way to get actual named dests, we just create them for any short content blocks
+makeContent :: Doc Text -> Doc Text
+makeContent cont | isEmpty cont = empty
+              | length (show cont) < 25 = vcat [
+                    selfClosingTag "HyperlinkTextDestination"
+                      [("Self", "HyperlinkTextDestination/"<>makeDestName (render Nothing cont)), ("Name","Destination"), ("DestinationUniqueKey","1")]
+                    , inTagsSimple "Content" $ flush cont
+                  ]
+              | otherwise = inTagsSimple "Content" $ flush cont
```
mb21, line 553: hm... not sure I understand what's going on here... why do add the `HyperlinkTextDestination` only if the content is less than 25 long?

src/Text/Pandoc/Writers/ICML.hs
```diff
@@ -277,13 +285,15 @@ hyperlinksToDoc (x:xs) = hyp x $$ hyperlinksToDoc xs
   where
     hyp (ident, url) = hdest $$ hlink
       where
-        hdest = selfClosingTag "HyperlinkURLDestination"
+        hdest = case Text.uncons url of
+              Just ('#', _) -> empty
+              _             -> selfClosingTag "HyperlinkURLDestination"
```
mb21, line 290: same :)

```suggestion
        hdest = if "#" `Text.isPrefixOf` url
                then empty
                else selfClosingTag "HyperlinkURLDestination"
```

src/Text/Pandoc/Writers/ICML.hs
lrosenthol: The comment above the method says it all
>-- |  NOTE: since we have no easy way to get actual named dests, we just create them for any short content blocks

AFAICT, there isn't anything in the Pandoc AST for a "destination", "bookmark" or "anchor", which is what I need here.   I know I could do it with a custom span attribute that I would read in filter, but here I need it in the "core" writer.   Instead, as a bit of a "hack", I just make any short content fragment as a link destination.

If you have a suggestion about how to set these bookmarks in the source - that I can find in the writer - I'm all ears!!

mb21: yes, for that, the pandoc AST follows the HTML/markdown approach where you just can have an `id` attribute on any element, then link to it with `#`. (and yes, for headers, the `id` are auto-generated..) e.g.

```
$ pandoc -t native
# foo

bar [baz]{#anchor}

^D

[Header 1 ("foo",[],[]) [Str "foo"]
,Para [Str "bar",Space,Span ("anchor",[],[]) [Str "baz"]]]
```

lrosenthol: Thanks - that was 1/3 of what I needed.  The second third was figuring out how that shows up inside the writer itself, and in looking at the HTML and DocX writers, I worked that out (first param to the element!).

Now I have to figure out how to pass it through to the actual content writer...

lrosenthol: Accomplished!

lrosenthol:
@mb21 I've implemented proper support for named destinations/links/anchors (whatever they are called) for most of the standard types.  I didn't do it for images and special divs - and will file a separate issue for those.

Hopefully with these changes, we should be good to merge...

lrosenthol:
@mb21 @jgm What else do you need from me in order to get this PR merged?  I'd like to move forward to addressing other ICML-related issues, and need this in main first (due to all the changes here).    Thanks!

mb21:
I'll try to take a closer look tomorrow... by just skimming the code so far I'm wondering:

- we're trying to have a named destination for all elements for some reason? or only for those that haven `id` attribute in the pandoc AST?
- do we need to make a distinction between what's called in HTML (and the pandoc AST) inline and block elements? (which roughly corresponds to what the ICML writer outputs as having character styles and paragraph styles applied). Because in pandoc, e.g. `Image` is an `Inline` element, but can still have an `id` set on it. Does InDesign support linking to that inline image, or only to the paragraph containing the image?
- just a thought: since we do have `WriterState` in the ICML writer, we could also think of moving things in there instead of passing in as function arguments (that's what I should have done with the `opts` already...). But yeah... in the end it doesn't really matter... the two are fundamentally equivalent, with passing in via arguments more verbose, but easier to understand for Haskell-newcomers...

lrosenthol:
Thanks a lot!  Really appreciate your time!

>we're trying to have a named destination for all elements for some reason? or only for those that haven id attribute in the pandoc AST?

only those that have an `id` attribute in the AST - but is that not all?  I would think that you can apply an `id` to any element - both inline and block level.

> If so, how would we handle pandoc -f html -t icml on this

That's a good question!  I will investigate how InDesign writes that out...

> since we do have WriterState in the ICML writer...

I don't know anything about the WriterState, so I just went with the extra arguments.  If you give me some insight into WriterState - and think that is the better approach - I'll make the changes.  You have more insight into Haskell and the "pandoc way" than I...

lrosenthol:
```
<div id="blockId">
  inline text <span id="inlineId">element</span>
</div>
```

would indeed end up with two `HyperlinkTextDestination` elements, one before `inline text` and one before `element`.

Which is what my code would currently output...

mb21:
> I would think that you can apply an id to any element - both inline and block level.

currently, the pandoc AST is somewhat limited: not all elements have `Attr`... but eventually we'll want to tackle this (#684).

Yeah, moving more stuff into the WriterState is probably best left to another pull request... since `style` and `ident` would both have to go in there... as it's things we collect to eventually output as part of the `CharacterStyles`...

mb21:
and what about a more contrived example?

```
<div id="blockId">
  <div id="blockId2">
    <span id="inlineId">
      <img id="inlineId2" src="dummy.jpg" />
    </span>
  </div>
</div>
```

I guess we should just output four `HyperlinkTextDestination` elements at the same place?

lrosenthol:
@mb21 I added a test with your specific multilevel HTML.  it only puts in the span IDs since destinations are tied to character ranges and not specific blocks (from what I can tell).

lrosenthol:
What else is necessary to get this merged??  @jgm @mb21

mb21:
Okay, yeah... guess that will be a little surprising, but maybe nested ids are a rare enough use-case to not hold up this pull.

It's been quite some time since I worked with ICML, and when I wrote this writer, I was quite new to Haskell... so the longer I look at it, the more things I discover I would do different today... but yes, if you've tested this enough in InDesign, and feel people exporting to ICML could use this, then we should probably merge it :-)

jgm:
This looks good to me.  I think @mb21 is right that using State to keep track of the style and identifier would be more elegant than threading through the parameters -- but of course it amounts to the same thing under the hood.

The one issue remaining is the overlong line in the commit message, flagged by CI above.  But actually it's probably best if you rebase your changes into one commit with a nice message summing up the changes.

lrosenthol:
@jgm will do - thanks!

and I'll also add a new issue #6665 to track doing a code cleanup

lrosenthol:
Ugh!  I just messed up the rebase... @jgm 

Should I just start this PR over off another branch or are you OK with this?

tarleb:
@lrosenthol it appears that there are git conflicts which would prevent merging, so I'm afraid these have to be resolved first. Can you rebase again?

lrosenthol:
OK - had to learn some new git commands, but I believe that everything is now good...watching the checks (definitely pass the commit length, so that's good - and no conflicts with base)

tarleb:
Thanks! The macos CI failure is not related to the changes, it seems to affect all PRs.

lrosenthol:
@jgm back to you for merging...

jgm:
@tarleb I wish I could figure out that macos failure, which keeps happening.
I have been able to get past it by altering stack.yaml, which expires the cache.
But that's ugly.
Maybe we should switch to using cabal for the macos build, to see if that helps.

jgm:
Many thanks and congrats on your first Haskell PR!

Issue #6603, opened 2020-08-09
LaTeX Reader and multirow, part 2


kysko:
Firstly, thanks to LRDC for his latex reader, and fixing a little problem recently.

But there's a few more issues with it that escaped me last time, `multirow` being particularly problematic. I prefer to combine them here rather as separate issues, even if it make a long text.

(Below, I was using *pandoc-nightly-windows-2020-08-08* which includes LRDC's latest fix for issue #6596.)

## multirow's second argument

From the documentation, `multirow`'s second curly-braced argument is `width`, the width in which to set the text.
However it seems the latex reader interprets the second argument as an alignment.
```
\begin{tabular}{r}
\multirow{2}{c}{hello}
\end{tabular}
```
will give the (simplified!) native output:
```
Table
 AlignRight,ColWidthDefault
 TableBody
  Row
   Cell AlignCenter (RowSpan 1) (ColSpan 1) [Plain [Str "hello"]]
```
Notice how the Cell is centered, although I demanded a right align.
The proper notation
```
\begin{tabular}{r}
\multirow{2}{5em}{hello}
\end{tabular}
```
will give a `Para` in the native output, not a `Table`.

Maybe the confusion comes from the fact that in ` \multicolumn{2}{r}`, the second argument *is* an alignment, and perhaps one thought `multirow` behaved the same way.

Note that trying to compile with `pdflatex`, the first example will give out an error ("missing number"), while the second won't (although here also there is something to say below).

## multirow and proper cell inputs in rows
The documentation says:
> You should leave the other rows empty at this column, otherwise the stuff created by `\multirow` will over-write it.

While not explicit, one can verify that litteraly empty cells must be written, as in the following:
```
\begin{tabular}{|r|r|}
\multirow{2}{5em}{hello} & A \\
& B \\
\end{tabular}
```
Notice how the `&` is present to separate from a preceding empty cell.
However, in the tests for the LaTex reader, one can see the following format is used:
```
\begin{tabular}{|r|r|}
\multirow{2}{c}{hello} & A \\
B \\
\end{tabular}
```
that is, by omitting the `&`.

The Native output is almost the same for both inputs, except that the "B" above is missing when written in the correct format.
As for compilation with `pdflatex` (once {c} is replaced by {5em} for example), both compile, but it is evident that the "B" is overwritten in the later format.
(/edited part reverted back, the above stands)

(
/Another edit!: on that note, the last example from the previous issue should then be written
```
\begin{tabular}{r}
[...]

PR #6608, opened 2020-08-10
[Latex Reader] Fixing issues with \multirow and \multicolumn table cells

LaurentRDC:
This PR fixes issue #6603

The main issues that need fixing are:
* [x] `\multirow` is not like `\multicolumn`; the second argument in `\multirow` is the text width;
* [x] Empty cells must be parsed, separated by the usual `&`;
* [x] Technically, the nesting of `\multirow` and `\multicolumn` is not interchangeable. `\multicolumn` must be at the top level.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -2387,9 +2387,11 @@ parseTableRow envname prefsufs = do
 
 parseTableCell :: PandocMonad m => LP m Cell
 parseTableCell = do
+  spaces
   updateState $ \st -> st{ sInTableCell = True }
   cell' <- parseMultiCell <|> parseSimpleCell
   updateState $ \st -> st{ sInTableCell = False }
+  spaces
   return cell'
 
 cellAlignment :: PandocMonad m => LP m Alignment
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -2391,10 +2391,10 @@ parseTableCell :: PandocMonad m => LP m Cell
 parseTableCell = do
   spaces
   updateState $ \st -> st{ sInTableCell = True }
-  cell' <- (    multicolumnCell 
-            <|> multirowCell 
-            <|> parseSimpleCell 
-            <|> parseEmptyCell
+  cell' <- ( multicolumnCell 
+         <|> multirowCell 
+         <|> parseSimpleCell 
+         <|> parseEmptyCell
            )
   updateState $ \st -> st{ sInTableCell = False }
   spaces
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -2383,16 +2383,26 @@ parseTableRow envname prefsufs = do
   cells <- mapM (\ts -> setInput ts >> parseTableCell) rawcells
   setInput oldInput
   spaces
-  return $ Row nullAttr cells 
+  -- Because of table normalization performed by Text.Pandoc.Builder.table, 
+  -- we need to remove empty cells
+  return $ Row nullAttr $ filter (\c -> c /= emptyCell) cells
 
 parseTableCell :: PandocMonad m => LP m Cell
 parseTableCell = do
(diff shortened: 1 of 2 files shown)


LaurentRDC:
I'm not sure how to treat the text width in `\multirow`, so it is currently ignored right now.

I've tested on the snippets from #6603 and they parse correctly.

kysko:
1) > I'm not sure how to treat the text width in `\multirow`, so it is currently ignored right now.

My guess would be to place it somehow in the cell's attributes. In an HTML output it would be a width/min-width like `style="width: 5em"`, so perhaps a keyval of e.g. `{"width", "5em"}`.

In general, since attributes are put in so many items, there would have to be some consensus for readers and writers on what some specific keyvals should mean and optionally be used for.

The special value `*` seem what a writer would naturally do; as for `=` (which seem to say don't adjust, even if the text overflows the cell borders?), not sure either, would have to see examples of why it's used.

2. Another detail from the documentation is the possible square brackets.

For example:
```
\multirow[c]{2}{5em}{some text}
```
Seems they brake the reading, leading to a Para... at least in previous nightly.
I don't know if it is taken care of in this current fix (although not mentioned in my issue).
The first square bracket indicate vertical alignment within the row span.
They should be at least ignored (first and second optional square brackets).
Should I open another issue?

LaurentRDC:
> Should I open another issue?

Let's keep everything related to `\multirow` in this PR.

The full form of `\multirow` is as follows:

```latex
\multirow[vpos]{nrows}[bigstruts]{width}[vmove]{text}
```

Given that the Pandoc abstract syntax tree supports vertical alignment (`vpos`), I think it's a good idea to add support to parse it. I will add it right now.

LaurentRDC:
Actually, as I read the Pandoc AST definition, table cells [can only have horizontal alignment](https://hackage.haskell.org/package/pandoc-types-1.21/docs/Text-Pandoc-Definition.html#t:Alignment). It'll therefore parse correctly but be ignored.

kysko:
yes, and that is why I suggested putting that extra info in a keyval, but it's moot if it means nothing to writers...

ignoring the square brackets is the safe way for now.

thanks

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -2383,16 +2383,26 @@ parseTableRow envname prefsufs = do
   cells <- mapM (\ts -> setInput ts >> parseTableCell) rawcells
   setInput oldInput
   spaces
-  return $ Row nullAttr cells 
+  -- Because of table normalization performed by Text.Pandoc.Builder.table, 
+  -- we need to remove empty cells
+  return $ Row nullAttr $ filter (\c -> c /= emptyCell) cells
```
jgm, line 2388: I'm a little confused by this comment; could you explain what is going on?

jgm: esp. in light of the comment at 2403

LaurentRDC: The comment above at line 2386-2387 refers to the fact that the `Text.Pandoc.Builder.table` function expects the table representation to be different than LaTeX. Consider a 2x2 table, with a 2x1 cell on the left:

```
+---+---+
| 1 | 2 |
+   +---+
| 1 | 3 |
+---+---+
```

`Text.Pandoc.Builder.table` takes in a list of rows, and the first row has an element which spans two rows. Therefore, since its a 2x2 table, only one element is required in the second row. `Text.Pandoc.Builder.table` will take the first element of the second row, and reject all "extras".

Contrast this to LaTeX, which represents the table above as follows:

```latex
\begin{tabular}{cc}
   \multirow{2}{1em}{1} & 2 \\
                                     & 3 \\
\end{tabular}
```
with an empty cell in the bottom left corner. This is what the comment at 2403 is about (as well as #6603) : syntax of LaTeX tables **requires** the presence of empty cells. Table cells that span more than one row/column must be accompanied by appropriate number of empty cells.

Therefore, to "translate" between what is syntactically-correct LaTeX and what `Text.Pandoc.Builder.table` expects, empty cells must be pruned.

jgm: I see, thanks

PR #6617, opened 2020-08-14
LaTeX reader SIrange fix

efharkin:
My previous PR implemented pandoc conversion of LaTeX `\SIRange` commands (https://github.com/jgm/pandoc/pull/6418), but the correct syntax is `\SIrange` (as noted by @pauliacomi [here](https://github.com/jgm/pandoc/pull/6418#issuecomment-674017020); see also this [SO question](https://tex.stackexchange.com/questions/133704/sirange-typeset-different-in-text-and-math-mode/133707) for an example of the correct syntax). That's what I get for implementing that feature on a machine without my usual LaTeX setup... Sorry about that.

This PR introduces a trivial fix by changing all references to `SIRange` to `SIrange` in the current master. Unit tests pass, and I double checked the output of pandoc against pdflatex just to be sure.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -259,9 +259,9 @@ dosiunitx = do
                       emptyOr160 unit,
                       unit]
 
--- converts e.g. \SIRange{100}{200}{\ms} to "100 ms--200 ms"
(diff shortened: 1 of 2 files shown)


jgm:
thanks!

pauliacomi:
Thanks @efharkin, looking forward to trying it out!

PR #6641, opened 2020-08-26
Make the setting-the-date example conditional

the-solipsist:
This makes the example a bit more realistic/valuable by checking if the metadata value "date" is already present, before changing the value.


tarleb:
Thanks!

Issue #6314, opened 2020-04-24
HTML writer - support new table features


jgm:
Add support for new table features introduced in
jgm/pandoc-types#66
including table attributes (including identifier), rowspan, colspan, table head and foot, multiple header lines, row headers, captions that allow block-level content and include an optional short caption.

tarleb:
I'm starting to work on this.

danlobo02:
subscribing

danlobo02:
Is this related or is there a ticket for html reader to support jgm/pandoc-types#66

tarleb:
The ticket for the HTML reader is #6312.

tarleb:
Progress notes: The hardest problem I'm facing is that, in order to apply the correct alignment to each cell, and row number to each row, we need to have a good idea of how the table grid will look. My current line of attack is building a separate T.P.Writers.Tables module which allows to create a rectangular `GridTable`. This should allow to reuse large parts of the current table writer functions, but makes it easy to find a cell's column and row number.

``` haskell
-- | Table row offset (i.e., the number of rows which have to be
-- moved up to find the topmost row belonging to a cell).
newtype RowOffset = RowOffset Int
  deriving (Eq, Num, Enum)

-- | Table column offset (i.e., the number of column which have to
-- be moved left to find the leftmost column belonging to a cell).
newtype ColOffset = ColOffset Int
  deriving (Eq, Num, Enum)

-- | Rectangular table which makes it easy to match table cells
-- with their column and row numbers.
newtype GridTable = GridTable [GridRow]

-- | Single row of a 'GridTable'.
newtype GridRow = GridRow [GridCell]

-- | Single cell of a 'GridTable'. Usually, only cells with zero
-- offsets should be rendered. Other cells serve as placeholders.
data GridCell = GridCell RowOffset ColOffset Cell

toGridTable :: [Row] -> GridTable
toGridTable = undefined -- TODO!
```

This should be useful for other writers as well.

I'd love to learn about alternative approaches and ideas.

jgm:
It would be great to get feedback from @despresc on this.

kysko:
I wrote an [experimental HTML table writer as a Lua filter](https://gist.github.com/kysko/8afb970ad018ae56a6e46e0f44350c99) in the past few weeks, just for the heck of it, knowing it would be useless once there was an official writer for that; so didn't publish it, but maybe now could give ideas. (Sorry if a bit messy.)
(I included some native formats on which I did my testings)

[...]

Issue #6312, opened 2020-04-24
HTML reader - support new table features


jgm:
Add support for new table features introduced in
jgm/pandoc-types#66
including table attributes (including identifier), rowspan, colspan, table head and foot, multiple header lines, row headers, captions that allow block-level content and include an optional short caption.

danlobo02:
subscribing

jgm:
@danlobo02 - note that you can subscribe to issues using the Subscribe button on the right; you don't need to make a comment.

tarleb:
Colspans and rowspans have been added in #6644. Table features which have not been added yet:
  -  intermediate headers
  -  footers
  -  attributes on all elements for which the information is available

tarleb:
I'd be happy to get some opinions on a problem I'm facing: Cell alignment cannot (or should not) be specified for all cells in a column, as the `align` attribute has been deprecated for `<col>` elements. So each cell must specify its own alignment. However, other formats don't support per-cell alignments, and expect column-wise alignment info. Should we try to detect column alignments in the HTML reader?

Put differently, given this table

``` html
<table>
  <tr><td style="text-align: left">1</td><td style="text-align: right">2</td></tr>
  <tr><td style="text-align: left">3</td><td style="text-align: right">4</td></tr>
</table>
```
What should the Markdown output be?

``` markdown
  --- ---
  1     2
  3     4
  --- ---
```
or
``` markdown
  --- ---
  1   2
  3   4
  --- ---
```

I lean towards the first version, as that is also what pandoc produces currently. But I believe that, in the presence of colspans and rowspans, the first version will require table normalization similar to what we do in the HTML writer.

jgm:
Yes, I agree, the first version.
By normalization, do you mean moving from a representation where the alignments are on each cell to one where the cell alignments are default and an alignment is specified on the column?
It seems to me that it would make most sense to do this in the reader, though I'm open to other ideas.

jgm:
@despresc may also have feedback on this.

ickc:
If I understand this question correctly, I think the reader should try its best to represent the input in the pandoc AST, and in this case means it is representing the align per cell, since the AST is capable to do that.

I don’t know much about the normalization already performed in pandoc, in this particular case, is it just checking if the whole column is sharing the same align and if it did then “normalize” it such that there’s no cell alignment but one single column alignment for that column? (And repeat for each column.) I think this kind of normalization probably shouldn’t hurt the performance a lot.

[...]

PR #6644, opened 2020-08-28
Support colspans and rowspans in HTML tables

tarleb:
Add support for cells spanning multiple columns or rows.

See: #6314

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE MultiWayIf          #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -30,6 +32,7 @@ module Text.Pandoc.Writers.HTML (
 import Control.Monad.State.Strict
 import Data.Char (ord)
 import Data.List (intercalate, intersperse, partition, delete, (\\))
+import Data.List.NonEmpty (NonEmpty((:|)))
 import Data.Maybe (fromMaybe, isJust, isNothing, mapMaybe)
 import qualified Data.Set as Set
 import Data.Text (Text)
@@ -53,6 +56,7 @@ import Text.Pandoc.Templates (renderTemplate)
 import Text.Pandoc.Walk
 import Text.Pandoc.Writers.Math
 import Text.Pandoc.Writers.Shared
+import Text.Pandoc.Writers.Tables
 import Text.Pandoc.XML (escapeStringForXML, fromEntities, toEntities,
                         html5Attributes, html4Attributes, rdfaAttributes)
 import qualified Text.Blaze.XHtml5 as H5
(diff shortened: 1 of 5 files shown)


tarleb:
The current method to build a grid table is simple, hacky, and not very efficient. A better method would be to use a zipper, and building the grid diagonally from top left to bottom right.

jgm:
Can you explain the problem that is being solved with this GridTable module?  Isn't the current pandoc-types representation of a Table pretty much isomorphic to HTML table syntax?  Shouldn't it be trivial to render it to HTML by just walking the tree?  I am obviously missing something!

tarleb:
The issue is that we need a way to map cells to their column in order to
apply the proper text alignment. Unlike before, it's not enough to look
at each row individually, but we need to know the actual table layout.

Let's say we have a table

    <table>
      <tr>
        <td rowspan=2>spanning</td>
        <td>one</td>
      </tr>
      <tr>
        <td>two</td>
      </tr>
    </table>

with

    [(AlignLeft, ColWidthDefault), (AlignCenter, ColWidthDefault)] :: [ColSpec]

We won't know in which column the cell in row two should be placed, and
which alignment to apply to the cell, unless we look at all previous
rows. That's what I'm using GridTable for.

jgm:
I see; I guess I was thinking, wrongly, that the alignments could be specified in the colgroup at the top.

tarleb:
I thought the same and am still confused as to why the `align` attribute on `col` elements is deprecated (and not even supported on Firefox).

despresc:
I suppose it might be nice to have the `table` builder also infer what the individual cell alignments are, since it already has to go through the whole table to fix cell dimensions and padding. That would have the nice side effect of making the cell alignment behaviour consistent across formats that support it.

The issue there, I think, is that some of the writers (LaTeX in particular) might want to know if the cell's alignment is *consistent* with the column alignments so that they don't output unnecessary cell alignment overrides. So the `Alignment` part of a `Cell` would need to be something like `(Alignment, NeedsOverride)` for that to work.

despresc:
Though `toLegacyTable` (and so all the writers) already normalize their input `Table`s, so constructing an intermediate table with explicit alignment information would not be too different.

mb21:
>  Shouldn't it be trivial to render it to HTML by just walking the tree?

I was hoping that as well.... so should we change the AST again, or...?

despresc:
The `GridTable` that's constructed doesn't really need to include the coordinates of the cell itself, at least not for the HTML writer, from what I understand of it. You only really need to add to the original `Cell` its inferred alignment, or maybe the `[Alignment]` list of the columns that the cell spans, and whether or not the cell is in the stub (the row head in a `TableBody`), though that might be better accomplished by creating a `GridBodyRow` that separates the stub cells in the row from the rest. That should be all that's necessary, and that data can be extracted from the table using the method that [placeRowSection](https://hackage.haskell.org/package/pandoc-types-1.21/docs/Text-Pandoc-Builder.html#v:placeRowSection) uses to lay out a table row on a grid for normalization.

A similar choice, but stylistically different, would be to stick a few helper functions somewhere that look something like

```haskell
renderTableHead
  :: Monad m
  => ([Alignment] -> Cell -> m a) -- ^ render a cell given alignments of spanned columns
  -> (Int -> Attr -> [a] -> m b) -- ^ render a row given the row number and rendered cells
  -> (Attr -> [b] -> m c) -- ^ render the entire head
  -> [ColSpec] -> TableHead -> m c
```

the idea being that the function also takes care of the table layout work like the construction of the `GridTable` does. The corresponding `renderTableBody` function would need to be more complex, since table bodies have more components to handle.

------

All of this information can be recorded directly in the `Table` type, and can be pretty easily produced using a modified `table` builder. I think the reason why it was decided not to do so originally is that the extra cached data in the table can become incoherent or ill-formed, making the `Table` type even looser than it is already. Separating out the stub cells in each `TableBody` row means that the column width of the stub might vary within a single `TableBody`, for instance. But maybe that kind of redundancy is better for ease of use?

Otherwise, if you'd like, I can write the `GridTable` constructor myself, since I'm probably the most familiar with how the method in `toLegacyTable` and the `normalizeTable` functions works.

tarleb:
> Otherwise, if you'd like, I can write a GridTable and related constructor myself for that data, since I'm probably the most familiar with how the method in toLegacyTable and the normalizeTable functions works.

That'd be great! I've written a less naïve toGridTable function in the meantime, but I'd much prefer a central function build with your expertise.

despresc:
I wrote an alternate `Writer.Tables` in https://github.com/jgm/pandoc/pull/6655, if you would like to look.

tarleb:
I rebased to the current master, but now the new "planets" test table is missing a column. I suspect that there be an issue with the table annotator. @despresc, can you help?

tarleb:
Found the problem, misunderstanding on my part.

tarleb:
I think this is ready to be merged now.

jgm:
I assume there are still some new table features in #6312 that are not yet addressed by this PR?

tarleb:
Still missing:

- intermediate headers
- footers
- attributes on all elements for which the information is available

I'll add a note to the issue.

PR #6655, opened 2020-09-01
Add Writers.Tables helper functions and types, add tests for those

despresc:
(I'm putting this here for comment. Not sure how this should be handled regarding https://github.com/jgm/pandoc/pull/6644.)

`Writers.Tables` includes an `AnnTable` type and a `toAnnTable` function that takes a `Table` and adds row numbers to rows, column numbers and the data of spanned columns to cells, and separates the row head cells from the row body cells. I think that should be all the HTML writer needs to render a table with a simple traversal. It should also contain the information needed for other writers to render a table without having to re-lay it on a grid, hopefully.

I'm interested in how ergonomic the `AnnTable` type is. The definitions are, when stripped of documentation:

```haskell
data AnnTable = AnnTable Attr Caption [ColSpec] AnnTableHead [AnnTableBody] AnnTableFoot
data AnnTableHead = AnnTableHead Attr [AnnHeaderRow]
data AnnTableBody = AnnTableBody Attr RowHeadColumns [AnnHeaderRow] [AnnBodyRow]
data AnnTableFoot = AnnTableFoot Attr [AnnHeaderRow]
data AnnHeaderRow = AnnHeaderRow Attr RowNumber [AnnCell]
data AnnBodyRow = AnnBodyRow Attr RowNumber AnnRowHead AnnRowBody
newtype RowNumber = RowNumber Int
type AnnRowHead = [AnnCell]
type AnnRowBody = [AnnCell]
data AnnCell = AnnCell (NonEmpty ColSpec) ColNumber Cell
newtype ColNumber = ColNumber Int
```

The `toAnnTable` function takes the components of a `Table` and constructs an `AnnTable` out of it. It's written defensively now: it doesn't assume that the `Table` components in its input are well-formed, so it normalizes them like the `table` builder. I'm not sure if that's desired, but it does protect the writers from ill-formed filter input.

I've added various example and property tests that `toAnnTable` behaves correctly. One of them, that it normalizes tables exactly like the `table` builder, is currently failing because the `table` builder is (I think) currently too strict. A pull request (https://github.com/jgm/pandoc-types/pull/77) should make this test succeed if it's merged. Otherwise I'll have to make a small change to `AnnTableBody` and `toAnnTable` to fix the test.

One last question: should rows be numbered continuously from the very start of a table? I think they would be, at least so that the even/odd pattern of rows isn't interrupted by table boundaries. That's what is assumed in `toAnnTable`, in any case.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -12,3 +12,8 @@ source-repository-package
     type: git
     location: https://github.com/jgm/pandoc-citeproc
     tag: 0.17.0.2
+
+source-repository-package
+    type: git
+    location: https://github.com/jgm/pandoc-types
+    tag: 8e9ca37802120f32ececac752d73463e2fc86811
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -553,6 +553,7 @@ library
                    Text.Pandoc.Writers.Math,
                    Text.Pandoc.Writers.Shared,
                    Text.Pandoc.Writers.OOXML,
+                   Text.Pandoc.Writers.Tables,
                    Text.Pandoc.Lua,
                    Text.Pandoc.PDF,
                    Text.Pandoc.UTF8,
@@ -820,6 +821,7 @@ test-suite test-pandoc
                   Tests.Writers.Powerpoint
                   Tests.Writers.OOXML
                   Tests.Writers.Ms
+                  Tests.Writers.Tables
   if os(windows)
     cpp-options:      -D_WINDOWS
   default-language: Haskell2010
--- a/src/Text/Pandoc/Writers/Tables.hs
+++ b/src/Text/Pandoc/Writers/Tables.hs
@@ -38,6 +38,7 @@ module Text.Pandoc.Writers.Tables
 where
 
 import           Control.Monad.RWS.Strict
+                                         hiding ( (<>) )
 import           Data.Generics                  ( Data
                                                 , Typeable
                                                 )
--- a/src/Text/Pandoc/Writers/Tables.hs
+++ b/src/Text/Pandoc/Writers/Tables.hs
@@ -1,0 +1,290 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TupleSections #-}
+
+{- |
+   Module      : Text.Pandoc.Writers.Tables
+   Copyright   : Copyright 2020 Christian Despres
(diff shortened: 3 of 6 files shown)


tarleb:
Looks good to me! I didn't have the chance yet to rebase my PR, but I expect it to be completely straight-forward.

jgm:
For future reference, an annotation `[API change]` should always be included when there's an API change, so we can record this in the changelog.  I could have added it but forgot.

despresc:
I will remember that. I suppose that's the case for `pandoc-types` too?

jgm:
Yes, that too.  Generally I will notice if something involves an API change, but it's helpful if these are marked really clearly in the commit messages, since one thing I need to do in changelogs is catalog the API changes.

tarleb:
Now that I actually started using it, I do have one suggestion: instead of making `AnnHeaderRow`, `AnnFooterRow` and `AnnBodyRow` separate types, it would be more convenient if we changed them into a single `AnnRow` type together with `data RowType = HeaderRow | FooterRow | BodyRow`. While this would allow to make invalid tables representable, I wouldn't expect bugs from this and it would shave off a bit of boilerplate in the writers.

tarleb:
To add a bit of bike-shedding: how would everyone feel about renaming the module to `AnnotatedTable` and to drop the "Ann" prefix from types? It might look a bit nicer if the module was imported qualified as `Ann`, so types would be referenced as `Ann.Cell` etc. But that, I guess, is just a matter of personal preference.

despresc:
Sure, the module could be renamed to `AnnotatedTable`. I can do that after your `Writers.HTML` request is merged, and qualify what's in there (should just be a search and replace). Or would you like me to change it in your branch before merging?

I put in a separate `AnnBodyRow` to distinguish the row header cells from the row body cells in that bit of the table. Would merging them into one type save a lot of effort? I think `RowType` would need a `BodyRow Int` branch to count the number of header cells in the row, at least.

Or did you mean the separate `TableHead` etc. types? I'm not sure those can be gotten rid of, because of the problem of having a row in the overall head (should go in `<thead>`) next to an intermediate head row (should go in a `<tbody>`).

(Sorry I didn't notice this until now).

despresc:
One possibility is to get rid of `HeaderRow`, and only have the existing `BodyRow` (renamed to just `Row`). That would allow row header cells to be separated from the row body, and allow a single function to operate on both. It's just that in a `TableHead` the row body part would be empty.

tarleb:
Thanks, that sounds good. I think I'd prefer to merge the HTML improvements first; maybe it would even make sense to wait until we've tackled another writer. That would give us some more insights into potential causes of boilerplate. I can open a new issue once we get there.

jgm:
sounds good to me

Issue #6538, opened 2020-07-16
asciidoctor images


docsAsCode:
In asciidoc (or asciidoctor) format we can use two kind of images : inline or block.
inline image syntax is : 
```
Click image:icons/play.png[Play, title="Play"] to get the party started.
```

block image syntax is (double `:`) : 
```
image::sunset.jpg[Sunset]
```

block image can be considered as figure in other language (rst, docbook).

In pandoc 2.10 only inline image is generated. Like you generate figure in rst or docbook, can you generate block image in asciidoc ?

Thx

mb21:
hm... and how should we decide whether to render inline or block?

jgm:
We can do what we do in other formats.
If the image is the only element in a Para, and has a title beginning with the prefix `fig:`, then we render it as a figure, stripping the prefix from the title.

kukimik:
Hello! I'd be happy to work on this issue. The change to make looks rather easy and isolated, but I'm new to pandoc and want to understand what's going on here, so this may take me a while.

tarleb:
Great! Don't hesitate to reach out if you have questions or run in to problems.

argent0:
@kukimik  I made the pull-request before seeing your claim. Sorry

kukimik:
@argent0 That's ok. I'll find myself another issue to work on. :wink:

docsAsCode:
Hello,

Happy to see the progress for my issue.

I forget to say block images support horizontal alignement: 

`image::sunset.jpg[Sunset,align=center]`

tarleb:
It may be better to raise this as a new issue once this one is closed.

PR #6671, opened 2020-09-08
Asciidoctor images

argent0:
Addresses  #6538 by implementing jgm's [proposed solution](https://github.com/jgm/pandoc/issues/6538#issuecomment-659536590).

I had to change `writer.asciidoctor` to pass the tests.

Concretely:

Given the file: `my-test.markdown`

```{markdown}
Inline-style: ![alt text](media/rId25.jpg)

Block image:


![alt text 2](media/rId26.jpg "image title")

The end.
````

This patch generates, after running:

```
$ stack run pandoc -- -f markdown test/my-test.markdown -t asciidoctor -o -
```

The following file:

```
Inline-style: image:media/rId25.jpg[alt text]

Block image:

image::media/rId26.jpg[alt text 2,title="image title"]

The end.

```
I think the current basic test doesn't distinguish between inline and block figures. Should the distinction be added to the basic test?

As mentioned in  #3177, this fix is brittle for the time being.

I have only just seen kukimik's comment.

### DIFF
--- a/src/Text/Pandoc/Writers/AsciiDoc.hs
+++ b/src/Text/Pandoc/Writers/AsciiDoc.hs
@@ -66,7 +66,8 @@ writeAsciiDoc opts document =
 -- | Convert Pandoc to AsciiDoctor compatible AsciiDoc.
 writeAsciiDoctor :: PandocMonad m => WriterOptions -> Pandoc -> m Text
 writeAsciiDoctor opts document =
-  evalStateT (pandocToAsciiDoc opts document) defaultWriterState{ asciidoctorVariant = True }
+  evalStateT (pandocToAsciiDoc opts document)
+    defaultWriterState{ asciidoctorVariant = True }
 
 type ADW = StateT WriterState
 
@@ -141,7 +142,8 @@ blockToAsciiDoc opts (Plain inlines) = do
 blockToAsciiDoc opts (Para [Image attr alternate (src,tgt)])
   -- image::images/logo.png[Company logo, title="blah"]
   | Just tit <- T.stripPrefix "fig:" tgt
-  = (\args -> "image::" <> args <> blankline) <$> imageArguments opts attr alternate src tit
+  = (\args -> "image::" <> args <> blankline) <$>
+    imageArguments opts attr alternate src tit
 blockToAsciiDoc opts (Para inlines) = do
   contents <- inlineListToAsciiDoc opts inlines
   -- escape if para starts with ordered list marker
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by argent0** [comment]
I see the `asciidoc` tests failing. According to https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#images
this is valid `asciidoc` syntax. So I will update the `asciidoc` test too.

**Review by tarleb — COMMENTED** [review]
Thank you for your contribution, the result of this looks good. Below I list are a few things that came to mind, in no particular order.

Maybe checkout the type classe `IsString`, of which `Doc Text` is an instance. It allows to write a value as if it was a string. (E.g. `Text 3 "foo"` can be shortened to just `"foo"`).

Pattern matching on the specific `Doc` constructors is clever, as it produces very compact code. But it is also fragile and would break if the return value changed, e.g. by changing the order in which values are concatenated. It would be more robust to construct the result directly. One option could be to construct a new image helper function which both `blockToAsciiDoc` and `inlineToAsciiDoc` can use.

**Comment by argent0** [comment]
I implemented @tarleb 's proposed refactoring of the common image code into a helper function that is called by `blockToAsciiDoc` & `inlineToAsciiDoc`.

I've run into some trouble though. The command

```
$ stack run pandoc -- -f markdown test/my-test.markdown -t asciidoctor -o -
```

generates:

```
Inline-style: image:media/rId25.jpg[alt text]                                                      
                                                                                                   
Block image:                                                                                       
                                                                                                   
image::media/rId26.jpg[alt text 2,title="image title"]Another paragraph.                           
                                                                                                   
The end.  
```

Notice the: "Another paragraph" at the the end of the block image.

Yet, the native output seems fine:

```
$ stack run pandoc -- -f markdown test/my-test.markdown -t native -o -
```
```
[Para [Str "Inline-style:",Space,Image ("",[],[]) [Str "alt",Space,Str "text"] ("media/rId25.jpg","
")]                                                                                                
,Para [Str "Block",Space,Str "image:"]                                                             
,Para [Image ("",[],[]) [Str "alt",Space,Str "text",Space,Str "2"] ("media/rId26.jpg","fig:image ti
tle")]                                                                                             
,Para [Str "Another",Space,Str "paragraph."]                                                       
,Para [Str "The",Space,Str "end."]]  
```

As the "Another paragraph" is in its own `Para`. :thinking: 

This is the input file `my-test.markdown`:

```
Inline-style: ![alt text](media/rId25.jpg)

Block image:


![alt text 2](media/rId26.jpg "image title")


Another paragraph.

The end.
```

Any idea, 1) if this is an issue; 2) where is the code to solve it?

Regards

tarleb:
Great, looking good! I believe it should be enough to append a `blankline` to the result in the block. Similar to what happens in the last line of `blockToAsciiDoc` for `Para` elements.

tarleb:
Almost ready to merge, just one tiny nitpick remaining: would you break the long lines 144 and 559 such that it has 80 chars or less? We try to keep all source lines below that limit.

I think it could also be clearer to use `Attr` instead of `(Text, [Text], [(Text, Text)])` in the type signature.

tarleb (approved):
Thanks, this looks good to me!

Leaving it to @jgm to double-check and merge.

argent0:
Thank you for all the help @tarleb.

jgm:
Looks good, thanks for the PR!

PR #6680, opened 2020-09-12
Fix hlint suggestions, update hlint.yaml

despresc:
Since `.hlint.yaml` is in the repo, I thought I might fix the suggestions that it gives.

One of these changes is slightly non-trivial, changing
```haskell
  where
    parseEmptyCell = optional spaces >> return emptyCell <* optional spaces
```
to
```haskell
  where
    parseEmptyCell = spaces $> emptyCell
```

in `Readers.LaTeX` (line 2190). Since `spaces` is just the normal parsec one re-exported from `Parsing`, that should be equivalent.


jgm:
I'm pretty impressed that hlint saw that those were equivalent.

despresc:
That was me, actually. `hlint` gave a different suggestion for that line, and I noticed the redundancy.

despresc:
(That was the only non-hlint thing done, other than adding some modules to "ignore: use camelCase").

Not totally sure why it's failing to build.

jgm:
Build failures are because of warnings (CI uses `-Werror` for the pandoc package).
For example:
```

160
src/Text/Pandoc/Readers/LaTeX.hs:34:1: warning: [-Wunused-imports]
161
    The import of ‘<$’ from module ‘Data.Functor’ is redundant
162
   |
163
34 | import Data.Functor (($>), (<$))
164
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
You can look through the build logs if you follow the Details links for the CI.

despresc:
I see, that's a bit earlier in the run. For some reason `stack` and `ghcid` aren't reporting that warning to me. Normally I fix them all.

jgm:
You can get different warnings for different compiler versions, so that may be the issue.

PR #6682, opened 2020-09-13
OpenDocument writer: New table cell support with column spans

pyssling:


jgm:
You may want to consider using the new Text.Pandoc.Writers.AnnotatedTable.
I know that this was needed to get things write for the HTML writer.  (The main issue, as I recall, is accurately discerning column alignments when you have colspans and rowspans.)
See discussion at https://github.com/jgm/pandoc/pull/6644
I'm not sure if this affects OpenDocument, but since it's an XML format I thought the same issue may arise.

pyssling:
@jgm I can have a look - but this really doesn't implement rowspans yet and I would prefer to tackle that separately as the need arises if that's ok. I think this is at least a step forward compared to the current state of things.

jgm:
Yes, I think the problem that module solves only emerges if you have rowspans.

But if you use that module now, it would make the path to supporting rowspans in the future much easier.

pyssling:
@jgm  I've updated the code to use the Annotated table and also implemented row spans. I think it should be good to go now.

jgm:
Awesome, thanks!

Issue #6662, opened 2020-09-04
Change markdown writer to use ATX headings by default


jgm:
After consensus reached on pandoc-discuss.

- [ ] Change `writerSetextHeaders` in default `WriterOptions` to False.
- [ ] Add `--setext-headings` option to change this option.
- [ ] Still respond to `--atx-headers`, as a no-op, for compatibility; but deprecate.
- [ ] Adjust T.P.App.Opt to add new option and change default.
- [ ] Adjust any tests that assume the other default.

the-solipsist:
A quick question: Why is `--markdown-headings=setext` a commandline option rather than a markdown extension (i.e., something like `-t markdown+setext`)?

jgm:
Because setext headings are part of the core default markdown features.

alerque:
This is a great change and I'm grad to see it happen, but just a postmortem on the release process:

This should have been at least a minor release bump, not a patch release bump. I use Pandoc in a lot of workflows for normalizing content formatting and would not have expected upgrading a patch version to cause a significant change in the default output formatting.

Nothing to do now that the harm is done, but maybe keep it in mind for next time.

tarleb:
Pandoc uses the [Haskell Package Versioning Policy](https://pvp.haskell.org/), so the jump from 2.11.1 to 2.11.2 *was* a minor release bump.

mb21:
Should we add a link to the Haskell PVP in https://pandoc.org/releases.html ?

jgm:
Good suggestion: I've added an explanation of the PVP to releases page.

PR #6693, opened 2020-09-21
Change writerSetextHeaders in default to False.

argent0:
Working on  #6662

First step

### DIFF
--- a/src/Text/Pandoc/Citeproc.hs
+++ b/src/Text/Pandoc/Citeproc.hs
@@ -532,7 +532,9 @@ deNote (Note bs:rest) =
       = Cite cs (concatMap noteInParens ils) : go zs
   go (x:xs) = x : go xs
   needsPeriod [] = True
-  needsPeriod (Str t:_) = not (T.null t) && isUpper (T.head t)
+  needsPeriod (Str t:_) = case T.uncons t of
+                            Nothing    -> False
+                            Just (c,_) -> isUpper c
   needsPeriod (Space:zs) = needsPeriod zs
   needsPeriod _ = False
   noteInParens (Note bs')
@@ -555,6 +557,25 @@ deNote (x:xs) = x : deNote xs
 removeFinalPeriod :: [Inline] -> [Inline]
 removeFinalPeriod ils =
   case lastMay ils of
+    Just (Span attr ils')
+      -> initSafe ils ++ [Span attr (removeFinalPeriod ils')]
+    Just (Emph ils')
+      -> initSafe ils ++ [Emph (removeFinalPeriod ils')]
+    Just (Strong ils')
+      -> initSafe ils ++ [Strong (removeFinalPeriod ils')]
+    Just (SmallCaps ils')
+      -> initSafe ils ++ [SmallCaps (removeFinalPeriod ils')]
     Just (Str t)
       | T.takeEnd 1 t == "." -> initSafe ils ++ [Str (T.dropEnd 1 t)]
+      | isRightQuote (T.takeEnd 1 t)
+        -> removeFinalPeriod
+             (initSafe ils ++ [Str tInit | not (T.null tInit)]) ++ [Str tEnd]
+             where
+               tEnd  = T.takeEnd 1 t
+               tInit = T.dropEnd 1 t
     _ -> ils
+ where
+  isRightQuote "\8221" = True
+  isRightQuote "\8217" = True
+  isRightQuote "\187"  = True
+  isRightQuote _       = False
--- a/test/command/pandoc-citeproc-82.md
+++ b/test/command/pandoc-citeproc-82.md
@@ -22,7 +22,7 @@ Title
 
 Some text.[^1]
 
-[^1]: Comment regarding text, supported by citation [@OCLC_i1099]
+[^1]: Comment regarding text, supported by citation [@OCLC_i1099].
 ^D
 Title
 =====
@@ -36,5 +36,5 @@ OCLC. "WorldCat." Accessed September 19, 2014.
 :::
 :::
 
-[^1]: Comment regarding text, supported by citation (OCLC, "WorldCat.")
+[^1]: Comment regarding text, supported by citation (OCLC, "WorldCat").
 ```


argent0:
Hi, the tests for `lhs` to `lhs` fail in an unexpected way

```
lhs to lhs:              FAIL (0.02s)
          
          --- lhs-test.markdown+lhs
          +++ /home/aner/source/git/pandoc/.stack-work/dist/x86_64-linux-tinfo6/Cabal-2.4.0.1/build/pandoc/pandoc --data-dir=../data --quiet lhs-test.native --wrap=preserve -r native -s -w markdown+lhs
          -   2 ========
```

Where I would expect it to fail like:

```
+ 1 # lhs test
- 1 lhs test
- 2 ======
```

Somehow the `#` disapears. :thinking: Any Idea what might be happening in this case?

If I understand correctly this is reading the `lhs-test.native` file, converting it to `markdown+lhs` and comparing the result to `lhs-test.markdown+lhs`.

* A test for the new `--setext-headings` would be necessary. Should I add it? (Any pointers?)
* We are using "headings" in `--setext-headings` and headers in `--atx-headers`. Should we use one word, both in the code and the flags?
* The code, after this patch, sets `optSetextHeaders` ("headers" used here) to `False`. Should the option be replaced to better reflect the new default? I.e. `optUseAtxHeaders` or something.

Regards

argent0:
The failed test:

```
ipynb
      reader:                                                        OK (0.02s)
      writer:                                                        FAIL (0.03s)                                                                                                                                                          
········
        --- ipynb/simple.ipynb
        +++ /home/aner/source/git/pandoc/.stack-work/dist/x86_64-linux-tinfo6/Cabal-2.4.0.1/build/pandoc/pandoc --data-dir=../data --quiet ipynb/simple.in.native -f native -t ipynb-raw_html-raw_tex+raw_attribute -s
        +  26     "## Pyout"
        -  26     "## Pyout\n"
```
is also unexpected. Why is there no `\n` at the end? :thinking:

jgm:
> is also unexpected. Why is there no \n at the end at this particular instance? 🤔

This is because in the ipynb output, the markdown is divided into lines, and the newline is omitted on the last line.
(I don't know why they do it that way, but that's how they do it.)

jgm:
> Somehow the # disapears. 🤔 Any Idea what might be happening in this case?

Could be a bug in the test setup?  That's really puzzling.  I can't look into it right now though.

> A test for the new --setext-headings would be necessary. Should I add it? (Any pointers?)

Something in `test/command` would be the easiest.

> We are using "headings" in --setext-headings and headers in --atx-headers. Should we use one word, both in the code and the flags?

We've been transitioning from "headers" to "headings," esp. in the public-facing parts (documentation, command-line options), but we've left "headers" in some places for backwards compatibility.  So I'm thinking "headings" is better for the new option, even though it's a bit odd to have both.  Note that after this change, the only command-line options with "headers" (in this sense) will be deprecated ones.

We could introduce `--atx-headings` as a synonym for `--atx-headers`, I suppose.

One question is whether `--atx-header(ing)s` should be deprecated, even if it's a no-op.  Maybe it's helpful to have an option that makes intentions clear.

 > The code, after this patch, sets optSetextHeaders ("headers" used here) to False. Should the option be replaced to better reflect the new default? I.e. optUseAtxHeaders or something.

`optSetextHeaders` is left over from an earlier time when ATX headings were the default!  In fact it makes more sense with this change than `optUseAtxHeaders` would.

Bigger question:  Perhaps instead of `--setext-headings` and `--atx-headers/headings`, we shold have:
```
--markdown-headings=setext|atx
```
I like that better, as it makes it clearer what syntax this affects.  If we did this, it would probably be good to change the Opt structure accordingly, and to make sure `markdown-headings: setext` works in defaults files.

argent0:
I'll see to:

- [ ]  implement `--markdown-headings=setext|atx`
- [ ]  remove `--setext-headings`
- [ ]  keep the deprecated `--atx-headers` and point to the new flag format

argent0:
I implemented the `--markdown-headings=setext|atx` flag. The default is `atx` style. It can be explicitly passed `--markdown-headings=atx` in case you want to future proof your script. The `--atx-headers`  is deprecated and recommends using the new flag.

argent0:
**About the failing `ipynb` reader test.**

```
--- ipynb/simple.out.native
        +++ /home/aner/source/git/pandoc/.stack-work/dist/x86_64-linux-tinfo6/Cabal-2.4.0.1/build/pandoc/pandoc --data-dir=../data --quiet ipynb/simple.ipynb -f ipynb-raw_html-raw_tex+raw_attribute -t native -s
```

This pull request's version generates the `Meta` present in `simple.ipynb`. That meta is absent in `test/ipynb/simple.out.native`. So I think the test should be updated. I saw this with vimdiff. But here is the test's output diff anyway:

```
        --- ipynb/simple.out.native
        +++ /home/aner/source/git/pandoc/.stack-work/dist/x86_64-linux-tinfo6/Cabal-2.4.0.1/build/pandoc/pandoc --data-dir=../data --quiet ipynb/simple.ipynb -f ipynb-raw_html-raw_tex+raw_attribute -t native -s
        +   1 Pandoc (Meta {unMeta = fromList [("jupyter",MetaMap (fromList [("kernelspec",MetaMap (fromList [("display_name",MetaString "Python 3"),("language",MetaString "python"),("name",MetaString "python3")])),("language_info",MetaMap (fromList [("codemirror_mode",MetaMap (fromList [("name",MetaString "ipython"),("version",MetaString "3")])),("file_extension",MetaString ".py"),("mimetype",MetaString "text/x-python"),("name",MetaString "python"),("nbconvert_exporter",Met
aString "python"),("pygments_lexer",MetaString "ipython3"),("version",MetaString "3.8.5")])),("nbformat",MetaString "4"),("nbformat_minor",MetaString "5")]))]})
        -   1 Pandoc (Meta {unMeta = fromList [("jupyter",MetaMap (fromList [("nbformat",MetaString "4"),("nbformat_minor",MetaString "5")]))]})
```

For example check the `kernelspec` tag, which is absent in `simple.out.native`.

jgm:
I'm confused about why a change of default from setext to atx headers in the markdown writer has any ramifications at all for metadata in ipynb.  Can you explain this?

argent0:
I'm having multiple issues with this:
1. the `ipynb` writer
2. its reader

Issue 2: With the reader: this PR writes metadata present in the source; absent in the test. So maybe this was an issue "slipping through the cracks".

> I'm confused about why a change of default from setext to atx headers in the markdown writer has any ramifications at all for metadata in ipynb. Can you explain this?

Finally, to address your question: 

```
pandocToNotebook = do
  ...
  let blockWriter bs = literal <$> writeMarkdown
           opts{ writerTemplate = Nothing } (Pandoc nullMeta bs)
  ...
  metadata' <- toJSON <$> metaToContext' blockWriter inlineWriter
                 (B.deleteMeta "nbformat" .
                  B.deleteMeta "nbformat_minor" $
                  jupyterMeta)
```

So `metadata'` depends on `writeMarkdown`
Then at `pandocToMarkdown`:

```
  metadata <- metaToContext'
               (blockListToMarkdown opts)
               (inlineListToMarkdown opts)
               meta
```

So it calls: `blockListToMarkdown` which calls `blockToMarkdown` where there are changes for the headings.

I'll look around those functions to see if I can spot anything.

argent0:
OK, so the metadata "changes" were introduce when I used jupyter notebook to check that I had valid output for ipynb. I think I should make the tests read only hereafter.

The markdown output for markdown+lhs is:

```
stack run pandoc -- --data-dir=../data --quiet --wrap=preserve -r native test/lhs-test.native -s -w markdown+lhs
lhs test

`unsplit` is an arrow that takes a pair of values and combines them to
return a single value:

> unsplit :: (Arrow a) => (b -> c -> d) -> a (b, c) d
> unsplit = arr . uncurry
>           -- arr (\op (x,y) -> x `op` y)

`(***)` combines two arrows into a new arrow by running the two arrows on a
pair of values (one arrow on the first item of the pair and one arrow on the
second item of the pair).

    f *** g = first f >>> second g

Block quote:

 > foo bar
```

That is: There are no headings. AFAIK markdown headings can't have leading spaces and `#` in the first column of a `lhs` causes problems:

```
7 15051 0 tmp $ stack ghc Main.lhs

Main.lhs:3:3: error: lexical error in pragma at character 'T'
  |
3 | # This is the police
  |   ^
8 15052 0 tmp $ cat Main.lhs 
> module Main where

# This is the police

> main = undefined
```

Perhaps the combination of `markdown+lhs` and `atx` headings should result in an error. Because the resulting file doesn't have `atx` markdown headings.

Let me know what you think.

jgm:
> Perhaps the combination of markdown+lhs and atx headings should result in an error.

That's a good thought.  Maybe it would make more sense to issue a warning in the markdown writer when we "downshift" an ATX heading to a paragraph.  Something like: "Rendering ATX heading as a paragraph, because # in the left-hand column is interpreted as CPP in lhs output. Consider using --markdown-headings=setext"  If they don't have any headings, they won't get the warning and things will be fine.  IF they do, they can react as they like.

jgm:
This would require adding a new constructor to LogMessage in Logging.

argent0:
## Details

I added the constructor 

```{haskell}
-- UnaccurateRepresentation block format message
UnaccurateRepresentation Block Text.Text Text.Text
```

That means that `Block` can't be represented accurately using `format` with a final message.

## Resulting Warning

```
$ stack run pandoc -- --data-dir=../data --wrap=preserve -r native test/lhs-test.native -s -w markdown+lhs
[WARNING] Format markdown+lhs can't accurately represent `Header 1 ("lhs-test",[],[]) [Str "lhs",Space,Str "test"]`
  Leading # is used for CPP pragams. Rendering as paragraph. Consider using --markdown-headings=setext.
lhs test
#
`unsplit` is an arrow that takes a pair of values and combines them to
return a single value:

> unsplit :: (Arrow a) => (b -> c -> d) -> a (b, c) d
> unsplit = arr . uncurry
>           -- arr (\op (x,y) -> x `op` y)

`(***)` combines two arrows into a new arrow by running the two arrows on a
pair of values (one arrow on the first item of the pair and one arrow on the
second item of the pair).

    f *** g = first f >>> second g

Block quote:

 > foo bar
```

jgm:
I'd prefer the constructor to be more targeted and the output simpler:

`CannotRenderASTHeading`

Render as: `"Rendering heading as paragraph, because # is not allowed in column 1 of lhs files.\nConsider using --markdown-headings=setext to force setext-style headings."`

argent0:
Hi, I've changed the constructors and error messages as requested. Is this change still going to be implemented? Please don't hesitate to  mention any issue with the code.

Ping (1/3)

jgm:
Yes!  I do plan to merge this.  But I'm holding off a bit, because this is an API change that would require a major version bump, and I want to make sure we don't need more point releases for 2.11 first.

I'll make a couple very minor comments, but this is looking good.

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -529,9 +529,26 @@ options =
 
     , Option "" ["atx-headers"]
                  (NoArg
-                  (\opt -> return opt { optSetextHeaders = False } ))
+                  (\opt -> do
+                    deprecatedOption "--atx-headers"
+                        "This is now `--markdown-headings=atx`"
```
jgm, line 534: Let's remove the backticks; that is markdown syntax and people
may not know what to make of it in a warning message.

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -529,9 +529,26 @@ options =
...
 
+    , Option "" ["markdown-headings"]
+                (ReqArg
+                  (\arg opt -> do
+                    headingFormat <-
+                      case arg of
+                        "setext" -> pure HeadingFormatSetext
+                        "atx" -> pure HeadingFormatAtx
+                        _ -> E.throwIO $ PandocOptionError $ T.pack
+                          ("Unknown markdown heading format: " ++ arg)
```
jgm, line 546: Add: "expecting setext or atx"?

src/Text/Pandoc/App/Opt.hs
```diff
@@ -130,7 +131,8 @@ data Opt = Opt
     , optPdfEngine             :: Maybe String -- ^ Program to use for latex/html -> pdf
     , optPdfEngineOpts         :: [String]   -- ^ Flags to pass to the engine
     , optSlideLevel            :: Maybe Int  -- ^ Header level that creates slides
-    , optSetextHeaders         :: Bool       -- ^ Use atx headers for markdown level 1-2
+    -- | Heading style for markdown level 1-2
+    , optMarkdownHeadingFormat :: MarkdownHeadingFormat
```
jgm, line 135: Instead of the comment on line 134, put a comment on line 135 using `-- ^ ` as above.

src/Text/Pandoc/Logging.hs
```diff
@@ -98,6 +98,8 @@ data LogMessage =
   | RunningFilter FilePath
   | FilterCompleted FilePath Integer
   | CiteprocWarning Text.Text
+  -- | UnaccurateRepresentation of block when writing format.
```
jgm, line 101: Let's remove this line.

src/Text/Pandoc/Logging.hs
```diff
@@ -378,3 +386,4 @@ messageVerbosity msg =
        RunningFilter{}               -> INFO
        FilterCompleted{}             -> INFO
        CiteprocWarning{}             -> WARNING
+       CannotRenderATXHeading{} -> WARNING
```
jgm, line 389: Align `->` with the above.

src/Text/Pandoc/Options.hs
```diff
@@ -254,7 +271,8 @@ data WriterOptions = WriterOptions
   , writerListings          :: Bool       -- ^ Use listings package for code
   , writerHighlightStyle    :: Maybe Style  -- ^ Style to use for highlighting
                                            -- (Nothing = no highlighting)
-  , writerSetextHeaders     :: Bool       -- ^ Use setext headers for levels 1-2 in markdown
+  -- | Header format to use in markdown
```
jgm, line 274: Instead of this line, use `-- ^` on the line with the option, as in the other cases.

src/Text/Pandoc/Options.hs
```diff
@@ -267,39 +285,40 @@ data WriterOptions = WriterOptions
...
+                  , writerColumns               = 72
+                  , writerEmailObfuscation      = NoObfuscation
+                  , writerIdentifierPrefix      = ""
+                  , writerCiteMethod            = Citeproc
+                  , writerHtmlQTags             = False
+                  , writerSlideLevel            = Nothing
+                  , writerTopLevelDivision      = TopLevelDefault
+                  , writerListings              = False
+                  , writerHighlightStyle        = Just pygments
+                  , writerMarkdownHeadingFormat = HeadingFormatAtx
```
jgm, line 311: Shouldn't the default be Setext?  (In principle we could have a different default than with command line pandoc, but I think it makes sense to keep the same default.)

jgm:
Note:  any changes in the public-facing API should be marked as "[API change]" in the commit messages.  This would include new constructors for exported types, new fields in exported record types, etc. If this isn't marked in commit messages, I may forget to note it in the changelog.

src/Text/Pandoc/App/CommandLineOptions.hs
argent0: I used your proposed message. Another one I thought was: "Valid options are: setext and atx". Both messages convey the same information.

src/Text/Pandoc/App/Opt.hs
argent0: Done. I was thinking about the 80 char line width guideline. Also I was trying to avoid modifying the spacing before the '::'s.

jgm: I'm sorry to reverse myself on this, but on reflection, I think that it's a bad idea to break the public API for aesthetic reasons. Let's just keep `optSetextHeaders` and use that.  Everything else can stay the same, and then we can integrate this change without a breaking API change.

Also, is it possible to rebase your changes on master so that they are easier to integrate, and possibly collapse them into a small number of commits?

argent0: > I think that it's a bad idea to break the public API for aesthetic reasons. Let's just keep optSetextHeaders and use that.

I'm not sure I understand what you mean.

In this patch
1. `atx-headings` are the new default markdown headings.
2. `--atx-headers` is deprecated.
3. `--markdown-headings=setext|atx` is the new way of controlling the markdown headings.

Did you mean to revert any of those? Or you meant keeping those but implement them using optSetextHeaders? I think you mean the latter.

> Also, is it possible to rebase your changes on master so that they are easier to integrate, and possibly collapse them 
> into a small number of commits?

I'll see if I can to collapse some commits on test cases.

jgm: I guess what I had in mind was avoiding the switch from `optSetextHeaders` to `optMarkdownHeadingFormat`.  That change would require a bump to 2.12.
I'm actually sort of conflicted on this.  I'm not ready to go to 2.12 yet, but otherwise we could merge this change.  On the other hand, the change does improve clarity.  (There may also be ramifications for the JSON and YAML instances that I'm not thinking of.)

jgm: In any case, collapsing to a small number of commits makes it easier to rebase in the future.

jgm: I can't remember whether your changes included also a change to `writerSetextHeaders` in `WriterOptions`.

Seems to me that if we're going to make a big API change, we might consider changing that as well as `optSetextHeaders`.

But maybe the principle of avoiding API changes except when necessary favors just leaving these types as they are?

argent0: Very well, I'll keep `optSetextHeaders` and make `----markdown-headings=setext|atx` modify it accordingly.

PR #6706, opened 2020-09-28
OpenDocument writer: Use references for internal links

pyssling:
Use references for internal links in cases where the link does
not contain custom text.

### DIFF
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -134,6 +134,7 @@ data Extension =
     | Ext_raw_html            -- ^ Allow raw HTML
     | Ext_raw_tex             -- ^ Allow raw TeX (other than math)
     | Ext_raw_markdown        -- ^ Parse markdown in ipynb as raw markdown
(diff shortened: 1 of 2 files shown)


pyssling:
Hi @jgm and @mb21 ,

This isn't ready for merging - it's rather something I need direction with.
Currently the OpenDocument writer uses links always - even for internal references. This commit enables it to use internal references instead of links and if the native_numbering extension is enabled it will also prefix references with chapter numbers.

The problems are that:
1. References ignore whatever link text may have been provided - instead they reuse the chapter title unconditionally. This is probably the wanted behavior in a lot of cases, but not all.
2. The format when numbering is enabled is "1.2.3 Title of chapter" which is not everyone's cup of tea.
3. For figures it includes the whole caption - which could be really really long.

I don't know if these issues have turned up in other writers, I looked at the Latex one for example, and as far as I can see it always ignores link text (but that was just from a quick look) and does whatever latex does according to styling (normally just a number is my guess). Any feedback on dealing with this issue much appreciated. Maybe the extension could take an argument like a "format string"?

pyssling:
Closed - there is a newer version which is now complete.

Issue #6710, opened 2020-10-01
Syntax highlighting implementation for OpenDocument/ODT output


niszet:
Hi,

I would like to use Pandoc's ODT output like a Docx output as much as possible. Syntax highlighting is one of missing feature of ODT output.

From an advice from [pandoc-discuss,](https://groups.google.com/g/pandoc-discuss/c/zo6uV9FRATU/m/4LPtpCqZBwAJ
) I update Writer/OpenDocument.hs to implement syntax highlighting for ODT. Its code are based on Docx's one. Thank you for the advice.

I will send a PR and I would appreciate if someone review it.

But, there are several steps to implement syntax highlighting for ODT output. I did only first part of followings.

* [x] Update `Code` of `Writer/OpenDocument.hs` for highlighting 
* [x] Update `CodeBlock` of `Writer/OpenDocument.hs` for highlighting (with changing paragraph structure like a docx style)
* [x] Update `Writer/ODT.hs` for generating styles into style.xml from skylighting.

ODT's CodeBlock has a different structure from Docx's CodeBlock. For implementing highlighting for CodeBlock, structure of paragraphs should be changed to Docx's structure (`CodeBlock` will be converted to `Code`s inside of `Para`)

Highlighting styles from skylighting should be generated automatically and it should be incorporated into styles.xml. To implement it, `Writer/ODT.hs` need to have `styleToOpenDocument` based on Docx's `styleToOpenXml`.

I will continue implement aboves. Please let me know if you have any opinions.

Thanks,
niszet

niszet:
#6711

jgm:
so far, so good!

niszet:
Thank you for merging! and sorry for bothering you by my PRs... 
PR #6717 is for checking a concept of modification (lack of updating of tests). (But now, I notice "Source_Code" should be "Source_20_Code". I should update again.)

Except above styles name issue, objectives of updates are noted in  https://github.com/jgm/pandoc/pull/6717#issue-497370008 .

Because styles will be updated from Writer/ODT.hs, this update does not contain styles file. But my local work (I put style file by my hand.), this code works as similar to docx's `CodeBlock` as below. (SourceCode paragraph and Code+LineBreak lines and it seems good.)

![image](https://user-images.githubusercontent.com/28832585/95006711-2303ec80-0642-11eb-947e-33525debdc85.png)

[...]

PR #6716, opened 2020-10-04
Syntax highlight for CodeBlock of OpenDocument

niszet:
By #6711, syntax highlighting for inline Code has been implemented.
Syntax highlighting for CodeBlock is also implemented based on Docx Writer.

* Paragraph style of CodeBlock is changed from "Preformatted_20_Text" to "Source_Code".
* As similar to docx writer, CodeBlock is decomposed to "Code + LineBreak" lines.
* To do this, Code is updated again to treat multiple lines.
* Styles will be incorpolated by Writer/ODT.hs

Further discussion will be written in #6710.


niszet:
Sorry, I will send a PR again.

PR #6718, opened 2020-10-04
Sort languages in `--list-highlight-languages` output

tarleb:
Languages appear to be sorted by their long name, which leads to
unexpected results: e.g., the long name of *m4* is *GNU m4*, so it is
listed between *gnuassembler* and *go*.


jgm:
Good idea

PR #6753, opened 2020-10-16
Man page: remove org from list of input formats supporting raw_tex

nfleisher:
Update pandoc's man page to reflect the fact that org is no longer a supported input format for the raw_tex extension

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2861,7 +2861,7 @@ their respective sections of [Pandoc's Markdown]:
   for the following formats (in addition to `markdown`):
 
   input formats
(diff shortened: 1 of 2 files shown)


tarleb:
Thank you for catching that! The man page is generated from MANUAL.txt, can you change it in there instead?

nfleisher:
Sure thing! What is best practice here: should I modify this pull request, or create a new one instead?

tarleb:
Both works, but it's easier for us if you modify this one. You can just add commits, GitHub makes it easy to squash them into one when merging.

PR #6754, opened 2020-10-16
DocBook reader: bibliomisc and anchor support

pyssling:
Support for bibliomisc and anchors which are used to link into
bibliomisc. Also refactor the bibliography into a bulletin list
(each under a bibliodiv if present). This is more consistent with
other DocBook behavior than just using paragraphs for each entry.


jgm:
I'm not sure about using bullet lists for the bibliography. In general, this looks bad -- hardly any real bibliography styles would want bullets.  Can you link to something that explains the behavior you're trying to be consistent with?

pyssling:
Asciidoctor default output - but honestly - the bulleted list is unimportant to me. I can as well go for distinct paragraphs - or please suggest something.

pyssling:
Ok, I reverted to the previous behavior. If someone in my organization wants it in a bulletin list I think we can do that through a custom filter or similar. Now it's just paragraphs.

jgm:
thanks!

PR #6755, opened 2020-10-16
OpenDocument Writer: Implement span with ident support

pyssling:
Spans containing an ident in the Attr will become bookmarks
in OpenDocument format.

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -188,15 +188,23 @@ formulaStyle mt = inTags False "style:style"
                                                   ,("style:horizontal-rel", "paragraph-content")
                                                   ,("style:wrap",           "none")]
 
+inBookmarkTags :: Text -> Doc Text -> Doc Text
+inBookmarkTags ident d =
+  selfClosingTag "text:bookmark-start" [ ("text:name", ident) ]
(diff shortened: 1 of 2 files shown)


jgm:
Could you do the same for Divs?

pyssling:
Please don't merge yet - I need to verify this a bit.

pyssling:
Ok, this should be correct - and is in fact better than what I hoped for - Divs in Pandoc are now (if they contain an ident) mapped to a section in OpenDocument. This is very clean and seems to be completely consistent with the intention of Div.

jgm:
Divs can be used for a number of things that might not be considered sections -- e.g., sidebars.  I don't know what the semantics of "section" is in opendocument, though -- if it's generic enough, this is probably right.

pyssling:
It seems to be very generic (though of course with more uses than this one):

The <text:section> element represents a named portion of content in a document.

Sections specify formatting properties for a region of text or  text that is automatically acquired from an external data source or document, or another text section.

https://docs.oasis-open.org/office/OpenDocument/v1.3/cs01/part3-schema/OpenDocument-v1.3-cs01-part3-schema.html#element-text_section

pyssling:
@jgm this one should be ready to go now - added a test case.

jgm:
looks good!

PR #6759, opened 2020-10-18
Implement data-external attribute support for HTML <img> element

### LINKED ISSUE #6543: Respect data-external="1" attribute also for native images
**Version**: `2.9.2.1`

**Input**
```
<html>
<head>
	<title>placeholder</title>
</head>
<body>
	<img data-external="1" src="https://static.fimfiction.net/images/logo-2x.png">
</body>
</html>
```

**Output**
```
> pandoc -f html -t epub -o test.epub test.html
> unzip -l test.epub 
Archive:  test.epub
  Length      Date    Time    Name
---------  ---------- -----   ----
       20  2020-07-17 22:35   mimetype
      251  2020-07-17 22:35   META-INF/container.xml
      160  2020-07-17 22:35   META-INF/com.apple.ibooks.display-options.xml
     1400  2020-07-17 22:35   EPUB/content.opf
      767  2020-07-17 22:35   EPUB/toc.ncx
      541  2020-07-17 22:35   EPUB/nav.xhtml
      496  2020-07-17 22:35   EPUB/text/title_page.xhtml
      914  2020-07-17 22:35   EPUB/styles/stylesheet1.css
    10638  2020-07-17 22:35   EPUB/media/file0.png <----------- logo-2x.png
      564  2020-07-17 22:35   EPUB/text/ch001.xhtml
---------                     -------
    15751                     10 files
```

brisad:
This makes it possible to use the `data-external` attribute with `` elements when converting HTML input to EPUB, just like it is done for raw HTML.

Using bash:
```
$ pandoc -f html -t epub -o test.epub <<< '<html><head><title>placeholder</title></head><body><img data-external="1" src="https://static.fimfiction.net/images/logo-2x.png"></body></html>'
$ unzip -l test.epub
Archive:  test.epub
  Length      Date    Time    Name
---------  ---------- -----   ----
       20  2020-10-18 14:56   mimetype
      251  2020-10-18 14:56   META-INF/container.xml
      160  2020-10-18 14:56   META-INF/com.apple.ibooks.display-options.xml
     1326  2020-10-18 14:56   EPUB/content.opf
      767  2020-10-18 14:56   EPUB/toc.ncx
      541  2020-10-18 14:56   EPUB/nav.xhtml
      496  2020-10-18 14:56   EPUB/text/title_page.xhtml
     1119  2020-10-18 14:56   EPUB/styles/stylesheet1.css
      582  2020-10-18 14:56   EPUB/text/ch001.xhtml
---------                     -------
     5262                     9 files
$ unzip -l test.epub | grep png
$ unzip -p test.epub EPUB/text/ch001.xhtml | grep img
<img src="https://static.fimfiction.net/images/logo-2x.png" data-external="1" />
```

Fixes #6543

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -5864,6 +5864,12 @@ with the `src` attribute.  For example:
       </source>
     </audio>
 
+When using HTML as the input format (as opposed to the raw HTML
+extension), then the `data-external="1"` attribute also works the same
+as described above, but only for the `<img>` element.  The other
+elements have no representation in pandoc's AST and any attributes
+will hence be ignored.
+
 # Jupyter notebooks
 
 When creating a [Jupyter notebook], pandoc will try to infer the
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -790,7 +790,7 @@ pImage = do
   let getAtt k = case fromAttrib k tag of
                    "" -> []
                    v  -> [(k, v)]
-  let kvs = concatMap getAtt ["width", "height", "sizes", "srcset"]
+  let kvs = concatMap getAtt ["data-external", "width", "height", "sizes", "srcset"]
   return $ B.imageWith (uid, cls, kvs) (escapeURI url) title (B.text alt)
 
 pCodeWithClass :: PandocMonad m => [(T.Text,Text)] -> TagParser m Inlines
--- a/src/Text/Pandoc/Writers/EPUB.hs
+++ b/src/Text/Pandoc/Writers/EPUB.hs
@@ -1060,7 +1060,8 @@ transformInline  :: PandocMonad m
                  => WriterOptions
                  -> Inline
                  -> E m Inline
-transformInline _opts (Image attr lab (src,tit)) = do
+transformInline _opts (Image attr@(_,_,kvs) lab (src,tit))
+  | isNothing (lookup "data-external" kvs) = do
     newsrc <- modifyMediaRef $ TS.unpack src
     return $ Image attr lab ("../" <> newsrc, tit)
 transformInline opts x@(Math t m)
--- a/test/Tests/Readers/HTML.hs
+++ b/test/Tests/Readers/HTML.hs
@@ -84,6 +84,10 @@ tests = [ testGroup "base tag"
           [ test html "anchor without href" $ "<a name=\"anchor\"/>" =?>
             plain (spanWith ("anchor",[],[]) mempty)
           ]
+        , testGroup "img"
+          [ test html "data-external attribute" $ "<img data-external=\"1\" src=\"http://example.com/stickman.gif\">" =?>
+            plain (imageWith ("", [], [("data-external", "1")]) "http://example.com/stickman.gif" "" "")
+          ]
         , testGroup "lang"
           [ test html "lang on <html>" $ "<html lang=\"es\">hola" =?>
             setMeta "lang" (text "es") (doc (plain (text "hola")))


MANUAL.txt
```diff
@@ -5864,6 +5864,12 @@ with the `src` attribute.  For example:
       </source>
     </audio>
 
+When using HTML as the input format (as opposed to the raw HTML
+extension), then the `data-external="1"` attribute also works the same
+as described above, but only for the `<img>` element.  The other
+elements have no representation in pandoc's AST and any attributes
+will hence be ignored.
```
brisad, line 5871: I this accurate? This was what I could determine from reading the code, but I'm not 100% sure.

jgm:
Just so i understand, it looks as if this PR does two separate things:

1.  modifies HTML reader so that data-external attribute has an effect (namely, adding `data-external` attribute in the Image tag).
2.  modifies EPUB writer so that data-external has an effect when set on a Markdown image element, not just in raw HTML.

Perhaps better to split into two separate PRs?

Something to note:  markdown `img{external=1}` will go to an img tag in HTML with attribute `data-external=1`.

MANUAL.txt
mb21: All AST elements with an [`Attr` type](https://github.com/jgm/pandoc-types/blob/master/src/Text/Pandoc/Definition.hs) can host attributes... not sure the HTML Reader passes them along only for images or all elements.... (you can try with `pandoc -f html -t native`

brisad:
> Just so i understand, it looks as if this PR does two separate things:

Exactly.

> 1. modifies HTML reader so that data-external attribute has an effect (namely, adding `data-external` attribute in the Image tag).

Yes. Since we're not dealing with raw HTML here we don't have access to the raw tags, but instead AST elements. Passing this specific attribute through the AST representation I can then read it on the EPUB writer side later. Is this an OK approach? I don't see any other way to do this.

> 2. modifies EPUB writer so that data-external has an effect when set on a Markdown image element, not just in raw HTML.

Yes, but is `Image` below really a Markdown element? Is it not a general representation of an image, regardless of format? 
https://github.com/jgm/pandoc/blob/b8989e07f82246899e691b764d02d933eda930c0/src/Text/Pandoc/Writers/EPUB.hs#L1063-L1064
> 
> Perhaps better to split into two separate PRs?

Okay. I can close this one and do two instead.

> Something to note: markdown `img{external=1}` will go to an img tag in HTML with attribute `data-external=1`.

Oh, interesting!

Markdown `img{external=1}` will have `external` as attribute in the native representation, but that's changed to `data-external` when going to HTML, as you point out. Since my proposed solution relies on `data-external` in the pandoc AST, this case will _not_ trigger the logic that avoids downloading media to the EPUB container.

With markdown `img{data-external=1}` instead, that will put the `data-external` attribute both in the native representation and the HTML. The resulting HTML is still the same, but since the native representation now has `data-external`, my solution will consider that case.

@jgm Do you think `img{external=1}` should trigger the same effect? In that case I'd need to check for both `data-external` and `external` attached to an `Image` in the EPUB writer.

MANUAL.txt
brisad: As far as I can understand, the HTML reader only passes on attributes that it knows about, at least for the four tags mentioned in this section of the manual. Here's the output reading some dummy data into native format:
```
$ pandoc -f html -t native <<< '<img src="localhost" data-external="1" foo="bar"><audio attr="1">A</audio><video attr="2">B</video><source src="y" data-external="1" type="audio/mpeg">C</source>'
[Plain [Image ("",[],[("data-external","1")]) [] ("localhost","")]
,Plain [Str "A"]
,Plain [Str "B"]
,Plain [Str "C"]]
```
or, using a pandoc version without my `data-external` change:
```
$ pandoc -f html -t native <<< (same as above)
[Plain [Image ("",[],[]) [] ("localhost","")]
,Plain [Str "A"]
,Plain [Str "B"]
,Plain [Str "C"]]
```
This tells me `<audio>`, `<video>`, and `<source>` aren't representable as native AST elements, and that pandoc's HTML reader silently drops these tags. This is what I wanted to explain in the manual, in case a user wonders why only the ``-tag supports `data-external="1"` when not using raw HTML. But, I'm happy to just remove this sentence if it adds any confusion. :-)

jgm:
When writing HTML, we add the `data-` prefix if the attribute is not already a valid HTML5 attribute name.  A good idea.

We've implemented the same logic in (much of) the HTML reader. See `toAttr`, which does the reverse transformation.  Not as clear that's a good idea -- though it helps with round-tripping -- because the code base isn't entirely consistent.  E.g. I see that in the markdown writer we check for a `data-emoji` attribute.

I'm not exactly sure what the best approach is here.  For now, checking for both `external` and `data-external` on the AST element things safest, but I'd also be interested in feedback about how to make everything more consistent and less error-prone.

jgm:
Oh, and I'm not sure why we don't just use `toAttr` on Image, Link, etc.

brisad:
Eight months later... Sorry about that.

I've divided this into two separate PRs now: #7429 and #7430

Instead of storing the attribute `data-external` on `Image`, I instead strip the `data-` and use `external` as key, just like `toAttr`. I think this makes things more consistent, and also supports the markdown variant `img{external=1}`.

PR #6763, opened 2020-10-20
adds \setupinterlinespace to fonts larger than normal

denismaier:
Without this \tfd etc don't adjust line spacing.


denismaier:
@jgm what am i missing here?

jgm:
Look at the logs for the failing CIs and search for FAIL.  This will show you the failing tests, e.g. writers-lang-and-dir.context, which uses --standalone and is thus affected by this change in the default template.

Hint:  `make TESTARGS=--accept` will revise the expected test outputs to fit the current executable's behavior.  So you can do that and then carefully review the changes to the tests using `git diff` to make sure they look right.

PR #6774, opened 2020-10-25
OpenDocument writer: Allow references for internal links

pyssling:
This commit adds two extensions to the OpenDocument writer,
`references_over_links` and `number_prefix_references`.

The first extension, `references_over_links`, substitutes document
internal links for references to headers, figures and tables.
Text in references is kept consistent with the referenced header,
table or figure which is an improvement if the document is edited
after being generated by pandoc.

The second extension `number_prefix_references` will prefix the
header references with the number according to the style of the
referenced heading in the final document. As noted in the MANUAL.txt
the document will need to have indexes updated for these numbers
to be generated - similarly to table of contents in OpenDocument.
Figure and table references are not number prefixed as the numbers
for those are inline in the caption.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3027,6 +3027,43 @@ This extension can be enabled/disabled for the following formats:
 output formats
 :  `odt`, `opendocument`
 
+#### Extension: `xrefs_name` ####
+
+Links to headings, figures and tables inside the document are
+substituted with cross-references that will use the name or caption
+of the referenced item. The original link text is replaced once
+the generated document is refreshed. This extension can be combined
+with `xrefs_number` in which case numbers will appear before the
+name.
+
+Text in cross-references is only made consistent with the referenced
+item once the document has been refreshed.
+
+This extension can be enabled/disabled for the following formats:
+
+output formats
+:  `odt`, `opendocument`
+
+#### Extension: `xrefs_number` ####
+
+Links to headings, figures and tables inside the document are
+substituted with cross-references that will use the number
+of the referenced item. The original link text is discarded.
+This extension can be combined with `xrefs_name` in which case
+the name or caption numbers will appear after the number.
+
+For the `xrefs_number` to be useful heading numbers must be enabled
+in the generated document, also table and figure captions must be enabled
+using for example the `native_numbering` extension.
+
+Numbers in cross-references are only visible in the final document once
+it has been refreshed.
+
+This extension can be enabled/disabled for the following formats:
+
+output formats
+:  `odt`, `opendocument`
+
 #### Extension: `styles` #### {#ext-styles}
 
 When converting from docx, read all docx styles as divs (for
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -17,6 +17,7 @@ module Text.Pandoc.Writers.OpenDocument ( writeOpenDocument ) where
 import Control.Arrow ((***), (>>>))
 import Control.Monad.State.Strict hiding (when)
 import Data.Char (chr)
+import Data.Foldable (find)
 import Data.List (sortOn, sortBy, foldl')
 import qualified Data.Map as Map
 import Data.Maybe (fromMaybe, isNothing)
@@ -35,6 +36,7 @@ import Text.DocLayout
 import Text.Pandoc.Shared (linesToPara, tshow, blocksToInlines)
 import Text.Pandoc.Templates (renderTemplate)
 import qualified Text.Pandoc.Translations as Term (Term(Figure, Table))
+import Text.Pandoc.Walk
 import Text.Pandoc.Writers.Math
 import Text.Pandoc.Writers.Shared
 import qualified Text.Pandoc.Writers.AnnotatedTable as Ann
@@ -54,6 +56,11 @@ plainToPara x         = x
 
 type OD m = StateT WriterState m
 
+data ReferenceType
+  = HeaderRef
+  | TableRef
+  | ImageRef
+
 data WriterState =
     WriterState { stNotes          :: [Doc Text]
                 , stTableStyles    :: [Doc Text]
@@ -69,6 +76,7 @@ data WriterState =
                 , stImageId        :: Int
                 , stTableCaptionId :: Int
                 , stImageCaptionId :: Int
+                , stIdentTypes     :: [(Text,ReferenceType)]
                 }
 
 defaultWriterState :: WriterState
@@ -86,6 +94,7 @@ defaultWriterState =
                 , stImageId        = 1
                 , stTableCaptionId = 1
                 , stImageCaptionId = 1
+                , stIdentTypes     = []
                 }
 
 when :: Bool -> Doc Text -> Doc Text
@@ -243,6 +252,12 @@ writeOpenDocument opts (Pandoc meta blocks) = do
                         meta
   ((body, metadata),s) <- flip runStateT
         defaultWriterState $ do
+           let collectInlineIdent (Image (ident,_,_) _ _) = [(ident,ImageRef)]
+               collectInlineIdent _                       = []
+           let collectBlockIdent (Header _ (ident,_,_) _)      = [(ident,HeaderRef)]
+               collectBlockIdent (Table (ident,_,_) _ _ _ _ _) = [(ident,TableRef)]
+               collectBlockIdent _                             = []
+           modify $ \s -> s{ stIdentTypes = query collectBlockIdent blocks ++ query collectInlineIdent blocks }
            m <- metaToContext opts
                   (blocksToOpenDocument opts)
                   (fmap chomp . inlinesToOpenDocument opts)
@@ -411,7 +426,7 @@ blockToOpenDocument o bs
                            inTags True "text:list" [ ("text:style-name", "L" <> tshow ln)]
                                       <$> orderedListToOpenDocument o pn b
       table :: PandocMonad m => Ann.Table -> OD m (Doc Text)
-      table (Ann.Table _ (Caption _ c) colspecs thead tbodies _) = do
+      table (Ann.Table (ident, _, _) (Caption _ c) colspecs thead tbodies _) = do
         tn <- length <$> gets stTableStyles
         pn <- length <$> gets stParaStyles
         let  genIds      = map chr [65..]
@@ -433,7 +448,7 @@ blockToOpenDocument o bs
                       then return empty
                       else inlinesToOpenDocument o (blocksToInlines c) >>=
                              if isEnabled Ext_native_numbering o
-                                then numberedTableCaption
+                                then numberedTableCaption ident
                                 else unNumberedCaption "TableCaption"
         th <- colHeadsToOpenDocument o (map fst paraHStyles) thead
         tr <- mapM (tableBodyToOpenDocument o (map fst paraStyles)) tbodies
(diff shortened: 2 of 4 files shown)


tarleb:
Thank you for opening this PR, that seems to be a useful pair of features.

In order to make this a good experience for everyone, I'd like to learn a bit about your motivation and expectation for this PR: Which statements fits best?
1. "This PR doubles as a welcome excuse to learn more about Haskell. A detailed code review would be welcome."
2. "The feature is the only thing that matters. Honestly don't care too much about the details, it would be great to get this merged ASAP."

Both options work for us and will cost us similar amounts of time, so feel free to indicate your choice.

pyssling:
Hi Albert,

I am in situation 1. with regards to the code. This code is already in use - but I am very open to improving it. My total Haskell experience (and knowledge) constitutes one course at university 19 years ago - and now a number of patches to pandoc. I'm sure the code is very crude from an experienced Haskell programmers perspective.

Thank you for your time.

tarleb:
Great, thanks for the feedback! I'll do an in-depth review within the next few days.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -431,33 +447,37 @@ blockToOpenDocument o bs
       figure attr caption source title | null caption =
         withParagraphStyle o "Figure" [Para [Image attr caption (source,title)]]
                                   | otherwise    = do
+        let (ref, _, _) = attr
```
tarleb, line 450: A common idiom is to destructure Attr in the definition:

``` haskell
figure  attr@(ident, _, _) caption ...
```

This names both the whole Attr triple as well as the identifier.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -398,7 +414,7 @@ blockToOpenDocument o bs
                            inTags True "text:list" [ ("text:style-name", "L" <> tshow ln)]
                                       <$> orderedListToOpenDocument o pn b
       table :: PandocMonad m => Ann.Table -> OD m (Doc Text)
-      table (Ann.Table _ (Caption _ c) colspecs thead tbodies _) = do
+      table (Ann.Table (ref, _, _) (Caption _ c) colspecs thead tbodies _) = do
```
tarleb, line 417: `ref` is usually called `ident` in the code base.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -635,6 +655,48 @@ inlineToOpenDocument o ils
...
+                              -- The number prefix on header references is empty
+                              -- as generated by pandoc - indices need to be
+                              -- refreshed for the number to be generated.
+                              then inTags False "text:bookmark-ref"
+                                          [ ("text:reference-format", "number" ),
+                                            ("text:ref-name", ident) ] mempty
+                                      <>
+                                      selfClosingTag "text:s" []
+                                      <>
+                                      inTags False "text:bookmark-ref"
```
tarleb, line 675: There is a lot of code duplication in here. This would probably also benefit from a new data representation mentioned above, as it should make it easier to fold the three`*Reference` functions into one.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -235,6 +242,15 @@ writeOpenDocument opts (Pandoc meta blocks) = do
                         meta
   ((body, metadata),s) <- flip runStateT
         defaultWriterState $ do
+           let isHeaderIdent (Header _ (ident,_,_) _) = [ident]
+               isHeaderIdent _ = []
+           modify $ \s -> s{ stHeaderRefs = query isHeaderIdent blocks }
```
tarleb, line 247: We iterate over all document elements three times. This could be reduced to a single pass by combining the three identifier lists into a single data type as mentioned above.

tarleb (commented):
There is a joke in the Haskell community that there a no advanced Haskellers, as everybody is intermediate forever. Hope my comments are helpful regardless.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -235,6 +242,15 @@ writeOpenDocument opts (Pandoc meta blocks) = do
                         meta
   ((body, metadata),s) <- flip runStateT
         defaultWriterState $ do
+           let isHeaderIdent (Header _ (ident,_,_) _) = [ident]
```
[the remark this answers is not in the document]
pyssling: done.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -69,6 +70,9 @@ data WriterState =
                 , stImageId        :: Int
                 , stTableCaptionId :: Int
                 , stImageCaptionId :: Int
+                , stHeaderRefs     :: [Text]
```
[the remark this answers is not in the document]
pyssling: Changed to `[(Text,Text)]` - let me know if there is some better way to capture types than by textual name.

src/Text/Pandoc/Writers/OpenDocument.hs
pyssling: done.

src/Text/Pandoc/Writers/OpenDocument.hs
pyssling: Fixed everywhere - I hope.

src/Text/Pandoc/Writers/OpenDocument.hs
pyssling: Refactored this area - no code duplication now I think.

src/Text/Pandoc/Writers/OpenDocument.hs
pyssling: Combined in what I hope is a correct way - I'm slightly unsure of the method of collecting inlines - but it seems to work based on the unit test.

src/Text/Pandoc/Writers/OpenDocument.hs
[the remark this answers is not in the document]
tarleb: Types are cheap and easy, so we could add a dedicated type for this:

```haskell
data ReferenceType
  = HeaderRef
  | TableRef
  | ImageRef
```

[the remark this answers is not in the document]
pyssling: Thank you! Much prettier.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -635,6 +650,46 @@ inlineToOpenDocument o ils
...
         return nn
 
+mkLink :: WriterOptions -> [(Text,ReferenceType)] -> Text -> Text -> Doc Text -> Doc Text
+mkLink o identTypes s t d =
+  let findIdentType ident = mapMaybe identToType identTypes where
+        identToType (it,tp)
+          | it == ident = Just (it,tp)
+          | otherwise   = Nothing
+      maybeIdentAndType = case T.uncons s of
+                            Just ('#', ident) -> listToMaybe (findIdentType ident)
```
tarleb, line 660: The Prelude comes with a [`lookup`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:lookup) to handle associated lists, but only returns the second element of the tuple. If we want both we'd have to reconstruct the tuple.

```haskell
-- would be shorter with language extension TupleSections enabled, but let's not go there.
(\x -> (ident, x)) <$> lookup ident identTypes 
```

Possibly simpler would be to cast some pointfree [Data.Foldable.find](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Foldable.html#v:find) magic:

```haskell
find ((ident ==) . fst) identTypes
```

The nice thing about the latter solution is that it works as long as `identTypes` is an instance of class `Foldable`. We wouldn't even have to touch such a line if we were to change `identTypes` to type `Map Text ReferenceType` later on.

tarleb (commented):
LGTM!

src/Text/Pandoc/Writers/OpenDocument.hs
tarleb: If you wonder how anyone can remember all those functions: we don't. The trick is to think about the types you need and to [put them into Hoogle](https://hoogle.haskell.org/?hoogle=(a%20-%3E%20Bool)%20-%3E%20%5Ba%5D%20-%3E%20Maybe%20a).

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -235,6 +243,12 @@ writeOpenDocument opts (Pandoc meta blocks) = do
                         meta
   ((body, metadata),s) <- flip runStateT
         defaultWriterState $ do
+           let collectInlineIdent (Image (ident,_,_) _ _) = [(ident,ImageRef)]
+               collectInlineIdent _                       = []
+           let collectBlockIdent (Header _ (ident,_,_) _)      = [(ident,HeaderRef)]
+               collectBlockIdent (Table (ident,_,_) _ _ _ _ _) = [(ident,TableRef)]
+               collectBlockIdent b                             = query collectInlineIdent b
+           modify $ \s -> s{ stIdentTypes = query collectBlockIdent blocks }
```
tarleb, line 251: Would this find an Image nested in a Table?
I forgot that Image is an inline and cannot be included into the `collectBlockIdent` function. I guess I was wrong, and we need at least two iterations.

src/Text/Pandoc/Writers/OpenDocument.hs
pyssling: done. Much better - thank you.

src/Text/Pandoc/Writers/OpenDocument.hs
pyssling: I've adjusted this now to make two iterations.

pyssling:
@jgm any chance you could give this one a review as well?

jgm:
I'm not sure I understand what these extensions do.  Can you explain a bit more, maybe with an example?

pyssling:
Well, the first option changes from links to references - in OpenDocument a reference is something that updates with the target text, so a heading of "Introduction" will have a reference called "Introduction", and if a later editor of the generated document changes "Introduction" to "Background" then the reference will also change to "Background" which the link will not.

The second option number prefixes references to headings. So instead of in the generated text having "(see Introduction)" we would have "(see 1. Introduction)" This is especially useful for printed documents where there are no links - chapter numbers become really useful.

I've tried to describe this quite thoroughly in the manual changes.

pyssling:
@jgm I'm wondering if you've had time to consider this and whether you would like me to look at alternative approaches.

The fact is I'm not the happiest with this approach. In the original format in my case the "References" identified here are of the type "xref" which do not have a "link text". Because the Link type in pandoc expects a link text the Docbook reader constructs one, and essentially what I'm doing the OpenDocument writer is removing the Link text from identified references and turning them back into references.

This could probably be done better by creating an Xref or InternalReference or similar inline type in the pandoc-types repo and extending the AST in that way. This is a bit more work - but I suspect there are many other formats that would make better use of xrefs rather than links.

Let me know how you wish to proceed.

jgm:
So with no extension a link looks like
```xml
<text:a xlink:type="simple" xlink:href="#chapter1" office:name=""><text:span text:style-name="Definition">The
Chapter</text:span></text:a>
```
With `references_over_links` it looks like
```xml
<text:bookmark-ref text:reference-format="text" text:ref-name="chapter1">The
Chapter</text:bookmark-ref>
```
And with `references_over_links+number_prefix_references`
```xml
<text:bookmark-ref text:reference-format="number" text:ref-name="chapter1"></text:bookmark-ref><text:s /><text:bookmark-ref text:reference-format="text" text:ref-name="chapter1">The
Chapter</text:bookmark-ref>
```
Correct?

pyssling:
Yes - the interesting thing to keep in mind is that the contents of:

```
<text:a xlink:type="simple" xlink:href="#chapter1" office:name=""></text:a>
```
is dynamic - i.e. if it header for `#chapter1` changes - the text in the reference is updated. People expect this behavior - but with Links - this isn't possible.

jgm:
So, do I understand correctly that the text provided, "The Chapter" , is just ignored in the second two cases?

pyssling:
It's a placeholder until the document is refreshed, visible but temporary. We could also skip it - and just have this empty until the document is refreshed.

jgm:
So, this means that if someone specifies a link in Markdown like
```
[My stuff](#mystuff)
```
then these extensions will ignore the description they write and substitute the actual name of the linked resource?

pyssling:
Yes - with a markdown document like this:

```
# Chapter My Stuffing {#mystuff}

[My stuff](#mystuff)
```

We end up with this output after running pandoc with `-t odt+references_over_links`:
image

In libreoffice we can now go the `Tools` menu, select `Update` and then `Update All`.

We then get this:

image

If we now update the chapter to title - without doing any updates or anything we get:

image

This is the behavior we want from our documents - that cross-references in the final document update when the document is edited. Hope this makes sense.

jgm:
I think this could be made clearer in the documentation.  People who understand the odt/opendocument format well will understand what you wrote, but many users won't.  Saying explicitly that the link text in links to internal targets will be replaced by the text from the linked section heading (or image or?) would be helpful.

pyssling:
Ok, done. Hope this is clear now.

MANUAL.txt
```diff
@@ -3027,6 +3027,36 @@ This extension can be enabled/disabled for the following formats:
...
+This typically requires the final document to have indexes refreshed
+in a native editor such as libreoffice as pandoc is not aware of the
+number style of the document.
+
+This further assumes that references are used within the document
+instead of links - see the `references_over_links` extension.
+
+output formats
+:  `odt`, `opendocument`
+
```
jgm, line 3059: I'm concerned that a reader might think "references" refers to markdown link references or something like that.  Lots of formats have things called "references."

Would using the term "crossreferences" or something like "xrefs" be clearer?  Maybe the extension should be `crossrefs` and `numbered_crossrefs`?  (The latter could automatically enable use of cross-references, so people wouldn't have to specify both.)

Do you think these would apply to other formats besides opendocument/docx?

It may be that these names are the best, but I think we should hash out alternatives first, because this is an API change and hard to reverse.

pyssling: What about `xrefs_for_internal_links` and `numbered_xrefs` ? I think that should be very clear - but my perception is slanted.

This should be applicable to OpenDocument, Docx most probably, and DocBook output formats. At least.

jgm: Yes, I like that better.

jgm: One thing that occurs to me, though, is that with `xrefs_for_internal_links` one might expect that *all* internal links would become xrefs.  `xrefs` by itself doesn't imply this and may for that reason be better (also shorter).  I think you'd probably need to look in the manual to figure out what it does, but that is probably true regardless of the name.

pyssling: Ok - `xrefs` and `numbered_xrefs` it is. And `numbered_xrefs` enables `xrefs`. Thanks for your patience.

jgm:
Just a thought:  why not have the `xrefs_number` extension *just* add the number, and `xrefs` *just* add the name?  Then the user could choose to have both together, or just one.  The reason I mention this is that I can think of a lot of contexts where I'd like to refer back to "Section 3.3" without having the whole name of the section in the link. In fact, this is the normal case in most academic writing: this is how LaTeX's `\ref` works, for examle.  With this revision, I'd suggest changing the extensions to something like `xrefs_name` and `xrefs_number` to make it a bit clearer what they do.

pyssling:
I guess it becomes very ODT specific then - and also a bit inconsistent.

The `xrefs` extension will always include name and number for figures and tables that have one. The `numbered_xrefs` extension only prefixes a number to the heading if heading numbers are enabled in the generated document.

This isn't quite the pick-and-choose scenario I'm afraid.
Another option would be to have 3 extensions:

`xrefs` - enables xrefs instead of links where supported - ODT, docx, DocBook
`numbered_xrefs` - ODT where there are heading numbers (and possibly Docx)
`number_only_xrefs` - ODT where there are heading, table and figure numbers (and possibly Docx)

This is quite hairy I'm afraid - not at all a clear cut case - ODT was not designed with this use-case in mind.

jgm:
Hm.  What does the xrefs extension now return for the "name and number" for figures and tables?  Is it "Figure 1" or is it "1 The caption of the figure" or what?

If the extension returns "Figure 1" for figures and the text of the section heading (e.g. "Introduction") for headings, that seems quite inconsistent.  I would have expected it to work in the same way in both cases.

pyssling:
You were right to push on this - I've now done quite a bit of work to make this make more sense and managed to find the right XML syntax to accomplish the `xrefs_name` and `xrefs_number` extensions.

Now `xrefs_name` will use only the caption text for figures and tables, ignoring the `Figure X: ` part of the caption. `xrefs_number` only uses the number. Enabling both will place the number before the name with a space between.

Quite a long journey - but it looks much better.

pyssling:
@jgm Is this good to go? Let me know if there's anything more we need.

jgm:
Looks good!

pyssling:
Thanks!

PR #6786, opened 2020-10-30
Updates to default CSS

mb21:
follow-up to https://github.com/jgm/pandoc/pull/6601

motivated by https://groups.google.com/g/pandoc-discuss/c/1oCCbois1G4

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2190,7 +2190,7 @@ ODT or pptx.
 ### Variables for HTML
 
 `document-css`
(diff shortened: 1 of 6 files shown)


jgm:
If you don't already know about this:  you can do `make TESTARGS=--accept` to update the tests using the current output.  Then it's important to inspect the tests very carefully, but this is far easier than updating them manually.

mb21:
Thanks, i always forget what command it is... waiting to recompile now ;)

PR #6787, opened 2020-10-30
Fix duplicate "class" attribute in HTML writer

andy-morris:
The html writer adds a class "odd" or "even" to each row in tables. If a row has any user-specified classes (e.g. from a filter), then this results in the tag in the output having two `class` attributes, the second one being ignored by browsers.

So this PR changes the table row code slightly so that the `even`/`odd` classes are just added to the existing `Attr` instead.


tarleb:
Thanks!

PR #6790, opened 2020-10-31
Lint code in PRs and when committing to master

tarleb:
This adds a new GitHub Actions workflow which checks the code with HLint for all PRs and commits to master. Most linter annotations in the code are resolved; the remaining warnings are disabled via configs.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,30 @@
+name: Lint
+
+# Trigger the workflow on push or pull request, but only for the
+# master branch
+on:
+  pull_request:
+    branch: [master]
+  push:
+    branch: [master]
+    paths-ignore:
+      - LICENSE
+      - README.md
+      - CHANGELOG.md
+      - stack.yaml
+      - .travis.yml
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    env:
+      hlint_script: https://raw.github.com/ndmitchell/hlint/master/misc/run.sh
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Download and run hlint
+        run: |
+          curl -sSL "${hlint_script}" | sh -s .
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -2,85 +2,107 @@
 # https://github.com/ndmitchell/hlint
 ##########################
 
-# This file contains a template configuration file, which is typically
-# placed as .hlint.yaml in the root of your project
-
-
 # Specify additional command line arguments
 #
 - arguments: [--color=auto, --cpp-ansi]
 
-
-# Control which extensions/flags/modules/functions can be used
-#
-# - extensions:
-#   - default: false # all extension are banned by default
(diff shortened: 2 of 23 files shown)


tarleb:
This is a bit of an experiment. I have generally had good experiences using linters in other projects, but there is always the danger that it gets in the way more than it helps.

jgm:
Yeah, I'm not sure.  I normally don't *always* apply hlint's suggestions, so I'm not sure I'd want to make it a requirement for commits.

Sometimes I find that some parentheses that aren't strictly necessary are helpful to make the code more readable by humans, for example.

tarleb:
That's very true. The PR disables multiple hints because they would have made things worse (IMHO).

I am using automatic linting checks in hslua; so far the experience has been good. However, my feeling is that unnecessary warnings happen more often than in other languages. Maybe this *would* get annoying quickly when used in a project of pandoc's size.

I haven't tried it yet, but maybe the [hint-man](https://github.com/apps/hint-man) bot could be a good alternative.

mb21:
Not sure if this has already been discussed, but we could also go full blown with https://github.com/tweag/ormolu

tarleb:
Ormolu and similar tools are great in a company environment IMHO. I wouldn't want to use them in pandoc though.

jgm:
Let's try it.  If the automatic lint action gets to be a pain, we can disable it.  But even if we do, these changes to hlint config and to the code are useful.

jgm:
I'm confused about how this works!  (Background: I was trying to add a matching lint step to my pre-push procedure.)
Can you explain it?
The main action is
```
curl -sSL "${hlint_script}" | sh -s .
```
where `hlint_script` = `https://raw.github.com/ndmitchell/hlint/master/misc/run.sh`.
Now if I got to that URL, I see
```
#!/bin/sh
curl -sSL https://raw.github.com/ndmitchell/neil/master/misc/run.sh | sh -s -- hlint $*
```
So, we download a script and run it, and the script says to download itself and pipe itself to hlint?  What is going on here?
I tried doing `hlint .` in the pandoc source directory, but I got many more warnings than the CI gives.

tarleb:
I took that line from the hlint docs' [https://github.com/ndmitchell/hlint#running-with-continuous-integration](CI section). Following all URL redirects and indirections, I end up at https://raw.githubusercontent.com/ndmitchell/neil/master/misc/run.sh, which downloads hlint from its GitHub release page. This *does* seem overly convoluted.

> I tried doing `hlint .` in the pandoc source directory, but I got many more warnings than the CI gives.

That's weird, I'm not getting any warnings right now. I don't know what's going on there.

tarleb:
FWIW, I'm running hlint 3.2.

tarleb:
My pre-commit hook is

``` shell
#!/bin/sh
git diff --cached --name-only | grep '.hs$' | xargs hlint --hint .hlint.yaml
```

jgm:
I've got an older hlint, that may be the issue.

jgm:
Upgraded hlint.  That did it. Thanks for the pre-commit hook.

PR #6791, opened 2020-10-31
DocBook reader: Table width support

pyssling:
Table width is not natively supported by docbook but is by
the docbook fo stylesheets through an XML processing instruction,
<?dbfo table-width="50%"?> . Implement support for this instruction
in the DocBook reader.

### DIFF
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -960,12 +960,22 @@ parseBlock (Elem e) =
                       let aligns = case colspecs of
                                      [] -> replicate numrows AlignDefault
                                      cs -> map toAlignment cs
+                      let parseWidth s = safeRead (T.filter (\x -> (x >= '0' && x <= '9')
+                                                                   || x == '.') s)
+                      let textWidth = case filterChild (named "?dbfo") e of
+                                        Just d  -> case attrValue "table-width" d of
+                                                     "" -> 1.0
+                                                     w  -> fromMaybe 100.0 (parseWidth w) / 100.0
+                                        Nothing -> 1.0
                       let widths = case colspecs of
                                      [] -> replicate numrows ColWidthDefault
                                      cs -> let ws = map toWidth cs
                                            in case sequence ws of
-                                                Just ws' -> let tot = sum ws'
-                                                            in  ColWidth . (/ tot) <$> ws'
+                                                Just ws' -> let colTot = sum ws'
+                                                                scale
+                                                                  | textWidth == 1.0 = (/ colTot)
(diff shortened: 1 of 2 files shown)


tarleb (commented):
The code looks good, just the indentation of the first case statement appears to be off.
I don't know enough about DocBook to comment on the logic.

pyssling:
Thank you, done.

tarleb (approved):

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -960,12 +960,22 @@ parseBlock (Elem e) =
                       let aligns = case colspecs of
                                      [] -> replicate numrows AlignDefault
                                      cs -> map toAlignment cs
+                      let parseWidth s = safeRead (T.filter (\x -> (x >= '0' && x <= '9')
+                                                                   || x == '.') s)
+                      let textWidth = case filterChild (named "?dbfo") e of
```
jgm, line 965: What is this `?dbfo`?

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -960,12 +960,22 @@ parseBlock (Elem e) =
                       let aligns = case colspecs of
                                      [] -> replicate numrows AlignDefault
                                      cs -> map toAlignment cs
+                      let parseWidth s = safeRead (T.filter (\x -> (x >= '0' && x <= '9')
```
jgm, line 963: Is this field just a decimal number with no units?  If so, do we need to filter?  If not, are we ignoring the units?

src/Text/Pandoc/Readers/DocBook.hs
pyssling: It's an XML processing instruction that's used by the docbook stylesheets to specify the table width in proportion to text. The CALS table model (from 1995) never included something width in relation to text - so this is what people came up with it seems.

src/Text/Pandoc/Readers/DocBook.hs
pyssling: It's percent - but the percent sign may or may not be there from what I've seen. I though this was safest - but there may be better ways.

jgm:
It would be nice to have a test of some kind.
At least a command test.

pyssling:
Added a testcase. Hopefully good now.

jgm:
Thanks -- I should have mentioned, we typically name the test cases after issues or PRs, so that it's easier to remember/look up what motivated them.  Thus 6791.md.
Alternatively you could add a link to this PR in the text.

pyssling:
done.

jgm:
thanks!

PR #6792, opened 2020-10-31
OpenDocument writer: Table width support

pyssling:
Use the same key-value in Attr for the table as the HTML writer
to implement table width support.

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -407,13 +407,14 @@ blockToOpenDocument o bs
              fromWidth (ColWidth w) | w > 0 = w
              fromWidth _                    = 0
              widths = map fromWidth mwidths
+             textWidth   = sum widths
              columnIds   = zip genIds widths
              mkColumn  n = selfClosingTag "table:table-column" [("table:style-name", name <> "." <> T.singleton (fst n))]
              columns     = map mkColumn columnIds
              paraHStyles = paraTableStyles "Heading"  pn aligns
              paraStyles  = paraTableStyles "Contents" (pn + length (newPara paraHStyles)) aligns
              newPara     = map snd . filter (not . isEmpty . snd)
-        addTableStyle $ tableStyle tn columnIds
+        addTableStyle $ tableStyle tn textWidth columnIds
         mapM_ addParaStyle . newPara $ paraHStyles ++ paraStyles
         captionDoc <- if null c
                       then return empty
@@ -684,14 +685,19 @@ listLevelStyle i =
                       , ("fo:margin-left", indent <> "in")
                       ]
 
-tableStyle :: Int -> [(Char,Double)] -> Doc Text
-tableStyle num wcs =
+tableStyle :: Int -> Double -> [(Char,Double)] -> Doc Text
+tableStyle num textWidth wcs =
     let tableId        = "Table" <> tshow (num + 1)
+        tableWidthAttr :: [(Text,Text)]
+        tableWidthAttr
+          | textWidth <= 1 && textWidth > 0 = [("style:rel-width",
+                                                T.pack (show (round (textWidth * 100) :: Int) <> "%"))]
+          | otherwise    = []
         table          = inTags True "style:style"
                          [("style:name", tableId)
                          ,("style:family", "table")] $
                          selfClosingTag "style:table-properties"
-                         [("table:align"    , "center")]
+                         (("table:align", "center") : tableWidthAttr)
         colStyle (c,0) = selfClosingTag "style:style"
                          [ ("style:name"  , tableId <> "." <> T.singleton c)
                          , ("style:family", "table-column"       )]
--- a/test/command/6792.md
+++ b/test/command/6792.md
@@ -1,0 +1,66 @@
+```
+% pandoc -f native -t opendocument -s --quiet
+[Table ("",[],[]) (Caption Nothing
+ [])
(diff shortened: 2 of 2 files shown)


src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -684,14 +684,18 @@ listLevelStyle i =
                       , ("fo:margin-left", indent <> "in")
                       ]
 
-tableStyle :: Int -> [(Char,Double)] -> Doc Text
-tableStyle num wcs =
+tableStyle :: Int -> [(Text,Text)] -> [(Char,Double)] -> Doc Text
+tableStyle num kvs wcs =
     let tableId        = "Table" <> tshow (num + 1)
+        tableAttrStyles []   = []
```
tarleb, line 690: I think this could be simplified by using [`find`](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-OldList.html#v:find) and [`maybeToList`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Maybe.html#v:maybeToList). Or [`mapMaybe`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Maybe.html#v:mapMaybe).

Of course, "simplified" here is subjective. I subscribe to the idea that explicit recursion is better avoided if possible, but YMMV.

src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -684,14 +684,18 @@ listLevelStyle i =
...
+        tableAttrStyles ((k,v):xs)
+          | k == "width" = [("style:rel-width",v)] ++ tableAttrStyles xs
+          | otherwise = tableAttrStyles xs
         table          = inTags True "style:style"
                          [("style:name", tableId)
                          ,("style:family", "table")] $
                          selfClosingTag "style:table-properties"
-                         [("table:align"    , "center")]
+                         ([("table:align", "center")] ++ (tableAttrStyles kvs))
```
tarleb, line 698: The first argument to the list concatenation is a singleton list, so this could be written using the `:` cons operator instead of list concatenation.

tarleb (commented):
Thanks! I left some minor code comments. LGTM otherwise!

src/Text/Pandoc/Writers/OpenDocument.hs
pyssling: I changed it to a mapMaybe - thank you for the assistance.

src/Text/Pandoc/Writers/OpenDocument.hs
pyssling: and done.

tarleb (approved):

tarleb:
Just to keep you in the loop: a new release is in preparation right now. That means merging may be delayed until after the release is out.

jgm:
Can you explain more fully what you mean by "table width support"?

pyssling:
Yes sorry - this is the width of the table as percent of the page-width. It's correctly transferred into the HTML writer.

jgm:
This should be determined by the relative widths of the columns, as specified in the table AST, right?

pyssling:
Column widths come in as a percentage of the table width as far as I can tell. I.e. you still need the full width of the table.

The AST types define the following:

-- | The width of a table column, as a fraction of the total table
-- width.

I.e. they seem to be relative to the total table width - not absolute.

jgm:
No, actually, the proper interpretation of the relative column widths in the table AST is as percentages of text width (not table width).

pyssling:
Ok, I can change it to do that and send a pull-request to fix the AST documentation then.

pyssling:
@jgm here is a pull-request to correct the AST documentation: https://github.com/jgm/pandoc-types/pull/85

If you could review and it's correct then I can refactor this pull-request and the one for the DocBook reader.

pyssling:
@jgm I've refactored this now to make use of the column widths as a percentage. Much more satisfying than looking at an undocumented key-value I have to say.

jgm:
I'm just wondering how this interacts with the relative column widths.  Do we support column widths yet in opendocument?  If we do, how do these interact with the table width?

pyssling:
Yes - column widths are supported and work from what I've seen - they are expressed in proportion to each other - i.e. they are without unit.
It works perfectly from what I can see - it really just scales the table in relation to text-width - doesn't affect the columns at all - they are still proportional as before.

jgm:
How about a test or two?  Easiest way is to add a command test, which could be named `test/command/6792.md`.

pyssling:
I'll see what I can do - this requires a full test - command tests do not work for styles as they can only manage opendocument output - not odt output.

pyssling:
I'm afraid this is beyond me - there is no framework for testing the ODT writer - just the ODT reader. I'm afraid I wouldn't know how (or have the time) to develop that sort of framework. I'm afraid these styles will remain untested for the time being.

jgm:
Since this modifies the opendocument writer, couldn't we just have a command test with `pandoc -t opendocument`?

pyssling:
Ok, thank you for pushing me on this. I figured out what was going on - the opendocument writer doesn't emit styles unless a template is used - so I got that working. Now there's a solid unit test. Thank you.

jgm:
Great. I don't think we need the special template. Just use the default one:
`pandoc -f native -t opendocument -s`.

pyssling:
done, sorry about that - didn't realize the template directory was accessible.

jgm:
great, thanks!

PR #6837, opened 2020-11-14
[API change] Atx headings new default.

argent0:
* Adds the --markdown-headings=atx|setext option.
* Deprecates --atx-headers

This is a remake of #6693

Adresses #6662


jgm:
Closed by f63b76e1698b0d7eba6b43ef45faaeee2b01b9ca

I made a few changes and committed this.  Thanks very much for your work on the issue.

The changes:

- changed the name of the LogMessage constructor, and added some further parameters so it is more informative
- fixed the logic for the warning in the Markdown writer
- changed test output rather than adding the `--markdown-headings=setext` option to them
- support `markdown-headings` in defaults files
- document the new option in MANUAL

argent0:
Thanks for your time.

I wrote a more informative constructor for the log message in the original PR but simplified it at your request. So I'm glad you changed your mind and decided to add more information.

I don't use the `lhs` format myself but, when writing markdown with several latex formulas, errors can be difficult to trace back to the source. So I appreciate more informative error messsges.

jgm:
Yes, your instinct was right about that!

PR #6838, opened 2020-11-15
JATS writer: move Table handling to separate module

tarleb:
This makes it easier to split the module into smaller parts.


tarleb:
I'm starting to get the new table features supported in JATS. I'm trying to avoid another huge module that causes problems when compiled on systems with limited memory. This is an attempt to split it up.

I'd like to learn whether this could serve as a model for other writers as well, esp. the large HTML writer.

jgm:
I'm fine with this, if you want to merge it.

PR #6840, opened 2020-11-16
CONTRIBUTING: describe hlint and how it's used

tarleb:


CONTRIBUTING.md
```diff
@@ -206,6 +206,32 @@ or Powerpoint to ensure that they weren't corrupted and that
...
 
+Code style
+----------
+
+Pandoc uses [hlint] to identify opportunities for code improvements,
+like redundant brackets, or pointing out unnecessary `Language`
+extensions. However, there sometimes are cases where there are good
+reasons to use code different from what hlint proposes. In these
+cases, the respective warning should be disabled in file
+`.hlint.yaml`.
```
alerque, line 217:
```suggestion
Pandoc uses [hlint] to identify opportunities for code improvements
like redundant brackets or unnecessary `Language`
extensions. However, sometimes there are cases where there are good
reasons to use code different from what hlint proposes. In these
cases, the respective warning should be disabled in the file
`.hlint.yaml`.
```

CONTRIBUTING.md
```diff
@@ -206,6 +206,32 @@ or Powerpoint to ensure that they weren't corrupted and that
...
+
+There should be no errors when running `hlint .`; this is checked by
+the continuous integration (CI) setup. It is recommended that
+contributors check their code with a local hlint installation, but
+relying on the CI is fine, too.
+
+A good way to ensure no new hints are triggered is to use a git
+[pre-commit hook], which runs hlint on all updated Haskell files
```
alerque, line 225:
```suggestion
A good way to ensure no new warnings are introduced is to use a Git
[pre-commit hook] which runs hlint on all updated Haskell files
```

alerque (changes requested):
Relatively minor English copy-editing.

CONTRIBUTING.md
tarleb: Thanks!

CONTRIBUTING.md
tarleb: Thanks, updated!

alerque (approved):

PR #6843, opened 2020-11-16
Added a sentence about `pagetitle` for HTML in the man page

xealits:
Let's inform users about the `pagetitle` variable, which is handy for HTML.

I am not sure which branch this patch should go to, `master` is set by default.  And I assumed `man/pandoc.1` is generated from this text file. Correct me if it's wrong.

	changed:      MANUAL.txt


jgm:
good idea.

PR #6851, opened 2020-11-17
JATS writer: support advanced table features

tarleb:


tarleb:
Todo:

- [ ] Tests
- [ ] Validation checks

tarleb:
I also checked that the resulting tables display nicely in the eLife [lens](https://lens.elifesciences.org) reader.

test/tables.jats_archiving
```diff
@@ -4,35 +4,31 @@
     <p>Demonstration of simple table syntax.</p>
   </caption>
   <table>
-    <col align="right" />
-    <col align="left" />
-    <col align="center" />
-    <col align="left" />
```
tarleb, line 10: I thought about keeping this, as it *should* work; however, the Lens reader doesn't respect the `align` attribute in `<col>` elements, so the alignment is always done for each cell individually (same as in the HTML writer).

tarleb:
This should be a fairly uncontroversial change, therefore taking the freedom to merge myself.

PR #6852, opened 2020-11-17
Preserve center environment

ip1981:


tarleb (commented):
You may want to run the test suite with `--accept`, e.g. via

    make test TESTARGS=--accept

I don't have a fully formed opinion on whether this is an desirable change or not. Conversions will be always lossy, and centering is a presentational detail that's not the focus.  OTOH, this would certainly be helpful when converting to HTML.

jgm:
Great, thanks!

PR #6853, opened 2020-11-17
Update org supported languages and keyword case

tecosaur:
See commit messages for details, these commits enact the following changes:
- update the list of languages supported by org
- change keywords to be lower case


tarleb:
Great, thanks!

Issue #6687, opened 2020-09-14
jats_publishing: author affiliation is missing?


denismaier:
The author affiliation is missing in the current jats templates. Is that on purpose?

See: https://github.com/jgm/pandoc/blob/34151e8da8e1863b7954d87c1713b97f3a944545/data/templates/article.jats_publishing#L88

It currently reads:

```
$if(author.email)$
<email>$author.email$</email>
$endif$
$if(author.aff-id)$
<xref ref-type="aff" rid="aff-$contrib.aff-id$"/>
$endif$
```

Shouldn't that be:

```
$if(author.email)$
<email>$author.email$</email>
$endif$
$if(author.affiliation)$
<aff>$author.affiliation$</aff>
$endif$
$if(author.aff-id)$
<xref ref-type="aff" rid="aff-$contrib.aff-id$"/>
$endif$
```

Should I start a PR, or am I missing something?

jgm:
I don't know a lot about JATS -- pinging @tarleb, @mfenner.

tarleb:
Might be an oversight, but I'm not sure. Could also be because there is no commonly accepted way of specifying this info, or because adding the same affiliation repeatedly would be undesirable. A separate, top-level `affiliation` field would probably be better to avoid duplication. Beamer, on the other hand, uses `institute` instad of `affiliation`, so that could lead to confusion. Maybe the current state is the result of having no clear solution to all that.

This is probably a good place to mention a new effort to agree on a method to specify this kind of metadata in YAML headers: <https://github.com/jam-schema/jams>. @mfenner already pointed us to https://github.com/datacite/bolognese which could be extremely handy in this context.

hubgit:
It does seem to be a bug/oversight that the affiliation is referenced by id (the `rid` attribute) but not added to the output.  I'd expect `author.aff-id` to be an array of affiliation identifiers, as well.

The affiliations should be listed separately in `aff` elements, ideally inside the `article-meta` element.

An alternative would be to add each author's affiliations in `aff` elements inside the relevant author's `contrib` element, in which case they don't need identifiers.

hubgit:
I think the actual array of author affiliations is stored in `author.institute`?

hubgit:
I've had a go at this, but it's not particularly easy as the data model isn't really there yet.

[...]

PR #6867, opened 2020-11-20
JATS writer: support author affiliations

tarleb:
Closes: #6687

### DIFF
--- a/data/templates/affiliations.jats
+++ b/data/templates/affiliations.jats
@@ -1,0 +1,35 @@
+$--
+$-- Affiliations
+$--
+$for(affiliation)$
+<aff id="aff-$affiliation.id$">
+$-- wrap affiliation if it has a known institution identifier
+$if(affiliation.group)$
+<institution content-type="group">$affiliation.group$</institution>
+$endif$
+$if(affiliation.department)$
+<institution content-type="dept">$affiliation.department$</institution>
+$endif$
+<institution-wrap>
+$if(affiliation.organization)$
+<institution>$affiliation.organization$</institution>
+$else$
+<institution>$affiliation.name$</institution>
+$endif$
+$if(affiliation.isni)$
+<institution-id institution-id-type="ISNI">$affiliation.isni$</institution-id>
+$endif$
+$if(affiliation.ringgold)$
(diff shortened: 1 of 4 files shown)


hubgit:
@tarleb I think this looks good. 

Will there be a corresponding addition to the JATS reader?

data/templates/affiliations.jats
```diff
@@ -0,0 +1,35 @@
+$--
+$-- Affiliations
+$--
+$for(affiliation)$
+<aff id="aff-$affiliation.id$">
+$-- wrap affiliation if it has a known institution identifier
```
hubgit, line 6: Should the `<institution-wrap>` opening line be below this comment, before all the `institution` elements?

Also it looks like `<institution-wrap>` is always added, without testing for the presence of an identifier (which is fine, just disagrees with the comment).

data/templates/affiliations.jats
```diff
@@ -0,0 +1,35 @@
...
+$if(affiliation.department)$
+<institution content-type="dept">$affiliation.department$</institution>
+$endif$
+<institution-wrap>
+$if(affiliation.organization)$
+<institution>$affiliation.organization$</institution>
+$else$
+<institution>$affiliation.name$</institution>
+$endif$
+$if(affiliation.isni)$
```
hubgit, line 19: Might it be more future-proof to leave the institution identifiers open and loop over an array of objects with `id` and `id-type` (and perhaps `vocab` and  `vocab-identifier`) properties?

tarleb:
Thanks! I plan to modify the reader eventually, but it will take a good bit longer.

data/templates/affiliations.jats
tarleb: I thought about doing that. The reason is that `<institution-wrap>`, IIUC, cannot contain PCDATA, so we cannot use commas to separate the different `<institute>` elements as done in the current version.

The comment is indeed outdated, thanks for pointing that out.

data/templates/affiliations.jats
tarleb: This is the *status quo* for all identifiers right now, including orcid, pmid, pmcid, etc. The current variant is consistent with that. We can add the generic variant on top of that.

PR #6877, opened 2020-11-22
HTML reader: extract submodules

tarleb:
Reducing module size should reduce memory use during compilation.

This is preparatory work to tackle support for more table features.


tarleb:
Merging now, as the changes make it easier for me to continue with table parsing. Happy to revert if it turns out to be unwanted.

PR #6884, opened 2020-11-24
Docx writer: Fix bullets/lists indentation

cholonam:
This pull request fixes the appearance of lists' bullets/numbers so that the first level is slightly indented to the right instead of right on the margin). This is similar to the way they are shown in other output formats like LaTeX or HTML, and also how they are shown in MS Word using the Default template (refer to the below screenshot).

I have regenerated the golden files and verified they open on Word 2010 and 2013.

image


jgm:
Thanks - I merged this manually via the command line, so I'll close this PR.

jgm:
By the way, in checking the effect your change had on the test suite, I discovered a bug that existed already before: #6894.

PR #6895, opened 2020-11-27
EPUB cover image: preserveAspectRatio="xMidYMid"

codelabor:
- as-is: preserveAspectRatio="none"
- to-be: preserveAspectRatio="xMidYMid"

It is not bad to have a full screen, but the cover image is distorted because the aspect ratio is not preserved.

* https://github.com/jgm/pandoc-templates/pull/15


jgm:
Thank you!

PR #6898, opened 2020-11-28
LaTeX reader: support (ignore) stackengine

ip1981:


jgm:
We support commands from some more common packages, but it would be madness to try to support every package. I've never heard for stackengine. What does it do?  Why do you need to ignore this?  Won't it be parsed as raw tex, which will be ignored anyway in most formats?

ip1981:
It turned out to be the simplest way to adjust (vertically) table cells in some cases. But since pandoc doesn't understand it, the content of the cells is lost in translation 😏

So the stackengine is a means to make an original document neat and this PR fixes translation to other forms.

jgm:
Could you work around this by giving a definition of the macro in your document:
```
\newcommand{\addstackgap}[1]{\rule{0pt}{1.5em}#1}
```
or something like that?

ip1981:
> Could you work around this by giving a definition of the macro in your document:
> 
> ```
> \newcommand{\addstackgap}[1]{\rule{0pt}{1.5em}#1}
> ```
> 
> or something like that?

I tries this (`\rule`) too :) It adds an unwanted `hr` in HTML for example.

jgm:
Maybe try with `\hspace`?

ip1981:
> Maybe try with `\hspace`?

And that too (assuming `\vspace`) :) And `\\[2em]` neither do the right thing:

Screenshot_2020-11-29_20-28-13

jgm:
I've just pushed a fix so that `\rule` with width 0 will no longer create a horizontal rule.
That may help with the first solution proposed above.

test/command/latex-stackengine.md
```diff
@@ -0,0 +1,18 @@
+# stackengine
```
jgm, line 1: This should be renamed `test/command/6898.md` -- that makes it easier for us to locate the issue to which a test corresponds.

PR #6908, opened 2020-12-02
docs: adds a new Pandoc filtering framework

jun-sheaf:
The new package linked for NodeJS supersedes the previous one. It contains significantly better typing, better documentation, better CI, and more.


jgm:
"supersedes" -- I didn't see a deprecation notice on the original package.
Does the author agree that this new package supersedes it?
IF not, should we perhaps link to both?

jun-sheaf:
The intention was to preserve the current docs style, however if you don't mind listing both, I've appended them both in a more stylistic way.

jgm:
great, thanks

PR #6922, opened 2020-12-04
Docbook writer: handle admonitions

jtojnar:
Similarly to https://github.com/jgm/pandoc/commit/d6fdfe6f2bba2a8ed25d6c9f11861774001f7a91, we should handle admonitions in Docbook writer as well.

### DIFF
--- a/src/Text/Pandoc/Writers/Docbook.hs
+++ b/src/Text/Pandoc/Writers/Docbook.hs
@@ -53,6 +53,13 @@ getStartLvl opts =
        TopLevelSection -> 1
        TopLevelDefault -> 1
 
+-- | Get correct name for the id attribute based on DocBook version.
+-- DocBook 4 used custom id attribute but DocBook 5 adopted the xml:id specification.
+-- https://www.w3.org/TR/xml-id/
+idName :: DocBookVersion -> Text
+idName DocBook5 = "xml:id"
+idName DocBook4 = "id"
+
 -- | Convert list of authors to a docbook <author> section
 authorToDocbook :: PandocMonad m => WriterOptions -> [Inline] -> DB m B.Inlines
 authorToDocbook opts name' = do
@@ -174,10 +181,7 @@ blockToDocbook opts (Div (id',"section":_,_) (Header lvl _ ils : xs)) = do
                                               then "section"
                                               else "sect" <> tshow n
                  _                    -> "simplesect"
-      idName = if version == DocBook5
-                 then "xml:id"
-                 else "id"
-      idAttr = [(idName, writerIdentifierPrefix opts <> id') | not (T.null id')]
+      idAttr = [(idName version, writerIdentifierPrefix opts <> id') | not (T.null id')]
       -- We want to add namespaces to the root (top-level) element.
       nsAttr = if version == DocBook5 && lvl == getStartLvl opts && isNothing (writerTemplate opts)
       -- Though, DocBook 4 does not support namespaces and
@@ -188,11 +192,12 @@ blockToDocbook opts (Div (id',"section":_,_) (Header lvl _ ils : xs)) = do
   title' <- inlinesToDocbook opts ils
   contents <- blocksToDocbook opts bs
   return $ inTags True tag attribs $ inTagsSimple "title" title' $$ contents
-blockToDocbook opts (Div (ident,classes,_) bs) =
-  let identAttribs = [("id", ident) | not (T.null ident)]
+blockToDocbook opts (Div (ident,classes,_) bs) = do
+  version <- ask
+  let identAttribs = [(idName version, ident) | not (T.null ident)]
       admonitions = ["attention","caution","danger","error","hint",
                      "important","note","tip","warning"]
-  in case classes of
+  case classes of
     (l:_) | l `elem` admonitions -> do
         let (mTitleBs, bodyBs) =
                 case bs of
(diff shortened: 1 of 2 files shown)


jtojnar:
Seems to work fine with `docbook`→`docbook` conversion:

```ShellSession
$ cabal run pandoc -v0 -- test.xml -f docbook -t docbook5
<section xml:id="sect">
  <title>sect</title>
  <note xml:id="foo">
    <title>foo</title>
    <para>
      content
    </para>
  </note>
</section
```

And markdown too, as long as I do not use a title:

```markdown
# sect
::: note
content
:::
```

```ShellSession
$ cabal run pandoc -v0 -- test.md -f markdown -t docbook
<section xml:id="sect">
  <title>sect</title>
  <note>
    <para>
      content
    </para>
  </note>
</section>
```

but once I add a heading it breaks:

```markdown
# sect
::: note
# foo
content
:::
```

```ShellSession
$ cabal run pandoc -v0 -- test2.md -f markdown -t docbook
<section xml:id="sect">
  <title>sect</title>
  <para>
  </para>
</section>
<section xml:id="foo">
  <title>foo</title>
  <para>
    content
  </para>
</section>
```

It does not even seem to visit the `Div` branch of `blockToDocbook`.

If I use the block syntax for the title too, as expected by the AST, it is parsed as `note` but then there is a paragraph in the title. And it is also quite ugly:

```markdown
# sect
::: note
:::: title
foo
::::
content
:::
```

```ShellSession
<section xml:id="sect">
  <title>sect</title>
  <note>
    <title><para>
      foo
    </para></title>
    <para>
      content
    </para>
  </note>
</section>
```

jgm:
It would be good to have some tests for this feature. (At least command tests (see test/command).)

jtojnar:
Thanks, wanted to ask about how to best test this but forgot about it. Command tests will work great, I think. Actually, just noticed `test/Tests` directory, which I somehow missed when looking the last time. I might just create unit test instead.

jtojnar:
I have added unit tests for the admonitions as well as the rest of the `Div` handling code paths. That should be nicer than command tests.

I have not added tests for `xml:id` usage since it is straightforward and the tests would be the same, except for using `xml:id` and `version = DocBook5`.

jtojnar:
I have also removed the attempt at specifying admonition titles from Markdown as they are not actually defined anywhere and my attempt fails for reasons I do not understand:

```patch
diff --git a/src/Text/Pandoc/Writers/Docbook.hs b/src/Text/Pandoc/Writers/Docbook.hs
index 398920839..c5c3f4208 100644
--- a/src/Text/Pandoc/Writers/Docbook.hs
+++ b/src/Text/Pandoc/Writers/Docbook.hs
@@ -203,6 +203,8 @@ blockToDocbook opts (Div (ident,classes,_) bs) = do
                 case bs of
                   -- Matches AST produced by the Docbook reader.
                   (Div (_,["title"],_) ts : rest) -> (Just (blocksToDocbook opts ts), rest)
+                  -- Matches AST generated from Markdown, TODO: I was not able to trigger this branch.
+                  (Header _lvl _ ts : rest) -> (Just (inlinesToDocbook opts ts), rest)
                   _ -> (Nothing, bs)
         admonitionTitle <- case mTitleBs of
                               Nothing -> return mempty
diff --git a/test/Tests/Writers/Docbook.hs b/test/Tests/Writers/Docbook.hs
index 1d53dcfe7..a0fc2736e 100644
--- a/test/Tests/Writers/Docbook.hs
+++ b/test/Tests/Writers/Docbook.hs
@@ -93,6 +93,19 @@ tests = [ testGroup "line blocks"
                                     , "  </para>"
                                     , "</attention>"
                                     ]
+          , "admonition-with-heading" =:
+                            divWith ("foo", ["attention"], []) (
+                              header 1 (text "This is title") <>
+                              para "This is a test"
+                            )
+                              =?> unlines
+                                    [ "<attention id=\"foo\">"
+                                    , "  <title>This is title</title>"
+                                    , "  <para>"
+                                    , "    This is a test"
+                                    , "  </para>"
+                                    , "</attention>"
+                                    ]
           , "single-child" =:
                             divWith ("foo", [], []) (para "This is a test")
                               =?> unlines
```


```haskell
admonition-with-heading: FAIL
  test/Tests/Helpers.hs:50:
  
  [Div ("foo",["attention"],[])
   [Header 1 ("",[],[]) [Str "This",Space,Str "is",Space,Str "title"]
   ,Para [Str "This",Space,Str "is",Space,Str "a",Space,Str "test"]]]
  - <attention id="foo">
  + <sect1 id="foo">
      <title>This is title</title>
      <para>
        This is a test
      </para>
  - </attention>
  + </sect1>
```

It would be nice to be able to use admonition titles from Markdown but we only have four admonitions with titles in Nixpkgs docs so getting much more common general admonitions merged takes priority.

jtojnar:
I guess it makes most sense to handle the AST produced by the Markdown reader so I have started stripping the `Para` inside `Div.title`.

Alternative would be changing the Docbook reader to use a `Header 1` for the admonition title but that would be potentially confusing or invalid for some readers/writers like it is for the `makeSections` function called by the DocBook writer. We would need to make sure all readers and writers support sectioning as described, for example, in the [HTML 5 specification](https://html.spec.whatwg.org/multipage/sections.html#headings-and-sections):

> Sections may contain headings of any rank, but authors are strongly encouraged to either use only `h1` elements, or to use elements of the appropriate rank for the section's nesting level.

jgm:
This looks good to me -- thanks!

PR #6926, opened 2020-12-06
Use Data.Monoid.First

mlang:


jgm:
Note that the haddocks for Data.Monoid.First say:
"Use of this type is discouraged. Note the following equivalence:"

src/Text/Pandoc/App/FormatHeuristics.hs
```diff
@@ -15,16 +15,13 @@ module Text.Pandoc.App.FormatHeuristics
...
 import System.FilePath (takeExtension)
 
 -- Determine default format based on file extensions.
 formatFromFilePaths :: [FilePath] -> Maybe Text
-formatFromFilePaths [] = Nothing
-formatFromFilePaths (x:xs) =
-  case formatFromFilePath x of
-    Just f     -> Just f
-    Nothing    -> formatFromFilePaths xs
+formatFromFilePaths = fmap getFirst . foldMap (fmap First . formatFromFilePath)
```
jgm, line 25: I'm curious, what would you say is the advantage of this change?  Although it shortens the code, it's now harder to see what it does.  It's also less obvious that it stops processing on the first match (though you could figure this out if you were good at reasoning about laziness in Haskell).

dixonary:
I would observe that `msum` will perform much of the same job (taking the first Just from a list, or Nothing if no such Just exists) with more obvious semantics.

tarleb:
This PR seemed stale, so I've taken the freedom to incorporate the ideas here in a separate commit. See 35688c42627dce4641f4f61253e3d3786452b61a.

Thanks @mlang for the PR, and thanks @dixonary for hinting at `msum`.

PR #6941, opened 2020-12-09
Docx writer: keep raw openxml strings verbatim

tarleb:


jjallaire:
I tried this out and it works exactly as expected. IMO this opens up a huge solution space for docx/pptx as you can now create arbitrarily sophisticated openxml structures and render markdown within them. Tables are certainly one example but you can imagine many others, especially within pptx.

tarleb:
Great to hear, thanks for testing. The remaining cleanup should be done within the next days.

tarleb:
Should be good now.

jgm:
Excellent!

Issue #6313, opened 2020-04-24
LaTeX writer - support new table features


jgm:
Adjust LaTeX writer to support new table features introduced in
jgm/pandoc-types#66
including table attributes (including identifier), rowspan, colspan, table head and foot, multiple header lines, row headers, captions that allow block-level content and include an optional short caption.

stroobandt:
# The CALS table model
Perhaps of interest to this discussion is the [the CALS table model](https://en.wikipedia.org/wiki/CALS_Table_Model).
[CALS](https://en.wikipedia.org/wiki/CALS_%28DOD%29) started out as a United States Department of Defense (DoD) initiative for creating a DTD standard for representing tables in SGML/XML.

CALS allows for cell merging.

CALS is also the historic precursor of the HTML table model. A little known fact is that CALS is supported in LaTeX through the `cals` package. In fact, it is easier to use and offers more capabilities than the more commonly employed `tabularx` package in LaTeX.

Here is the documentation of the `cals` package in LaTeX:

 - <http://www.ctan.org/tex-archive/macros/latex/contrib/cals> (demo & doc)
 - <http://tug.org/TUGboat/tb32-2/tb101parashchenko.pdf> (user's manual)

the-solipsist:
Just FYI: There's an existing open issue on the CALS table model: https://github.com/jgm/pandoc/issues/890 (opened by @stroobandt himself in 2013).

stroobandt:
@the-solipsist Thanks for pulling [CALS](https://en.wikipedia.org/wiki/CALS_Table_Model) out of the woodwork again!

It surely is a nice standard to work from and it is already supported in at least LaTeX and by definition, in HTML.
Furthermore, it allows to merge cells. This is currently not supported in Pandoc Markdown.

If the CALS standard is good enough for the US military, it is probably also good enough for the rest of us.

tarleb:
Possibly incomplete list of missing features:

- [ ] Row headers
- [ ] Intermediate header rows (body headers)
- [ ] footers

Should we open separate issues for each of these? Implementing table footers might be an excellent first issue.

PR #6950, opened 2020-12-14
LaTeX writer: support colspans and rowspans in tables.

tarleb:
See also: #6313

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -254,7 +254,10 @@ $if(highlighting-macros)$
 $highlighting-macros$
 $endif$
 $if(tables)$
-\usepackage{longtable,multirow,booktabs,array}
+\usepackage{longtable,booktabs,array}
+$if(multirow)$
+\usepackage{multirow}
+$endif$
 \usepackage{calc} % for calculating minipage widths
 $if(beamer)$
 \usepackage{caption}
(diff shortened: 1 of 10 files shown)


tarleb:
- [ ] add tests
- [ ] fix big producing invalid LaTeX for some tables

tarleb:
Potentially controversial choices in this PR:
- Package `multirows` is now required in the default template.
- Content of row-spanning cells is centered vertically. This is the default, but doesn't always look good in the resulting PDF.
- Tables with colspans or rowspans always fill the full columnwidth. Otherwise cells tend to get too wide.

jgm:
> Tables with colspans or rowspans always fill the full columnwidth. Otherwise cells tend to get too wide.

I'm not sure what you mean here.  Why would forcing tables to fill the full columnwidth make cells narrower?  Wouldn't it make them wider?

Seems to me there are cases where you'd want a narrow table with a colspan or rowspan.
 
> multirow

Slight drawback to this is that it's in texlive-latex-extra on debian, so bringing it in requires bringing in a lot else. I've tried to avoid requiring things in texlive-latex-extra so far -- though maybe we already depend on something there.

I assume there's no good way to do rowspan without this package?

I suppose one option would be to include the package only if rowspans are actually used.  Or we could just bite the bullet. It's entirely possible that we already require packages in texlive-latex-extra.

tarleb:
The multirow package is now only loaded if really needed.

Please ignore my previous comment about table widths. I found that this was based on a faulty test of mine.

jgm:
Great, thanks!

PR #6967, opened 2020-12-20
Fixed table with file extensions and interpreters

wuffi:
Assigned .pl file extension to perl interpreter and .rb to ruby


tarleb:
Whoops. Thanks!

PR #6972, opened 2020-12-21
added support for writing nested tables to asciidoc

### LINKED ISSUE #6942: asciidoc writer doesn't support nested tables
AsciiDoc supports nested tables, see https://asciidoctor.org/docs/user-manual/#nested-tables  
In the nested table pipes are replaced by exclamation marks. Below find input, expected output, command, actual output.

```
$ cat nested.html
<!doctype html>
<html>
<head>
<meta charset="utf-8">
<title> NestedTables </title>
</head>
<body>
<table>
 <tr>
  <td >
   <table>  <tr> <td> a1 </td> <td> a2 </td> </tr>  </table>
  </td>
  <td>b</td>
 </tr>
 <tr>
   <td>c</td> <td>d </td>
 </tr>
</table>
</body>
</html>

$ cat expected.asciidoc
[width="100%",cols="50%,50%",]
|===
a|
[cols=",",]
!===
!a1 !a2
!===

|b
|c |d
|===

$ pandoc -f html -t asciidoc nested.html 
[width="100%",cols="50%,50%",]
|===
a|
[cols=",",]
|===
|a1 |a2
|===

|b
|c |d
|===
```

timo-a:
This PR fixes #6942.

A new field is introduced in the WriterState to track the recursion level when writing nested tables.
Tables within tables ("level 1") are written correctly and deeper nested tables are written as level 1 and enclosed by a comment explaining that the nesting depth is not supported by asciidoc.

I did not write any new tests because I don't know how. 
I found `pandoc/test/tables.asciidoc` and that it is used but I don't know which direction the conversion goes and how to extend the source table without affecting other writers' tests.
A test would verify that for:
[table.asciidoc.txt](https://github.com/jgm/pandoc/files/5725516/table.asciidoc.txt)
[table.html.txt](https://github.com/jgm/pandoc/files/5725518/table.html.txt)
[tableThriceNested.asciidoc.txt](https://github.com/jgm/pandoc/files/5725519/tableThriceNested.asciidoc.txt)
[tableThriceNested.html.txt](https://github.com/jgm/pandoc/files/5725520/tableThriceNested.html.txt)

```
stack exec pandoc -- -f html -t asciidoc tableThriceNested.html -o tableThriceNested.asciidoc
stack exec pandoc -- -f html -t asciidoc table.html -o table.asciidoc
```

### DIFF
--- a/src/Text/Pandoc/Writers/AsciiDoc.hs
+++ b/src/Text/Pandoc/Writers/AsciiDoc.hs
@@ -36,7 +36,6 @@ import Text.DocLayout
 import Text.Pandoc.Shared
 import Text.Pandoc.Templates (renderTemplate)
 import Text.Pandoc.Writers.Shared
-import Prelude hiding (log)
 
 
 data WriterState = WriterState { defListMarker       :: Text
@@ -63,7 +62,6 @@ defaultWriterState = WriterState { defListMarker      = "::"
                                  , inList             = False
                                  , hasMath            = False
                                  , tableNestingLevel  = 0
-                                 , log = []
                                  }
 
 -- | Convert Pandoc to AsciiDoc.
--- a/src/Text/Pandoc/Writers/AsciiDoc.hs
+++ b/src/Text/Pandoc/Writers/AsciiDoc.hs
@@ -45,6 +45,10 @@ data WriterState = WriterState { defListMarker       :: Text
                                , asciidoctorVariant  :: Bool
                                , inList              :: Bool
                                , hasMath             :: Bool
+                               -- |0 is no table
+                               -- 1 is top level table
+                               -- 2 is a table in a table
+                               , tableNestingLevel   :: Int
                                }
 
 defaultWriterState :: WriterState
@@ -56,6 +60,7 @@ defaultWriterState = WriterState { defListMarker      = "::"
                                  , asciidoctorVariant = False
                                  , inList             = False
                                  , hasMath            = False
+                                 , tableNestingLevel  = 0
                                  }
 
 -- | Convert Pandoc to AsciiDoc.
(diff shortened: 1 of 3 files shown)


jgm:
Thanks for this!  For tests, it's probably easiest to add a "command test" -- see `test/command`. It should be fairly self-explanatory what to do if you look at some examples.
Instead of adding a comment to the source, I think it would make more sense to use pandoc's logging mechanism to emit a warning.  This could be a "skipped content" warning if you're going to omit the table.  Or maybe better to include the nested table even though it won't be right, and print a special warning. Maybe we need a new constructor on LogMessage for this case.

timo-a:
I replaced the comments with log messages and added command tests.  
BTW is there a way to execute a single command test?

jgm:
To run a single test you can do
```
stack test --test-arguments='-p pandoc-citeproc-82'
```

jgm:
I think I'd actually prefer just using `BlockNotRendered` and not adding a new constructor to `LogMessage`.  The overly nested table would be omitted and BlockNotRendered would signal the issue.  (In truth, I don't think this will ever come up...who uses triply nested tables?)

timo-a:
OK, I switched to `BlockNotRendered`

src/Text/Pandoc/Writers/AsciiDoc.hs
```diff
@@ -45,6 +47,11 @@ data WriterState = WriterState { defListMarker       :: Text
                                , asciidoctorVariant  :: Bool
                                , inList              :: Bool
                                , hasMath             :: Bool
+                               -- |0 is no table
+                               -- 1 is top level table
+                               -- 2 is a table in a table
+                               , tableNestingLevel   :: Int
+                               , log :: [LogMessage]
```
jgm, line 54: Could we change this to `logMessages` to avoid the need to shadow `log` from prelude?

test/command/nested-table-to-asciidoc-6942.md
```diff
@@ -0,0 +1,81 @@
...
+!===
+!a1 !a2
+!===
+
+|b
+|c |d
+|===
+```
+
+A table within a table within a table cannot be converted because asciidoc only supports two levels of tables.
```
jgm, line 39: We try to keep our source files wrapped to <= 80 chars.

jgm: Of course that isn't always possible in these command tests (because the output or input may need to exceed 80 chars), but that's the style we use for the textual parts.

test/command/nested-table-to-asciidoc-6942.md
```diff
@@ -0,0 +1,81 @@
...
+   <td>c</td> <td>d </td>
+ </tr>
+</table>
+</body>
+</html>
+^D
+[INFO] Not rendering Table ("",[],[]) (Caption Nothing []) [(AlignDefault,ColWidth 0.5),(AlignDefault,ColWidth 0.5)] (TableHead ("",[],[]) []) [TableBody ("",[],[]) (RowHeadColumns 0) [] [Row ("",[],[]) [Cell ("",[],[]) AlignDefault (RowSpan 1) (ColSpan 1) [Plain [Str "a1"]],Cell ("",[],[]) AlignDefault (RowSpan 1) (ColSpan 1) [Table ("",[],[]) (Caption Nothing []) [(AlignDefault,ColWidthDefault),(AlignDefault,ColWidthDefault)] (TableHead ("",[],[]) []) [TableBody ("",[],[]) (RowHeadColumns 0) [] [Row ("",[],[]) [Cell ("",[],[]) AlignDefault (RowSpan 1) (ColSpan 1) [Plain [Str "1"]],Cell ("",[],[]) AlignDefault (RowSpan 1) (ColSpan 1) [Plain [Str "2"]]]]] (TableFoot ("",[],[]) [])]]]] (TableFoot ("",[],[]) [])
+[width="100%",cols="50%,50%",]
+|===
+a|
```
jgm, line 72: What is this `a` in line 72?

src/Text/Pandoc/Writers/AsciiDoc.hs
jgm: Is this even used?

test/command/nested-table-to-asciidoc-6942.md
timo-a: Letters before `|` hold style information about the following cell.  
`... cell0 a| cell1 ...` indicates that the content of cell1 is to be treated as an asciidoc file, so nested tables, list, source code blocks are all possible as content.
https://asciidoctor.org/docs/user-manual/#nested-tables

In line 72, the `a` enables a nested table in the first cell of the first row.

test/command/nested-table-to-asciidoc-6942.md
timo-a, line 39:
```suggestion
A table within a table within a table cannot be converted because asciidoc only
supports two levels of tables.
```

src/Text/Pandoc/Writers/AsciiDoc.hs
timo-a, line 54:
```suggestion
```

src/Text/Pandoc/Writers/AsciiDoc.hs
```diff
@@ -56,6 +63,8 @@ defaultWriterState = WriterState { defListMarker      = "::"
                                  , asciidoctorVariant = False
                                  , inList             = False
                                  , hasMath            = False
+                                 , tableNestingLevel  = 0
+                                 , log = []
```
timo-a, line 67:
```suggestion
```

src/Text/Pandoc/Writers/AsciiDoc.hs
```diff
@@ -36,6 +36,8 @@ import Text.DocLayout
 import Text.Pandoc.Shared
 import Text.Pandoc.Templates (renderTemplate)
 import Text.Pandoc.Writers.Shared
+import Prelude hiding (log)
```
timo-a, line 39:
```suggestion
```

src/Text/Pandoc/Writers/AsciiDoc.hs
timo-a: It is not used after all, thanks for catching that, I removed it.

jgm:
Great, thanks!

PR #6982, opened 2020-12-25
Add the `description` meta tag to the HTML templates

jerry-sky:
The `description` meta tag will make the generated HTML documents more complete and SEO-friendly.

### DIFF
--- a/data/templates/default.html4
+++ b/data/templates/default.html4
@@ -13,6 +13,9 @@ $endif$
 $if(keywords)$
   <meta name="keywords" content="$for(keywords)$$keywords$$sep$, $endfor$" />
 $endif$
+$if(description)$
+  <meta name="description" content="$description$" />
+$endif$
   <title>$if(title-prefix)$$title-prefix$ – $endif$$pagetitle$</title>
   <style type="text/css">
     $styles.html()$
--- a/data/templates/default.html5
+++ b/data/templates/default.html5
@@ -13,6 +13,9 @@ $endif$
 $if(keywords)$
   <meta name="keywords" content="$for(keywords)$$keywords$$sep$, $endfor$" />
 $endif$
+$if(description)$
+  <meta name="description" content="$description$" />
+$endif$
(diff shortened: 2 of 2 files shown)


mb21:
Weird that we didn't have that already... then again, from https://moz.com/learn/seo/meta-description:

> Google [announced](https://webmasters.googleblog.com/2009/09/google-does-not-use-keywords-meta-tag.html) in September of 2009 that neither meta descriptions nor meta keywords factor into Google's ranking algorithms for web search. 

Although sometimes it's indeed used for the text snippet displayed in Google's search results page.

jerry-sky:
> Although sometimes it's indeed used for the text snippet displayed in Google's search results page.

Yes, that’s what I’m aiming for, and I’m aware that Google does not take those meta tags into consideration, but from what I have found it seems that the `description` meta tag is in fact used when displaying the website in Google Search.

jgm:
No objections to providing a way to get descriptions in there.  But this way of doing it is likely to cause problems.  The `descriptions` metadata field will be rendered as HTML, and would need escaping to be put into an attribute, wouldn't it?  What if the quotes or emphasis occur in it, for example?

This makes me think we should provide a doctemplates `pipe` to escape HTML; it would simplify these templates.

jgm:
In the HTML writer we'll need something like
```
maybe id (defField "description-meta") descriptionMeta .
```
(around line 325) and then (around line 257)
```
let descriptionMeta = stringifyHTML <$> lookupMetaInlines "description" meta
```
and then you'd use `description-meta` in the template.  This will ensure that the values there are properly escaped.

jgm:
If you don't have a Haskell development environment set up, I can do that part for you.

jerry-sky:
> If you don't have a Haskell development environment set up, I can do that part for you.

Yes, I would really appreciate that, because I’ve never had any experiences with Haskell development. This PR is more of a _suggestion_.

jgm:
OK, I'll merge this then and add something to the writer.

PR #6985, opened 2020-12-26
Powerpoint writer: allow arbitrary OOXML in raw inline elements

tarleb:
The raw text is now included verbatim in the output. Previously is was parsed
into XML elements, which prevented the inclusion of partial XML snippets.


tarleb:
Pinging @jjallaire in case you'd like to take a look already. This is only for raw inlines, not for raw blocks.

It seems that raw blocks are currently discarded when producing pptx; changing that requires more elaborate refactoring.

jjallaire:
Thanks! I gave it a try and it does indeed work as expected.

tarleb:
Do you happen to have a small example which we could for testing? I failed to come up with a good a test case.

jjallaire:
This is what I used to confirm that XML tags were left alone and not parsed:

```lua

local para = pandoc.Para({
  pandoc.Str("Foo "),
  pandoc.RawInline("openxml", "<a:r>"),
  pandoc.RawInline("openxml", "<a:rPr i=\"1\"/><a:t>Bar</a:t>"),
  pandoc.RawInline("openxml", "</a:r>")
})
```

I also tried to find something that would intersperse Pandoc tokens w/ openxml RawInline. The best thing I could think of was a filter that would auto-link certain words. However, I found that the link target is just an alias to a URL stored elsewhere in the bundle so wouldn't have worked:

```xml
<a:r>
  <a:rPr>
      <a:hlinkClick r:id="rId2"/>
   </a:rPr>
   <a:t>link</a:t>
</a:r>
```

I think the most interesting examples are indeed of blocks (for custom slide layouts). That said, that's probably the *only* good example we can come up with for PowerPoint, and given that it requires refactoring it would probably be better to just directly support additional layouts (as is done now for 2-column layout) since the number of relevant layouts is small and nearly covered as is:

e518ef5c-ab01-427e-9d55-e80712065809

The biggest issue I've seen in approximating these default layouts is that it appears as if pptx columns support only 1 block and don't support widths. For example, the following markdown:

```markdown
## Slide 

::: {.columns}

::: {.column width=20%}
Elephant

![](elephant.png)
:::

::: {.column width=80%}
Walrus

![](walrus.png)
:::

:::
```

Is rendered as:

Screen Shot 2020-12-27 at 6 59 43 AM

tarleb:
I found and fixed a small bug. The tests now required a tiny change, which makes me confident that everything is working as intended.

This change now covers inlines as well as *some* blocks. I'll open a separate PR for the remaining blocks in a moment, as that change is more involved and I'm not sure it's actually necessary.

PR #6988, opened 2020-12-27
Powerpoint writer: support raw OOXML blocks everywhere

tarleb:


tarleb:
This is the second and last part required to pass all raw OOXML blocks through. To be perfectly honest, I don't fully understand the conditions under which a raw block might be discarded. From a first reading of the code, it seems to me that the proposed changes are actually unnecessary. I was already half done when I realized this, so here's a PR just in case.
Cc: @jjallaire

jjallaire:
This may prove difficult to test (or perhaps even difficult for filter writers to take advantage of). There appear to be side effects of AST rendering that can't be reproduced in a filter writing raw blocks. For example, here's a simple two-column slide:

```markdown
## Slide One

::: {.columns}
::: {.column}
First column
:::

::: {.column}
Second column
:::
:::
```

I inspected the generated slide xml and then replaced the column layout with the following raw block:

````markdown
## Slide One

```{=openxml}
<p:sp>
  <p:nvSpPr>
    <p:cNvPr id="3" name="Content Placeholder 2"/>
    <p:cNvSpPr>
      <a:spLocks noGrp="1"/>
    </p:cNvSpPr>
    <p:nvPr>
      <p:ph sz="half" idx="1"/>
    </p:nvPr>
  </p:nvSpPr>
  <p:spPr/>
  <p:txBody>
    <a:bodyPr/>
    <a:lstStyle/>
    <a:p>
      <a:pPr lvl="0" marL="0" indent="0">
        <a:buNone/>
      </a:pPr>
      <a:r>
        <a:rPr/>
        <a:t>First column</a:t>
      </a:r>
    </a:p>
  </p:txBody>
</p:sp>
<p:sp>
  <p:nvSpPr>
    <p:cNvPr id="4" name="Content Placeholder 3"/>
    <p:cNvSpPr>
      <a:spLocks noGrp="1"/>
    </p:cNvSpPr>
    <p:nvPr>
      <p:ph sz="half" idx="2"/>
    </p:nvPr>
  </p:nvSpPr>
  <p:spPr/>
  <p:txBody>
    <a:bodyPr/>
    <a:lstStyle/>
    <a:p>
      <a:pPr lvl="0" marL="0" indent="0">
        <a:buNone/>
      </a:pPr>
      <a:r>
        <a:rPr/>
        <a:t>Second column</a:t>
      </a:r>
    </a:p>
  </p:txBody>
</p:sp>
```
````

When rendering this second document, the generated slide xml is identical (as expected), however there is one diff in the pptx:

Screen Shot 2020-12-28 at 9 29 08 AM

This is in the slide's `_rels` file and it appears to be pointing the slide to a different layout (presumably a single column layout). This results in the following in the rendered pptx:

Screen Shot 2020-12-28 at 9 30 39 AM

So it appears as if it's not enough to emit slide layout openxml in a filter, you also need to write into `_rels` to indicate the slide's layout. This is obviously not something that a filter can do.

To me this argues that no changes to pptx openxml handling are warranted, because most of the things that users will try to accomplish won't work (even for inlines, the way pptx composes paragraphs you can't inject a styling inline into a paragraph the way you can with docx).

tarleb:
OK, thanks a lot for checking. I was afraid something like this would be the case. I'll close this then.

I'd like to look into this some more, but don't have the time right now. Just in case you'd be interested to let somebody work on this: there are currently a number of [excellent people](https://twitter.com/FlavioCorpa/status/1341755832367722497) looking for Haskell work.

jjallaire:
Okay, we may indeed want to dig into this further. In that case I'll be back in touch for recommendations on people that would be well suited for the project.

PR #6989, opened 2020-12-28
Wrap HTML5 body in <main> tag

phikal:
Wrapping the actual content of a HTML document in a [main](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main) tag, makes it easier for tools like [web readers](https://blog.mozilla.org/firefox/reader-view/) to identify what is actual part of the document, and what is a header.

This patch just wraps the `$body$` in such a tag, improving how documents generated by pandoc can be parsed. From my testing, it doesn't break anything.

### DIFF
--- a/data/templates/default.html5
+++ b/data/templates/default.html5
@@ -59,7 +59,9 @@ $endif$
 $table-of-contents$
 </nav>
 $endif$
+<main>
 $body$
+</main>
 $for(include-after)$
 $include-after$
 $endfor$


tarleb:
Thanks for the PR. My understanding of the `<main>` element was that it's used to mark the part of a page which is (somewhat) unique to that page. I.e., while it excludes navigation to other parts of a website, title and page-specific header *do* count as main content. I might be wrong, of course. Could you explain a bit how we can test this ourselves, and how it improves accessibility of an average page generated with pandoc?

phikal:
> Could you explain a bit how we can test this ourselves, and how it improves accessibility of an average page generated with pandoc?

I'm no expert, so my test is simple: 

1. I generate a page without the main tag, and a page with the main tag
2. I open both sites using Firefox' web reader
3. The first one includes the header, if you were to have the reader read the site out loud (which I'm guessing helps people with eyesight issues), would try to read that out too.
4. The second one doesn't, and is thus perfrable.

tarleb:
@jooyoungseo, it would be great to hear your thoughts on this.

jooyoungseo:
Either `<main>` or `<div role = "main">` should be used for a specific part of a web page. Let me give an example...

```html
<body>
  <header>
    <nav>
      <!-- Page navigation could go here. -->
      ...
    </nav>
  </header>
  <main>
    <!-- Main content is here. -->
  </main>
  <footer>
    ...
  </footer>
</body>
```

In sum, substituting `<main>` for `<body>` seems risky.

phikal:
JooYoung Seo <notifications@github.com> writes:

> [1:text/plain Hide]
>
> Either `<main>` or `<div role = "main">` should be used for a specific part of a web page. Let me give an example...
>
> ```html
> <body>
>   <header>
>     <nav>
>       
>       ...
>     </nav>
>   </header>
>   <main>
>     
>   </main>
>   <footer>
>     ...
>   </footer>
> </body>
> ```
>
> In sum, substituting `<main>` for `<body>` seems risky.

This suggestion isn't to replace <body> by <main>, but to wrap the
translated markdown content in a <main> tag.

-- 
	Philip K.

jooyoungseo:
I got it. However, in my humble opinion, it still has some risky assumption. We cannot overgeneralize `$body$` as a main landmark region. What if a user would pass the example code that I shared just like above?

That would result in having double main.

Like I explained, the main landmark should be used for a specific area region of content (not as a general container).

As an everyday screen reader user, I would get confused if the whole content passed to `$body$` was treated as a general main landmark.

phikal:
In that case, would it be better to conditionally allow adding the
<main> tag? Or if it's a more common case, to remove it?

-- 
	Philip K.

jooyoungseo:
Come to think of it, we could conditionally apply `<main>` only when there's no user-customized `<main>` found in `$body$`. However, I am not 100% sure if this will save us. Some post-processors could add `<main>` later. It's your call (@tarleb).

tarleb:
Thank you for the very insightful comments, @jooyoungseo! It seems to me that the benefits of this change depend on assumptions which may not hold true in general. There's certainly room for more semantic markup, but we'd need to document all assumptions really well if we wanted to do that in the default template.

@phikal, thank you for your efforts to improve pandoc, they are most welcome!

@jooyoungseo One more question if you allow: let's assume we were to build a separate, more semantic template for articles, and were to wrap everything in an `<article>` element. Would you say the article's headline should be part of the `<main>` content? How about a table of contents for the article?

jooyoungseo:
@tarleb, I am not sure whether I have correctly understood your meaning of head line of an article (`<article>`) tag, but I would simply sue `<h*>` tag for article title like below:

```html
<article>
  <h1>Pandic is Cool!</h1>
  <!-- Article body here. -->
</article>
```

Or, the other way around would work as well (I would recommend the following structure):

```html
<h1>Pandoc is Cool!</h1>
  <article>
    <!-- Article body here. -->
  </article>
```

As you know, the heading level should vary depending on the page structure.

Typically, `<main>` is used only one time per page, and we can put sub-sections using `<article>` or `<section>` under the `<main>` landmark region.

For TOC container, we could use `<nav aria-label = "Table of Contents">`, or `<div role = "navigation" aria-label = "Table of Contents">` something like that.

Please let me know if there would be anything I could further assist.

PR #6998, opened 2021-01-02
Replace org-mode’s verbatim from code to codeWith.

phaazon:
This adds the `"verbatim"` class so that exporters can apply a specific
style on it. For instance, it will be possible for HTML to add a CSS
rule for code + verbatim class.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/Inlines.hs
+++ b/src/Text/Pandoc/Readers/Org/Inlines.hs
@@ -572,7 +572,7 @@ underline :: PandocMonad m => OrgParser m (F Inlines)
 underline = fmap B.underline    <$> emphasisBetween '_'
 
 verbatim  :: PandocMonad m => OrgParser m (F Inlines)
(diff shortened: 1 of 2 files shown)


jgm:
@tarleb what  do you think?

tarleb:
I'd say that's a reasonable change. We'll deviate slightly from the default Emacs exporters, which treat code and verbatim exactly the same. But being able to differentiate between the two appears well worth it.

@phaazon, the tests in file `test/Tests/Readers/Org/Inline.hs` must be adjusted to match the new behavior.

phaazon:
Sure, I wanted to have your opinion on the change before @tarleb. I’m updating the change.

jgm:
Don't know if you know about this feature of the HTML writer:
```
% pandoc -t html
<span class="kbd">hi</span>
<p><kbd>hi</kbd></p>
```

tarleb:
Test failure due only to https://github.com/hslua/hslua/issues/92

Thanks!

Issue #7016, opened 2021-01-09
Fix JATS citations and references


tarleb:
Pandoc contains a special jats.csl to create properly formatted JATS references via pandoc-citeproc. This is no longer supported, and citations are misformatted.

The old method produced [`<element-citation>`](https://jats.nlm.nih.gov/archiving/tag-library/1.1/element/element-citation.html), so styling and presentation of citations was left to the JATS reader. It would be great if we could produce [`<mixed-citation>`](https://jats.nlm.nih.gov/archiving/tag-library/1.1/element/mixed-citation.html) elements, as element order and punctuation can be taken from the CSL. I suspect that simple mixed-citation elements are also very easy to produce.

I'd also like to look into tagging the components of a `<mixed-citation>`, but I think I need help to get started. If I understand correctly, then a good first step would be to define a new type and provide a `CiteprocOutput` instance. Is that correct? I guess we'd still have to find a way to integrate that in pandoc's processing pipeline.

jgm:
Could you say more about the problem?  If it's a problem with citeproc, it could perhaps be fixed there.

jgm:
Can you give an example of the sort of output you'd like to produce?

tarleb:
The issue is that pandoc-citeproc supported a mechanism that allowed the generation of raw JATS XML via a CSL style file. We still have `data/jats.csl` for this, and it is still used as the default when producing JATS. As a result, citations look like this:

    <xref alt="{{jats}}&lt;xref ref-type=&quot;bibr&quot; rid=&quot;{{/jats}}{{jats}}ref-{{/jats}}1{{jats}}&quot;&gt;{{/jats}}1{{jats}}&lt;/xref&gt;{{/jats}}" rid="ref-Upper_writers_1974">{{jats}}&lt;xref
  ref-type=&quot;bibr&quot;
  rid=&quot;{{/jats}}{{jats}}ref-{{/jats}}1{{jats}}&quot;&gt;{{/jats}}1{{jats}}&lt;/xref&gt;{{/jats}}</xref>

The references section used to look nice, but I think in-text citations never fully worked in JATS; I wrote some fairly complex hacks to get good results with pandoc-scholar, and always wanted to port some of the mechanics back into pandoc. I just didn't get around to it.

I don't think this needs fixing in citeproc, but in the writer. I'll see if I can get back basic support, i.e. citations that are nicely formatted but not tagged otherwise. Tagging will be more interesting, and I'll probably need help there.

[...]

PR #7018, opened 2021-01-10
JATS writer: fix citations

tarleb:
Citations are now output as `<mixed-citation>` elements. This preserves CSL formatting, but semantic information about authors, pages, etc. is lost.

See: #7016


jgm:
Maybe add a command test to show how this works and guard against future regressions?  There are already csl files you can use in the test/command directory.

jgm:
Sorry, I should have left the commits separate rather than squashing them.

PR #7021, opened 2021-01-12
LaTeX Reader: support for '\includesvg', '\svgpath'

k-alin:
LaTeX reader includes .svg files provided using '\includesvg', which may be not be located in the current working directory, whether or not the file name provided has the extension .svg.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -1983,9 +1983,10 @@ graphicsPath = do
   return mempty
 
 svgSourceExtn :: Text -> Text
-svgSourceExtn src = case (T.isSuffixOf (T.pack ".svg") src) of
-                           True -> src
-                           False -> T.pack $ (T.unpack src)++".svg"
+svgSourceExtn src = if T.isSuffixOf (T.pack ".svg") src then
+                      src
+                    else
+                      T.pack $ (T.unpack src) ++ ".svg"
 
 splitBibs :: Text -> [Inlines]
 splitBibs = map (str . T.pack . flip replaceExtension "bib" . T.unpack . trim) . splitTextBy (==',')
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -1986,7 +1986,7 @@ svgSourceExtn :: Text -> Text
 svgSourceExtn src = if T.isSuffixOf (T.pack ".svg") src then
                       src
                     else
-                      T.pack $ (T.unpack src) ++ ".svg"
+                      T.pack $ T.unpack src ++ ".svg"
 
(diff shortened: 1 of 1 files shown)


jgm:
Are these commands core latex, or defined in a package? If so, what package?

k-alin:
It's part of the svg package: https://ctan.org/tex-archive/graphics/svg

tarleb:
Thank you for the PR, @k-alin. I hope to have a look soon.

To explain the holdup a little: pandoc cannot support every LaTeX package, so we are a bit conservative when it comes to support non-standard packages. The ultimate merge decision lies with @jgm, and depends on the expected usefulness for others and the complexity that the feature adds to the code.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -222,7 +222,7 @@ mkImage options (T.unpack -> src) = do
    let attr = ("",[], kvs)
    let alt = str "image"
    defaultExt <- getOption readerDefaultImageExtension
-   let exts' = [".pdf", ".png", ".jpg", ".mps", ".jpeg", ".jbig2", ".jb2"]
+   let exts' = [".pdf", ".png", ".jpg", ".mps", ".jpeg", ".jbig2", ".jb2", ".svg"]
```
tarleb, line 225: I think this change could be problematic, as it would modify the behavior of `\includegraphics`. Maybe the list of default extensions could be passed in as a parameter?

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -1629,6 +1633,7 @@ blockCommands = M.fromList
    , ("lstinputlisting", inputListing)
    , ("inputminted", inputMinted)
    , ("graphicspath", graphicsPath)
+   , ("svgpath", graphicsPath)
```
tarleb, line 1636: If I understand correctly, then this would also modify the paths searched by `\includegraphics`, which is probably harmless but potentially undesired.

src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -1977,6 +1982,12 @@ graphicsPath = do
   getResourcePath >>= setResourcePath . (<> ps)
   return mempty
 
+svgSourceExtn :: Text -> Text
+svgSourceExtn src = if T.isSuffixOf (T.pack ".svg") src then
```
tarleb, line 1986: The idiomatic way to format `if`...`then`...`else` in Haskell is to put each of these key words on their own line, together with the corresponding expression:

```haskell
if test
then result1
else result2
```

tarleb (commented):
Looks good, but I'm worried about the interaction with `graphicx` commands.

jgm:
I have added support for the svg package in commit 1393037f4eb04ebad849d79b53a8b3e354aa0bdb

PR #7025, opened 2021-01-12
Add optional `csquotesoptions` to default LaTeX template.

bluebirch:


jgm:
Both this and `csquotes` should be documented under Variables for LaTeX in MANUAL.txt.

PR #7036, opened 2021-01-17
Lua filters: allow empty table as shorthand for empty caption

tarleb:


tarleb:
Test balloon for more flexible methods to define tables. Other ideas:

- Allow `nil` to be accepted as an empty caption.
- Accept a list of blocks as caption.
- Accept a list of blocks as simple `Cell`.

More advanced:

- Allow list of cells as `Row nullAttr`.
- Allow list of rows as `TableBody`.
- Allow `nil` to signal empty `TableHead`, `TableBody`.

Questionable:

- Accept `nil` as `nullAttr`.

jgm:
I think I'm missing some context here.

tarleb:
Sorry, I should have linked to the discussion that motivated this: https://groups.google.com/g/pandoc-discuss/c/gHSbJmb493E/m/LzwA1UMxDAAJ

Tables, as they are now, are still difficult to work with in Lua filters. This PR is a first attempt to improve the situation.

I'd like to note that I'm still trying to avoid adding any significant amount of Lua code until I get a real idea which marshalling strategy gives the best trade-offs between performance and maintainability. But captions as well as blocks as table cells seem small and important enough to bend that rule a little.

tarleb:
Closing this, as I believe that we can reach a better solution after #7191 is done. I also don't want to deal with the resulting merge conflicts.

PR #7042, opened 2021-01-20
JATS writer: use element citations

tarleb:


src/Text/Pandoc/Writers/JATS/References.hs
```diff
@@ -0,0 +1,125 @@
...
+      let yeartxt = tshow year
+      in inTags' "year" [("iso-8601-date", yeartxt)] $ literal yeartxt
+    _                      -> empty
+
+toNameElement :: Name -> Doc Text
+toNameElement name = inTags' "name" [] . mconcat $
+  [ maybe mempty (inTags' "surname"     [] . literal) (nameFamily name)
+  , maybe mempty (inTags' "given-names" [] . literal) (nameGiven name)
+  , maybe mempty (inTags' "suffix"      [] . literal) (nameSuffix name)
+  ]
```
jgm, line 120: In addition to "suffix" many names have "dropping-particle" and "non-dropping-particle."
If JATS doesn't have elements for this, I think you  want to add dropping-particle to the end of given-names and add non-dropping-particle to the beginning of surname (with spaces separating).

tarleb: JATS only has a [`<prefix>` element](https://jats.nlm.nih.gov/publishing/tag-library/1.2/element/prefix.html), but it's intended usage are titles and honorifics.

After some reading, I believe the correct method to use [`<name-alternatives>`](https://jats.nlm.nih.gov/publishing/tag-library/1.3d1/element/name-alternatives.html) with different versions for display and indexing. Then we can also add a `<string-name>` containing the nameLiteral field.

Cc @hubgit, who probably could offer deeper insights into proper, JATS4R conformant, tagging.

jgm: Note that nameLiteral is generally only used when the other fields are not, so that's not going to work.

I think the best thing to do is what I suggested:  when particles are not explicitly set off in their own fields, they are generally incorporated into given names (dropping) and surnames (nondropping).

jgm: Oh, and you'll also have to handle the case where nameLiteral has a value but the other fields don't.

tarleb: Ok, done.

I just noted that all my tests are local, will add them tomorrow.

tarleb:
Should be good now, I think.

src/Text/Pandoc/Extensions.hs
```diff
@@ -88,6 +88,7 @@ data Extension =
                                   --   does not affect readers/writers directly; it causes
                                   --   the eastAsianLineBreakFilter to be applied after
                                   --   parsing, in Text.Pandoc.App.convertWithOpts.
+    | Ext_element_citations   -- ^ Use element-citation elements for citations
```
jgm, line 91: Might be helpful here to say "JATS citations"

src/Text/Pandoc/Writers/JATS/References.hs
hubgit: There's not too much about names in [the JATS4R recommendations for citations](https://jats4r.org/citations):

> `<name>` / `<string-name>` / `<collab>`. The `<name> `or` <string-name>` element should be used for an individual named author. Use `<name>` when the ancestor element is `<element-citation>`, and `<string-name>` when the parent is `<mixed-citation>`.

> When groups of authors (consortia) are cited under a group name or an institution is cited as the contributor, the `<collab>` element should be used.

The [PMC tagging guidelines](https://www.ncbi.nlm.nih.gov/pmc/pmcdoc/tagging-guidelines/article/dobs.html#dob-refs) have some examples of citations where `<given-names>` or `<surname>` includes text that might be in the "dropping-particle"/"non-dropping-particle" data:

```
6. (a) De Albuquerque IL, Galeffi C, Casinovi CG, Marini-Bettolo GB. Gazz Chim Ital. 1964;94:287. (b) Alcantara, A. F. de C.; Souza, MR.; Pilo-Veloso, D. Fitoterapia. 2000;71:613.
```

```xml
<ref id="R6">
  <label>6</label>
  <element-citation publication-type="journal">
    <label>(a)</label>
    <name>
      <surname>De Albuquerque</surname>
      <given-names>I. L.</given-names>
    </name>
    <name>
      <surname>Galeffi</surname>
      <given-names>C.</given-names>
    </name>
    <name>
      <surname>Casinovi</surname>
      <given-names>C. G.</given-names>
    </name>
    <name>
      <surname>Marini-Bettolo</surname>
      <given-names>G. B.</given-names>
    </name>
    <source>Gazz. Chim. Ital.</source>
    <year>1964</year>
    <volume>94</volume>
    <fpage>287</fpage>
  </element-citation>
  <element-citation publication-type="journal">
    <label>(b)</label>
    <name>
      <surname>Alcantara</surname>
      <given-names>A. F. de C.</given-names>
    </name>
    <name>
      <surname>Souza</surname>
      <given-names>M. R.</given-names>
    </name>
    <name>
      <surname>Pilo-Veloso</surname>
      <given-names>D.</given-names>
    </name>
    <source>Fitoterapia</source>
    <year>2000</year>
    <volume>71</volume>
    <fpage>613</fpage>
  </element-citation>
</ref>
```

PR #7048, opened 2021-01-25
Lua: do not silently load files from user-defined userdir

tarleb:
A custom initialization file `init.lua` must be placed in the user's
default data directory, or the system-wide default directory. Built-in
Lua packages are only loaded from the system's default data directory.
Loading from a different directory by overriding the default path, e.g.
via `--data-dir`, is no longer supported for these scripts.


tarleb:
Superseded by #7051.

PR #7050, opened 2021-01-26
LaTeX writer: change BCP47 lang tag from jp to ja

### LINKED ISSUE #7047: Correct BCP47 tag in lang doesn't work but incorrect one does work when producing PDF from md with latex template
The section of [Language variables](https://pandoc.org/MANUAL.html#language-variables) states that the variable of `lang` key in the YAML section is based on BCP47. However, some correct tags do not work when producing a PDF file from a md file via pandoc with pandoc's [default.latex](https://github.com/jgm/pandoc/blob/3f9eee473dfa4eca7001cfe9ec57399de7a28e4a/data/templates/default.latex#L331) template.

Let's consider the tag of Japanese language as an example. Since its correct BCP47 tag is `ja`, I set `lang: ja` in the YAML section. However, this caused the following error hindering the document generation:

```bash
Error producing PDF.
! Argument of \str_uppercase:n has an extra }.
<inserted text>
                \par
l.64   \setotherlanguage[]{}
```

However, I replace `ja` with `jp` (Regional Code for Japan, so it's incorrect tag for the Japanese language, in a strict sense), I can get the PDF I want.

Would you have any idea about why this is happening, or would you mind adding some notes that some tags of BCP47 might not work and that regional tags might work for `lang` key to the MANUAL?

# MWE

## test-lang.md

    ---
    #lang: ja  ## Correct BCP47 tag for Japanese language, NOT WORKABLE for pandoc
    lang: jp   ## INCORRECT BCP47 tag for Japanese language, workable for pandoc
    
    ## Japanese font setting
    header-includes: |
      \usepackage{zxjatype}
      \setCJKmainfont{SourceHanSansJP-Medium.otf}
    ---
    
    こんにちは世界！
    
    ::: {lang=jp}
    > こんにちは世界！
    :::

## Bash command

```bash
pandoc test-lang.md --pdf-engine=xelatex -o test-lang.pdf --template=my/path/to/template.tex
```

mb21:
fixes #7047

or do we want to keep the existing (wrong) `jp` tag for backwards-compatibility? `jp` identifies the country, not the language... and we don't have countries in that list anywhere else... so...


jgm:
I think jp is just a mistake.  On the other hand, it's possible I suppose that somebody figured out that this works and uses it.  Maybe keep it in for now, but add a comment that it's in there for backwards compatibility?

jgm:
On second thought, seems better just to fix the mistake, and let people know about it in the changelog.
It should never have worked with 'jp'.

PR #7052, opened 2021-01-26
Lua: disable implicit execution of user-provided init file

tarleb:
Pandoc used to load the file `init.lua` from a user data directory
during Lua initialization. This feature has been removed, as it was in
conflict with pandoc's philosophy of only loading files that have been
passed explicitly. This also solves a slight security issue, as write
permissions to a user's data directory automatically resulted in the
power to run arbitrary Lua commands.


tarleb:
Not sure if we really want this, but if we do, then it's only a click away now.

tarleb:
Closing, but keeping the branch around just in case.

Issue #6674, opened 2020-09-10
Lua filter round-trip tests "double free or corruption" error


jgm:
We're seeing this in linux CI:
```
 Lua
3592
    Lua filters
3593
      inline elements can be round-tripped through the lua stack:    OK (0.10s)
3594
        +++ OK, passed 20 tests.
3595
      block elements can be round-tripped through the lua stack:     OK (0.14s)
3596
        +++ OK, passed 20 tests.
3597
      meta blocks can be round-tripped through the lua stack:        double free or corruption (!prev)
3598
Test suite test-pandoc: FAIL
```
https://github.com/jgm/pandoc/runs/1098511351?check_suite_focus=true

@tarleb any ideas?

jgm:
This may be related to the recent pandoc-types changes?

tarleb:
Yikes. This could be due to changes in HsLua 1.2: that version contains a number of low-level changes, including improvements to the way functions are exposed to Lua. This might take a while to debug; we should probably set the hslua < 1.2 constraint again.

tarleb:
OK, I have tightened the version constraints for hslua again.

The error is a mystery to me. I'm unable to reproduce it locally despite running the tests many times (same compiler version, same OS as far as I can tell).

tarleb:
Unfortunately, I can still provoke this error and similar failures by running the test suite with the number of QuickCheckTests to a large (four digit) value. Most failures are related to double-free issues, but there's the occasional round-trip failure without a crash. I assumed those validation failures would be repeatable by setting the quickcheck seed, but that not the case.

At least the error is rare, it occurs once every 100-500 CI runs by my estimate.

I'm reopening this and will have to dig deeper. I'll also loosen the hslua version constraints again, as that did not help.

jgm:
Here's another instance of it
https://github.com/jgm/pandoc/runs/1579669659?check_suite_focus=true#step:7:3909

tarleb:
So far I double-checked my memory allocating Haskell code, but didn't find anything. Guess I'll have to learn how to use Valgrind to really track this down.

tarleb:
Using Valgrind, I now can see some memory handling problems, but don't quite know how to proceed. My interpretation of the valgrind output is that there is something going wrong in Lua's memory allocator, but that would be very surprising.

```
      meta blocks can be round-tripped through the lua stack:     ==670637== Thread 5:                                                                                                                             
==670637== Invalid write of size 8                                                                                                                                                                                 
[...]

PR #7059, opened 2021-01-29
rST writer: add option for list tables, closes #4564

### LINKED ISSUE #4564: Option to use list-tables in rST writer.

danse:
this changes the API adding a global `--list-table` option which only
affects the rST writer, and causes all tables to be rendered using the
list table syntax documented here
http://docutils.sourceforge.net/docs/ref/rst/directives.html#list-table

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -952,6 +952,10 @@ header when requesting a document from a URL:
     placement of link references is affected by the
     `--reference-location` option.
 
+`--list-tables`
+
+:   In writing reStructuredText, render tables using the [list table syntax].
+
 `--reference-location=block`|`section`|`document`
 
 :   Specify whether footnotes (and references, if `reference-links` is
@@ -1292,6 +1296,7 @@ header when requesting a document from a URL:
     auxiliary files, use `--pdf-engine-opt=-outdir=foo`.
     Note that no check for duplicate options is done.
 
+[list table syntax]: https://docutils.sourceforge.io/docs/ref/rst/directives.html#list-table
 [Dublin Core elements]: https://www.dublincore.org/specifications/dublin-core/dces/
 [ISO 8601 format]: https://www.w3.org/TR/NOTE-datetime
 [Encoding issue with the listings package]:
--- a/src/Text/Pandoc/App/CommandLineOptions.hs
+++ b/src/Text/Pandoc/App/CommandLineOptions.hs
@@ -520,6 +520,11 @@ options =
                   (\opt -> return opt { optReferenceLinks = True } ))
                  "" -- "Use reference links in parsing HTML"
 
+    , Option "" ["list-tables"]
+                 (NoArg
+                  (\opt -> return opt { optListTables = True } ))
+                 "" -- "Use list tables in rendering rST"
+
     , Option "" ["reference-location"]
                  (ReqArg
                   (\arg opt -> do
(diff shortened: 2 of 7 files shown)


danse:
I'll check the linter errors

danse:
Followed the linter hints and (force) pushed

danse:
I get the following failure on Linux 8.8.4, to be honest it doesn't seem related to these changes

```
    defaults-inheritance-2.md
      #1:                                                            Error: Circular defaults file reference in 'command/defaults7.yaml'
```

danse:
opened #7061 to handle the build error

danse:
The error in continuous integration seems due to #6674

jgm:
Thanks for the update.  As noted in the issue, my main reservation about merging this is that it adds a new command-line option and alters WriterOptions (an API change that could  break existing code).  Also, people might want certain tables to be list and others not.  The ideal solution, it seems to me, is to make this triggered by an attribute on the table (and thus on a per-table basis). And we now do have table attributes.  The problem is that you still can't set attributes on tables in most formats (though this could always be done with a filter).

NickVolynkin:
> people might want certain tables to be list and others not

Theoretically, yes. But in many cases the team styleguide defines one preferred style for tables in rST. In my previous and current job it's the list-table.

Having two types of tables in a large documentation set is a likely situation, indeed. One has legacy code and so on. But within a single document most often there's one style.

Also, in my experience, a common case for *writing* rST is migrating a large set of documents from some other format to rST. In such case I would use list-table as the default setting and pick usual, ascii-art tables for special cases like cell spans.

>  The problem is that you still can't set attributes on tables in most formats (though this could always be done with a filter).

Writing filters is way more difficult than just using `pandoc` as a CLI application. A lot of users won't be able to write a filter.

tarleb:
I think we wouldn't lose anything if we went with the attribute-controlled solution first, as the command line switch could still be added later. @danse, do you want to update the PR, or would you prefer for somebody else to make those modifications?

theletterf:
Any update on this one? Nobody really uses grid tables for anything serious, AFAIK. :)

danse:
apologies for answering late on this one. I am not working at this project anymore so the change of scope has been a blocker for too long. As i wrote in #4564, i am willing to devote some time in the next future to merge these changes. Since @jgm's position seems to be quite firm, and we have no data about users' preferences, i will work towards the direction of using attributes. I hope to also include a sample script to set these attributes for all tables, so hopefully it will be easy enough for most users to switch to list tables

jgm:
My position isn't firm. I'm open to a global option that uses list tables for all tables in a document. I just don't know enough about how these are used to make the decision.

tarleb:
Side note: The ConTeXt writer uses the `ntb` extension to switch between "natural" and "extreme" tables. It might not be a good idea to add more extensions, but it could still make sense in this case.

danse:
> My position isn't firm. I'm open to a global option that uses list tables for all tables in a document. I just don't know enough about how these are used to make the decision.

Well, if we went for the attributes, there would be no way to produce list tables for some users, although i think that a Pandoc script would significantly help.

When i was working on automated conversion of documents we had guidelines about how to format `.rst`, so we considered list tables more convenient and just used them in all cases, using a fork.

Anyway i might be biased towards merging this as it is, and i agree that reducing global options is desirable.

> Side note: The ConTeXt writer uses the `ntb` extension to switch between "natural" and "extreme" tables. It might not be a good idea to add more extensions, but it could still make sense in this case.

I did not work much with extensions. If i understand correctly they pollute a global namespace as well. By looking at the manual i do not find advantages in using extensions rather than options.

danse:
Superseded by [an updated pull request](https://github.com/jgm/pandoc/pull/8320)

Issue #6199, opened 2020-03-19
EPUB Metadata support belongs-to-collection?


ivysrono:
I'm using pandoc 2.9.2
https://pandoc.org/MANUAL.html#creating-epubs-with-pandoc
https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-belongs-to-collection

It has been supported by Sigil and Calibre GUI, however, I do not find how to set it.

mb21:
I don't understand your question. Either way, for questions, please use the [pandoc-discuss mailing list](https://groups.google.com/d/forum/pandoc-discuss) instead of this issue tracker. Thanks!

jgm:
I think the request is to add `belongs-to-collection` to the list of EPUB metadata attributes supported by pandoc.

jgm:
However, I don't know whether this concerns pandoc's epub reader or epub writer. Perhaps @ivysrono can comment.

ivysrono:
> I think the request is to add `belongs-to-collection` to the list of EPUB metadata attributes supported by pandoc.

You are right.


> However, I don't know whether this concerns pandoc's epub reader or epub writer. Perhaps @ivysrono can comment.

I try to convert markdown to epub, so it may be epub writer?

brainchild0:
It seems reasonable to consider a mechanism for providing values  generically of the metadata fields, without any particular awareness of the specific field. The data might be given in YAML or directly in XML. The main drawback is  that without checking per-field validity, user misuse might lead to problems in the target.

jgm:
Closed by PR #7063.

PR #7063, opened 2021-01-31
ePub writer: `belongs-to-collection` metadata

Desbeers:
This is a pull-request to support the `belongs-to-collection` metadata attribute for ePubs.

See the request on https://github.com/jgm/pandoc/issues/6199

It adds two fields to the EPUB.hs writer:

- `belongs-to-collection`: The name of the collection.
- `group-position`: the number of the publication relative in the collection.

When a `belongs-to-collection` is found; `collection-type` = "series" will be added as well. I know that that can also be "set" but I never saw that in real-life.

The patch also reformats the other metadata a bit; that's only to keep the source-formatting nice :-)

The two new attributes are also added to the MANUAL.txt.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -5932,6 +5932,15 @@ The following fields are recognized:
 
 `rights`
   ~ A string value.
+  
+`belongs-to-collection`
+  ~ A string value.  identifies the name of a collection to which
+    the EPUB Publication belongs.
+
+`group-position`
+  ~ The `group-position` field indicates the numeric position in which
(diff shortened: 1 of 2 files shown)


jgm:
I'm baffled about the CI failure.  It is most likely an out of memory error in linking, but why we are getting that now, I don't know. I tried rerunning but to no avail.  @tarleb any ideas?

tarleb:
I think the problem is that this was opened when the main branch was failing. Closing and reopening to see if that fixes it.

jgm:
Thanks!

PR #7073, opened 2021-02-03
LaTeX template: Update to iftex package

adunning:
Load the iftex package directly rather than via the ifxetex and ifluatex compatibility wrappers, which have been merged into a single package that is part of the LaTeX core. The capitalization of the commands has been changed for compatibility with older versions of TeX Live that have the version of iftex by the Persian TeX Group. This had been removed in <https://github.com/jgm/pandoc/commit/2845794c0c31b2ef1f3e6a73bb5b109da4c74f37> for compatibility with BasicTeX, but that is no longer an issue.


jgm:
Thanks, good to know this!

PR #7079, opened 2021-02-05
DocBook reader: Support informalfigure

pyssling:
Add support for informalfigure.


jgm:
What about using `getFigure` here instead of `getBlocks`?
If the element doesn't contain a title, then it will do much the same thing (except that it will also have the id information available).  And when we add a Figure element in the future, we'll presumably want both of these things to produce one.

pyssling:
done.

jgm:
Thanks!

PR #7086, opened 2021-02-06
pandoc.cabal: use common stanza to reduce duplication

tarleb:


jgm:
It would be good to get this tested with cabal 2.2 now that we have that in CI.
Can you rebase on top of that change (which I just merged)?

jgm:
Good -- this is nice!

Issue #7077, opened 2021-02-04
Teplate parial fallback to data dir doesn't work with absolute path to template


aDotInTheVoid:
Version: pandoc-nightly-macos-2021-02-04

In directory `/tmp/pandoc_bug`

File README.md
```
text
```
template.html
```html
<style>
$styles.html()$
</style>

$body$
```

Running `./pandoc README.md --template template.html` gives the expected output (see below).

However `/pandoc README.md --template /tmp/pandoc_bug/template.html` gives this error
```
pandoc: /tmp/pandoc_bug/styles.html: openBinaryFile: does not exist (No such file or directory)
```
While this is true, it should fall back to finding `styles.html`  [in the templates subdirectory of the user data directory](https://pandoc.org/MANUAL.html#partials). Indeed this is what happens in the case of the first example

<details>

<summary> Expected output </summary>

```html
<style>
html {
  line-height: 1.5;
  font-family: Georgia, serif;
  font-size: 20px;
  color: #1a1a1a;
  background-color: #fdfdfd;
}
body {
  margin: 0 auto;
  max-width: 36em;
  padding-left: 50px;
  padding-right: 50px;
  padding-top: 50px;
  padding-bottom: 50px;
  hyphens: auto;
  word-wrap: break-word;
  text-rendering: optimizeLegibility;
  font-kerning: normal;
}
@media (max-width: 600px) {
  body {
    font-size: 0.9em;
    padding: 1em;
  }
}
@media print {
  body {
    background-color: transparent;
    color: black;
    font-size: 12pt;
  }
  p, h2, h3 {
    orphans: 3;
    widows: 3;
  }
  h2, h3, h4 {
    page-break-after: avoid;
  }
}
p {
  margin: 1em 0;
}
a {
  color: #1a1a1a;
}
a:visited {
  color: #1a1a1a;
}
img {
  max-width: 100%;
}
h1, h2, h3, h4, h5, h6 {
  margin-top: 1.4em;
}
h5, h6 {
  font-size: 1em;
  font-style: italic;
}
h6 {
  font-weight: normal;
}
ol, ul {
  padding-left: 1.7em;
  margin-top: 1em;
}
li > ol, li > ul {
  margin-top: 0;
}
blockquote {
  margin: 1em 0 1em 1.7em;
  padding-left: 1em;
  border-left: 2px solid #e6e6e6;
  color: #606060;
}
code {
[...]

PR #7088, opened 2021-02-07
 Document no template fallback for absolute path

aDotInTheVoid:
Reopening of https://github.com/jgm/pandoc-website/pull/44, but targeted at the correct website

See jgm/pandoc#7077


jgm:
Thanks

PR #7105, opened 2021-02-18
Add option to disable vertical content centering

### LINKED ISSUE #7104: RevealJS vertical center option not working
RevealJS allows to disable the automatic vertical centering of the slide content by using: `center: false` in  Reveal.initialize

Source:
https://revealjs.com/presentation-size/#center

This is not working with pandoc as there is no center variable which gets read.
Manually editing the HTML works

maurerle:
fixes #7104

### DIFF
--- a/data/templates/default.revealjs
+++ b/data/templates/default.revealjs
@@ -83,7 +83,7 @@ $endif$
       // https://revealjs.com/config/
       Reveal.initialize({
 $if(center)$
-        // Determines wether slide content should be vertically centered
+        // Determines whether slide content should be vertically centered
         center: $center$,
 $endif$      
 $if(controls)$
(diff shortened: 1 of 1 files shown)


data/templates/default.revealjs
```diff
@@ -82,6 +82,10 @@ $endif$
       // Full list of configuration options available at:
       // https://revealjs.com/config/
       Reveal.initialize({
+$if(center)$
+        // Determines wether slide content should be vertically centered
```
jgm, line 86: wether -> whether

jgm:
thanks!

maurerle:
This is flawed as center defaults to true and setting it to false does not trigger the if(center), therefore changes nothing.

This should check if center is defined.

jgm:
You're right -- please open a new issue for this so we can track it.
Note that the template language doesn't allow you to check definedness or distinguish between false and undefined.  We may need some special support for setting this variable in the writer.

PR #7108, opened 2021-02-19
Update default ODT style

Oni:
As of now, the default style for ODT documents has a "First paragraph" style that inherits from "Standard" style and has no top or bottom margin. All subsequent paragraphs have "Text_20_body" style that inherits from "Standard" and add "0.0598in" margins on top and bottom. This makes the final document a bit ugly since the first paragraph has a small gap ("0.0598in") towards the second one, and all subsequent have double that.

The proposed fix makes "First paragraph" inherit from "Text_20_body" instead so that it also has a consistent margin.

Another approach would be to inherit "Text_20_body" and add a 0 margin on top.


jgm:
That makes sense to me.

PR #7109, opened 2021-02-19
Update epub.css

Jmuccigr:
Correct "title" to "main" and add missing items (subtitle, publisher, and rights). Explanatory comments.

### DIFF
--- a/data/epub.css
+++ b/data/epub.css
@@ -7,14 +7,19 @@ h3 { text-align: left; }
 h4 { text-align: left; }
 h5 { text-align: left; }
 h6 { text-align: left; }
-/* For title, author, and date on the cover page */
-h1.title { }
+/* For (sub)title, author, date, and publisher on the cover page */
+h1.main { }
+h1.subtitle { font-size: smaller; }
 p.author { }
 p.date { }
+p.publisher { }
+div.rights { }
+/* For navigation items (e.g., TOC) */
 nav#toc ol,
 nav#landmarks ol { padding: 0; margin-left: 1em; }
 nav#toc ol li,
 nav#landmarks ol li { list-style-type: none; margin: 0; padding: 0; }
+/* Other */
 a.footnote-ref { vertical-align: super; }
 em, em em em, em em em em em { font-style: italic;}
 em em, em em em em { font-style: normal; }


data/epub.css
```diff
@@ -7,14 +7,19 @@ h3 { text-align: left; }
 h4 { text-align: left; }
 h5 { text-align: left; }
 h6 { text-align: left; }
-/* For title, author, and date on the cover page */
-h1.title { }
+/* For (sub)title, author, date, and publisher on the cover page */
+h1.main { }
```
jgm, line 11: If you look at the default template you'll see it has
```
$for(title)$
$if(title.text)$
  <h1 class="$title.type$">$title.text$</h1>
$else$
  <h1 class="title">$title$</h1>
$endif$
$endfor$
```
Thus, if they use a structured title (e.g. `title { text = ___, type = ___ }`) then "main" will be used (or maybe something else, depending on the value of "type").  If they just specify `title` as an unstructured field, `title` will be used.

Perhaps we could improve this by changing the first case to
```
  <h1 class="title $title.type$">$title.text$</h1>
```
so that the `title` class is consistently applied.
Anyway, I'd suggest we make that change to the template, but also leave out `main` from the default template.
(I'm actually not quite sure what motivated these structured titles -- would need to dig through git history.)

jgm, line 11: If you look at the default template you'll see it has
```
$for(title)$
$if(title.text)$
  <h1 class="$title.type$">$title.text$</h1>
$else$
  <h1 class="title">$title$</h1>
$endif$
$endfor$
```
Thus, if they use a structured title (e.g. `title { text = ___, type = ___ }`) then "main" will be used (or maybe something else, depending on the value of "type").  If they just specify `title` as an unstructured field, `title` will be used.

Perhaps we could improve this by changing the first case to
```
  <h1 class="title $title.type$">$title.text$</h1>
```
so that the `title` class is consistently applied.
Anyway, I'd suggest we make that change to the template, but also leave out `main` from the default template.
(I'm actually not quite sure what motivated these structured titles -- would need to dig through git history.)

Jmuccigr: I saw there was something going on in the template - which I don't really understand - but I'm going by what I get in the output epub3 document (title_page.xhtml) when I have "title" in my yaml:

```
<section epub:type="titlepage" class="titlepage">
  <h1 class="main">My Title</h1>
```

jgm: You must have a structured object for `title` in your YAML metadata, then.
```
---
title:
  type: main
  text: My Title
...
```

Jmuccigr: Ah, yes, I do. I was just copying the example in the Guide. If I switch to an unstructured title in yaml, I do indeed get h1 with a class of title (and if I keep subtitle in on its own, it gets a p).

Given that the Guide example gives an h1 with class main and that it does no harm being in the css if not used, would a solution be to have both main and title in there for h1, perhaps with a comment? I'm thinking of this css file as more of a template for the user to customize, so the comments are helpful.

For subtitle, it seems to go either with p or h1, depending on whether so just having it work for either would seem to be a solution:

p.subtitle, h1.subtitle { }

jgm: If we're going to have `h1.main`, then shouldn't we also have `h1.subtitle`, `p.ROLE` for various possible creator roles, et.c?

PR #7116, opened 2021-02-24
Include Header.Attr.attributes as XML attributes on section

lifeunleaded:
This implements the command line option --docbook-section-attrib which,
if used, adds key-value pairs found in the attributes list of Header.Attr as
XML attributes on the corresponding section element.

It checks that the key names are allowed according to DocBook 4 or 5
respectively. Any key name not allowed as an XML attribute name is
dropped, as are xml:id (for DocBook 5)/id (for DocBook 4) to not
intervene with computed identifiers.

The default is to not enable it, to avoid suddenly populating attributes
without changing anything else.

### DIFF
--- a/src/Text/Pandoc/Writers/Docbook.hs
+++ b/src/Text/Pandoc/Writers/Docbook.hs
@@ -168,7 +168,7 @@ blockToDocbook :: PandocMonad m => WriterOptions -> Block -> DB m (Doc Text)
 blockToDocbook _ Null = return empty
 -- Add ids to paragraphs in divs with ids - this is needed for
 -- pandoc-citeproc to get link anchors in bibliographies:
-blockToDocbook opts (Div (id',"section":_,_) (Header lvl _ ils : xs)) = do
+blockToDocbook opts (Div (id',"section":_,_) (Header lvl (_,_,attrs) ils : xs)) = do
   version <- ask
   -- Docbook doesn't allow sections with no content, so insert some if needed
   let bs = if null xs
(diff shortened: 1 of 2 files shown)


tarleb:
Hi @lifeunleaded, thank you for the PR and for your patience. The PR looks good to me, just two questions:

- I'm wondering whether the new option is needed, or if we can just always enable the new code. Do you think there would be downsides if we removed the option?
- Would it make sense to use `filter` with a predicate like `isSectionAttr :: DocBookVersion -> Text -> Bool`? The `mapMaybe` was a bit confusing to me.

lifeunleaded:
> Hi @lifeunleaded, thank you for the PR and for your patience. The PR looks good to me, just two questions:

Thank you for looking @tarleb, sorry it took a while to get back to this. 

> * I'm wondering whether the new option is needed, or if we can just always enable the new code. Do you think there would be downsides if we removed the option? The reason I'm asking is that option proliferation is a problem in pandoc, and it would be nice if we could avoid adding another.

For me, the option can be omitted. I added it mostly as a courtesy in case anyone has pandoc in an automated toolchain and experiences regression test failures when their DocBook output suddenly has attributes it didn't have. But I see your point, and option proliferation is probably a bigger problem, so I'll take out the option.

> * Would it make sense to use `filter` with a predicate like `isSectionAttr :: DocBookVersion -> Text -> Bool`? The `mapMaybe` was a bit confusing to me.

I'll try refactoring that. mapMaybe just felt natural as I wanted to return a list but with non-allowed items removed.

lifeunleaded:
@tarleb Done, I think. Much more elegant with the filter. I expanded the tests and attribute qualifications so at least valid docbook should be generated even if an existing toolchain has populated Header.Attr.attributes.

Since the command line option was the only user documentation I provided, should I be adding this somewhere else now?

tarleb:
Thank you, looks beautiful!

> Since the command line option was the only user documentation I provided, should I be adding this somewhere else now?

I think the new feature gives a very natural behavior, so the only thing we need to care about is the release changelog. The changlog is semi-automatically created from commits, the most helpful would be to squash the two commits and document the change in the commit message.

Tiny nitpick: could you restore the empty line in CommandLineOptions? Otherwise, commands like `git log .../CommandLineOptions.hs` will include this change, which can be a nuisance.

lifeunleaded:
Of course, the empty line was just an oversight, sorry about that. There'll be a forcepush with a squashed commit shortly.

lifeunleaded:
@tarleb Thank you for all the patient reviewing. I think I cleaned it up now.

tarleb:
Thanks!

PR #7121, opened 2021-02-26
Wrap url colours in braces in the default LaTeX template

LoicGrobol:
The current default LaTeX template writes the url colours without braces in the hypersetup block. This makes it impossible to use colour that contain commas (as in `linkcolor: "rgb:red,34;green,136;blue,51"` using xcolor extended colour expressions) because the comma clashes with the key=value parsing.

This PR fix this by simply adding braces around the colour names.


jgm:
Good improvement, thanks!

PR #7122, opened 2021-02-26
Fix/update URLs and use HTTP**S** where possible

salim-b:
I've changed a lot of `http:` URLs of which the respective web servers have properly configured HTTPS to their `https` equivalents.

Plus a bunch of URLs have simply been updated.

Cf. https://github.com/jgm/pandoc/pull/6090


jgm:
Thanks! I can fix up the tests.

PR #7123, opened 2021-03-04
Relaxes haddock-library upper bound

jkachmar:
Relates to https://github.com/commercialhaskell/stackage/issues/5890

I haven't tested this locally, but the `haddock-library-1.10.0` changelog makes me think that this should be fairly safe; I figured letting CI run the compile/test loop would be "good enough" but let me know if you think I should pull the changes down and run a sanity check.


jkachmar:
Whoops, well look like I'm wrong!

Looks like this fix will actually be somewhat involved 😞; I'll close this PR for the time being, and open another if I get the opportunity to try and actually patch things up appropriately.

PR #7131, opened 2021-03-09
HTML template CSS: add dark mode with --highlight-style-dark

### LINKED ISSUE #7125: HTML: add `prefers-color-scheme: dark` to the default css
Hello I would like to add `prefers-color-scheme: dark` media query to the default CSS generated when using this command `pandoc -s -o f.html f.md`, something like this:

```
@media (prefers-color-scheme: dark) {
  html {
    background-color: #1E1F21;
    color: #EEEFF1;
  }
  a {
    color: #EEEFF1;
  }
  a:visited {
    color: #EEEFF1;
  }
  blockquote p: {
    color: #606060;
  }
  hr {
    background-color: #EEEFF1;
  }
}
```

My rationale behind this enhancement is to make it easier to read the rendered HTML at night by default.

mb21:
closes #7125

these values take precedence over values set by custom variables

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -837,6 +837,12 @@ header when requesting a document from a URL:
     To generate the JSON version of an existing style,
     use `--print-highlight-style`.
 
+`--highlight-style-dark=`*STYLE*|*FILE*
+
+:   Specifies the coloring style to be used in highlighted source code
+    in [dark mode] for HTML-based output formats. The default is `breezeDark`.
+    See `--highlight-style` for more information.
+
 `--print-highlight-style=`*STYLE*|*FILE*
 
 :   Prints a JSON version of a highlighting style, which can
@@ -6202,6 +6208,10 @@ scheme used by the Python library pygments (though pygments is not actually
 used to do the highlighting).  To see a list of highlight styles,
 type `pandoc --list-highlight-styles`.
 
+For HTML-based output formats, the color scheme used for [dark mode]
+can be selected using the `--highlight-style-dark` option.
+It defaults to `breezeDark`.
+
 If you are not satisfied with the predefined styles, you can
 use `--print-highlight-style` to generate a JSON `.theme` file which
 can be modified and used as the argument to `--highlight-style`. To
@@ -6223,6 +6233,7 @@ definitions](https://github.com/KDE/syntax-highlighting/tree/master/data/syntax)
 To disable highlighting, use the `--no-highlight` option.
 
 [skylighting]: https://github.com/jgm/skylighting
+[dark mode]: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
 
 # Custom Styles
 
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -177,3 +177,25 @@ div.csl-indent {
   margin-left: 2em;
 }
 $endif$
+@media screen and (prefers-color-scheme: dark) {
+  html {
+    color: #d3d3d3;
+    background-color: #101010;
+  }
+  a, a:visited {
+    color: #d3d3d3;
+  }
+  blockquote {
+    border-color: #707070;
+    color: #d3d3d3;
+  }
+  pre, code {
+    background-color: #1b1b1b;
+  }
+  hr {
+    background-color: #d3d3d3;
+  }
+  tbody, th {
+    border-color: #d3d3d3;
+  }
+}
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -177,6 +177,7 @@ div.csl-indent {
   margin-left: 2em;
 }
 $endif$
+$if(document-css)$
 @media screen and (prefers-color-scheme: dark) {
   html {
     color: #d3d3d3;
@@ -202,4 +203,5 @@ $endif$
   $highlighting-css-dark$
   $endif$
 }
+$endif$
 
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -189,15 +189,14 @@ $endif$
     border-color: #707070;
     color: #d3d3d3;
   }
+  pre, code, div.sourceCode {
+    background-color: #1b1b1b;
+  }
   hr {
     background-color: #d3d3d3;
   }
   tbody, th {
     border-color: #d3d3d3;
   }
-  pre, code, div.sourceCode {
-    background-color: #1b1b1b;
-  }
-  code span.fu { color: #093bb9; }
-  code span.va { color: #2725c0; }
 }
+
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -189,14 +189,17 @@ $endif$
     border-color: #707070;
     color: #d3d3d3;
   }
-  pre, code, div.sourceCode {
-    background-color: #1b1b1b;
-  }
   hr {
     background-color: #d3d3d3;
   }
   tbody, th {
     border-color: #d3d3d3;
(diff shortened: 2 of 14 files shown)


jgm:
How does this interact with syntax highlighting?

mb21:
@jgm good point! some of the darker colors of the default syntax highlighting theme are a bit hard to read:

image

but in my opinion not really harder than the output of `--highlight-style=breezeDark`...

we could of course lighten up the background-color of `div .sourceCode` in dark mode...

mb21:
I see three solutions to this:

1. lighten up `.sourceCode` background in dark mode (the question is to what? something that's almost white or something that's almost black)
2. inject the CSS supporting dark-mode only if there are no code blocks with syntax highlighting in the document (could be overriden by a variable set by the user)
3. Figure out a way to load a completely different [skylighting](https://github.com/jgm/skylighting) theme in dark mode

mb21:
Okay, I've lightened up the background and overwrite the two darkest colors we have in the default pygments theme:

image

mb21:
ah, but overwriting those two values would also overwrite them when the user specifies another theme... so reverted that part while keeping the slightly lighter background.

image

if we're unhappy with this, we could add a `highlighting-css-dark` variable [around here](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/HTML.hs#L322) and populate it with, say, `styleToCss "breezeDark"`.

jgm:
We can't assume anything about the highlighting theme they're using -- it could even be a custom theme.

mb21:
True. But what about the reverse? Should the theme assume a light background and dark font? e.g. `pandoc --print-highlight-style pygments` gives:

```
    "text-color": null,
    "background-color": null,
```

mb21:
Anyway, I've added a `highlight-style-dark` option, which seems like the best solution anyway.

jgm:
I have mixed feelings about this.  In general, I'm resistant to proliferating command-line options.  If we do add `-highlighting-theme-dark`, though, it would also need to be supported via defaults files (I didn't see that above but maybe I missed it).

I'm still wondering whether there might be an alternative approach which derives the "dark" highlighting colors from the regular ones (inverting or something -- not sure whether there's a transformation that would give good results).  If so, we could just have a function that maps a highlighting theme on to a "dark-mode" version.  And then everything would work transparently without the need for a new option, and without the user needing to think of that.

mb21:
> it would also need to be supported via defaults files

ah yes, I assumed this would somehow get auto-generated from the command-line args... but I could add it...

---

Some more screenshots...

Using `breezeDark` (i.e. what's currently on this branch):

image

---

Doing a simple invert:

```css
div.sourceCode {
  filter: invert(1);
}
pre.sourceCode {
  /* since pygments theme doesn't set a font color, we need to set it so that we can invert it */
  color: black;
}
```

image

---

> I'm resistant to proliferating command-line options.

yes, I understand that. And at first I was hesitant as well, but in the end it seems like the "proper" solution? If the user can set the `--highlight-theme`, there should also be a way to set the theme for dark mode? But I don't feel strongly either way...

jgm:
I actually don't mind the "inverted" version in your screen shot.  Is it objectionable?

(It would be worth seeing what the inverted versions of all the highlight styles look like; I assume this is just the inverted default style.)

mb21:
>  Is it objectionable?

I don't know.. the aggressive pink on black background certainly has a different feel to it than the default pygments theme in light mode... but I don't mind :)

But to make the inverting really a full solution, we'd have to change the output of `pandoc --print-highlight-style pygments` to contain `"text-color": "black",` instead of `"text-color": null`. Otherwise, the text inherits the CSS from darkmode on the html body (white), which is then invered to black, and you don't see it.

jgm:
As a color-blind person, I'm probably not the best one to judge such things...

jgm:
It might not be desirable to set the text-color to black.  As things are, text-color for code in the default style will just follow your document's text-color.  If you set that to something else, then the black code might look bad.

Is there a way to do this in the CSS itself -- set the text-color in the Div to whatever the document's text-color is?

jgm:
I wonder if invert(0.7) or something like that might lead to more aesthetically pleasing results?

What I like better about the inverted version is that it does not change the fact that the code block's background color is not different from the document's.

mb21:
> Is there a way to do this in the CSS itself -- set the text-color in the Div to whatever the document's text-color is?

probably... but we'd only want that when the pygments theme is active... (or any theme that doesn't set the color)... so we'd still need a template variable for that, which is a hack as well...

> What I like better about the inverted version is that it does not change the fact that the code block's background color is not different from the document's.

yeah, although arguably if we want that, a theme should set `background-color: null` (like the pygments does, just likes it sets `text-color: null`)

jgm:
> yeah, although arguably if we want that, a theme should set background-color: null (like the pygments does, just likes it sets text-color: null)

Yes, exactly.  I guess what I mean is this:  I like the fact that the dark-mode and light-mode versions agree with each other in this respect.

mb21:
Ah, by not setting background-color on pre, and code at all, I actually got the invert approach working: https://github.com/jgm/pandoc/pull/7226

mb21:
Update: that color-inverting approach (in the other PR) works okay with light themes (like the default pygments), but looks quite horrible in dark mode when applying a dark theme e.g. `--highlight-style=breezeDark`. So it's either

- we set a variable in the writer

      let isDarkTheme = selectedTheme `elem` ["breezeDark", "espresso", "zenburn"]

  and only apply the inverting-CSS when that variable is true, or:

- add the `--highlight-style-dark` option (set to an appropriate default, so the user doesn't have to worry about it)

jgm:
I think checking for particular theme names is not the thing to do, since they might be using a custom theme.
To do this right we'd have to inspect the theme itself and determine from the colors used whether it would look better inverted or non-inverted.

mb21:
> With the `--highlight-style-dark` approach, users who use dark mode will have bad results until they figure out that they need to specify this option.

Not sure I understand this part... if the the option is not specified, it would default to a dark theme (`breezeDark` in this PR), to give good results when the output document is viewed in dark mode.

jgm:
Could we set `isLightTheme` automatically by checking backgroundColor on the style?  If it's > 0x8000 we count it as a light theme?

mb21:
> Could we set `isLightTheme` automatically by checking backgroundColor on the style?

Probably... although I'm not sure where and why we would use that boolean variable then?

mb21:
> Could we set `isLightTheme` automatically by checking backgroundColor on the style?

To determine whether we should inject the inverting CSS? Yeah, probably... we'd have to read out the style file though (also if it's a user-supplied style file, which I don't think we have in `writerHighlightStyle`) It seems like a little bit too much magic for my tastes ;-)

jgm:
All you have to do is `backgroundColor <$> writerHighlightStyle opts` and it gives you a `Maybe Color` (where `Color = RGB Word8 Word8 Word8`.
So we'd just need to write a function `isLight :: Color -> Bool`.
(I noticed you closed the inverting PR -- that's the one this suggestion relates to, and I think it's still a good option.)

mb21:
I pushed a commit to the other PR, but `join (backgroundColor <$> writerHighlightStyle opts)` seems to be always `Nothing` for some reason... will try to investigate if I have time..

mb21:
Okay, I've add a commit with that... seems to work. Even if it means that people have to use a custom template to set a different theme for dark mode. But I guess it's and edge case anyway and we'll get feedback one way or another...

PR #7139, opened 2021-03-11
Update current-date.lua filter

Ddedalus:
The correct placeholder for day is `%d`, not `%e`, see [Lua docs](https://www.lua.org/pil/22.1.html)

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -390,7 +390,7 @@ current date, if a date isn't already set:
 ``` lua
 function Meta(m)
   if m.date == nil then
-    m.date = os.date("%B %e, %Y")
+    m.date = os.date("%B %d, %Y")
     return m
   end
 end


tarleb:
Thanks, using `%d` makes more sense.

But note that the `%e` is actually valid, as the [`os.date` doc](https://www.lua.org/manual/5.3/manual.html#pdf-os.date) confirms:

> If `format` is not \"`*t`\", then `date` returns the date as a string, formatted according to the same rules as the ISO C function `strftime`.

And `%e` is documented in `strftime` with

> Same as %d, except single digit is preceded by a space [1-31].

tarleb:
I'm going to cherry-pick the commit to adjust the commit message.

tarleb:
On second thought, using `%d` is not equivalent in our case. With `%d` we get `01`, whereas `%e` gives ` 1`.

Regardless, thanks for opening the PR.

PR #7148, opened 2021-03-14
RFC: Use custom Prelude based on relude.

jgm:
The Prelude now longer exports partial functions, so
a large number of uses of these functions in the
code base have been rewritten.

A .ghci file has been added; this is necessary for
ghci to work properly with the custom Prelude.

Currently there are lots of compiler warnings.
We should either fix these or go to using a custom
Prelude that changes less than relude.


jgm:
@tarleb This was kind of an experiment, and I'm still not sure what to think.

There would be some advantages and some disadvantages.  Another possibility would be a completely home-grown custom prelude.  But I thought I'd see if you and others who spend time with the pandoc code base had comments before continuing to work on this branch. (It needs a lot of code cleanup to remove the compiler warnings, but this branch does build, and the tests and benchmarks pass.)

tarleb:
I mostly like this, and I had even been pondering to do the same thing. Relude seem strike a good balance between usability and safety, without going overboard with new ideas.

On the other hand, there are *a lot* more changes than I expected, which has me skeptical. I'm also not sure if this would make it easier or more difficult for new contributors to get started. Relude is a lot saner in many aspects, but it also adds another layer with which one has to become familiar.

There is a [long discussion](https://discourse.haskell.org/t/towards-a-prevalent-alternative-prelude/1986) about blessing an "official" alternative prelude. Relude has come up there as well. Maybe we should wait and see if this reaches a conclusion before continuing?

tarleb:
Of course, if the idea is to shape this discussion, then choosing a prelude for pandoc would certainly add lot of weight. Relude is my favorite existing alternative, so I wouldn't mind. But I'm also dreading the frustration of having to go back and forth between pandoc and libraries for which it would be overkill.

As a side note: there seems to be a very unfortunate non-technical [disagreement](https://twitter.com/vrom911/status/1370473947783905282) between relude maintainers and people at the Haskell Foundation. So the chances of relude being blessed are rather slim, IMHO. But I'd rather not be part of these politics.

jgm:
I have some of the same concerns.  Once the good base became entangled with relude's assumptions, it wouldn't be easy to extricate it.

Of course, an alternative to adopting relude would be to create our own custom prelude, that departs less from the official prelude but (like relude) bans partial functions and provides easy alternatives.  A minimalistic version of this would be to hide head, tail, init, last, minimum, maximum, minimumBy, maximumBy, read, and include all the functions defined in Safe.  Or we could include NonEmpty, as relude does, and some functions to work with that.

In addition to the banishment of partial functions, another thing I like about relude is that `Debug.Trace` functions are always available, but attached to a compiler warning.  IT would be nice not to have to add the `import Debug.Trace` every time you want to trace something.

If we wanted to, we could also have the custom prelude export functions from Data.Maybe, Data.Either, Control.Monad, etc., that are tedious to import manually.

An alternative that would keep the standard Prelude would be to adopt hlint rules that get triggered by partial functions and wire them into a CI check and git commit hook.

jgm:
Here's a list (possibly not complete) of modules that currently use partial functions:
```
src/Text/Pandoc/Readers/Markdown.hs
src/Text/Pandoc/Readers/HTML.hs
src/Text/Pandoc/Readers/Docx.hs
src/Text/Pandoc/Readers/DocBook.hs
src/Text/Pandoc/Readers/Docx/Lists.hs
src/Text/Pandoc/Readers/Textile.hs
src/Text/Pandoc/Readers/RST.hs
src/Text/Pandoc/Readers/Haddock.hs
src/Text/Pandoc/Readers/Txt2Tags.hs
src/Text/Pandoc/Readers/Vimwiki.hs
src/Text/Pandoc/Readers/LaTeX/SIunitx.hs
src/Text/Pandoc/Readers/LaTeX/Citation.hs
src/Text/Pandoc/Readers/HTML/Table.hs
src/Text/Pandoc/Citeproc/BibTeX.hs
src/Text/Pandoc/Readers/Org/Blocks.hs
src/Text/Pandoc/Citeproc.hs
src/Text/Pandoc/Parsing.hs
src/Text/Pandoc/Readers/Odt/Generic/XMLConverter.hs
src/Text/Pandoc/Shared.hs
src/Text/Pandoc/Writers/ZimWiki.hs
src/Text/Pandoc/Writers/Markdown.hs
src/Text/Pandoc/Writers/AnnotatedTable.hs
src/Text/Pandoc/Writers/RTF.hs
src/Text/Pandoc/Writers/ConTeXt.hs
src/Text/Pandoc/Writers/Textile.hs
src/Text/Pandoc/Writers/MediaWiki.hs
src/Text/Pandoc/Writers/Ms.hs
src/Text/Pandoc/Writers/Docbook.hs
src/Text/Pandoc/Writers/Powerpoint/Presentation.hs
src/Text/Pandoc/Writers/Man.hs
src/Text/Pandoc/Writers/HTML.hs
src/Text/Pandoc/Writers/FB2.hs
src/Text/Pandoc/Writers/Org.hs
src/Text/Pandoc/Writers/Muse.hs
src/Text/Pandoc/Writers/Powerpoint/Output.hs
src/Text/Pandoc/Writers/AsciiDoc.hs
src/Text/Pandoc/Writers/DokuWiki.hs
src/Text/Pandoc/Writers/LaTeX.hs
src/Text/Pandoc/Writers/JATS/Table.hs
src/Text/Pandoc/Writers/RST.hs
src/Text/Pandoc/Writers/Shared.hs
src/Text/Pandoc/Writers/LaTeX/Table.hs
src/Text/Pandoc/Writers/LaTeX/Citation.hs
src/Text/Pandoc/Writers/Markdown/Inline.hs
```

tarleb:
I like the idea of having a custom prelude. Getting rid of dangerous functions and not having to import `fromMaybe` yet again would be wonderful.

There's this Haskell adage that writing a monad tutorial and creating a new alternative prelude are rites of passage. I'm behind on the tutorial, but maybe I can skip ahead to the prelude stage.

jgm:
The one drawback I see is that it might be a little hard to find the documentation for our custom prelude. (It is an internal module and wouldn't appear in the API haddocks.)

mb21:
Just a thought: could also have a custom prelude that just re-exports (a subset of) relude. Not sure this is a good idea though...

tarleb:
I just thought of a downside of a custom prelude: it will make it impossible to split the code into multiple packages, unless the prelude gets published as its own package.

jgm:
That's a good point. I suppose the prelude could be published as a module of the core package, though, and used in the others.  But maybe this is all more trouble than it's worth.  (Or we could go back to considering using one of the already available preludes like relude...)

PR #7150, opened 2021-03-16
RFC: Add bespoke prelude

jgm:
This is a draft for comment.  It adds a custom prelude without depending on relude.
It is much more minimal, at the moment, than relude.
It adds warnings to the the commonly used unsafe functions, without removing them, and provides safe alternatives, as well as some convenient modules and definitions in scope.
(Some functions are currently left out, like foldr1.)

Eventually we might want to replace head, last, etc. with alternatives that work on NonEmpty, but this will require more changes in pandoc.


tarleb:
I'm torn. As you said, we'd buy safety and convenience at the cost of additional code and reduced documentation.

I think I'm still in favor; we could also publish the prelude as a package on Hackage, but that would add more overhead.

jgm:
I'd rather not publish yet another prelude on Hackage.
We could always generate API docs and upload them to the pandoc website on each release.
And maybe, if the prelude weren't too extensive, it wouldn't need much documentation.  You'd learn that `mzero` is included when you tried to import it and got a compiler warning.  Similarly, for things like `head` (if we switched to using NonEmpty) we could show a warning with detailed instructions, as relude does.

PR #7153, opened 2021-03-16
CONTRIBUTING: suggest using a `cabal.project.local` file

tarleb:


jgm:
Probably going to change it in cabal.project.

jgm:
Odd.  I've tried changing my default stack install to use `-j4`, and it makes no difference at all to compile time.

PR #7156, opened 2021-03-18
Update bounds for random

lehins:
This PR updates bounds for random-1.2 which was released about a year ago.

If possible a new revision on hackage would be awesome, since it is blocking commercialhaskell/stackage#5474


jgm:
I am waiting for QuickCheck to make this change, since there's no way for me to test it until QuickCheck will accept this version of random.

lehins:
@jgm QuickCheck does support random-1.2 starting with version QuickCheck-2.14.1

jgm:
Okay, thanks.  Don't know why I thought it hadn't.
I can see some tests need updating (random UIDs).

PR #7171, opened 2021-03-22
Support toc-title in revealjs

### LINKED ISSUE #7170: Support toc-title in revealjs
Among other HTML formats, the HTML5 template includes the content of `toc-title` as heading:

https://github.com/jgm/pandoc/blob/99e1b67b74ddcbf7ffc8add640493f73471cf50a/data/templates/default.html5#L54-L61

Is there any reason against doing something similar for `revealjs`? Otherwise there's only an untitled slide containing the TOC, which doesn't look appealing.

fkohrt:
Fixes #7170

### DIFF
--- a/data/templates/default.revealjs
+++ b/data/templates/default.revealjs
@@ -59,6 +59,9 @@ $endif$
 $endif$
 $if(toc)$
 <section id="$idprefix$TOC">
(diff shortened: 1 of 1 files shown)


jgm:
Great, thanks!

Issue #7159, opened 2021-03-18
Typo in the Pandoc Lua filters documentation


Polirecyliente:
In the Pandoc Lua filters documentation available at https://pandoc.org/lua-filters.html, it says:

> "The return of a filter function must one of the following:"

Which is wrong, it should read:

> "The return of a filter function must be one of the following:"

The missing verb is "be", which is necessary in that place to say that the following 3 returns of a filter function, are the valid ones.

jgm:
Thanks. If you want, you can easily create a pull request by editing
https://github.com/jgm/pandoc/blob/master/doc/lua-filters.md
(click the edit button, then make your change, and there's a convenient button at the bottom to create a new branch for a PR).

Polirecyliente:
Thanks to @TatianaPorras aka La Tats xD ... next time I'll just send the Pull Request directly.

PR #7179, opened 2021-03-28
Fixed typo #7159

TatianaPorras:
This commit fixes the typo in the Lua filters documentation, see #7159

The Lua filters documentation (file https://github.com/jgm/pandoc/blob/master/doc/lua-filters.md) had a grammatical error, it said "The return of a filter function must one of the following:", which lacks the verb "be", so the correct phrase is "The return of a filter function must be one of the following:".


tarleb:
Thanks.

PR #7191, opened 2021-03-31
Update to HsLua 2.0

tarleb:
This updates pandoc to use the improved hslua framework. Both this PR and hslua are work in progress.

Filter performance improves by about 5%, but the planned marshalling modifications will change this again.

Suggestions for further improvements to hslua are welcome.


tarleb:
The goals of this refactoring are:

- better error messages when retrieving elements from Lua fails;
- better error messages for failing function calls;
- improved maintainability, esp. for modules and exposed Haskell functions;
- improve readability by writing more explicit code, use less typeclass magic (this one is debatable, as it slightly increases code verbosity);
- placing function definitions and documentation side-by-side; generate documentation;
- online documentation.

tarleb:
Progress so far: everything works as before, just faster. Especially unmarshalling has become more efficient.

Some very unscientific micro-benchmarks:

| filter                             | filter runtime change |
|------------------------------------|-------------------------|
| function Para(_) return nil end    | ±0%                     |
| function Str(s) return s end       | -13%                    |
| function Pandoc(d) return d end    | -15%                    |
| function Str(s) return {s} end     | -30%                    |

Preliminary tests with the alternative marshalling method show that both marshalling and unmarshalling will be even faster, but creation of new elements and the attribute access will be slower. It's probably going to about even-out in most filters.

tarleb:
More progress: the new marshalling method seems to work and has comparable, if not even better, performance. Element properties are pushed to Lua on demand, so a filter like

``` lua
function Pandoc (doc)
  print(doc.meta)
  return doc
end
```

takes only around 1% of the old time when run over the manual.

Error handling is better as well. A filter like

``` lua
function Str (_)
  return pandoc.Str(true)
end
```

currently runs without error, while the new version gives

```
string expected, got boolean
        while retrieving function argument string
        while retrieving arguments for function Str
stack traceback:
        [C]: in field 'Str'
        ../testing/string-error.lua:2: in function 'Str'
```

There is still room for improvement, and are some details are still left to be figured out. But all in all I'm confident that the new approach will work, and that it will improve the situation for maintainers and users.

tarleb:
Explaining are some of the trade-offs:

- Currently all AST elements are represented in Lua using just tables and strings; the new method uses Lua "userdata" values, storing pointers to the Haskell values.
- To get mutable attributes, we must still create Lua tables and associate them with the userdata, as otherwise a Lua statement like `div.content[1] = para` would only modify the content list, but not the `div` object. We do this by using Lua's "uservalue" feature, which allow to attach a "caching" table associated to userdata objects. This is done on demand, so we have lazy loading for all element properties.
- Jumping between Haskell and Lua is slow, so I had to implement some parts in C. However, these snippets are part of the hslua package and should be unnoticeable.
- Element constructors like `pandoc.Str` create new values in Haskell, so arguments are checked right away, making it easier to track down problems. The new constructors are slightly slower because of extra work, but the effect is far less than I feared.
- It's still possible to set element properties to invalid values. For example, running `str.text = true` will only cause an error if the `str` value is unmarshalled back to Haskell. Catching this early is difficult and expensive.

What's left to do:

- HsLua's handling of sum-types is not very elegant yet.
- Better handling of lists; ideally, lists would be lazy as well.
- Various fine-tuning.

jgm:
That all sounds good so far!

jgm:
@tarleb what is your time line on this?  I'm contemplating a 2.14 release at some point soon.
Also, does this involve API changes?

tarleb:
I set myself a deadline for mid July to release hslua-2.0. This PR could be merged once the release is out, it would already lead to some improvements. I'd like to have the remaining refactoring done by September. Progress is slow right now, and I expect to need the full time.

There are some remaining changes to the docx writer that I'd like to get into the 2.14 release -- I found that I broke some things in tables and need to fix that. I should be done within the next few days.

tarleb:
I try to avoid behavior changes, but here is one where a different behavior probably makes sense:
When assigning HTML-like attributes to an element, we currently convert them to `Attr` before assigning. This gives the following behavior:

```lua
local span = pandoc.Span 'test'
local attribs = {id = "foo", width = "27"}
span.attr =  attribs
assert(span.attr.attributes.width == 27)
assert(span.attr.id == nil)
```

I think this is confusing, so the current draft converts only when unmarshalling the structure back to Haskell.

```lua
local span = pandoc.Span 'test'
local attribs = {id = "foo", width = "27"}
span.attr =  attribs
assert(span.attr.attributes.width == 27) -- error: `attributes` is not a table.
assert(span.attr.id == "foo")
```

It's a very special corner-case. and I wouldn't expect anyone to notice the difference anyway.

jgm:
Good news.  I'm all in favor of asymptotic convergence to perfection!

Question: do the module changes mean we won't be using e.g. `pandoc.RawBlock()`?  Will anything change in the user-facing interface for Lua filters?  Will existing filters need to be changed, or is it backwards compatible?

jgm:
Anyway, let me know when ready to merge!

PR #7196, opened 2021-04-01
Org writer: Use LaTeX style maths deliminators

tecosaur:
Like Markdown, Org supports LaTeX fragments. However, while many Markdown flavours like the TeX-style `$` / `$$`, as mentioned in the Org manual [1], Org works better with LaTeX-style `\( \)` / `\[ \]` delimiters. It also supports LaTeX environments (`\begin{*} ... \end{*}`) without any form of escaping, but that is best left to another patch.

\[1\]: https://orgmode.org/manual/LaTeX-fragments.html

### DIFF
--- a/src/Text/Pandoc/Writers/Org.hs
+++ b/src/Text/Pandoc/Writers/Org.hs
@@ -404,8 +404,8 @@ inlineToOrg (Str str) = return . literal $ escapeString str
 inlineToOrg (Math t str) = do
   modify $ \st -> st{ stHasMath = True }
   return $ if t == InlineMath
-              then "$" <> literal str <> "$"
-              else "$$" <> literal str <> "$$"
+              then "\\(" <> literal str <> "\\)"
+              else "\\[" <> literal str <> "\\]"
 inlineToOrg il@(RawInline f str)
   | isRawFormat f = return $ literal str
   | otherwise     = do
--- a/test/writer.org
+++ b/test/writer.org
@@ -600,14 +600,14 @@ Ellipses...and...and....
   :END:
 
 - \cite[22-23]{smith.1899}
-- $2+2=4$
-- $x \in y$
-- $\alpha \wedge \omega$
-- $223$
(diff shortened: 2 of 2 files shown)


tarleb:
I don't understand. To me it seems that this PR removes support for
TeX-style math (which *is* supported by at least the org HTML exporter),
and duplicates support for LaTeX-style math. Can you explain more?

tecosaur:
As it is currently, pandoc will export LaTeX maths snippets in a format that Org does not recognise as maths. The cause of this is mentioned in this point of that manual page:

> To avoid conflicts with currency specifications, single ‘$’ characters are only recognized as math delimiters if the enclosed text contains at most two line breaks, is directly attached to the ‘$’ characters with no whitespace in between, and if the closing ‘$’ is followed by whitespace, punctuation or a dash. For the other delimiters, there is no such restriction, so when in doubt, use ‘\(...\)’ as inline math delimiters.

So, I can create a simple markdown file like
```
Say I have some $a$s and some $b$s. They are recognised as inline maths.
```

Clearly, these are parsed as maths elements:
```
~$ pandoc -f markdown -t latex inline-maths.md
Say I have some \(a\)s and some \(b\)s. They are recognised as inline maths.
```

Now, if I convert to Org
```
~$ pandoc -f markdown -t org inline-maths.md
Say I have some $a$s and some $b$s. They are recognised as inline maths.
```

Except, these are not maths elements in Org, due to the requirements quoted above, specifically
- "the closing ‘$’ is followed by whitespace, punctuation or a dash"

The form that Org better supports, and that acts as pandoc expects, is LaTeX-style delimiters.

tarleb:
Thank you, I understand now. That was the information I was missing.

We need to fix the tests before we can merge, could you take a look?

tecosaur:
Glad I was able to clear that up :slightly_smiling_face: I've updated the test suite to match, hopefully this all looks good to you now.

Issue #5722, opened 2019-09-01
Handle new pandoc-citeproc locator syntax with --biblatex option


seifferth:
The new locator syntax introduced in pandoc-citeproc <https://github.com/jgm/pandoc-citeproc/pull/362> does not work with the `--biblatex` switch in pandoc 2.7.3 (full version below).

Input:

```
[@citekey{xii}]
```

output of `pandoc -t latex --biblatex`

```latex
\autocite[xii\}]{citekey}
```

expected output:

```latex
\autocite[xii]{citekey}
```

Full pandoc version information: pandoc 2.7.3, Compiled with pandoc-types 1.17.6, texmath 0.11.2.2, skylighting 0.8.2

jgm:
The `{..}` locator syntax is part of pandoc-citeproc; pandoc itself just has the notion of a citation suffix, and all the parsing of this into locator + other stuff is done in pandoc-citeproc.  Since pandoc-citeproc doesn't get involved when you use `--biblatex`, the parsing isn't done.

We could move the code for doing this parsing to pandoc, and call it from pandoc-citeproc, I suppose.  But then we'd need code to combine a locator + suffix into a single suffix, for biblatex.  A simpler but probably adequate approach might be to remove the curly braces.

seifferth:
I understand that the special locator syntax is specific to pandoc-citeproc. Yet ideally, one would be able to use the same source to get adequate results using either pandoc-citeproc or biblatex. I believe simply removing the curly braces is a reasonable approach, since biblatex has it's own heuristics for parsing locators anyway.

PR #7220, opened 2021-04-11
Translate citation locator braces to \pnfmt when using --biblatex

seifferth:
BibLaTeX provides \pnfmt, which allows users to force treating
special parts of a citation postnote as pages. This seems to be
equivalent to pandoc-citeproc's locator braces, which provide the
same functionality. Since locator braces are already handled by
the LaTeX Writer anyway (they are stripped), it might be a good
idea to simply translate them into LaTeX when possible.

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX/Citation.hs
+++ b/src/Text/Pandoc/Writers/LaTeX/Citation.hs
@@ -24,12 +24,16 @@ import Text.DocLayout (Doc, brackets, empty, (<+>), text, isEmpty, literal,
 import Text.Pandoc.Walk
 import Text.Pandoc.Writers.LaTeX.Types ( LW )
 
+data CitationPackage = Biblatex
+                     | Natbib
+                       deriving (Eq, Show)
+
 citationsToNatbib :: PandocMonad m
                   => ([Inline] -> LW m (Doc Text))
                   -> [Citation]
                   -> LW m (Doc Text)
 citationsToNatbib inlineListToLaTeX [one]
-  = citeCommand inlineListToLaTeX c p s k
+  = citeCommand Natbib inlineListToLaTeX c p s k
   where
     Citation { citationId = k
              , citationPrefix = p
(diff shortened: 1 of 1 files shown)


jgm:
But not all locators are page numbers....

seifferth:
I just saw that we already talked about this issue in
<https://github.com/jgm/pandoc/issues/5722>. I completely forgot
about that, to be honest.

> But not all locators are page numbers....

Using `\pnfmt` should still work, I believe. Apparently,
it is more of a suggestion to biblatex than a strong
override (such as using `\pno` would be). To quote the manual
(<http://mirrors.ctan.org/macros/latex/contrib/biblatex/doc/biblatex.pdf>,
p. 117):

> `\pnfmt{<text>}`
>
> This command formats its argument `<text>` in the same format as
> `postnote`. The command can be used to format a page range while adding
> additional text in the postnote argument of a cite command.
>
>     \autocite[\pnfmt{378-381, 383} and more]{sigfridsson}

Preliminary testing with biblatex suggests that it might work:

```latex
\documentclass{article}
\begin{filecontents}{\jobname.bib}
@article{tenen-wythoff-2014,
    author = "Dennis Tenen and Grant Wythoff",
    title = "Sustainable Authorship in Plain Text using Pandoc and Markdown",
    year = "2014",
    journal = "The Programming Historian",
    volume = "3",
    doi = "https://doi.org/10.46430/phen0041",
}
\end{filecontents}
\usepackage[style=authoryear]{biblatex}
\addbibresource{\jobname.bib}
\begin{document}
  Arabic number as locator
    \autocite[\pnfmt{20}, additional note]{tenen-wythoff-2014}.
  Roman number as locator
    \autocite[\pnfmt{iv}, additional note]{tenen-wythoff-2014}.
  Part as locator
    \autocite[\pnfmt{pt. III}, additional note]{tenen-wythoff-2014}.
  Something else as locator
    \autocite[\pnfmt{(a)}, additional note]{tenen-wythoff-2014}.
\end{document}
```

The latex above produces the following citations:

Arabic number as locator (Tenen and Wythoff 2014, p. 20, additional note).
Roman number as locator (Tenen and Wythoff 2014, p. iv, additional note).
Part as locator (Tenen and Wythoff 2014, pt. III, additional note).
Something else as locator (Tenen and Wythoff 2014, (a), additional note).

Adding pnfmt around an explicitly delimited citation locator with biblatex
shouldn't hurt, at least.

The implementation is still lacking, of course, as it doesn't support
whitespace in locators yet. It is also inconsistent with pandoc-citeproc,
as my implementation allows multiple locators, while pandoc-citeproc
only allows one. Maybe I could reuse (part of) the postfix parsing
functionality from pandoc-citeproc directly? It has been integrated into
pandoc if I understand correctly, hasn't it?

jgm:
Sorry I've let this languish.
Yes, the code you're talking about is in Text.Pandoc.Citeproc.Locator - parseLocator.

PR #7235, opened 2021-04-20
list of figures before list of tables in LaTeX and ConTeXt templates

jdutant:
See https://groups.google.com/g/pandoc-discuss/c/xqqWWleVO3s


jgm:
Thank you

PR #7238, opened 2021-04-23
Add note about default export behavior of org-mode in 

### LINKED ISSUE #7236: In Org mode, Header with level > 3 are not recognized as headers
Pandoc version: 2.12

Also behaves this way on https://pandoc.org/try/?text=***+Level+3%0A****+Level+4%0A%0A&from=org&to=org&standalone=0

pandoc-org-mode-header-misread-bug

channikhabra:
Since it seem to be a common source of confusion, this PR adds a note in org-mode documentation referring users to consult org-mode's export documentation when they find themselves in a pinch.

Fixes #7236

### DIFF
--- a/doc/org.md
+++ b/doc/org.md
@@ -36,20 +36,20 @@ The following export keywords are supported:
   defaults to stdout unless a target has to be given as a command
   line option.
 
-> *Note*: Pandoc tries to be compatible with org-mode when exporting an org
-> document. If you find some behavior confusing, please do refer to org-mode
-> [Export-Settings](https://orgmode.org/manual/Export-Settings.html)
-> documentation. For example, a common confusion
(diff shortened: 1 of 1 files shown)


tarleb:
Thank you, looks good! Just one thing: The PR uses a block quote to create a box around the note. This is a common method, but semantically not the best. How about using a fenced div instead, e.g. `::: note`? One could even add some `header-includes` directive to style the block, ensuring that the output looks pretty.

channikhabra:
@tarleb Since pandoc.org uses bootstrap anyway, I've added div with `.alert .alert-info` classes instead of blockquotes. Please check and suggest changes.
I haven't figured out how to generate pandoc.org content locally, but with some tinkering in browser's inspector, I think this is how it should end up looking like:

image

tarleb:
Great, thanks

tarleb:
BTW, extra kudos for using link titles.

PR #7241, opened 2021-04-23
writers: Recognize extra syntax definitions

jtojnar:
Since Skylighting does not support Nix syntax, we need to add custom syntax definition using --syntax-definition flag, otherwise files with nix extension will be recognized as bash scripts by the default syntax map. This means that code blocks with nix class will end up with language=bash when converted to DocBook and other formats.

Unfortunately, the writers always use the default syntax map so adding custom definition will not help.

This patch corrects that, allowing the writers to see all custom definitions.

The LaTeX reader suffers from the same issue but it has not been fixed there since it would require wider changes.

## Test

```ShellSession
$ printf '```nix\nfoo\n```' | cabal exec -- pandoc -f markdown -t docbook
<programlisting language="bash">
foo
</programlisting>
$ printf '```nix\nfoo\n```' | cabal exec -- pandoc -f markdown -t docbook --syntax-definition=<(curl -s 'https://gitlab.com/rycee/presentations/-/raw/da77a89576fe8eed2a8b4f73e29b31895d45c98f/2019-12-cph-nur/nix-syntax.xml')
<programlisting language="nix">
foo
</programlisting>
```

### DIFF
--- a/src/Text/Pandoc/Highlighting.hs
+++ b/src/Text/Pandoc/Highlighting.hs
@@ -52,12 +52,12 @@ highlightingStyles =
    ("breezedark", breezeDark),
    ("haddock", haddock)]
 
-languages :: [T.Text]
-languages = [T.toLower (sName s) | s <- M.elems defaultSyntaxMap]
+languages :: SyntaxMap -> [T.Text]
+languages syntaxmap = [T.toLower (sName s) | s <- M.elems syntaxmap]
 
-languagesByExtension :: T.Text -> [T.Text]
-languagesByExtension ext =
-  [T.toLower (sName s) | s <- syntaxesByExtension defaultSyntaxMap (T.unpack ext)]
+languagesByExtension :: SyntaxMap -> T.Text -> [T.Text]
+languagesByExtension syntaxmap ext =
+  [T.toLower (sName s) | s <- syntaxesByExtension syntaxmap (T.unpack ext)]
 
 highlight :: SyntaxMap
           -> (FormatOptions -> [SourceLine] -> a) -- ^ Formatter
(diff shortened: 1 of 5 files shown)


jtojnar:
Fixed the `-Wname-shadowing` warning.

src/Text/Pandoc/Highlighting.hs
```diff
@@ -52,12 +52,12 @@ highlightingStyles =
...
-languages = [T.toLower (sName s) | s <- M.elems defaultSyntaxMap]
+languages :: SyntaxMap -> [T.Text]
+languages syntaxmap = [T.toLower (sName s) | s <- M.elems syntaxmap]
 
-languagesByExtension :: T.Text -> [T.Text]
-languagesByExtension ext =
-  [T.toLower (sName s) | s <- syntaxesByExtension defaultSyntaxMap (T.unpack ext)]
+languagesByExtension :: SyntaxMap -> T.Text -> [T.Text]
+languagesByExtension syntaxmap ext =
+  [T.toLower (sName s) | s <- syntaxesByExtension syntaxmap (T.unpack ext)]
```
jgm, line 61: Note that this would be a breaking API change.  (Not a reason against, necessarily, but should be noted very clearly in the commit log.)

jgm:
It looks good otherwise. I just merged it, and (doh!) forgot to add the commit message noted above.

src/Text/Pandoc/Highlighting.hs
jtojnar: Should I add something like `+ Text.Pandoc.Highlighting: languages and languagesByExtension now take an additional SyntaxMap argument. Get it from writerOpts or use Skylighting’s defaultSyntaxMap. [API change]`.

Or I can create new functions and make the old ones call them with `defaultSyntaxMap`.

PR #7242, opened 2021-04-24
Markdown writer: Cleaner (code)blocks with single class

jtojnar:
When a block only has a single class and no other attributes, it is not necessary to wrap the class attribute in curly braces – the class name can be placed after the opening mark as is.

This will result in bit cleaner output when pandoc is used as a markdown pretty-printer.

## Test

````ShellSession
$ printf '```nix\nfoo\n```' | cabal exec -- pandoc -f markdown -t markdown
``` nix
foo
```
$ printf ':::warning\nfoo\n:::' | cabal exec -- pandoc -f markdown -t markdown
::: warning
foo
:::
````

### CODE REVIEW

File: src/Text/Pandoc/Writers/Markdown.hs [hunk 1]
```diff
@@ -513,7 +519,7 @@ blockToMarkdown' opts (CodeBlock attribs str) = do
      backticks = endline '`'
      tildes = endline '~'
      attrs  = if isEnabled Ext_fenced_code_attributes opts
-                 then nowrap $ " " <> attrsToMarkdown attribs
+                 then nowrap $ " " <> classOrAttrsToMarkdown attribs
```
jtojnar, line 522: I wonder if I should keep the space in case of single class. Traditionally, there is not one.

jgm: I like the space.

jtojnar: Cool, looks like both are widely supported: https://babelmark.github.io/?text=%60%60%60+nix%0Afoo%0A%60%60%60

jgm:
Thanks!

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -279,6 +279,12 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
               escAttrChar '\\' = literal "\\\\"
               escAttrChar c    = literal $ T.singleton c
 
+-- | (Code) blocks with a single class can just use it standalone,
+-- no need to bother with curly braces.
+classOrAttrsToMarkdown :: Attr -> Doc Text
+classOrAttrsToMarkdown ("",[cls],_) = literal cls
```
jgm, line 285: This needed to check also for empty attributes: see #7397

PR #7257, opened 2021-04-29
Docx reader: add handling of vml image objects (jgm#4735)

mbrackeantidot:
As explained on the mailing list as well:

Currently, when parsing a docx, a v:imagedata tag inside a w:r/w:object/v:shape
is ignored. But it could be treated the same as a v:imagedata inside a
w:r/w:pict, both represent images linked by the r:id attribute.

In fact, we now use a pre-processing to transform a
w:r/w:object/v:shape/v:imagedata to a w:r/w:pict/v:imagedata before transforming
with Pandoc and we have tested this on many documents.


src/Text/Pandoc/Readers/Docx/Parse.hs
```diff
@@ -709,17 +709,24 @@ elemToParPart ns element
...
-       -- Todo: check out title and attr for deprecated format.
        Just s -> expandDrawingId s >>= (\(fp, bs) -> return $ Drawing fp "" "" bs Nothing)
        Nothing -> throwError WrongElem
+elemToParPart ns element
+  | isElem ns "w" "r" element
+  , Just objectElem <- findChildByName ns "w" "object" element
+  , Just shapeElem <- findChildByName ns "v" "shape" objectElem
+  , Just imagedataElem <- findChildByName ns "v" "imagedata" shapeElem
+  , Just drawingId <- findAttrByName ns "r" "id" imagedataElem
+  = expandDrawingId drawingId >>= (\(fp, bs) -> return $ Drawing fp "" "" bs Nothing)
```
jgm, line 729: Big improvement in code clarity!

jgm:
Thanks!

PR #7259, opened 2021-04-30
Org writer: inline latex envs need newlines

### LINKED ISSUE #7252: Inline TeX not put on own line with Org export
Pandoc 2.13

If I have a `.md` file with the following content
```md
A maths environment follows
\begin{align*}
  x &= 1 \\ y &= 2
\end{align*}
Here is some standard text.
```

I can tell from `pandoc -f markdown -t texinfo` that the TeX block is interpreted as TeX.
```texinfo
A maths environment follows @tex
\begin{align*}
  x &= 1 \\ y &= 2
\end{align*}
@end tex Here is some standard text.
```

However, when I export to Org, pandoc creates an incorrect/invalid document:
```org
A maths environment follows \begin{align*}
  x &= 1 \\ y &= 2
\end{align*} Here is some standard text.
```

Org supports inline TeX with `\begin{} .... \end{blocks}`, however, `\begin` must be on its own line
> Environments of any kind. The only requirement is that the ‘\begin’ statement appears on a new line, preceded by only whitespace. [[https://orgmode.org/manual/LaTeX-fragments.html](https://orgmode.org/manual/LaTeX-fragments.html)]

i.e.

```org
A maths environment follows
\begin{align*}
  x &= 1 \\ y &= 2
\end{align*}
Here is some standard text.
```

I imagine this is a small tweak, but I'm not quite sure what. Hence an issue rather than a PR :slightly_smiling_face:

tecosaur:
Closes #7252 (discssion there).

As specified in https://orgmode.org/manual/LaTeX-fragments.html, an
inline \begin{}...\end{} LaTeX block must start on a new line.

### DIFF
--- a/src/Text/Pandoc/Writers/Org.hs
+++ b/src/Text/Pandoc/Writers/Org.hs
@@ -407,6 +407,8 @@ inlineToOrg (Math t str) = do
               then "\\(" <> literal str <> "\\)"
               else "\\[" <> literal str <> "\\]"
 inlineToOrg il@(RawInline f str)
+  | elem f ["tex", "latex"] && T.isPrefixOf "\\begin" str =
+    return $ cr <> literal str <> cr
   | isRawFormat f = return $ literal str
   | otherwise     = do
       report $ InlineNotRendered il


tarleb:
Thanks!

PR #7267, opened 2021-05-05
epub Writer: Fix belongs-to-collection XML id choice

nuew:
The EPUB writer previously used the same XML id for both the book identifier and the EPUB collection. This causes an error on epubcheck.


tarleb (approved):
Thank you, looks good!

PR #7287, opened 2021-05-14
Markdown writer: convert definition lists to bullet lists if `definition_lists` option is disabled

### LINKED ISSUE #1039: Please render definition lists as nested unordered lists when definition_lists extension is disabled
2013-10-27 21:14, John MacFarlane skrev:

 I like the idea of changing the behavior this way when the definition_lists extension is disabled.  Could you put up an issue on the issue tracker?

 +++ BP Jonsson [Oct 27 13 16:03 ]:
 I'm just trying to write a script which turns this

```
 Term 1
 :   Definition 1.1
 :   Definition 1.2

      Definition 1.2.a

 :   Definition 1.3
 Term 2
 :   Definition 2.1
      :   Definition 2.1.1
 :   Definition 2.2
```

 into this

```
 -   Term 1
      -   Definition 1.1
      -   Definition 1.2

          Definition 1.2.a

      -   Definition 1.3
 -   Term 2
      -   Definition 2.1
          -   Definition 2.1.1
      -   Definition 2.2
```

 since that seems to me to be the semantically closest equivalent in vanilla markdown.

 In fact I wonder if this couldn't be made the built-in behavior/format for definition lists in markdown_strict and markdown_github output? The current behavior utterly destroys the semantics, which surely makes noone happy:

```
 Term 1
 Definition 1.1

 Definition 1.2

 Definition 1.2.a

 Definition 1.3

 Term 2
 Definition 2.1
 Definition 2.1.1

 Definition 2.2


```

 /bpj

link2xt:
Fixes #1039


test/command/1039.md
```diff
@@ -0,0 +1,34 @@
...
+
+    -   Definition 1.2
+
+        Definition 1.2.a
+
+    -   Definition 1.3
+
+-   Term 2
+
+    -   -   Definition 2.1
```
link2xt, line 29: This is how reader parses this, `Definition 2.1` is parsed as definition, not label. So the result is not what #1039 author expected, but it's correct.

link2xt:
Do not forget to "hide whitespace changes" when reviewing:

jgm:
Thanks for this!  But I'm not quite sold on the proposal yet -- see my comment in the original issue.

PR #7289, opened 2021-05-15
add support for institute

twsh:
This pull request adds support for the institute variable to all the HTML slide show templates.

In each case I have placed the institute after the author, and using the same HTML class (p, H3, span).

The formats are:

* revealjs
* s5
* dzslides
* slidy
* slideous


tarleb:
LGTM. Adding some documentation in the "Variables for HTML slides" in the manual would be great.

alerque (approved):

jgm:
Thanks! - I'll add a note in the manual.

Issue #6639, opened 2020-08-26
Support ZWNJ for LaTeX


lemzwerg:
The standard Unicode way to break a ligature is to insert U+200C, the zero-width non-joiner character.  This is especially useful for German: At word compound boundaries, ligatures should be suppressed.  Examples: `beruf<200C>lich`, `Schaf<200C>fell`, `Dorf<200C>idylle`.

Both polyglossia and babel support this with the `"|` shorthand: `beruf"|lich`. Note that, at least for German, `"|` does not only break up the ligature, it also inserts a small horizontal space of 0.03em if there is no line break (and we are not within a PDF string; see `tex/generic/babel-german/german.ldf` in TeXLive for more).

Looking into `LaTeX.hs` I see that there is already support for U+200B, the zero-width space character.  I thus ask to do the same for U+200C.  However, it's not clear to me how to implement it: The simplest solution would be to add

```Haskell
'\x200C' -> emits "\"|"  -- zero-width non-joiner
```

but normally users disable babel shortshands since they can't be used in Pandoc...

jgm:
This is about the LaTeX writer, I gather.
If you use xelatex or lualatex, won't these unicode characters just be emitted literally in the LaTeX output?

lemzwerg:
Yes, this is about the LaTeX writer.

And yes, xelatex and friends emit ZWNJ literally.  However, most fonts don't contain a glyph for it, causing a visible warning.  Additionally, as explained earlier, it's not sufficient for German, thus my request for a special handling similar to zero-width space if.

At least for scripts like Latin, Cyrillic, or Greek the German approach (i.e., inserting some extra horizontal space) is OK, I believe.  For other scripts, in particular Arabic and Indic scripts, Pandoc should probably do nothing.

jgm:
So if we escape `'\x200C'` as `"\"|"` would that be enough?
I didn't really understand your comment 
"normally users disable babel shortshands since they can't be used in Pandoc..."

lemzwerg:
> So if we escape `'\x200C'` as `"\"|"` would that be enough?

Yes, provided Babel shorthands are active.

> I didn't really understand your comment "normally users disable
> babel shortshands since they can't be used in Pandoc..."

Only within a LaTeX document I can write `f"|l` to break the ligature.  With Pandoc, I get `"|` literally.  For this reason, many users (from what I gather from the internet) explicitly disable Babel shorthands in customized LaTeX templates for Pandoc by saying
```
[...]

PR #7290, opened 2021-05-15
LaTeX writer: convert zero width non-joiner char into TeX command

tarleb:
Closes: #6639

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX/Util.hs
+++ b/src/Text/Pandoc/Writers/LaTeX/Util.hs
@@ -120,6 +120,10 @@ stringToLaTeX context zs = do
          '\'' | ctx == CodeString -> emitcseq "\\textquotesingle"
          '\160' -> emits "~"
          '\x200B' -> emits "\\hspace{0pt}"  -- zero-width space
+         '\x200C' -> emits ("\\texorpdfstring{\\nobreak" <>
+                            "\\discretionary{-}{}{\\kern.03em}" <>
+                            "\\nobreak\\hskip\\skip11}{}")
+                           -- zero-width non-joiner
          '\x202F' -> emits "\\,"
          '\x2026' -> emitcseq "\\ldots"
          '\x2018' | ligatures -> emitquote "`"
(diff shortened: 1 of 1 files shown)


tarleb:
Please read https://github.com/jgm/pandoc/issues/6639#issuecomment-841722028 before merging.

jgm:
The original issue has the babel definition
```
\def\ZWNJ{%
  \texorpdfstring{%
    \TextOrMath{%
      \penalty\@M
      \discretionary{-}{}{\kern.03em}%
      \ifvmode
      \else
        \nobreak
        \hskip\z@skip
      \fi}{}%
  }{}%
}
```
The definition you use here is simpler. How did you arrive at it?  Are the simplifications okay?

tarleb:
I *think* it's ok, but it might not be. The simplification of leaving out `\TextOrMath` is ok, because we don't emit the command in math mode; also `\nobreak` is in the TeX core and defined as `\penalty\@M`. I get less sure when it comes to the missing `\nobreak\hskip\z@skip`: it *might* be ok. Maybe I don't understand TeX well enough to see why it would be needed in this position, and my tests all returned the expected result. If we want to play it safe, we could use `\nobreak\hskip\skip11`, which should have the same meaning, to avoid a pesky `\makeatletter`. I'm also slightly unsure about `\ifvmode`, as I fail to see how the engine could be in vertical mode at that point, but this might also just show my limited understanding of TeX.

tarleb:
To expand on the `\nobreak\hskip\z@skip` info a little: This command sequence comes from the babel macro `\def\bbl@allowhyphens{\ifvmode\else\nobreak\hskip\z@skip\fi}`, which is documented as follows:

> In several languages compound words are used. This means that when TeX has to hyphenate such a compound word, it only does so at the `-` that is used in such words. To allow hyphenation in the rest of such a compound word, the macro `\bbl@allowhyphens` can be used.

All my tests (as well as my limited understanding of the matter) indicate that the macro is not necessary in this case. E.g., hyphenation of `Dorf&zwnj;idyllengenießerin` works as expected for various box sizes.

tarleb:
Argh. Made an error while testing: turns out the `\nobreak\hskip\skip11` *is* needed for proper hyphenation. I've amended the commit.

tarleb:
Closing in favor of #7291.

PR #7291, opened 2021-05-15
LaTeX template: define commands for zero width non-joiner character

tarleb:
Closes: #6639

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -156,10 +156,11 @@ $if(CJKmainfont)$
   \fi
 $endif$
 \fi
-%% Support for special Unicode characters
-% Zero width non-joiner
+$if(zero-width-non-joiner)$
+%% Support for zero-width non-joiner characters.
 \makeatletter
 \def\zerowidthnonjoiner{%
+  % Prevent ligatures and adjust kerning, but still support hyphenating.
   \texorpdfstring{%
     \textormath{\nobreak\discretionary{-}{}{\kern.03em}%
       \ifvmode\else\nobreak\hskip\z@skip\fi}{}%
@@ -172,6 +173,8 @@ $endif$
   \catcode`^^^^200c=\active
   \protected\def ^^^^200c{\zerowidthnonjoiner}
 \fi
+%% End of ZWNJ support
+$endif$
 $if(beamer)$
 $if(theme)$
 \usetheme[$for(themeoptions)$$themeoptions$$sep$,$endfor$]{$theme$}
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -156,6 +156,22 @@ $if(CJKmainfont)$
   \fi
 $endif$
 \fi
+%% Support for special Unicode characters
+% Zero width non-joiner
+\makeatletter
+\def\zerowidthnonjoiner{%
+  \texorpdfstring{%
+    \textormath{\nobreak\discretionary{-}{}{\kern.03em}%
+      \ifvmode\else\nobreak\hskip\z@skip\fi}{}%
+  }{}%
+}
+\makeatother
+\ifPDFTeX
+  \DeclareUnicodeCharacter{200C}{\zerowidthnonjoiner}
+\else
+  \catcode`^^^^200c=\active
+  \protected\def ^^^^200c{\zerowidthnonjoiner}
+\fi
 $if(beamer)$
 $if(theme)$
 \usetheme[$for(themeoptions)$$themeoptions$$sep$,$endfor$]{$theme$}
--- a/src/Text/Pandoc/Writers/LaTeX/Types.hs
+++ b/src/Text/Pandoc/Writers/LaTeX/Types.hs
@@ -40,7 +40,7 @@ data WriterState =
   , stCsquotes      :: Bool          -- ^ true if document uses csquotes
   , stHighlighting  :: Bool          -- ^ true if document has highlighted code
   , stIncremental   :: Bool          -- ^ true if beamer lists should be
-                                     --   displayed bit by bit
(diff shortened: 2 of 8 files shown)


tarleb:
Please read https://github.com/jgm/pandoc/issues/6639#issuecomment-841722028 before merging.

data/templates/default.latex
```diff
@@ -156,6 +156,20 @@ $if(CJKmainfont)$
...
+% Zero width non-joiner
+\def\zerowidthnnonjoiner{%
+  \texorpdfstring{%
+    \textormath{\nobreak\discretionary{-}{}{\kern.03em}}{}%
+  }{}%
+}
+\ifPDFTeX
+  \DeclareUnicodeCharacter{200C}{{\zerowidthnnonjoiner}}
+\else
+  % the next two lines both contain an invisible zero width non-joiner char
```
lemzwerg, line 169: Why not using the `^^^^` notation instead?

data/templates/default.latex
```diff
@@ -156,6 +156,20 @@ $if(CJKmainfont)$
   \fi
 $endif$
 \fi
+%% Support for special Unicode characters
+% Zero width non-joiner
+\def\zerowidthnnonjoiner{%
```
lemzwerg, line 161: What's the reason of using an additional 'n' character in the name?

tarleb: Late night coding :D

data/templates/default.latex
tarleb: The engine appears to be unhappy about anything but the literal unicode char. I had tried `\char"200C`, but that doesn't work either.

lemzwerg: You mean that `^^^^200c` doesn't work?  If this is the case maybe `^` has been redefined?

tarleb: Oh, turns out I was using a capital `C` as required by `\DeclareUnicodeCharacter`, but had to use a lowercase `c` in this case. TeX is weird sometimes.

data/templates/default.latex
```diff
@@ -156,6 +156,21 @@ $if(CJKmainfont)$
...
+\def\zerowidthnonjoiner{%
+  \texorpdfstring{%
+    \textormath{\nobreak\discretionary{-}{}{\kern.03em}%
+      \ifvmode\else\nobreak\hskip\skip11\fi}{}%
+  }{}%
+}
+\ifPDFTeX
+  \DeclareUnicodeCharacter{200C}{{\zerowidthnonjoiner}}
+\else
+  % the next two lines both contain an invisible zero width non-joiner char
```
lemzwerg, line 170: This comment line is now redundant.

data/templates/default.latex
```diff
@@ -156,6 +156,20 @@ $if(CJKmainfont)$
...
+%% Support for special Unicode characters
+% Zero width non-joiner
+\def\zerowidthnonjoiner{%
+  \texorpdfstring{%
+    \textormath{\nobreak\discretionary{-}{}{\kern.03em}%
+      \ifvmode\else\nobreak\hskip\skip11\fi}{}%
+  }{}%
+}
+\ifPDFTeX
+  \DeclareUnicodeCharacter{200C}{{\zerowidthnonjoiner}}
```
lemzwerg, line 168: Is there a reason for the double braces?  It seems to me that they are not needed.

jgm:
I think that, on the whole, I favor this approach over inserting a big string of tex commands every time the non-joiner character appears.  One might object to the addition of a large amount of "noise" to the default template.  I suppose we could make its inclusion conditional on the presence of these characters in the document -- I don't know whether that's a good idea.

data/templates/default.latex
```diff
@@ -156,6 +156,20 @@ $if(CJKmainfont)$
   \fi
 $endif$
 \fi
+%% Support for special Unicode characters
+% Zero width non-joiner
+\def\zerowidthnonjoiner{%
+  \texorpdfstring{%
+    \textormath{\nobreak\discretionary{-}{}{\kern.03em}%
+      \ifvmode\else\nobreak\hskip\skip11\fi}{}%
```
lemzwerg, line 164: [Sorry for the new round of comments]

I'm not sure that `\skip11` is more readable than `\z@skip`.  Is this an intentional change in comparison to the code from babel-german?  If yes, wouldn't it be better to give `\skip11` a name?

tarleb: Agreed. The `\skip11` came from trying to avoid having to use the verbose `\makeatletter`/`\makeatother` combination which I, personally, find quite annoying. OTHO, `\skip11` is *really* cryptic, so let's go for the more verbose variant.

lemzwerg: You don't have to use `\makeatletter` and `\makeatother`.  Simply use `0pt plus0pt minus0pt` instead of `\skip11`.

src/Text/Pandoc/Writers/LaTeX/Types.hs
```diff
@@ -40,7 +40,7 @@ data WriterState =
   , stCsquotes      :: Bool          -- ^ true if document uses csquotes
   , stHighlighting  :: Bool          -- ^ true if document has highlighted code
   , stIncremental   :: Bool          -- ^ true if beamer lists should be
-                                     --   displayed bit by bit
+  , stZwnj          :: Bool          -- ^ true if document a ZWNJ character
```
lemzwerg, line 43: `s/true if document/true if document contains/`

lemzwerg:
Looks good, thanks!

tarleb:
I added a commit that makes the inclusion conditional.

jgm:
Looks good!

PR #7304, opened 2021-05-20
Adds new enumerate itemgroup for ordered lists

### LINKED ISSUE #5016: Remove width attribute for ordered list in ConTeXt writer
I want to remove [width attribute](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/ConTeXt.hs#L231-L236) in ConTeXt writer. This attribute is hardcoded. Would it be possible to turn off or set value of the attribute in a lua filter?

denismaier:
- change ordered list from itemize to enumerate
- adds new itemgroup for ordered lists
- add fontfeature for table figures
- remove width from itemize in context writer

Closes #5016

### DIFF
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -51,6 +51,7 @@ $endif$
 
 % use microtypography
 \definefontfeature[default][default][script=latn, protrusion=quality, expansion=quality, itlc=yes, textitalics=yes, onum=yes, pnum=yes]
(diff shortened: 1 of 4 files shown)


denismaier:
Does anyone know why that fails? The test suite passes here.

tarleb (approved):

tarleb:
Very cool! The CI fails because the `import Data.List.NonEmpty ...` line is no longer needed, thereby causing a warning.

Issue #4102, opened 2017-11-27
Markdown reader: Small caps not recognized with other attributes


adunning:
If one combines `.smallcaps` with another attribute in a span, this is not converted into native SmallCaps. For example:

```shell
pandoc -t latex << EOT
[test1]{.smallcaps lang=la}

[test2]{.smallcaps}

EOT
```

Result:

```latex
\textlatin{test1}

\textsc{test2}
```

mb21:
The expected output for `[test1]{.smallcaps lang=la}` would be the following?

    \textsc{ \textlatin{test1} }

adunning:
The problem is that it's recognized as a span class, but not actual small caps (it doesn't work in .docx either); the output of `echo '[test]{.smallcaps}' | pandoc -t native` is `[Para [SmallCaps [Str "test"]]]`.

adunning:
(In other words, it has nothing to do with the LaTeX writer; I was only using that format as a readable example.)

jgm:
Parsing it as SmallCaps would lose the lang attribute -- I
think that's why we didn't do it that way.

However, I suppose we could create this:

    Span ("",[],[("lang","fr")]) [SmallCaps [Str "test"]]


+++ Andrew Dunning [Nov 27 17 20:08 ]:

ickc:
> Parsing it as SmallCaps would lose the lang attribute -- I think that's why we didn't do it that way.

It has always been like this in pandoc even without the "native small caps syntax" such as `[test2]{.smallcaps}`. The problem is the native AST element representing small caps does not allow attribute. So even in the old syntax, only a very particular combination would be parsed as the small caps AST element.

(That's why when I first PR the support of `[SC]{.smallcaps}` syntax this behavior isn't changed. So in a sense this is a "not a regression" kind of issue.)

In this sense it is related to #684.

PR #7307, opened 2021-05-21
Markdown reader: allow more attributes in special spans

tarleb:
Spans with "smallcaps" as the first class are converted to *SmallCaps*
elements. While previously no other classes or attributes were allowed,
additional classes, attributes, and an identifier are not permitted and
kept in a *SmallCaps* wrapping *Span* element.

The same change is applied to underline spans, where the first class
must be either "ul" or "underline".

Closes: #4102

### DIFF
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE FlexibleContexts    #-}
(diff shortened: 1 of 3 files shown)


tarleb:
I'm having second thoughts about this change, so I'm putting it here as a PR.

jgm:
I think this makes sense, but why require that the special class be the first one?
(I guess it makes pattern-matching faster, but probably only slightly.)

tarleb:
My thought here was that it might give the best balance between new functionality and unintended conversions. I wouldn't be too surprised to see `ul` be used as a class in other contexts, and the proposed solution still allows to do so.

tarleb:
I've added a second commit (intended to be squashed) which removes the limitation of expecting the special class as the first, allowing it to occur in any position.

jgm:
I really don't know what's best here.  Your concern about people wanting to use a span with class `ul` and not wanting that to be parsed as an Underline element gives me pause.  On the other hand, the convention that the special class must occur first seems a bit too "magic" and might surprise people.  And even with this, we might get some people who want just a span with class `ul`.

tarleb:
Maybe we should just keep the current behavior and document the behavior as intentional? Or, but that's a bit of a stretch, we could hide the new behavior behind the `attributes` extension.

I note that neither the `.underline` nor `.ul` class are documented yet in the manual; the changelog only states "Support new Underline element in readers and writers". Removing support for `.ul` should not support too many problems.

jgm:
I don't think I'd want to remove the current behavior for `ul`, if that's what you meant.  That would cause existing documents to break.  It should be documented, though.

jgm:
Note that in a recent commit 649133fe5ecf6a2408eeb3c01bc654f71754a590 `.ul` and `.underline` were documented.

Maybe I'll take out the documentation for `.ul`, just in case we decide to remove it.

hftf:
> Note that in a recent commit [649133f](https://github.com/jgm/pandoc/commit/649133fe5ecf6a2408eeb3c01bc654f71754a590) `.ul` and `.underline` were documented.
> 
> Maybe I'll take out the documentation for `.ul`, just in case we decide to remove it.

I believe the Pandoc-flavored Markdown Writer still outputs `[a]{.ul}` from the input `[Underline [Str "a"]]`, doesn't it?

https://github.com/jgm/pandoc/blob/407de98b5e3971c84c7e89de5f2a9d317b4d4557/src/Text/Pandoc/Writers/Markdown/Inline.hs#L304-L317

(Speaking of the inconsistent output of Underline between Pandoc-flavored Markdown and the other Markdown flavors, this could be a good opportunity to decide/discuss whether "raw HTML" like `<u>` belongs in Pandoc-flavored Markdown output, even with `bracketed_spans` or `native_spans` enabled. See also https://github.com/jgm/pandoc/issues/5825#issuecomment-683109845; I'll try to get that new issue filed soon.)

jgm:
So, the suggestion is to parse `<u>...</u>` as a native Underline element, and to generate it from an Underline element?
That does have a certain attraction.  On the other hand, it isn't really consistent with how other raw HTML bits are treated.  (`<em>..</em>` isn't parsed as Emphasis, for example).

tarleb:
The second commit in this PR removes the limitation of the special class (like *smallcaps*) having to be the first class, making it possible to use markup like `[Populus Romanus]{.foo .smallcaps lang="latin"}`.

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1819,30 +1819,48 @@ bracketedSpan = do
...
--- no other attributes are set or if style is "font-variant:small-caps"
--- (and no other attributes are set).
+    return $ case smallCapsAttr attr of
+      Just ("", [], []) -> B.smallcaps <$> lab
+      Just scAttr       -> B.spanWith scAttr . B.smallcaps <$> lab
+      Nothing           -> case underlineAttr attr of
+        Just ("", [], []) -> B.underline <$> lab
+        Just ulAttr       -> B.spanWith ulAttr . B.underline <$> lab
+        Nothing           -> B.spanWith attr <$> lab
+
```
jgm, line 1829: Maybe, to be consistent with the recent change to the HTML writer, we should allow `[test]{.underline .smallcaps}` to turn into `Underline [SmallCaps [Str "test"]]`?  Not sure, what do you think?

jgm: This could be done fairly elegantly with a fold over the classes, similarly to what I did in the HTML writer.

tarleb: Nice! Done.

PR #7313, opened 2021-05-25
Support GHC 9.0.1

tarleb:


jgm:
Added to 2.14 release project, because I think it's worth having this in the next release if at all possible (unless we're blocked by dependencies).

tarleb:
Should be ok now. Changes are fairly minimal:

- upgraded haddock-library to new major version; only one small constructor change affects the code.
- changed order for TH JSON instances; dependencies for other instances must go first.
- adjusted code to avoid warnings due to the smarter pattern completeness checks.
- use `Alt Maybe` instead of `First` and `Optional`, as the latter is deprecated and causes a deprecation warning.

jgm:
Perfect, thank you.

PR #7316, opened 2021-05-26
Add a short guide to pandoc's sources

tarleb:
Pandoc's architecture is fairly stable, so it might help newcomers to have a document that explains the basics.


tarleb:
Cc @argent0, who will be working on the "better figures" feature during this year's GSoC.

argent0:
Very nice thank you.

I noticed a possibly misused word at line 13:

```
13  very module architecture, rendering it an excellent learning
```

Perhaps you meant `modular`.

jgm:
Note that there's already a bit of a guide to the code organization in CONTRIBUTING.md (section entitled "The Code").
And on the architecture, there's `doc/using-the-pandoc-api.md`.  Should this new material be folded into CONTRIBUTING.md, with references to the latter?

jgm:
One way to organize a guide like this might be:

- Modifying a writer
- Modifying a reader
- Adding a new writer
- Adding a new reader
- Adding a command-line option
- Adding an exension
- Other

That is, it could be organized by the type of change, with each type explaining what needs to be changed.  (To add a writer, you need to add the writer module, add it to exposed-modules in pandoc.cabal, modify T.P.Writers, modify the MANUAL.txt, add tests.)

tarleb:
Thanks @argent0, good catch.

I hadn't considered the section in CONTRIBUTING; merging seems like a good idea.

My idea for the guide was that it would be usable by people interested in modifying pandoc, as well as those who merely want to learn the internals for intellectual reasons¹. Less of a "cookbook" and more of a "blueprint", which shows how components interact with each other. But I like the step-by-step approach, too, maybe I should add a recipes section?

¹ Inspired by [the Lua sources reading list](https://www.reddit.com/r/programming/comments/63hth/ask_reddit_which_oss_codebases_out_there_are_so/c02pxbp?utm_source=share&utm_medium=web2x&context=3) by Mike Pall.

argent0:
Perhaps extending the Readers section to mention that they use `Pandoc.Builder` from `pandoc-types` to build the AST.

> # Readers
> 
> Readers generate the internal representation from their input using `Text.Pandoc.Builder` in the `pandoc-types` package. These are convenient functions for building the AST programmatically. 
> 
> The same distinction that applies to writers also applies to
> readers. Readers for XML formats use XML parsing libraries, while
> plain text formats are parsed with [parsec].

tarleb:
I added a section on the T.P.Builder module and started to merge "The code" from CONTRIBUTING.md.

tarleb:
I'm not sure yet how to best handle the overlap with `using-the-pandoc-api.md`. Linking to the relevant sections of that doc should do, but some minor redundancy might be unavoidable.

doc/short-guide-to-pandocs-sources.md
```diff
@@ -0,0 +1,260 @@
...
+  - `Text.Pandoc.Highlighting` contains the interface to the
+    skylighting library, which is used for code syntax highlighting.
+  - `Text.Pandoc.ImageSize` is a utility module containing functions for
+    calculating image sizes from the contents of image files.
+  - `Text.Pandoc.MIME` contains functions for associating MIME types
+    with extensions.
+  - `Text.Pandoc.Lua.*` implement Lua filters.
+  - `Text.Pandoc.Options` defines reader and writer options.
+  - `Text.Pandoc.PDF` contains functions for producing a PDF from a
+    LaTeX source.
```
mb21, line 238:
```suggestion
  - `Text.Pandoc.PDF` contains functions for producing a PDF.
```

jgm:
I think it would be great to merge some version of this, @tarleb.
Even if it's not quite finished.

Issue #7321, opened 2021-05-27
Page detection in suffix is too eager


tstenner:
``` markdown
---
references:
- id: fenner2012a
  title: One-click science marketing
  author:
  - family: Fenner
    given: Martin
  container-title: Nature Materials
  volume: 11
  issue: 4
  publisher: Nature Publishing Group
  type: article-journal
  issued:
    year: 2012
---

[@fenner2012a, $a^2+b^2=c^2$]

[@fenner2012a, with $a^2 + b^2 = c^2$]
```

With `pandoc --csl https://raw.githubusercontent.com/citation-style-language/styles/master/apa.csl test.md -C -t plain` pandoc 2.13 (citeproc 0.3.0.9) gives the following result:

```
(Fenner, 2012, p. a^2+b^2=c^2)

(Fenner, 2012, with a² + b² = c²)
```

In the first example, the equation in the suffix is interpreted as page number, incorrectly prefixed with "p." and then not parsed as inline math. Adding any word before the equation causes the suffix to be rendered correctly as seen in the second example.

jgm:
Transferring to pandoc, because the code that parses the locator from the suffix is there.

jgm:
You should be able to work around this by using the braces syntax for explicit locators:
`[fenner2012a, {}$a^2+b$]`

tstenner:
Thanks, that brought me on the right track.
Your example swallows the , `(Fenner 2012a² + b)`, but `[@fenner2012a{}, $a^2+b$]` works as intended.

I have pushed a PR that adds this example to the manual.

PR #7322, opened 2021-05-27
Add citation example for locators and suffixes

tstenner:
Some suffixes can be misinterpreted as locators (as in `(Fenner, 2012, p. 1-15)`) (#7321).
The manual mentions the case where a suffix can be forced to be parsed as locator, but the opposite direction isn't obvious.


jgm:
Good idea

PR #7333, opened 2021-05-30
Mention GladTeX for EPUB export

humenda:


mb21:
Thanks, probably we should add a link to https://pandoc.org/MANUAL.html#math-rendering-in-html as well?

humenda:
Sure, should I do this?

doc/epub.md
```diff
@@ -141,8 +141,12 @@ Pandoc has an EPUB3 writer.  It renders LaTeX math into MathML, which
 EPUB3 readers are supposed to support (but unfortunately few do).
 
 Of course, this isn't much help if you want EPUB2 output (`pandoc -t epub2`)
-or target readers that don't support MathML. Then you should try using the
-`--webtex` option, which will use a web service to convert the TeX to an image.
+or target readers that don't support MathML. Then you have two options:
+
+1.  Use the `--webtex` option, which will use a web service to convert the TeX to an image.
```
jgm, line 146: We try to keep lines < 80 columns.

doc/epub.md
```diff
@@ -141,8 +141,12 @@ Pandoc has an EPUB3 writer.  It renders LaTeX math into MathML, which
...
+2.  Use the `--gladtex` option to convert maths into SVG images on your local
+    machine. The added benefit is the increased accessibility for blind users
+    that are able to see the LaTeX sources of the formulas instead of the image.
```
jgm, line 149: Is this "added benefit" specific to `--gladtex` or does it apply to `--webtex` too?
(`--webtex` gives you an alt attribute with the original tex.)

Also, as already noted, a link to the part of the manual that gives details on these options would be great.  (Or you could just make `--gladtex` and `--webtex` into links.)

humenda:
>> +    machine. The added benefit is the increased accessibility for blind users
>>+    that are able to see the LaTeX sources of the formulas instead of the image.
>
>Is this "added benefit" specific to `--gladtex` or does it apply to `--webtex` too?
>(`--webtex` gives you an alt attribute with the original tex.)

Changed.

>Also, as already noted, a link to the part of the manual that gives details on these options would be great.  (Or you could just make `--gladtex` and `--webtex` into links.)

Should they link to <https://pandoc.org/MANUAL.html#math-rendering-in-html>?

jgm:
You can actually link directly to the option.  They get put into spans with identifiers, e.g
```
<span id="option--webtex" class="option-anchor"><span class="nowrap"><code class="option-def">--webtex</code></span>[<span class="nowrap"><code>=</code></span><em>URL</em>]</span>
```

humenda:
Thanks. I have updated the PR. I also adapted the manual slightly.

MANUAL.txt
```diff
@@ -1415,13 +1415,17 @@ results only for basic math, usually you will want to use
...
         pandoc -s --gladtex input.md -o myfile.htex
-        gladtex -d myfile-images myfile.htex
-        # produces myfile.html and images in myfile-images
+        gladtex -d image_dir myfile.htex
+        # produces myfile.html and images in image_dir
+
+    Alternatively, you can use GladTeX as a Pandoc filter:
+
+        pandoc -t json input.md | gladtex -d image_dir -P - \
+            |pandoc -f json -t html -o myfile.html -
```
jgm, line 1428: This is cool, I had no idea gladtex could be used this way!
I'd suggest formatting like
```
pandoc -t json input.md | gladtex -d image_dir -P - | \
    pandoc -f json -t html -o myfile.html
```
(You don't need the `-` for stdin on the last pandoc invocation, and I think it's clearer with the pipe on the line before.  If you'd rather have the pipe on the second line, how about leaving a space after it?)

humenda:
Thanks again. There's actually a third way on how to use GladTeX as a filter:

```
# if set, GladTeX will behave like a pandoc filter, also allows specifying
# additional args
export GLADTEX_ARGS=""
pandoc -F gladtex -t html doc.md
```

But this means that people could not use spaces in the `-d` option and it's
not the GladTeX, but the pandoc manpage. Le t me know whether this should be
mentioned. If not, you can go ahead and merge :).

jgm:
```
# if set, GladTeX will behave like a pandoc filter, also allows specifying
# additional args
export GLADTEX_ARGS=""
pandoc -F gladtex -t html doc.md
```

> But this means that people could not use spaces in the `-d` option and it's
not the GladTeX, but the pandoc manpage. 

I wonder if we shouldn't have this *instead of* the thing you have above (with the pipes).
But I don't quite understand the limitation.  with `GLADTEX_ARGS=""`, does it default to `image_dir` for the argument of `-d`?  Are you saying you can do `GLADTEX_ARGS='-d "dir with spaces"'`?
If we do use this, we should use `--filter` instead of `-F`.

humenda:
Well, until now, quoting in `GLADTEX_ARGS` was not supported, but I have just
added it. I can rework the example to use `--filter` and a path with spaces
for a proof-of-concept. The version of GladTeX that supports this should be
out in a few weeks time. I could open a new PR later on as well.

MANUAL.txt
```diff
@@ -1415,13 +1415,18 @@ results only for basic math, usually you will want to use
...
 
         pandoc -s --gladtex input.md -o myfile.htex
-        gladtex -d myfile-images myfile.htex
-        # produces myfile.html and images in myfile-images
+        gladtex -d image_dir myfile.htex
+        # produces myfile.html and images in image_dir
+
+    Alternatively, you can use GladTeX as a Pandoc filter:
+
+        pandoc -t json input.md | gladtex -d image_dir -P - |
```
jgm, line 1427: I think you need
`| \`
(line continuation after pipe)

humenda:
Thanks, fixed.

MANUAL.txt
```diff
@@ -1415,13 +1415,18 @@ results only for basic math, usually you will want to use
...
-        gladtex -d myfile-images myfile.htex
-        # produces myfile.html and images in myfile-images
+        gladtex -d image_dir myfile.htex
+        # produces myfile.html and images in image_dir
+
+    Alternatively, you can use GladTeX as a Pandoc filter:
+
+        pandoc -t json input.md | gladtex -d image_dir -P - | \
+            pandoc -f json -t html -o myfile.html
+
```
jgm, line 1429: Now that I think about it, I don't think this description of filter usage really belongs in the manual's description of the `--gladtex` option.  (`--gladtex` isn't used here at all.)  I'd suggest putting this here: https://github.com/jgm/pandoc/wiki/Pandoc-Filters#math-related

humenda:
Agreed, these are separate operation modes. Removed.

jgm:
looks good now!
I'll have to give gladtex a try.

humenda:
>looks good now!

Thanks for the continued feedback.

>I'll have to give gladtex a try.

I'm looking forward to your feedback and criticism :).

PR #7357, opened 2021-06-05
DocBook: correct list of admonitions

jtojnar:
In https://github.com/jgm/pandoc/pull/6922, I accidentally added non-existent admonitions to the DocBook writer (taken from rST). Let’s remove them. Though `danger` is actually supported since DocBook 5.2 so let’s add it to the DocBook reader.


jgm:
Thanks!

PR #7358, opened 2021-06-05
CommonMark writer: do not throw away attributes when Ext_attributes is enabled

jtojnar:
`﻿Ext_attributes` covers at least the following:

- `Ext_fenced_code_attributes`
- `Ext_header_attributes`
- `Ext_inline_code_attributes`
- `Ext_link_attributes`

### DIFF
--- a/src/Text/Pandoc/Writers/Markdown.hs
+++ b/src/Text/Pandoc/Writers/Markdown.hs
@@ -322,9 +322,12 @@ blockToMarkdown' opts (Div attrs ils) = do
     case () of
          _ | isEnabled Ext_fenced_divs opts &&
              attrs /= nullAttr ->
-                nowrap (literal ":::" <+> classOrAttrsToMarkdown attrs) $$
-                chomp contents $$
-                literal ":::" <> blankline
+                let attrsToMd = if variant == Commonmark
+                                then attrsToMarkdown
+                                else classOrAttrsToMarkdown
+                in nowrap (literal ":::" <+> attrsToMd attrs) $$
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -286,7 +286,7 @@ classOrAttrsToMarkdown attrs = attrsToMarkdown attrs
 
 linkAttributes :: WriterOptions -> Attr -> Doc Text
 linkAttributes opts attr =
-  if isEnabled Ext_link_attributes opts && attr /= nullAttr
```
jtojnar, line 289: I wonder if we could just reuse the `linkAttributes` from `Inline`.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -322,7 +322,7 @@ blockToMarkdown' opts (Div attrs ils) = do
     case () of
          _ | isEnabled Ext_fenced_divs opts &&
              attrs /= nullAttr ->
-                nowrap (literal ":::" <+> classOrAttrsToMarkdown attrs) $$
+                nowrap (literal ":::" <+> (if variant == Commonmark then attrsToMarkdown else classOrAttrsToMarkdown) attrs) $$
```
jgm, line 325: Reminder to keep lines <= 80 chars.

jtojnar: Sorry, fixed.

jgm:
Looks good, thanks!

PR #7360, opened 2021-06-06
Docbook writer: Interpret links without contents as cross-references

jtojnar:
This is what MyST does:

https://myst-parser.readthedocs.io/en/latest/using/syntax.html#targets-and-cross-referencing

### DIFF
--- a/src/Text/Pandoc/Writers/Docbook.hs
+++ b/src/Text/Pandoc/Writers/Docbook.hs
@@ -429,7 +429,9 @@ inlineToDocbook opts (Link attr txt (src, _))
   | otherwise = do
       version <- ask
       (if "#" `T.isPrefixOf` src
-            then inTags False "link" $ ("linkend", writerIdentifierPrefix opts <> T.drop 1 src) : idAndRole attr
+            then let tag = if null txt then "xref" else "link"
+                 in  inTags False tag $
+                     ("linkend", writerIdentifierPrefix opts <> T.drop 1 src) : idAndRole attr
             else if version == DocBook5
                     then inTags False "link" $ ("xlink:href", src) : idAndRole attr
                     else inTags False "ulink" $ ("url", src) : idAndRole attr )
--- a/test/Tests/Writers/Docbook.hs
+++ b/test/Tests/Writers/Docbook.hs
@@ -32,9 +32,11 @@ which is in turn shorthand for
 -}
 
 infix 4 =:
-(=:) :: (ToString a, ToPandoc a)
-     => String -> (a, String) -> TestTree
+(=:), testDb4, testDb5 :: (ToString a, ToPandoc a)
+                       => String -> (a, String) -> TestTree
 (=:) = test docbook
+testDb4 = test docbook
+testDb5 = test docbook5
 
 lineblock :: Blocks
 lineblock = para ("some text" <> linebreak <>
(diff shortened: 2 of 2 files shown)


tarleb (commented):
Thank you for the PR and sorry for the late reply! This somehow got lost.

I think this is a good and very sensible change. There is one minor issue with code style, but otherwise this seems good to go.

tarleb:
For others looking at this PR, here are the docs for [xref](https://tdg.docbook.org/tdg/5.1/xref.html).

test/Tests/Writers/Docbook.hs
```diff
@@ -32,9 +32,11 @@ which is in turn shorthand for
 -}
 
 infix 4 =:
-(=:) :: (ToString a, ToPandoc a)
-     => String -> (a, String) -> TestTree
+infix 4 =::
+(=:), (=::) :: (ToString a, ToPandoc a)
+            => String -> (a, String) -> TestTree
 (=:) = test docbook
+(=::) = test docbook5
```
jtojnar, line 39: I do not really like this operator, it looks like assigning type.
Ideally we would have `(=:5)` but Haskell does not seem to allow numbers in operators.

src/Text/Pandoc/Writers/Docbook.hs
```diff
@@ -430,7 +430,11 @@ inlineToDocbook opts (Link attr txt (src, _))
   | otherwise = do
       version <- ask
       (if "#" `T.isPrefixOf` src
-            then inTags False "link" $ ("linkend", writerIdentifierPrefix opts <> T.drop 1 src) : idAndRole attr
+            then
```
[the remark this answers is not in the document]
jtojnar: Moved the positive body next to `then`. Not sure what the proper alignment of `let…in` is, though, there are at least two in this file.

test/Tests/Writers/Docbook.hs
tarleb: Agree, it's a bit confusing. I think we could define `testdb5` or `db5` and set ``infix 4 `db5` ``. Or we could do fancy stuff with unicode symbols like `⚄` and `⚃`, but, uuuh, maybe not.

jtojnar: Or maybe just simple `test docbook5 "test name" $`

tarleb: :+1:

src/Text/Pandoc/Writers/Docbook.hs
```diff
@@ -429,7 +429,9 @@ inlineToDocbook opts (Link attr txt (src, _))
   | otherwise = do
       version <- ask
       (if "#" `T.isPrefixOf` src
-            then inTags False "link" $ ("linkend", writerIdentifierPrefix opts <> T.drop 1 src) : idAndRole attr
+            then let tag = if null txt then "xref"else "link"
```
tarleb, line 432: missing space between `"xref"` and `else`

test/Tests/Writers/Docbook.hs
jtojnar: Done.

jtojnar:
Hmm, turns out there was some specialization so we need helper still. Or maybe we could specify `default ToString String`?

tarleb:
Wouldn't that (`default ToString`) still give a warning? We could use `Language TypeApplications` and write `test @Text`, but defining `testDB5` or similar seems like the easiest solution.

jtojnar:
Seems to work now.

tarleb:
Excellent, thanks.

@jgm I'm wondering if we should hide it behind the `xrefs_name` extension, but I don't think that's necessary.

PR #7389, opened 2021-06-17
DocBook reader: Preserve callouts and xrefs in code

jtojnar:
When converting DocBook documents like https://github.com/NixOS/nixpkgs/blob/c06cb69ca0dc02b37e27104ac8415c3cab173328/nixos/modules/programs/plotinus.xml#L27, these need to be dealt with manually.

### DIFF
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -1076,7 +1076,10 @@ strContentRecursive = strContent .
   (\e' -> e'{ elContent = map elementToStr $ elContent e' })
 
 elementToStr :: Content -> Content
-elementToStr (Elem e') = Text $ CData CDataText (strContentRecursive e') Nothing
+elementToStr (Elem e') =
+  case qName (elName e') of
+        n | n == "xref" || n == "co" -> Text $ CData CDataText (showElement e') Nothing
+        _ -> Text $ CData CDataText (strContentRecursive e') Nothing
 elementToStr x = x
 
 parseInline :: PandocMonad m => Content -> DB m Inlines


src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1076,7 +1076,10 @@ strContentRecursive = strContent .
   (\e' -> e'{ elContent = map elementToStr $ elContent e' })
 
 elementToStr :: Content -> Content
-elementToStr (Elem e') = Text $ CData CDataText (strContentRecursive e') Nothing
+elementToStr (Elem e') =
+  case qName (elName e) of
+        n | n == "xref" || n == "co" -> Text $ CData CDataText (show e) Nothing
```
jtojnar, line 1081: The show is just a placeholder until I find out a function to stringify the [`Element`](https://hackage.haskell.org/package/xml-conduit-1.9.1.1/docs/Text-XML.html#t:Element).

jtojnar:
Ideally, the AST would support rich code blocks and the dumbing down would be handled in writers but until then let’s preserve the data.

jtojnar:
It might be also useful to control this behaviour with an extension.

jgm:
Can you explain the motivation for this change, and what it does?
There is not even a test to show how the behavior changes.

jtojnar:
In NixOS manual, we have code like

```
<programlisting>
<xref linkend="opt-services.xserver.desktopManager.gnome.enable"/> = true;
</programlisting>
```

docbook-xsl renders it as

> <pre>
> <a href="https://nixos.org/manual/nixos/stable/options.html#opt-services.xserver.desktopManager.gnome.enable">services.xserver.desktopManager.gnome.enable</a> = true;
> </pre>

----

We are trying to switch our documentation from DocBook to CommonMark and using pandoc for the conversion.

Currently, pandoc parses the code as

> ```
>  = true;
> ```

since `xref` is an empty element and `elementToStr` will return empty string for it.

We would like it to produce the following

> ```
> <xref linkend="opt-services.xserver.desktopManager.gnome.enable"/> = true;
> ```

The `xref` tag is in the code block verbatim, letting us to deal with it manually. (Probably using some pandoc filter.)

----

I will add a test but first I would like to know if this behaviour is reasonable or if it should be optional (off-by-default extension).

jgm:
I don't really like the idea of including a string in the code block that actually isn't verbatim code.
Would it be possible for pandoc to resolve the xref?  I'm not sure how that happens -- is all the information needed for doing that elsewhere in the docbook document?

jtojnar:
In our use case, the anchors are located in different files (that are `XInclude`d into the final book along with the file containing the `xref`s). So we would need some other mechanism to resolve the links, which would be even more complex than this. Leaving it up to the user (e.g. to resolve them using a filter script) would probably be cleaner. The issue is it requires preserving the data.

But I agree that mixing code with metadata without any way to distinguish them would be ugly. Alternative would be keeping this behind an extension flag `Ext_docbook_preserve_markup_in_codeblocks` and when a XML element were detected in a code block XML element, it would set the raw contents of the XML element as the `CodeBlock` contents (and add a `raw_docbook` class to the code block to distinguish it).

Much nicer option would be modifying the AST to support rich code blocks – this is supported by other languages than DocBook (HTML, rST). But yeah, it would be a pretty big backwards compatibility break. (Or adding `RichCodeBlock` if that is more acceptable.)

jtojnar:
Or maybe an extension that would keep the elements that would cause information loss as `RawBlock`, to be handled manually. That feels like kosher solution that would be easier to implement than a `AST` changes.

But I could even do this short term and work on the AST changes long term if you think it is a good idea.

jgm:
A pragmatic solution would be to preprocess, resolving these references, before passing to pandoc.

PR #7401, opened 2021-06-22
Docx parser: nested fields and pageref fields

mbrackeantidot:
Pageref fields can be hyperlinks, so I implemented treating them as hyperlinks. These fields are often used in tables of contents.

To make them work properly, I had to make nested fields work properly, as a table of contents is also a field, so pagerefs are usually nested in another field. Making nested fields work also fixes a bug, for example in the test document I added (nested_instrText.docx).


mbrackeantidot:
@jkr As discussed on the mailing list, here's my pull request.

mbrackeantidot:
Don't merge yet. I will add a new commit. I've found another bug. Apparently not every field has a fldCharType="separate" element.
The current commits shouldn't change, though, so feel free to comment on those. The bug was already there and merits its own commit.

mbrackeantidot:
I fixed the other bug as well. So the pull request is ready. @jkr

jkr:
Sorry that I've been incommunicado. Having finally looked at this, and especially trusting the tests, this looks good to me. Milan -- apologies for the long wait.

jkr:
Actually -- one quick question, @mbrackeantidot, and sorry I didn't notice this before: I just wanted to make sure about the reason for the change in the outcome for `0_level_header` test? It's because it's using a docx-generated TOC and therefore produces the sort of links your change picks up, right?

PR #7403, opened 2021-06-23
reveal.js template: Properly set boolean config options

### LINKED ISSUE #7402: reveal.js: Explicitly setting boolean options to `false` fails
## Problem

Since Pandoc 2.14.0.3 (more specifically https://github.com/jgm/pandoc/commit/14b2eb2aeb21b3dee7d07c7348ebd2c586d6a866), all [reveal.js config options](https://revealjs.com/config/) are explicitly set to their default values. 

While leaving the default of  a boolean variable untouched (1) or flipping it to `true` (2) works fine, setting it to `false` (3) fails.

### 1

```sh
echo "---
title: 'Test'
---

# My presentation

blabla
" | pandoc --to=revealjs --from=markdown --self-contained | grep mouseWheel
```

```
        mouseWheel: false,
```

### 2

```sh
echo "---
title: 'Test'
mouseWheel: true
---

# My presentation

blabla
" | pandoc --to=revealjs --from=markdown --self-contained | grep mouseWheel
```

```
        mouseWheel: true,
```

### 3

```sh
echo "---
title: 'Test'
mouseWheel: false
---

# My presentation

blabla
" | pandoc --to=revealjs --from=markdown --self-contained | grep mouseWheel
```

```
        mouseWheel: ,
```

## Workaround

Explicitly setting the boolean `false` as a string works:

```sh
echo "---
title: 'Test'
mouseWheel: 'false'
---

# My presentation

blabla
" | pandoc --to=revealjs --from=markdown --self-contained | grep mouseWheel
```

```
        mouseWheel: false,
```

## Pandoc version
2.14.0.3

salim-b:
Fixes #7402

### DIFF
--- a/data/templates/default.revealjs
+++ b/data/templates/default.revealjs
@@ -91,11 +91,11 @@ $endif$
       // https://revealjs.com/config/
       Reveal.initialize({
         // Display controls in the bottom right corner
-        controls: $controls$,
+        controls: $if(controls)$true$else$false$endif$,
 
         // Help the user learn the controls by providing hints, for example by
         // bouncing the down arrow when they first encounter a vertical slide
-        controlsTutorial: $controlsTutorial$,
+        controlsTutorial: $if(controlsTutorial)$true$else$false$endif$,
 
         // Determines where controls appear, "edges" or "bottom-right"
         controlsLayout: '$controlsLayout$',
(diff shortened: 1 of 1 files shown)


jgm:
This won't work by itself -- we'd also need changes to the HTML writer, which currently sets the default values of these variables to strings.  The *string* `false` counts as a true value.
I think the change I just pushed to doctemplates is a better path forward.

Issue #7416, opened 2021-06-29
Duplicated alt text attribute on html output


argent0:
**Explain the problem.**

When an `alt` attribute is specified in the input two alternative texts are inserted in the output HTML.

```
$ pandoc -f markdown -t html
![caption](../media/rId25.jpg "title"){alt="alt"}

<figure>
<img src="../media/rId25.jpg" title="title" alt="alt" alt="caption" /><figcaption aria-hidden="true">caption</figcaption>
</figure>
```

I believe the duplication should be avoided when the `alt` attribute is explicitly provided.

**Expected behavior**

```
$ pandoc -f markdown -t html
![caption](../media/rId25.jpg "title"){alt="alt"}

<figure>
<img src="../media/rId25.jpg" title="title" alt="alt" /><figcaption aria-hidden="true">caption</figcaption>
</figure>
```

**Background**

Related issue:  #3928

[Discussion](https://github.com/tarleb/pandoc/discussions/12)

**Pandoc version?**

```
$ pandoc --version
pandoc 2.14.0.2
Compiled with pandoc-types 1.22, texmath 0.12.3, skylighting 0.10.5.1,
citeproc 0.4.0.1, ipynb 0.1.0.1
```

jgm:
Closed by #7417.

PR #7425, opened 2021-07-02
HTML5 writer, remove aria-hidden when explicit atl text is provided.

argent0:
Related to #7416 

This PR removes the `aria-hidden=true` attribute from the HTML5 output when an explicit alt text is provided.

```
% pandoc -f markdown -t html
![caption](../media/rId25.jpg "title"){alt="alt"}

^D
<figure>
<img src="../media/rId25.jpg" title="title" alt="alt" /><figcaption>caption</figcaption>
</figure>
```

Quoting from the [w3 recommendation](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden), emphasis mine:

> aria-hidden (state)
>
> Indicates whether the element is exposed to an accessibility API. See related aria-disabled.
>
> User agents determine an element's hidden status based on whether it is rendered, and the rendering is usually controlled by CSS. For example, an element whose display property is set to none is not rendered. An element is considered hidden if it, or any of its ancestors are not rendered or have their aria-hidden attribute value set to true.
>
> Authors MAY, with caution, **use aria-hidden** to hide visibly rendered content from assistive technologies only if the act of hiding this content is intended to improve the experience for users of assistive technologies **by removing redundant or extraneous content**. Authors using aria-hidden to hide visible content from screen readers MUST ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.

It made sense to use it when the content was "redundant", but when an explicit alt text is provided, this is no longer the case.

**Further discussion**

[Discussion](https://github.com/tarleb/pandoc/discussions/12#discussioncomment-943251)

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -659,17 +659,20 @@ dimensionsToAttrList attr = consolidateStyles $ go Width ++ go Height
 figure :: PandocMonad m
        => WriterOptions -> Attr -> [Inline] -> (Text, Text)
        -> StateT WriterState m Html
-figure opts attr txt (s,tit) = do
+figure opts attr@(_, _, attrList) txt (s,tit) = do
   html5 <- gets stHtml5
   -- Screen-readers will normally read the @alt@ text and the figure; we
   -- want to avoid them reading the same text twice. With HTML5 we can
   -- use aria-hidden for the caption; with HTML4, we use an empty
   -- alt-text instead.
+  -- When the alt text differs from the caption both should be read.
   let alt = if html5 then txt else [Str ""]
   let tocapt = if html5
-                  then H5.figcaption !
-                       H5.customAttribute (textTag "aria-hidden")
-                                          (toValue @Text "true")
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -659,17 +659,19 @@ dimensionsToAttrList attr = consolidateStyles $ go Width ++ go Height
...
+  -- When the alt text differs from the caption both should be read.
   let alt = if html5 then txt else [Str ""]
   let tocapt = if html5
-                  then H5.figcaption !
-                       H5.customAttribute (textTag "aria-hidden")
-                                          (toValue @Text "true")
+                  then (H5.figcaption !) $ if isJust (lookup "alt" attrList)
+                            then mempty
+                            else H5.customAttribute (textTag "aria-hidden")
+                                    (toValue @Text "true")
```
jgm, line 674: This change makes good sense to me! I'd only suggest, for conformity to local style, moving the `if` clause to the next line so that the `then` is lined up with the condition of the `if`.

jgm:
Thanks!

PR #7429, opened 2021-07-06
Recognize data-external when reading HTML img tags

brisad:
If an  in HTML has a `data-external` attribute, keep it as a
native attribute with the `data-` part of the key stripped.  This
information can be used to tell writers such as EPUB to not
incorporate images into the resulting document.

### DIFF
--- a/src/Text/Pandoc/Readers/HTML.hs
+++ b/src/Text/Pandoc/Readers/HTML.hs
@@ -728,11 +728,14 @@ pImage = do
   let alt = fromAttrib "alt" tag
   let uid = fromAttrib "id" tag
   let cls = T.words $ fromAttrib "class" tag
+  let external = case fromAttrib "data-external" tag of
+                   "" -> []
+                   v  -> [("external", v)]
   let getAtt k = case fromAttrib k tag of
                    "" -> []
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -728,11 +728,14 @@ pImage = do
   let alt = fromAttrib "alt" tag
   let uid = fromAttrib "id" tag
   let cls = T.words $ fromAttrib "class" tag
+  let external = case fromAttrib "data-external" tag of
+                   "" -> []
+                   v  -> [("external", v)]
```
jgm, line 733: Instead of handling this one attribute specially, why don't we use `toAttr` (defined in T.P.Readers.HTML.Parsing), as we do for many other tags?  This would carry over all the attributes, not just `external`.
Before calling `toAttr` we'd want to strip out `alt`, `src`, and `title`, since these are handled separately.  We could get rid of the special treatment of kvs on l. 734.

brisad: Yes, that was much better. I've updated the PR with those changes.

jgm:
great, thanks

PR #7433, opened 2021-07-08
DocBook reader: add support for citerefentry

jtojnar:
For referencing UNIX manual pages https://docbook.org/tdg/en/html-ng/citerefentry.html

### DIFF
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -97,7 +97,7 @@ List of all DocBook tags, with [x] indicating implemented,
 [x] chapterinfo - Meta-information for a Chapter
 [ ] citation - An inline bibliographic reference to another published work
 [ ] citebiblioid - A citation of a bibliographic identifier
-[ ] citerefentry - A citation to a reference page
+[x] citerefentry - A citation to a reference page
 [ ] citetitle - The title of a cited work
 [ ] city - The name of a city in an address
 [x] classname - The name of a class, in the object-oriented programming sense
@@ -1112,6 +1112,10 @@ parseInline (Elem e) =
         "segmentedlist" -> segmentedList
         "classname" -> codeWithLang
         "code" -> codeWithLang
+        "citerefentry" -> do
+             let title = maybe mempty strContent $ filterChild (named "refentrytitle") e
+             let manvolnum = maybe mempty (\el -> "(" <> strContent el <> ")") $ filterChild (named "manvolnum") e
+             return $ codeWith ("",["interpreted-text"],[("role","manpage")]) (title <> manvolnum)
         "filename" -> codeWithLang
         "envar" -> codeWithLang
         "literal" -> codeWithLang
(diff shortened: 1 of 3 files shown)


jtojnar:
I could also add support for converting this AST back to DocBook.

And support for rST syntax: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-manpage

jtojnar:
Actually, rST parses as `{"t":"Code","c":[["",["interpreted-text"],[["role","manpage"]]],"ls(1)"]}` so it might make sense to parse DocBook as that too and we would get rST reader/writer support for free.

jgm:
Is this element specific to man pages?  That certainly isn't clear from the documentation.

jtojnar:
As far as I was able to tell, it was originally intended for referring to UNIX manual pages, either part of the same DocBook document as `refentry` element, or external – hence the [`manvolnum`](https://tdg.docbook.org/tdg/5.1/manvolnum.html) element:

> The notion of a volume number is historical. UNIX manual pages (“man pages”), for which refentry was devised, were typically stored in three-ring binders. Each bound manual was a volume in a set and contained information about a particular class of things. For example, volume 1 was for user commands, and volume 8 was for administrator commands.

These days, [`refentry`](https://tdg.docbook.org/tdg/5.1/refentry.html) is more general, for example the linked element documentation pages are each a `refentry`.

As per the _Processing expectations_ section of [`citerefentry`](https://tdg.docbook.org/tdg/5.1/citerefentry.html), the element is supposed to be a hyperlink to a `refentry` (when in the same document) but pandoc does not support `refentry` tag at the moment so that is moot.

----

But I do not see much of a difference between reference and manual pages so it makes sense to me to use `role=manpage`. Especially since this will likely cover the vast majority of uses.

jgm:
Thanks for the info. Somehow it doesn't seem right to me to emit such a rST-specific AST element.  Nothing else in he DocBook reader parses into something with class `interpreted-text`.  I'd rather have something more abstract that can be manipulated. Perhaps
```xml
<citerefentry><refentrytitle>nix.conf</refentrytitle><manvolnum>5</manvolnum></citerefentry>
```
could become
```haskell
Code ("",["citerefentry"],[]) "nix.conf(5)"
```
unless it's important to be able to access the volume number as an attribute.

jtojnar:
Yeah, that’s what I had at the start but I liked being able to unify stuff. But I guess we can keep it distinct and use Lua filter when converting between from DocBook to rST.

```lua
function Code(elem)
  elem.classes = elem.classes:map(function (x)
    if x == 'citerefentry' then
      elem.attributes['role'] = 'manpage'
      return 'interpreted-text'
    else
      return x
    end
  end)

  return elem
end
```

----

I considered keeping the `manvolnum` as an attribute but it should be easy to parse it out if someone needs it.

jtojnar:
It would be nice if pandoc could preserve semantics between formats that support it but yeah, these special classes are not very nice representation.

Opened https://github.com/jgm/pandoc/pull/7437 where it produces `Code` with just `.citerefentry` class.

jtojnar:
Resolved in #7437.

PR #7437, opened 2021-07-09
DocBook reader: add support for citerefentry

jtojnar:
Originally intended for referring to UNIX manual pages, either part of the same DocBook document as [`refentry`](https://tdg.docbook.org/tdg/5.1/refentry.html) element, or external – hence the [`manvolnum`](https://tdg.docbook.org/tdg/5.1/manvolnum.html) element. These days, `refentry` is more general, for example the element documentation pages linked are each a refentry.

As per the *Processing expectations* section of [`citerefentry`](https://tdg.docbook.org/tdg/5.1/citerefentry.html), the element is supposed to be a hyperlink to a refentry (when in the same document) but pandoc does not support refentry tag at the moment so that is moot.

This roughly corresponds to a [`manpage` role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-manpage) in rST syntax, which produces a `Code` AST node with attributes `.interpreted-text role=manpage` but that does not fit DocBook parser.


jgm:
great, thanks!

Issue #4535, opened 2018-04-05
Removing non-Haskell dependencies (C dependencies)


aolney:
Following from [this discussion](https://groups.google.com/d/msgid/pandoc-discuss/89055b69-bdfa-4ca7-aea5-97ca8dece032%40googlegroups.com) there seems to be a value to locating non-Haskell dependencies and replacing them with Haskell.

@jgm has identified the following (comments his):

- [ ] skylighting - depends on pcre - not clear how we can eliminate that without
a pure Haskell regex library that matches pcre's features.

- [ ] yaml - depends on libyaml - there are some proposals in Google Summer
of Code for writing a better pure Haskell yaml parser, so this may be
done in the relatively near future.

- [ ] commonmark/gfm - depends on libcmark - I have been working on a pure
Haskell commonmark parser, which isn't yet published but which could
substitute for libcmark (at the cost of reduced performance).
https://github.com/jgm/commonmark-hs

- [ ] confirm absence of other dependencies using [HackageDB](https://hackage.haskell.org/package/hackage-db-2.0.1/docs/Distribution-Hackage-DB.html) to chase down all of pandoc's transitive dependencies and exhaustively enumerate those that depend on foreign libraries.

jgm:
The following cabal files in pandoc's transitive dependencies contain c-sources stanzas.
In some cases they may only be used for testing or benchmarking.

```
time.cabal:72:    c-sources: lib/cbits/HsTime.c
time.cabal:179:    c-sources: test/unix/Test/Format/FormatStuff.c
aeson.cabal:156:    c-sources: cbits/unescape_string.c
aeson.cabal:166:  c-sources: cbits/unescape_string.c
cryptonite.cabal:240:  C-sources:         cbits/cryptonite_chacha.c
cryptonite.cabal:273:    C-sources:         cbits/decaf/p448/arch_ref64/f_impl.c
cryptonite.cabal:284:    C-sources:         cbits/decaf/p448/arch_32/f_impl.c
cryptonite.cabal:296:    C-sources: cbits/curve25519/curve25519-donna-c64.c
cryptonite.cabal:298:    C-sources: cbits/curve25519/curve25519-donna.c
cryptonite.cabal:315:    c-sources:      cbits/cryptonite_rdrand.c
cryptonite.cabal:321:    C-sources:       cbits/aes/x86ni.c
cryptonite.cabal:326:    C-sources:       cbits/aes/generic.c
[...]

PR #7456, opened 2021-07-24
Stop using the HTTP package.

mt-caret:
Given https://github.com/osener/markup.rocks/issues/21 and https://github.com/jgm/pandoc/issues/4535, I've recently been working on getting pandoc to build with ghcjs, which I've been able to do here: https://github.com/mt-caret/pandoc-ghcjs. Unfortunately, while the build succeeds there seems to be some problems with the JavaScript emitted so it does not work as-is (I'd love some help understanding what's wrong if anybody is interested). In the meantime, there seem to be relatively trivial changes that I can upstream to make pandoc more ghcjs-friendly.

Pandoc only depends on the urlEncode function in the package, which is also provided by http-types. The HTTP package also depends on the network package, which has difficulty building on ghcjs. I've created some benchmarks and fuzzing tests (https://github.com/mt-caret/bench-url-encode), and based on the results there unfortunately http-type's implementation seems to be 50% slower.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -598,7 +598,6 @@ library
                    Text.Pandoc.UTF8,
                    Text.Pandoc.Templates,
                    Text.Pandoc.XML,
-                   Text.Pandoc.Network.HTTP,
                    Text.Pandoc.SelfContained,
                    Text.Pandoc.Highlighting,
                    Text.Pandoc.Logging,
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -709,6 +709,7 @@ library
                    Text.Pandoc.XML.Light.Types,
                    Text.Pandoc.XML.Light.Proc,
                    Text.Pandoc.XML.Light.Output,
+                   Text.Pandoc.Network.HTTP,
                    Text.Pandoc.CSS,
                    Text.Pandoc.CSV,
                    Text.Pandoc.RoffChar,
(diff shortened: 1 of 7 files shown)


jgm:
Even so, avoiding a dependency on HTTP seems a win.

mt-caret:
> Unfortunately this is not the only transitive dependency on network in pandoc (cabal-plan will give you a nice graph).

Yeah, I'm planning to spend more time working on removing the network dependency (or fixing it so it builds on ghcjs). Another issue is how everything depends on zlib; my guess is we need to add a bunch of functionality to https://github.com/GaloisInc/pure-zlib in order to use it as a drop-in replacement for zlib, which looks pretty painful.

> Even so, avoiding a dependency on HTTP seems a win.
Sounds good! Is there anything else I should do before merging?

pandoc.cabal
```diff
@@ -599,6 +598,7 @@ library
                    Text.Pandoc.UTF8,
                    Text.Pandoc.Templates,
                    Text.Pandoc.XML,
+                   Text.Pandoc.Network.HTTP,
```
jgm, line 601: I'd prefer to make this an internal (non-exported) module.

mt-caret: Done.

pandoc.cabal
```diff
@@ -598,7 +598,6 @@ library
                    Text.Pandoc.UTF8,
                    Text.Pandoc.Templates,
                    Text.Pandoc.XML,
-                   Text.Pandoc.Network.HTTP,
```
jgm: Shouldn't this now appear in other-modules?

mt-caret: Sorry about that; fixed.

jgm:
Thanks!

PR #7458, opened 2021-07-25
Added numbersections to description of --number-sections

marekkowalczyk:
Added name of variable 'numbersections' corresponding to option '--number-sections'.


jgm:
Thanks for this.  I'm not going to merge it, because (a) pandoc.1 is a generated file (from MANUAL.txt), and (b) we describe the variable elsewhere, and I think it would clutter things to much if we talked about variables in connection with all the command-line options.

marekkowalczyk:
Your call but as a user I found this mismatch between variable and option names frustrating and it cost me a lot of time to find out the solution.

PR #7461, opened 2021-07-26
Allow --reference-location to affect HTML output

bitonic:
This works similarly to the markdown writer, but with special care
taken to skipping section divs with what regards to the block level.

The change also takes care to not modify the output if `EndOfDocument`
is used.

I implemented this to implement side notes without JavaScript -- see https://mazzo.li/posts/value-speculation.html for an example.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -961,7 +961,7 @@ header when requesting a document from a URL:
 :   Specify whether footnotes (and references, if `reference-links` is
     set) are placed at the end of the current (top-level) block, the
     current section, or the document. The default is
-    `document`. Currently only affects the markdown writer.
+    `document`. Currently only affects the markdown and HTML writers.
 
 `--markdown-headings=setext`|`atx`
 
--- a/man/pandoc.1
+++ b/man/pandoc.1
@@ -993,7 +993,7 @@ Specify whether footnotes (and references, if \f[C]reference-links\f[R]
 is set) are placed at the end of the current (top-level) block, the
 current section, or the document.
 The default is \f[C]document\f[R].
-Currently only affects the markdown writer.
+Currently only affects the markdown and HTML writers.
 .TP
 \f[B]\f[CB]--markdown-headings=setext\f[B]\f[R]|\f[B]\f[CB]atx\f[B]\f[R]
 Specify whether to use ATX-style (\f[C]#\f[R]-prefixed) or Setext-style
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -74,9 +74,11 @@ import Text.TeXMath
 import Text.XML.Light (elChildren, unode, unqual)
 import qualified Text.XML.Light as XML
 import Text.XML.Light.Output
+import Data.String (fromString)
 
 data WriterState = WriterState
     { stNotes        :: [Html]  -- ^ List of notes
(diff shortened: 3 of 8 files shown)


jgm:
Current docs for this option:

`--reference-location=block`|`section`|`document`              
                                                                
:   Specify whether footnotes (and references, if `reference-links` is
    set) are placed at the end of the current (top-level) block, the
    current section, or the document. The default is
    `document`. Currently only affects the markdown writer.

The proposal is to allow this to affect footnotes in the HTML writer as well.

bitonic:
@jgm I've updated the manual. Let me know if anything else is needed.

jgm:
This looks good to me.

Question:

- What happens when you use `--section-divs` with notes at end of section (or block)?  Do the notes get inside the section div or do they come after it? Which behavior makes more sense? There is no test of this behavior.
- That's especially important because in EPUB and some slide show outputs we implicitly use section divs.  At least some informal testing should be done to see how this feature affects HTML slide show formats (revealjs, slidy) and epub.
- Probably it should be documented in the manual that it affects these formats, too (assuming it does).

bitonic:
@jgm they would be outside the section block (right after the section). I do not know what makes more sense, but if you advise I'll change the behavior (if different from what's already implemented) and add the tests.

jgm:
I think either of these structures could make sense:
```
<section>
contents
<div>footnotes<ldiv>
</section>
```
or
```
<section>
contents
</section>
<div>
footnotes
</div>
```
I'm not sure which is better.  In any case this should be a tested case.

bitonic:
@jgm I've added the test.

jgm:
Great.  Can you explain why it is important to add the extra class e.g. `footnote-end-of-document`?

Also, have you tested, at least informally, with epubs and HTML slide shows (revealjs)?
I just want to make sure nothing strange happens with these cases.

bitonic:
@jgm My use case for those classes is that I want to style sidenotes (like in this article https://mazzo.li/posts/vectorized-atan2.html ) differently from footnotes (like in this article https://mazzo.li/posts/haskell-backprop-short.html ). I thought that class names would be the cleanest way to do that.

I have not tested epub or revealjs. If you tell me how to test (e.g. "modify this file in this way and check XYZ") I'll follow up.

jgm:
Just use `--reference-location=block` (or `section`) when you produce an epub.  E.g. `pandoc MANUAL.txt -o m.epub --reference-location=block`.  Then see how it looks in an epub reader, if you have one available.

Slide shows:  create a markdown file with a few sections and some footnotes, then `-t revealjs -s -o slides.html --reference-location=block` etc.

bitonic:
I tried it and it seems to work fine. Here's the ebook with `--reference-location=block`:

image

It's similar with `--reference-location=section`. And with the slides:

image

jgm:
Have you tried slides with `section`?  I'm particularly interested to see how that works.

bitonic:
I have, yes. For ebooks, it's at the end of the section, with the `<hr>`:

image


And this is with the slides -- it's in its own slide, much like if it were at the end of the document, but per-section:

image

I can think of several things that could be configurable:

1. Resetting the footnote number at each section / block. This is probably the best default for books actually, but not for blog posts (my case).
2. When the `<hr>` is placed.

But I think going forward with this is still an improvement.

jgm:
Agreed, thanks!

PR #7466, opened 2021-07-30
LaTeX table writer: Increase column width precision

peterfab9845:
In cases where a table row's content is longer than the column width, the table is scaled to the text width.
However, only two decimal places are used for each column's size.
In some cases, with many columns, the cumulative rounding can result in a noticeable overrun out of the text area (shown here with the `showframe` package):
image

(This particular case can be fixed with a larger columns setting, because `$\alpha$` isn't as wide as it seems, but the overall issue remains.)

This PR simply adds two extra decimal places to the column widths, minimizing the effect:
image

Reproduction:
[test.md](https://github.com/peterfab9845/pandoc/files/6904737/test.md)
`pandoc test.md -o test.pdf`

Version:
```
pandoc 2.14.0.2
Compiled with pandoc-types 1.22, texmath 0.12.3, skylighting 0.10.5.1,
citeproc 0.4.0.1, ipynb 0.1.0.1
```


jgm:
Thanks!

PR #7490, opened 2021-08-11
Pptx writer: Select slide layouts by names

undergroundquizscene:
Currently, slide layouts are chosen from the reference doc according to order. Powerpoint names the layout files according to their order in its visual editor, with file names `slideLayout1.xml`, `slideLayout2.xml`, etc. In pandoc, slides of type `MetadataSlide` (which powerpoint’s standard template calls the “Title Slide”) always use `slideLayout1.xml`, and similar choices are made for other types of slide.

With this PR I propose choosing the slide layouts by their "name" XML attribute. That is, look at all slide layouts in the reference doc, and choose the first layout with a name that case-insensitively matches an expected title for each type of slide. For example, based on the current reference doc (and powerpoint's default layouts), for the `MetadataSlide` use the first layout with a name like "Title Slide". If any name isn't found in the provided reference doc, fall back to pandoc's default pptx reference doc (and issue a warning). If the name isn't present there either, error out.

This PR is a work in progress, which I'm opening early to get feedback on the approach. I plan to make the following additional changes before merging:

- Read more layouts, either layouts matching the `slideLayoutn.xml` pattern or by reading the list of them from the slideMaster files
- Write more helpful and comprehensive warning/error messages
- Document the new way of choosing layouts in the manual

While working on this, I've found it helpful to have haskell-language-server available in the nix-shell: is it ok for me to commit that change?


jgm:
It would be good to get feedback from @jkr who wrote the pptx writer.

undergroundquizscene:
> - Read more layouts, either layouts matching the slideLayoutn.xml pattern or by reading the list of them from the slideMaster files

This is now done: I've used the pattern, following the existing examples in the code.

jgm:
> While working on this, I've found it helpful to have haskell-language-server available in the nix-shell: is it ok for me to commit that change?

Yes, that's absolutely fine, but it should be a separate PR.

undergroundquizscene:
> > While working on this, I've found it helpful to have haskell-language-server available in the nix-shell: is it ok for me to commit that change?
> 
> Yes, that's absolutely fine, but it should be a separate PR.

Perfect, I'll make a separate PR now

jgm:
> With this PR I propose choosing the slide layouts by their "name" XML attribute. That is, look at all slide layouts in the reference doc, and choose the first layout with a name that case-insensitively matches an expected title for each type of slide. For example, based on the current reference doc (and powerpoint's default layouts), for the MetadataSlide use the first layout with a name like "Title Slide". If any name isn't found in the provided reference doc, fall back to pandoc's default pptx reference doc (and issue a warning). If the name isn't present there either, error out.

I'd like to understand this better. (Probably the reason I'm not is that I don't use powerpoint.)  Is this all correct?

- the reference doc contains a number of "slide layouts," currently numbered numerically
- the layouts are matched up to slides in order, with slide 1 using layout 1, slide 2 layout 2, etc. [Is this right? what if there aren't enough layouts defined?]
- you propose to keep these numbered slide layouts, but allow the possibility of adding named slide layouts to the reference doc [is that right?]
- pandoc would try to associate a slide with a layout with a matching name, falling back to the old method (with a warning?] if this fails

What counts as a matching name?  Are you matching the slide title?  Exact match or something else?

undergroundquizscene:
> I'd like to understand this better. (Probably the reason I'm not is that I don't use powerpoint.) Is this all correct?
> 
>     * the reference doc contains a number of "slide layouts," currently numbered numerically
> 
>     * the layouts are matched up to slides in order, with slide 1 using layout 1, slide 2 layout 2, etc. [Is this right? what if there aren't enough layouts defined?]
> 
>     * you propose to keep these numbered slide layouts, but allow the possibility of adding named slide layouts to the reference doc [is that right?]
> 
>     * pandoc would try to associate a slide with a layout with a matching name, falling back to the old method (with a warning?] if this fails
> 
> 
> What counts as a matching name? Are you matching the slide title? Exact match or something else?

Not quite all correct: I'll explain.

- The reference doc contains a number of slide layouts, numbered numerically.
- Each slide uses one of these layouts, chosen based on the type of slide it is. Currently:
  - The "Title Slide" (as PowerPoint calls it in the default Office template), which comes at the start of the presentation and shows metadata like the name of the presentation and its author, uses the layout `ppt/slideLayouts/slideLayout1.xml`.
  - "Section Headers", which contain only a (section) title, use the layout `ppt/slideLayouts/slideLayout3.xml`
  - "Title and Content" slides, which contain a heading and some body content and are generated by pandoc for most normal slides, use the layout `ppt/slideLayouts/slideLayout2.xml`
  - "Two Content" slides, which have two columns of content, use the layout `ppt/slideLayouts/slideLayout4.xml`.
- I propose selecting the layouts based on their name, so that for the title slide pandoc will look at all layouts in the reference doc, and choose the first one whose name (read from an XML attribute) matches "Title Slide". (And similarly for each other type of slide.)
  - At the moment the match is just a case-insensitive match, but we can definitely expand that if it's desirable
- If no layout is found with the right name, pandoc will issue a warning and take the layout with that name from the default reference doc.
  - The names I've used are the ones that the layouts in the default reference doc currently have, so those won't need to be changed

Is that explanation clearer?

jgm:
Yes, that helps (and so did reading the code).

If I understand correctly, the point of this is just to make it easier for authors of a reference.docx to see which layout goes with which slides.

At any event, some documentation should be added in the manual, so people will know what to do.

undergroundquizscene:
> If I understand correctly, the point of this is just to make it easier for authors of a reference.docx to see which layout goes with which slides.

That's correct, yes. (Except that it's a reference.pptx, of course)

> At any event, some documentation should be added in the manual, so people will know what to do.

Agreed, I'll add some.

src/Text/Pandoc/Logging.hs
```diff
@@ -76,6 +76,7 @@ data LogMessage =
   | InlineNotRendered Inline
   | BlockNotRendered Block
   | DocxParserWarning Text
+  | PptxParserWarning Text
```
jgm, line 79: Since this is an API change, make sure your commit message clearly indicates this (if it doesn't already), so I can document it in the changelog.

src/Text/Pandoc/Writers/Powerpoint/Output.hs
```diff
@@ -318,6 +359,40 @@ presentationToArchive opts meta pres = do
...
+            throwError (PandocSomeError ("Couldn't find layout named \""
+                                         <> layoutTitle <> "\" in the provided "
+                                         <> "reference doc or in the default "
+                                         <> "reference doc included with pandoc."))
+          (Nothing, Just ((element, path, entry) :| _)) -> do
+            P.logOutput (PptxParserWarning ("Couldn't find layout named \""
+                                           <> layoutTitle <> "\" in provided "
+                                           <> "reference doc. Falling back to "
+                                           <> "the default included with pandoc."))
+            pure SlideLayout { slElement = element
```
jgm, line 383: Should these both be PptxParserWarnings? (One is PandocSomeError)

src/Text/Pandoc/Logging.hs
undergroundquizscene: Ok, will do.

src/Text/Pandoc/Writers/Powerpoint/Output.hs
undergroundquizscene: No, I've only used a warning when we have a default we can fall back to. If a named layout isn't found in either the provided reference doc or the default reference doc then we can't do the conversion and have to throw an error. (Since the default reference doc does contain layouts with these names, I'd only expect this to happen if it was changed or moved/removed.)

I guess at this point we could fall back to selecting one of the first 4 layouts (the behaviour before this change)? That would give a successful output if the layouts have been renamed in the provided reference doc but left in the right order, and the default reference doc is missing/incorrect.

undergroundquizscene:
@jgm Do you know what the [`cdLine`](https://hackage.haskell.org/package/xml-1.3.14/docs/Text-XML-Light-Types.html#v:cdLine) attribute in Text.XML.Light's CData type is? I'm getting test failures reporting that it's changed, but I'm not sure why and I don't know what it actually means. The XML looks the same to me, as far as I can see.

I'm inclined to ignore it in `compareXML`, but not knowing what it is gives me pause.

src/Text/Pandoc/Writers/Powerpoint/Output.hs
jgm: OK, I understand now.  I don't know about the fallback question.  Probably simplest just to have an error as you do here.

jgm:
`cdLine`:  I think it's to track line numbers, but I'm not completely sure.

note:  in most of the parsers that use XML, we've moved from using xml (Text.XML.Light) to using Text.Pandoc.XML.Light.  This reproduces the Text.XML.Light types, but with Text instead of String.  The Text.Pandoc.XML.Light module also has replacement's for the xml package's parsers; these replacements use xml-conduit under the hood (faster and more accurate) but they copy the old xml API, so we've been able to use them as drop-in replacements.

It looks like there are still a few places in the code where Text.XML.Light is used directly.  It might be worth swapping these out, unless it causes problems.

undergroundquizscene:
> `cdLine`: I think it's to track line numbers, but I'm not completely sure.

Ok, I think I'll stop checking it in the comparison. I'll try accepting the tests first, and take a closer look at the resulting changes in the XML (rather than relying on the test's diff output): that may give me more of a hint.

> It looks like there are still a few places in the code where Text.XML.Light is used directly. It might be worth swapping these out, unless it causes problems. It would be nice, eventually, to remove the dependency on the `xml` package.

Sure, I'll have a go at swapping it out, I'm sure it'll be straightforward.

undergroundquizscene:
> > It looks like there are still a few places in the code where Text.XML.Light is used directly. It might be worth swapping these out, unless it causes problems. It would be nice, eventually, to remove the dependency on the `xml` package.
> 
> Sure, I'll have a go at swapping it out, I'm sure it'll be straightforward.

Ah, not entirely straightforward: I've made the changes, but the module `Text.Pandoc.XML.Light` isn't exposed by pandoc, and I'm attempting to import it into the test code. I don't know cabal super well, is there a way to expose the module just to the test target? (If not, will I just duplicate that module in the test code?)

undergroundquizscene:
> > `cdLine`: I think it's to track line numbers, but I'm not completely sure.
> 
> Ok, I think I'll stop checking it in the comparison. I'll try accepting the tests first, and take a closer look at the resulting changes in the XML (rather than relying on the test's diff output): that may give me more of a hint.

Ah, I've found the issue: the slide master and its relationships file in `data` both have a doctype one line 1 and the rest of the xml on line 2. Until now, we've copied those across to the output archive without modifying them. Now that I'm modifying them, the doctype and rest of the xml are both on line 1. Since we don't output the doctype in the diff output, it's hard to see when that's happened. I'll do a little more testing to confirm that nothing else has changed, and then accept the changes.

jgm:
> I don't know cabal super well, is there a way to expose the module just to the test target? (If not, will I just duplicate that module in the test code?)

No, there isn't.  And duplicating the module is not a good solution.  This may be why I didn't change these when adding T.P.XML.Light!

undergroundquizscene:
> > > `cdLine`: I think it's to track line numbers, but I'm not completely sure.
> > 
> > 
> > Ok, I think I'll stop checking it in the comparison. I'll try accepting the tests first, and take a closer look at the resulting changes in the XML (rather than relying on the test's diff output): that may give me more of a hint.
> 
> Ah, I've found the issue: the slide master and its relationships file in `data` both have a doctype one line 1 and the rest of the xml on line 2. Until now, we've copied those across to the output archive without modifying them. Now that I'm modifying them, the doctype and rest of the xml are both on line 1. Since we don't output the doctype in the diff output, it's hard to see when that's happened. I'll do a little more testing to confirm that nothing else has changed, and then accept the changes.

I've taken a slightly different tack in the end: In the diff testing, I've stopped checking that the `cdLine` values are the same. If everything else in the document is the same except these values, I think they can always be parsed as the same XML document, so I don't think we need to require them to be the same.

undergroundquizscene:
> No, there isn't. And duplicating the module is not a good solution. This may be why I didn't change these when adding T.P.XML.Light!

Ah yeah, that's fine: I'll leave things as they are then.

jgm:
OK, I think this is good to go, then.  Thanks!

undergroundquizscene:
> OK, I think this is good to go, then. Thanks!

Great, thanks very much!

PR #7505, opened 2021-08-17
Escape backslashes in haddock

undergroundquizscene:
Any literal backslash needs to be escaped: these are currently showing
up [in hackage](https://hackage.haskell.org/package/pandoc-2.14.1/docs/Text-Pandoc-UTF8.html#v:toTextLazy) as “‘r’” instead of “‘\r’”.


jgm:
Thanks, didn't know!

PR #7510, opened 2021-08-19
Clarify that each YAML block is a separate YAML document

wlupton:
See the ["Each YAML block is a YAML document?" pandoc-discuss thread](https://groups.google.com/g/pandoc-discuss/c/icIxqbKptDc/m/q2dTCNpAAQAJ) for background.

I am proposing just adding a new paragraph. I did wonder whether "YAML object" could also be changed to "YAML document" (I think it occurs twice) but decided not to propose this in the PR.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -4179,6 +4179,10 @@ A document may contain multiple metadata blocks.  If two
 metadata blocks attempt to set the same field, the value from
 the second block will be taken.
 
+Each metadata block is handled internally as an independent YAML document.
+This means, for example, that any YAML anchors defined in a block cannot be
+referenced in another block.
+
 When pandoc is used with `-t markdown` to create a Markdown document,
 a YAML metadata block will be produced only if the `-s/--standalone`
 option is used.  All of the metadata will appear in a single block


MANUAL.txt
```diff
@@ -4179,6 +4179,10 @@ A document may contain multiple metadata blocks.  If two
 metadata blocks attempt to set the same field, the value from
 the second block will be taken.
 
+Each metadata block is an independent YAML document.  This means, for
```
alerque, line 4182: This may or may not be relevant to most Pandoc users, but as a programmer using YAML in other contexts I'm used to its features that are explicitly designed to combine other parts of other documents. What this paragraph describes is not a YAML limitation, it is a scoping limitation in Pandoc itself. I would suggest tweaking the wording here a little bit to announce that this is a limitation imposed by Pandoc's YAML handling. Maybe it will save some like myself who might expect more.

```suggestion
Each metadata block is handled internally as an independent YAML document.  This means, for
```

wlupton: I made this change (sorry, I didn't notice the above form until after I'd pushed it). Thanks!

alerque (approved):

PR #7514, opened 2021-08-20
Adjust `fenced_divs` syntax to resemble Markdown directive syntax

### LINKED ISSUE #7480: Make `fenced_divs` more consistent with Markdown Directive Syntax
### Pandoc's `fenced_div` Extension

Pandoc's [`fenced_div` extension](https://pandoc.org/MANUAL.html#divs-and-spans) currently supports the following syntax for creating divs:

Example 1 ([try pandoc](https://pandoc.org/try/?text=%3A%3A%3A%3A%3A+%7B%23special+.sidebar+key%3Dvalue%7D%0AHere+is+a+paragraph.%0A%0AAnd+another.%0A%3A%3A%3A%3A%3A&from=markdown&to=html5&standalone=0))

```
::::: {#special .sidebar key=value}
Here is a paragraph.

And another.
:::::
```

```
<div id="special" class="sidebar" data-key="value">
<p>Here is a paragraph.</p>
<p>And another.</p>
</div>
```

Example 2 ([try pandoc](https://pandoc.org/try/?text=%3A%3A%3A+Warning+%3A%3A%3A%3A%3A%3A%0AThis+is+a+warning.%0A%0A%3A%3A%3A+Danger%0AThis+is+a+warning+within+a+warning.%0A%3A%3A%3A%0A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A%3A&from=markdown&to=html5&standalone=0))
```
::: Warning ::::::
This is a warning.

::: Danger
This is a warning within a warning.
:::
::::::::::::::::::
```
```
<div class="Warning">
<p>This is a warning.</p>
<div class="Danger">
<p>This is a warning within a warning.</p>
</div>
</div>
```

### Limitations of `fenced_div`

Strangely, Pandoc does not allow you to combine the two `fenced_div` syntax styles:

```
:::: container {.theorem #id key=value} ::::::
This is a theorem.
::::
```

```
<p>:::: container {.theorem #mytheorem key=value} :::::: This is a theorem. ::::</p>
```

I think this is a perfectly natural thing to do.  Given the current behavior, I would expect the output to be something like this:

```
<div id="mytheorem" class="container theorem" data-key="value">
<p>This is a theorem.</p>
</div>
```
 
### Markdown Directive Syntax

Some markdown implementations support the following [de-facto standard for directive syntax](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444), with syntax for inline, leaf, and block directives.  Perhaps in a future proposal it would be desirable to add these to the Pandoc AST for further processing, but for now I would be satisfied if the syntax for the `fenced_div` extension were updated to more closely align with the directive syntax.

Inline Directive 
```
:name[content]{#id .class1 .class2 key=val}
```

Leaf Directive
```
:: name [content] {key=val}
```

Block Directive 
```
::: name [inline-content] {key=val}
contents, which are sometimes further block elements
:::
```

### Alternatives

* Instead of modifying the behavior of `fenced_divs`, create a separate extension `directive_divs`
* Incorporate Markdown directive syntax into the pandoc AST instead of converting to divs (this would be a much larger change)
* Re-write any directive syntax in my markdown documents to use the format expected by pandoc (quite undesirable)

### References

Directive Syntax

* Talk.CommonMark, ["Generic directives/plugins syntax"](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444)
* RemarkJS, [`remark-directive` plugin](https://github.com/remarkjs/remark-directive)
* markdown-it, [`markdown-it-directive` plugin](https://github.com/hilookas/markdown-it-directive)

Possibly Related Issues

* #4037

benrbray:
Resolves #7480.  Adds support for blocks of the form

```
:::: container {.theorem #id key=value} ::::::
This is a theorem.
::::
```
```html
<div id="id" class="container theorem" data-key="value">
<p>This is a theorem.</p>
</div>
```

Previously, you could specify only the attributes list XOR the name, but now you can do one, the other, or both.   As discussed in the issue, this change was made to more closely resemble Markdown's [de-facto standard for directive syntax](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444).  While this change does not fully support directive syntax, it makes documents which use directive syntax more compatible with pandoc.

Note that Markdown directives also allow `[inline content]` like below, but this is currently unsupported:

```
(unsupported)
::: name [inline-content] {key=val}
contents, which are sometimes further block elements
:::
```

For now, I see two options:

* **Option 1:**  Don't support `[inline content]`.  I think this omission would be fine, since pandoc always creates divs from fenced sections, so it's not clear where the text should go anyway.
* **Option 2:**  Parse `[inline content]` but throw it away.  This would ensure that pandoc is compatible with documents that use block directive syntax.

If we go with Option 1, this PR is ready to merge.  If we go with Option 2, I would just need to make a small change.  

Full support for directive syntax will, I think, require an AST change, which would be a much larger project.  I think this PR is a good compromise until then!

### DIFF
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -2063,7 +2063,11 @@ divFenced = do
     string ":::"
     skipMany (char ':')
     skipMany spaceChar
-    attribs <- attributes <|> ((\x -> ("",[x],[])) <$> many1Char nonspaceChar)
+    attribs <- attributes <|> do
+      name <- pure <$> many1Char nonspaceChar
+      skipMany spaceChar
+      (ident, classes, keyval) <- option nullAttr attributes
+      return (ident, name ++ classes, keyval)
     skipMany spaceChar
     skipMany (char ':')
     blankline
--- a/test/command/7480.md
+++ b/test/command/7480.md
@@ -1,0 +1,10 @@
+```
+% pandoc
+:::: container {.theorem #id key=value} ::::::
+content
+::::
+^D
+<div id="id" class="container theorem" data-key="value">
+<p>content</p>
+</div>
+```


benrbray:
Note that the [original implementation](https://github.com/jgm/pandoc/issues/7480#issuecomment-895885505) I posted in the issue thread was buggy and didn't handle cases like `::: {.cls #id}` correctly.  So, the code here is slightly different.

benrbray:
@jgm is there anything holding back this PR?  No rush, just wondering if this feature needs any changes or further discussion.  Thanks!

benrbray:
@jgm Gentle reminder about this PR.  Thanks!

jgm:
Still considering it.

Sporarum:
Hey, as a new pandoc user, I was really surprised it didn't work, as it does work for code blocks

I this still being considered ?

pchampin:
> Hey, as a new pandoc user, I was really surprised it didn't work, as it does work for code blocks

ditto!

PR #7515, opened 2021-08-20
LaTeX-parser: restrict \endinput to current file

noctux:
In LaTex, `\endinput` only concerns the handling of the file containing the directive.
So if an `\endinput` is encountered within an included file, the remaining content in the parent file is still relevant and valid.
Unfortunatly, pandoc does not respect this scoping of `\endinput`.
This patch attempts to implement the correct behaviour.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -242,6 +242,7 @@ extra-source-files:
                  test/command/sub-file-chapter-1.tex
                  test/command/sub-file-chapter-2.tex
                  test/command/bar.tex
+                 test/command/bar-endinput.tex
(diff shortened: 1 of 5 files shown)


jgm:
Thanks for this!
The linter found something you need to put in extra-source-files in pandoc.cabal.

noctux:
Thanks for the review (and pandoc in general. It's incredibly useful).

Ah, sorry, good catch. I believe the `.md` should already be covered by the glob for `test/command/*.md`, so this should only really concern `bar-input.tex`. Hopefully fixed now.

jgm:
great, thanks

PR #7517, opened 2021-08-20
pptx: Make first heading title if slide level is 0

undergroundquizscene:
# Summary

At the moment, the pptx writer adds a slide break before any table, “columns” div, or paragraph starting with an image, unless the only thing before it on the same slide is a heading at the slide level. In that case, the item and heading are kept on the same slide, and the heading is used as the slide title (inserted into the layout’s “title” placeholder).

However, if the slide level is set to 0 (as was recently enabled) this makes it impossible to have a slide with a title which contains any of those items in its body.

This PR proposes changing this behaviour: now if the slide level is 0, then items will be kept with a heading of any level, if the heading’s the only thing before the item on the same slide.

# Screenshots

Here are some screenshots showing the result of the changes.

## H1 with Image

Firstly, on the following markdown document:

```markdown
# Hello

![An image](./Image.png)
```

- Currently, a slide level of 0 produces two slides, with the heading on the first and the image on the second. The heading is also not used as the slide title.

  H1 and Image (slide level 0, current)

- After these changes, a slide level of 0 produces only one slide, with the heading as the title and the image as the content.

  H1 and Image (slide level 0, new)

## H1 with Table

Secondly, the same results with a table:

```markdown
# Hello

+---+
| 1 |
+---+
```

- Current:
  H1 and Table (slide level 0, current)
- New:
  H1 and Table (slide level 0, new)

## H1 and H2 with Table (Possible downside to changes)

Finally, the one possible downside I've found:

```markdown
# Hello

## There

+---+
| 1 |
+---+
```

With two headings followed by a table (or image or columns div), the current version of pandoc doesn't use either heading for a title, but with the changes we lose the content of the second heading.

- Current:
  H1, H2, and Table (slide level 0, current)
- New:
  H1, H2, and Table (slide level 0, new)


jgm:
Losing content seems like a bad drawback!

undergroundquizscene:
> Losing content seems like a bad drawback!

Yes, that's fair 😅 I will look at how to prevent loss of content on Monday and see what I can do.

undergroundquizscene:
@jgm Ok, I've fixed the lost content – the new code now gives the following output for the last case:

H1, H2, and Table (slide level 0, new)

I've also added a second commit, that I wanted to check with you before squashing into the first. With the first commit, if the first thing on a slide is a heading, and either the heading is at the slide level, or the slide level is 0, the heading will be used as the title of the slide.

I wondered whether we might change that? The second commit makes it so that, if the first thing on a slide is a heading which is **at or below** the slide level, then that heading is used as the slide title (regardless of the slide level value). To my mind, this makes pandoc slightly more forgiving: if making slide breaks with horizontal rules, you don't need to make sure your slide headings have the right level.

However, it makes it slightly more awkward to make a slide with an empty title and a heading in the body. Where at the moment you can do this (with slide level set to 1):

```markdown
# Slide 1

Some content

---

## Not a slide title

The above heading is in the slide body.
```

With the change I'm proposing you would have to add an extra empty heading:

```markdown
# Slide 1

Some content

---

#

## Not a slide title

The above heading is in the slide body.
```

What do you reckon?

jgm:
I don't think I'm in favor of the suggested change.  Better to give authors the flexibility -- sometimes they might want a lower-level heading that isn't a slide title -- even if this also means they have more responsibility.

undergroundquizscene:
> I don't think I'm in favor of the suggested change. Better to give authors the flexibility -- sometimes they might want a lower-level heading that isn't a slide title -- even if this also means they have more responsibility.

Ok, that sounds good – I'll remove the second commit.

jgm:
Sorry, I pushed a broken test. Please rebase and CI should pass.

undergroundquizscene:
> Sorry, I pushed a broken test. Please rebase and CI should pass.

Thanks, will do now

Issue #5097, opened 2018-11-24
Add support for blank slides with powerpoint output.


W1M0R:
### Original (24 November 2018)

There is currently no way to specify an intentional blank slide for powerpoint output.

For blank slides with reveal.js, I can specify markdown input like this:

```markdown
# [Blank Slide]{.hidden} {data-background-color="black"}

[This slide is intentionally left blank. It contains notes about the next slide. Apply the blank slide layout.]{.hidden}
```

`.hidden` is my own CSS style that applies `display: none`. This allows headings and content to still display (for information purposes) in non-slide output formats (e.g. pure html or pdf).

I would like to have the powerpoint output to also know that the slide should be blank. I have a reference.pptx file with the blank layout in the slide master as I would like to have it (e.g. no text with a completely black background), but I don't have a way to specify that the 'blank' layout should apply to the slide.

Blank pages are useful: https://en.wikipedia.org/wiki/Intentionally_blank_page

Perhaps YAML frontmatter can be used to specify the slide layout for powerpoint output specific? It may even be made more general to apply to all slide output formats, but a powerpoint-only solution would satisfy my requirements.

### Edit (18 December 2018)

Just some clarifications about what I mean.

It is possible to create 'blank content' slides for powerpoint output using markdown input. But the powerpoint layout used for the blank slide is the same layout as for the non-blank slides. Powerpoint master slide layouts support a variety of predefined layouts, that can each be styled separately. I think Pandoc supports four of those predefined layouts, and applies them automatically based on the content of the slide. One of the predefined powerpoint layouts, is called 'Blank' or 'None'. I would like this predefined Blank layout to be applied to slides that have no content. If I edit the reference document or template (let's say reference.pptx), and change the properties of the Blank layout (let's say to use a black background instead of the default white), then the blank slides in my pandoc-generated pptx file, should have a black background.

If this can't be achieved using the predefined Blank layout, then perhaps it should be possible to style a powerpoint slide (i.e. set its background color) using Pandoc classes or some other method.

jkr:
I see the need for this. But it's tricky because we have to remove empty slides that are produced by the way slides are split (see cdbe45e8).

So we need some way to indicate an intentional blank slide. Perhaps two consecutive hrules would be a good way to signify this from the markdown side? @jgm: what do you think?

jkr:
It doesn't work at the moment. The issue was that the process of making
slides requires some fairly opinionated splitting (images have to create
[...]

PR #7522, opened 2021-08-24
pptx: WIP: Support more layouts

undergroundquizscene:
At the moment, the pptx writer only supports four slide layouts: “Title Slide” (used for the automatically generated metadata slide), “Section Header” (used for headings above the slide level), “Two Column” (used when there’s a columns div containing at least two column divs), and “Title and Content” (used for all other slides).

With this PR, I'd like to propose adding support for some additional layouts:
- Comparison
  + This would be used on any two-column slide where at least one column includes text followed by non-text (an image or a table), with the text in each column added to the “body” placeholder for that column and the non-text added to the ObjType placeholder.
- Blank
  + This would be used on any input slide that’s blank.
- Content with Caption
  + This would be used on any non-two-column slide that contains text and non-text, with the non-text content added to the ObjType placeholder and the text added to the “body” placeholder.

These are all layouts which exist within the default reference doc, so if the user provides a reference doc without these layouts, we’ll be able to fall back to the default reference doc.

I’ve only implemented support for Comparison so far (no tests yet, though I’ve tried it out and it works), but wanted to open the PR now to start the conversation.


src/Text/Pandoc/Writers/Powerpoint/Output.hs
```diff
@@ -173,6 +174,10 @@ getSlideLayouts = asks envSlideLayouts >>= maybe (throwError e) pure
 data ContentType = NormalContent
                  | TwoColumnLeftContent
                  | TwoColumnRightContent
+                 | ComparisonUpperLeftContent
+                 | ComparisonUpperRightContent
+                 | ComparisonLowerLeftContent
+                 | ComparisonLowerRightContent
```
undergroundquizscene, line 180: I'm not mad about this, because basically I'll need a new constructor for each placeholder in every supported layout. Maybe a better choice would be to specify the placeholder type and index here?

```
newtype ContentType 
  = ContentType PHType Int
  -- ^ The placeholder type to use for the content, 
  -- and the index of the placeholder to use.
```

jgm:
Please rebase on master, there was a broken test.

src/Text/Pandoc/Writers/Powerpoint/Output.hs
undergroundquizscene: Ok, I think I will – thanks

undergroundquizscene:
With the blank layout I've added, this PR would fix #5097

MANUAL.txt
```diff
@@ -1179,9 +1179,31 @@ header when requesting a document from a URL:
...
+        - Content with Caption
+          - This layout is used for any non-two-column slides which contain text
+            followed by non-text (e.g. an image or a table).
+        - Blank
+          - This layout is used for any slides which only contain blank content,
+            e.g. a slide containing only speaker notes, or a slide containing
+            only a non-breaking space.
+        - Title and Content
+          - This layout is used for all slides which do not match the criteria
+            for another layout.
```
undergroundquizscene, line 1206: Is this the best part of the manual to describe this layout-choosing logic in? It's relevant to using a reference doc, but it's also used with the default reference doc when the user doesn't specify their own. It feels like it would fit in the "structuring the slide show" section, perhaps in a new subheading?

jgm: I agree, this makes sense under a new subheading. (It should be made clear that this part applies only to pptx.)  The section on reference docs could then have a link to that subheading.

Note: instead of using sublists with just one element, it would be better to use definition lists.

undergroundquizscene: > I agree, this makes sense under a new subheading. (It should be made clear that this part applies only to pptx.) The section on reference docs could then have a link to that subheading.

Perfect, I'll do that.

> Note: instead of using sublists with just one element, it would be better to use definition lists.

Ah, of course! Thanks

undergroundquizscene:
(Note: I've kept the commits separate for now, but once this is approved I'd like to squash a number of them together before merging)

jgm:
Let me know when you consider this ready.

undergroundquizscene:
> Let me know when you consider this ready.

I've made all the content changes I want to make: except for squashing some commits it's ready. Will I go ahead and squash the commits? (i.e. Are there any other changes you would like me to make?)

jgm:
I think this looks good. Go ahead and rebase.

undergroundquizscene:
Ok, good to go :)

PR #7532, opened 2021-08-30
Do not leak working directory in TikZ filter

jeroendehaas:
The TikZ filter in `lua-filters.md` leaks the working directory into its output. I noticed this when inspecting Jupyter Notebooks generated via Pandoc. This pull request attempts to address this by leaving out the working directory in the target file name.


jgm:
Thanks!

PR #7538, opened 2021-09-02
Add --ipynb-output options html, latex

ickc:
The effect is basically similar to `best`, but specifying the format would be html/latex manually.

This is useful for example when pandoc is called in multiple pipes, say from markdown to json, then some filtering, then from json to html.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1282,14 +1282,18 @@ header when requesting a document from a URL:
     the EPUB-specific contents.  The default is `EPUB`.  To put
     the EPUB contents in the top level, use an empty string.
 
-`--ipynb-output=all|none|best`
+`--ipynb-output=all|none|best|html|latex`
 
 :   Determines how ipynb output cells are treated. `all` means
     that all of the data formats included in the original are
     preserved.  `none` means that the contents of data cells
     are omitted.  `best` causes pandoc to try to pick the
     richest data block in each output cell that is compatible
-    with the output format.  The default is `best`.
+    with the output format. `html|latex` is similar to `best`,
+    but manually choose the one best for `html|latex` respectively,
+    which can be useful when multiple pipes are used so that the
+    output format is not immediately known to pandoc.
+    The default is `best`.
 
 `--pdf-engine=`*PROGRAM*
 
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -237,6 +237,10 @@ convertWithOpts opts = do
                        _ | readerNameBase /= "ipynb" -> id
                        IpynbOutputAll  -> id
                        IpynbOutputNone -> (filterIpynbOutput Nothing :)
+                       IpynbOutputHTML -> (filterIpynbOutput (Just $
+                                     Format "html") :)
+                       IpynbOutputLaTeX -> (filterIpynbOutput (Just $
+                                     Format "latex") :)
                        IpynbOutputBest -> (filterIpynbOutput (Just $
                                      if htmlFormat format
                                         then Format "html"
--- a/src/Text/Pandoc/App/CommandLineOptions.hs
+++ b/src/Text/Pandoc/App/CommandLineOptions.hs
@@ -686,6 +686,8 @@ options =
                       "all" -> return opt{ optIpynbOutput = IpynbOutputAll }
                       "best" -> return opt{ optIpynbOutput = IpynbOutputBest }
                       "none" -> return opt{ optIpynbOutput = IpynbOutputNone }
+                      "html" -> return opt{ optIpynbOutput = IpynbOutputHTML }
+                      "latex" -> return opt{ optIpynbOutput = IpynbOutputLaTeX }
                       _ -> E.throwIO $ PandocOptionError
                              "ipynb-output must be all, none, or best")
                  "all|none|best")
--- a/src/Text/Pandoc/App/Opt.hs
+++ b/src/Text/Pandoc/App/Opt.hs
@@ -75,6 +75,8 @@ data IpynbOutput =
     IpynbOutputAll
   | IpynbOutputNone
   | IpynbOutputBest
+  | IpynbOutputHTML
+  | IpynbOutputLaTeX
   deriving (Show, Generic)
 
 instance FromYAML IpynbOutput where
@@ -83,6 +85,8 @@ instance FromYAML IpynbOutput where
       "none"  -> return IpynbOutputNone
       "all"   -> return IpynbOutputAll
       "best"  -> return IpynbOutputBest
+      "html"  -> return IpynbOutputHTML
+      "latex"  -> return IpynbOutputLaTeX
       _       -> fail $ "Unknown ipynb output type " ++ show t
 
 -- | Data structure for command line options.


jgm:
Is this still needed when JSON filters are called on the command line using `--filter`?  OR is it only when you explicitly do `-t json`?

ickc:
I did not test, but my case is that I have a pipeline script in Python doing batch processing (with caches). There are 2 passes, one for from x to json (then to panflute AST), then it does some work, then it goes from (panflute AST) to JSON to output. So there's no way to can pandoc in one pass.

Regardless of workflow, I think it is not uncommon to treat `pandoc --to json ... | some_filter | pandoc --from json ...` as some sort of equivalent to `pandoc -F some_filter ...` although there will always be side-effects (that pandoc is smarter when called once.) This PR let those people manually overriding it.

Another use-case would be something I did to pannb I released yesterday (see a [test case here](https://github.com/ickc/pannb/blob/master/tests/test_pandoc_filter.py#L15-L21) that is an example of the 2 passes call above.) What it does is that

1. in ipynb, you have output cells
2. use pandoc to convert from ipynb, now those output will be selected according to `--ipynb-output`, and I force it to be `html` here
3. then the filter call pandoc to convert `raw_html` to native pandoc AST

In this case I think the only thing that can accomplish what I need is to add `--ipynb-output=html` option (without reinventing `pandoc --from ipynb ...` which is not impossible as it is just JSON too but is not leveraging the right thing.)

jgm:
> There are 2 passes, one for from x to json (then to panflute AST), then it does some work, then it goes from (panflute AST) to JSON to output. So there's no way to can pandoc in one pass.

I don't understand this.  Why can't you have one filter that reads pandoc JSON, converts this to panflute AST, does some work on this, converts the AST to JSON, and outputs this?

ickc:
> I don't understand this. Why can't you have one filter that reads pandoc JSON, converts this to panflute AST, does some work on this, converts the AST to JSON, and outputs this?

I not sure I understand you, do you mean a single filter calling pandoc directing, like `pandoc -F some_filter_that_is_going_to_call_pandoc_but_is_also_called_by_pandoc ...`?

That's just the decision of that workflow. It is a Python workflow, so Python is the caller, so you can imagine that there's only one single Python Interpreter has been invoked, and it calls pandoc whenever that single Python process feels like it.

That can be merits of reasons to do it (e.g. not to have the overhead of keep starting Python instances), but basically this is a picture of using pandoc as a library, not cli. [Although we cannot get around the fact that in the end we're calling pandoc cli, but the API is hiding that fact and acts as if it is a library. In ideal world we'd have pandoc as a real library to so that we can get around the overhead of calling pandoc cli multiple times (which I have other tricks to pack them and call pandoc only once in non-standalone cases...)]

One of the actual reason of that decision is caching. The input and input pandoc args might not change and in that case we only use the cache and then do stuffs in between and call `pandoc --from json ...` if the output pandoc args has changed for example. The caching not only cached the `pandoc --to json ...` work, but also the json to panflute AST construction which is not negligible.

ickc:
And to clarify

> There are 2 passes, one for from x to json (then to panflute AST), then it does some work, then it goes from (panflute AST) to JSON to output. So there's no way to call pandoc in one pass.

It means

There are 2 passes, one batch with many files convert from x to json, then cached, then some optional stuffs (including reading or even changing the metadata from inputs and optionally create a relationship between them), then another batch convert them to put. So it is a pipeline script (kind like makefile with arbitrary no. targets.)

jgm:
I see, thanks for the clarification.

jgm:
Couldn't you just use `--ipynb-output=all` for the steps that create JSON, and then `--ipynb-output=best` at the end when you're creating html or latex?

ickc:
One example use is that the filter pannb processes the ipynb output cells to convert from raw-block to native AST. So at the time the filter got the content, it has to already see the filtered results. And by the time it passes back to pandoc, pandoc will not see the cells output as it was, but a bunch of ASTs, so that it cannot further filter it out.

E.g. https://ickc.github.io/pannb/example-output/ The table you're seeing is the result of `pandoc -F pannb x.ipynb -o x.rst` where the rst is further processed by Sphinx documentation tool. The fact that we see a properly rendered table there is because the HTML table has been converted to pandoc AST before it reaches back to pandoc.

I think another way of saying it is that we want `--ipynb-output` to choose the "best" output as if the output will be HTML, but we are not really outputing to HTML, whether it is JSON (because we break up the pipes) or some other formats such as RST (because the filter is doing more things to convert from HTML.)

jgm:
I'm still not quite getting it. I understand that you want your filter to have access to the HTML output.  But with `all` you'd have that available to you, and you could transform it as you see fit.  Why does the filtering have to take place at the beginning of the pipeline rather than the end?  Can't your filter just ignore the output formats it's not interested in, knowing that they'll be weeded out at the end?

ickc:
> Why does the filtering have to take place at the beginning of the pipeline rather than the end?

There’s no beginning or the end. I think you may be thinking passing a filter to pandoc. We’re not, the first pass is Python calling pandoc to JSON, then Python does works including filtering, the last pass is pandoc from JSON. No filter has been passed to pandoc.

> Can't your filter just ignore the output formats it's not interested in, knowing that they'll be weeded out at the end?

Because it is aggressively converting any raw block to native AST. ipynb cells are only part of the picture. We are (or I am) composing a lot of different things together so may be what you are suggesting is ok one at a time. But I’m already talking about at least 3 things composing already. (A pipeline that Python calls pandoc not the other way around, a filter that can either be called by pandoc or inside Python by the said pipeline that performs multiple actions to ipynb output cells, and an action that aggressively convert any raw block to native.)

ickc:
Even if we’re willing to call pandoc 3 times, where the first to JSON, then Python does some work including filtering, then call pandoc only to filter the output, then Python does some more work that aggressively convert all raw blocks, etc.

the 2nd call to pandoc still won’t work without this patch as it is converting from JSON to JSON.

ickc:
The only way it would work will be letting the last call to pandoc to call the filter directly (-F) and in that case another Python process is invoked. And that’s the very thing I want to avoid. That’s why I was saying pandoc has never called any filter above.

ickc:
Could you briefly mention what’s the issue here? Is the solution too ugly, or incomplete?

ickc:
Actually, forget about composing. That's too complicated to understand the core issue.

Consider `pandoc -F pannb x.ipynb -s -o x.rst --ipynb-output=html`. (I mentioned this above, but this is the simplest case we can understand the core issue.) I don't think what you suggested would work in this case, because according to best strategy, it would not have selected the raw html, and hence we won't be able to convert from raw html to ast. Eg consider the cell output is an html table. Then in the output we would not have a table element unless we have `--ipynb-output=html`.

We could say this is a hack. And it is. But often the richness ipynb output is html (as the Jupyter is often run in web browser), so that's what we want to capture and convert it to pandoc AST.

jgm:
> Could you briefly mention what’s the issue here? Is the solution too ugly, or incomplete?

*Any* additional complexity is a cost, and needs to be justified by a clear need.  So it's important for me to understand why this problem can't be solved with what we currently have.

Let's just focus on the simpler case you give at the end.  Couldn't this be handled as:

```
pandoc x.ipynb -s -t json --ipynb-output=all | pannb | pandoc -f json -s -o x.rst
```
The json that gets passed to the `pannb` filter will include the HTML table, possibly as one of several output options.  In this case can't the filter (a) convert the HTML table to whatever, and (b) remove the extraneous output cells?

ickc:
I feel like we are talking past each other. I don't think there's any other solutions that can simultaneously does everything I've described so far. Whatever you're suggesting, I think it is only limited to a very specific situation.

In this case

> `pandoc x.ipynb -s -t json --ipynb-output=all | pannb | pandoc -f json -s -o x.rst`

May be it's ok with this particular example with the filter being more careful. A bunch of questions from this suggestion:

What would happen if the cell output contains 3 formats, and then the raw-block with HTML disappeared after the filter (that got turned into AST), and at last `ipynb-output=best` is specified? May be it would works, but the behavior is undefined given what `ipynb-output` is described, i.e. it is up to implementation details of pandoc. Now the programmer (filter writer) need to probe more about how pandoc actually work in this case. `ipynb-output=html` is much more natural from the programmer's perspective because it is telling pandoc to "do `best` pretending it is to HTML", which is a more defined behavior.

Now, the next level is about how this can be reused when multiple output formats are needed (imagine some pipeline program or makefile like that has a similar recipe for multiple output formats.) What `ipynb-output` should I use in the 2nd pandoc pipe? `ipynb-output=all` won't work as we need to filter out unneeded formats in output cells; `ipynb-output=best` is not safe as its action is output format dependent; `ipynb-output=none` is too aggressive as the some cells might not have html raw blocks, and the filtering should fall back to plain text for example.

So I think the final defense is really just think about what `--ipynb-output=html|latex` is doing: **it is giving the end-user (including filter writer/programmer) who need something like `--ipynb-output=best` (namely, currently the only strategy to include only one output format per output cell) but with a little bit more control.**

The solution I came up with after looking at the code is the least effort path as basically I'm reusing the `best` strategy, just that we're now telling it not to determine that by itself, but leave that to me to decide.

So that's why I asked that question, because to me, `--ipynb-output=html` is "clearly needed", but it is a bit ugly and incomplete. For example, basically now I'm exposing the implementational detail of pandoc's `best` strategy for us to choose from, but if in the future the best strategy is determined differently, then this exposed choice might not make sense any more, and for compatibility's sake it would be next to impossible to make that change.

I think I've said enough. Please feel free to decide if you will accept this. While I do hope this get merged, it won't be the end of the world if it doesn't.

Thanks!

ickc:
> Why can't the filter just remove these other raw blocks from the output cell when it incorporates the HTML table into the AST?

This is what I want to avoid. It is starting to replicating pandoc's functionality (in filtering the cell.)

For example, in order to do that, the filter has to be specifically written for this application (that converting raw-block of a certain format within the output cell div) and it has to be contextual (that looks for adjacent elements.) But the way it is written in pannb currently is 3 independent filters where one can cherry-pick what they want, where the pannb cli is a shortcut to use all 3 simultaneously.

> Isn't that just what you want when you're producing the final output? (We're talking about the second invocation of pandoc in the pipe here, the one that produces an rst file or whatever.)

E.g. what if we are converting to LaTeX/PDF, and there's an output format suitable for LaTeX output, but the filter has already convert from the HTML raw-block to AST? Then it is likely the 2nd pipe filtering of `--ipynb-output=best` would have 2 outputs. This situation is so far hypothetical, but the point is it actually creates more "edge case" that one might not be able to foresee in the beginning.



> I'm sorry for all the questions. I need to get clear about why the feature is needed...and I'm still not.

I appreciate this. This is what makes pandoc special! I was just asking for clarification on what was the issue.

ickc:
I changed my mind and now think that my approach is over-generation. It cannot handle all general situation (any output cells from the Python ecosystem), and for those specific to mine I have control over, there are better solutions.

I may have questions on how different mime types are chosen wrt output formats but probably a question for pandoc-discuss.

P.S. Just write down a trick that resemble what this patch is doing for future reference: if you have to have a behavior similar to the options here, create a simple filter that intercept the JSON/AST that put it in stderr or whatever, emit error or delete the whole doc to end pandoc early, and use `pandoc -F intercept_filter --to html ...`. This functions almost the same as described here except it is less efficient and has a worse interface.

PR #7543, opened 2021-09-03
Improve order of Image fields

quinn-dev:
Since `identifier`, `classes` and `attributes` are aliases for `attr`, they should follow directly after the `attr` field to avoid confusion. This is the case for all types documented above `Image`.


jgm:
Thanks!

PR #7544, opened 2021-09-03
pptx: Miscellaneous fixes

undergroundquizscene:
The commits in this PR consist of a few miscellaneous fixes:

- Fix a typo in the generated XML
- Copy all themes and embedded fonts from the reference doc into the output file
- Don't add duplicate relationships for slide layouts
- Fix the relationship (rId) numbering in presentation.xml and presentation.xml.rels

I can split them into multiple PRs if that would be better.


undergroundquizscene:
The deleted-layouts tests currently produce pptx files that PowerPoint says need to be repaired: I'm looking at fixing this. It's caused by introducing tests with multiple masters.

undergroundquizscene:
> The deleted-layouts tests currently produce pptx files that PowerPoint says need to be repaired: I'm looking at fixing this. It's caused by introducing tests with multiple masters.

I'm having some trouble figuring out what's wrong with this, so I think I'll remove the commit supporting multiple masters from this PR: I'll work on that on another branch so the other changes can be merged in the meantime.

jgm:
Just checking -- is this ready to review, squashed the way you want it, etc.?

undergroundquizscene:
Yup, good to go!

Issue #7506, opened 2021-08-17
man writer: \f[C] is not visible on terminal


joeyh:
**Explain the problem.**

The man writer uses `\f[C]` to mark up code blocks. On my Debian linux system, the resulting man page displays with code blocks indistinguishable from normal text. When a man page uses code blocks inside a sentence, with the expectation that they'll be visually different, this can make them harder to read.

pandoc's own man page demonstrates this problem to some extent. In the DESCRIPTION, there are some options mentioned, which are code blocks in the input markdown. They are displayed as regular unformatted text. (Only checked this with the 2.9.2.1 man page.)

The groff man page does not mention a 'C' font. It does say:

       the utf8 device (for terminal
       emulators) has only font descriptions for the roman, italic,
       bold, and bold-italic styles (R, I, B, and BI, respectively).

So, it seems like it would be a good idea for the man page to use one of those font styles for code blocks.

**Pandoc version?**

2.9.2.1. current git master seems to also hardcode `\f[C]` based on examining the code

jgm:
I see the point of this request. Still:

- terminal is just one possible target for man output.  groff also produces HTML, and for that `\f[C]` gives the results we want.
- I looked at some man pages and didn't see a general convention of using a bold or italic font for inline code or literal text.  (More examples welcome.)

The ideal solution, I suppose, would be to use a groff macro that changes the font depending on the target format (if that's even possible).

jgm:
```
.ie "\f[C]x\f[]"x" \
true
.el \
false
```
prints "true" with `man` and "false" with `groff -mman -Thtml`.  So in principle it would be possible to make the output conditional on the output format.

jgm:
Proof of concept:
```
.\" Check to see if C font is rendered; if not, render as B
.if "\f[C]x\f[]"x" \{\
. ftr C B
.\}

.TH "" "" "" "" ""
.hy
.SH Hi
Hello \f[C]there\f[].
```
This renders as bold with `man` and as `<tt>` with `-Thtml`.
In fact, you could just add this to a custom man template and use current pandoc.

jgm:
@joeyh PR #7545 in a possible solution. I'm not sure, though - using a custom font might cause problems for people (if any) who use `pandoc -t man` to generate fragments rather than full templated man pages.

pmachata:
Came in support of the request. I consider the terminal to be the primary medium for man pages, and when my inline command-line snippets etc. are rendered in plain text, the result is confusing to read.

[...]

PR #7545, opened 2021-09-03
Man writer: use custom font V for inline code.

jgm:
The V font is defined conditionally, so that it renders
like C in output formats that support that, and like B
in those that don't (e.g. the terminal).

We could just redefine C, but this would affect code
blocks, too, and putting them all in boldface looks ugly,
I think.

Possible drawback: fragments created by pandoc's man
writer will presuppose a nonstandard V font.

Addresses #7506.


jgm:
Merged manually.

PR #7548, opened 2021-09-04
Fix typos in pandoc.path docs

quinn-dev:
`iff` -> `if`, two occurrences.


jgm:
Here I suspect `iff` is meant as an abbreviation of `if and only if`, but because we have an "otherwise" clause, I agree `if` is simpler.

PR #7554, opened 2021-09-08
pptx: Fix logic for choosing Comparison layout

undergroundquizscene:
There was a mistake in the logic used to choose between the Comparison
and Two Content layouts: if one column contained only non-text (an image
or a table) and the other contained only text, the Comparison layout was
chosen instead of the desired Two Content layout.

This PR fixes that logic:

> If either column contains text followed by non-text, use Comparison.
  Otherwise, use Two Content.

It also adds a test asserting this behaviour.


undergroundquizscene:
I don't believe the test failure is to do with my changes:

> meta blocks can be round-tripped through the lua stack:                                                                       FAIL (0.36s)
>        *** Failed! Falsified (after 16 tests):
> [...]
>  cabal: Tests failed for test:test-pandoc from pandoc-2.14.2.
> Use --quickcheck-replay=248230 to reproduce.
>
> Use -p '/meta blocks can be round-tripped through the lua stack/' to rerun this test only.

I will rebase to see if running the tests again is successful with a different random seed.

undergroundquizscene:
@jgm I'd like to get this merged before the next release, if possible: I introduced the bug after the last release, so merging before the next would mean the bug isn't in any releases. (No rush getting to this, I'm just tagging you to make sure you see it before the next release.)

jgm:
Understood. Looks like it needs rebasing again?

undergroundquizscene:
> Understood. Looks like it needs rebasing again?

Ah yes, looks like it: will do that now.

undergroundquizscene:
Rebased and ready to go!

jgm:
ok, merged!

PR #7556, opened 2021-09-09
Ipynb reader: improve Ipynb.Raw handling

ickc:
This issue centers around Ipynb.Raw.

- fixes a typo in the [mime type of rst](https://docutils.sourceforge.io/FAQ.html#what-s-the-official-mime-type-for-restructuredtext-data)
- add most (i.e. those relevant to pandoc) selectable formats in jupyterlab's "Raw NBConvert Format", which includes a different mime-type for rst.
- format -> raw_mimetype. See jupyter/nbformat#229. Essentially while in the nbformat spec, format should be used, in the actual implementation of nbconvert, lab and notebook (i.e. the official ecosystem around ipynb) uses raw_mimetype instead. So for practicality for now we should use `raw_mimetype` as it is the only implemented key in the wild. We'll see how nbformat is resolving that issue.

Edit: also matches the above behavior in the ipynb writer too. Especially there's a backward incompatible changes documented in commit 346df3b.

### DIFF
--- a/src/Text/Pandoc/Readers/Ipynb.hs
+++ b/src/Text/Pandoc/Readers/Ipynb.hs
@@ -101,7 +101,7 @@ cellToBlocks opts lang c = do
               "text/latex"      -> "latex"
               "application/pdf" -> "latex"
               "text/markdown"   -> "markdown"
-              "text/x-rsrt"     -> "rst"
+              "text/x-rst"      -> "rst"
               _                 -> format
       return $ B.divWith ("",["cell","raw"],kvs) $ B.rawBlock format' source
     Ipynb.Code{ codeOutputs = outputs, codeExecutionCount = ec } -> do
--- a/src/Text/Pandoc/Readers/Ipynb.hs
+++ b/src/Text/Pandoc/Readers/Ipynb.hs
@@ -97,12 +97,16 @@ cellToBlocks opts lang c = do
       let format = fromMaybe "ipynb" $ lookup "format" kvs
       let format' =
             case format of
-              "text/html"       -> "html"
-              "text/latex"      -> "latex"
-              "application/pdf" -> "latex"
-              "text/markdown"   -> "markdown"
-              "text/x-rst"      -> "rst"
-              _                 -> format
+              "text/html"             -> "html"
+              "slides"                -> "html"
+              "text/latex"            -> "latex"
+              "application/pdf"       -> "latex"
+              "pdf"                   -> "latex"
+              "text/markdown"         -> "markdown"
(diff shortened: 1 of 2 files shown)


jgm:
>  Essentially while in the nbformat spec, format should be used, in the actual implementation of nbconvert, lab and notebook (i.e. the official ecosystem around ipynb) uses raw_mimetype instead. So for practicality for now we should use raw_mimetype as it is the only implemented key in the wild. We'll see how nbformat is resolving that issue.

Why not try raw_mimetype first, then fall back to format?
That way we cover all the bases.  This can be done with
```hs
fromMaybe "ipynb" $ lookup "raw_mimetype" kvs <|> lookup "format" kvs
```

jgm:
Also, what do we do on the writer side?  Do we use `format`?  Should we perhaps use *both*?  (I hate to depart from the spec'd behavior.)

ickc:
force-pushed from 556bf4e to 859d4fd was for rebasing the latest commits from the other PR.

ickc:
@jgm, I didn't consider the writer.

Thinking about that, another concern would be

```haskell
                  "rst"      -> "text/x-rst"
```

because there's no official mime-type for rst, and it is declared [here](https://docutils.sourceforge.io/FAQ.html#what-s-the-official-mime-type-for-restructuredtext-data) that it should be `text/x-rst`, while the ipynb official ecosystem is using `text/restructuredtext` instead.

I don't know a good solution to raw_mimetype vs format in writing ipynb, duplicating seems excessive. And `format` is not really used in the wild AFAIK. I'd prefer stick with `raw_mimetype` only.

I think how it is resolved in https://github.com/jupyter/nbformat/issues/229 will eventually informs how it should be handled here, but in the interim, may be `raw_mimetype` only?

jgm:
OK, you have a better sense than I do of what makes sense for this.

ickc:
@jgm, it should be ready for final review.

The reader has no backward incompatible changes except for "text/x-rsrt" while I guess is a typo that can be safely discarded.

The writer has backward incompatible changes where commit 346df3b has documented them.

The writer also added a bunch of format mapping, in particular I'm not 100% certain for those maps to html.

jgm:
Could you squash some of these commits (e.g. the one importing `<|>`) into a smaller number of logically separate commits, each with a good description of what it is doing?

ickc:
I rewrote the history, and it is complaining the commit messages is longer than 78 characters.
However, I edited them so that none of them has more columns than that.
May be the "Check commit message length" is wrong?

jgm:
thanks!

Issue #5689, opened 2019-08-16
--incremental option does not work for pptx output


jooyoungseo:
I have used the latest Pandoc V2.7.3 to see if "--incremental" option works for PowerPoint list elements.
It does not seem it works for pptx output.
Do you have any plans for improving this feature for the PPTX output?

ciaochiaociao:
pandoc -i ... also does not work

PR #7559, opened 2021-09-09
pptx: Add support for incremental lists

undergroundquizscene:
- Support -i option
- Support incremental/noincremental divs
- Support older block quote syntax
- Add tests

One thing not clear from the manual is what should happen when the input
uses a combination of these things. For example, what should the
following produce? (With `-i` set? Without it?)

```md
::: {.incremental .nonincremental}
- are
- these
- incremental?
:::

::: incremental
::::: nonincremental
- or
- these?
:::::
:::

::: nonincremental
> - how
> - about
> - these?
:::
```

In this commit I’ve taken the following approach:

- incremental/nonincremental divs have the highest priority, so they’re
  unaffected by the block quoting syntax
  - I chose this because they’re the newest syntax
- the innermost incremental/nonincremental div is the one which takes
  effect
- a div with both classes acts as if it has neither
- block-quoting changes the behaviour if outside of an
  incremental/nonincremental div

I’ve added some tests to verify this behaviour.

This PR closes https://github.com/jgm/pandoc/issues/5689.


undergroundquizscene:
Does the approach I've taken with the priority of the incremental/nonincremental divs and block quotes seem ok?

jgm:
Have you tried testing this markdown with beamer and revealjs to see if the behavior you've outlined fits what we do already?

undergroundquizscene:
I have not, I'll do that today.

undergroundquizscene:
> I have not, I'll do that today.

I haven't got to this today, but I'll do it soon :)

undergroundquizscene:
Ok, I've investigated the behaviour of the beamer and revealjs outputs. In both formats, with or without the `--incremental` option, I see the following:

- in heading attributes, the `incremental` class overrides the `nonincremental` class
- the innermost incremental/nonincremental div takes effect
- block quoted lists within incremental/nonincremental divs are inverted
- incremental/nonincremental divs within block quotes are not inverted

I'll update this PR to match that behaviour, and comment here once that's done.

undergroundquizscene:
I have made those changes now, so once the tests have finished this PR should be ready to merge.

jgm:
Excellent.

PR #7562, opened 2021-09-10
Use babel, not polyglossia, with xelatex

jgm:
Previously polyglossia worked better with xelatex, but
that is no longer the case, so we simplify the code so that
babel is used with all latex engines.

This involves a change to the default LaTeX template.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -142,11 +142,11 @@ When using LaTeX, the following packages need to be available
 contains images), [`hyperref`], [`xcolor`],
 [`ulem`], [`geometry`] (with the `geometry` variable set),
 [`setspace`] (with `linestretch`), and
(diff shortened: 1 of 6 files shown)


alerque:
Wow, my first reaction to this was to flail and object, but apparently the situation really has reversed since I was doing new work in LaTeX.

jgm:
I should have added this link as background:
https://chat.stackexchange.com/transcript/message/59070629#59070629

jgm:
If people who do multilingual work could test with this branch, it would give me more confidence.
(including with RTL languages)

tstenner:
Just for completeness: the [discussion on tex.stackexchange](https://tex.stackexchange.com/q/482396/13147) was also in favor of babel.

PR #7567, opened 2021-09-13
pptx: Support specifying slide background images

undergroundquizscene:
In the reveal-js output, it’s possible to use reveal’s
`data-background-image` class on a slide’s title to specify a background
image for the slide.

With this PR, it’s possible to use `background-image` in the same
way for pptx output. Only the “stretch” mode is supported, and the
background image is centred around the slide in the image’s larger axis,
matching the observed default behaviour of PowerPoint.

- Support `background-image` per slide.
- Add tests.


undergroundquizscene:
This PR is a draft because I still need to update the manual.

I also have a question: rather than using `background-image`, should I use `data-background-image` instead? `data-background-image` is supported by the revealjs output, but I think that's because the class is passed through to reveal, which handles it. The data prefix seems unnecessary to me, which is why I chose `background-image`, but I guess this is now inconsistent. What do you think?

undergroundquizscene:
I have now updated the manual. I have a question about the manual (not really related to my changes): how are the `#option--x` link targets generated? E.g. there's a working `#option-css` target, and I've guessed in my changes that there's a corresponding target for the `--reference-doc` option section. I've read through the manual a bit and can't see anything describing automatic identifiers on things other than headings: is this something done specially for the manual?

jgm:
On the first point:  the HTML writer adds a `data-` prefix to attributes that aren't already legal HTML attributes.  So, if you're targeting reveal.js, you can use either `background-image` or `data-background-image` in a markdown file, and they'll do the same thing. (I think the documentation should talk about the former only.) For pptx the `data-` prefix is irrelevant; I'd just use `background-image`.

jgm:
On the second: with a lua filter; see `tools/option-anchors.lua` in the pandoc website repository.

undergroundquizscene:
> On the first point: the HTML writer adds a `data-` prefix to attributes that aren't already legal HTML attributes. So, if you're targeting reveal.js, you can use either `background-image` or `data-background-image` in a markdown file, and they'll do the same thing. (I think the documentation should talk about the former only.) For pptx the `data-` prefix is irrelevant; I'd just use `background-image`.

Ah, excellent – thanks for the explanation. Will I update the manual to only mention `background-image` in the reveal.js sections, maybe with a note explaining that the data prefix is added?

> On the second: with a lua filter; see tools/option-anchors.lua in the pandoc website repository.

Perfect, thanks!

undergroundquizscene:
(Accidentally pushed a commit intended for another branch: I've removed it now)

jgm:
> Will I update the manual to only mention background-image in the reveal.js sections, maybe with a note explaining that the data prefix is added?

That makes sense I think.

undergroundquizscene:
Manual updated – once the tests have passed, I believe this is ready to go.

MANUAL.txt
```diff
@@ -6057,55 +6057,71 @@ the [Beamer User's Guide] may also be used: `allowdisplaybreaks`,
...
 
 ```
 ---
 title: My Slide Show
 parallaxBackgroundImage: /path/to/my/background_image.png
 title-slide-attributes:
-    data-background-image: /path/to/title_image.png
-    data-background-size: contain
+    background-image: /path/to/title_image.png
+    background-size: contain
```
jgm, line 6117: I'm afraid that *here*, at least, it does need to have the `data-` prefix. The reason is that this material is processed directly by the template, so the keys and values get passed through without change.

jgm: It's only when `background-image` occurs as an attribute on an AST element that the `data-` gets added.

jgm:
Maybe, to be on the safe side, we should check for a `data-background-image` attribute if `background-image` isn't found. After all, people may already be using `data-background-image` in their slide shows, and it would be good if this still worked with pptx.

undergroundquizscene:
Yes, that sounds sensible – I'll rebase now and make that change.

undergroundquizscene:
Change made, ready to go!

MANUAL.txt
undergroundquizscene: Ah, sorry – I missed these comments. I will update the manual now

Issue #7547, opened 2021-09-04
Textile writer: Use native definition list syntax


Atlas48:
Hello! I was notifying that pandoc doesn't support textile's native definition list syntax, and instead opts for a HTML workaround.

Textile may have added this at a later date, and overall, it's one character away from the exact syntax org uses for definition lists.

### Org
```org
- Cookies :: cream
- Sugar :: milk
```
### Textile
```textile
- Cookies := cream
- Sugar := Milk
```
This probably means we can copy over the definition list writer function from org and change it slightly.

I *could* do it, but I consider my experience with haskell too limited, and my confidence, too low. I'll try and work on a fix if I can.

jgm:
One issue is how this works when the definition contains multiple block-level elements (e.g. paragraphs, or a sublist), and how it works when there are multiple definitions. The writer would need to deal with both these cases.  If the native syntax can't handle this, we might need to fall back to HTML.

Can you link to a syntax description for textile definition lists?

Atlas48:
> Can you link to a syntax description for textile definition lists?

I can link the line from the official textile spec where it's [defined](/textile/textile-spec/blob/a0615116ddf341bd57f54564af96e018bd5b9731/page_layout.yaml#L142).

As far as the specification's concerned, there's no such thing as a nested definition list or a mixed definition list.

Atlas48:
I've also attempted to haphazardly port some of org's code over in my own [branch](../compare/master...Atlas48:textile-fix), if you want me to file a separate pull request, holler at me.

Atlas48:
@jgm Calling for code review of my branch. Bear in mind I am basically cargo-cult programming here.

PR #7571, opened 2021-09-16
Changed definition list output for textile.

Atlas48:
Requesting a code review, I've merely copied over org's definition and changed it slightly as they are almost syntactically identical.

### DIFF
--- a/src/Text/Pandoc/Writers/Textile.hs
+++ b/src/Text/Pandoc/Writers/Textile.hs
@@ -241,8 +241,8 @@ blockToTextile opts x@(OrderedList attribs@(start, _, _) items) = do
         return $ vcat contents <> (if level > 1 then "" else "\n")
 
 blockToTextile opts (DefinitionList items) = do
-  contents <- withUseTags $ mapM (definitionListItemToTextile opts) items
-  return $ "<dl>\n" <> vcat contents <> "\n</dl>\n"
+  contents <- mapM (definitionListItemToTextile opts) items
+  return $ vcat contents $$ blankline
 
 -- Auxiliary functions for lists:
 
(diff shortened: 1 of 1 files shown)


Atlas48:
See issue #7547.

src/Text/Pandoc/Writers/Textile.hs
```diff
@@ -282,8 +282,7 @@ definitionListItemToTextile :: PandocMonad m
 definitionListItemToTextile opts (label, items) = do
   labelText <- inlineListToTextile opts label
   contents <- mapM (blockListToTextile opts) items
-  return $ "<dt>" <> labelText <> "</dt>\n" <>
-          T.intercalate "\n" (map (\d -> "<dd>" <> d <> "</dd>") contents)
+  return $ "- " <> labelText <> " := " <>) contents) $$ cr
```
jgm, line 285: This is going to give you bad syntax if contents contains multiple paragraphs or other block level elements, won't it?  (See my original query on the issue.)

We could, as before, fall back to using a raw HTML definition list in this case.

In any case, I'm surprised that no tests failed with this change.  Aren't we testing definition lists with the textile writer?

Atlas48: I have no idea what I'm doing whatsoever, I just thought that this would be a nice jumping-off point.

Textile does implement a "close definition list" token, `=:`, but I think me implementing it wouldn't be 100% the best, I can try though.

tarleb:
Hi @Atlas48, just checking if you need help in getting this PR into a mergeable state? It would be nice to get this into pandoc at some point.

Atlas48:
@tarleb yes please, I have only so much clue about writing haskell.

Issue #5660, opened 2019-07-25
Powerpoint writer - footer, slide numbers, date


Grandement:
Hi,

When I use the date keyword the date is displayed only on the title page. But not on other slides. Can we implement something to enable it? Or enable the date, footer or slide numbers in general?

E.g. the date is applied to the title slide but in my reference pptx it is set to "do not show on title slide". It is set in the final pptx as well but displayed on the title page and not on the others.

![image](https://user-images.githubusercontent.com/12670931/61852304-76fd3800-aeb9-11e9-9b79-74667625bb17.png)

agusmba:
This is a tough one.

On the one side I see the benefits of getting a pptx ready to go, but on the other hand I'd also like the pptx to be easily editable in PowerPoint (thus I'd be against hardcoding footer values in the pptx that do not follow standard PowerPoint behavior).

Due to the lack of variable (document property) substitution in PowerPoint (unlike in Word), the templating part of headers/footers must be done via master slides, with sensible use of the Header/Footer option you showed above.

I'm not sure (I didn't implement this feature) but I feel it would be quite complex for pandoc to read and understand the master slides of your reference document so that it could mimic the use of said Header/Footer option.

Personally I just use that option manually from PowerPoint after pandoc generates the pptx, although it can be a pain to always relay on this manual step.

Maybe there could be a way of making pandoc *activate* the Header/Footer option so that when the pptx opens it is populated (like the content index in docx), but I have no idea if this is feasible or not.

serussell:
@agusmba -- can you link to the documentation about how to do document property variable substitution, based on (I assume) header variables in the source document?  I've been wanting to do this, to get things like the subtitle into a docx footer in the template.

agusmba:
Can't remember if there's a detailed guide on this, but it's pretty easy.

Edit your docx reference doc and insert in your footer any DocumentProperty you want (subtitle in your case). If it doesn't exist, create it beforehand, the value doesn't matter, it will be substituted by the correct one by pandoc.

I can't write comfortably on my phone, if there's a need I could write a guide in a few weeks.

serussell:
@agusmba , see, I tried that.  It removed the subtitle from the title page, so I decided that meant that only one replacement was made in the document.  Perhaps I did something wrong; `pandoc` supposed to replace *all* occurrances of styled keyword texts?  I'll try again.

agusmba:
>  It removed the subtitle from the title page

Mmm, that's strange.

[...]

PR #7572, opened 2021-09-16
pptx: Support footer content defined in the reference doc

undergroundquizscene:
In PowerPoint, you can define footer content (a date, either fixed or automatically updating, arbitrary text content, and the slide number, optionally starting from a specific value) in the master, which is then duplicated across all slides (though can optionally be left off the title slide).

This PR adds support for that footer content: if a footer is specified in the master, it will be reproduced across all slides (except the title slide, if that option is disabled).


undergroundquizscene:
This PR is WIP for now, as I still need to add tests and documentation.

I also have a question: if the user's reference doc contains a date footer, but they've also specified a date in the YAML metadata block, what should we do? At the moment, this code ignores the YAML date, and includes only the date from the reference doc.

As I see it, there are a couple of other options:

- Use the YAML block date only on the automatically-generated title slide (possibly only if the reference doc footer content says not to display it on the title slide)
- Ignore the date in the reference doc footer, if there's a date in the YAML block
- Ignore only a fixed date in the reference doc footer, if there's a date in the YAML block
  - The other kind of date is an automatic one: one which PowerPoint updates to today's date any time you open the file

jgm:
Not sure about the date issue -- what do you think?  I guess a fixed date in a reference doc wouldn't really be useful.  So I'm happy to have that replaced by metadata date.  But if it's an automatic date, I'm not sure.

undergroundquizscene:
I think we should replace fixed dates in the footer with the date from the YAML block, but leave automatic dates as they are. If the user wants the fixed date from the reference, they can always remove the date from the YAML block, or set it to the same value as the reference doc. On the other hand, automatically updating dates is a feature that isn't available via the markdown source, so I'd rather keep it available via the reference doc.

I'll make that change today but keep it in a separate commit pending any further discussion.

undergroundquizscene:
This PR is pretty much ready to go – there are just two outstanding issues. First, there's a lint failure caused by some other code (so I believe once that's fixed the checks here will all pass), and I'd like to squash the two commits into one if we're happy with the metadata date replacing a fixed date from the reference doc.

jgm:
I fixed the linter issue.

undergroundquizscene:
Thanks, I'll rebase

undergroundquizscene:
This PR fixes https://github.com/jgm/pandoc/issues/5660

Issue #4828, opened 2018-08-09
Powerpoint Output: bullet points are rendered as indented to level 2


rich-iannone:
The issue here was originally posted in https://github.com/rstudio/rmarkdown/issues/1418, where the OP expressed that this output (bullet points appearing as indented to level 2) is not the expected behaviour, requiring manual editing after rendering.

To reproduce, use the following (bullets.md):

````
### Slide with Bullets

- Bullet 1
- Bullet 2
- Bullet 3
````

I ran this with pandoc (v.2.2.1):

```
pandoc +RTS -K512m -RTS bullets.md --to pptx --from markdown --output bullets.pptx
```

Here's a screenshot of bullets.pptx, showing the bullet characters appearing reasonably far from the left border of the text box.

<img width="883" alt="bullets-indented" src="https://user-images.githubusercontent.com/5612024/43918135-13e1860a-9bc7-11e8-887a-e0b8cc47fc11.png">

jgm:
@jkr when you have a chance, could you look at this and #4827?

jkr:
Yes -- sorry I've been offline this summer. (Newly associated, so
checked out a bit.) I will take a look at these today, and get something
up soon.

dvbuntu:
I've noticed this recently as well.  Any update or workaround on this?

jkr:
Sorry to take so long on this. This is just PowerPoint's default template's formatting. The bullets are actually level 1. If you want it to be indented differently, you need to use a different template (`reference-doc`, in pandoc-speak).

Run pandoc on the above example to produce a PowerPoint doc.

~~~
$ cat input.text                                         
### Slide with Bullets

- Bullet 1
- Bullet 2
- Bullet 3
~~~

~~~
$ pandoc input.text -o output.pptx                             
~~~

Note the "lvl = 1" in the output
~~~
$ unzip -p output.pptx ppt/slides/slide1.xml|xmllint --format -

<?xml version="1.0" encoding="UTF-8"?>
<p:sld xmlns:a="http://schemas.openxmlformats.org/drawingml/2006/main" xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships" xmlns:p="http://schemas.openxmlformats.org/presentationml/2006/main">
  <p:cSld>
    <p:spTree>
      <p:nvGrpSpPr>
        <p:cNvPr id="1" name=""/>
        <p:cNvGrpSpPr/>
        <p:nvPr/>
[...]

PR #7578, opened 2021-09-17
WIP: pptx: Fix list level numbering

undergroundquizscene:
In PowerPoint, the content of a top-level list is at the same level as
the content of a top-level paragraph – the only difference is that a
list style has been applied.

At the moment, the pptx writer increments the paragraph level on each
list, turning what should be top-level lists into second-level lists.

This commit changes that logic. When we encounter a list:

- if we’re at level 0 and we’re not inside a list, stay at level 0
- otherwise, increment the level


undergroundquizscene:
One thing I'm not certain of is whether I should change how the writer handles continuation paragraphs within lists.

At the moment, continuation paragraphs are treated as paragraphs at the same level as the list: before my changes, this means all paragraphs in a top-level list are indented by one level.

Given this input markdown:

```
# Slide

Paragraph.

1. Top-level

   Continuation
   
   1. Sub-list
   
      With Continuation

   2. (still sub-list)

2. (back to top-level)

Paragraph.
```

The output pptx looks like this:

output-old

After my changes, they're not indented at all, so the continuation paragraphs are identical (from PowerPoint's perspective) to any top-level paragraphs outside bullets.

The same input markdown now produces this output:

output-new

What I think might be better would be to indent any continuation paragraphs by one level (and not indent to a new level on reaching a new list). This is an output representation that isn't currently produced by any markdown input that I can think of (so there shouldn't be confusion caused by different inputs producing the same output), and it has the added benefit of fixing list numbering in the presence of continuation paragraphs:

output-proposed

What do you think?

undergroundquizscene:
This PR fixes https://github.com/jgm/pandoc/issues/4828

jgm:
That looks right to me -- at least in so far as the indentation goes.
It would be nice also to (a) make sure there's some space between the list marker (e.g. 1.) and the following text, and (b)make sure the text following the list marker lines up (exactly, not just approximately) with the continuation paragraph.
We have a way of doing this in the docx writer, whose details I can't recall.

undergroundquizscene:
Ok, I'll make the change to indent continuation paragraphs rather than the lists themselves.

I'll look into the spacing as well, see how that might be done: whatever worked in the docx writer might also work here alright.

undergroundquizscene:
I've had an initial look at how to handle the spacing. I think it could be done by specifying the right combination of values for the `marL` and `indent` attributes on the first paragraph of each list with continuation paragraphs. To figure out the right value, we'd need to read the list level styles from the current master, which would be fine. I'm having a bit of trouble understanding what the correct values to use for the two attributes though, especially for indenting the text without indenting the list marker.

If I understand the docx writer correctly, I think it's doing a similar thing with the indentation in [mkLvl](https://github.com/jgm/pandoc/blob/3de0d5a9778e5be02d47e1de607c642948cb02e9/src/Text/Pandoc/Writers/Docx.hs#L652-L654)?

jgm:
Yes, if I recall, the docx writer assumes a fixed amount of indent per indentation level.  Reading this from styles would be better of course!

undergroundquizscene:
I haven't got this working yet, but I think I'm close. I wanted to sum up what I've found so far.

The spec defines `marL` and `indent`, but it's not very clear:

- `marL`
  
   > Specifies the left margin of the paragraph. This is specified in addition to the text body inset and applies only to this text paragraph. That is the text body inset and the marL attributes are additive with respect to the text position. If this attribute is omitted, then a value of 347663 is implied.

- `indent`

   > Specifies the indent size that is applied to the first line of text in the paragraph. An indentation of 0 is considered to be at the same location as marL attribute. If this attribute is omitted, then a value of -342900 is implied.

The list level values for these in pandoc's default reference doc are:

1. marL="342900" indent="-342900" 
2. marL="742950" indent="-285750" 
3. marL="1143000" indent="-228600"
4. marL="1600200" indent="-228600"
5. marL="2057400" indent="-228600"
6. marL="2514600" indent="-228600"
7. marL="2971800" indent="-228600"
8. marL="3429000" indent="-228600"
9. marL="3886200" indent="-228600"

The values seem to work as follows:

- Positive `marL` values add space to the left of the paragraph, indenting it.
- For lists, `indent` adds space between the list marker (e.g. the bullet for bulleted lists) and the item text. Positive values add space to the right of the marker, pushing the text to the right. Negative values instead push the marker to the left, eating into the margin. 
- If the indent value is negative and its absolute value is greater than the positive marL value, then the text is pushed to the right once the margin has been exhausted. E.g., there is no visible difference between (marL, indent) values of (400000, -400000), (200000, -400000), and (0, -400000).

For examples that are just lists, like the following, setting `marL` to the `marL` from the next indentation level and `indent` to (marL + indent - new marL) produces the result we want.

```markdown
# Slide

- First point
  - some
  - sub
  - points
- Second point
```

However, this doesn't work correctly where the continuation paragraphs have no bullet style – the initial paragraphs end up over-indented. I'm not sure yet how to work out how much to decrease `marL` by to make that work.

jgm:
> However, this doesn't work correctly where the continuation paragraphs have no bullet style – the initial paragraphs end up over-indented. I'm not sure yet how to work out how much to decrease marL by to make that work.

The docx writer doesn't have this issue, does it?  Maybe look at what it is doing with continuation paragraphs?  (It has been a while, so I don't have it in my head.)

undergroundquizscene:
> The docx writer doesn't have this issue, does it? Maybe look at what it is doing with continuation paragraphs? (It has been a while, so I don't have it in my head.)

I don't know, I'll take a look. Is it worth merging these changes while I work out the spacing? I won't have much time to work on this from now on, so it may take a while.

jgm:
I'm open to merging now if this is already strictly better than what we have.
Can you post a sample of the sort of markdown input that leads to the residual spacing issue, and an image of how it looks in powerpoint?

undergroundquizscene:
Sure, no problem. I haven't yet pushed the spacing changes to this branch, because I don't think they're ready to merge. At the moment, given this markdown:

```markdown
# Slide

Paragraph.

1. Top-level

   Continuation
   
   1. Sub-list
   
      With Continuation

   2. (still sub-list)
   
      - continuation
      - list

2. (back to top-level)

Paragraph.
```

This PR produces a powerpoint which looks like this:

Demo

With my spacing changes, the output currently looks like the following. I believe there are two problems to address: the spacing is applied to all list items, when it should only be applied to the first paragraph of items with continuation paragraphs; and the amount of spacing lines up with the text after a bullet, rather than with the bullet/a non-bulleted item.

Demo with Spacing

(So to be clear, I'm suggesting merging this PR as-is, which produces the output in the first picture, and making a new PR to continue working on improving the spacing.)

jgm:
One thing that's odd about the current output is that there appears to be no space at all between the list markers and the text after them at the second level, at least with the enumerated lists.  Shouldn't there be a tab there?  And can't the tab stop be set up so it aligns exactly with where the continuation paragraphs will start?

undergroundquizscene:
> One thing that's odd about the current output is that there appears to be no space at all between the list markers and the text after them at the second level, at least with the enumerated lists. Shouldn't there be a tab there? And can't the tab stop be set up so it aligns exactly with where the continuation paragraphs will start?

Yeah, I've noticed that before (it's a pre-existing problem) and found it odd, but hadn't had a chance to look yet. I've looked into how PowerPoint behaves (by reproducing the above slide within a new presentation in PowerPoint), and it doesn't manage that spacing by inserting a tab. Instead it uses the `marL` and `indent` attributes I mention above, but it looks like it overrides the presentation-wide values on numbered list items and indented non-list paragraphs.

Here's a picture of the slide I created in PowerPoint:

Manual Example

And here's the xml for that slide (indented with `xmllint -format -`):

<details>
  <summary>slide1.xml</summary>

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<p:sld xmlns:a="http://schemas.openxmlformats.org/drawingml/2006/main" xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships" xmlns:p="http://schemas.openxmlformats.org/presentationml/2006/main">
  <p:cSld>
    <p:spTree>
      <p:nvGrpSpPr>
        <p:cNvPr id="1" name=""/>
        <p:cNvGrpSpPr/>
        <p:nvPr/>
      </p:nvGrpSpPr>
      <p:grpSpPr>
        <a:xfrm>
          <a:off x="0" y="0"/>
          <a:ext cx="0" cy="0"/>
          <a:chOff x="0" y="0"/>
          <a:chExt cx="0" cy="0"/>
        </a:xfrm>
      </p:grpSpPr>
      <p:sp>
        <p:nvSpPr>
          <p:cNvPr id="4" name="Title 3">
            <a:extLst>
              <a:ext uri="{FF2B5EF4-FFF2-40B4-BE49-F238E27FC236}">
                <a16:creationId xmlns:a16="http://schemas.microsoft.com/office/drawing/2014/main" id="{6E4D81EF-7C57-AA4B-A4BB-D3097564DD5F}"/>
              </a:ext>
            </a:extLst>
          </p:cNvPr>
          <p:cNvSpPr>
            <a:spLocks noGrp="1"/>
          </p:cNvSpPr>
          <p:nvPr>
            <p:ph type="title"/>
          </p:nvPr>
        </p:nvSpPr>
        <p:spPr/>
        <p:txBody>
          <a:bodyPr/>
          <a:lstStyle/>
          <a:p>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>Slide</a:t>
            </a:r>
          </a:p>
        </p:txBody>
      </p:sp>
      <p:sp>
        <p:nvSpPr>
          <p:cNvPr id="5" name="Content Placeholder 4">
            <a:extLst>
              <a:ext uri="{FF2B5EF4-FFF2-40B4-BE49-F238E27FC236}">
                <a16:creationId xmlns:a16="http://schemas.microsoft.com/office/drawing/2014/main" id="{901565FB-C48D-A640-8DDF-8B4D89204E12}"/>
              </a:ext>
            </a:extLst>
          </p:cNvPr>
          <p:cNvSpPr>
            <a:spLocks noGrp="1"/>
          </p:cNvSpPr>
          <p:nvPr>
            <p:ph idx="1"/>
          </p:nvPr>
        </p:nvSpPr>
        <p:spPr/>
        <p:txBody>
          <a:bodyPr/>
          <a:lstStyle/>
          <a:p>
            <a:pPr marL="0" indent="0">
              <a:buNone/>
            </a:pPr>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>Paragraph.</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr marL="514350" indent="-514350">
              <a:buFont typeface="+mj-lt"/>
              <a:buAutoNum type="arabicPeriod"/>
            </a:pPr>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>Top-level</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr marL="457200" lvl="1" indent="0">
              <a:buNone/>
            </a:pPr>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>Continuation</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr marL="914400" lvl="1" indent="-457200">
              <a:buFont typeface="+mj-lt"/>
              <a:buAutoNum type="arabicPeriod"/>
            </a:pPr>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>Sub-list</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr marL="914400" lvl="2" indent="0">
              <a:buNone/>
            </a:pPr>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>With continuation</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr marL="914400" lvl="1" indent="-457200">
              <a:buFont typeface="+mj-lt"/>
              <a:buAutoNum type="arabicPeriod"/>
            </a:pPr>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>(still sub-list)</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr lvl="2"/>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>continuation</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr lvl="2"/>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>list</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr marL="514350" indent="-514350">
              <a:buFont typeface="+mj-lt"/>
              <a:buAutoNum type="arabicPeriod"/>
            </a:pPr>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>(back to top-level)</a:t>
            </a:r>
          </a:p>
          <a:p>
            <a:pPr marL="0" indent="0">
              <a:buNone/>
            </a:pPr>
            <a:r>
              <a:rPr lang="en-US" dirty="0"/>
              <a:t>Paragraph.</a:t>
            </a:r>
          </a:p>
        </p:txBody>
      </p:sp>
    </p:spTree>
    <p:extLst>
      <p:ext uri="{BB962C8B-B14F-4D97-AF65-F5344CB8AC3E}">
        <p14:creationId xmlns:p14="http://schemas.microsoft.com/office/powerpoint/2010/main" val="1569627031"/>
      </p:ext>
    </p:extLst>
  </p:cSld>
  <p:clrMapOvr>
    <a:masterClrMapping/>
  </p:clrMapOvr>
</p:sld>
```
</details>

Happily, this example shows the indentation we want on the second level (where "sub-list" lines up with "with continuation"), so I can copy what's done for that level to get the behaviour we want.

jgm:
Unrelated question:  I've noticed that part of your previous changes to the pptx tests have considerably multiplied the amount of time and disk space these tests take.  Almost all the tests now run in four different configurations (with different reference docs).  All told, the pptx test files take up almost 8 MB (and all of this goes into the source tarball).

Question:  Would it be sufficient to test a subset of the test cases with these four configurations?  That would really help us to slim down the tarball.  (In the past we've sometimes hit up against Hackage upload limits.)

undergroundquizscene:
> Unrelated question: I've noticed that part of your previous changes to the pptx tests have considerably multiplied the amount of time and disk space these tests take. Almost all the tests now run in four different configurations (with different reference docs). All told, the pptx test files take up almost 8 MB (and all of this goes into the source tarball).

Ah, sorry! I hadn't thought about the space taken up by the files.

> Question: Would it be sufficient to test a subset of the test cases with these four configurations? That would really help us to slim down the tarball. (In the past we've sometimes hit up against Hackage upload limits.)

Absolutely – I'll make a separate PR for that this evening, it shouldn't take long at all.

undergroundquizscene:
I made some good progress on Wednesday! I've changed the writer's output to line up paragraphs and numbered lists, which mostly matches the behaviour I've observed from PowerPoint. (The one exception is that the indentation used on the first level didn't match the rest, so I've changed it.) I've also updated the master in the default reference doc, to line up bulleted lists in the same way as regular paragraphs and numbered lists.

This means that anyone using the default reference doc will get nicely lined-up continuation paragraphs. However, people using their own reference doc may not have matched the bullet indentation (the `bodyStyle`) with the numbered list indentation (the `otherStyle`), especially if they made their reference doc by copying the current one. Also, people using a reference doc that's regularly used to make PowerPoint files directly may notice the small difference in how pandoc does the spacing for the first indentation level.

I think these drawbacks are probably ok: people using a custom reference doc probably know how to adjust the spacing for these levels within their reference doc, which pandoc will respect. And I don't anticipate it being important to people that the indentation increment for the first level is different from all the rest.

Here's what the example markdown above produces with these new changes:

New Screenshot

I'll fix the tests, and then I think this is ready to go! (@jgm Thanks so much for updating the test diffing to work on the pretty-printed output, btw – it's been a real pain to work with and I didn't know how to fix it)

undergroundquizscene:
I'll squash the last two commits, but I think keeping it as three separate commits makes sense to me: that sound ok?

jgm:
Looks good to me, thanks for all your work on this!

Issue #7335, opened 2021-05-31
`--to native` produces Haskell that `hindent` has problems with


kjambunathan:
This is a haskell file output from `writenative`: [testdoc-skeletal-2-output-native.hs.txt](https://github.com/jgm/pandoc/files/6567825/testdoc-skeletal-2-output-native.hs.txt)

When I put the file in `haskell-mode`--yes, I run emacs--and do `M-x hindent-reformat-region`, `hindent` complains with the follwoing message 

```
hindent-reformat-region-as-is: hindent: <interactive>:2:1: Parse error: ;
```
I thought I could get around the problem by passing `{ writerWrapText = WrapNone  }` to `writeNative'.   The problem persists.

So, there are two issues:

1. Native output shouldn't produce invalid haskell when it adds newlines to haskell output 
2. `writeNative` should honor  `writerWrapText` directive.

jgm:
I don't know about this.  The output by itself isn't something you can load into ghci anyway, because there's no top-level declaration.  If you add one, you'll have to indent the whole thing, but keeping the relative indents the same it will work:
```hs
x=
  [Plain [Cite [Citation {citationId = "Brandom1994", citationPrefix = [], citationSuffix = [], citationMode = NormalCitation, citationNoteNum = 0, citationHash = 1}] [Str "(Brandom",Space,Str "1994)"]]
  ,Plain [Cite [Citation {citationId = "Brandom1994", citationPrefix = [], citationSuffix = [], citationMode = AuthorInText, citationNoteNum = 0, citationHash = 2}] [Str "Brandom",Space,Str "(1994)"]]
  ,Plain [Cite [Citation {citationId = "Brandom1994", citationPrefix = [Str "see "], citationSuffix = [Str " chapter 7"], citationMode = NormalCitation, citationNoteNum = 0, citationHash = 3}] [Str "(see Brandom",Space,Str "1994,",Space,Str "chap.",Space,Str "7)"]]
  , Div ("refs",["references"],[])
   [Div ("ref-Brandom1994",[],[])
    [Para [Str "Brandom,",Space,Str "Robert.",Space,Str "1994.",Space,Emph [Str "Making",Space,Str "It",Space,Str "Explicit"],Str ".",Space,Str "Harvard",Space,Str "University",Space,Str "Press."]]]]
```
I don't feel too strongly about this, though. We could move the commas to the end of the previous lines.
[...]

PR #7584, opened 2021-09-20
Use pretty-simple to create native output

### LINKED ISSUE #7580: Proposal: Use pretty-simple to render native AST
Currently we use our own renderer.  Apparently it causes some difficulties for hindent (#7335); it also produces output that can be hard to read and diff for long paragraphs.  Why not use something standard and off-the-shelf?

jgm:
Previously we used our own homespun formatting.
But this produces over-long lines that aren't ideal
for diffs in tests. Easier to use something off-the-shelf
and standard.

Closes #7580. Note that this doesn't help with #7335, since
commas are still used at the beginning of lines in pretty-show's
output. But I'm inclined to think #7335 is just a problem with
hindent; this is perfectly legitimate Haskell syntax.

In a first effort, we used pretty-show for this, but pretty-simple
has more configuration options than pretty-show (e.g. it allows
 2-space indent and configurable line width), and seems to give better 
results.  It is also around 20% faster.

Note that with this change the native writer is considerably
slower than before (by a factor of about 10).  I don't think this is a
problem, since this writer is just used for diagnostics and tests; for
performance sensitive serialization, json should be used.

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -583,7 +583,8 @@ library
                  xml-conduit           >= 1.9.1.1  && < 1.10,
                  unicode-collation     >= 0.1.1    && < 0.2,
                  zip-archive           >= 0.2.3.4  && < 0.5,
-                 zlib                  >= 0.5      && < 0.7
+                 zlib                  >= 0.5      && < 0.7,
+                 pretty-simple         >= 4.0      && < 4.1
   if os(windows) && arch(i386)
      build-depends: basement >= 0.0.10,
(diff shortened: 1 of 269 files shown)


jgm:
Supersedes #7581.

jgm:
pretty-simple is a also significantly faster than pretty-show.

jgm:
Now I'm leaning to preferring pretty-show, just because it doesn't add any dependencies (skylighting already depends on it).

PR #7588, opened 2021-09-22
add abstract to default html template

jmbuhr:
This adds an (optional) abstract to the default html template to go along with the abstract in other formats (LaTeX, ConTeXt, AsciiDoc, and docx, see https://pandoc.org/MANUAL.html#metadata-variables). It is inspired by the default html template of Rmarkdown (https://github.com/rstudio/rmarkdown/blob/main/inst/rmd/h/default.html) and makes switching formats more seamless.

(Originally wrongly PRd at https://github.com/jgm/pandoc-templates/pull/17)


data/templates/default.html4
```diff
@@ -46,6 +46,12 @@ $endfor$
 $if(date)$
 <h3 class="date">$date$</h3>
 $endif$
+$if(abstract)$
+<div class="abstract">
+<p class="abstract">Abstract</p>
```
jgm, line 51: I would remove this `<p>` element. After all, an abstract can contain multiple paragraphs, lists, etc.  If the abstract contains block-level content, it will be rendered with `<p>` tags and then you'll get invalid HTML with this setup.

data/templates/styles.html
```diff
@@ -161,6 +161,16 @@
 $if(csl-entry-spacing)$
   margin-bottom: $csl-entry-spacing$;
 $endif$
+$if(abstract)$
+p.abstract{
+  text-align: center;
+  font-weight: bold;
```
jgm, line 167: If we do include this, I'd prefer a more vanilla formatting.  Normally abstracts in journals are not boldface.

jgm:
Sorry, I hadn't noticed these.  Using a div is good.
I don't think I like the `<p>Abstract</p>` part, though.  It's a general design principle in pandoc to avoid English labels, since people write in many different languages.  Better to simply include the abstract itself without the label.  In most cases it should be obvious that it is an abstract; if not, the word "Abstract" can always be added to the abstract in the metadata, or some custom CSS could be used that adds this content.

jmbuhr:
I agree with this sentiment and it is something I hadn't thought about, though not including the word abstract would make it inconsistent with the default pdf / tex template:

image

On the other hand, the docx template also does not include the word abstract:

image

jgm:
I'm merging but this is going to need some changes which I'll make separately.

PR #7602, opened 2021-10-01
epub: Add EPUB3 subject metadata (authority/term)

nuew:
This adds the ability to specify EPUB3 specific refinements on the subject tag, specifically `authority` and `term`.

Use of the plain text subject metadata, or use in EPUB2 mode will function as before.


jgm:
Looks good ,thanks!

PR #7606, opened 2021-10-05
Avoid blockquote when parent style has more indent

mbrackeantidot:
When a paragraph has an indentation different from the parent (named)
style, it used to be considered a blockquote. But this only makes sense
when the paragraph has more indentation. When a paragraph has less
indentation than its parent style, it probably shouldn't become a blockquote.
So this merge adds a check for the indentation of the parent style.


jgm:
Thanks, this is great.

jgm:
@mbrackeantidot - do you think that the issue mentioned here could be due to this change?
<https://groups.google.com/d/msgid/pandoc-discuss/d5e31c5d-b3a8-43d9-bc86-d38f11429c6bn%40googlegroups.com>

jgm:
I've fixed it

mbrackeantidot:
I see. I probably misinterpreted firstLine and hanging by not considering that they're always positive. Thanks for fixing @jgm .

PR #7607, opened 2021-10-05
Support for <indexterm>s when reading DocBook

### LINKED ISSUE #7427: <indexterm> support in DocBook reader?
The documentation within the DocBook reader [suggests](https://github.com/jgm/pandoc/blob/c8ab8bccf25ad01b9a98b2e59c84c151c5b0c371/src/Text/Pandoc/Readers/DocBook.hs#L211) that `<indexterm>` is supported. However, when I have `<indexterm>` elements, they are interpreted as just any regular old phrase.

Here's an example of 2 index terms in a DocBook source fragment:

```
<primary>IM</primary>
</indexterm>
<indexterm>
<primary>Instant Messaging</primary>
</indexterm>
```

These turn up in the JSON AST without anything to hook on for a filter:

```
{"t":"Str","c":"IM"},{"t":"SoftBreak"},{"t":"Str","c":"Instant"},{"t":"Space"},{"t":"Str","c":"Messaging"}
```

I'm using Pandoc to convert DocBook into LaTeX, and it works surprisingly well (except for a few Lua filters I had to write). `<indexterm>` support is, for now, the one missing piece. If you're interested to have this supported and not interested in implementing this (this being open source and all and you being free to spend your time as you please), I'll consider submitting a patch request.

bigsmoke:
This PR resolves #7427.  This was my first Haskell patch ever, so I'm sure that my additions, although functional, are an eyesore to seasoned Haskell coders. Please feel even more free than you usually would to correct/criticize my code.

### DIFF
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -316,7 +316,7 @@ List of all DocBook tags, with [x] indicating implemented,
 [ ] postcode - A postal code in an address
 [x] preface - Introductory matter preceding the first chapter of a book
 [ ] prefaceinfo - Meta-information for a Preface
-[ ] primary - The primary word or phrase under which an index term should be
+[x] primary - The primary word or phrase under which an index term should be
     sorted
 [ ] primaryie - A primary term in an index entry, not in the text
 [ ] printhistory - The printing history of a document
@@ -385,7 +385,7 @@ List of all DocBook tags, with [x] indicating implemented,
 [o] screeninfo - Information about how a screen shot was produced
 [ ] screenshot - A representation of what the user sees or might see on a
     computer screen
-[ ] secondary - A secondary word or phrase in an index term
+[x] secondary - A secondary word or phrase in an index term
 [ ] secondaryie - A secondary term in an index entry, rather than in the text
 [x] sect1 - A top-level section of document
 [x] sect1info - Meta-information for a Sect1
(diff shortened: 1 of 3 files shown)


src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1079,6 +1079,17 @@ elementToStr :: Content -> Content
 elementToStr (Elem e') = Text $ CData CDataText (strContentRecursive e') Nothing
 elementToStr x = x
 
+indextermNaryTextAsAttr :: Text -> Element -> Maybe (Text, Text)
+indextermNaryTextAsAttr n e = case findChild q e of
+        Nothing -> Nothing
+        Just naryEl -> Just (n, (strContent naryEl))
```
jgm, line 1085: Here it's idiomatic to omit the parentheses around `(strContent naryEl)`.

jgm:
Should we include the pagenum attribute?

What about `see` and `seealso` elements (there can be multiple of the latter)?

jgm:
A test would be nice, too!
Easiest would be a "command test", since there isn't currently a Tests.Reader.DocBook module.

bigsmoke:
> Should we include the pagenum attribute?

I don't think so. That is only used within `<index>` sections, which, like we discussed in #7608, makes most sense to continue to ignore in Pandoc.
 
> What about `see` and `seealso` elements (there can be multiple of the latter)?

I will follow up with a new PR request with (partial) support for `<see>` and `<seealso>` elements. They will be more straight-forward and less flaky to express as attributes to the span than `<primar>`, `<secondary>` and `<tertiary>`, because the latter actually have a mixed content model (which AsciiDoc doesn't have, so that it doesn't become a problem for me personally)..

> A test would be nice, too!
> Easiest would be a "command test", since there isn't currently a Tests.Reader.DocBook module.

I will included a test too. Thank you for the pointer! ☺

bigsmoke:
I have improved the patch. You stimulated me to go a bit further and also do the Right Thing™ with mixed-content child element of `<indexterm>`, including a test case to ensure that the existing `strContentRecursive` that I reused indeed did what I hoped it would: to flatten the mixed content of, for example, a `<primary>` element into a string suitable for use in an attribute.

Thanks for sticking with me, @jgm ☺

bigsmoke:
I reworded one of my commit messages and did a `git push --force-with-lease`, because one of the GitHub sanity check choked on its length 5 days ago.

jgm:
Great, thanks.

PR #7609, opened 2021-10-05
Keep <index> elements from DocBook source in AST

### LINKED ISSUE #7608: Pass through <index> elements from DocBook sources
`<index>` elements are currently ignored. However, when going from DocBook to LaTeX to print, this may be very necessary (and indeed is in my own publication pipeline.

`<index>` is kept in the AST when reading DocBook.

bigsmoke:
This resolves #7608. Let me know if this bites anything.


jgm:
This does two things:  instead of skipping the index element, we get

1. a section with an identifier (which can be used to locate the index in the target format)
2. the contents of the section -- whatever getBlock does with the children of the index element

I'm not sure about the ramifications of 2; when you try this with real docbook documents, what kind of contents get placed in the section?

bigsmoke:
For the same reason that I closed issue #7608, let me close this PR also.

PR #7664, opened 2021-11-03
Update bash_completion.tpl

StaticPH:
Update bash-completion and fix possible environment pollution

- Declare `highlight_styles` as a function-local scope variable; keeps `highlight_styles` out of the shell session environment when sourcing completion from a file rather than adding `eval "$(pandoc --bash-completion)"` to `.bashrc`
- Add argument completion for `--print-highlight-style`, `--eol`, and `--markdown-headings`

### DIFF
--- a/data/bash_completion.tpl
+++ b/data/bash_completion.tpl
@@ -4,7 +4,7 @@
 
 _pandoc()
 {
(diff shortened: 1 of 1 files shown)


jgm:
Great, thanks

Issue #7656, opened 2021-11-03
pandoc --columns not function in sub files


ajeep8:
If these's a pipe table in main.md file, it will wrap lines only when line content length >72.
so if the line content lenght is long but <=72, pandoc to pdf, the table will expand far to pdf page right and can not see all the content.

I can walk around this problem using:
```
pandoc --columns=10 main.md -o main.pdf
```
(set columns a very small value than 72) the pipe table line will wrap even the line content<=72.

But, if the pipe table is in sub files, this trick will not function.

I test 3 filters:
[lua-filters](https://github.com/pandoc/lua-filters):include-files.lua: sub file pipe table only wrap when pipe table line content >80, even I set --columns=10
[panda.lua](https://github.com/CDSoft/panda): same as include-files.lua
[pandoc-include](https://github.com/DCsunset/pandoc-include): sub file pipe table only wrap when pipe table line content >72, even I set --columns=10

I didn't find any process for columns or table in these 3 filter source code, so maybe bug in pandoc? I'm using pandoc 2.14.2

jgm:
This is basically the same issue as pandoc/lua-filters#195, no?
As noted there, the underlying issue is that pandoc.read (which these filters use to parse the contents of the subfiles) doesn't take reader options.  That could be changed, I imagine, but @tarleb could comment further.

One more reason a native support for includes in markdown would be nice, see #553.

tarleb:
We could, as a first step, allow to use the values from the `PANDOC_READER_OPTIONS ` global in `read`, but the global is read-only.  Adding a constructor for a read- and writeable object, which would make it possible to use options different from default/global options, would be a little more work, but not hard either.

jgm:
Here's a thought.  We have From/ToJSON instances for ReaderOptions and all the derivative types it uses.  Couldn't you piggy-back on those?  That is: add Text.Pandoc.Lua.Marsrhaling.Value for aeson Value, and then use the To/FromJSON instances to convert to and from Value.  That gives you "free" bidirectional marshalling that is guaranteed to match up with the JSON marshalling.

tarleb:
We did that for our first Lua filters draft implementation, and it worked reasonably well IIRC. I still have [hslua-aeson](https://hackage.haskell.org/package/hslua-aeson) lying around for this purpose, but it doesn't support aeson 2.0 yet. It would be helpful to marshal Extensions, I think. I'll give it a go again.

ajeep8:
[...]

PR #7667, opened 2021-11-04
Lua: allow to pass custom reader options to `pandoc.read`

tarleb:
Reader options can now be passed as an optional third argument to
`pandoc.read`. The object can either be a table or
`PANDOC_READER_OPTIONS`; the latter ensures that the same options used
for main input parsing are re-used to parse the input to `pandoc.read`.

Closes: #7656

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -2742,7 +2742,33 @@ format, and functions to filter and modify a subtree.
 [`sha1`]{#pandoc.sha1}
 
 :   Alias for [`pandoc.utils.sha1`](#pandoc.utils.sha1)
-    (DEPRECATED).
+    (DEPRECATED, use `pandoc.utils.sha1` instead).
+
+## Other constructors
+
+[`ReaderOptions (opts)`]{#pandoc.readeroptions}
+
+:   Creates a new [ReaderOptions] value.
+
+    Parameters
+
+    `opts`:
+    :   Either a table with a subset of the properties of a
+        [ReaderOptions] object, or another ReaderOptions object.
+        Uses the defaults specified in the manual for all
+        properties that are not explicitly specified. Throws an
+        error if a table contains properties which are not present
+        in a ReaderOptions object. ([ReaderOptions]|table)
+
+    Returns: new [ReaderOptions] object
+
+    Usage:
+
+        -- copy of the reader options that were defined on the command line.
+        local cli_opts = pandoc.ReaderOptions(PANDOC_READER_OPTIONS)
+
+        -- default reader options, but columns set to 66.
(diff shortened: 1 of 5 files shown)


tarleb:
Might need a little more work, e.g. documentation. I'm also not sure whether we should provide a `pandoc.ReaderOptions` constructor and require the argument to have that type:

```lua
local opts = pandoc.ReaderOptions {tab_stops = 2}    -- or, e.g., pandoc.ReaderOptions(PANDOC_READER_OPTIONS)
return pandoc.read('*CommonMark*', 'commonmark', opts)
```

One advantage of that would be that it is easier to document.

jgm:
I suppose another advantage of pandoc.ReaderOptions would be that you could throw an error on unknown options (giving them feedback if they get the field names wrong).

The important thing is to get as much consistency as possible in the interface.  E.g., if you have to do something similar with Attributes, and can't just pass in a table, it could be a reason to do the same here.

tarleb:
Most things are very permissive: whenever an `Attr`, `Attributes`, or `ListAttributes` is expected, it's also possible to pass in an equivalent table or even just an ID string. A Lua string can be used as a plain string, as `Str`, or as `Plain [Str _]`. If a list is expected, we generally allow a single item instead, which is then treated as a singleton list. So I'd lean towards at least allowing tables to be used instead of full `pandoc.ReaderOptions` elements.

On the other hand, we don't support that yet for Citations, but using `{id='einstein1905', mode='NormalCitation'}` could be convenient at times. Maybe I should open a new issue for this question.

jgm:
That's okay with me -- should I merge this then?

tarleb:
I've added a `pandoc.ReaderOptions` constructor and changed the code towards a better and more maintainable method of marshalling Lua tables into ReaderOptions tables.

Docs need updating, but good to merge otherwise.

jgm:
I'll keep this open pending documentation (otherwise we're more likely to forget).

tarleb:
OK, done.

jgm:
Question:  Is there a reason there's not a Pushable instance for ReaderOptions?  If there were, I could just have ReaderOptions be the optional second parameter of Reader (in custom readers) rather than using the gloabl PANDOC_READER_OPTIONS.

tarleb:
> Question:  Is there a reason there's not a Pushable instance for ReaderOptions?  If there were, I could just have ReaderOptions be the optional second parameter of Reader (in custom readers) rather than using the gloabl PANDOC_READER_OPTIONS.

Not really.  I wanted to avoid any confusion on whether `push` would act
like `pushReaderOptions` or like `pushReaderOptionsReadonly`, but the
expectation would probably to have a modifiable object.

Having a second parameter seems like the nicer interface. I'll add a
Pushable instance in a minute.

PR #7670, opened 2021-11-05
Add disableLayout variable in revealjs template

cderv:
This allows to modify it using Pandoc variable. Default value is correctly set to false by Pandoc.

This follows change in 
https://github.com/jgm/pandoc/commit/14b2eb2aeb21b3dee7d07c7348ebd2c586d6a866#diff-ecb2d745902425cf54bfa01ba4522c26288c56cf06be7be034312601f2129b4bR137

where `disableLayout` was added in template but was hard coded to FALSE. 

Unless there is a good reason to do that, I think it should be a variable to allow modify the value. Currently, it requires to patch the template to achieve that.

### DIFF
--- a/data/templates/default.revealjs
+++ b/data/templates/default.revealjs
@@ -137,7 +137,7 @@ $endif$
 
         // Disables the default reveal.js slide layout (scaling and centering)
         // so that you can use custom CSS layout
-        disableLayout: false,
+        disableLayout: $disableLayout$,
 
         // Vertical centering of slides
         center: $center$,


jgm:
What does this allow you to do, exactly?  We're already including the default reveal.js css, plus some pandoc css that is needed for pandoc elements.  How does it change things if you set this variable to True?

cderv:
This configuration will allow to disable auto scaling and centering feature of revealjs: https://revealjs.com/presentation-size/#byol

I am setting this variable to TRUE when debugging sometime and for specific presentation with content that does not play well with those reveal features and where custom Layout is needed.

So it is a revealjs configuration that you may need in specific case and that Pandoc does not allow to change currently.

jgm:
OK, thanks!

fkohrt:
@cderv Just FYI, one can also change a reveal variable using the query string, e.g. `presentation.html?disableLayout=true`

PR #7680, opened 2021-11-11
Lua: load `re` module available into global of the same name

tarleb:


jgm:
@tarleb I'm noticing that now, when I run the test suite and I have lpeg installed in my regular lua path, I get errors: e.g.,
```
        Exception: PandocLuaError "error loading module 'lpeg' from file '/opt/homebrew/lib/lua/5.4/lpeg.so':\n\tdlopen(/opt/homebrew/lib/lua/5.4/lpeg.so, 6): Symbol not found: _lua_getiuservalue\n  Referenced from: /opt/homebrew/lib/lua/5.4/lpeg.so\n  Expected in: flat namespace\n in /opt/homebrew/lib/lua/5.4/lpeg.so"
        CallStack (from HasCallStack):
          error, called at test/Tests/Lua.hs:260:15 in main:Tests.Lua
          runLuaTest, called at test/Tests/Lua.hs:218:43 in main:Tests.Lua
        Use -p '/module '\''re'\'' is available/' to rerun this test only.
```
Can you see what is going on here?
Setting LUA_PATH and LUA_CPATH to null fixes it.

tarleb:
It looks like LUA_CPATH includes libraries that were compiled against Lua 5.4. Can you pin to Lua 5.3?

I'm preparing to switch to Lua 5.4 at some point, but it's not ready yet.

tarleb:
Or could hslua/hslua#94 be to blame? I'm not sure if that's a valid bug.

jgm:
Could we make it work in the following way?

- By default the variables `lpeg` and `re` are loaded with the versions built into pandoc
- The system version of lpeg (in the regular lua search path) is only loaded if the script explicitly does `require "lpeg"`

I'm thinking the present behavior will lead to a lot of problems, e.g. filters that work on one system won't work on another because a different version of lua is installed.

tarleb:
True, that's a serious problem. I favor the current mechanism, so I've pushed a change which makes module loading more robust. However, I can still implement the proposed alternative if you prefer.

jgm:
I think it's just surprising for pandoc to be loading external Lua libraries even when the script doesn't explicitly say to do this -- so I'd prefer the alternative, yes.

jgm:
I can confirm, though, that your latest fix allows the test suite to succeed even when my lua path points to 5.4.

tarleb:
:+1: That makes sense; I pushed the change.

jgm:
@tarleb Even after cd91f72843359c5305842fa8afbec4a2d72629fa I'm still getting this in the test suite.  As I understand the change, this shouldn't be happening, should it?
```
        Exception: PandocLuaError "error loading module 'lpeg' from file '/opt/homebrew/lib/lua/5.4/lpeg.so':\n\tdlopen(/opt/homebrew/lib/lua/5.4/lpeg.so, 6): Symbol not found: _lua_getiuservalue\n  Referenced from: /opt/homebrew/lib/lua/5.4/lpeg.so\n  Expected in: flat namespace\n in /opt/homebrew/lib/lua/5.4/lpeg.so"
        CallStack (from HasCallStack):
          error, called at test/Tests/Lua.hs:270:15 in main:Tests.Lua
          runLuaTest, called at test/Tests/Lua.hs:232:45 in main:Tests.Lua
```

jgm:
Note, the failures occur not just on things that test 'require', but on e.g. "inline elements can be round-tripped through the lua stack"

tarleb:
I'm stumped. I've been staring at the code now for the past 10 minutes, but am at a complete loss as to why this happens. Do the tests pass if that commit is reverted?

jgm:
By the way, I can reproduce outside the tests, using e.g.
```
% stack exec pandoc -- -L t.lua
hi
Error running filter t.lua:
error loading module 'lpeg' from file '/opt/homebrew/lib/lua/5.4/lpeg.so':
	dlopen(/opt/homebrew/lib/lua/5.4/lpeg.so, 6): Symbol not found: _lua_getiuservalue
  Referenced from: /opt/homebrew/lib/lua/5.4/lpeg.so
  Expected in: flat namespace
 in /opt/homebrew/lib/lua/5.4/lpeg.so
```

tarleb:
I managed to reproduce it, too. Still don't understand it, but now I can debug it. Sorry for the mess.

jgm:
Just tested reverting that commit -- indeed, the problem goes away.

jgm:
Looks like on line 89 of T.P.Lua.Init we're getting an ErrRun status rather than OK.

tarleb:
Replacing the first `Lua.pop` after that with `Lua.forcePeek (Lua.peekString Lua.top) >>= Lua.failLua` shows that it tries to load the library from the `LUA_CPATH`. I really don't get what's going on here. Maybe some linking issues? Maybe I need to rename the exported `luaopen_lpeg` function.

tarleb:
Oh, I get it now. It's `re` that's to blame, because it contains `require'lpeg'. I guess I can work around that by loading both modules normally, but then remove them from the registry of loaded modules.

jgm:
Or...we could just not load `re` by default.
Let people load it with `require` if they need it.

tarleb:
I'm fine with either solution. I'll create a PR for both options.

jgm:
Won't loading them normally still raise the error if it's installed with lua 5.4?

jgm:
I'm confused -- I don't see the `re.lua` code anywhere in the `lpeg` package on Hackage.  Where is pandoc supposed to find this?

tarleb:
A `re.h` file is generated from `cbits/lpeg-1.0.2/re.lua` and then used in `re.c`. See the Makefile in `cbits`. Possibly not the best design, but it seemed like an easy solution.

jgm:
Ah, very sneaky!  Well then, what about this solution?  Just remove the `require "lpeg"` before you create re.h.  Replace it with something like `local m = lpeg`.

jgm:
I wasn't sure why the lpeg library didn't just load re in the normal way, by using a normal lua `loadstring` (or whatever it's called) and passing in a bytestring of the text of `re.lua`?
If you did that, wouldn't this problem go away?  Why treat it like a C library when it isn't?

tarleb:
There were a couple of reasons:
- It makes it easier to ship it as a black box. The library user doesn't need to know how to load the string.
- The interface is uniform; each library is loaded by calling the loader function.
- If the library is configured to rely on shared libraries, the loader can just throw an error. It was less clear to me what the bytestring should contain in that case.

I debated myself a good while over this. Still not sure whether it was the right decision.

PR #7682, opened 2021-11-11
Quick and dirty fix to allow `custom-style` attributes for inline images

cpkio:
This fix allows Image to pass custom attributes to the element, thus allowing to style inline images  in MS Word. Using a filter (Moonscript):
```
  if element.classes[1] == 'icon'
    table.insert(element.attributes, { "custom-style", "StyleName" })
  return element
```
you can assign MS Word styles to Images.

Please review. It's dirty and I look at Haskell code only a week long, maybe it can be done smarter.

### DIFF
--- a/src/Text/Pandoc/Writers/Docx.hs
+++ b/src/Text/Pandoc/Writers/Docx.hs
@@ -1231,9 +1231,15 @@ inlineToOpenXML' opts (Link _ txt (src,_)) = do
                         M.insert src i extlinks }
               return i
   return [ Elem $ mknode "w:hyperlink" [("r:id",id')] contents ]
-inlineToOpenXML' opts (Image attr@(imgident, _, _) alt (src, title)) = do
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -1313,8 +1319,8 @@ inlineToOpenXML' opts (Image attr@(imgident, _, _) alt (src, title)) = do
               , spPr
               ]
             ]
-        imgElt = mknode "w:r" [] $
-          mknode "w:drawing" [] $
+        imgElt = mknode "w:r" [] [ mknode "w:rPr" []
+                                   [ mknode "w:rStyle" [("w:val", fromStyleId inlineImageStyle)] ()], mknode "w:drawing" [] $
```
jgm, line 1323: The problem with this is that it always adds a w:rStyle, even when there isn't a style -- see the test failures.

PR #7700, opened 2021-11-18
[#7699] check :class: for custom roles

### LINKED ISSUE #7699: RST Reader does not check the :class: field for custom roles
**Explain the problem.**

The relevant RST documentation is under the [role directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html#custom-interpreted-text-roles):

> The following option is recognized by the "role" directive for most base roles:
>
> `class` : *text*
> Set the "classes" attribute value on the element produced (inline, or element associated with a base class) when the custom interpreted text role is used. If no directive options are specified, a "class" option with the directive argument (role name) as the value is implied. See the class directive above.

As shown in the example below, Pandoc 2.16.1 always uses the role name for the element class, and always ignores the `:class:` field in a custom role.

Example RST:

```rst
.. role: classy
  :class: myclass

:classy:`hello`
```

Pandoc command: `pandoc -f rst -t html5 -o test.html test.rst`

Output HTML5:

```html
<p><span class="classy">hello</span></p>
```

Expected output:

```html
<p><span class="myclass">hello</span></p>
```

**Pandoc version?**
2.16.1, Windows

willj-dev:
This PR fixes #7699. It includes three new tests:

**Custom role with class field**

Input:

```rst
.. role:: classy
    :class: myclass

:classy:`a`
```

This is expected to create a `Span` with a singleton class list: `["myclass"]`

**Custom role with a class field containing multiple whitespace-separated classes**

Input:

```rst
.. role:: classy
    :class: myclass1 myclass2
       myclass3

:classy:`a`
```

This is expected to create a `Span` with three classes: `["myclass1", "myclass2", "myclass3"]`

**Custom role with an inherited class field**

Input:

```rst
.. role:: classy
    :class: myclass1
.. role:: classier(classy)
    :class: myclass2

:classier:`a`
```

This is expected to create a `Span` with two classes: `["myclass2", "myclass1"]`

### DIFF
--- a/src/Text/Pandoc/Readers/RST.hs
+++ b/src/Text/Pandoc/Readers/RST.hs
@@ -919,14 +919,22 @@ addNewRole roleText fields = do
         (baseRole, baseFmt, baseAttr) =
                getBaseRole (parentRole, Nothing, nullAttr) customRoles
         fmt = if parentRole == "raw" then lookup "format" fields else baseFmt
-        annotate :: [Text] -> [Text]
-        annotate = maybe id (:) $
-            if baseRole == "code"
-               then lookup "language" fields
-               else Nothing
-        attr = let (ident, classes, keyValues) = baseAttr
-        -- nub in case role name & language class are the same
-               in (ident, nub . (role :) . annotate $ classes, keyValues)
+
+        updateClasses :: [Text] -> [Text]
+        updateClasses oldClasses = let
+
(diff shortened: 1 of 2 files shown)


willj-dev:
(sorry for the force push, used the wrong email for my original commit!)

jgm:
Excellent, thanks!

PR #7701, opened 2021-11-19
Lua: fix global module loading

tarleb:
Prevent issues if Lua is misconfigured on the system.

### DIFF
--- a/src/Text/Pandoc/Lua/Init.hs
+++ b/src/Text/Pandoc/Lua/Init.hs
@@ -14,10 +14,11 @@ module Text.Pandoc.Lua.Init
   ( runLua
   ) where
 
-import Control.Monad (forM_, when)
+import Control.Monad (forM, forM_, when)
 import Control.Monad.Catch (throwM, try)
 import Control.Monad.Trans (MonadIO (..))
 import Data.Data (Data, dataTypeConstrs, dataTypeOf, showConstr)
+import Data.Maybe (catMaybes)
 import HsLua as Lua hiding (status, try)
 import GHC.IO.Encoding (getForeignEncoding, setForeignEncoding, utf8)
 import Text.Pandoc.Class.PandocMonad (PandocMonad, readDataFile)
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Lua/Init.hs
```diff
@@ -78,23 +78,43 @@ initLuaState = do
...
             Lua.pop 1  -- ignore error message
             -- Try loading via the normal package loading mechanism.
             Lua.getglobal "require"
             Lua.pushName pkgname
             Lua.call 1 1  -- Throws an exception if loading failed again!
+            return False
 
         -- Module on top of stack. Register as global
         Lua.setglobal pkgname
+        return usedInternal
```
jgm, line 107: How about
```
return (usedInternal, pkgname)
```
and then avoid the `zip` on line 112 and the `_` on 113?

jgm: Or, perhaps even clearer, return `Just pkgname` instead of True and `Nothing` instead of False.

jgm:
I think I prefer this solution to the "no re" solution, now that I understand it.
Sorry, tests failed because of something I've since fixed, so you'll want to rebase.

src/Text/Pandoc/Lua/Init.hs
tarleb: Much nicer. :+1:

tarleb:
Updated and rebased. Tests pass, so I hope it's ok if I just merge.

jgm:
I pulled this in and I still get a test failure (this is on another machine):
```
  Lua
    Lua filters
      global modules
        module 'lpeg' is available via `require`:                                                                                   FAIL (0.40s)
          test/Tests/Lua.hs:225:
          expected: OK
           but got: ErrRun
          Use -p '/module '\''lpeg'\'' is available via `require`/' to rerun this test only.
        module 're' is available via `require`:                                                                                     FAIL (0.38s)
          test/Tests/Lua.hs:229:
          expected: OK
           but got: ErrRun
          Use -p '/module '\''re'\'' is available via `require`/' to rerun this test only.

2 out of 2947 tests failed (157.62s)
```
Note that on this machine (as on the one I was using before) I have lpeg installed but it's in lua 5.4.
If I reset LUA_CPATH and run the test suite again,  it passes.
Is this expected?  Or is it supposed to fall back to the in-built lpeg if there are problems loading the system one?  (I know we discussed this but can't remember what was decided.)

If it's expected behavior, I will work around it by having the test suite reset LUA_CPATH, so that tests can succeed regardless of the test machine's lua setup.

jgm:
I've changed the two failed tests so they reset the lua path and cpath before doing `require`. This ensures that what the test is testing is the same on every system, regardless of its Lua setup. @tarleb let me know if this doesn't seem like a good idea.

tarleb:
Thanks, that's the exact change I was planning to make after reading your first message.

Probably no longer relevant, but the version collisions can be worked around by setting `LUA_CPATH_5_3`; Lua should ignore `LUA_CPATH` if a version specific env var is set.

jgm:
So there's a commit that says:

>   Make loading of global LPeg modules more robust.
      Ignore errors if the normal package mechanism failed; this not only
      covers the case of modules being unavailable on the system, but also
      works if the modules are present, but fail to load for some reason.
      This makes the built-in package version a true fallback.

Has that behavior been superseded?  Or should it work that way and fall back to the inbuilt lpeg in this case?

tarleb:
That is indeed outdated now. I tried to describe the current behavior in `lua-filters.md`, but it may be a little too terse:

> Pandoc uses a built-int version of the library,
> unless it has been configured by the package maintainer to
> rely on a system-wide installation.

> Note that the result of `require 'lpeg'` is not necessarily
> equal to this value; the `require` mechanism prefers the
> system's lpeg library over the built-in version.

PR #7364, opened 2021-06-09
Use the new 'simpleFigure' builder function in the readers.

argent0:
## Reading input with figures

Reading figures is now supported for two formats natively and one format through a Lua filter. This functionality has to be enabled with a new [pandoc extension](https://pandoc.org/MANUAL.html#extensions): `native_figures`.

```
#### Extension: `native_figures` ####

Use pandoc's native `Figure` element for content inside `<figure>` tags, in the case of HTML, or `figure` environments, in case of LaTeX. This, in turn, allows some writers to produce more accurate representations of figures. It also allows the use of the `Figure` element in filters, for custom figure output.

This extension can be enabled/disabled for the following formats:

input formats
: `latex` `html`
```

The choice of an extension was made to introduce the new behavior with minimal disruption of the old one.

### Reading HTML 5 figures

This version of pandoc can now read figures from HTML into its internal representation:

**Now**
```
$ pandoc -f html+native_figures -t native
<figure class="important">
  <img src="../media/rId25.jpg" />
  <ul> <li> ITEM </li> </ul>
  <figcaption> CAP2 </figcaption>
</figure>
^D
[Figure ("",["important"],[]) (Caption Nothing [Plain [Str "CAP2"]])\
	[ Plain [Image ("",[],[]) [] ("../media/rId25.jpg","")]
	, BulletList [[Plain [Str "ITEM"]]]]]
```

The difference with the old handling is that, now, elements inside a figure are not limited to only one image:

**Before**
```
$ pandoc-before -f html -t native
...
^D
[Para [Image ("",[],[]) [Str "CAP2"] ("../media/rId25.jpg","fig:")]]
```

Notice the missing `ITEM`. Handling of attributes and classes has also become more granular and accurate. Notice the missing `important` class above.

### Reading LaTeX figures

It can also read LaTeX figures:

**Now**
```
pandoc -f latex+native_figures -t native
\begin{figure}
  \begin{subfigure}[b]{0.5\textwidth}
    \begin{subfigure}[b]{0.5\textwidth}
      \centering
      \includegraphics{test/media/rId25.jpg}
      \caption{CAP1.1}
    \end{subfigure}
    \begin{subfigure}[b]{0.5\textwidth}
      \centering
      \includegraphics{test/media/rId25.jpg}
      \caption{CAP1.2}
    \end{subfigure}
    \caption{CAP1}
    \label{fig:inner1}
  \end{subfigure}
  \begin{subfigure}[b]{0.5\textwidth}
[...]

PR #7703, opened 2021-11-20
Capture `alt-text` in JATS figures

tarleb:


tarleb:
This is a result of this year's GSoC and taken almost verbatimfrom  the respective commit by Aner in PR #7364. Only the test has been updated, as the native representation has changed since this was written.

Cc: @argent0

Issue #3177, opened 2016-10-23
Explicit Figure element in Block


jgm:
Currently we represent figures in the AST using this hack:  a figure is an `Image` whose title attribute starts with `fig:` and which is by itself in a `Para`.

Short of a full-featured figure environment in the AST, it would make sense to move to a less hacky representation:  a `Div` with class `figure` containing the image (which need not have a title starting with `fig:`).  This would involve changes to readers and writers.

Indeed, if we did this, we could support figures containing multiple images, via explicit Divs.

jgm:
Another advantage is that attributes could be added explicitly to the Div.

    <div class="figure floatRight">
    ![my image](img.jpg){.imageclass}
    </div>

See #3094.

hrehfeld:
Seconded, I just spent an hour figuring out why multiple images in a paragraph don't create a figure. Is there any way to create a figure right now (I guess creating Rawblocks will work?)

Relevant code is here? https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/HTML.hs#L450 Why is the match only for one image and not multiple ones in the first place?

jgm:
If multiple images were allowed, which one would form the figure's caption?  (There is only one caption.)  What would determine how the images are arrayed in the figure?  In retrospect, some kind of more explicit syntax for figures would have been desirable, and maybe that's the direction we should move in.

hrehfeld:
Layout: Hm, I only use html and latex backends, but both of those handle multiple images in a figure in a reasonable way without extra specification. 

Caption: None unless explicitely stated? That's legal in both html and latex iirc. However I believe Paras do not support extra data like attrs...

jgm:
+++ Hauke Rehfeld [Oct 24 16 16:42 ]:
>   Seconded, I just spent an hour figuring out why multiple images in a
>   paragraph don't create a figure. Is there any way to create a figure
>   right now (I guess creating Rawblocks will work?)

You can paste the images together into one image, I suppose,
or use a filter.

>   Relevant code is here?
>   [1]https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/HT
>   ML.hs#L450 Why is the match only for one image and not multiple ones in
>   the first place?

Good question.  I suppose that since I'm in a field where
pictorial figures aren't used much, I didn't realize how
common figures with multiple images and a single caption
are.

But how would it work to allow a paragraph with multiple
images (and nothing else) to form a figure?  From which
[...]

PR #7704, opened 2021-11-20
Support complex figures.

tarleb:
* It provides a specific representation for figures in the pandoc's AST.
* It uses the `SimpleFigure` pattern synonym to replace the previous
  construction:

  [Para [Image ("",[],[]) [Str "CAP2"] ("../media/rId25.jpg","fig:")]]


tarleb:
This is rebased and updated version of #7364. @argent0, can you take a look and check that I didn't changed something against your intentions?

There are mostly just small changes compared to the commit in #7364: The tests are updated to match the new native output, and the changes to the test suite have been reverted (as I think those would warrant a separate PR). Also, the Lua marshaling code for figures needed an update to match the new mechanism.

Thanks again for the great GSoC work @argent0!

jgm:
Note: When this gets merged, we will also need a new pandoc-types release, and it would be a good idea at that point to do jgm/pandoc-types#91 as well.

MANUAL.txt
```diff
@@ -3176,6 +3176,17 @@ In the `context` output format this enables the use of [Natural Tables
...
+
+Use pandoc's native `Figure` element for content inside `<figure>` tags, in the
+case of HTML, or `figure` environments, in case of LaTeX. This, in turn, allows
+some writers to produce more accurate representations of figures. It also
+allows the use of the `Figure` element in filters, for custom figure output.
+
+This extension can be enabled/disabled for the following formats:
+
+input formats
+: `latex` `html`
```
jgm, line 3189: Question:  do we really need to gate this under an extension?  What is the motivation for that?
An alternative would be to always use Figure elements to represent figures.

argent0: The motivation to gate this under an extension is to keep backward compatibility. Some users may still rely on the original representation in their lua-filters.

> An alternative would be to always use Figure elements to represent figures.

Or this extension could be turned on by default allowing users to fallback to the original behaviour.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -536,23 +538,40 @@ pPara = do
...
+             _ -> mzero
+
+pNativeFigure :: PandocMonad m => TagParser m Blocks
+pNativeFigure = try $ do
+  TagOpen tag attrList <- lookAhead $ pSatisfy (matchTagOpen "figure" [])
+  --let (ident, classes, kvs) = toAttr attr
+  contents <- pInTags tag (many $ Left <$> pInTags "figcaption" block <|> (Right <$> block))
+
+  let (captions, rest) = partitionEithers contents
+  -- I should capture the caption
```
argent0, line 573: The comments should be removed

argent0: What this code does is capture all the tags inside a figure but keeping the `<figcaptions>` separated.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -600,6 +600,8 @@ blockToMarkdown' opts (OrderedList (start,sty,delim) items) = do
 blockToMarkdown' opts (DefinitionList items) = do
   contents <- inList $ mapM (definitionListItemToMarkdown opts) items
   return $ mconcat contents <> blankline
+blockToMarkdown' opts (Figure (ident, classes, kvs) _ body) =
+  blockToMarkdown' opts (Div (ident, ["figure"] `union` classes, kvs) body)
```
argent0, line 604: Here I added the `"figure"` class to the `div`. I'd like to know how this aligns with the goal of keeping English words out of markdown.

src/Text/Pandoc/Writers/MediaWiki.hs
```diff
@@ -210,6 +211,9 @@ blockToMediaWiki x@(DefinitionList items) = do
         contents <- local (\s -> s { listLevel = listLevel s <> ";" }) $ mapM definitionListItemToMediaWiki items
         return $ vcat contents <> if null lev then "\n" else ""
 
+blockToMediaWiki (Figure (ident, classes, kvs) _ body) =
+  blockToMediaWiki (Div (ident, ["figure"] `DL.union` classes, kvs) body)
```
argent0, line 215: Ditto

src/Text/Pandoc/Writers/Muse.hs
```diff
@@ -276,6 +276,7 @@ blockToMuse (Table _ blkCapt specs thead tbody tfoot) =
     isSimple = onlySimpleTableCells (headers : rows) && all (== 0) widths
 blockToMuse (Div _ bs) = flatBlockListToMuse bs
 blockToMuse Null = return empty
+blockToMuse (Figure {}) = return empty
```
argent0, line 279: This old behaviour slipped through the cracks from the phase where `Figure` elements were simply ignored.
This should now change to the phase were `Figure` is treated as a `Div` as a default handling strategy:

```{haskell}
blockToMusc (Figure attr _ bs) = blockToMuse (Div attr bs)
```

src/Text/Pandoc/Writers/TEI.hs
```diff
@@ -191,6 +191,7 @@ blockToTEI _ HorizontalRule = return $
   selfClosingTag "milestone" [("unit","undefined")
                              ,("type","separator")
                              ,("rendition","line")]
+blockToTEI _ (Figure {}) = return empty
```
argent0, line 194: Ditto
```{haskell}
blockToTEI opts (Figure attr _ bs) = blockToTEI opts (Div attr bs)
```

src/Text/Pandoc/Writers/Texinfo.hs
```diff
@@ -254,11 +254,39 @@ blockToTexinfo (Table _ blkCapt specs thead tbody tfoot) = do
...
+  return $ text ("@float" ++ floatType body) $$ content $$ (
+      if isEmpty captionText
+         then empty
+         else inCmd "caption" captionText
+    ) $$ text "@end float"
+  where
+  floatType [SimpleFigure {}] = " Figure"
+  floatType [Table {}] = " Table"
+  floatType _ = ""
+
```
argent0, line 289: The modifications in the output here are backed by [this documentation](https://www.gnu.org/software/texinfo/manual/texinfo/html_node/_0040float.html).

src/Text/Pandoc/Writers/XWiki.hs
```diff
@@ -122,6 +122,10 @@ blockToXWiki (DefinitionList items) = do
   contents <- local (\s -> s { listLevel = listLevel s <> ";" }) $ mapM definitionListItemToMediaWiki items
   return $ vcat contents <> if Text.null lev then "\n" else ""
 
+blockToXWiki (Figure attr _ body) = do
+  content <- blockToXWiki $ Div attr body
+  return $ intercalate content ["(((\n", ")))"]
+
```
argent0, line 128: This creates a [group](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiSyntax/?syntax=2.1&section=Groups)

argent0 (changes requested):

MANUAL.txt
jgm: I'll comment in the main comments.

jgm:
@tarleb @argent0 I'm tempted to go for a wholesale move to using this new Figure constructor, rather than introducing it under an extension.

Yes, this would mean that some existing filters would need to be rewritten, but it would avoid pain in the future.  I worry that a situation in which some filters are presupposing Figure elements and others aren't might produce a lot of confusion.

If we did this, it would mean replacing all uses of the "implicit figures" pattern (`Para [Image ... "fig:"]`) with `Figure`.  It would be relatively easy to perform this switch in the readers; all we'd have to do is change `simpleFigure` so it produces a `Figure`.  The other piece would be revising the writers so they all handle a `Figure` intelligently.  That would be more work, I gather.

Thoughts?

argent0:
When I was writing this I had in mind two different concepts that became:

1. `simpleFigure`: a graph, a picture, with an optional caption. Position in the document relatively more important.
2. `Figure`: A sub-document, a document withing a document, with an optional paragraph. Position in the document relatively less important.

I was thinking in using a constructor (`Block`) for the simpleFigure, and most of the work towards that goal as already been done.

But, we would need two constructors.

jgm:
I would prefer to avoid having two separate figure concepts.  Why can't a simple figure just be a special case of `Figure` (with a single image as its contents, and a single Plain as its caption)?

tarleb:
@argent0 you still have full collaborator access to my fork, please don't

tarleb:
Or, if you prefer, you could also update your original PR. E.g.,

``` shell
git remote add -f tarleb https://github.com/tarleb/pandoc
git switch figures-gsoc
git reset --hard tarleb/figures-gsoc
git push --force-with-lease
```

I'd close this one then.

argent0:
> I would prefer to avoid having two separate figure concepts. Why can't a simple figure just be a special case of `Figure` (with a single image as its contents, and a single Plain as its caption)?

Yes, this can be a better alternative something like:

`Figure attr caption body`: for full figures

`Figure attr caption [Image ...]` for a `simpleFigure`,why would there be a need to restrict the caption to single `Para` though?

> @argent0 you still have full collaborator access to my fork, please don't hesitate to make changes if needed.

I'll make the contributions there then (rather than updating my PR).

jgm:
I'm actually not sure why we need any conceptual distinction between figure types at all.  There will be simpler figures and more complex ones, because sometimes the `[Block]` contained in the `Figure` element will be simple and sometimes it will be more complex.  But they are both figures in the same sense.

LaTeX just has one `figure` environment; similarly HTML.  The issue of whether figures float is a complex one; in LaTeX all figures float, though you can specify preferences for locations.  In HTML they don't, because it's not a paged format so floating doesn't make so much sense.  Pandoc is going to have to abstract from those differences.

mb21:
Is this going back to [this comment by despresc](https://github.com/jgm/pandoc/issues/3177#issuecomment-687670695)? I don't think this was ever discussed properly through...

argent0:
> Is this idea of "two figure types" going back to [this comment by despresc](https://github.com/jgm/pandoc/issues/3177#issuecomment-687670695)? I don't think this was ever discussed properly through...

Yes, it's partially based on that discussion. I was also mainly thinking about "articles" which are only a part of pandoc's scope.

mb21:
Ah, the [pandoc-types commit](https://github.com/argent0/pandoc-types/commit/5b32019746d1692c593c080f9a8ee45c7d25ecba) associated with this PR doesn't have a `Figure` AST element yet. So far we just have the pattern synonym, correct?

@jgm is the plan to merge this without having settled on the details of the `Figure` AST element, or? I'm happy to provide input if needed (but unfortunately currently short on time to get my hands dirty with the code).

tarleb:
The type constructor went through a good bit of evolution. I squashed all changes into a cleaned-up, single commit here: jgm/pandoc-types#94. We aimed to keep the focus of the GSoC on figures and didn't want to take tables into account at that time.

jgm:
I'm not sure where we stand on my original query, so reposting:

>  @tarleb @argent0 I'm tempted to go for a wholesale move to using this new Figure constructor, rather than introducing it under an extension.
>
> Yes, this would mean that some existing filters would need to be rewritten, but it would avoid pain in the future. I worry that a situation in which some filters are presupposing Figure elements and others aren't might produce a lot of confusion.
>
> If we did this, it would mean replacing all uses of the "implicit figures" pattern (Para [Image ... "fig:"]) with Figure. It would be relatively easy to perform this switch in the readers; all we'd have to do is change simpleFigure so it produces a Figure. The other piece would be revising the writers so they all handle a Figure intelligently. That would be more work, I gather.
>
> Thoughts?

As I understand it (I haven't looked back recently, so maybe things changed) this code does not change the `simpleFigure` builder to create a Figure block, and it continues to gate use of the Figure block under extensions.

tarleb:
I'd be in favor of an all-in approach.

jgm:
If we're doing all-in, then we'd need to make sure all writers handle Figure elements properly.

I don't think we're there yet, right?

(Eventually we'd also want to remove the old support for "single image in para with title beginning with fig:", though maybe it's better to keep this support in the writers for now...?  But in any case, proper functioning of Figure in the writer should not depend on this.)

tarleb:
@argent0 I think that, since this the PR contains the fruits of your work, it would be great if you could bring this to completion. However, if you think you might not have the bandwidth for this at the moment, then that's fine, too. Let me know in that case, and I can take it from there.

argent0:
> If we're doing all-in, then we'd need to make sure all writers handle Figure elements properly.
> I don't think we're there yet, right?

@jgm all writers handle Figure elements either in:

1. A format specific way.
2. A fallback `Div`-like way.

> (Eventually we'd also want to remove the old support for "single image in para with title beginning with fig:", though maybe it's better to keep this support in the writers for now...? But in any case, proper functioning of Figure in the writer should not depend on this.)

Concretely. Are we talking about changing the pattern from:

```{haskell}
pattern SimpleFigure :: Attr -> [Inline] -> Target -> Block
pattern SimpleFigure attr figureCaption tgt <-
    Para [Image attr figureCaption
        (isFigureTarget -> Just tgt)]  where
  SimpleFigure attr figureCaption tgt =
    Para [Image attr figureCaption (second ("fig:" <>) tgt)]
```

to

```{haskell}
pattern SimpleFigure :: Attr -> [Inline] -> Target -> Block
pattern SimpleFigure attr figureCaption tgt <-
    Figure attr (Caption Nothing [Para figureCaption]) [Para [Image ?? ?? tgt]]
  SimpleFigure attr figureCaption tgt =
    Figure attr (Caption Nothing [Para figureCaption]) [Para [Image ?? ?? tgt]]
```
? It doesn't look right to me.

jgm:
I wasn't thinking that `SimpleFigure` would need to change; rather, that we just wouldn't include clauses that match on `SimpleFigure` in the writers.  Only real `Figure` blocks would render as figures.  Does this make sense?  (`SimpleFigure` would no longer have any real use.)

On the reader side, the markdown reader would produce a Figure block if `implicit_figures` is enabled and an image is alone in its paragraph.

argent0:
It makes sense. Although currently `SimpleFigure` and `Figure` aren't handled _exactly_ the same in the writers (HTML, for example). I'll commit a change for the HTML writer and RFC here.

argent0:
I removed the `SimpleFigure` pattern from the HTML writter. Naturally this breaks many tests that expect the `SimpleFigure` construction to work.

When considering the `Figure` constructor I always had in mind the complementary `SimpleFigure`. So I'm not sure what's the desired behavior now. Maybe it's better that @tarleb continues working on this in order to shorten the iteration loop.

tarleb:
I've picked-up work on this again; currently working to improve support in all formats and to remove the `SimpleFigure` handlers from writers. Fixing some small bugs, too.

tarleb:
My current working branch is `figure-by-format`, which has one commit per format. I found this easier to work with, but will squash everything once I'm done.
See https://github.com/jgm/pandoc/compare/master...tarleb:pandoc:figure-by-format?expand=1

tarleb:
OK, I believe this is ready now: There is good reader support for HTML, LaTeX, and JATS thanks to the work by Aner. We also have solid writer support for the "important" output formats, including LaTeX, ConTeXt, HTML, Docx, DocBook, and JATS.

jgm:
This is a very big patch to review.  Could you update the commit message to give a full description of what this change does, clearly marking behavior changes and any API changes?  Also, I note that it depends on an unreleased version of pandoc-types.  Can you say what changes are needed in pandoc-types for his commit to work?

tarleb:
This necessary change to pandoc-types is <https://github.com/jgm/pandoc-types/pull/94>.

I'll add more detailed per-format descriptions to the PR and split it into (easily squashable) commits to make it easier to review.

tarleb:
I've updated the commit with a description of all changes.

Many formats don't support figures, so they get turned into Div elements. We should probably define a shared function for this, to make sure that figure captions are handled consistently in these cases.

jgm:
> Many formats don't support figures, so they get turned into Div elements. We should probably define a shared function for this, to make sure that figure captions are handled consistently in these cases.

That sounds like a good idea to me.

jgm:
> - FB2 writer: The figure caption is added as alt text to the images in
  the figure; pre-existing alt texts are kept.

> - HTML writer: the alt text is no longer constructed from the caption,
  as was the case with implicit figures. This reduces duplication, but
  comes at the risk of images that are missing alt texts. Authors should
  take care to provide alt texts for all images.

Is there a reason for this asymmetry?  Maybe FB2 writer should behave like the HTML writer and not add the caption as alt text?

jgm:
> - Markdown writer: figures are output as implicit figures if possible,
  and via HTML otherwise.

I didn't check the code here, but obviously we should only use HTML if `raw_html` is enabled; otherwise we could just emit the figure contents (perhaps in a div)?

jgm:
> - OpenDocument writer: A separate paragraph is generated for each block
  element in a figure, each with style `FigureWithCaption`. Behavior for
  single-image figures therefore remains unchanged.

When this kind of "unwrapping" is used, is the caption repeated for each single-image figure?  Or how does that work?

jgm:
I haven't had a chance to look over the code yet -- just the commit message. A few questions above.  This is really a ton of work, and should really improve things!

One problem is that now there is no good way to represent complex figures in Markdown. It is tempting to add a figure_divs extension that just puts them in Divs with class `figure`.  Maybe that isn't the ideal Markdown syntax for figures, but having something might be better than having nothing.

tarleb:
> Is there a reason for this asymmetry? Maybe FB2 writer should behave like the HTML writer and not add the caption as alt text?

I seem to remember that FB2 figures can only contain a single image, and this was an attempt to create images while preserving as much info as possible. I wanted to check again, but the page with the FB2 spec is down, and I can't find it anywhere else.

tarleb:
> I didn't check the code here, but obviously we should only use HTML if raw_html is enabled; otherwise we could just emit the figure contents (perhaps in a div)?

I think I forgot about that; fill fix it with the next push.

jgm:
> HTML writer: the alt text is no longer constructed from the caption,
  as was the case with implicit figures. This reduces duplication, but
  comes at the risk of images that are missing alt texts. Authors should
  take care to provide alt texts for all images.

This raised a question for me.  I could probably answer it by reading the code, but maybe you can just tell me.  How does `implicit_figures` work in the context of these changes?  Does it create a Figure element on the reader side?  Or does it get applied on the writer side, special treatment of Para [Image], so that figures created with `implicit_figures` are handled completely differently?

In fact, there doesn't seem to be any other way to create a figure presently, in markdown.  And `implicit_figures` doesn't give you a way to set the alt text independently of the caption, does it?  So does the above mean that if you go markdown -> HTML, your figures will suddenly lack alt text where they had it previously? That seems undesirable, especially if there is no alternative.

tarleb:
I was about to say that implicit figures could we written with an explicit `alt` attribute, as in `Caption{alt="description"}`, but that's not true: all image attributes are currently transferred to the Figure element, so the alt text would be lost. That's not good.

A possible solution could be to set the `alt` attribute to the stringified caption if the image wouldn't have a caption otherwise. Also, we should keep part of the attributes on the Image element instead of setting those attributes on the figure. Maybe only the ID should be transferred, so classes and key-value attributes (like `width`, `height`) would be left on the image.

jgm:
> A possible solution could be to set the alt attribute to the stringified caption if the image wouldn't have a captionan alt value otherwise.

That sounds sensible.

> Also, we should keep part of the attributes on the Image element instead of setting those attributes on the figure. Maybe only the ID should be transferred, so classes and key-value attributes (like width, height) would be left on the image.

Maybe we could use current behavior of the HTML renderer as a guide?  What gets transferred to the figure element currently?

jgm:
Looks like the HTML reader currently ignores attributes on the figure element but passes those on the img element through to the image.

We do need to figure out some way to avoid these changes breaking current workflows with `implicit_figure`.  That is my main hesitation, currently, about merging.

tarleb:
The Markdown reader now uses the image description as figure caption, while also adding it to the image. This would lead to duplication in HTML, so we check for this case in the HTML writer and add the `aria-hidden` attribute if caption and image description are equal. The old behavior will be preserved this way for most people.

test/fb2/images.fb2
```diff
@@ -1,2 +1,24 @@
...
...yUZeIpj6pRxOJjDGO5qNpbNLsWTiSyt/dex5mR1Ui+9Phxe72w+moI1mmViSbG7xQNNpHlrhKBnkkcbIwi5WZGegpgMYYq6oCoKoiamZAfB1TWCoq+o/2DxzEiU/TNHXeBCGEqqqKGKGIDPOFTedDY8x8M5HEIRlXi5D7TtrsVpv5cDbZyB69eL7a6IjIfD6NNBHJy1A3tcRJCOV4Nmq1OiqZUgFznonzZVWIECoqXjWpijJGevVipFkZqipYKt6JE1XXjj5tJHk+n41jptrtrifqptNpq9WaTMcqyXon9VlqxHgyi5GjWa6qeZ5DJcsyM6OomVm04WQwnY3XIGkCQGEhzktjORoN8yp3zqVpI0lSUNXB59OhiChE1RdaSK0XSTNqoSSbmQe8F8Bi4jRN00WRr3TY7jSzJAEAQZIkCBUQo1aqmtSTHmJkrGApQtu7lodKKXDQGCyaaJo4mgeAEEPM667sd59/dyoQpUYDUQBCpZiqqmrNIgCEIKlgWS1OTl4OYl9gRoo40VgrRhGpP0G9qMZgFieTKgaKOFXQgolaCE7MBOrgnEO0aKWfLI7rWRaAnC4ATv3ZpiLiVQFYoCmzLAuhHI5eMsInIBkNUXTZUAERAVlvFSrziZoZFxOaaK3HAJOw1K9ArQXKMl/MJ7Ld6Z0l/1m86qMCGEXEe7+EW0+g4iFljCHxLQGcWq0NUp/Ur52K+Vc3ohrrFrs8PzhlJgBiKt7MYoxVKIpiIRudtdc9QSxheTPnHIAYCaNzTlXNzKg+YWRQpgKQRjGIB8Pyb0BJnv1ikFoYOwBEBKAEqFQlaRalFtqkc440n3IZrNp7Z4LExMpgy4SAVBaFVp9zhBCIaBQnDqhPSQgKIAJPksal+AG8ogpRnQCIUVRFoSEE9QIaLNCpqiPFQUK0/wPxadi/ncvxsAAAACV0RVh0ZGF0ZTpjcmVhdGUAMjAxMS0wMi0yOFQwMjo1NTowMiswMTowMGbLlncAAAAldEVYdGRhdGU6bW9kaWZ5ADIwMTEtMDItMjhUMDI6NTU6MDIrMDE6MDAXli7LAAAAEXRFWHRqcGVnOmNvbG9yc3BhY2UAMix1VZ8AAAAgdEVYdGpwZWc6c2FtcGxpbmctZmFjdG9yADF4MSwxeDEsMXgx6ZX8cAAAAABJRU5ErkJggg==</binary></FictionBook>
```
jgm, line 2: What happened to the embedded binary image content which was in the original test output?

src/Text/Pandoc/Shared.hs
```diff
@@ -47,6 +47,7 @@ module Text.Pandoc.Shared (
                      compactify,
                      compactifyDL,
                      linesToPara,
+                     figureDiv,
```
jgm, line 50: This definitely needs to be announced as an API change in the commit message.
(Worth also checking the rest of the changes to make sure there are no other API changes.)

test/command/4677.md
```diff
@@ -2,8 +2,5 @@
 % pandoc --to "markdown-bracketed_spans-fenced_divs-link_attributes-simple_tables-multiline_tables-grid_tables-pipe_tables-fenced_code_attributes-markdown_in_html_blocks-table_captions-smart"
 ![Caption](img.png){#img:1}
 ^D
-<figure>
 <img src="img.png" id="img:1" alt="Caption" />
-<figcaption aria-hidden="true">Caption</figcaption>
-</figure>
```
jgm, line 8: Why isn't this an "implicit_figure"?

test/writer.docbook5
```diff
@@ -1320,7 +1320,9 @@ or here: &lt;http://example.com/&gt;
       From <quote>Voyage dans la Lune</quote> by Georges Melies (1902):
     </para>
     <figure>
-      <title>lalune</title>
+      <title>
+        lalune
+      </title>
```
jgm, line 1325: I think we generally try to use indentation for block-level content but keep it unindented for inline-level; was there a reason this changed?

test/writer.xwiki
```diff
@@ -623,8 +623,8 @@ or here: <http://example.com/>
 
 From “Voyage dans la Lune” by Georges Melies (1902):
 
-[[image:lalune.jpg||alt="lalune" title="fig:Voyage dans la Lune"]]
-
+(((
+[[image:lalune.jpg||alt="lalune" title="Voyage dans la Lune"]])))
```
jgm, line 627: It's odd that there's a line break after `(((` but not before `)))`, is there a reason for this?

test/command/4677.md
tarleb: This can't be an implicit figure as we'd loose due to `link_attributes` being disabled. But it should still be a figure. Fixed now.

test/fb2/images.fb2
tarleb: Weirdly enough, the test doesn't seem to check whether the image is embedded. I've added it back regardless.

test/writer.docbook5
tarleb: Oops, my bad. Fixed.

test/writer.xwiki
tarleb: I took a look at the [docs](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiSyntax/?syntax=2.1&section=Groups) and couldn't find a reason, so I've added a newline before the final `)))`.

src/Text/Pandoc/Shared.hs
tarleb: Added a bullet point to the commit message. I checked again and didn't find any other API changes.

test/command/6925.md
```diff
@@ -23,12 +23,12 @@ a
 ::: thm
 **Theorem 1**. *a*
 
-![image](1.png)
+![](1.png)
```
jgm, line 26: Here the alt text disappears -- won't that mean we no longer have a figure?  An `implicit_figure` needs a non-empty alt text.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -677,6 +669,31 @@ blockToMarkdown' opts (OrderedList (start,sty,delim) items) = do
...
+  case body of
+    [Plain [Image imgAttr alt (src, ttl)]]
+      | isEnabled Ext_implicit_figures opts
+      , Just descr    <- combinedAlt alt
+      , Just imgAttr' <- combinedAttr imgAttr
+      , isEnabled Ext_link_attributes opts || imgAttr' == nullAttr
+        -> do
+          -- use implicit figures if possible
+          let tgt' = (src, fromMaybe ttl $ T.stripPrefix "fig:" ttl)
+          contents <- inlineListToMarkdown opts [Image imgAttr' descr tgt']
```
jgm, line 690: I think we need a nonempty alt for implicit figures.  So maybe the `_` case for `combinedAlt` should be `Just "image"`?  That preserves earlier behavior.

test/command/6925.md
tarleb: Fixed.;

jgm:
Taking a deep breath...here we go!

tarleb:
I got to admit, I'm properly scared. Hoping for timely feedback from the Quarto dev team.

tarleb:
Congratulations @argent0!

PR #7705, opened 2021-11-21
Markdown, CommonMark: add support for wiki links. [API change]

tarleb:
Adds commonmark extensions `wikilinks_title_after_pipe` and
`wikilinks_title_before_pipe`. The former enables links of style `[[Name
of page|Title]]` and the latter `[[Title|Name of page]]`. Titles are
optional in both variants, so this works for both:
`[[https://example.org]]`, `[[Name of page]]`.

The writer is modified to render links with title `wikilink` as a
wikilink if a respective extension is enabled.

Pandoc will use `wikilinks_title_after_pipe` if both extensions are
enabled.

Closes: #2923

### DIFF
--- a/src/Text/Pandoc/Writers/Markdown/Inline.hs
+++ b/src/Text/Pandoc/Writers/Markdown/Inline.hs
@@ -651,6 +651,9 @@ inlineToMarkdown opts lnk@(Link attr@(ident,classes,kvs) txt (src, tit)) = do
                 case txt of
                       [Str s] | escapeURI s == srcSuffix -> True
                       _       -> False
+  let useWikilink = tit == "wikilink" &&
+                    (isEnabled Ext_wikilinks_title_after_pipe opts ||
+                     isEnabled Ext_wikilinks_title_before_pipe opts)
   let useRefLinks = writerReferenceLinks opts && not useAuto
   shortcutable <- asks envRefShortcutable
   let useShortcutRefLinks = shortcutable &&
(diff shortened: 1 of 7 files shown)


tarleb:
Still needs documentation, but I'd already welcome feedback on the general direction.

jgm:
Two thoughts:

- As noted in the linked issue, in GitHub wikis spaces are converted to - and other symbols are encoded ; I guess the idea would be not to do this conversion automatically, requiring a filter to be used?  (Otherwise we'd have to support N different URL munging schemes...)
- Would it be worth adding a parser for these syntaxes to the regular markdown reader, so the extensions work there as well as commonmark?

tarleb:
Given the diversity of wikis with divergent behavior, requiring a filter to match the exact results seems somewhat reasonable to me.

I don't really have a strong opinion about implementing it in regular Markdown. On one hand, it would be nice to nudge more users to use CommonMark. But people relying on citations may not be able to switch. Also, @blackheaven did half of the work already in #6458. I think supporting it there as well would probably be the right choice.

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -557,6 +558,11 @@ inlineToMarkdown opts lnk@(Link attr@(ident,classes,kvs) txt (src, tit)) = do
       | useAuto -> return $ literal srcSuffix
       | otherwise -> return linktext
     _ | useAuto -> return $ "<" <> literal srcSuffix <> ">"
+      | useAuto && useWikilink -> return $ "[[" <> literal srcSuffix <> "]]"
+      | useWikilink && isEnabled Ext_wikilinks_title_after_pipe opts -> return $
+        "[[" <> literal src <> "|" <> linktext <> "]]"
```
jgm, line 563: I think there might be a problem using `linktext` here.
In the commonmark extension, we parse the page title as raw text, so e.g. `*hi*` is not emphasized text.
It's not entirely clear how we should handle the case where we have a Link designated as a "wikilink" but the link text contains formatting.  One option would be to use `stringify`.  Or perhaps we could do what you're doing here, which would create raw markdown formatting.

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -557,6 +558,11 @@ inlineToMarkdown opts lnk@(Link attr@(ident,classes,kvs) txt (src, tit)) = do
       | useAuto -> return $ literal srcSuffix
       | otherwise -> return linktext
     _ | useAuto -> return $ "<" <> literal srcSuffix <> ">"
+      | useAuto && useWikilink -> return $ "[[" <> literal srcSuffix <> "]]"
```
jgm, line 561: is there a reason for using this rather than the more portable <...> syntax?

jgm:
@tarleb any thoughts on these comments?

tarleb:
I'm sorry, I didn't get around to it yet. It'll probably take another month or so until I can go back and complete this.

skywind3000:
any progress ?? really want this feature

src/Text/Pandoc/Writers/Markdown/Inline.hs
tarleb: I changed this to use stringify.

src/Text/Pandoc/Writers/Markdown/Inline.hs
tarleb: The motivation here was to preserve as much information as possible, esp the fact that this is a wikilink. That "wiki-autolink" syntax appears to enjoy wide support across implementations. But just falling back to explicit title/url wiki syntax could be saver, if a little less concise.

tarleb:
Should be good to go: the Markdown reader now uses basically the same algorithm as CommonMark, with two exceptions:  single `]` are allowed within the link, while newlines are not. This is modeled after GitHub wiki syntax. (Their implementation is maddening.)

PR #7706, opened 2021-11-22
Add .yml to Citeproc formatFromExtension

jkrenzer:
Make Citeproc recognize files with .yml extension as YAML-bibliography.

This is a shoot at solving #7707.


jgm:
Thanks!

Issue #7709, opened 2021-11-23
Add pandoc.text() to Lua API?


jgm:
Could we export pandoc.text, wrapping Text.Pandoc.Builder.text?
(A simple way to convert a string to a sequence of Str and Space elements.)
@tarleb

tarleb:
Sure! We'd probably need a different name though, as `pandoc.text` is already populated by the `text` module. Maybe `pandoc.utils.text` would do?

We could change the way strings are handled in situations where they are passed to functions that expect a list of inline values. I believe filter breakage would be minimal. The `pandoc.utils.text` function would then be equivalent to `local text = function(s) return pandoc.Emph(s).content end`.

Currently a plain string `s` is treated as `{pandoc.Str(s)}` when passed to a function that expects inlines. It should have been designed differently right from the start, but the old architecture had made that difficult.

jgm:
Actually, I love the idea of using the text constructor for plain strings; that would be by far the most convenient thing.
Is that impossible?

PR #7712, opened 2021-11-23
Lua: split strings into words when treating them as Inline list

tarleb:
Using a Lua string where a list of inlines is expected will cause the
string to be split into words, replacing spaces and tabs into
`pandoc.Space()` elements and newlines into `pandoc.SoftBreak()`.

The previous behavior was to treat the string `s` as `{pandoc.Str(s)}`.
The old behavior can be recovered by wrapping the string into a table
`{s}`.

### DIFF
--- a/src/Text/Pandoc/Lua/Marshaling/AST.hs
+++ b/src/Text/Pandoc/Lua/Marshaling/AST.hs
@@ -67,6 +67,7 @@ import Text.Pandoc.Lua.Marshaling.ListAttributes
   (peekListAttributes, pushListAttributes)
 
 import qualified HsLua as Lua
+import qualified Text.Pandoc.Builder as B
 import qualified Text.Pandoc.Lua.Util as LuaUtil
 
 instance Pushable Pandoc where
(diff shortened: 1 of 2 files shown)


tarleb:
As mentioned in #7709.

jgm:
This is great. Is this behavior documented?

tarleb:
I definitely need to start documenting this and the other behaviors. It will probably spur some necessary discussions, e.g. on handling metadata, captions, and tables.

tarleb:
BTW, I'd like to extract the AST marshaling code into a separate package. The repo could live in the pandoc organization.  Do you have any concerns that would speak against such a step? Or should we do with an internal library instead, thereby avoiding the overhead of yet another repository?

jgm:
I don't mind extracting this code into a separate library, if that's what you prefer.  But what is the reason for doing this?  Is it that these functions would be useful to others, not just in pandoc?

tarleb:
It's actually more trivial than that. For one, it forces me to draw clear API boundaries. Also, I'm planning to break the code into separate modules with the help of hs-boot files; I feel like that kind of build-complexity is better kept out of pandoc's code base.

So no strong reasons, it's more just personal preference.

jgm:
Well, it's fine with me!

tarleb:
Preliminary repo: https://github.com/tarleb/hslua-pandoc-types.

The corresponding PR is #7719.

Issue #7421, opened 2021-07-01
Rowcolors colors table captions and headings when multiple tables span pages


jamesgarriss:
**The Goal**

I am converting Markdown to PDF.  The content includes multiple, long tables that often break over the pages.  My goal is to alternate the coloring of the rows.

**The Problem**

The problem that I'm seeing is that the coloring appears on table captions and headings, not on the first table but on following tables.  I'm not sure if this is because I'm doing something poorly or because there's a bug in Pandoc.

To re-create the problem:

1. Create a Markdown file with multiple, long tables (see attached ipad.md).
2. Convert Markdown to PDF and observe that everything looks good.

```
pandoc ipad.md
  --output=ipad.pdf
```

3. Extract the template.

```
pandoc -D latex > new_template.latex
```

4. Modify the template in two places to use `rowcolors` (see attached new_template.latex, which I had to rename to .txt, as GitHub won't allow me to upload a .latex).
  a. Find `\usepackage{xcolor}` and change to `\usepackage[table]{xcolor}`.
  b. Find `\usepackage{longtable,booktabs,array}` and add the following line `\rowcolors{1}{}{yellow}`.
5. Convert Markdown to PDF using the template.

```
pandoc ipad.md
  --output=ipad.pdf
  --template=new_template.latex
```

6. Open the resulting PDF (see attached ipad.pdf).  Notice that caption and heading for tables 2 and 4 are highlighted yellow.

It seems like there should be a smarter/simpler way to solve this problem, but it was all I could figure out.  If there's not, then I assume this is a bug in Pandoc.

**Pandoc Version** 

pandoc 2.14.0.1
macOS 11.4 (Big Sur)

[ipad.md](https://github.com/jgm/pandoc/files/6747868/ipad.md)
[new_template.txt](https://github.com/jgm/pandoc/files/6747884/new_template.txt)
[ipad.pdf](https://github.com/jgm/pandoc/files/6747874/ipad.pdf)

mb21:
Well, if you're modifying your latex template and things aren't as expected, I wouldn't say that's a bug in pandoc. 

For questions like this, please use the [pandoc-discuss mailing list](https://groups.google.com/d/forum/pandoc-discuss) instead of this issue tracker. Thanks!

jgm:
[...]

PR #7716, opened 2021-11-25
Add header colors

bollenn:
Create option to adjust table header colors and/or row style.

This is the upgraded to todays version of https://github.com/jgm/pandoc/pull/3935

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -266,6 +266,9 @@ $highlighting-macros$
 $endif$
 $if(tables)$
 \usepackage{longtable,booktabs,array}
+\newcommand{\BeforeTableStyle}{}
+\newcommand{\TableHeaderRowStyle}{}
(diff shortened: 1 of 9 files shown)


jgm:
I'm not crazy about this, because I like pandoc to generate vanilla LaTeX that doesn't presuppose that custom macros have been defined in the preamble.

What sort of styling are you interested in achieving?

bollenn:
Tent to agree, not a big fan neither but it works :rofl: 

Using https://github.com/jgm/pandoc/issues/7421 I'm getting closer to our company style, still need some fixing in the header font style but...

bollenn:
Just noticed this came close to how sphinx does the trick.

See `sphinxattablestart`, `sphinxstyletheadfamily`  https://github.com/sphinx-doc/sphinx/blob/4442f32e7ac94a27ac44d8533279c7eb9179efef/doc/latex.rst#macros

PR #7720, opened 2021-11-28
Lua: remove `pandoc.utils.text`

tarleb:
The new `pandoc.Inlines` function behaves identical on string input, but
allows other Inlines-like arguments as well.

The `pandoc.utils.text` function could be written as

    function pandoc.utils.text (x)
      assert(type(x) == 'string')
      return pandoc.Inlines(x)
    end

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -3121,24 +3121,6 @@ Usage:
     -- outputs "Moin"
     print(pandoc.utils.stringify(inline))
 
-### text {#pandoc.utils.text}
-
-`text (words)`
-
-Converts a string to `Inlines`, treating interword spaces as
-`Space`s or `SoftBreak`s. If you want a single `Str` with literal
-spaces, use `pandoc.Str`.
-
-Parameters:
-
-`words`
-:  markup-less text (string)
-
-Returns:
-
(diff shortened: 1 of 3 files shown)


jgm:
Just wondering about discoverability.  What does the documentation for the Inlines function say?

tarleb:
This is what we have now:

> [`Inlines (inline_like_elements)`]{#pandoc.inlines}
>
> Converts its argument into an [Inlines](#type-inlines) list:
>
> -   copies a list of [Inline] elements into a fresh list; any
>     string `s` within the list is treated as `pandoc.Str(s)`;
> -   turns a single [Inline] into a singleton list;
> -   splits a string into `Str`-wrapped words, treating
>     interword spaces as `Space`s or `SoftBreak`s.
>
> Parameters:
>
> `inline_like_elements`:
> :   List in which each element can be treated as an [Inline]
>     values, or just a single such value.
>
> Returns: [Inlines] list

I don't feel strongly about keeping or removing `pandoc.utils.text`; please feel free to close this PR if the minor duplication in functionality is acceptable.

tarleb:
I just realized that I had created the branch in the main repo instead of my fork. Sorry 'bout that.

PR #7729, opened 2021-12-03
Adding a writer for Markua 0.10

binaarinen:
This change will add a simple writer to support outputs in Markua 0.10. The 
writer setup is similarly to the Markdown writer to conform to the pandoc 
structure. It is influenced by the Markdown, XWiki and TEI writers and can be 
easily extended to include more functionality in the future.

## Project
The project was done as part of an assignment under Prof. Dr. Farhad D.Mehta at
the Ostschweizer Fachhochschule - OST (university of applied science) by:
    - Tim Wisotzki (tim.wisotzki@ost.ch)
    - Samuel Lemmenmeier (samuel.lemmenmeier@ost.ch)
If possible we would be happy to get some feedback soon, since the project will
conclude with that.

## Functionality
 - convert the Pandoc AST to Markua syntax (as defined in the 
     [Markua Manual](https://leanpub.com/markua/read)) wherever possible 
     (some features are not supported by Markua) to the most general Markua 
     style.
 - common options for Markua (citeproc, bibliography, columns)
 - when the Markua Manual does not include a unique specification, the not
     final [Manual Spec](https://markua.com) was used as a guideline.

## Special considerations
the following assumtions and/or defaults were used for the conversion form AST 
to Markua:
 - attributes are, when available, always added as an attribute-list above the 
    block or around an inline span
 - if the data for footnotes and bibliography are available they are 
    automatically added. 
 - images are inserted as figues and if they have a caption, alt or title, these
    are only inserted in the attribute list.
 - links are always in span
 - divisions will be returned without <div> tags and attribute
 - since Markua doesn't support raw blocks, but they can be used for filters, 
    the are returned in original form, without attributes
 - math will always have the format latex


src/Text/Pandoc/Writers/Markua.hs
```diff
@@ -0,0 +1,501 @@
...
+                       <> cr <> headRef <> nestle contents <> blankline
+
+blockToMarkua opts (Table _ blkCapt specs thead tbody tfoot) = do
+    let (caption, aligns, widths, headers, rows) = toLegacyTable
+                                                   blkCapt
+                                                   specs
+                                                   thead 
+                                                   tbody 
+                                                   tfoot
+    let numcols = maximum (length aligns :
```
tarleb, line 218: Checking all these things is good, defensive programming. I believe that `toLegacyTable` guarantees all lengths to be equal, so the code below could be simplified.

The use of `maximum` is ok here, as the list can never be empty. However, we generally try to avoid partial functions. How about `foldr (max . length) 0 [...]`?

src/Text/Pandoc/Writers/Markua.hs
```diff
@@ -0,0 +1,501 @@
...
+    let eqs = literal $ T.replicate level "#"
+    return $ blankline <> attributes <> cr 
+            <> eqs <> space <> contents <> blankline
+
+blockToMarkua _ HorizontalRule = return $ blankline <> "* * *" <> blankline
+
+-- div is also used to compute the bibliography
+-- if --citeproc and --bibliography is set (div with refs id is generated),
+-- then a bibliography header level 1 with a attr list above is inserted
+blockToMarkua opts (Div attr blks) = do
```
tarleb, line 190: Pattern matching might result in cleaner code. How about something like `blockToMarkua opts (Div ("refs",_,_) blks)` to handle the bibliography?

src/Text/Pandoc/Writers/Markua.hs
```diff
@@ -0,0 +1,501 @@
...
+    let citeId = if attrIdPrefix == "ref-"
+                 then T.drop 4 attrId
+                 else T.empty
+    let headRef = literal (if not $ T.null citeId
+                           then T.cons '[' (T.append citeId "] ")
+                           else T.empty)
+    -- attr only used for bibliography
+    let attributes = attrToMarkua (citeId,[],[])
+    let biblioHeader = literal (if attrId == "refs"
+                                then "{id: bibliography}\n#Bibliography"
```
tarleb, line 203: There should probably be a space after `#`, matching other headers. Could we use `blockToMarkua opts (Header ("bibliography",[],[]) "Bibliography"` to generate that part?

tarleb (commented):
Thank you, this is such a great project to do!

I've added a few first comments. Before I continue, I'd like to ask whether you'd prefer
1. a review focused on functionality, or
2. detailed review, also looking at code style and adding my (very subjective) ideas for possible improvements?

It would be helpful for us to know your time constraints, i.e., when the project has to be finished.

It's not a requirement, but do you think you may have time to maintain the writer in the future? It would be excellent to have someone to help if bugs are found by users at a later point.

jgm:
A basic question:  is markua so different from markdown that it makes sense to implement this in a separate module rather than as a variant of markdown?  (That is how we currently handle commonmark and plain output.)

binaarinen:
Thank you for your fast replies and the initial feedback.
We will start working on it right away.

***feedback***
We will have to hand in the project on the 21st of December (we know it's rather short notice). Clearly it would be great if we could get the request well enough to be merged by then and we are happy to make adjustments to improve the writer. Since we don't have unlimited time left, we would prioritise some functional feedback, but of course we would also be greatful about any other observations.


***include in markdown***
It might have been possible to include it as a variant of markdown, but because of the following differences we decided to implement it as a separate writer:
 - Markua handles linebreaks different then markdown, which affects basicly all functions.
 - Attribute lists are denoted and placed differently
 - Markua has some significant differences in it's featureset

A list of some of the main differences can also be found in the [Manual](https://leanpub.com/markua/read#markdown-diffs)

jgm:
> Markua handles linebreaks different then markdown, which affects basicly all functions.

We already have a switch in the markdown writer for `hard_line_breaks`.  As far as I can see, this gives the same behavior described in the Markua manual.

> Attribute lists are denoted and placed differently

This is true, but it could be handled by a new extension or switch in the markdown writer.

> Markua has some significant differences in it's featureset

Almost all of those differences are already supported by pandoc markdown extensions, e.g. list numbering features, gfm tables, definition lists, fenced code blocks.

Of those differences that aren't supported, many are irrelevant (e.g. `+` not working for bullet lists -- the markdown writer doesn't use `+` anyway for this), or the stricter standards for ATX headers, or the lack of indented code blocks (which the writer doesn't use, provided `fenced_code_blocks` is enabled).

The whitespace differences seem to be handled already by current extensions.

In Markua all URLs must be absolute; that wouldn't be enforced by the current writer, but I'm not sure how one would generate an absolute URL if the reader gives you a relative one.

One difference that does seem significance is mapping `_` to underline instead of emphasis.

The format for math is a bit different.

The blurb syntax seems new, but it's isomorphic with pandoc fenced divs and could be supported with a switch, couldn't it?  Does your writer produce this syntax, and if so from what elements in the pandoc AST?

The directives are new, but I'm not sure what in the pandoc AST should map to these (if anything).

Before merging this module, I'd need to be convinced that it makes sense to have a separate module, rather than implementing some markdown variants in the existing markdown writer (as we already do for multimarkdown, gfm, commonmark, php markdown extra, and plain).  The advantage of the latter approach is that lots of code can be reused instead of being duplicated (e.g., code for lists and pipe tables).  This can be important for maintainability in the future.

binaarinen:
Thank you for your feedback!
Of course we want to provide a consistent solution with the rest of the project. We are currently taking a closer look at integrating the markua part with the markdown writer and if its possible to cover all functions necessary for markua.

jgm:
I think that, as long as there is no Markua reader, it probably doesn't make sense to add any new markdown extensions (T.P.Extensions). The Markua variant can be handled more simply by adding a new Variant constructor in T.P.Writers.Markdown.  (See how `plain` is treated.)  Then it's just a matter of checking the variant in clauses for lists, tables, etc., and making any necessary adjustments.

binaarinen:
We integrated the markua writer within the markdown writer as you suggested and updated the pull request accordingly.
Please let us know if further improvements are necessary and would be happy if we could contribute to the project.

jgm:
Thank you -- this looks good on first glance.  I'll take a deeper look before long1
And we'll run the CI.

jgm:
Can you rebase this into one commit on top of current master?

MANUAL.txt
```diff
@@ -252,6 +252,7 @@ header when requesting a document from a URL:
     - `markdown_mmd` ([MultiMarkdown])
     - `markdown_phpextra` ([PHP Markdown Extra])
     - `markdown_strict` (original unextended [Markdown])
+    - `markua` ([Markua])
```
jgm, line 255: This adds a Markua writer but not a reader, so we shouldn't list `markua` in the `--from` formats.

pandoc.cabal
```diff
@@ -19,7 +19,7 @@ description:     Pandoc is a Haskell library for converting from one markup
                  .
                  - light markup formats (many variants of Markdown,
                    reStructuredText, AsciiDoc, Org-mode, Muse, Textile,
-                   txt2tags)
+                   txt2tags, Markua)
```
jgm, line 22: I'd consider Markua a "variant of Markdown," so not needing separate mention here.

src/Text/Pandoc/Extensions.hs
```diff
@@ -601,6 +603,8 @@ getAllExtensions f = universalExtensions <> getAll f
   getAll "mediawiki"       = autoIdExtensions <>
     extensionsFromList
     [ Ext_smart ]
+  getAll "markua"          = extensionsFromList
+    [ Ext_citations]
```
jgm, line 607: Is this a meaningful extension for Markua?  It doesn't have automatic citation support, does it?

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -77,6 +81,25 @@ writeCommonMark opts document =
...
+                  enableExtension Ext_pipe_tables $
+                  enableExtension Ext_fancy_lists $
+                  enableExtension Ext_startnum $
+                  enableExtension Ext_strikeout $
+                  enableExtension Ext_subscript $
+                  enableExtension Ext_superscript $
+                  enableExtension Ext_definition_lists $
+                  enableExtension Ext_smart $
+                  enableExtension Ext_footnotes $
+                    writerExtensions opts }
```
jgm, line 100: Question: does it make sense to have *any* variation in extensions for Markua, given that the whole format is specified?  E.g. would it ever make sense to add the `grid_tables` extension?  If not, then maybe best simply to hard code a list of extensions here.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -328,7 +351,8 @@ blockToMarkdown' opts (Div attrs ils) = do
   variant <- asks envVariant
   return $
     case () of
-         _ | isEnabled Ext_fenced_divs opts &&
+         _ | variant == Markua -> contents <> blankline
```
jgm, line 354: It would be possible to use Divs with special classes to produce Markua blurbs and asides.  Would that make sense?

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -408,6 +432,9 @@ blockToMarkdown' opts b@(RawBlock f str) = do
       | f `elem` ["markdown", "markdown_github", "markdown_phpextra",
                   "markdown_mmd", "markdown_strict"]
          -> return $ literal str <> literal "\n"
+    Markua -> do
+        report $ NoTranslation "raw content is writen out in original form and"
```
jgm, line 436: I don't understand this...  NoTranslation has a specific use, to indicate that there is no translation for a given term.
I'm not sure what you're trying to warn about here, but another LogMessage constructor should be used.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -581,27 +615,33 @@ blockToMarkdown' opts t@(Table _ blkCapt specs thead tbody tfoot) = do
   return $ nst (tbl $$ caption'') $$ blankline
 blockToMarkdown' opts (BulletList items) = do
   contents <- inList $ mapM (bulletListItemToMarkdown opts) items
-  return $ (if isTightList items then vcat else vsep) contents <> blankline
+  variant <- asks envVariant
+  return $ (if isTightList items || variant == Markua then vcat else vsep) 
+                  contents <> blankline
```
jgm, line 620: Can you explain this?  Is it that list items *can't* have blank lines between them in Markua?

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -581,27 +615,33 @@ blockToMarkdown' opts t@(Table _ blkCapt specs thead tbody tfoot) = do
...
-                           (delim == OneParen || delim == TwoParens)
-                       then OneParen -- commonmark only supports one paren
-                       else DefaultDelim
+  let delim' | isEnabled Ext_fancy_lists opts =
+               case variant of
+                   Markua -> if delim == TwoParens then OneParen else delim
+                   _ -> delim
+             | variant == Commonmark && --commonmark only supports one paren
+                   (delim == OneParen || delim == TwoParens) = OneParen
+             | otherwise = DefaultDelim
```
jgm, line 633: I wonder if it makes more sense to put the check for Markua outside of this guard (like the check for Commonmark below).  Then we don't need to enable `fancy_lists` extension.  This is connected to the question above, whether it makes sense to have extensions for Markua.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -581,27 +615,33 @@ blockToMarkdown' opts t@(Table _ blkCapt specs thead tbody tfoot) = do
...
   let attribs = (start', sty', delim')
   let markers  = orderedListMarkers attribs
-  let markers' = map (\m -> if T.length m < 3
-                               then m <> T.replicate (3 - T.length m) " "
-                               else m) markers
+  let markers' = case variant of
+                        Markua -> map (\m -> m <> T.replicate 1 " ") markers
```
jgm, line 637: `T.replicate 1 " "` == `" "`

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -581,27 +615,33 @@ blockToMarkdown' opts t@(Table _ blkCapt specs thead tbody tfoot) = do
...
+                        _ -> map (\m -> if T.length m < 3
+                                   then m <> T.replicate (3 - T.length m) " "
+                                   else m) markers
   contents <- inList $
               zipWithM (orderedListItemToMarkdown opts) markers' items
-  return $ (if isTightList items then vcat else vsep) contents <> blankline
+  return $ (if isTightList items || variant == Markua then vcat else vsep)
```
jgm, line 643: Same question as above on bullet lists.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -725,19 +768,22 @@ bulletListItemToMarkdown opts bs = do
...
+  variant <- asks envVariant
   let sps = case writerTabStop opts - T.length marker of
                    n | n > 0 -> literal $ T.replicate n " "
                    _ -> literal " "
   let ind = if isEnabled Ext_four_space_rule opts
                then writerTabStop opts
                else max (writerTabStop opts) (T.length marker + 1)
-  let start = literal marker <> sps
+  let start = case variant of
+              Markua -> literal marker
```
jgm, line 785: Don't you want at least one space after the list marker?

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -95,6 +96,13 @@ escapeText opts = T.pack . go' . T.unpack
                   , isAlphaNum x -> c : '_' : x : go xs
                 _                -> c : go cs
 
+-- Escape the escape character, as well as formatting pairs
+escapeMarkuaString :: Text -> Text
+escapeMarkuaString s = foldr (uncurry T.replace) s $ zip ["--", "**", "//",
+                                                          "^^", ",,"]
+                                                          ["~-~-", "~*~*",
+                                                           "~/~/", "~^~^",
+                                                           "~,~,", "~~"]
```
jgm, line 105: I don't think it's a good idea to use `zip` in this way -- better just to use an explicit list, where each item is paired with its escaped counterpart. (Think what would happen if one item got deleted from just one of these lists, everything would pair up with the wrong counterpart.)

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -116,9 +124,62 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
               escAttrChar '\\' = literal "\\\\"
               escAttrChar c    = literal $ T.singleton c
 
+attrsToMarkua:: Attr -> Doc Text
+attrsToMarkua attributes
+    | null list = empty
+    | length list == 1 = braces $ head list
+    |  otherwise = braces $ hcat (head list : (map (\x -> ", " <> x)
```
jgm, line 131: In general, it's better to avoid using `head` (since it's a partial function and can throw runtime errors).
Why not use a case?  Match on
``` hs
  [] -> empty
  [x] -> braces x
  xs -> braces $ intercalate ", " xs
```

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -116,9 +124,62 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
...
+                                                   (tail list)))
+        where attrId = case attributes of
+                        ("",_,_) -> []
+                        (i,_,_)  -> [literal $ T.append "id: " i]
```
jgm, line 135: More idiomatic would be `"id: " <> i`

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -116,9 +124,62 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
...
+                               (_,_,[]) -> []
+                               (_,_,keyvalues) -> map (\(k,v) -> escAttr k
+                                              <> ": " <>
+                                              escAttr v)
+                                              (filter (/=(T.empty,T.empty))
+                                              (map preprocessKeyValues
+                                                   keyvalues))
+              escAttr          = mconcat . map escAttrChar . T.unpack
+              escAttrChar '"'  = literal "\""
+              escAttrChar c    = literal $ T.singleton c
```
jgm, line 152: Why not `escAttr = T.replace "\"" "\\\""`?

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -116,9 +124,62 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
...
+              attrClasses = case attributes of
+                             (_,[],_) -> []
+                             (_,classes,_) -> map (escAttr . ("class: " <>))
+                                classes
+              attrKeyValues = case attributes of
+                               (_,_,[]) -> []
+                               (_,_,keyvalues) -> map (\(k,v) -> escAttr k
+                                              <> ": " <>
+                                              escAttr v)
+                                              (filter (/=(T.empty,T.empty))
```
jgm, line 147: I'd probably use `(mempty, mempty)` instead of `(T.empty, T.empty)` -- for Text `mempty` is `T.empty`.

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -116,9 +124,62 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
...
+              list = concat [attrId, attrClasses, attrKeyValues]
+
+              -- if attribute key is alt, caption, title then content
+              -- gets wrapped inside quotes
+              -- if some value is automatically set to "fig:", then this
+              -- attribute gets removed
+              preprocessKeyValues :: (Text, Text) -> (Text, Text)
+              preprocessKeyValues (key,value)
+                 | value == "fig:" = (T.empty,T.empty)
+                 | T.head value == '\"' = (key, value)
```
jgm, line 163: Instead of using `(T.empty, T.empty)` for the values you're going to ignore, why not return `Maybe (Text, Text)` -- so this would be Nothing.  Then you can use `mapMaybe` on ln. 147 above.

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -116,9 +124,62 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
...
+                 | key == "alt" ||
+                   key == "caption" = (key, T.cons '\"' (T.snoc value '\"'))
```
jgm, line 165: Even though this might be a tiny bit more efficient, for code clarity I think you can't beat `"\"" <> value <> "\""`.
Or even better define your own `inquotes`.

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -283,6 +344,7 @@ inlineToMarkdown opts (Span attrs ils) = do
              _ -> id
          $ case variant of
                 PlainText -> contents
+                Markua -> "[" <> contents <> "]" <> attrsToMarkua attrs
```
jgm, line 347: Is this right?  Markua documentation says to use backticks for spans?
(I'm no expert, just glancing at it.)

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -396,60 +458,79 @@ inlineToMarkdown opts (Quoted DoubleQuote lst) = do
...
-        | isEnabled Ext_tex_math_double_backslash opts ->
-            return $ "\\\\[" <> literal str <> "\\\\]"
-        | otherwise -> (\x -> cr <> x <> cr) `fmap`
-            (texMathToInlines DisplayMath str >>= inlineListToMarkdown opts)
+inlineToMarkdown opts (Math InlineMath str) = do
+  variant <- asks envVariant
+  if variant == Markua
+      then do
+        let attr = addKeyValueToAttr ("",[],[]) ("format", "latex")
+        inlineToMarkdown opts (Code attr str)
```
jgm, line 496: From the docs it looks like the preferred shorter form is `` `e=mc^2`$ ``.

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -469,6 +550,9 @@ inlineToMarkdown opts il@(RawInline f str) = do
       | f `elem` ["markdown", "markdown_github", "markdown_phpextra",
                   "markdown_mmd", "markdown_strict"]
          -> return $ literal str
+    Markua -> do
+        report $ NoTranslation "raw content is writen out in original form and"
```
jgm, line 554: See above

jgm: Why not use `renderEmpty` here? It emits the right kind of warning.

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -557,6 +641,8 @@ inlineToMarkdown opts lnk@(Link attr@(ident,classes,kvs) txt (src, tit)) = do
     PlainText
       | useAuto -> return $ literal srcSuffix
       | otherwise -> return linktext
+    Markua -> return $ "[" <> linktext <> "](" <> (literal src) <> ")" <> 
+        attrsToMarkua (addKeyValueToAttr attr ("title", tit))
```
jgm, line 645: Do you want to do this even when `tit` is empty?

test/tables.markua
```diff
@@ -0,0 +1,58 @@
...
+|     |     |     |     |
+|----:|:----|:---:|----:|
+|  12 | 12  | 12  |  12 |
+| 123 | 123 | 123 | 123 |
+|   1 | 1   |  1  |   1 |
+
+Multiline table without column headers:
+
+|        |     |      |                                                       |
+|:------:|:----|-----:|-------------------------------------------------------|
```
jgm, line 56: Does markua support this way of indicating alignment in pipe tables with colons?  Not every implementation does.

test/writer.markua
```diff
@@ -0,0 +1,718 @@
...
+    If you want, you can indent every line, but you can also be lazy and just
+    indent the first line of each block.
+
+[^3]: This is *easier* to type. Inline notes may contain
+    [links](http://google.com) and `]` verbatim characters, as well as
+    [bracketed text].
+
+[^4]: In quote.
+
+[^5]: In list.
```
jgm, line 718: Have you run this whole test output through Markua and checked to make sure everything looks right when rendered?
(Ditto with the table test?)

MANUAL.txt
binaarinen: We changed it accordingly

pandoc.cabal
binaarinen: We changed it accordingly

src/Text/Pandoc/Extensions.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: We looked into the idea and it's now possible to create blurbs and asides by providing a div with class "blurb" or "aside".

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: We agree, that it doesn't make much sense to change certain options. However we enabled some, that will create the correct version also used by markdown (ie. pipetables) in order not to write duplicate code.

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: We used NoTranslation, because it seemed to be the most fitting, since markua doesn't support raw element. It is now changed to the "renderEmpty" you suggested, which is more fitting.

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: The fancy_lists extension is used, because Markua supports the use of "special" enumerators (like roman numbers). Even though in this case it would be more elegant to test for Markua in a different way, this would require to add a special case in every time fancy-lists are possible. To keep the code clean and without unnecessary duplications, we opted for using the option instead as with it the Markdown part is identical to Markua.

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: Yes, the space was already passed along with the marker. We changed this slightly to improve the readability/understanding.

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: Yes, you are correct, that must have slipped through our checks and is now corrected

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We changed it accordingly

test/tables.markua
binaarinen: Yes Markua supports this [feature](https://leanpub.com/markua/read#tables)

test/writer.markua
binaarinen: Unfortunately there is currently no automatic test-tool for Markua, but we manually checked the files again, to make sure that everything is proper Markua.

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We are not entirely sure where this applies, but through the process of applying your feedback we changed the entire function slightly.

binaarinen:
Thank you for your detailed feedback.
I'm sorry it took us a while to get to it, but we incorporated the changes to the code and rebased it into one commit.

src/Text/Pandoc/Writers/Markdown.hs
jgm: I understand; I wasn't suggesting that you not enable the extensions here.
The question is whether you should be adding them to `writerExtensions opts`, which implies that users can set extensions not on this list.  (In other words, could `writerExtensions opts` be replaced by `mempty`?  If there are no writer options it makes sense for the *user* to enable, then I think it could.)

src/Text/Pandoc/Writers/Markdown.hs
jgm: Have tests for this feature been added?

src/Text/Pandoc/Writers/Markdown/Inline.hs
jgm: My point was that `escAttr = T.replace "\"" "\\\""` seems to do exactly what the three line definition above does.  Or am I wrong?

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -327,8 +351,15 @@ blockToMarkdown' opts (Div attrs ils) = do
   contents <- blockListToMarkdown opts ils
   variant <- asks envVariant
   return $
-    case () of
-         _ | isEnabled Ext_fenced_divs opts &&
+     case () of
+         _ | variant == Markua -> 
+                   case () of
+                        () | classes' == ["blurb"] -> prefixed "B> " contents <> blankline
+                           | classes' == ["aside"] -> prefixed "A> " contents <> blankline
```
jgm, line 358: Blurbs can have other classes attached (e.g. in the documentation, `warning`).
So we shouldn't assume that there will be just one class, `blurb`. Instead, test to see that `"blurb"` is an element of `classes'`, and ditto for aside.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -327,8 +351,15 @@ blockToMarkdown' opts (Div attrs ils) = do
...
+                           -- | necessary to enable option to create a bibliography
+                           | (take 3 (T.unpack id')) == "ref" -> contents <> blankline
+                           | otherwise -> mempty
```
jgm, line 361: I don't think we should be completely omitting the contents of Divs that don't have these special classes!
Instead, just fall back to emitting the contents.

src/Text/Pandoc/Writers/Markdown.hs
```diff
@@ -581,27 +620,33 @@ blockToMarkdown' opts t@(Table _ blkCapt specs thead tbody tfoot) = do
...
+  let markers' = case variant of
+                        Markua -> markers
+                        _ -> map (\m -> if T.length m < 3
+                                   then m <> T.replicate (3 - T.length m) " "
+                                   else m) markers
   contents <- inList $
               zipWithM (orderedListItemToMarkdown opts) markers' items
-  return $ (if isTightList items then vcat else vsep) contents <> blankline
+  return $ (if isTightList items then vcat else vsep) 
+                contents <> blankline
```
jgm, line 649: Let's restore the original (lack of) indentation if it fits in 80 columns.

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -116,9 +122,56 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
...
+              list = concat [attrId, attrClasses, attrKeyValues]
+
+              -- if attribute key is alt, caption, title then content
+              -- gets wrapped inside quotes
+              -- attribute gets removed
+              preprocessKeyValues :: (Text, Text) -> (Text, Text)
+              preprocessKeyValues (key,value)
+                 | key == "alt" ||
+                   key == "caption" ||
+                   key == "title" = (key, inquotes value)
```
jgm, line 156: I don't understand why these attributes are being treated differently in this way??

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -116,9 +122,56 @@ attrsToMarkdown attribs = braces $ hsep [attribId, attribClasses, attribKeys]
...
+              intercalateDocText [x] = x
+              intercalateDocText (x:xs) = x <> ", " <> (intercalateDocText xs)
+
+-- | Add a (key, value) pair to Pandoc attr type
+addKeyValueToAttr :: Attr -> (Text,Text) -> Attr
+addKeyValueToAttr (attrlistId,classes,keyvalues) (key,value) 
+    | not (T.null key) && not (T.null value) = (attrlistId,
+                                                classes,
+                                                (key,value): keyvalues)
+    | otherwise = (attrlistId,classes,keyvalues)
```
jgm, line 169: For consistency, maybe use `ident` instead of `attrlistId`?

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -396,60 +450,77 @@ inlineToMarkdown opts (Quoted DoubleQuote lst) = do
...
+          _ | isEnabled Ext_tex_math_dollars opts ->
+                return $ "$" <> literal str <> "$"
+            | isEnabled Ext_tex_math_single_backslash opts ->
+                return $ "\\(" <> literal str <> "\\)"
+            | isEnabled Ext_tex_math_double_backslash opts ->
+                return $ "\\\\(" <> literal str <> "\\\\)"
+            | otherwise -> 
+                texMathToInlines InlineMath str >>=
+                  inlineListToMarkdown opts .
+                    (if variant == PlainText then makeMathPlainer else id)
```
jgm, line 500: This is OK, but it might be nicer to avoid this nesting and just have a case statement, with the first case being:
```hs
  _ | variant == Markua -> return $ "`" etc.
```

src/Text/Pandoc/Writers/Markdown/Inline.hs
```diff
@@ -396,60 +450,77 @@ inlineToMarkdown opts (Quoted DoubleQuote lst) = do
...
+                        (url <> urlEncode str, str))
+          _ | isEnabled Ext_tex_math_dollars opts ->
+                return $ "$$" <> literal str <> "$$"
+            | isEnabled Ext_tex_math_single_backslash opts ->
+                return $ "\\[" <> literal str <> "\\]"
+            | isEnabled Ext_tex_math_double_backslash opts ->
+                return $ "\\\\[" <> literal str <> "\\\\]"
+            | otherwise -> (\x -> cr <> x <> cr) `fmap`
+                (texMathToInlines DisplayMath str >>= inlineListToMarkdown opts)
+
```
jgm, line 523: see above on InlineMath

jgm (changes requested):
This is looking good.  I've made a few more comments.

Also, some small comments on your commit message.
Best if the top line says "Add a writer for Markua 1.0".
In the message, give a link to Markua's site.
Also, note the API changes using [API change]:  basically, it's just the newly exported function writeMarkua from T.P.Writers.Markdown.
You don't really need the rest, but keep the PR # and, if there's an issue in our tracker requesting Markua support, indicate that as well.

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: we changed it

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: we now added testcases

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: We did not manage to simplify the function like that and still get the required results

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: we changed it to handle other classes aswell

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: we change it so write the contents

src/Text/Pandoc/Writers/Markdown.hs
binaarinen: its back to one line

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: we changed it to fit to the others

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: we changed it

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: we changed it

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: The value of these keys should be in quotes, but others (ie. the format in Math) is written without, that's why they are separated

src/Text/Pandoc/Writers/Markdown/Inline.hs
binaarinen: Thank You, we appreciate your fast feedback and we adjusted the code accordingly.

binaarinen:
Thank you. We appreciate your fast feedback and adjusted the code accordingly.

src/Text/Pandoc/Lua/Marshal/Reference.hs
```diff
@@ -0,0 +1,101 @@
+{-# LANGUAGE LambdaCase           #-}
```
jgm, line 1: There's a bunch of stuff in this commit, like this for example, that doesn't relate to the change.
Can you rebase on current master HEAD?

binaarinen: I'm sorry, we forgot to do that, it's now rebased

jgm:
Many thanks for your contribution to pandoc!

PR #7732, opened 2021-12-06
Lua: update to latest pandoc-lua-marshal (0.1.1)

tarleb:
- `walk` methods are added to `Block` and `Inline` values; the methods
  are similar to `pandoc.utils.walk_block` and
  `pandoc.utils.walk_inline`, but apply to filter also to the element
  itself, and therefore return a list of element instead of a single
  element.

- Functions of name `Doc` are no longer accepted as alternatives for
  `Pandoc` filter functions. This functionality was undocumented.


doc/lua-filters.md
```diff
@@ -834,6 +834,38 @@ or `pandoc.Blocks`.
...
+`walk(lua_filter)`
+
+Applies a Lua filter to the element. Just as for full-document
+filters, the order in which elements are handled are Inline ->
+Inlines -> Block -> Blocks.
+
+Note that the result of this function is not a single Block, but
+a list of Block values ([Blocks]), as filters can delete
+elements or replace them with multiple Block values.
+
```
jgm, line 850: Should be `[/Block]` not `[Blocks]`?
And why talk of blocks, when this can be used on inlines too?

I guess I hadn't thought much about the return type before, though you brought it up.
It's a disanalogy with pandoc's Haskell `walk` that it returns a different type.  Maybe that's okay.
I suppose an alternative might be to have it return multiple values (basically, `table.unpack` the list).  Then someone who knows that the top element isn't going to be deleted or multiplied can just take the first return value, and others can `table.pack` the result.

jgm: Further question:  can you use `:walk` on a list (Haskell type `[Inline]` or `[Block]`)?

jgm: Lua manual says:

> A constructor also collects all results from a call, without any adjustments:

>    a = {foo0()}         -- a = {}  (an empty table)
>    a = {foo1()}         -- a = {'a'}
>    a = {foo2()}         -- a = {'a', 'b'}

So if we returned multiple values, it could be forced to a list by just adding braces.
Of course, you'd be in trouble if you didn't use the braces because you assumed it would return exactly one value, and this assumption were wrong.

tarleb:
Release pending; other than that it should be good to merge.

jgm:
Excellent.  The code is much cleaner now in the examples using walk.

PR #7737, opened 2021-12-07
workaround for possible pdfroff bug

jarnosz:
The `--no-toc-relocation` option avoids a double pass through the intermediate pdfroff files, fixing an empty page at the beginning, and pagination and faulty toc construction. The final output is identical to regular `groff -ms` output.

The `-Dutf8` simplifies the call to `groff` from within `pdfroff`.

### DIFF
--- a/haskell.yml
+++ b/haskell.yml
@@ -1,0 +1,40 @@
+name: Haskell CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: windows-latest
+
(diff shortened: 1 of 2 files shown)


jgm (commented):
I'm not sure what this commit is for, or why it's part of this PR.
Can you remove it?

jgm (commented):
This commit looks good, except that in the commit description lines should be hard-wrapped to under 80 columns per line.

jarnosz:
Ouch! I messed up: I committed to the master branch instead of labeling with the issue number. I'll revert or delete right away...

PR #7741, opened 2021-12-08
Custom reader: pass list of sources instead of concatenated text

tarleb:
The first argument passed to Lua `Reader` functions is no longer a plain
string but a richer data structure. The structure can easily be
converted to a string by applying `tostring`, but is also a list with
elements that contain each the *text* and *name* of each input source as
a property of the respective name.

A small example is added to the custom reader documentation, showcasing
its use in a reader that creates a syntax-highlighted code block for
each source code file passed as input.

Existing readers must be updated.

### DIFF
--- a/doc/custom-readers.md
+++ b/doc/custom-readers.md
@@ -17,7 +17,7 @@ install any additional software to do this.
 A custom reader is a Lua file that defines a function
 called `Reader`, which takes two arguments:
 
-- a string, the raw input to be parsed
+- the raw input to be parsed, as a list of sources
 - optionally, a table of reader options, e.g.
   `{ columns = 62, standalone = true }`.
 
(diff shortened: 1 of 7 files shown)


jgm:
What error is raised if you pass in a string instead of a sources structure?
Would it be reliable to catch this and pass in a string, to achieve backwards-compatibility? Or is this too much of a mess, given that we just rolled out custom readers and breakage might not be too painful at this stage?

tarleb:
Errors will typically contain `(string expected, got pandoc Sources)`. Checking for that and retrying would be a bit ugly, but not terrible either.

I'd still vote for a clean break though.

tarleb:
I added the check and it seems to work reliably. A deprecation warning is reported before conversion is tried a second time.

doc/custom-readers.md
```diff
@@ -27,14 +27,19 @@ which is automatically in scope.  (Indeed, all of the utility
 functions that are available for [Lua filters] are available
 in custom readers, too.)
 
+Each input source contains the *text*, and the *name* of an input
+file as a property of the respective name. The input sources
+list, as well as each of its element, can be converted to a
```
jgm, line 32: its elements

doc/custom-readers.md
```diff
@@ -27,14 +27,19 @@ which is automatically in scope.  (Indeed, all of the utility
 functions that are available for [Lua filters] are available
 in custom readers, too.)
 
+Each input source contains the *text*, and the *name* of an input
+file as a property of the respective name. The input sources
```
jgm, line 31: I find this a bit hard to understand.  "the *name* of an input file as a property of the respective name".  An example could help?

tarleb:
I rewrote the new paragraph a little, but I'm still not quite satisfied. I definitely wouldn't mind to see it replaced.

jgm:
This is better.  Was I hallucinating, or did you have code before that caught the error that would be raised if they assumed that the argument to Reader was a string?

tarleb:
The code is in a second commit in this branch so it can either be removed or squashed. Should I add a sentence about it to the docs?

jgm:
Oh -- I missed it.
I think we don't need to document it; it's there to support backwards-compatibility.

PR #7742, opened 2021-12-08
Lua reader: pass sources as third argument to Reader function

tarleb:


tarleb:
Closing in favor of #7741.

PR #7745, opened 2021-12-10
fix(IpynbOutput)!: rank always favors output format

ickc:
Previously, with fmt == f case and Image have a rank of 1.
In the end, e.g. from ipynb to html conversion, if both html and image exists, it actually prefers the image.
This commit changes this, so that fmt == f is always highest rank, and rank never collides.
This is achieved by keeping fmt == f case having rank 1, and every other rank increased by 1.

# Example

[just-plotly.ipynb.zip](https://github.com/jgm/pandoc/files/7691334/just-plotly.ipynb.zip)

```bash
pandoc -s -o just-plotly-pandoc.html just-plotly.ipynb -V header-includes='<script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js" integrity="sha512-c3Nl8+7g4LMSTdrm621y7kf9v3SDPnhxLNhcjFJbKECVnmZHTdo+IRO05sNLTH/D3vA6u1X32ehoLC7WFVdheg==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>'    
```

Before this change, (and add the flag `--media=...`), pandoc would pick up the image data from plotly.

After this change, pandoc would pick up the html data, which shows an interactive plot.

Sidenotes:

From
https://github.com/jupyter/nbconvert/blob/6c8d9c8f8374f8785967e0e622ad10826228c193/share/jupyter/nbconvert/templates/lab/index.html.j2#L15-L19
and
https://github.com/jupyter/nbconvert/blob/5d2c5e2b79534c11678b73e707feb74d7827a557/nbconvert/exporters/html.py#L45-L46
nbconvert automatically included `require.min.js` from cdn.

Should we copy it here, namely, when converting from ipynb to html, `require.min.js` is automatically added? (Which is problematic if it is broken into pipes.)

Or perhaps we should just document it in the pandoc manual, what do you think, @jgm?

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -383,9 +383,12 @@ extra-source-files:
                  test/pptx/*.pptx
                  test/pptx/**/*.pptx
                  test/pptx/**/*.native
+                 test/ipynb/*.native
                  test/ipynb/*.in.native
                  test/ipynb/*.out.native
                  test/ipynb/*.ipynb
+                 test/ipynb/*.out.ipynb
+                 test/ipynb/*.out.html
                  test/txt2tags.t2t
                  test/twiki-reader.twiki
                  test/tikiwiki-reader.tikiwiki
--- a/src/Text/Pandoc/Shared.hs
+++ b/src/Text/Pandoc/Shared.hs
@@ -727,17 +727,17 @@ filterIpynbOutput mode = walk go
                  where
                   rank (RawBlock (Format "html") _)
                     | fmt == Format "html" = 1 :: Int
-                    | fmt == Format "markdown" = 2
-                    | otherwise = 3
+                    | fmt == Format "markdown" = 3
+                    | otherwise = 4
                   rank (RawBlock (Format "latex") _)
                     | fmt == Format "latex" = 1
-                    | fmt == Format "markdown" = 2
-                    | otherwise = 3
+                    | fmt == Format "markdown" = 3
+                    | otherwise = 4
                   rank (RawBlock f _)
                     | fmt == f = 1
-                    | otherwise = 3
-                  rank (Para [Image{}]) = 1
-                  rank _ = 2
+                    | otherwise = 4
+                  rank (Para [Image{}]) = 2
+                  rank _ = 3
                   removeANSI (CodeBlock attr code) =
                     CodeBlock attr (removeANSIEscapes code)
                   removeANSI x = x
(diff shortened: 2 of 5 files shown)


jgm:
What does require.min.js do?  Why is it needed?

> when converting from ipynb to html, require.min.js is automatically added? 

That's a bit awkward.  Generally we try to avoid making writer behavior depend on the reader.  Is require.min.js always needed?

jgm:
I think you'll need to update pandoc.cabal too to make sure these test files are included in the package.  Add `test/ipynb/*.html`.

jgm:
Also, I'd prefer if you could squash these into one commit, and hard-wrap the commit message to 80 cols. Otherwise looks good. Thanks!

ickc:
> What does require.min.js do? Why is it needed?
> 
> > when converting from ipynb to html, require.min.js is automatically added?
> 
> That's a bit awkward. Generally we try to avoid making writer behavior depend on the reader. Is require.min.js always needed?

From its website:

>RequireJS is a JavaScript file and module loader. It is optimized for in-browser use, but it can be used in other JavaScript environments, like Rhino and Node. Using a modular script loader like RequireJS will improve the speed and quality of your code. [RequireJS](https://requirejs.org/)

Not an experienced Javascript programmer. From my understanding, it is a module loader like that in node.js but for the browser.

E.g. within the HTML in the above ipynb, there's a script tag, within which there's `require(['plotly'], function(Plotly) {`.

So plotly is not directly loaded, but require.js as a module loader loaded it.

My guess is that since require.js is loaded from cdnjs.cloudflare.com..., require.js would try to load plotly automatically via relative import from the same domain. But I'm not so sure.

So it seems like it is how nbconvert (and their ecosystem) expects things to work for anything that needs js (other than MathJax where nbconvert always loads.) In this case nothing is needed to be done from the nbconvert side to support new jupyter extensions, where they just need to make sure their js is also available in that cdn. I guess this is how all extensions in jupyter works when they need js, such as holoview, bokeh, etc.

I think what can be done here is either

1. write it in pandoc MANUAL and let the user do it themselves when needed.
2. html writer try to detect if there's anything ipynb-input like, such as having cell, output, etc., and if so add the require.js as what nbconvert will always do
3. somehow detect if `require()` is used in the document, such as inside a javascript tag? Seems a little bit trying too smart here...
4. always load require.js but seems a bit like bloatware...

Personally I am ok with option (1).

Oh, direct answer to the question

> Is require.min.js always needed?

No, but it is always loaded in nbconvert. As long as none of the contents has javascript that uses `require(...)`, then it is not needed. In more traditional notebook usage where stuffs are at most png or svg images, this is fine. But much of the appeal in Jupyter nowadays are its interactiveness and many visualization packages uses javascript.

There's a nice summary by pyviz on visualization tools in the Python ecosystem, which is not exhaustive: (orange network is js)

[](https://pyviz.org/overviews/index.html)

ickc:
@jgm, ready to merge.

PR #7749, opened 2021-12-11
docs(ipynb): documents media related tips & tricks

ickc:
For `require.js`, See discussion in #7745, in particular,

https://github.com/jgm/pandoc/pull/7745#issue-1076570646
https://github.com/jgm/pandoc/pull/7745#issuecomment-991381269

I added a commit to documents the use of `--extract-media` or `--self-contained` too. nbconvert behavior is similar to using `--self-contained` here.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -6441,33 +6441,16 @@ soft line breaks in Markdown cells; `--atx-headers` will
 cause ATX-style headings to be used; and `--preserve-tabs` will
 prevent tabs from being turned to spaces.
 
-## Image media
-
-When the ipynb input has media such as svg, png, pdf,
-you may want to use `--self-contained` for any of the supported
-output formats including HTML, or use `--extract-media` in general.
-
-## Javascript media
-
-When converting from ipynb, some Jupyter extensions,
-especially those that uses javascript for visualization,
+Regarding media, `--self-contained`/`--extract-media` can be used
+to include them in the output format,
+whereas some javascript media
 expects the presence of
-[`require.js`](https://github.com/requirejs/requirejs).
-In this case when converting say from ipynb to html,
-you can add the following
-either via `header-includes` in metadata or
-`--include-in-header` on the command line:
+[`require.js`](https://github.com/requirejs/requirejs),
+where you may load this using `header-includes`, such as
 
-```
+```html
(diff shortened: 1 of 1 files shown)


ickc:
@jgm, I take the approach that simply document this behavior. Let me know if you think this is enough.

ickc:
@jgm, did you have a chance to look at the proposed doc and see if this is ok? Thanks.

MANUAL.txt
```diff
@@ -679,7 +679,7 @@ header when requesting a document from a URL:
     necessary, and adjust the images references in the document
     so they point to the extracted files.  Media are downloaded,
     read from the file system, or extracted from a binary
-    container (e.g. docx), as needed.  The original file paths
+    container (e.g. docx, ipynb), as needed.  The original file paths
```
jgm, line 682: Strictly speaking, ipynb is not a "binary container".  Maybe remove the word "binary" on 681?

MANUAL.txt
```diff
@@ -6441,6 +6441,35 @@ soft line breaks in Markdown cells; `--atx-headers` will
...
+</script>
+```
+
+such as
+
+```
+pandoc -s -o require.html require.ipynb \
+-V header-includes='<script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"></script>'
+```
+
```
jgm, line 6472: I'm reluctant to add this to the manual, which is already overly long.
Perhaps this information should be in the FAQ on the website?

ickc: I tried to shorten the description to be more succinct.
I think people are more used to get information like this in the manual than the FAQ,
(where I never went to look for information, but I should now knowing it exists.)

If you insist to make it to FAQ instead,
is it https://pandoc.org/faqs.html ?
Should it be committed to https://github.com/jgm/pandoc-website and which file?

ickc:
We will rebase after the discussion above is settle.

MANUAL.txt
jgm: FAQs live in this repository, in doc/faqs.md.
They get copied to the website from this repo on each release.

jgm: Is there a specific error people would get that would tell us that they need require.js?  If so, it would be natural to have a FAQ like "I'm getting such-and-such error when I try to use JavaScript in ipynb. What does this mean?"

jgm:
I added a FAQ.

PR #7751, opened 2021-12-13
Lua: support topdown traversals

tarleb:
The traversal order of filters can now be selected by setting the key
`traverse` of the filter to either `'topdown'` or `'typewise'`; the
default remains `'typewise'`.

Topdown traversals can be cut short by returning `false` as a second
value from the filter function. No child-element of the returned element
is processed in that case.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -3,6 +3,11 @@ tests: True
 flags: +embed_data_files
 constraints: aeson >= 2.0.1.0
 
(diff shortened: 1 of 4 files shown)


jgm:
Great -- this will really simplify some things people have wanted to do.
(In fact, there's a recentish pandoc-discuss thread on this, if you want to comment there.)

Issue #7752, opened 2021-12-15
Get list of references in Lua filter


tarleb:
I'd like to get the list of references for a document, essentially by exposing the function `T.P.Citeproc.getReferences` to Lua. Here's what I think would be a good way to approach this. The basic idea is to improve JSON support in filters, so `Reference` values can be opaque objects for now.

1. A global function `tojson` is added to the pandoc.utils module (or maybe a separate `pandoc.json` module?). The function uses aeson to do the encoding.
2. Modeled on Lua's `tostring` function, the `tojson` function checks for a `__tojson` metamethod and returns the result of that function if it exists.
3. A `__tojson` metamethod is added to all AST elements that are instances of `ToJSON`.
4. A new `Reference` type is added to Lua; it has no properties except for the `__tostring`, and `__tojson` metamethods.
5. The `getReferences` function is exported as `pandoc.utils.references`.

This would give access to references without us having to decide which Reference properties to expose, and how to bring case-insensitive behavior to Lua. Also, having `tojson` seems like a generally useful feature.

Does it sound like a sensible course of action?

Edit: There'd also be a function `fromjson` which converts a JSON string to a Lua object. It could take an optional second parameter to identify the target type. E.g., `fromjson(input, pandoc.ReaderOptions)`

jgm:
Could you clarify why you want this?/what you'd be able to do with it that you can't do now?

tarleb:
I'm trying to generate [Crossref XML](https://data.crossref.org/reports/help/schema_doc/4.4.2/schema_4_4_2.html) output. The XML should include information on all citations in an article, which I'd like to generate in a filter (or maybe in a custom writer). I don't see a clean solution to do this right now.

jgm:
`getReferences` returns a `[Reference Inlines]`.
If you `tojson` that (using the aeson instance), then you'll get pandoc json for the Inlines content.  That may be pretty inconvenient to work with in lua (even converted to a lua table).
Wouldn't it be more ergonomic to have `getReferences` return a Lua table in which the Inlines content is Lua Inlines objects?

tarleb:
Quite likely, yes. I guess I had assumed that I'd do some hacky back-and-forth through JSON to get the output I need.

[...]

PR #7756, opened 2021-12-17
Lua: add function `pandoc.utils.references`

tarleb:
List with all cited references of a document.

Closes: #7752


tarleb:
The code can be shortened to just

``` haskell
pushReference = pushValue . Aeson.toJSON
```

once there is a suitable `hslua-aeson` that works with the hslua and aeson versions that we use. I released a new version just now that works with hslua-2.0, but it also requires aeson-2.0, which is not on stackage yet.

PR #7765, opened 2021-12-20
Add text wrapping to HTML output.

### LINKED ISSUE #7764: wrapping HTML output (blaze-html dependency)
The [documentation](https://pandoc.org/MANUAL.html#general-writer-options) presently says "Automatic wrapping does not currently work in HTML output." @jgm responds on this list "It's because we use an external library (blaze-html) to produce the HTML, and it doesn't do wrapping."

I [don't see this issue documented on the blaze repo](https://github.com/jaspervdj/blaze-html/issues), should it?

jgm:
New internal module, T.P.Writers.Blaze, exporting `layoutMarkup`.
This converts a blaze Html structure into a doclayout Doc Text.

Note that much test output has changed. This needs to be carefully
examined.

Closes #7764.


tarleb (approved):
I like this change! Getting rid of a special case would be worth the performance penalty.

jgm:
By the way, if the performance penalty is an issue, you can always do `--wrap=none` and then it will be just as fast as before.  (This code path avoids doclayout entirely, unless you're using a template/standalone mode.)

PR #7768, opened 2021-12-21
Lua: add new library function `pandoc.utils.type`.

tarleb:
The function behaves like the default `type` function from Lua's
standard library, but is aware of pandoc userdata types. A typical
use-case would be to determine the type of a metadata value.


tarleb:
This is basically a more general replacement for `.t` fields on metadata value.

jgm:
looks good!

Issue #5221, opened 2019-01-13
Missing cli options in lua filter


hason:
I want  to react to certain options (e.g. `--top-level-division`) in Lua filter. What about create a global variable `PANDOC_OPTIONS` for all options or `PANDOC_WRITER_OPTIONS` for options which were provided to the writer?

tarleb:
Making writer options available to filters could be useful. We should probably do this for JSON filters as well as for Lua filters. JSON filters should be kept as powerful as Lua filters when possible.

jgm:
If we do this, should we provide the raw argv strings or rather a structured Opt?

Note that Text.App.Opt.Opt provides Generic and ToJSON instances.

PR #7784, opened 2021-12-30
Lua: provide global `PANDOC_WRITER_OPTIONS` [API change]

tarleb:
Closes: #5221


tarleb:
The template and syntax map are not exposed, as there is no JSON instance for them and there is no clear benefit of making them available.

src/Text/Pandoc/Lua/Global.hs
```diff
@@ -58,8 +60,11 @@ setGlobal global = case global of
   PANDOC_READER_OPTIONS ropts -> do
     pushReaderOptionsReadonly ropts
     Lua.setglobal "PANDOC_READER_OPTIONS"
+  PANDOC_WRITER_OPTIONS ropts -> do
```
jgm, line 63: Could be less confusing to use `wopts` here.

jgm:
Question: would it make sense to use pushViaJSON for the ReaderOptions as well?  This might help keep things in sync in the future, if ReaderOptions change, with less manual intervention.

tarleb:
My one argument against `pushReaderOptions = pushViaJSON` would be that JSON uses dashes in field names, which are a less nice to use in Lua: `reader.strip_comments = true` vs `reader['strip-comments'] = true`.  But for many of the fields, going through JSON would give a much nicer and consistent interface. I'll hack up a PR for that.

jgm:
That's true.  On the other hand, if it's uniform between JSON and Lua, that's less to remember.
One possibility would be to have pushViaJSON systematically change `-` to `_` in field names, but maybe that's not worth it.

tarleb:
I hadn't considered the possibility of replacing `-` with `_`, that seems nice. The only interesting differences would then be

- `indented_code_classes` would no longer be a `List` (but I want to fix that with the next release of hslua-aeson), and
- `abbreviations` are currently a table with the abbreviations as keys and `true` as values, which is the idiomatic way to represent sets in Lua; it would become a simple List when going through JSON.

I don't believe many people would be impacted by either of these.

jgm:
I guess the "set" representation of abbreviations would be more efficient if it were actually used in a lua filter.  But it probably doesn't matter much.  After all, if efficiency were an issue, the filter could always construct the "set" like table from the list like one.

tarleb:
I think I want to wait for the next hslua-aeson release before doing this, as the released version as a few issues that might get in the way. But there is e58a5ceed8bb629efbfd670bf162b65b7853ab7f as a first step into that direction.

jgm:
Sounds good.  By "doing this", I assume you mean changing `PANDOC_READER_OPTIONS` to use via json.  Is this PR for `PANDOC_WRITER_OPTIONS` ready to merge?

tarleb:
The code should be ok, but I just noticed that the API change isn't really described in the commit message; I will fix that.

tarleb:
Better now, I think.

PR #7788, opened 2021-12-31
Lua: allow to control the PandocMonad type used in pandoc.read

tarleb:
The default behavior of `pandoc.read` is to run the parser in the IO
environment, meaning that it can access the file-system to include
subfiles, e.g. when reading LaTeX or reStructuredText. However, the
mediabag is always discarded, which can be undesirable when reading
binary input like docx or epub.

A fourth option is added to `pandoc.read` that controls the behavior of
read. It can be one of the strings `'io'`, `'global'`, or `'sandbox'`:

-   'io' is the default and gives the old behavior as described above.
-   'global' uses the same environment that was used to read the input
    files; the parser has full access to the file-system and the
    mediabag.
-   'sandbox' works like 'global', but prohibits file-system access.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -3232,6 +3232,14 @@ Returns: the transformed inline element
 
 Parse the given string into a Pandoc document.
 
+The parser is run in the same environment that was used to read
+the main input files; it has full access to the file-system and
+the mediabag. This means that if the document specifies files to
+be included, as is possible in formats like LaTeX,
+reStructuredText, and Org, then these will be included in the
+resulting document. Any media elements are added to those
+retrieved from the other parsed input files.
+
 Parameters:
 
 `markup`:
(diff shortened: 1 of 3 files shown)


tarleb:
An alternative could be to always use the `sandbox` or `global` behavior described above. Then we wouldn't have to add yet another parameter. My main objective here is to preserve the mediabag, as otherwise we lose images from binary input.

tarleb:
I tend towards removing `io`, as it uses a separate CommonState, meaning for example that log messages are lost.

jgm:
Using `global` by default might make sense and would avoid the need for another parameter.  But if you think that there might be cases where using `sandbox` is helpful, this is fine with me. I agree about removing the old behavior.

tarleb:
I'd expect the non-sandboxed behavior to cover all use-cases, so I removed the extra parameter.

PR #7789, opened 2021-12-31
Lua: provide `pandoc.write` function

tarleb:


doc/lua-filters.md
```diff
@@ -2014,6 +2016,109 @@ Fields:
...
     `reject-changes`, and `all-changes` (string)
 
+## WriterOptions {#type-writeroptions}
+
+Pandoc writer options
+
+Fields:
+
+`cite_method`
+:   How to print cites (JSON-like)
```
jgm, line 2026: I don't know how intelligible "JSON-like" is going to be to the reader.
Maybe it would be worth adding a note at the beginning that "JSON-like" means that the value will be the same as in pandoc's JSON serialization of WriterOptions.  But even that has the drawback that most users won't know what that serialization looks like, and won't know how to find out.  (The cli tool will never generate this.)  Maybe, then, we should just list the legal values here?

tarleb: OK, done. I think this warants at least some basic tests, so docs and code won't fall out of sync as easily. I'll add some basic checks.

doc/lua-filters.md
```diff
@@ -2021,6 +2023,121 @@ Fields:
...
+:   Number of levels to include in TOC (integer)
+
+`top_level_division`
+:   Type of top-level divisions; one of 'top-level-part',
+    'top-level-chapter', 'top-level-section', or
+    'top-level-default'. The prefix `top-level` may be omitted
+    when setting this value. (string)
+
+`variables`
+:   Variables to set in template; string-indexed table (table)
```
jgm, line 2133: I guess `variables` is pretty useless unless you can specify a template.
How hard would it be to wrap `compileTemplate`?

src/Text/Pandoc/Writers.hs
```diff
@@ -25,7 +25,6 @@ module Text.Pandoc.Writers
     , writeBibLaTeX
     , writeCommonMark
     , writeConTeXt
-    , writeCustom
```
jgm, line 28: Since any API change might break something, I'd lean towards being conservative and leaving this here.  It's not doing any harm, is it? It's true that it's not consistent with Text.Pandoc.Readers, but a less intrusive change to restore consistency would be to export `readCustom` from that module.

tarleb: I probably should have put this in the commit message: the *reason* for removing it is to avoid a circular dependency (Writers -> Writers.Custom -> Lua -> Writers). Reader/Writer consistency is just my justification for why I believe this change is ok.

doc/lua-filters.md
tarleb: Not hard. We'd need a (opaque) userdata type for `Template Text` and a fairly short function definition of maybe 6 lines of code. Were would that function live? `pandoc.compile_template` or maybe `pandoc.template.compile`?

tarleb: I've added two commits, providing `pandoc.compile_template`.  I'd probably prefer `pandoc.template.compile` better, just in case we decide to expose more template-related functions.

src/Text/Pandoc/Writers.hs
jgm: oh, I see. Well, that's a stronger reason!  (And that's probably the reason readCustom is not in T.P.Readers)

doc/lua-filters.md
jgm: I don't have strong feelings about the naming issue.  I think the other function that would be useful is getDefaultTemplate.

tarleb: I created a `pandoc.template` module with two functions, `compile` to compile a template, and `default` to retrieve the default template for a writer. I left out `compileDefaultTemplate` as it can be built from the other two.

jgm: >  left out compileDefaultTemplate as it can be built from the other two.

But compileDefaultTemplate handles the partials for you; that might be complicated to reproduce with lua?

tarleb: I believe the following should do, but I might have missed something:

``` lua
function compile_default_template (writer)
  return pandoc.template.compile(pandoc.template.default(writer))
end
```

See also the last test in file `pandoc-template.lua`.

tarleb: I should add that `compile` now uses `runWithDefaultPartials` if no template path is given.

jgm: OK, if you've tested it with partials, I'm satisfied.

Issue #7730, opened 2021-12-05
Numbered lists without numbers deleted during conversion


felagund:
On the file attached, I run
`pandoc -f odt -t html example_lost_text.odt > lost_text.html`

I expected to get something like:
```
<h1 id="title">Title</h1>
<ol style="list-style-type: none">
<li>This is a numbered list where the number was removed by pressing backspace when cursor was in front of “T” in “This”. It will get lost during conversion.</li>
</ol>
<h2 id="subtitle">Subtitle</h2>
<ol type="1">
<li>This won’t get lost.</li>
</ol>
```

Instead I got
```
<h1 id="title">Title</h1>
<ol type="1">
</ol>
<h2 id="subtitle">Subtitle</h2>
<ol type="1">
<li>This won’t get lost.</li>
</ol>
```
As can be seen, the first paragraph after title is missing after conversion. The file includes sume custom styles but I think this is generic to numbered lists (at least).

**Pandoc version?**
2.16.2 on Ubuntu 20.10 (downloaded deb from this site) 

**File in question that is not converted correctl**
[example_lost_text.odt](https://github.com/jgm/pandoc/files/7655844/example_lost_text.odt)

EDIT: I just updated the expected result, but I am not that picky about the style property - the text should not disappear above all, it could also translate to `<p>` I guess.

jgm:
@MarLinn can you see what is going on here?

jgm:
Closed by #7793.

PR #7793, opened 2022-01-02
Odt: Add list-header

TuongNM:
This is a fix for #7730

According to the [specifications](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part3-schema/OpenDocument-v1.3-os-part3-schema.html#__RefHeading__1415152_253892949) a list-header is a type of list-item. This is what gets created if you delete the number or bullet point of a list-item.

Since this is a list-item, I treated it as exactly the same.


TuongNM:
Fixed the commit message format and force-pushed again.

jgm:
Great, thank you!

PR #7794, opened 2022-01-02
Switch to hslua-2.1

tarleb:
This allows for some code simplification and improves stability.


jgm:
What's the status of this change?  Is it close to ready?

tarleb:
I keep finding things I want to fix or improve, so my guess is that it's going to take another three to six weeks. It would be good to have hslua 2.1 be the last major release before going from Lua 5.3 to Lua 5.4, so I want to get it right. (Also, pandoc/dockerfiles maintenance eats up a frightening amount of time right now.)

The only somewhat important fix in hslua 2.1 is the one for pandoc/pandoc-lua-marshal#3; the rest is more about general robustness and reverting questionable design decisions. I believe it's ok to delay this a little.

jgm:
I guess the question is whether the next pandoc release should wait for this change?

tarleb:
I'd go with "don't wait". I found the bug through property testing, it should not be triggered by regular real-world input.

BTW, this is the same bug that was triggering the occasional test failures in the old Lua system. I found that is was still there, just masked by the new marshaling method. Finally got to the bottom of it though: hslua failed to request sufficient space on the Lua stack, which caused the Lua stack to overflow in some cases. This only happens for deeply nested structures, which is probably why nobody ever triggered the bug "in the wild".

jgm:
Glad  to hear you  finally uncovered the cause!

tarleb:
Good to go now.

jkachmar:
Related to https://github.com/commercialhaskell/stackage/issues/6416; whenever `pandoc` cuts a release we can unrestrict the bounds on `hslua-2.1.0` (et al.).

Issue #6731, opened 2020-10-07
Lua custom writer interface extensions


wlupton:
A recent pandoc-discuss thread discussed use of Lua custom writers for generating standalone documents. The problem at the moment is that:
* The writer is responsible for generating the toc, but it can't discover that a toc has been requested (because the supplied variables don't include `toc`)
* Even if the writer knew to generate a toc, it doesn't have a way of setting the `table-of-contents` variable (which is where the template will expect to find the generated toc)

Therefore I'd like to suggest:
* The writer should have access to all the writer options, e.g., via a `PANDOC_WRITER_OPTIONS` global variable
* The writer's `Doc()` function's `variables` argument should include those [variables-set-automatically](https://pandoc.org/MANUAL.html#variables-set-automatically) that can reasonably be set at this stage (this could include `toc`)
* The writer's `Doc()` function should be able to modify variables, e.g. (if this is feasible) by modifying its `variables` argument

I suppose if we have (1) then we don't need (2) (not `toc` anyway). In fact I can see that (2) might be regarded as undesirable, because it's insufficient (you need (1) for `toc-depth` and there's no need for `toc-depth` to be a variable because it's not relevant in a template).

One other note (which could go into a separate issue): the man page doesn't document the `table-of-contents` variable. I gather there's some history here (`toc` used to be used both indicate the need for a toc and to contain the generated toc).

tarleb:
Related (same request but for filters): #5221.

tarleb:
Reopening: the new code only adds the global variable, but setting variables is not possible yet.

PR #7795, opened 2022-01-02
Lua writer: allow variables to be set via second return value of `Doc`

tarleb:
New templates variables can be added by giving variable-value pairs as a
second return value of the global function `Doc`.

Example:

    function Doc (body, meta, vars)
      vars.date = vars.date or os.date '%B %e, %Y'
      return body, vars
    end

Closes: #6731

### DIFF
--- a/doc/custom-writers.md
+++ b/doc/custom-writers.md
@@ -50,3 +50,19 @@ the functions in `sample.lua` according to your needs.
 
 ``` {.lua include="sample.lua"}
 ```
+
+# Template variables
+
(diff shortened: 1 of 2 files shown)


jgm:
This definitely seems useful, if a bit unprincipled (since now the Lua filter isn't *just* an AST filter).  Despite this reservation, I'd probably incline towards thinking it's worth doing.

jgm:
Oh, sorry, this is about writers, not filters!  No problem then.

PR #7806, opened 2022-01-06
Org reader: support counter cookies in lists

lucasvreis:
Hi, this is my first contribution to Pandoc. I'm also somewhat new to Haskell, so even though I followed the guidelines and read similar code from other readers, you may need to revise my code styling.

This adds support for counter cookies in org lists. Such cookies are used to override the item counter in ordered lists. In org it is possible to set the counter at any list item, but since Pandoc AST does not support this, I restricted the usage to setting an offset for the entire ordered list, by using the cookie in the first list item.

Note that even though unordered lists do not have counters, Org Mode still parses such cookies in unordered lists and suppresses them in the output, so in the PR I did the same.

Also, even though org-list-allow-alphabetical is disabled in Emacs by default, for some reason alphabetical cookies are always parsed and used in Org Mode regardlessly of whether this option is enabled or the list style is decimal, so I did the same.

E.g.
```
 2. test
 3. test
```
Is parsed as an ordered list starting at 1, as before. This also conforms to Org Mode behaviour.

```
 1. [@2] test
 2. test
```
Is now parsed as an ordered list starting at 2, so that it conforms to Org Mode behaviour.

Note that when parsing
```
 1. [@2] test
 2. [@9] test
```
the second cookie is silenced and the entire list starts at 2. This is because the current Pandoc AST does not support expressing a change in the counter at a specific item.

The support for alphabetical cookies means that
```
 1. [@c] test
 2. test
```
is parsed as an ordered list starting at 3, since this is what Org Mode does in all exporters I checked (regardless of alphabetical settings).

If there are changes that should be made or more tests that should be written, please let me know 🙂

### DIFF
--- a/src/Text/Pandoc/Readers/Org/BlockStarts.hs
+++ b/src/Text/Pandoc/Readers/Org/BlockStarts.hs
@@ -25,8 +25,10 @@ module Text.Pandoc.Readers.Org.BlockStarts
 
 import Control.Monad (void)
 import Data.Text (Text)
-import qualified Data.Text as T
 import Text.Pandoc.Readers.Org.Parsing
+import Text.Pandoc.Definition as Pandoc
+import Text.Pandoc.Shared (safeRead)
+import Data.Char (toLower)
 
 -- | Horizontal Line (five -- dashes or more)
 hline :: Monad m => OrgParser m ()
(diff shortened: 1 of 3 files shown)


lucasvreis:
Regarding alphabetical lists, I'm planning to open a follow-up PR supporting alphabetical lists when `Ext_fancy_lists` is enabled. But I decided to include alphabetical cookies here because they are always enabled in org.

jgm:
Great!  I'll let @tarleb comment on this, since he knows most about the org reader.

tarleb:
Great addition, thank you!

lucasvreis:
@jgm @tarleb thanks! I also finished implementing the `fancy_lists` extension for org [here](https://github.com/lucasvreis/pandoc/commit/c67355b0ca6efd845723e65e622e3d08eb58b73d). Should I open a new PR or open an issue first? I am not sure if it's ok to reuse the extension name or I should add another. I'm also unsure how I should restructure the org tests to add tests to a specific extension, and what I would need to change in the docs.

tarleb:
A new PR would be welcome, no need for an accompanying issue in this case.

The module `Tests.Readers.Org.Inline.Smart` contains some tests which set a specific extension. You can also use a "command" test, if you prefer, see `test/command` -- those are true integration tests, with all the advantages and disadvantages.

Reusing `fancy_list` is a good idea, best to avoid adding more extensions unless really necessary.

Thanks!

PR #7808, opened 2022-01-06
MediaWiki writer: Remove redundant display text for wiki links

lollipopman:
Prior to this commit the MediaWiki writer always added the display
text for a wiki link:

    * [[Help|Help]]
    * [[Bubbles|Everyone loves bubbles]]

However the display text in the first example is redundant since
MediaWiki uses the target as the default display text. The result being:

    * [[Help]]
    * [[Bubbles|Everyone loves bubbles]]

### DIFF
--- a/src/Text/Pandoc/Writers/MediaWiki.hs
+++ b/src/Text/Pandoc/Writers/MediaWiki.hs
@@ -449,10 +449,13 @@ inlineToMediaWiki (Link _ txt (src, _)) = do
   case txt of
      [Str s] | isURI src && escapeURI s == src -> return src
      _  -> return $ if isURI src
-              then "[" <> src <> " " <> label <> "]"
-              else "[[" <> src' <> "|" <> label <> "]]"
-                     -- with leading / it's a link to a help page
-                     where src' = fromMaybe src $ T.stripPrefix "/" src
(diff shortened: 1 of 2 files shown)


lollipopman:
This is my first pull request to pandoc and my first attempt at writing Haskell, so any feedback is appreciated! I wasn't sure the best place to add tests for this change, if someone could provide direction that would be appreciated.

P.S. Thanks for this great piece of software!

jgm:
Great!  This looks good, we'll see what the CI says.
As for tests, there are two possibilities. One is to add a Haskell module test/Tests/Writers/MediaWiki.hs (and integrate it into the tests).  The other (which I think in this case is probably preferable) is just to add a "command test."  You can see examples of this in `test/command/`; it should be pretty clear from the examples what to do.

lollipopman:
@jgm thanks for the feedback, I have added a test based on the format of other tests, let me know what you think.

jgm:
It's best if you name the command test 7808.md; that way we have an easy way of finding the corresponding issue or PR.

lollipopman:
@jgm renamed, thanks

jgm:
Looks good! Congrats on your first Haskell PR.

lollipopman:
@jgm thanks!

Issue #7810, opened 2022-01-07
Org writer: some issues and possible improvements with lists


lucasvreis:
Hi, while writing tests for a org writer PR about lists I found a few corner cases related to newlines that are not handled very well. So I am filing this issue to keep a record of them.
1. This one is actually a bug since it generates invalid org ([see this observation](https://github.com/jgm/pandoc/issues/7557#issuecomment-917047836)): the org file
	```org
	- 
	  #+begin_example
	    aa
	  #+end_example
	- test
	```
	"round trips" (`--from org --to org`) to
	```org
	- #+begin_example
	  aa
	  #+end_example
	
	 - test
	```
	which is not valid org as jgm noted (only plain or para should appear on the bullet line).
2. This next one is also present in RST (the code was copied I guess): when a list item is empty, space is added after the item so that
	  ```
	  - a
	  - 
	  - a
	  - a
	  ``` 
	  "round trips" (`--from org --to org`) to
	  ```
	  -  a
	  -  
	  
	  -  a
	  -  a
	  ```
	
3. This is a suggestion that I think is related to item 2: currently, nested lists (even "tight" ones) are written with unnecessary newlines around them, like so:
	  ```
	  -  a
	  
	     -  b
	     -  b
	  
	  -  a
	  -  a
	  ```
	  maybe it is possible to write without such newlines (like the markdown writer).

jgm:
On 3:  the code contains a comment that says

>   -- ensure that sublists have preceding blank line

but I have no idea why (I didn't write this writer).  It seems to me that this is not necessary at all in org.

jgm:
I believe I've now fixed all three issues.

PR #7811, opened 2022-01-07
Org writer: support starting number cookies

lucasvreis:
This complements #7806 by supporting writing Org ordered lists that start at a specific number.

This commit also slightly changes another thing: before, the org writer indented items to match the markers length like so
```
9.  nine
10. ten
```
I was confused about what it should do when a counter cookie was present (should its length be counted as a marker, or ignored as part of the text like checkboxes?). But then I saw that org mode in emacs did not like this indentation anyway, because when you trigger the automatic formatting (`C-c C-c` above the list) emacs removes the extra spaces. So after this commit, the aligment is removed and the last list is written as
```
9. [@9] nine
10. ten
```
Is this ok?

Also, is one test enough? I did wrote another one with an empty item that would not pass for mysterious reasons, only to discover that the issue was not related to this (see #7810) 🙃

### DIFF
--- a/src/Text/Pandoc/Writers/Org.hs
+++ b/src/Text/Pandoc/Writers/Org.hs
@@ -207,10 +207,9 @@ blockToOrg (OrderedList (start, _, delim) items) = do
                     x         -> x
   let markers = take (length items) $ orderedListMarkers
                                       (start, Decimal, delim')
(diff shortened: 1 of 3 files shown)


lucasvreis:
Oops, I was not aware of those tests. I think they are not run when I filter `-p Org`, tomorrow I will fix them

jgm:
Looks good , thanks!

PR #7812, opened 2022-01-07
Org reader: support alphabetical (fancy) lists

lucasvreis:
As discussed in #7806, this adds support for alphabetical lists in org by enabling the extension Ext_fancy_lists, mimicking the behaviour of Org Mode when org-list-allow-alphabetical is enabled.

Enabling Ext_fancy_lists will also make Pandoc differentiate between the delimiters of ordered lists (periods or closing parentheses). Org does this differentiation by default when exporting to some formats (e.g. plain text) but does not in others (e.g. html and latex), so I decided to copy Pandoc's markdown reader behaviour.

One thing I'm unsure: since this is a separate extension, should the tests be moved to a separate file? Or are they fine in List?

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3162,6 +3162,15 @@ output format.
 Some aspects of [Pandoc's Markdown citation syntax](#citations)
 are also accepted in `org` input.
 
+#### Extension: `fancy_lists` {#org-fancy-lists}
+
+Some aspects of [Pandoc's Markdown fancy lists](#extension-fancy_lists) are also
+accepted in `org` input, mimicking the option `org-list-allow-alphabetical` in
+Emacs. As in Org Mode, enabling this extension allows lowercase and uppercase
+alphabetical markers for ordered lists to be parsed in addition to arabic ones.
+Note that for Org, this does not include roman numerals or the `#` placeholder
+that are enabled by the extension in Pandoc's Markdown.
+
 #### Extension: `element_citations` ####
 
 In the `jats` output formats, this causes reference items to
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -536,6 +536,7 @@ getAllExtensions f = universalExtensions <> getAll f
     extensionsFromList
     [ Ext_citations
     , Ext_smart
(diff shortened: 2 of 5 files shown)


jgm:
Question:  why put this under an extension at all?  If org allows these forms of list markers, and doesn't allow you to disable parsing them, then pandoc shouldn't make it an option either; it should just be default behavior.  Then we don't have to have two slightly different meanings for `fancy_lists` depending on whether it's md or org...

If it does remain under `fancy_lists`, one difference that should be documented is that (as far as I can see from the code) org doesn't allow roman numeral enumerations.

lucasvreis:
@jgm alphabetical lists are not allowed by default in org mode, one must set the variable `org-list-allow-alphabetical` to `t` to use them. I think that's why who wrote the Pandoc reader did not allow them as well. So I thought it made sense to put it behind an extension, especially because it could give headaches to people who do not enable them in org and have lines starting as "K. F. Gauss.." etc (such would be parsed as lists).

The `.` and `)` delimiters however are always allowed in org and the Pandoc supported them too, but before the reader would read both as a `DefaultDelim`. As I said, emacs differentiates them in some exporters like plain text but does not in others. So although both are always allowed I made the reader specify the delim type in the AST only when the extension is enabled. But for me it would also make sense to do this always, I don't have much opinion on this...

lucasvreis:
@jgm  
> If it does remain under `fancy_lists`, one difference that should be documented is that (as far as I can see from the code) org doesn't allow roman numeral enumerations.
Yes, roman are not allowed. In the PR there is a little section in manual.txt where I mentioned this and other differences, do you think it is ok?

tarleb:
LGTM! Could you add a brief note on this to `doc/org.md`?

lucasvreis:
@tarleb sure!

jgm:
Thanks for the explanation; I'm sorry I missed the added part of the manual. Anyway, this looks good!

lucasvreis:
@jgm no worries! I have not written the `doc/org.md` note yet as tarleb suggest, should I do it in another pr since this one is merged? or it is not necessary?

jgm:
Oh yes, another PR would be good for that!

PR #7822, opened 2022-01-10
Docx writer: Handle bullets correctly in lists by not reusing numIds

### LINKED ISSUE #7689: .tex to .docx – issues with itemizing (enumerating)
**Explain the problem.**
.tex source code: https://www.overleaf.com/read/jgwnszzhftjc
pandoc's .docx output:
image
The .docx output has the sentence "extra line after equation" incorrectly itemized as number "2".

**Pandoc version?**
pandoc 2.16.1 on macOS 10.15.7

brisad:
Make sure that we only create one bullet per list item in docx.  In
particular, when a div is a list item, its contained paragraphs will
now no longer wrongly get individual bullets.

This is accomplished by making sure that for each list, we only use
the associated numId once.  Any repeated use would add incorrect
bullets to the document.

Closes #7689


src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -975,14 +975,11 @@ listItemToOpenXML opts numid (first:rest) = do
         _             -> False
   -- Prepend an empty string if the first entry is another
   -- list. Otherwise the outer bullet will disappear.
-  let (first', rest') = if isListBlock first
-                           then (Plain [Str ""] , first:rest)
-                           else (first, rest)
-  first'' <- withNumId numid $ blockToOpenXML opts first'
-  -- baseListId is the code for no list marker:
-  rest''  <- withNumId baseListId $ blocksToOpenXML opts rest'
+  let bs' = if isListBlock (head bs) then (Plain [Str ""]:bs) else bs
```
jgm, line 978: We try to avoid using unsafe partial functions like `head`.
You can use `headMay` from `Safe` or use pattern matching (e.g. a case statement).

src/Text/Pandoc/Writers/Docx/Types.hs
```diff
@@ -111,6 +111,7 @@ data WriterState = WriterState{
        , stDelId          :: Int
        , stStyleMaps      :: StyleMaps
        , stFirstPara      :: Bool
+       , stNumIdUsed      :: Bool
```
jgm, line 114: Put a comment on this to say what it's for?  This wasn't obvious to me.

jgm (changes requested):

src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -975,7 +975,11 @@ listItemToOpenXML opts numid bs = do
         _             -> False
   -- Prepend an empty string if the first entry is another
   -- list. Otherwise the outer bullet will disappear.
-  let bs' = if isListBlock (head bs) then (Plain [Str ""]:bs) else bs
+  let bs' = case bs of
+                 [] -> []
```
jgm, line 979: ghc 9 now gives a "redundant pattern match" warning; it has gotten smart.
We could get rid of that by deleting 968.

brisad: Sorry for requesting review before CI was complete. I have an older ghc version locally, and thought what could possibly go wrong? I've removed the line now. Thanks for the review!

jgm:
Great, thank you!

Issue #7798, opened 2022-01-03
asciidoc checklists


arcnmx:
**Explain the problem.**

Converting the following:

```bash
pandoc -f gfm -t asciidoctor <<EOF
- [ ] a
- [x] b
EOF
```
produces:

```asciidoc
* ☐ a
* ☒ b
```

instead of using the checklist syntax ([see asciidoctor docs](https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/)):

```asciidoctor
* [ ] a
* [x] b
```

This seems related to [the recently fixed org-mode reader/writer support for a similar feature](https://github.com/jgm/pandoc/pull/7098).

**Pandoc version?**

pandoc 2.14.0.3, NixOS

jgm:
Yes, if someone wants to implement this change to the AsciiDoc writer, the code needed should be very similar to the changes to the Org writer linked above.

tarleb:
Fixed in #7832.

arcnmx:
Hm, is `* [X]` equivalent to the documented `* [x]` or `* [*]`?
As far as I can tell, asciidoctor does not recognize the former (which appears to be what was implemented) as a checklist.

tarleb:
Well, that's an issue then. @ricnorr, do you have time to look into this?

jgm:
I can fix this, it's a simple fix.

PR #7832, opened 2022-01-14
asciidoc writer, checklists fix

ricnorr:
Requesting a code review. See issue #7798

### DIFF
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -434,9 +434,6 @@ getDefaultExtensions "jats_articleauthoring" = getDefaultExtensions "jats"
 getDefaultExtensions "opml"            = pandocExtensions -- affects notes
 getDefaultExtensions "markua"          = extensionsFromList
                                           []
-getDefaultExtensions "asciidoc" =  extensionsFromList
-                                     [Ext_auto_identifiers,
-                                      Ext_task_lists]
 getDefaultExtensions _                 = extensionsFromList
                                           [Ext_auto_identifiers]
 
@@ -502,9 +499,7 @@ getAllExtensions f = universalExtensions <> getAll f
   getAll "muse"            = autoIdExtensions <>
     extensionsFromList
     [ Ext_amuse ]
-  getAll "asciidoc"        = autoIdExtensions <>
-    extensionsFromList
-    [ Ext_task_lists ]
+  getAll "asciidoc"        = autoIdExtensions
   getAll "plain"           = allMarkdownExtensions
   getAll "gfm"             = getAll "commonmark"
   getAll "commonmark"      =
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -499,7 +499,9 @@ getAllExtensions f = universalExtensions <> getAll f
   getAll "muse"            = autoIdExtensions <>
     extensionsFromList
     [ Ext_amuse ]
-  getAll "asciidoc"        = autoIdExtensions
+  getAll "asciidoc"        = autoIdExtensions <>
+    extensionsFromList
+    [ Ext_task_lists ]
   getAll "plain"           = allMarkdownExtensions
   getAll "gfm"             = getAll "commonmark"
   getAll "commonmark"      =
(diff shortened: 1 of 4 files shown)


jgm:
In general this looks very good.

Some questions:  I checked the asciidoc documentation for task lists, and found this:
https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/
This suggests that

1) task lists are a built-in feature of asciidoc, not an extension that would make sense to disable, and
2) they only work for bullet lists, not ordered lists.

I might be wrong in one of these impressions -- in which case correct me.
But if these things are right, then I think the code for task lists should be removed from the clause for OrderedList.  And either we should make `task_lists` part of the *default* extensions for asciidoc (not just "all extensions"), or we should make this code unconditional so that it is not sensitive to an extension.  (Either of these would also avoid something I thought was awkward in the test suite, using "all extensions" as if it is a default in `defopts`.)

ricnorr:
> In general this looks very good.
> 
> Some questions: I checked the asciidoc documentation for task lists, and found this: https://docs.asciidoctor.org/asciidoc/latest/lists/checklist/ This suggests that
> 
> 1. task lists are a built-in feature of asciidoc, not an extension that would make sense to disable, and
> 2. they only work for bullet lists, not ordered lists.
> 
> I might be wrong in one of these impressions -- in which case correct me. But if these things are right, then I think the code for task lists should be removed from the clause for OrderedList. And either we should make `task_lists` part of the _default_ extensions for asciidoc (not just "all extensions"), or we should make this code unconditional so that it is not sensitive to an extension. (Either of these would also avoid something I thought was awkward in the test suite, using "all extensions" as if it is a default in `defopts`.)

Yes, you are right, review fix please

src/Text/Pandoc/Extensions.hs
```diff
@@ -434,6 +434,9 @@ getDefaultExtensions "jats_articleauthoring" = getDefaultExtensions "jats"
 getDefaultExtensions "opml"            = pandocExtensions -- affects notes
 getDefaultExtensions "markua"          = extensionsFromList
                                           []
+getDefaultExtensions "asciidoc" =  extensionsFromList
```
jgm, line 437: This should be formatted like the others, so the `=` lines up.

test/Tests/Writers/AsciiDoc.hs
```diff
@@ -90,11 +87,6 @@ tests = [ testGroup "emphasis"
                                            [ "* [ ] a"
                                            , "* [X] b"
                                            ]
-          , test (asciidocWithOpts def) "bullet without task_lists" $
-               bulletList [plain "☐ a", plain "☒ b"] =?> unlines
-                                           [ "* ☐ a"
-                                           , "* ☒ b"
-                                           ]
```
jgm: If we leave `task_lists` as an extension that can be disabled, then it makes sense to test what happens when it is disabled.  But I'm still not sure why we should make task list rendering something you *can* disable with an extension.

So, I'd favor

- removing `task_list` from the available extensions for asciidoc.
- changing the code to do the task list transformation if asciidoctorVariant is set in WriterState, otherwise not.  (A cursory glance through the original asciidoc user guide https://asciidoc-py.github.io/userguide.html doesn't show anything about checklists, so it seems to be an asciidoctor-only feature.)

With these changes, task list syntax will be used automatically with `-t asciidoctor` but not with `-t asciidoc`, which I think is what we want.  Extensions should be used only when it makes sense to disable them.

ricnorr: Fixed, review please

src/Text/Pandoc/Shared.hs
```diff
@@ -665,8 +666,12 @@ taskListItemToAscii = handleTaskListItem toMd
...
-  then handleItem bls
+  then handleTaskListItemNoExtsCheck handleInlines bls
   else bls
+
+handleTaskListItemNoExtsCheck :: ([Inline] -> [Inline]) -> [Block] -> [Block]
+handleTaskListItemNoExtsCheck handleInlines bls =
+  handleItem bls
   where
     handleItem (Plain is : bs) = Plain (handleInlines is) : bs
     handleItem (Para is  : bs) = Para  (handleInlines is) : bs
```
jgm, line 677: I'd like to avoid the API change this would require.
Why not just call `handleTaskListItem`, and add `Ext_task_lists` to the extensions parameter when calling it?

ricnorr: done

jgm:
Great, thanks!

Issue #2750, opened 2016-03-02
MathJax Setup


ickc:
Hi,

I want to have a custom setup of MathJax, like this:

```html
<script type="text/x-mathjax-config">
	MathJax.Ajax.config.path["Contrib"] = "https://cdn.mathjax.org/mathjax/contrib";
	MathJax.Hub.Register.StartupHook("TeX Jax Ready",function () {
		MathJax.Hub.Insert(MathJax.InputJax.TeX.Definitions.macros,{
			cancel: ["Extension","cancel"],
			bcancel: ["Extension","cancel"],
			xcancel: ["Extension","cancel"],
			cancelto: ["Extension","cancel"]
		});
	});
	MathJax.Hub.Config({
		TeX: {
			equationNumbers: { autoNumber: "AMS"},
			extensions: ["[Contrib]/physics/physics.js", "[Contrib]/siunitx/siunitx.js"]
		}
	});
</script>
```

I tried to dig into the documentation and also search through the issues, but there seems to have no reference on it.

Note that according to MathJax, the setup has to come before the MathJax is loaded. So if I put the above code in the beginning of the Markdown file, and using `pandoc --mathjax -s -o` the mathjax is put in the head and hence the setup will comes after it.

So what's the best way to do this? Template? Are there any way to put something in the head of the html without touching the template? And are there any way to config the mathjax in addition to the link to cdn?

Thanks.

jgm:
Use a custom template and put this before the part of the
header where mathjax gets inserted.

ickc:
I see. Thanks!

KeAWang:
Do you have an example of how to do this?

ickc:
First, you need to follow the doc from MathJax to setup your MathJax config.

Assuming you already done that, the problem with using that in pandoc is that the location of loading the MathJax javascript and the `header-includes` is wrong. I think for a while @jgm has considered to change this ordering (there was a PR somewhere in the past.) But in the end he decided to not change that.

So it means your only bet of getting that work is to have a custom template that you have swap the order of that 2 manually. See [Pandoc - Pandoc User’s Guide](https://pandoc.org/MANUAL.html#templates) for more about how you get the template.

If you have more question, provide a reproducible set of files and instructions you have gone through so far that didn't work, and I may be able to give you some more hints.

ickc:
[...]

PR #7833, opened 2022-01-14
fix(html-template)!: load header-includes before math for MathJax config

ickc:
MathJax expect the config comes before loading the MathJax script.
This change of order allows one to config MathJax via header-includes,
which loads before the MathJax script.

This potentially is a breaking change.
However, the only kind math supported by pandoc that is configurable
seems to be katex, and according to
src/Text/Pandoc/Writers/HTML.hs
the way it is configured is hard-coded (katex doesn't seem to offer
MathJax style config that is independent of loading MathJax.)
So it seems it is safe to change this order without breaking
others' documents.

c.f. #2750

Edit: see https://docs.mathjax.org/en/latest/web/configuration.html#configuration-using-an-in-line-script
for the doc to config MathJax.

### DIFF
--- a/data/templates/default.html4
+++ b/data/templates/default.html4
@@ -23,12 +23,12 @@ $endif$
 $for(css)$
   <link rel="stylesheet" href="$css$" type="text/css" />
 $endfor$
-$if(math)$
-  $math$
-$endif$
 $for(header-includes)$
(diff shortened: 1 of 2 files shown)


jgm:
looks like a good idea!

Issue #7831, opened 2022-01-13
Lua error w/ Pandoc 2.17: "Error running Lua: pandoc CommonState expected, got pandoc CommonState"


jjallaire:
I am seeing this error _sometimes_ occur when running pandoc. (but its actually not deterministic, sometimes files succeed and sometimes they fail). I'd like to find a reproducible case for this -- any idea on what sorts of states could trigger this?

Possibly totally unrelated but I'll mention it here in case it gives a clue: I have several other documents that sometimes succeed and sometimes fails with the message `pandoc failed`. It doesn't appear to be deterministic either. These files all rendered w/o trouble in Pandoc 2.16.2.

jjallaire:
I should add that these render are using quarto (https://quarto.org) which makes very extensive use of Lua filters (we have dozens and dozens of filters chained together). It's possible the non-determinism exists because there is some sort of memory corruption that doesn't always manifest. I know that there were many changes to the Lua subsystem made in Pandoc 2.17 -- if there are things you can think of that could tickle a memory management problem that would give me something to start with.

 Also, if there is a way to run pandoc with enhanced diagnostics (esp. for Lua) I could also try that.

jgm:
@tarleb will be able to figure out which of the many Lua subsystem changes might be relevant to this.  (He'd also know about diagnostics.)

jgm:
If you can give us a complete example that sometimes gives this error, it would help.  (Including all the Lua filters used...ideally, disentangled from quarto and using just pandoc.)

jgm:
One thing that would be helpful to know is whether any of your filters use `pandoc.read`.

jjallaire:
We do indeed use `pandoc.Read` frequently.

That said, I discovered `pandoc --verbose` which helped me narrow down which filter exhibited the problem, and ran with only that filter (which doesn't use `pandoc.Read`). It turned out that the following idiom was causing the crash (code shortened for brevity, our code included a list of every language that Pandoc highlights):

```lua
local kHighlightClasses = pandoc.List({
  "abc",
  "actionscript",
  "ada",
  "agda",
  "apache",
  "asn1",
  "asp",
})

function isHighlightClass(class)
  return kHighlightClasses:includes(class)
end
```

We changed to this and the crash no longer occurred:

```lua
local kHighlightClasses = {
  "abc",
  "actionscript",
  "ada",
  "agda",
  "apache",
  "asn1",
  "asp"
}

function isHighlightClass(class)
[...]

PR #7834, opened 2022-01-14
Use updated pandoc-lua-marshal

tarleb:
Fixes: #7831


tarleb:
Fixed via 96db10c8d7284beca356da037eacd4391328813e instead.

jgm:
Glad you found the fix so fast!

PR #7844, opened 2022-01-17
Docx writer: Separate tables even with RawBlocks between

### LINKED ISSUE #7724: bug: `org` → `docx` writer loses whitespace when separated by `#+latex` keyword
Hi, 

**Explain the problem.**
Exporting from `org`, the `docx` writer seems to eat up too much white space between tables if there's a `#+latex:` line in between. It does not seem to happen with e.g. html.

Example: 
```sh 
$ cat test0.org 
* test0
| a | b |

| c | d |

$ cat test1.org 
* test1
| a | b |

#+latex:

| c | d |

$ pandoc test*.org -f org -t docx -o test.docx
```
The output is basically

> 
> **test0**
> |a| b|
> |--|--|
> 
> |c| d|
> |--|--|
> 
> **test1**
> |a| b|
> |--|--|
> |c |d | 
 
But there shouldn't have been a difference between the two sections; both sections should have been like `test0`. See screenshot at end.

HTML and LaTeX exports work as expected (e.g. `pandoc test*.org -f org -t {html,latex} --standalone`).

**Pandoc version?**
pandoc v2.14.1 on GNU/Linux and Windows. 

image

Thanks,
Rasmus

brisad:
Adjacent docx tables need to be separated by an empty paragraph. If
there's a RawBlock between tables which renders to nothing, be sure to
still insert the empty paragraph so that they will not collapse
together.

Fixes #7724


brisad:
Hi, this solution feels pretty hacky. But I don't know if there's a good way to detect whether a block will render or not?

jgm:
It looks fine to me.  Maybe it would be slightly more elegant to do something like
```hs
blocksToOpenXML opts = fmap concat . mapM (blockToOpenXML opts) . separateTables . filter isForeignRawBlock
```
rather than using `dropWhile`?  But it's a matter of taste; if you like that suggestion better, go ahead and change it, otherwise we'll merge this!

Thanks!

brisad:
Yes, I agree that is more elegant, and easier to understand. I actually started out with something similar, but later it somehow got baked in into `separateTables`. Can't remember exatly why, but I think I was unsure if it was OK to remove the blocks completely.

I'll push an update later with the suggested approach.

By the way, by foreign, you mean non-openxml blocks, right? Then I will negate the filter expression, so that they are excluded.

jgm:
Sorry, yes, I meant `filter (not . isForeignRawBlock)` (the name was just a suggestion).

jgm:
Thanks!

Issue #7845, opened 2022-01-18
KindleGen has been replaced by KindlePreviewer


bkmgit:
**Explain the problem.**

KindleGen has been replaced by KindlePreviewer. Update link in documentation. See
- https://www.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211
- https://github.com/ciromattia/kcc/issues/371

PR #7846, opened 2022-01-18
Update link KindleGen -> KindlePreviewer

bkmgit:
Addresses https://github.com/jgm/pandoc/issues/7845

### DIFF
--- a/doc/epub.md
+++ b/doc/epub.md
@@ -7,11 +7,9 @@ Starting with version 1.6, pandoc can produce output in the [EPUB]
 electronic book format. EPUB books can be viewed on iPads, Nooks, and
 other electronic book readers, including many smart phones. (They can
 also be converted to Kindle books using the GUI only [KindlePreviewer] 
-on Windows and Mac OSX.  [KindleGen] which offers a command line
-interface has been deprecated. Binaries of the last KindleGen release
-for [Linux], [Mac OSX] and [Windows] can still be found on the internet,
-though may not be suitable for the latest releases of these operating
-systems.)
+on Windows and Mac OSX.  [KindleGen] has which offers a command line
+interface and supports Linux, Mac OSX and Windows has been deprecated.
+Binaries can still be found on the internet.)
 
 This means that it's now very easy to produce an electronic book!
 Let's try it.
@@ -158,9 +156,6 @@ of the image,  increasing accessibility for blind users.
 
 [KindlePreviewer]: https://www.amazon.com/Kindle-Previewer/b?node=21381691011
 [KindleGen]: https://www.amazon.com/gp/feature.html?docId=1000765211
-[Linux]: https://archive.org/details/kindlegen2.9
-[Mac OSX]: https://web.archive.org/web/20190905040839/https://www.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211
-[Windows]: https://archive.org/details/kindlegen_win32_v2_9
 [EPUB]:  https://en.wikipedia.org/wiki/EPUB
 [Pro Git]: https://git-scm.com/book/en/v2
 [Creative Commons]: https://creativecommons.org/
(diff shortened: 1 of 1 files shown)


alerque (changes requested):
I don't think this is a clear win. Yes Amazon no longer supports `kindlegen`, but their replacement for it is not a drop in replacement. The new "Kindle Previewer" es not have the CLI features of `kindlegen` and it is not even available for Linux. The last released version of the original `kindlegen` may not be endorsed by Amazon but it is available through various package managers and the internet archive. It is more suited for use in many workflows that would involve Pandoc.

If any change is made it should mention both apps with a note that the new one is GUI and Win/Mac only, the deprecated one is CLI and available for Win/Mac/Linux.

jgm:
@alerque if we do that we should say something about how one can find the cli kindlegen.

bkmgit:
Dropping of Linux support is unfortunate. Versions of KindleGen can be found at
- Linux https://archive.org/details/kindlegen 
- Mac OSX https://web.archive.org/web/20190905040839/https://www.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211
- Windows https://archive.org/details/kindlegen_win32_v2_9
If these are mentioned, they will become outdated (Mac OSX is 32 bit only, and will only work on Intel based Macs), and if one chooses a random link, one may install malware.

Some workflows are moving to [Calibre](https://calibre-ebook.com/), for example
https://github.com/DistributedProofreaders/guiguts/issues/208

[Amazon](https://kdp.amazon.com/en_US/help/topic/G79CTKR8BX79E96L) recommends Pandoc.

bkmgit:
Maybe also helpful is https://github.com/ystyle/kaf-cli/releases/kindlegen/

alerque:
Here is a better link to the Linux kindlegen package: https://archive.org/details/kindlegen2.9

The archive.org link is similar but it was repackaged by somebody and doesn't have all the original files. This is the same gzip file originally posted by Amazon as can be verified by checksums. There are several others floating around out there that claim to be newer versions or reposts that have been tampered with, so verifying the checksums is actually important rather than just running some binary you found on the net!

alerque (approved):

bkmgit:
Ok, added links.

alerque (approved):

doc/epub.md
```diff
@@ -6,8 +6,10 @@ author: John MacFarlane
 Starting with version 1.6, pandoc can produce output in the [EPUB]
 electronic book format. EPUB books can be viewed on iPads, Nooks, and
 other electronic book readers, including many smart phones. (They can
-also be converted to Kindle books using [KindlePreviewer] on Windows 
-and Mac OSX.)
+also be converted to Kindle books using the GUI only [KindlePreviewer] 
+on Windows and Mac OSX.  [KindleGen] has which offers a command line
+interface and supports Linux, Mac OSX and Windows has been deprecated.
+Binaries can still be found on the internet.)
```
mb21: typo and a suggestion:

```suggestion
on Windows and Mac OSX. [KindleGen] – which offers a command line
interface and supports Linux, Mac OSX and Windows – has been deprecated,
but binaries can still be found on the internet.)
```

bkmgit: Thanks.

alerque (approved):

mb21 (approved):
thanks, looks good to me!

perhaps @jgm would like to have a last word...

doc/epub.md
```diff
@@ -156,6 +158,9 @@ of the image,  increasing accessibility for blind users.
 
 [KindlePreviewer]: https://www.amazon.com/Kindle-Previewer/b?node=21381691011
 [KindleGen]: https://www.amazon.com/gp/feature.html?docId=1000765211
+[Linux]: https://archive.org/details/kindlegen2.9
+[Mac OSX]: https://web.archive.org/web/20190905040839/https://www.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211
+[Windows]: https://archive.org/details/kindlegen_win32_v2_9
```
jgm, line 161: I think it's confusing to make the reference link `[Linux]` refer to this rather than to something about Linux.
Better to make the reference `[KindleGenLinux]` (or something like that), you can then form the link as `[Linux][KindleGenLinux]`.

bkmgit: Thanks. Done.

jgm:
Great, thanks.

PR #7849, opened 2022-01-18
Add pandoc-lua-marshal to Nix shell

evenbrenden:
Add missing dependency to Nix shell. `pandoc-lua-marshal` is currently broken in Nixpkgs, so building in the Nix shell still fails (assuming Nixpkgs is used as input). Nevertheless, this is one step closer to a working Nix build.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -57,6 +57,7 @@ let
     network
     network
     network-uri
+    pandoc-lua-marshal
     pandoc-types
     parsec
     process


alerque (approved):

tarleb:
I don't have any experience with nix, but do you know what we'd need to fix in pandoc-lua-marshal?

evenbrenden:
> I don't have any experience with nix, but do you know what we'd need to fix in pandoc-lua-marshal?    
    
This would typically be fixed in Nixpkgs, the repo that manages the dependencies listed in `shell.nix`. From what I gather, new and updated packages are automatically pulled from Hackage to Nixpkgs. `pandoc-lua-marshal` was added in [this PR](https://github.com/NixOS/nixpkgs/pull/148359), but the build currently fails because the `hslua` and `tasty-lua` packages in Nixpkgs are too old. These will probably be updated on a later pull from Hackage. I can take a closer look.

tarleb:
I see, thanks. Might be time for me to learn some nix.

evenbrenden:
https://github.com/NixOS/nixpkgs/pull/160733 fixes `pandoc-lua-marshal`.

PR #7850, opened 2022-01-19
Add diff syntax highlighting language to listings

bollenn:
Add `diff` syntax highlighter to the `listings` module

Idea borrowed from https://tex.stackexchange.com/a/106129


jgm:
I think this is too much in the default template for a special case that's really a workaround for something missing in listings.  You can do all of this with `--include-in-header`.  Just make sure your code block is like this:
````
``` {language=Diff}
...
```
````
to work around the missing "Diff" in the lookup table.

bollenn:
Confirm the alternative solution works well... thx

PR #7851, opened 2022-01-19
Search for metadata files in $DATADIR/metadata

### LINKED ISSUE #5876: Support metadata directory in data directory
Since both templates and defaults can be placed in their respective `$DATA_DIR/templates/` and `$DATA_DIR/defaults/` why not also metadata files in `$DATA_DIR/metadata/`?

evenbrenden:
If files specified with --metadata-file are not found in the working
directory, look in $DATADIR/metadata.

Closes #5876.

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -219,7 +219,7 @@ convertWithOpts opts = do
       case optMetadataFiles opts of
         []    -> return mempty
         paths -> mconcat <$>
-           mapM (\path -> do raw <- readFileStrict path
+           mapM (\path -> do raw <- readMetadataFile path
                              yamlToMeta readerOpts (Just path) raw) paths
 
     let transforms = (case optShiftHeadingLevelBy opts of
--- a/src/Text/Pandoc/Class/PandocMonad.hs
+++ b/src/Text/Pandoc/Class/PandocMonad.hs
@@ -46,6 +46,7 @@ module Text.Pandoc.Class.PandocMonad
   , getResourcePath
   , readDefaultDataFile
   , readDataFile
+  , readMetadataFile
   , fillMediaBag
   , toLang
   , setTranslations
@@ -585,6 +586,25 @@ readDataFile fname = do
          if exists
             then readFileStrict (userDir </> fname)
             else readDefaultDataFile fname
+
+-- | Read metadata file from the working directory or, if not found there, from
+-- the metadata subdirectory of the user data directory.
+readMetadataFile :: PandocMonad m => FilePath -> m B.ByteString
+readMetadataFile fname = do
+  existsInWorkingDir <- fileExists fname
+  if existsInWorkingDir
+     then readFileStrict fname
+     else do
+       dataDir <- getUserDataDir
+       case dataDir of
+         Nothing ->
+           throwError $ PandocCouldNotFindMetadataFileError $ T.pack fname
+         Just userDir -> do
+           let path = userDir </> "metadata" </> fname
+           existsInUserDir <- fileExists path
+           if existsInUserDir
+              then readFileStrict path
+              else throwError $ PandocCouldNotFindMetadataFileError $ T.pack fname
 
 -- | Read file from from the default data files.
 readDefaultDataFile :: PandocMonad m => FilePath -> m B.ByteString
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -644,7 +644,9 @@ header when requesting a document from a URL:
     files specified later on the command line will be preferred
     over those specified in earlier files. Metadata values
     specified inside the document, or by using `-M`, overwrite
-    values specified with this option.
+    values specified with this option. The file will be searched
+    for first in the working directory, and then in the `metadata`
+    subdirectory of the user data directory (see `--data-dir`).
 
 `-p`, `--preserve-tabs`
 
@@ -1512,6 +1514,7 @@ Nonzero exit codes have the following meanings:
    93 PandocIpynbDecodingError
    94 PandocUnsupportedCharsetError
    97 PandocCouldNotFindDataFileError
+   98 PandocCouldNotFindMetadataFileError
    99 PandocResourceNotFound
 ----- ------------------------------------
 
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -251,6 +251,9 @@ extra-source-files:
                  test/command/3510-export.latex
                  test/command/3510-src.hs
                  test/command/3971b.tex
+                 test/command/5876.yaml
+                 test/command/5876/metadata/5876.yaml
+                 test/command/5876/metadata/command/5876.yaml
                  test/docbook-chapter.docbook
                  test/docbook-reader.docbook
                  test/docbook-xref.docbook
--- a/src/Text/Pandoc/Error.hs
+++ b/src/Text/Pandoc/Error.hs
@@ -54,6 +54,7 @@ data PandocError = PandocIOError Text IOError
                  | PandocFilterError Text Text
                  | PandocLuaError Text
                  | PandocCouldNotFindDataFileError Text
+                 | PandocCouldNotFindMetadataFileError Text
                  | PandocResourceNotFound Text
                  | PandocTemplateError Text
                  | PandocAppError Text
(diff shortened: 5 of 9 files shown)


src/Text/Pandoc/App.hs
```diff
@@ -218,9 +218,13 @@ convertWithOpts opts = do
...
-        paths -> mconcat <$>
-           mapM (\path -> do raw <- readFileStrict path
-                             yamlToMeta readerOpts (Just path) raw) paths
+        paths -> do
+          mconcat <$>
+            mapM
+              (\path -> do
+                raw <- readMetadataFile path
+                yamlToMeta readerOpts (Just path) raw)
+              paths
```
jgm, line 227: Here it would be helpful if your change were the minimal one necessary: just changing `readFileStrict` to `readMetadataFile` and leaving everything else the same.  With the reformatting, it becomes more difficult to see what has changed.  (Also, there's now a redundant `do`.)

src/Text/Pandoc/Class/PandocMonad.hs
```diff
@@ -586,6 +587,23 @@ readDataFile fname = do
...
             else readDefaultDataFile fname
 
+-- | Read metadata file from the working directory or, if not found there, from
+-- the metadata subdirectory of the user data directory.
+readMetadataFile :: PandocMonad m => FilePath -> m B.ByteString
+readMetadataFile fname = do
+  existsInWorkingDir <- fileExists fname
+  if existsInWorkingDir then readFileStrict fname
+  else do
+    dataDir <- getUserDataDir
```
jgm, line 597: The style of the rest of the pandoc code is:
```
if foo
   then bar
   else baz
```
Better to keep to a consistent style.

src/Text/Pandoc/Class/PandocMonad.hs
```diff
@@ -586,6 +587,23 @@ readDataFile fname = do
...
+    case dataDir of
+      Nothing ->
+        throwError $ PandocCouldNotFindDataFileError $ T.pack fname
+      Just userDir -> do
+        let path = userDir </> "metadata" </> fname
+        existsInUserDir <- fileExists path
+        if existsInUserDir then readFileStrict path
+        else throwError $ PandocCouldNotFindDataFileError $ T.pack fname
```
jgm, line 605: This isn't really the appropriate error since it's not a data file.

test/command/5876-datadir-metadata.md
```diff
@@ -0,0 +1,18 @@
+```
```
jgm, line 1: Usually we just name these with the number, e.g. 5876.md.

test/command/5876-datadir-metadata/metadata/test-metadata.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
jgm, line 1: You'll need to make sure this test file is included in extra-source-files in pandoc.cabal, or it won't be included in the source distribution.
Maybe also worth testing two further cases:

- the case where the file exists in the working directory AND in DATADIR/metadata
- the case where the file exists in neither place (yielding an error)

src/Text/Pandoc/Class/PandocMonad.hs
evenbrenden: I could add `PandocCouldNotFindMetadataFileError`? That would be consistent with `readDataFile` and `readMetadataFile`.

test/command/5876-datadir-metadata.md
evenbrenden: I'll try to write the other tests that you proposed too. There appears to be a convention for enumerating tests that belong to the same issue, e.g. `5876a.md`, `5876b.md`. Is that OK?

Should mention that [CONTRIBUTING.md](https://github.com/jgm/pandoc/blob/master/CONTRIBUTING.md#tests) suggests using both issue number and feature in the test file names.

src/Text/Pandoc/App.hs
evenbrenden: Absolutely, sorry about that. Changed in fixup.

src/Text/Pandoc/Class/PandocMonad.hs
evenbrenden: Changed in fixup.

src/Text/Pandoc/Class/PandocMonad.hs
evenbrenden: Added `PandocCouldNotFindMetadataFileError`.

test/command/5876-datadir-metadata.md
evenbrenden: Changed in fixup.

src/Text/Pandoc/Class/PandocMonad.hs
evenbrenden: Note that a missing metadata file specified with `--metadata-dir` results in the more general `PandocIOError` (via `readFileStrict`) in `master`. This PR changes that.

test/command/5876-datadir-metadata/metadata/test-metadata.yaml
evenbrenden: Added resources to `extra-source-files` as well as two new tests. There's also the case where the file exists in the working directory, but not in the user directory, which is covered by multiple existing tests. I could still add it under this issue for completeness, though.

test/command/5876-datadir-metadata.md
jgm: Oh, didn't realize that about contributing.md!
I guess it's probably not bad advice, and it looks like we've done it sometimes.
So, as you like.

For multiple tests belonging to the same issue, simply include them all in the same file.  A command test file can contain any number of command tests (separate code blocks).  We don't really have a convention of putting them in separate files with letter suffixes.

test/command/5876-datadir-metadata/metadata/test-metadata.yaml
jgm: Note that you shouldn't put 7851.md and the like in extra-source-files, because it's already covered by a glob (`test/command/*.md`).  However, the yaml file in the subdirectory wasn't covered by a glob.  Note that `make check-cabal` will automatically check to make sure that there aren't committed test files that aren't in the extra-source-files stanza.  I don't think we necessarily need the extra test case you describe, but it wouldn't hurt.  Up to you.

test/command/5876-datadir-metadata.md
evenbrenden: Ah, I see. I gathered all tests in `5876.md` then. I think an extra suffix is unnecessary in this case since the tests are pretty clear about what is being tested.

test/command/5876-datadir-metadata/metadata/test-metadata.yaml
evenbrenden: Great, I ran the check and everything should be included now. I also added a fourth test for when there's no user directory specified and the file does not exist. Then all four possible outcomes of `readMetadataFile` are covered in `5876.md`.

evenbrenden:
Squashed the fixups into a single commit.

jgm:
This looks great!  Just a couple small things now.

1.  Update list of Exit Codes in manual (heading `# Exit codes`) to add the new 98 code
2. Modify your commit message so it clearly marks the two API changes with `[API change]`.  These are the new constructor on PandocError and the new exported function from T.P.Class (via PandocMonad).

evenbrenden:
> This looks great! Just a couple small things now.
> 
>     1. Update list of Exit Codes in manual (heading `# Exit codes`) to add the new 98 code
> 
>     2. Modify your commit message so it clearly marks the two API changes with `[API change]`.  These are the new constructor on PandocError and the new exported function from T.P.Class (via PandocMonad).

Like this?

jgm (approved):
Great.

jgm:
Now that I've merged this, a potential issue occurs to me.
(Perhaps it also arises for data files and templates.)
Suppose someone says `--metadata-file ../foo.yaml`.  If `foo.yaml` isn't found in the parent directory of the working directory, do we really want to look for it in the parent of the user data directory, which would be *outside of the user data directory*?

evenbrenden:
Hm, can't imagine that anyone wants to do that on purpose. I checked the behaviour for `--template`, which is different: If the path specified isn't found directly, `$DATADIR/templates` is searched by filename. That is, the path is stripped. This means that you can't go outside of the data directory, but also that you can't search subdirectories of `$DATADIR/templates`. Not sure what the best solution is, but I think it should be consistent for any file type. I can open an issue for further discussion.

jgm:
One approach would be to allow subdirectories in both cases, but only if they don't contain `..`.
But yeah, we should track this on a separate issue.

evenbrenden:
Created https://github.com/jgm/pandoc/issues/7861.

PR #7868, opened 2022-01-27
Don't read files outside of user data directory

evenbrenden:
If a file path does not exist relative to the working directory, but
it does exist relative to the user data directory, and it exists outside
of the user data directory, do not read it. This applies to readDataFile
and readMetadataFile in PandocMonad and, by extension, any module that
uses these by passing them relative paths.

Fix bug in makeCanonical where consecutive ".."s was removed.

### DIFF
--- a/src/Text/Pandoc/Class/PandocMonad.hs
+++ b/src/Text/Pandoc/Class/PandocMonad.hs
@@ -574,11 +574,25 @@ getDefaultReferencePptx = do
      Nothing   -> foldr addEntryToArchive emptyArchive <$>
                      mapM pathToEntry paths
 
--- | Read file from user data directory or,
--- if not found there, from the default data files.
+-- | Checks if the file path is relative to a parent directory.
+isRelativeToParentDir :: FilePath -> Bool
+isRelativeToParentDir fname =
+  let canonical = makeCanonical fname
+   in length canonical >= 2 && take 2 canonical == ".."
+
+-- | Returns possible user data directory if the file path refers to a file or
+-- subdirectory within it.
+checkUserDataDir :: PandocMonad m => FilePath -> m (Maybe FilePath)
+checkUserDataDir fname =
+  if isRelative fname && not (isRelativeToParentDir fname)
+     then getUserDataDir
+     else return Nothing
+
+--- | Read file from user data directory or,
+--- if not found there, from the default data files.
 readDataFile :: PandocMonad m => FilePath -> m B.ByteString
 readDataFile fname = do
-  datadir <- getUserDataDir
+  datadir <- checkUserDataDir fname
   case datadir of
        Nothing -> readDefaultDataFile fname
        Just userDir -> do
@@ -595,7 +609,7 @@ readMetadataFile fname = do
   if existsInWorkingDir
      then readFileStrict fname
      else do
-       dataDir <- getUserDataDir
+       dataDir <- checkUserDataDir fname
        case dataDir of
          Nothing ->
            throwError $ PandocCouldNotFindMetadataFileError $ T.pack fname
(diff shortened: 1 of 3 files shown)


src/Text/Pandoc/Class/PandocMonad.hs
```diff
@@ -638,6 +652,7 @@ makeCanonical :: FilePath -> FilePath
 makeCanonical = Posix.joinPath . transformPathParts . splitDirectories
  where  transformPathParts = reverse . foldl' go []
         go as     "."  = as
+        go ("..":as) ".." = ["..", ".."] <> as
```
evenbrenden, line 655: Maybe this should be in a separate PR?

jgm: Is this really a bug?  Seems to me that this case will never occur.
We start with the accumulator as an empty list.
The only way something can get added to that list is if it matches the third clause `go as x`.  But we will never match that with `x` == `".."`, because `".."` always matches the second clause and gets discarded.  Am I missing something?

evenbrenden: Without line 655, `makeCanonical "../../file"` will evaluate to `"file"`. I think it goes like this:

1. `go [] ".." = ".." : []` (657)
2. `go [_:[]] ".." = []` (656)
3. `go [] "file" = ["file"]` (657)

With line 655:

1. `go [] ".." = ".." : []` (657)
2. `go ("..":[]) ".." = ["..", ".."] <> []` (655)
3. `go ["..", ".."] "file" = "file" : ["..", ".."]` (657)

I see that I messed up the alignment, will fix.

jgm: Oh right!  We don't match that second line when `as` is empty.

src/Text/Pandoc/Class/PandocMonad.hs
```diff
@@ -574,11 +574,25 @@ getDefaultReferencePptx = do
...
--- | Read file from user data directory or,
--- if not found there, from the default data files.
+-- | Checks if the file path is relative to a parent directory.
+isRelativeToParentDir :: FilePath -> Bool
+isRelativeToParentDir fname =
+  let canonical = makeCanonical fname
+   in length canonical >= 2 && take 2 canonical == ".."
+
+-- | Returns user data directory if it exists and if the file path exists
+-- within it.
```
jgm, line 584: This comment might be a bit misleading, since we aren't really checking for the existence of the file path.
Not sure how to word it better, though.

src/Text/Pandoc/Class/PandocMonad.hs
jgm: I'd prefer if this were  a separate commit (not necessarily in a separate PR), so it stands out in the history.

src/Text/Pandoc/Class/PandocMonad.hs
evenbrenden: Hm. Maybe that says something about the function itself. Trying something else, what do you think?

src/Text/Pandoc/Class/PandocMonad.hs
evenbrenden: Done!

jgm:
I don't see why the `placeholder` test file is needed?

evenbrenden:
> I don't see why the `placeholder` test file is needed?

Git doesn't check in empty folders and we want the data directory to exist for the test.

jgm:
OK, great!

jgm:
@evenbrenden See <https://groups.google.com/d/msgid/pandoc-discuss/bf4007b6-45b6-aa07-d7da-7fc4b424ac2d%40reagle.org>

I'm wondering whether this change (or maybe one of the other related ones) may have messed up defaults file lookups in the user data directory.  Can you have a look?

jgm:
@evenbrenden I think this is due to my commit 545c0911aa5d7d91280c5213c6d57b3e634ef1e5

Issue #5990, opened 2019-12-13
better "Defaults file" documentation


allefeld:
The documentation under [Defaults](https://pandoc.org/MANUAL.html#default-files) should be improved, for example like this:

![image](https://user-images.githubusercontent.com/1738353/70836590-85aca300-1df8-11ea-9a1d-de65896812b5.png)

I can share the underlying Markdown code, and also extend it to include all possible entries.

jgm:
Yes, please do share.

allefeld:
Will do, but I have to extend it to all entries first.

allefeld:
Here it is. A complete list of command line options (according to  `pandoc --help`), and an almost complete list of their defaults file counterparts.

Markdown source and pdf:
[defaults.txt](https://github.com/jgm/pandoc/files/3985757/defaults.txt)
[defaults.pdf](https://github.com/jgm/pandoc/files/3985759/defaults.pdf)

It would be good if you could have a look, there are a few things I could not figure out, and others where I'm not certain I got it right. Let me know, and I'll make a new version.

One thing that is confusing me particularly: If I understand the YAML specification correctly, an omitted value is interpreted as `null`, which in a boolean setting in most languages is treated as equivalent to `false`. Consistent with that, `highlight-style:  ` and `highlight-style: null`  mean `--no-highlight`. Yet `output-file: ` has the meaning `--output=-`. Why not just require `output-file: -`? Or would that violate YAML syntax?

I also see some (unnecessary?) changes from command line options to defaults file entries. Generally, `--option=VALUE` translates to `option: VALUE`. If an option can be repeated on the command line, it instead translates to

    options:
      - VALUE

So far, so logical.

Deviating from that:

-   `--read` (which already has the alias `--from`) translates into `readER:`. Same for `--write`.
-   The non-repeatable `--resource-path=`, which takes a string with `:` as separator, translates into a YAML list, not needing a separator. Similar for `--indented-code-classes=`.
-   Different options `--filter=` and `--lua-filter=` translate into the same `filters:`.
-   The repeatable options `--css`, `--include-in-header`, `--include-before-body`, `--include-after-body`, and `--metadata` do not have an appended `s` in translation. Granted, that would be awkward. But logical. :)

A more consistently logical translation could make this table unnecessary.

[...]

PR #7872, opened 2022-01-29
New defaults files documentation

allefeld:
I took a stab at integrating the new defaults file documentation, as discussed in #5990, into the Pandoc manual.

I took out the original example defaults file code close to the beginning, and added the tables in subsections at the end. In the process, I slightly changed the text which was referencing the example before, and added some text referencing the tables after. It could probably be done more smoothly, but I didn't want to modify the existing text more than minimally. Please have a look.

I also noticed an inconsistency in the manual; it uses 5 times the term "default file" and 9 times the term "defaults file". I took the liberty to change that to always use "defaults file".

I checked the new version by `make man/pandoc.1` as well as `pandoc -s MANUAL.txt -o MANUAL.html`. It displays decently, but not perfectly. As mentioned in the issue, the different tables appear at different width, in both the man page and the HTML. That's though they are grid tables with exactly the same width and column width in the Markdown source. I don't know how to fix that.


allefeld:
The second commit includes information from comments in the example defaults file. I omitted a few where the information can be obtained from the corresponding command line option documentation.

jgm:
Great! Let me take a look and see if I can fix the table issue.

jgm:
OK, when used in the context of the website (with the CSS we have there), the tables all look the same width, so I don't think this will be a problem.  Same with PDF output.  The man page, though, doesn't look good; this may require some work on the man writer.

jgm:
The problem in the man output is that the code blocks get turned into .IP (indented paragraph), which looks really bad in this particular context.

jgm:
OK, I've got a solution to the man page issue.
(A bit of hackery in the manfilter.lua.)

allefeld:
Great, that was fast! I'm looking forward to the new documentation to get onto the website...

jgm:
It should be there now...

Issue #7880, opened 2022-02-01
Topdown traversal / alternative interface for custom writers


tarleb:
I was thinking about how to support the new table style in custom writers (#6573), the asymmetry between custom reader and writer, and how writers for non context-free languages sometimes require obscure hacks (e.g. https://groups.google.com/d/msgid/pandoc-discuss/e249e0e0-d71e-4476-9d1f-12f7d3f2acf5%40googlegroups.com). That's why I started to think about an alternative interface that could be setup in parallel to the existing one.

One idea I had was to check for the existence of a global `Writer` function and to use it if present. It would get two arguments, the `Pandoc` document and the writer options, along the lines of custom readers.  The full responsibility of rendering the document would then be on that function.

As a convenience, we could add a helper `pandoc.writer.make` that works similar to a Lua filter, but expects functions to return strings (or `Doc` elements in case we decide to support that via a doclayout module). It would generate a function that converts any AST element to a string.

``` lua
local write_html
write_html = pandoc.writer.make{
  -- expects string return values
  Space = function (_) return ' ' end,

  -- allows to use pandoc.write
  Str = function (s) return pandoc.write(pandoc.Pandoc(s.text), 'html') end,

  -- recurse on element content
  Emph = function (emph)
    return ('<em>%s</em>'):format(write_html(emph.content))
  end,
  Para = function (p)
    -- closing tag omitted intentionally, as suggested by
    -- some HTML style guides.
    return ('<p>%s'):format(write_html(p.content))
  end,

  -- Improved error handling by making this the default implementation for
  -- catch-all functions of a type.
  Block = function (b) error(('no converter for block of type %s')format(b.t)) end
}

function Writer(doc, writer_options)
  local head = ('<head><title>%s</title>\n'):write_html(doc.meta.title)
  local body = '<body>' .. write_html(doc.blocks)
  return '<html>\n' .. head .. body
end
```

A possible downside that I see is that the additional capabilities could blur the distinction between filter and writer, but that could be a good thing in cases where a filter is coupled to a custom writer.

jgm:
One thing people have sometimes requested is the ability to just define certain functions in a custom writer, and have everything else "the same as usual." e.g. override the treatment of Link in HTML while keeping everything else the same.

[...]

PR #7885, opened 2022-02-02
ConTeXt writer: tag paragraphs

tarleb:
Paragraphs are enclosed by `\startparagraph` and `\stopparagraph`
commands. This ensures better tagging results in PDF output.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3579,6 +3579,15 @@ In the `context` output format this enables the use of [Natural Tables
 Natural tables allow more fine-grained global customization but come
 at a performance penalty compared to extreme tables.
 
+#### Extension: `tagging` ####
+
+Enabling this extension with `context` output will produce markup
+suitable for the production of tagged PDFs. This includes
+additional markers for paragraphs and alternative markup for
+emphasized text. The `emphasis-command` template variable is set
+if the extension is enabled. Combine this with the `pdfa` variable
+to generate accessible PDFs.
+
 
 # Pandoc's Markdown
 
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -114,6 +114,9 @@ $endif$
 \setupxtable[foot][]
 \setupxtable[lastrow][bottomframe=on]
 
+$if(emphasis-commands)$
+$emphasis-commands$
+$endif$
 $if(highlighting-commands)$
 $highlighting-commands$
 $endif$
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -129,6 +129,7 @@ data Extension =
     | Ext_subscript           -- ^ Subscript using ~this~ syntax
     | Ext_superscript         -- ^ Superscript using ^this^ syntax
     | Ext_styles              -- ^ Read styles that pandoc doesn't know
+    | Ext_tagging             -- ^ Output optimized for PDF tagging
     | Ext_task_lists          -- ^ Parse certain list items as task list items
     | Ext_table_captions      -- ^ Pandoc-style table captions
     | Ext_tex_math_dollars    -- ^ TeX math between $..$ or $$..$$
@@ -602,6 +603,7 @@ getAllExtensions f = universalExtensions <> getAll f
     [ Ext_smart
     , Ext_raw_tex
     , Ext_ntb
+    , Ext_tagging
     ]
   getAll "textile"         = autoIdExtensions <>
     extensionsFromList
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -42,6 +42,7 @@ import Text.Pandoc.Writers.Shared
 import Text.Printf (printf)
 
 import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Map.Strict as Map
 import qualified Text.Pandoc.Writers.AnnotatedTable as Ann
 
 data WriterState =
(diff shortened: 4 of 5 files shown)


tarleb:
Demonstrating the difference (extracted from the PDF with `pdfinfo -struct-text`):

#### before

```
Div
  "Term Paper TitleBird2022-01-23"
  Sect "section"
    Div
      H (block)
        "Knuth"
    Div
      "Thus, I came to the conclusion that the designer of a new system must not onlybe the implementer and first large–scale user; the designer should also write thefirst user manual.The separation of any of these four components would have hurt TeX significantly. IfI had not participated fully in all these activities, literally hundreds of improvementswould never have been made, because I would never have thought of them orperceived why they were important.But a system cannot be successful if it is too strongly influenced by a single person.Once the initial design is complete and fairly robust, the real test begins as peoplewith many different viewpoints undertake their own experiments."
```

#### after

```
  "Term Paper TitleBird2022-01-23"
  Sect "section"
    Div
      H (block)
        "Knuth"
    Div
      P (block)
        "Thus, I came to the conclusion that the designer of a new system must not onlybe the implementer and first large–scale user; the designer should also write thefirst user manual."
      P (block)
        "The separation of any of these four components would have hurt TeX significantly. IfI had not participated fully in all these activities, literally hundreds of improvementswould never have been made, because I would never have thought of them orperceived why they were important."
      P (block)
        "But a system cannot be successful if it is too strongly influenced by a single person.Once the initial design is complete and fairly robust, the real test begins as peoplewith many different viewpoints undertake their own experiments."
```

I'm just not sure if the additional verbosity is worth it.

CC: @denismaier @klpn

jgm:
Sounds like a win to me, but let's see what the ConTeXt experts say.

denismaier:
In general that's a useful addition, especially when going directly to PDF. Maybe, if you convert to context sources, some might the less verbose alternative. Maybe a new command line option could be useful?

jgm:
I suppose we could add an extension like `tags` or `pagaraph_tags`. But I'm not sure how important this would be.  If someone is using pandoc to generate ConTeXt that will then be hand-edited, and they don't want these things, they could always pipe the output through
```
sed -E -e '/\\(start|stop)paragraph/d'
```
Interested in more feedback on this from ConTeXt users...

jgm:
Actually, the ConTeXt writer has access to variables, so why don't we just activate this feature if the `pdfa` variable is set? Would that be sensitive?

tarleb:
Checking the `pdfa` variable would make sense, IMHO.

It seems that there are number of additional cases where we could improve tagging, e.g., in lists or for emphasized text: the ConTeXt wiki recommends to define `\definehighlight[emph][style={\em}]` and use `\emph{text}` instead of the normal `{\em text}`, as the former produces better tagging. The [Export page](https://wiki.contextgarden.net/Export) in the wiki has a couple of additional examples. The end result looks quite different from "normal" ConTeXt, so yet another extension would be justifiable, too.

jgm:
Checking the `pdfa` variable is easy but slightly unprincipled.  (Variables are supposed to be for template inclusion, so it's always a bit odd when they affect the body too.)  So maybe adding a `tagging` extension would make sense.  Not sure.

tarleb:
I'm tempted to leave things as they are, but to use tagging as motivation for the new writer style and `make_variant` function that you [suggested](https://github.com/jgm/pandoc/issues/7880#issuecomment-1028270298). Tagging-friendly ConTeXt would be a prime usecase for this.

tarleb:
I just found out about the effect of `--section-divs` on ConTeXt output (#2609). I think it might make sense to hide the suggested behavior behind that switch.

jgm:
I think having a separate tagging extension might be more principled. It wouldn't really be obvious why `--section-divs` ALSO puts tags around paragraphs.
Or maybe it wouldn't it be that bad just to do this by default for ConTeXt?
It's the way of the future, presumably, if we want tagging.

tarleb:
I see, that's true. If we merge this, would it make sense to let the `--section-divs` behavior be the default? It seem like that would be the most consistent.

jgm:
> If we merge this, would it make sense to let the --section-divs behavior be the default?

Agreed. I guess that would mean that we're only targeting ConTeXt IV, since older versions don't support the `\start/stopsection`.  But at this point that's probably quite sensible.  I think I'd be in favor of the simplest solution, and this is probably it.

The question @denismaier raised above is about the increased verbosity. I don't know how much of an issue that is for ConTeXt user.

tarleb:
I was informed on the ConTeXt mailing list that using `\startparagraph ... \stopparagraph` leads to problems in some cases, e.g. in list items. The workaround is to use `\bpar ... \epar` instead. I don't understand yet whether it's preferable to always use those commands, or to use them only where `\startparagraph` would lead to unexpected results.

tarleb:
I went ahead with the additional extension: if `tagging` is enabled, all paragraphs are wrapped in `\bpar`/`\epar` commands. Furthermore, we then generate `\definehighlight` commands for all used emphasis types and inject them via the `emphasis-commands` template variable.

Docs haven't been updated yet.

MANUAL.txt
```diff
@@ -3579,6 +3579,15 @@ In the `context` output format this enables the use of [Natural Tables
...
 at a performance penalty compared to extreme tables.
 
+#### Extension: `tagging` ####
+
+Enabling this extension with `context` output will produce markup
+suitable for the production of tagged PDFs. This includes
+additional markers for paragraphs and alternative markup for
+emphasized text. The `emphasis-command` template variable is set
+if the extension is enabled. Combine this with the `pdfa` variable
+to generate accessible PDFs.
```
jgm, line 3589: Note: maybe we should have a section of the manual (near the end I guess) about producing accessible PDFs?

tarleb: Agreed. I realize now that the last sentence should be dropped here, as the `pdfa` variable only affects the metadata, not tagging.

tarleb: My tests indicate that `context` will always produce tagged PDF, but there are small errors in the tags unless the `tagging` extension is enabled. I will clean that up.

PR #7897, opened 2022-02-05
Lua: support single-function custom writers

tarleb:
Check for the global function `Writer` in a custom writer. If it exists, then it is called with the document and writer options as input. Otherwise we fall back to classic custom writers, requiring a global function for each AST element type.

Closes: #7880.

### DIFF
--- a/doc/custom-writers.md
+++ b/doc/custom-writers.md
@@ -15,8 +15,17 @@ install any additional software to do this.
 
 [Lua]: https://www.lua.org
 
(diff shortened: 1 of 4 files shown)


tarleb:
One example is included in the docs. I've dumped a more complex one as a [gist]: it's a full Markdown writer that uses the `pandoc.layout` module to access to DocLayout functionality.

jgm:
I like this.

PR #7900, opened 2022-02-07
Docx: single numbering ID for examples - fixes #7895

### LINKED ISSUE #7895: Interrupted sequences generated in DOCX when using "Numbered example lists"
Numbered example list is a great feature in markdown that enables single consistent numbering thought the document. A case where this is especially crucial is sequential paragraph numbering.

The downside of using "Numbered example lists" is that conversion to DOCX results in restarts of native numbering and unnecessary explicit re-entry of the next value whenever a sequence is interspersed with other content. For example:

```markdown

#	Chapter 1

@a.  Example from paragraph 1 with some content

@a.  Example from paragraph 2 with some content

#	Chapter 2

@a.  Example from paragraph 3 with some content

@a.  Example from paragraph 4 with some content

@a.  Example from paragraph 5 with some content

```

Although the above paragraphs will get numbered sequentially (1., 2., 3., 4., 5.) in DOCX, the used sequence gets restarted when at "Example from paragraph 3...". At that point a new sequence is used and explicit value "3" in entered.

The downside of this sequence restarts is that it is extremely cumbersome to do any numbering modifications in DOCX (e.g. when a document is sent to a colleague that uses only Microsoft Word).

It would therefore be a real time-saver if Pandoc could follow intended use of numbering in DOCX and therefore reuse the same identifier when "Numbered example lists" are used i.e. `<w:numId w:val="same_numeric_value"/>` in `word/document.xml` from DOCX archive would have the same value in this case.

mjfs:


jgm:
Thanks.  But will this work in all cases?  Does it work regardless of the example list delimiters used? e.g.
```
@a. one

(@b) two
```

jgm:
> The first one seems like the logical choice - if you agree with this I could add this to the implementation.

Seems reasonable to me.

> I am wondering how often does it actually happen that inconsistent use of delimiters in the process of numbering content is actually the desired end result.

Not very often, if ever.  My guess is that we'll be okay if we let the first one control the delimiters on the rest.

jgm:
I don't think we actually want that behavior.  If you try with `pandoc -t plain` you'll see that it doesn't work that way in other formats.  Rather, there is a single sequence of numbers, but they get different delimiters depending on what was in the original.

I think we should keep the single sequence.  IF it's convenient to preserve the delimiters, this could be nice, but as noted above it's unlikely to be a problem in practice if we just use the first example list item's delimiter for all the items in the sequence.

mjfs:
It must be noted that the effect of the proposed change is exactly what you describe in the second paragraph - and what I understood you considered reasonable in the post before last:

- a) There is a single uninterrupted sequence in the document
- b) Delimiter of this sequence is set by the first occurrence
- c) The result is an idiomatic document in the target format

Unfortunately, as mentioned, there is currently no way of implementing an uninterrupted numbering in OpenXML/DOCX (or in OpenDocument/ODT for that mater) that would enable both:

1. an editable numbering - allowing insertion of new items (with auto-renumbering), customizing numbering style etc. as well as 
2. enable different delimiters within that same numbering sequence for a specific (sub)group of the sequence

I understand that the behavior that limits the Examples to 1 delimiter throughout the document is not shared between all formats. But neither are other functionalities - e.g. probably nobody ever complained that there is an option to:

- (i) add a table of content that allows refreshing in Microsoft Word or
- (ii) use idiomatic numbering of headings in a DOCX template or 
- (iii) use native numbering of figures and tables or 
- (iv) import or output editorial comments of the content or
- (v) ...

Formats that primarily target plain text editors do no support the functionally listed above due to obvious reasons. One could hardly say that insisting on the lowest common denominator would be beneficial on the other hand. Furthermore, one could even say that there is a reason that most (all?) word processors support changing only the attributes (e.g. delimiter, spacing) of the entire numbering sequence (and not a subset of it). For everything else there is support for free-style named sequences where few limits exists regarding formatting etc.

Lastly, perhaps also worth point out is that Pandoc's manual does not state anywhere that one can mix delimiters within a single Example sequence, nor does it list an example of such behavior. If there were to be a change enforcing a single delimiter for other formats, it would therefore not break any specification. And, as discussed above, it is hard to find a good use case for inconsistent numbering of items within a single sequence.

I understood that you support the implementation in this pull request for this specific format (please correct me if not). Hopefully the arguments above also show that for at least some if not all Pandoc supported format the Examples implementation should follow similar logic.

jgm:
Oh, I'm sorry, I misread your example.  I didn't realize that you were using regular list items, not example list items.

jgm:
Thanks, merged.

Issue #7898, opened 2022-02-06
Cannot create PDF when using a custom writer


tarleb:
Pandoc checks that the target format matches the chosen PDF generator and refuses to produce a PDF if they mismatch. A custom writer will never match, even if the writer produces a variant of the supported format.

Possible solutions:

1. Disable format checking when going through a custom writer. This seems like the easiest solution.
2. Expose `makePDF` to Lua so writers can do the conversion themselves. It would require significantly more effort, but may be the cleaner solution.
3. Leave as is, and document the need for a  separate compilation steps when going through custom writers. Inconvenient, but probably still okay.

Not sure what's best here.

jgm:
Not sure either.  I think I'd be okay with 1 though, and it's pretty easy.

PR #7901, opened 2022-02-07
PDF: allow custom writer as format if engine is explicitly specified

tarleb:
Closes: #7898

### DIFF
--- a/src/Text/Pandoc/App/OutputSettings.hs
+++ b/src/Text/Pandoc/App/OutputSettings.hs
@@ -254,6 +254,10 @@ pdfWriterAndProg mWriter mEngine =
       go Nothing Nothing       = Right ("latex", "pdflatex")
       go (Just writer) Nothing = (writer,) <$> engineForWriter writer
       go Nothing (Just engine) = (,engine) <$> writerForEngine (takeBaseName engine)
+      go (Just writer) (Just engine) | isCustomWriter writer =
+           -- custom writers can produce any format, so assume the user knows
+           -- what they are doing.
+           Right (writer, engine)
       go (Just writer) (Just engine) =
            case find (== (baseWriterName writer, takeBaseName engine)) engines of
                 Just _  -> Right (writer, engine)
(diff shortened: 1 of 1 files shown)


tarleb:
The only remaining inconvenience is that a template has to be specified on the command line, e.g.

    pandoc -t pdfa.lua --pdf-engine=context --template=default.context input.md

jgm:
I'm not sure what you mean by "allow an empty template."

Actually, I don't think it's so bad if they have to specify the template.
I wonder, though...would it work if you put `default.pdfa.lua` in the templates subdirectory of userdata, or made this a symlink to default.context?

tarleb:
> I'm not sure what you mean by "allow an empty template."

I meant to set the `writerTemplate` to `Nothing`, despite `--standalone` being implied.

> I wonder, though...would it work if you put `default.pdfa.lua` in the templates subdirectory of userdata, or made this a symlink to default.context?

Yes, that works. However, it's rather fagile:

1. The writer should be in the current working directory, as pandoc will search the template relative to the `$DATADIR/template` directory.
2. the writer name should not contain a `-` or `+`, as everything after these is interpreted as extension. I.e., if the writer is named `pdfa-3a.lua`, then the template must be named `default.pdfa`.

I should probably open a new issue for these things.

jgm:
I'm still not clear on the precise change you have in mind ; maybe clearer if you just add a commit that I can look at!

tarleb:
Right, sorry. I've added another commit.

tarleb:
FWIW, this is the full code for the pdfa.lua writer that's fueling most of my motivation for this. The remaining motivation comes from the crossref-xml.lua writer, but that writer is currently still a collection of filters and hacks.

``` lua
local add_tags = {
  traverse = 'topdown',
  Para = function (p)
    return pandoc.Para(
      {pandoc.RawInline('context', '\\startparagraph\n')} ..
      p.content ..
      {pandoc.RawInline('context', '\n\\stopparagraph')}
    )
  end,
  Note = function (n)
    -- "fixing" paragraphs in footnotes leads to layout issues.
    return n, false
  end
}

function Writer (doc, opts)
  local default = pandoc.template.default('context')
  opts.template = pandoc.template.compile(default)
  opts.variables.pdfa ='3a'
  return pandoc.write(doc:walk(add_tags), 'context', opts)
end
```

tarleb:
The downside of this change is that a custom writer cannot know for certain if pandoc was called with `--standalone` or not.

tarleb:
I'm removing the second commit again, the longer I thought about it the less I liked it.

There are some alternative ideas spooking around in my mind, I'll try to forge a separate PR.

Issue #5202, opened 2019-01-07
DokuWiki reader: parse table alignment


link2xt:
DokuWiki format allows to [align table cells with whitespace](https://www.dokuwiki.org/wiki:syntax#tables). Pandoc has no way to represent individual cell alignment, but it may still be useful to align the whole column according to header alignment, if header is present.

This would allow Pandoc to correctly parse such tables:
```
^          right^   center   ^left            ^
|  right-aligned|  centered  |left-aligned    |
|  right-aligned|  centered  |left-aligned    |
```

damon-sava-stanley:
Howdy, I've taken a stab at this (in the linked pull request), though it is a bit of a bodge.

PR #7908, opened 2022-02-11
Add DokuWiki table alignment for #5202

damon-sava-stanley:
Within each cell, determine the cell alignment as per
https://www.dokuwiki.org/wiki:syntax#tables. The current approach, as
per the issue treats the first row's alignment as determining
that of the entire column. Given this, it wastes some work in
determining an alignment for every cell.

### DIFF
--- a/src/Text/Pandoc/Readers/DokuWiki.hs
+++ b/src/Text/Pandoc/Readers/DokuWiki.hs
@@ -19,8 +19,8 @@ import Control.Monad
 import Control.Monad.Except (throwError)
 import Data.Char (isAlphaNum, isDigit)
 import qualified Data.Foldable as F
-import Data.List (transpose)
 import Data.Maybe (fromMaybe, catMaybes)
+import Data.Bifunctor (second)
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Text.Pandoc.Builder as B
@@ -96,7 +96,11 @@ codeTag f tag = try $ f
 -- | Parse any inline element but softbreak.
 inline' :: PandocMonad m => DWParser m B.Inlines
 inline' = whitespace
-      <|> br
+      <|> inline''
+
+-- | Parse any inline element but whitespace.
+inline'' :: PandocMonad m => DWParser m B.Inlines
+inline'' = br
       <|> bold
       <|> italic
       <|> underlined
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Readers/DokuWiki.hs
```diff
@@ -471,19 +479,24 @@ table = do
   let (headerRow, body) = if firstSeparator == '^'
                             then (head rows, tail rows)
                             else ([], rows)
-  let attrs = (AlignDefault, ColWidthDefault) <$ transpose rows
+  -- Since Pandoc only has column level alignment, we have to make an arbitrary
+  -- choice of how to reconcile potentially different alignments in the row.
+  -- Here we end up assuming that the alignment of the header / first row is
+  -- what the user wants to apply to the whole thing.
+  let attrs =  map (\(a, _) -> (a, ColWidthDefault)) . head $ rows
```
jgm, line 486: `head` makes me nervous because it's partial, even though here `many1` guarantees a nonempty list.
Could we use `headMay` from Safe, or a pattern match, or some other non-partial function?

src/Text/Pandoc/Readers/DokuWiki.hs
```diff
@@ -492,11 +505,23 @@ tableRowEnd = try $ tableCellSeparator <* manyTill spaceChar eol
...
   where
-    normalCell = char '|' *> manyTill inline' (lookAhead tableCellSeparator)
-    headerCell = char '^' *> manyTill inline' (lookAhead tableCellSeparator)
+    cellContent = do
+      -- https://www.dokuwiki.org/wiki:syntax#tables
+      -- DokuWiki represents the alignment of cells with two spaces padding.
+      tableCellSeparator
+      cellInline <- manyTill inlineUnconsolidatedWhitespace (lookAhead tableCellSeparator)
+      let left  = (==2) . length . filter (== B.space) . take 2 $ cellInline
+      let right = (==2) . length . filter (== B.space) . take 2 . reverse $ cellInline
```
jgm, line 517: Maybe more straightforward:  `[B.space, B.space] `isPrefixOf` cellInline` and similarly for right with `isSuffixOf`?

jgm:
Looks good, thanks!  I made a couple minor comments in the code.

damon-sava-stanley:
Thanks for the suggestions! I've integrated them. (Apologies for the git churn above, it's been a minute on my part.)

jgm:
Excellent, thanks for this!

Issue #7907, opened 2022-02-10
Put id attributes on TOC entries so they can be linked to


MuntashirAkon:
**Describe your proposed improvement and the problem it solves.**
Pandoc has an option to generate table of contents. It also nicely links the section titles making it a bit easy to navigate around the generated HTML file. However, there is no way of going back to the table of content from the section that the reader has navigated to.

Here's a link to a documentation generated by `make4ht` as an example: https://muntashirakon.github.io/AppManager/en/

As you can see in the example,
1. Section numbers do not have hyperlinks, only the titles do, thus, making it easier to distinguish the titles from the section number
2. It is possible go back to the table of contents via the hyperlink in the section title.

**Describe alternatives you've considered.**
I can continue to use `make4ht` but it has other issues. Other option is to write a script to emulate that myself.

jgm:
I think this is best handled with customization on top of what pandoc does.
For example, on <https://pandoc.org> we use a small Lua script to insert the anchor links that appear on hover:

```lua
-- Adds anchor links to headings with IDs.
function Header (h)
  if h.identifier ~= '' then
    -- an empty link to this header
    local anchor_link = pandoc.Link(
      {},                  -- content
      '#' .. h.identifier, -- href
      '',                  -- title
      {class = 'anchor', ['aria-hidden'] = 'true'} -- attributes
    )
    h.content:insert(1, anchor_link)
    return h
  end
end
```

You could modify this to make it link to the TOC.  Everyone is going to have different preferences for this kind of thing, and making the section headings links by default is not what everyone wants.  (Actually, pandoc used to do this long ago but we stopped.)

tarleb:
I think these links will come in handy for this: 

- <https://stackoverflow.com/q/69891283/>
- <https://groups.google.com/g/pandoc-discuss/c/1xIAG684yrM/m/SANzToYFFwAJ>

MuntashirAkon:
> Hence the second link, which shows how you can create your own toc with a Lua filter.

Ah, thanks!

jgm:
Maybe we should modify the HTML writer so that every TOC entry gets its own id attribute, generated in some predictable way based on the id it links to?  That would facilitate this.

jgm:
I'll reopen this issue and rebrand it accordingly.

damon-sava-stanley:
[...]

PR #7913, opened 2022-02-11
Put id attributes on TOC entries #7907

damon-sava-stanley:
Naming scheme of id is "toc-" + id of linked to header/section.
In Shared, will effect HTML, Markdown, Powerpoint, and RTF.


jgm:
This looks good.  I'm still not absolutely sure we want to make this change, though.  I'll comment on the orginial issue.

jgm:
On second thought, I think I'm persuaded that this would be a good change. The problem is that you can modify the link targets with a filter, but the filter doesn't give you any way to add ids to the toc unless they're added by the writer.  That seems a compelling enough reason to add the ids.

PR #7929, opened 2022-02-19
Relax upper bound for hslua, allow hslua-2.2.

tarleb:
This change allows to use pandoc with Lua 5.4.


tarleb:
Lua 5.4 would be used by default after this is merged. Packagers may still include Lua 5.3 instead by building pandoc with `--constraint='hslua <2.2'`.

tarleb:
Adding more info about the Lua 5.3 / Lua 5.4 switch: This should be there unnoticeable to script authors, [incompatible changes](https://www.lua.org/manual/5.4/manual.html#8.1) are mostly negligible.

An argument against merging this: allowing both 5.3 and 5.4 could lead to confusion, as the pandoc version would not allow to deduce the included Lua version. This is already the case, but to a lesser extend: e.g., Lua 5.3.3, shipped in Ubuntu Docker images, behaves slightly different in rare edge cases than the default Lua 5.3.6.

Most major distributions already ship Lua 5.4 packages, with the exception of Ubuntu LTS. However, the next LTS release will include Lua 5.4.4, same as the version shipping with hslua. So if we are ok with breaking Ubuntu LTS builds for 2 months, then we could just require hslua-2.2 instead of allowing both hslua-2.1 and hslua-2.2.

jgm:
If the differences btw 5.3 and 5.4 are minor, then I think this is okay. I'd rather avoid breaking Ubuntu LTS builds.  Maybe add an issue as a reminder to tighten this constraint in two months?

PR #7933, opened 2022-02-21
Org reader fix: more flexible LaTeX environments

lucasvreis:
Looking at the definition of `org-element-latex-environment-parser`, one sees that Org is more flexible than Pandoc about LaTeX environments. In fact, it parses every char just after `\begin{xxx}` until `\end{xxx}` as content for the environment, so all the following examples are valid environments that did not work before:
```org
\begin{equation} e = mc^2 \end{equations}
```
```org
\begin{tikzcd}[ampersand replacement=\&]
	A \& B \\
	C \& D
	\arrow[from=1-1, to=1-2]
	\arrow["f", from=2-1, to=2-2]
\end{tikzcd}
```
A question: is there any reason why the verbatim case was made to parse a line like `f(x) = x^2` inside environments as `Str "f(x) = x^2"` instead of `[Str "f(x)", Space, Str "=", Space, Str "x^2"]`? If I understood things properly the latter seems more correct to me in respect to what Org does, so this was changed too.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/BlockStarts.hs
+++ b/src/Text/Pandoc/Readers/Org/BlockStarts.hs
@@ -58,7 +58,6 @@ latexEnvStart = try $
   skipSpaces *> string "\\begin{"
              *> latexEnvName
              <* string "}"
-             <* blankline
  where
    latexEnvName :: Monad m => OrgParser m Text
    latexEnvName = try $ mappend <$> many1Char alphaNum <*> option "" (textStr "*")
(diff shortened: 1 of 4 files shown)


tarleb:
Thanks, I think you are completely right with everything you changed!

Minor request: we semi-automatically generate the changelog from the commit messages. Could you make sure that it reads well in this context?

tarleb (approved):

lucasvreis:
@tarleb did I did it right? I'm not sure if commit message markup should be in markdown 😅

PR #7951, opened 2022-03-08
add reference to install via winget

Sa1Gur:
I found out that pandoc can be installed via winget


jgm:
Thanks!

Sa1Gur:
You are welcome. Thank you for perfect converter!

Issue #7295, opened 2021-05-16
default.latex. Move metadata variables - title, author, date, etc - above $header-includes$?


JohnLukeBentley:
I suggest/request that in default.latex the basic metadata variables - title, author, date, etc. - be moved above `$header-includes$`. (I generally use the practice of putting these soon after \documentclass, mimicking what's generally done when working with html).

The motivation is to be able to reference those metadata variables in one's included header code. In my case I'm wanting to put some of these in custom margins.

If the suggestion seems right I suppose all of the following from default.latex would warrant being moved up.

```
$if(title)$
\title{$title$$if(thanks)$\thanks{$thanks$}$endif$}
$endif$
$if(subtitle)$
$if(beamer)$
$else$
\usepackage{etoolbox}
\makeatletter
\providecommand{\subtitle}[1]{% add subtitle to \maketitle
  \apptocmd{\@title}{\par {\large #1 \par}}{}{}
}
\makeatother
$endif$
\subtitle{$subtitle$}
$endif$
\author{$for(author)$$author$$sep$ \and $endfor$}
\date{$date$}
```

jgm:
I think that would be fine.

JohnLukeBentley:
Your commit tests fine at my end. Thanks very much!

jgm:
I should have known better than to mess with this!  See #7362.  Perhaps the best approach will be to move the metadata bits immediately before header-includes.  However, I do like having this material in a conspicuous location, either at the beginning or the end of the preamble, so I'm not sure.

gitphilip:
Hi!

For me this update seems to create an issue if I want to include the fontawesome package in the metadata yaml.
In particular, I load fontawesome in the header (I think there is no other way, or?) and want to use the mail icon in the author field already (when using beamer). Previously this worked, but now it seems that moving the metadata before the header causes this issue. More precisely, this doesn't work:

```yaml
author: | 
 | \textbf{My name}
 | \scriptsize \faEnvelope{} \href{mailto:mail@mail.com}{mail@mail.com} \normalsize
```
I load fontawesome via an external header file (I already checked `header-includes` does also not work):

`--include-in-header=../resources/slides_header.tex`

So is there a workaround for this issue, or will it persist and require me to manually change the template?

jgm:
This is like the problem in #7362, but (I think) without a workaround.
@JohnLukeBentley can you comment?
We have a tension here:

[...]

Issue #5811, opened 2019-10-11
incompatibilities with hyperref latex package


brainchild0:
The LaTeX template contains a large number of `\usepackage` and other declarations, followed by an interpolation of `header-includes`, in which users  will load further needed packages.

One of the packages loaded by the template is *hyperref*.  

**Unfortunately, some packages will not load after hyperref is loaded.**  

Hyperref is apparently a fussy package, as documented in [ the manual][man] and expounded in [ a blog article][blog]:

> The hyperref package documentation says: “Make sure it comes last of your loaded packages”. The reason is that it redefines many LaTeX commands. It’s a rule of thumb that helps to avoid errors.

**Problems can therefore be avoided by moving the use of    hyprref  to follow after the user header (i.e. `$header-includes$`).**  

In fact, the template already does exactly so for select packages (e.g. *Load polyglossia as late as possible*):

``` latex
$for(header-includes)$
$header-includes$
$endfor$
$if(lang)$
\ifxetex
  % Load polyglossia as late as possible: uses bidi with RTL langages (e.g. Hebrew, Arabic)
  \usepackage{polyglossia}
  \setmainlanguage[$polyglossia-lang.options$]{$polyglossia-lang.name$}
$for(polyglossia-otherlangs)$
  \setotherlanguage[$polyglossia-otherlangs.options$]{$polyglossia-otherlangs.name$}
$endfor$
\else
  \usepackage[shorthands=off,$for(babel-otherlangs)$$babel-otherlangs$,$endfor$main=$babel-lang$]{babel}
$if(babel-newcommands)$
  $babel-newcommands$
$endif$
\fi
$endif$
$if(dir)$
\ifxetex
  % Load bidi as late as possible as it modifies e.g. graphicx
  \usepackage{bidi}
\fi
\ifnum 0\ifxetex 1\fi\ifluatex 1\fi=0 % if pdftex
  \TeXXeTstate=1
  \newcommand{\RL}[1]{\beginR #1\endR}
  \newcommand{\LR}[1]{\beginL #1\endL}
  \newenvironment{RTL}{\beginR}{\endR}
  \newenvironment{LTR}{\beginL}{\endL}
\fi
$endif$
$if(natbib)$
\usepackage[$natbiboptions$]{natbib}
\bibliographystyle{$if(biblio-style)$$biblio-style$$else$plainnat$endif$}
$endif$
$if(biblatex)$
[...]

PR #7960, opened 2022-03-11
Latex template hyperref last

chroetz:
In the default.latex template, the hyperref package is moved down to be loaded last.

As mentioned in the [manual](https://ctan.net/macros/latex/contrib/hyperref/doc/hyperref-doc.html#x1-50003) of the LaTeX package [hyperref](https://ctan.org/pkg/hyperref), this package should be loaded last. 
This rule is applied to the default.latex template.

This fixes a bug: Links in the table of contents of PDF output referred to one page before the page where the respective section begins, if that section begins at the very top of the page.


jgm:
Thanks.  See #5811 for a discussion of whether hyperref should go before or after header-includes.
There are some reasons to put it before, so that hyperref commands can be used in header-includes.
Since we also want header-includes to come before title, author, etc. (see #7295 for discussion), this means also before `\title`.

It seems to me that it should be safe to move header-includes forward, so they come right before `\title`, etc., and to move hyperref so it comes right before header-includes.

More radically, we could consider moving hyperref after header-includes but before `\title`.  This would have advantages if people load packages in header-includes that need to be modified by hyperref.  The drawback mentioned in #5811 is that users couldn't use hyperref commands in their header-includes. But, as noted in the comment there, it seems that an author could simply put `\usepackage{hyperref}` in header-includes if it's needed there; the redundant `\usepackage` that comes later wouldn't cause problems.

Comments welcome!

chroetz:
Thanks for the comment. Sorry that I didn't see that this issue was already discussed.

My original problem of links in the toc that point to the wrong page is caused by using
```
\usepackage{titlesec}
\titleformat{...}
```
in *header-includes*. It can be solved by putting `\usepackage{hyperref}` only after that.

So, I think it would be best to put `\usepackage{hyperref}` after *header-includes* and let users use `\usepackage{hyperref}` in the *header-includes* if they need it there. As was mentioned and according to [this](https://tex.stackexchange.com/questions/79110/loading-a-package-more-than-once), only the first `\usepackage{hyperref}` is loading the package.

I close the PR as this was obviously a bit premature and I don't think I have the necessary insight to make sure that the change is robust.

jgm:
@chroetz I think that's a good suggestion.  Are you planning to submit a new PR?
Alternatively, I could just make the change.

chroetz:
> Are you planning to submit a new PR?
> Alternatively, I could just make the change.

I'd prefer if you make the change. 
Thanks!!

jgm:
Done.

memeplex:
I'm not entirely sure but I think this change may be causing some trouble. I'd like to configure hyperref to show underlined links, but my `\hypersetup` fragment specified in `header-includes` comes before this fragment introduced by pandoc at the end:

```
\hypersetup{
  hidelinks,
  pdfcreator={LaTeX via pandoc}}
```

For now I think `boxlinks: true` is good enough for my needs, so this is just to notice that `\hypersetup` is perhaps being introduced too late in the preamble (maybe it fixes worse issues that way, I don't know).

jgm:
See above https://github.com/jgm/pandoc/pull/7960?notification_referrer_id=NT_kwDNC-SvMzI1MjU5NzY4ODozMDQ0#issuecomment-1065334769
which I believe includes a workaround for you.

Issue #1865, opened 2015-01-09
writer for AZW3 (Kindle native format)


ousia:
@jgm, I wonder whether a writer for the [Kindle native format](https://en.wikipedia.org/wiki/Amazon_Kindle#Proprietary_formats_.28AZW.2C_KF8.29) could be added to issue #1852.

jgm:
I hadn't heard of that.  Nice idea.  As far as I can see, it's not too
different from EPUB.  For reference:
<http://kindlegen.s3.amazonaws.com/AmazonKindlePublishingGuidelines.pdf>

I don't see myself getting to this in the near future, but someone else
might want to take on the project, starting with a refactoring of the
EPUB writer code to make it more modular, so we can re-use common bits.

cmal:
looking forward to this

lachrimae:
Hey, I intend to start work on this and I haven't contributed before. What documentation would be key for understanding the interface the writer will have to implement? What are the most important modules it will interact with? I'm going to model the AZW3 writer off of the EPUB writer, so some things will be obvious, but any advice would be appreciated as well.

mb21:
The EPUB writer is somewhat special in that it basically calls the HTML writer and does some pre- and post-processing and then zips everything. If EPUB and AZW3 are very, very similar, it might even make sense to implement azw3 as an [extension](http://pandoc.org/MANUAL.html#extensions) to the EPUB writer.

jonay2000:
I see this hasn't had much progress recently. If you don't mind I'll give it a go :)

PR #7991, opened 2022-03-28
Refactor of Writers/EPUB.hs

jonay2000:
After a comment by [JGM](https://github.com/jgm/pandoc/issues/1865#issuecomment-407898949) we refactored EPUB.hs to make it more readable. This includes adding more comments, splitting up the main conversion function (which was 500+ lines long). We think that readability is improved like this, but more refactoring may be necessary before the kindle writer will be worked on. 

If you disagree with some of the changes, let us know!

### DIFF
--- a/src/Text/Pandoc/Writers/EPUB.hs
+++ b/src/Text/Pandoc/Writers/EPUB.hs
@@ -254,7 +254,7 @@ addMetadataFromXML e@(Element (QName name _ (Just "dc")) attrs _ _) md
   | name == "coverage" = md { epubCoverage = Just $ strContent e }
   | name == "rights" = md { epubRights = Just $ strContent e }
   | name == "belongs-to-collection" = md { epubBelongsToCollection = Just $ strContent e }
-  | name == "group-position" = md { epubGroupPosition = Just $ strContent e }  
+  | name == "group-position" = md { epubGroupPosition = Just $ strContent e }
   | otherwise = md
   where getAttr n = lookupAttr (opfName n) attrs
 addMetadataFromXML e@(Element (QName "meta" _ _) attrs _ _) md =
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Writers/EPUB.hs
```diff
@@ -812,7 +693,245 @@ pandocToEPUB version opts doc = do
...
+                                   "foreword", "preface", "frontispiece",
+                                   "seriespage", "titlepage"]
+               backMatterTypes = ["appendix", "colophon", "bibliography",
+                                  "index"]
+
+  zipWithM chapToEntry [1..] chapters
+
+-- | Splits the blocks into chapters and creates a corresponding reftable
+createChaptersAndReftable :: WriterOptions -> [Block] -> ([Chapter], [(Text, Text)])
+createChaptersAndReftable opts secs = do
```
jgm, line 838: Why is this written with `do` notation? There is no monad.  I think you could just write
```
createChaptersAndReftable opts secs = (chapters, reftable)
 where
  chapterHeaderLevel = ...
  etc.
```

jgm:
These look like good improvements in clarity (save the one thing I flagged in a comment).
I think it's probably possible to improve even further -- e.g. the functions with many positional arguments aren't so clear and it might be worth considering introducing records to tame the complexity (as I did already with EPUBMetadata).  But (with the one small change noted above) I'd be happy to merge this as it stands, assuming CI passes.

src/Text/Pandoc/Writers/EPUB.hs
ivardb: Indeed, my mistake, it should be fixed now, let me know if you want us to combine the commits into one.

jgm:
Thanks! I can squash the commits through the interface here.

PR #8002, opened 2022-03-30
Lua: allow to pass Sources to pandoc.read

tarleb:
Sources, the data type passed to the `Reader` function in custom
readers, are now accepted as input to `pandoc.read`.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -3414,7 +3414,7 @@ retrieved from the other parsed input files.
 Parameters:
 
 `markup`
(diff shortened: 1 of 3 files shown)


tarleb:
My main motivation here is convenience, but it's also nice that `--file-scope` can be made to work with custom readers this way. It's also a requirement for the idea discussed on [pandoc-discuss](https://groups.google.com/g/pandoc-discuss/c/tZgBVpfTYaQ/m/29TopL9fAgAJ), but I don't plan to implement this now.

PR #8014, opened 2022-04-10
Add HTML to formats including abstract

castedo:
Default HTML template is also including metadata variable `abstract`


jgm:
thanks!

PR #8036, opened 2022-04-23
Fix <200c> character rendering

hossein-lap:
Rename \textormath to \TextOrMath for rendering Zero Width Non-Joiner
Maybe a typo?!


tarleb:
Thanks, good change. It's not a typo, but apparently `\textormath` is defined by `babel`, while `\TextorMath` is, from what I found, in the LaTeX core and the preferred command nowadays.

Issue #813, opened 2013-04-04
New Feature: internal links to tables and figures and headers


GeraldLoeffler:
It's currently possible to include internal links to sections. I'd like to propose a similar feature for links to figures/images and tables.

It may make sense to provide this feature only if the figure/image or table that is being linked to has a caption. In that case Pandoc can today automatically generate a number for the figure or table and include it in the caption, e.g. "Figure 15".

At the most basic, the text of the link would be provided by the user, as is currently the case for links to sections.

Of course it would be very convenient if the automatically generated number for the figure or table would also be used for the text of the link, e.g. "as can be seen in Figure 15, blah", where "Figure 15" would be the internal link whose text is auto-generated from the figure it points to.

scaramouche1:
The lack of internal cross-references is the major stumbling block for Pandoc's "world domination" in academia.  This thread has made great progress at discussing the problem and providing solutions for cross-referencing figures.

But the problem of cross references is more general, and thus it makes sense to try to solve it in a more general way.  For instance, mathematical documents cross-reference theorems and equations, social science cross reference hypotheses, philosophy cross references examples, and most disciplines cross reference section numbers.  For Pandoc to be really helpful to all these disciplines, a general solution must be devised.

I propose a slight modification to the # and @ notations to allow for all types of cross-references.  Here's the informal specification:

Anchors (#) and references (@) can optionally include a type descriptor.  For instance, #fig:cat, #eq:force, #sec:intro, etc.  The descriptor is what's in between '#' and ':'.  Example uses:

**Referencing a figure**

    ![This is a cat](cat.png) {#fig:cat}
    
    As seen in Figure @fig:cat.

**Referencing an equation**

    $$F = ma$$ {#eq:force}
    
    As seen in Equation @eq:force.

**Referencing a hypothesis**

    Hypothesis {#hyp:temp}: Temperature increases in summer.
    
    As mentioned in Hypothesis @hyp:temp.

This notation also serves to cite theorems, proofs, etc. (@thm:, @proof:).

**Referencing a section number**

    # Introduction {#sec:intro}
    
    As mentioned in Section @sec:intro.

This is in addition to the implicit referencing of headers already in place.

**How to deal with possible clash with bibliographic references**

There proposed notation has is small clash with the notation used to cite bibliographic references.  I propose that references that include a ":" are first searched inside the document, and only if there's no internal match they are searched in the bibliographic system.  Alternatively, ':' could be forbidden for bibliographic references.

[...]

PR #8049, opened 2022-04-29
Org reader: allow attrs for Org tables

### LINKED ISSUE #6441: Org reader support for ATTR_HTML tables
Support for `#+ATTR_HTML` lines for figures was added in https://github.com/jgm/pandoc/pull/2927 

This issue ticket is a feature request to extend the  `#+ATTR_HTML` support for HTML tables.

```
#+ATTR_HTML: :class table-dark
| h1 | h2 |
|----|----|
| c1 | c2 |
```

I'd expect to see the `table-dark` class in the HTML generated by the `hakyll` tool, which uses the Pandoc reader for org files:

    <table class="table-dark">

Instead I see:

    <table>

@tarleb suggests in https://github.com/jgm/pandoc/pull/2927#issuecomment-640238922 that Table attributes are not supported yet, but could be. This ticket is to track an implementation of support for e.g.

    #+ATTR_HTML: :class table-dark

leungbk:
Fixes #6441.


src/Text/Pandoc/Readers/Org/Blocks.hs
```diff
@@ -634,25 +634,29 @@ orgTable = try $ do
 
   let caption = fromMaybe mempty (blockAttrCaption blockAttrs)
   let orgTbl = normalizeTable <$> rowsToTable rows
+  let attr = case blockAttrKeyValues blockAttrs of
+        [] -> nullAttr
+        _ -> (fromMaybe mempty (blockAttrLabel blockAttrs), [], blockAttrKeyValues blockAttrs)
```
leungbk, line 639: Something I didn't consider when I first wrote this: if `blockAttrKeyValues blockAttrs` is `[]` but the `blockAttrs` has a non-empty `blockAttrLabel` property, should we have `attr` as `(fromMaybe mempty (blockAttrLabel blockAttrs), [], [])`?

leungbk: I think the answer is yes.

leungbk: Actually, I guess it's not so simple since having no `blockAttrName` and a non-null `blockAttrLabel` can produce a duplicate `id` in the HTML because of the div-wrapping performed in this function. I'll need to look at this again tomorrow evening.

leungbk:
> Thanks for this, looks good! Can we generalize and reuse the code in `figure` to get the attributes, or are there subtle differences that would make this difficult?

Because of the need to wrap tables in a div to ensure compatibility with pandoc-citeproc, I've decided to not use the name/label as the attr ID; the name/label will be used as an HTML ID in that enclosing div.

In contrast, given how figures are represented, it makes sense to pass non-empty name/label as the attr ID there. Until https://github.com/jgm/pandoc/issues/813 is resolved, we will have to continue to treat tables and figures differently in this manner. For this reason, I don't think it is presently worth it to reuse the code in `figure`.

tarleb:
I assume you mean pandoc-crossref, not pandoc-citeproc. Does that mean pandoc-crossref expects the wrapping Div and won't work if all attributes are added to the table? That's surprising to me.

To be honest, I had forgotten about the wrapper Div. Now I wonder whether we should go the next step and get rid of the wrapper. Compat with pandoc-crossref would be a sufficient reason to keep it for now, so I just to make sure: are you positive that folding the Attr into the table will cause compatibility problems?

leungbk:
> are you positive that folding the Attr into the table will cause compatibility problems?

I tried it just now after also removing the Div and found no issues.

tarleb:
Great, thank you! Nice and clean :+1:

PR #8065, opened 2022-05-11
Improved reading <xref> elements in DocBook files

### LINKED ISSUE #8064: xref usage with docx ends in "figure_title"
when converting docbook to docx, the references to an image end in "figure_title"
picture

[out.txt](https://github.com/jgm/pandoc/files/8670305/out.txt)

renders to

[test.docx](https://github.com/jgm/pandoc/files/8670325/test.docx)

with rendering command:

pandoc --from docbook --to docx out.txt -o "test.docx" --toc --toc-depth=3    && "test.docx"

Although it has correct picture caption titles...

frerich:
`<xref>` elements can also be used to link to `<figure>` elements. Alas, the DocBook reader was not aware of this and thus generated a link text which just said 'figure_title'.

https://tdg.docbook.org/tdg/4.5/figure.html explains that `<figure>` elements can contain `<title>` elements, so let's try to use that if it is available.

Technically, this logic could be a _lot_ more sophisticated. The [docbook-xsl project's XSLT stylesheet for handling `<xref>` elements](https://github.com/docbook/xslt10-stylesheets/blob/6ff683948c5a85949e4b7661f302e8b5f12f7bf2/xsl/html/xref.xsl) is a whopping 1300+ lines of code, and a lot of that is dedicated to a comprehensive heuristic which tries to come up with a sensible link text depending on the target element being referenced. However, it took seven years for someone to point out that Pandoc's handling of `<xref>` elements is incomplete, so maybe we don't need to go all the way just yet.

Closes #8064 .

### DIFF
--- a/src/Text/Pandoc/Readers/DocBook.hs
+++ b/src/Text/Pandoc/Readers/DocBook.hs
@@ -1342,6 +1342,7 @@ parseInline (Elem e) =
                   "sect5"        -> descendantContent "title" el
                   "cmdsynopsis"  -> descendantContent "command" el
                   "funcsynopsis" -> descendantContent "function" el
+                  "figure"       -> descendantContent "title" el
                   _              -> qName (elName el) <> "_title"
           where
             xrefLabel = attrValue "xreflabel" el
--- a/test/docbook-xref.docbook
+++ b/test/docbook-xref.docbook
@@ -24,6 +24,9 @@ cross-reference text: <xref linkend="ch02"/>.
 <listitem><para>A link to an
 <sgmltag>funcsynopsis</sgmltag> element: <xref linkend="func01"/>.
 </para></listitem>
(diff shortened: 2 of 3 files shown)


jgm:
Thanks very much!

frerich:
> Am I since 7 years the only one trying to survive in corporate environment writing in asciidoctor(-pdf), where .docx is needed as standard documents and want to get there via docbook and pandoc?

I don't know, maybe you are! :-) However, maybe you're also just the first person to exercise the `<xref>` element such that it points to `<figure>` elements.

gxurma:
If you would be so kind and build a "nightly" windows version for me of this latest state, so I would be able to test it further. ;)

jgm:
Nightlies are built automatically.  See the actions tab.

PR #8073, opened 2022-05-14
Allow environment variable interpolation in highlight-style in defaults files

tomtomjhj:
This patch makes the following `--defaults` configuration work as expected. 
```yaml
highlight-style: ${.}/my.theme
```


jgm:
Thanks!

PR #8096, opened 2022-06-01
ConTeXt writer: unify link handling

tarleb:
Autolinks, i.e. links with content that's the same as the linked URL,
are now marked with the `\url` command. All other links, both internal
and external, are created with the `\goto` command, leading to shorter,
slightly more idiomatic code. As before, autolinks can still be styled
via `\setupurl`, other links via `\setupinteraction`.

### DIFF
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns        #-}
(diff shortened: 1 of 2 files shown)


tarleb:
I'm not a ConTeXt expert, but this I believe the change leads to nicer code without any other behavior changes.

jgm:
I'm not a ConTeXt user, so I'm reluctant to sign off on this without further input.  @denismaier does this look like a good change?  It sure looks simpler to me, but perhaps there are implications I'm not aware of?

denismaier:
I'd say this is a good change, but one possible improvement: Wouldn't it make sense to wrap the url in `\hyphenatedurl`, see https://wiki.contextgarden.net/URL
WDYT?

tarleb:
I did that initially, but found that the `\url` command already uses `\hyphenatedurl` internally. See the source: https://source.contextgarden.net/strc-ref.mkvi#l1822

denismaier:
Ah, good to know. Thanks for pointing me to that! I did a quick test:

```tex
\setuppapersize[A6]
\setupinteraction[state=start]

\starttext

asdfasdfasdf \useURL[url1][http://example.com/asdf/asdf/asdf/asdf]\from[url1]

asdfasdfasdf \goto{\hyphenatedurl{http://example.com/asdf/asdf/asdf/asdf}}[url(http://example.com/asdf/asdf/asdf/asdf)]

\stoptext
```

Both variants give the same results. 

Question is if that isn't still a bit too verbose:

Why not something like this?

```tex
\setuppapersize[A6]
\setupinteraction[state=start]
\def\mySimpleLinkCmd#1{\goto{\hyphenatedurl{#1}}[url(#1)]}
\def\mySecondSimpleLinkCmd#1{\useURL[url123][#1]\from[url123]}

\starttext
asdfasdfasdf \mySimpleLinkCmd{http://example.com/asdf/asdf/asdf/asdf}

asdfasdfasdf \mySecondSimpleLinkCmd{http://example.com/asdf/asdf/asdf/asdf}

asdfasdfasdf \useURL[url1][http://example.com/asdf/asdf/asdf/asdf]\from[url1]

asdfasdfasdf \goto{\hyphenatedurl{http://example.com/asdf/asdf/asdf/asdf}}[url(http://example.com/asdf/asdf/asdf/asdf)]

\stoptext
```

tarleb:
The difference between the variants `\useURL[url1][http://example.com/]\from[url1]` and `\goto{\hyphenatedurl{http://example.com/}}[url(http://example.com/)]` is that only the former is affected by `\setupurl` while the latter isn't. I didn't want to break that feature, hence the `\url` version that makes it more explicit.

I believe general pandoc philosophy is to keep the number of necessary definitions in the template as low as possible. We violate that principle whenever the code would otherwise become extremely verbose or when it simplifies customization (e.g., #5016). It's always disruptive for people who must then update their template, and I'm not sure we could justify it in this case.

denismaier:
I see. That's reasonable. Keep it that way then...
Wouldn't it make sense then to define urls only once and reuse afterwards?

denismaier:
What about " Wouldn't it make sense then to define urls only once and reuse afterwards?"

tarleb:
I believe that would be difficult to do nicely, as the current `url1` labels are not very readable. Repeated URLs are probably also a rather rare occurrence, so it may not be worth the effort to do this.

denismaier:
Yes. That's probably the case.

Issue #5539, opened 2019-05-31
ConTeXt Writer: Start and Stop Section Levels


DaveJarvis:
# Background
Using Pandoc 2.7.2, consider the following document:

``` markdown
# Inflation Theory
## 13,813 ± 58
Time did not start with an explosion...

### Relevance
Our Universe's first three minutes produced vast amounts of protons...
```

This results in a document similar to the following:

``` tex
\chapter[title={Inflation Theory},reference={inflation-theory}]

\section[title={13,813 ± 58},reference={section}]

Time did not start with an explosion...

\subsection[title={Relevance},reference={relevance}]

Our Universe's first three minutes produced vast amounts of protons...
```

# Problem
By generating `\chapter`, `\section`, and `\subsection`, etc., it puts constraints around embedding documents within documents. As Hans Hagen wrote on the ConTeXt mailing list:

> I don't know as I never use these coding methods. But it might make sense to check all these mappings onto context, for instance I think that pandoc uses `\section` and such while in fact it should use `\startsectionlevel[title={}] ... \stopsectionlevel` so that one can embed documents in other documents.

# Expected Output
The expected outputs include a standalone document and a document fragment.

## Standalone Document
The relevant portions of the standalone document might resemble:

``` tex
\definestructureconversionset
  [documentstructure]
  [numbers, numbers, numbers]
  [numbers]

\setupheads [sectionconversionset=documentstructure]

\starttext

\startsectionlevel [title={Inflation Theory},reference={inflation-theory}]
  \startsectionlevel [title={13,813 ± 58},reference={section}]
    Time did not start with an explosion...

    \startsectionlevel [title={Relevance},reference={relevance}]
      Our Universe's first three minutes produced vast amounts of protons...
    \stopsectionlevel
  \stopsectionlevel
\stopsectionlevel

\stoptext
```

The `[numbers, numbers, numbers]` code reflects the depth of heading levels in the document. If the document has 5 levels, then `[numbers, numbers, numbers, numbers, numbers]` may be necessary.

## Document Fragment
A document fragment could look as follows:

``` tex
\startsectionlevel [title={Inflation Theory},reference={inflation-theory}]
  \startsectionlevel [title={13,813 ± 58},reference={section}]
    Time did not start with an explosion...

    \startsectionlevel [title={Relevance},reference={relevance}]
      Our Universe's first three minutes produced vast amounts of protons...
    \stopsectionlevel
  \stopsectionlevel
\stopsectionlevel
```

[...]

PR #8106, opened 2022-06-06
ConTeXt writer: use "sectionlevel" environment for headings

tarleb:
The document hierarchy is now conveyed using the
`\startsectionlevel`/`\stopsectionlevel` by default. This makes it easy
to include pandoc-generated snippets in documents at arbitrary levels.
The more semantic environments "chapter", "section", "subsection", etc.
are used if the `--top-level-division` command line parameter is set to
a non-default value.

Closes: #5539

### DIFF
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -80,6 +80,10 @@ $endif$
 \setuphead[subsubsubsection]   [style=\sc]
 \setuphead[subsubsubsubsection][style=\it]
 
+\definesectionlevels
+   [default]
+   [section, subsection, subsubsection, subsubsubsection, subsubsubsubsection]
+
 $if(headertext)$
 \setupheadertexts$for(headertext)$[$headertext$]$endfor$
 $endif$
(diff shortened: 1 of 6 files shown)


denismaier:
I'm not sure I understand. What happens if you have a `--top-level-division`? Will this change the setup, or will this use the environments directly?

tarleb:
Without `--top-level-division` (or `--top-level-division=default`), `# hello` will be rendered as `\startsectionlevel[title={hello}] ... \stopsectionlevel`, whereas with `--top-level-division=chapter` we get `\startchapter[title={hello}] ... \stopchapter`.

The rationale here is that the default should be something that just works, and `\startsectionlevel` fits that bill: it works no matter where or how it is used. With `--top-level-division=chapter` the user wants to enforce a certain behavior, so the more semantic `\startchapter` seems more appropriate here. AFAIU, there is no need to adjust `\definesectionlevels` in this case.

Does this make sense? We can still change it if this is somehow problematic.

PR #8112, opened 2022-06-06
Allow placing custom readers and writers in data subdir

tarleb:


src/Text/Pandoc/Class/PandocMonad.hs
```diff
@@ -668,6 +669,29 @@ withPaths (p:ps) action fp =
   catchError ((p </> fp,) <$> action (p </> fp))
              (\_ -> withPaths ps action fp)
 
+-- | If @fp@ doesn't exist, but the data file relative to @/subdir/@
+-- does, then return that; otherwise return @fp@.
+fallbackDataFile :: PandocMonad m
```
jgm, line 674: I wonder if a more perspicuous name would be something like `findFile`?

src/Text/Pandoc/Class/PandocMonad.hs
```diff
@@ -668,6 +669,29 @@ withPaths (p:ps) action fp =
   catchError ((p </> fp,) <$> action (p </> fp))
              (\_ -> withPaths ps action fp)
 
+-- | If @fp@ doesn't exist, but the data file relative to @/subdir/@
+-- does, then return that; otherwise return @fp@.
+findFile :: PandocMonad m
```
jgm, line 674: Sorry to revise my original suggestion, but what about `findFileWithDataFallback`?
`findFile` should be reserved for something less special-purpose.

jgm: Also, could we use this function to simplify `readMetadataFile`?

tarleb: Both done

src/Text/Pandoc/Writers/Custom.hs
```diff
@@ -25,31 +26,33 @@ import qualified Text.Pandoc.Lua.Writer.Classic as Classic
...
-  -- Most classic writers contain code that throws an error if a global
-  -- is not present. This would break our check for the existence of a
-  -- "Writer" function. We resort to raw access for that reason, but
-  -- could also catch the error instead.
-  let rawgetglobal x = do
-        pushglobaltable
-        pushName x
-        rawget (nth 2) <* remove (nth 2) -- remove global table
+writeCustom luaFile opts doc = do
+  luaFile' <- fromMaybe luaFile <$> findFileWithDataFallback "readers" luaFile
```
jgm, line 30: should this be `"writers"` instead of `"readers"`?

tarleb: Whoops. Careless copy-pasting.

jgm:
This looks good to go.

Issue #8099, opened 2022-06-02
markdown images without valid path targets are excluded from docx output


jjallaire:
We are observing an issue with docx output whereby images must exist prior to calling Pandoc (or they are substituted with their description). This is likely by design but does interfere w/ some more dynamic rendering pipelines (described below).

Consider the following files:

```
elephant.png
subdir/doc.md
```

Where the content of `subdir/doc.md` is:

```
![](../elephant.png)
```

Rendering to docx as follows gives a warning:

```bash
$ pandoc subdir/doc.md --to docx --output doc.docx
[WARNING] Could not fetch resource ../elephant.png: replacing image with description
```

Note that this does not occur for other compound file formats like pdf and epub.

In the case of our application, we are appending together a bunch of .md files (which could be in various sub-directories) and have a Lua filter that fixes up all of the paths so that they are relative to the working directory where pandoc is called from. However, in the case of docx our filter never sees the path so fails to include the image. It would be terrific if docx would tolerate the missing file and allow the Lua filters to fix up the path.

The repro described above can be found at <https://github.com/jjallaire/docx-image-paths>

jjallaire:
Note that there is something about the ".." in the path that triggers this behavior because if I move the `elephant.png` into `subdir` (and fix the image to be `![](elephant.png)`) then things work correctly.

tarleb:
Maybe the `rebase_relative_paths` extension could help here (`--from=markdown+rebase_relative_paths`). That should tell pandoc to look for resources relative to the input document.

jjallaire:
In our case though we are actually passing the markdown as one concatenated file (and passing that file on stdin) so I don't think `rebase_relative_paths` would work.

jgm:
I'm not really understanding the problem.  docx rendering happens last, after filters, so how can the docx rendering mess up what you're trying to do with filters?

The message "replacing image with description" comes from `fillMediaBag`.  I guess we don't use that for PDF and EPUB output (probably mostly because these modules were written before we had a media bag).

If you're concatenating markdown files that live in different directories, then the information needed for relative path resolution is just going to be lost, isn't it?

jjallaire:
The problem is that as a result of the file path not being resolved the `Image` is removed entirely from the AST (and replaced by a `Span` of class `image`). I updated the example repo with this filter:

```lua
function Pandoc(doc)
  print(pandoc.write(doc, "native"))
end
```

Re-running demonstrates the issue:

```bash
[...]

PR #8114, opened 2022-06-07
Populate mediabag after filters have run.

tarleb:
The mediabag is filled with document resources after the filters have
run. This allows, for example, filter authors to modify image paths
before pandoc tries to fetch the images.

Lua filters that rely on a filled mediabag can use the new
`pandoc.mediabag.fill` function to perform that action in the filter.

Closes: #8099

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -306,17 +306,16 @@ convertWithOpts opts = do
                                 >>= r readerOpts
              ByteStringReader r ->
                mconcat <$> mapM (r readerOpts . inputToLazyByteString) inputs)
-            >>=
-              (   (if not (optSandbox opts) &&
-                      (isJust (optExtractMedia opts)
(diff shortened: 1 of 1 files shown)


tarleb:
I'm not sure if this the best option, but I don't have any better ideas.

alerque (approved):

jgm:
Do any of the filters now in the pandoc-filters repository presuppose a filled media bag?  I'm just wondering how much of a breaking change this would be.

tarleb:
None of the filters in that repo retrieve values from the mediabag. The only mediabag functions that are used are `insert` and `fetch`.

tarleb:
I searched the entirety of GitHub and found one Lua filter that expects the current behavior. That one filter tries to correct relative paths, just as #8099.

JSON filters are more difficult to query, and I can't really imagine many cases where JSON filters would be affected by this change. I couldn't find filters that checked for an image-replacing Span.

PR #8115, opened 2022-06-07
fillMediaBag: Keep attributes of original image on Span

tarleb:
Images that cannot be fetched are replaced with a Span that contains the
image's description. The span now also retains all original image
attributes and inherits all attributes of the image. Furthermore, the
classes `image` and `placeholder` are added, and path and title are
store in attributes `original-image-src` and `original-image-title`,
respectively.

Closes: #8099

### DIFF
--- a/src/Text/Pandoc/Class/PandocMonad.hs
+++ b/src/Text/Pandoc/Class/PandocMonad.hs
@@ -688,14 +688,23 @@ fillMediaBag d = walkM handleImage d
                   report $ CouldNotFetchResource src
                             "replacing image with description"
                   -- emit alt text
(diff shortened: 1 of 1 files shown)


tarleb:
This allows to reconstruct the original Image element from its replacement Span; it's an alternative to #8114.

Personally, I think I'd prefer #8114.

alerque:
I'm not sure I like this as an *alternative* to #8114, but as an additional tool in the toolbox passing through enough data to be able to reconstruct the original sounds like a great bonus.

jgm:
I don't see a downside to merging this in addition to #8114.

PR #8116, opened 2022-06-07
ConTeXt writer: support complex table structures.

tarleb:
The following table feature are now supported in ConTeXt:

- Colspans,
- rowspans,
- multiple bodies,
- row headers,
- multi-row table head and foot.


tarleb:
@denismaier it'd be great if you could have a look when time permits.

tarleb:
Missing: 

- [ ] column widths,
- [ ] table widths, and
- [ ] table references.

tarleb:
Implementation details:

-  I also switched to a modern syntax for Natural Tables.
- With xtables, the last row in each body part is now marked as `lastrow`, so it can be styled with `\setupxtable[lastrow][...]`. This is helpful when handling border settings.

jgm:
@denismaier if you're able to look next week, we can always make adjustments.

denismaier:
I've had a quick look. I usually don't really use extreme tables, but from AFAICS this looks reasonable.

PR #8117, opened 2022-06-08
org reader: recognize {webp,jxl} files as images

contrun:
Currently, webp files are not recognized as images. This pr would make `echo "[[file:test.webp]]" | stack run -- -f org -t html -i - -o -` outputs `<p></p>` instead of `<p><a href="test.webp">file:test.webp</a></p>`.

I also added support for [JPEG XL](https://en.wikipedia.org/wiki/JPEG_XL) images.


tarleb:
Thanks!

PR #8121, opened 2022-06-12
add `--embed-resources` flag

### LINKED ISSUE #7331: Enable usage of "self-contained" without "standalone" for HTML
Using the option [`self-contained`](https://pandoc.org/MANUAL.html#options-affecting-specific-writers) currently implies the option `standalone` when producing HTML output. This is not desired in all use cases, especially not when using `Pandoc` as an external renderer in certain context (Content Management System, Templating etc.).

It would be useful to have an option to produce data URIs (i.e. not requiring external images) for a part of included document, rendering only HTML body in that case without other HTML declarations (that are a part of the master document).

Introducing an additional `Pandoc` option would perhaps be the cleanest way of addressing the mentioned use case, since it would not break existing usage, e.g. something like `embed-resources` (or `embedded-references` or `embed-media`

ebobrow:
* add `--embed-resources` flag for `--self-contained` without implying `--standalone`
* deprecate `--self-contained` flag

closes #7331

### DIFF
--- a/src/Text/Pandoc/App/CommandLineOptions.hs
+++ b/src/Text/Pandoc/App/CommandLineOptions.hs
@@ -439,7 +439,14 @@ options =
 
     , Option "" ["self-contained"]
                  (NoArg
-                  (\opt -> return opt { optSelfContained = True }))
+                  (\opt -> do
+                    deprecatedOption "--self-contained" "use --embed-resources --standalone"
+                    return opt { optSelfContained = True }))
+                 "" -- "Make slide shows include all the needed js and css (deprecated)"
+
+    , Option "" ["embed-resources"]
+                 (NoArg
+                  (\opt -> return opt { optEmbedResources = True }))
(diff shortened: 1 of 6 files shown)


src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -439,7 +439,14 @@ options =
 
     , Option "" ["self-contained"]
                  (NoArg
-                  (\opt -> return opt { optSelfContained = True }))
+                  (\opt -> do
+                    deprecatedOption "--self-contained" "use `--embed-resources --standalone`"
```
jgm, line 443: Let's omit the `` ` `` (backticks), since we don't use them on other deprecation option warnings.

jgm:
Thanks very much for this. It looks good, except for a small thing I've noted.

ebobrow:
Thanks! I removed the backticks.

jgm:
Excellent, many thanks!

PR #8124, opened 2022-06-13
Make clear that walk performs a deep copy

badumont:
Relevant discussion here: https://groups.google.com/g/pandoc-discuss/c/-NoL3Ft9AC8

I felt that it could be useful to inform that walk is the only way to get a deep copy of a List. I wrote this where I think that users will likely loop up for this feature, but it may be out of place.


tarleb (approved):
Thank you!

PR #8127, opened 2022-06-15
Update fonts MIME following IANA recommendation

lewer:
Hello.

MIME for **ttf** fonts was changed in https://github.com/jgm/pandoc/pull/6464.

But according to https://www.iana.org/assignments/media-types/media-types.xhtml#font, MIME should be **font/ttf** and not **application/font-sfnt** which is listed as deprecated on the same page.

**font/ttf** is supported by EPUB 3 (https://www.w3.org/publishing/epub3/epub-spec.html#sec-core-media-types)

This PR also changes MIME for **otf** and **woff** fonts.


mb21:
thanks, btw. this list if originally from https://github.com/Happstack/happstack-server/blob/master/src/Happstack/Server/FileServe/BuildingBlocks.hs, I'm sure they'd appreciate a PR as well...

PR #8129, opened 2022-06-15
LaTeX: Handle bold inside tt

### LINKED ISSUE #7525: Latex: Bold inside tt fails, tt inside bold works
**Explain the problem.**

The following file:

```
\usepackage{beramono}
\documentclass[]{book}
\begin{document}
\texttt{\textbf{bold1}}
\textbf{\texttt{bold2}}
\end{document}
```

generates the following HTML:

```
<p><code>bold1</code> <strong><code>bold2</code></strong></p>
```

This is wrong, because both words should be bold. The standard tex tt font has no bold variant, which is probably why this bug was not noticed.

**Pandoc version?**

```
orge:mwe john$ pandoc --version
pandoc 2.14.0.1
Compiled with pandoc-types 1.22, texmath 0.12.3, skylighting 0.10.5.1,
citeproc 0.4, ipynb 0.1.0.1
```

Command line:

```
pandoc mwe.tex -o mwe.html
```

ebobrow:
parse `\texttt{\textbf{bold code}}` as `Strong [Code ("",[],[]) "bold code"]`

closes #7525

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -347,7 +347,7 @@ inlineCommands = M.unions
     , ("textmd", extractSpaces (spanWith ("",["medium"],[])) <$> tok)
     , ("textrm", extractSpaces (spanWith ("",["roman"],[])) <$> tok)
     , ("textup", extractSpaces (spanWith ("",["upright"],[])) <$> tok)
-    , ("texttt", ttfamily)
+    , ("texttt", formatCode nullAttr <$> tok)
     , ("sout", extractSpaces strikeout <$> tok)
     , ("alert", skipopts >> spanWith ("",["alert"],[]) <$> tok) -- beamer
     , ("textsuperscript", extractSpaces superscript <$> tok)
@@ -368,7 +368,7 @@ inlineCommands = M.unions
     , ("it", extractSpaces emph <$> inlines)
     , ("sl", extractSpaces emph <$> inlines)
     , ("bf", extractSpaces strong <$> inlines)
-    , ("tt", code . stringify . toList <$> inlines)
+    , ("tt", formatCode nullAttr <$> inlines)
     , ("rm", inlines)
     , ("itshape", extractSpaces emph <$> inlines)
     , ("slshape", extractSpaces emph <$> inlines)
@@ -407,8 +407,8 @@ inlineCommands = M.unions
     , ("hypertarget", hypertargetInline)
     -- hyphenat
     , ("nohyphens", tok)
-    , ("textnhtt", ttfamily)
-    , ("nhttfamily", ttfamily)
+    , ("textnhtt", formatCode nullAttr <$> tok)
+    , ("nhttfamily", formatCode nullAttr <$> tok)
     -- LaTeX colors
     , ("textcolor", coloredInline "color")
     , ("colorbox", coloredInline "background-color")
@@ -546,9 +546,6 @@ coloredInline stylename = do
   skipopts
   color <- braced
   spanWith ("",[],[("style",stylename <> ": " <> untokenize color)]) <$> tok
-
-ttfamily :: PandocMonad m => LP m Inlines
-ttfamily = code . stringify . toList <$> tok
 
 processHBox :: Inlines -> Inlines
 processHBox = walk convert
--- a/src/Text/Pandoc/Shared.hs
+++ b/src/Text/Pandoc/Shared.hs
@@ -70,6 +70,7 @@ module Text.Pandoc.Shared (
                      eastAsianLineBreakFilter,
                      htmlSpanLikeElements,
                      filterIpynbOutput,
+                     formatCode,
                      -- * TagSoup HTML handling
                      renderTags',
                      -- * File handling
@@ -106,7 +107,7 @@ import Data.Char (isAlpha, isLower, isSpace, isUpper, toLower, isAlphaNum,
                   generalCategory, GeneralCategory(NonSpacingMark,
                   SpacingCombiningMark, EnclosingMark, ConnectorPunctuation))
 import Data.Containers.ListUtils (nubOrd)
-import Data.List (find, intercalate, intersperse, sortOn, foldl')
+import Data.List (find, intercalate, intersperse, sortOn, foldl', groupBy)
 import qualified Data.Map as M
 import Data.Maybe (mapMaybe, fromMaybe)
 import Data.Monoid (Any (..))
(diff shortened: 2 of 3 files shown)


src/Text/Pandoc/Readers/LaTeX.hs
```diff
@@ -547,8 +547,17 @@ coloredInline stylename = do
...
+ttfamily =
+  fromList
+    . concatMap (toList . fmt)
+    . groupBy (\a b -> not (isStrong a || isStrong b))
+    . toList
+  where
+    isStrong (Strong _) = True
+    isStrong _ = False
+    fmt [Strong x] = strong $ code $ stringify x
+    fmt xs = code $ stringify xs
```
jgm, line 560: This will certainly work for the special case of textbf inside texttt.  But I wonder if we couldn't generalize it a bit?  After all, someone might have emph or textsc or any number of other kinds of inline formatting inside texttt...

jgm: One option would be to group Str and Space and SoftBreak (maybe other) elements together into chunks, and leave other inlines separate.  Then:

- for the chunks consisting entirely of Str, Space, SoftBreak, you simply put them inside Code.
- for the other chunks, use `walk` to do the same transformation recursively on their `[Inline]` children.

jgm: Indeed, this code seems general enough to be used in other readers, and maybe could go in Shared?  I have a vague memory that we do something like this already in one of the other readers, but I can't recall which.

ebobrow:
Thanks for the review! I moved the function to Shared, but I couldn't find the other reader you were talking about.

src/Text/Pandoc/Shared.hs
```diff
@@ -779,6 +780,31 @@ filterIpynbOutput mode = walk go
...
+    . B.toList
+  where
+    isPlaintext (Str _) = True
+    isPlaintext Space = True
+    isPlaintext SoftBreak = True
+    isPlaintext (Quoted _ _) = True
+    isPlaintext _ = False
+    fmt xs
+        | all isPlaintext xs = B.toList $ B.code $ stringify xs
+        | any isPlaintext xs = B.toList $ formatCode $ B.fromList xs
```
jgm, line 799: Under what condition could we have `any isPlaintext xs` but not `all isPlaintext xs`?

src/Text/Pandoc/Shared.hs
```diff
@@ -779,6 +780,31 @@ filterIpynbOutput mode = walk go
...
+        | otherwise = walk
+                          ( \ys ->
+                              if length ys == 1 && not (isPlaintext $ head xs)
```
jgm, line 802: I'd like to avoid partial functions like `head` (even though in this case you're checking length). I think maybe you could do something more elegant like

```
\case
  [y] | not (isPlaintext y) ->
  ys ->
```

src/Text/Pandoc/Shared.hs
ebobrow: This is for something with multiple inlines, like

```latex
\texttt{\sout{Normal. \underline{Underlined.}}}
```

While walking through this expression, there will be a point where one of the inlines in `\sout` is handled, but not the other:

```
Strikeout [ Str "Normal. ", Underline [ Code ("",[],[]) "Underlined." ] ]
```

jgm: I see. Instead, why not at line 803 have `concatMap fmt . groupBy (\a b -> isPlaintext a && isPlaintext b)`?
Then we could be confident that `fmt` would never apply to mixed lists.  That might make things a bit clearer.

ebobrow: Agreed, that looks nicer. I've extracted the `concatMap fmt . groupBy (\a b -> isPlaintext a && isPlaintext b)` into a separate function. Please let me know if you can think of a better name for it (or if I should leave it inlined).

ebobrow:
Sorry for all the force-pushes. I've cleaned some stuff up and I tihnk it's looking better now.

jgm:
I want to suggest these changes, which I think clarify the code even further, and also give us the flexibility to add attributes to the Code elements.
```
index 3f34e0b8a..988cc3055 100644
@@ -347,7 +347,7 @@ inlineCommands = M.unions
     , ("textmd", extractSpaces (spanWith ("",["medium"],[])) <$> tok)
     , ("textrm", extractSpaces (spanWith ("",["roman"],[])) <$> tok)
     , ("textup", extractSpaces (spanWith ("",["upright"],[])) <$> tok)
-    , ("texttt", formatCode <$> tok)
+    , ("texttt", formatCode nullAttr <$> tok)
     , ("sout", extractSpaces strikeout <$> tok)
     , ("alert", skipopts >> spanWith ("",["alert"],[]) <$> tok) -- beamer
     , ("textsuperscript", extractSpaces superscript <$> tok)
@@ -368,7 +368,7 @@ inlineCommands = M.unions
     , ("it", extractSpaces emph <$> inlines)
     , ("sl", extractSpaces emph <$> inlines)
     , ("bf", extractSpaces strong <$> inlines)
-    , ("tt", formatCode <$> inlines)
+    , ("tt", formatCode nullAttr <$> inlines)
     , ("rm", inlines)
     , ("itshape", extractSpaces emph <$> inlines)
     , ("slshape", extractSpaces emph <$> inlines)
@@ -407,8 +407,8 @@ inlineCommands = M.unions
     , ("hypertarget", hypertargetInline)
     -- hyphenat
     , ("nohyphens", tok)
-    , ("textnhtt", formatCode <$> tok)
-    , ("nhttfamily", formatCode <$> tok)
+    , ("textnhtt", formatCode nullAttr <$> tok)
+    , ("nhttfamily", formatCode nullAttr <$> tok)
     -- LaTeX colors
     , ("textcolor", coloredInline "color")
     , ("colorbox", coloredInline "background-color")
diff --git a/src/Text/Pandoc/Shared.hs b/src/Text/Pandoc/Shared.hs
index 0d6a1b559..d6d24e4e6 100644
--- a/src/Text/Pandoc/Shared.hs
+++ b/src/Text/Pandoc/Shared.hs
@@ -780,13 +780,14 @@ filterIpynbOutput mode = walk go
                     | otherwise = ""
         go x = x
 
--- | Move formatting within code inlines to the outside,
--- i.e. `Strong [ Code ("",[],[]) "Strong code" ]` (issue #7525)
-formatCode :: Inlines -> Inlines
-formatCode =
-  B.fromList
-    . fmt
-    . B.toList
+-- | Reformat 'Inlines' as code, puting the stringlike parts in 'Code'
+-- elements while bringing other inline formatting outside.
+-- The idea is that e.g. `[Str "a",Space,Strong [Str "b"]]` should turn
+-- into `[Code ("",[],[]) "a ", Strong [Code ("",[],[]) "b"]]`.
+-- This helps work around the limitation that pandoc's Code element can
+-- only contain string content (see issue #7525).
+formatCode :: Attr -> Inlines -> Inlines
+formatCode attr = B.fromList . fmt . walk fmt . B.toList
   where
     isPlaintext (Str _) = True
     isPlaintext Space = True
@@ -796,13 +797,8 @@ formatCode =
     fmt = concatMap go . groupBy (\a b -> isPlaintext a && isPlaintext b)
       where
         go xs
-          | all isPlaintext xs = B.toList $ B.code $ stringify xs
-          | otherwise = walk
-                            ( \case
-                                  [y] | not (isPlaintext y) -> [y]
-                                  ys -> fmt ys
-                            )
-                            xs
+          | all isPlaintext xs = B.toList $ B.codeWith attr $ stringify xs
+          | otherwise = xs
 
 --
 -- TagSoup HTML handling
```

ebobrow:
Done, thanks! Unfortunately I can't remove `walk` from the `otherwise` branch because it stops `code` from being applied.

jgm:
> Unfortunately I can't remove walk from the otherwise branch because it stops code from being applied to the innermost Inlines.

I don't think that's true!  Remember, `walk` is a bottom-up traversal, so in `walk fmt` `fmt` will be applied to the innermost inline lists first.  The above code passed all your tests -- do you have a test case it doesn't pass?

src/Text/Pandoc/Shared.hs
```diff
@@ -779,6 +780,26 @@ filterIpynbOutput mode = walk go
                     | otherwise = ""
         go x = x
 
+-- | Reformat 'Inlines' as code, puting the stringlike parts in 'Code'
```
jgm, line 783: puting -> putting

ebobrow:
Perfect, thanks for all your help! I can work on a second PR after this is merged.

jgm:
Great, thanks!

PR #8130, opened 2022-06-16
Allow aeson 2.1

tarleb:


tarleb:
The tests still succeed with

    cabal test --constraint='aeson == 2.1.0.0' --allow-newer

PR #8132, opened 2022-06-17
Prepare for liftA2 being exported from Prelude

googleson78:
CLC has approved the proposal to export `liftA2` from `Prelude`
(https://github.com/haskell/core-libraries-committee/issues/50).
It means that the entirety of `Applicative` will now be exported from `Prelude`.

The implementation of the proposal is delayed at least to GHC 9.6,
but one can already future-proof code to be
compliant with this change in a backwards-compatible way. No CPP required.

Migration guide and more info:
https://github.com/haskell/core-libraries-committee/blob/main/guides/export-lifta2-prelude.md (not yet merged, draft can be seen [here](https://github.com/googleson78/core-libraries-committee/blob/lifta2-guide/guides/export-lifta2-prelude.md)

### DIFF
--- a/src/Text/Pandoc/Parsing/Types.hs
+++ b/src/Text/Pandoc/Parsing/Types.hs
@@ -19,7 +19,8 @@ module Text.Pandoc.Parsing.Types
   )
 where
 
-import Control.Applicative (liftA2)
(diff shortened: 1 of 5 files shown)


tarleb (approved):
LGTM

Issue #8059, opened 2022-05-07
Org-mode reader: `#+pandoc-emphasis-pre` doesn't work as expected


adql:
**Explain the problem.**
Adding characters to `#+pandoc-emphasis-pre` as [described in the manual](https://pandoc.org/org.html#emphasis-rules) doesn't work as expected. Interestingly, adding to `#+pandoc-emphasis-post` does.

Minimal working example `test.org`:
```
#+pandoc-emphasis-pre: "-\t ('\"{T"
#+pandoc-emphasis-post: "-\t\n .,:!?;'\")}[t"

1. T/est/ with T allowed as pre
2. /Tes/t with t allowed as post
3. Normal /emphasis/, and in {/brackets/}
```

Command: `pandoc -o test.md test.org`

Expected `test.md` result:
```markdown
1.  T*est* with T allowed as pre
2.  *Tes*t with t allowed as post
3.  Normal *emphasis*, and in {*brackets*}
```

Actual result:
```markdown
1.  T/est/ with T allowed as pre
2.  *Tes*t with t allowed as post
3.  Normal *emphasis*, and in {*brackets*}
```

Exporting to Pandoc AST confirms that the problem is with the reader.

Also completely replacing the strings with `"T"` and `"t"` respectively achieves the same result.

**Pandoc version?**
Pandoc 2.18 on Manjaro Linux (`pandoc-2.18-linux-amd64.tar.gz` from the [release page](https://github.com/jgm/pandoc/releases/tag/2.18)). Also happens on https://pandoc.org/try

adql:
The relevant test in `tests/Tests/Readers/Org/Meta.hs`
```haskell
[ "Changing pre and post chars for emphasis" =:
  T.unlines [ "#+pandoc-emphasis-pre: \"[)\""
              , "#+pandoc-emphasis-post: \"]\\n\""
              , "([/emph/])*foo*"
              ] =?>
  para ("([" <> emph "emph" <> "])" <> strong "foo")
```
, which tests adding the non-standard `[` to `pre`, passes flawlessly. Possibly the bug occurs on alphanumeric chars (?) – I tried manipulating the test with `T`, `t`, and `3`, all fail.

adql:
I tried more cases and it seems like this is more general that I first thought. The only characters I managed so far to have as `pre` are various parentheses, `$` and `+`. Alphanumeric chars, `!`, `%`, `#` all fail. Probably others as well.

tarleb:
Related issue: #6070

PR #8134, opened 2022-06-19
Org reader: Fix `Post` standing for `Pre` in `setEmphasisPreChar`

adql:
Small typing mistake which causes e.g. `)/emph/)` and `(/noemph/)` on the default pre-chars.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/Meta.hs
+++ b/src/Text/Pandoc/Readers/Org/Meta.hs
@@ -217,7 +217,7 @@ setSelectedTags tags st =
 
 setEmphasisPreChar :: Maybe [Char] -> OrgParserState -> OrgParserState
 setEmphasisPreChar csMb st =
-  let preChars = fromMaybe (orgStateEmphasisPostChars defaultOrgParserState) csMb
+  let preChars = fromMaybe (orgStateEmphasisPreChars defaultOrgParserState) csMb
   in st { orgStateEmphasisPreChars = preChars }
 
 setEmphasisPostChar :: Maybe [Char] -> OrgParserState -> OrgParserState


adql:
I discovered this by chance while trying to work on #8059. It appeared in [this commit](https://github.com/jgm/pandoc/commit/621e86402339209ea340dd36dfbd3d9eefc28e85) about 4 years ago, go figure how this hasn't disturbed anyone so far (couldn't find a relevant issue).

tarleb:
Thank you, good catch! If I understand correctly, then the bug was triggered only when the new "pre" setting could not be parsed.

adql:
As I understand, it actually affected all parsing when no `pandoc-emphasis-pre` is set, i.e. the default behavior (since `setEmphasisPreChar` pulls the default from `defaultOrgParserState` when no setting is given, and until now it accidentally pulled the `Post`).

Kind of surprising that it went unnoticed, but then the only affected characters in the default setting are the directional ones `(` and `{`. Maybe it's not very common to emphasize directly after an opening parenthesis.

tarleb:
The function `setEmphasisPreChar` is called when the org-file contains a `#+pandoc-emphasis-pre` line. Pandoc treats the rest of that line as a Haskell expression that defines a string, i.e., it should be surrounded by double quotes. If that succeeds, then the pre chars are updated. If, however, the line does not contain a valid Haskell string, then `setEmphasisPreChar` is called with `Nothing` as the first argument. Only then does the `fromMaybe` evaluate the fallback value (its first argument).

The idea behind that mechanism is that a sole `#+pandoc-emphasis-pre`, without any arguments, will reset to the default behavior. That mechanism, as you found, was broken. The value was set to the allowed post chars instead of reset to its initial value.

adql:
That's interesting. Where is the default behavior actually defined then?

I should mention that `#+pandoc-emphasis-pre` actually hardly works at all. I tested several cases and it seems like #8059 is broader than at first seemed. Most characters that I tried to add failed. The test in the suite only tests for `[`, which happens to pass.

Issue #8120, opened 2022-06-12
Extensions cannot be used with custom writers


asmaier:
**Explain the problem.**
I created a custom writer in Lua for kramdoc based on pandoc markdown:

```
function Writer (doc, opts)
    local filter = {
      Math = function(elem)

        local math = elem

        if elem.mathtype == 'DisplayMath' then
            local delimited = '\n$$' .. elem.text ..'$$\n\n'
            math = pandoc.RawInline('markdown', delimited)
        end

        if elem.mathtype == 'InlineMath' then
            local delimited = '$$' .. elem.text ..'$$'
            math = pandoc.RawInline('markdown', delimited)
        end 
        
        return math
      end           
    }
    return pandoc.write(doc:walk(filter), 'markdown', opts)
  end
```
When using it like this it works:

`pandoc -t kramdown.lua -o vector.md vector.tex`

However when I try to activate or deactivate extensions by adding or substracting the extensions name to my custom writer (as described at https://pandoc.org/MANUAL.html#extensions) it doesn't work, e.g. when trying to deactivate the extension for `smart` quotes:

```
pandoc -t kramdown.lua-smart -o vector.md vector.tex
Error running Lua:
cannot open kramdown.lua-smart: No such file or directory
```
Obviously activating/deactivating extensions for custom writers doesn't work like for the inbuilt supported formats (I also tried the combinations `kramdown-smart.lua` or `kramdown-smart`). Is this a bug/missing feature or is there another way to activate/deactivate extensions when using custom writers?

**Pandoc version?**
Pandoc 2.18 on Mac OS 10.15.7

tarleb:
That's an interesting problem. Let's start with the workaround first: an easy way would be to set a variable, e.g.:

    pandoc -t kramdown.lua -V extensions=-smart ...

and then change the filter like that:

    local exts = opts.variables.extensions or ''
    return pandoc.write(doc:walk(filter), 'markdown' .. exts, opts)

---

I like the idea of specifying extensions for custom writers in the way you propose. There are technical details that must be answered, like

- Should we allow arbitrary extensions on custom readers, or just those that pandoc knows about?
- How would that information get passed to the Writer function?

We should probably do the same for readers.

jgm:
> How would that information get passed to the Writer function?

We already have an `extensions` field in `PANDOC_WRITER_OPTIONS` and `PANDOC_READER_OPTIONS`.

[...]

PR #8141, opened 2022-06-22
Lua: support format extensions in custom writers.

tarleb:
Custom writers can define the extensions that they support via the
global `writer_extensions`. The variable's value must be a table with
all supported extensions as keys, and their default status as values.

E.g., the below specifies that the writer support the extensions `smart`
and `sourcepos`, but only the `smart` extension is enabled by default:

    writer_extensions = {
      smart = true,
      sourcepos = false,
    }


tarleb:
Here it is in action, based on the example in #8120:

``` lua
writer_extensions = {
  smart = true,
}

function Writer (doc, opts)
  local filter = {
    Math = function (elem)
      local raw_md = elem.mathtype == 'DisplayMath'
        and '\n$$' .. elem.text ..'$$\n\n'
        or '$$' .. elem.text ..'$$'
      return pandoc.RawInline('markdown', raw_md)
    end
  }
  local exts = pandoc.List{}
  local enabled_exts = pandoc.List(opts.extensions)
  for ext in pairs(writer_extensions) do
    local polarity = enabled_exts:includes(ext) and '+' or '-'
    exts:insert(polarity .. ext)
  end
  return pandoc.write(doc:walk(filter), 'markdown' .. table.concat(exts), opts)
end
```

It get's the job done, but feels rather clunky.

tarleb:
Another downside: the following assert may fail:

``` lua
function Writer (_, opts)
  assert(pandoc.List(PANDOC_WRITER_OPTIONS.extensions) ==
         pandoc.List(opts.extensions))
  return ''
end
```

The value in `PANDOC_WRITER_OPTIONS.extensions` is computed in `getWriter`, *before* the Lua file is loaded, but `opts.extensions` is not know till *after* the file was loaded.

jgm:
Let me know if this is ready.
Re the downside mentioned above:  should we deprecate the globals like PANDOC_WRITER_OPTIONS? They should not be necessary with new-style writers, right?

tarleb:
Ready from my side.

The handling of `PANDOC_WRITER_OPTIONS` already changed in #8343 (support for bytestring writers). There docs mentioned the new behavior, but I've now added a deprecation note as well.

jgm:
Excellent!

PR #8142, opened 2022-06-22
Remove extra soft break for tasklist

black-desk:
Browser will display the extra newline character between checkbox and
text as a space, which make tasklist items cannot be aligned.

I just remove it.

图片

before:
图片
after: 
图片

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -502,7 +502,7 @@ listItemToHtml opts bls
       let checkbox  = if checked
                       then checkbox' ! A.checked ""
                       else checkbox'
-          checkbox' = H.input ! A.type_ "checkbox" ! A.disabled "" >> nl
+          checkbox' = H.input ! A.type_ "checkbox" ! A.disabled ""
       isContents <- inlineListToHtml opts is
       bsContents <- blockListToHtml opts bs
       return $ constr (checkbox >> isContents) >> bsContents
--- a/test/command/tasklist.md
+++ b/test/command/tasklist.md
@@ -6,10 +6,8 @@ tests adapted from <https://github.github.com/gfm/#task-list-items-extension->
 - [x] bar
 ^D
 <ul class="task-list">
-<li><input type="checkbox" disabled="" />
-foo</li>
-<li><input type="checkbox" disabled="" checked="" />
-bar</li>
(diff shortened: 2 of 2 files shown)


jgm:
This seems a good change! We need tests to be updated before CI will pass.

jgm:
Thanks!

mb21:
But usually you would want space between the check-mark and the text?

Also, pandoc markdown was according to GitHub Markdown spec before this PR: https://github.github.com/gfm/#task-list-items-extension-

black-desk:
I think the space between text and box should be done by css but not literal space.

tarleb:
> I think the space between text and box should be done by css but not literal space.

Agreed, but this brings us into a dilemma: either we break compatibility with the GFM spec that gave rise to this feature, or we produce overly-specific HTML.

My preference is with keeping this PR, but the choice should be an informed decision.

black-desk:
> My preference is with keeping this PR, but the choice should be an informed decision.

Agreed, but I not familiar with Haskell at all. I not even understand the line I changed :(

Sorry, I cannot help :(

tarleb:
No worries. What I mean is just that it's important to discuss the issue that @mb21 brought up. From my point of view, there are no further changes necessary.

jgm:
Thanks for bringing this up, I hadn't considered this possible drawback.
I think it's okay if our HTML output departs slightly from GitHub's as long as semantically it's the same. However, we do want to ensure that there's a space between the check box and the text.  Could we add something to our default CSS for this?

black-desk:
the screenshots I show in https://github.com/jgm/pandoc/pull/8142#issue-1280674790 had apply my own css:

```css
input[type="checkbox"] {
        margin-right:0;
        margin-left: -1.5em;
        width: 1.5em;
}
```

the default situation is weird:

图片

because the checkbox is part of the <p>, and pandoc's css set:

```css
ul.task-list{list-style: none;}
```

:(

but in my option of view, the checkbox is some kind of ::mark which should be a sibling of <p>

图片

jgm:
OK, that makes sense. Maybe we should have a separate issue (since this PR is now closed) suggesting making the checkbox a sibling of the block-level content, and suggesting some default CSS for it?

black-desk:
> OK, that makes sense. Maybe we should have a separate issue (since this PR is now closed) suggesting making the checkbox a sibling of the block-level content, and suggesting some default CSS for it?

done, #8151

Issue #8168, opened 2022-07-07
Don't output line breaks when writing DocBook `<programListing>` elements


studoot:
Currently, Pandoc's code for writing DocBook inserts line breaks between`<programlisting>` opening and closing tags. This is a problem because this element contains verbatim text, where whitespace (including line breaks) is interpreted as being significant, leading to extraneous space at the top of code blocks.

Here are a couple of screen shots showing the difference between DocBook (transformed into PDF) straight out of Pandoc:

![bad](https://user-images.githubusercontent.com/799344/177810447-1cd665db-844c-4f1a-b187-88252cf6dd82.png)

and then where I've manually edited out the line breaks:

![good](https://user-images.githubusercontent.com/799344/177810439-fdb3ec4e-a076-49ea-ba38-223c70016542.png)

Note the gap between border and text in the first image...

The [code for this](https://github.com/jgm/pandoc/blob/7fdc01ac0d91a037b48847939b762654174b4125/src/Text/Pandoc/Writers/Docbook.hs#L256-L257) is pretty straightforward, so I'm in the process of raising a pull request (#8169) that simply removes the `cr <>` fragments in that function.

studoot:
Closing, as a closer inspection of DocBook references indicate leading and trailing linebreaks in `<programlising>` elements should be allowable and not propagate through to XSL:FO output (which is where I was witnessing the extra line).

PR #8169, opened 2022-07-07
Remove leading and trailing line breaks within `programlisting` DocBo…

studoot:
Implements #8168 by removing output of line breaks within DocBook `programlisting` elements.


studoot:
Closing, as a closer inspection of DocBook references indicate leading and trailing linebreaks in `<programlising>` elements should be allowable and not propagate through to XSL:FO output (which is where I was witnessing the extra line).

PR #8173, opened 2022-07-10
Fix mistakes in the User’s Guide

ivan-pan:
I corrected some mistakes such as `yyyy[mm[dd]]]` (instead of `yyyy[mm[dd]]`) and `LaTex` (instead of `LaTeX`).

**Note:** `` `beameroption` `` appears with the first backtick in the HTML output, I do not know why that is.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -5471,7 +5471,7 @@ containing file, relative to the working directory, and prepend
 the resulting path to the link or image path.
 
 The use of this extension is best understood by example.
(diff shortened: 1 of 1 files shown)


jgm:
Terrific -- thanks for this.

jgm:
Looks like an earlier version of MANUAL.txt had
```
``beameroption`
```
It appears that this has been fixed already in e1dc292a0c19a33dffd69d37784c7a5926e684f5

PR #8184, opened 2022-07-14
Lua: extend pandoc.system module.

tarleb:
The module now has the additional functions `list_directory`,
`make_directory`, and `remove_directory`. This makes it easier to write
cross-platform scripts that need to inspect or modify the file system.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -4417,6 +4417,60 @@ Returns:
 
 -   The current working directory (string).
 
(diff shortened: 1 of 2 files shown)


jgm:
I guess the main use for create/delete directory would be to create temporary directories?  In that case, I wonder if it would be better to export functions that do that?  I don't think the Lua interface contains general-purpose functions to create and delete files (or does it?), so it seems a bit odd to have such functions for directories.

PR #8188, opened 2022-07-16
Fix links in manual

bewuethr:
I've bumped into two links that broke because the section headings they link to have changed.


jgm:
Thanks!  I guess that's the downside of implicit reference links.

PR #8202, opened 2022-07-26
Support rowspans and colspans in grid tables

tarleb:


tarleb:
This is unfinished work, so this PR is currently more of a preview and an RFC. The new parser doesn't yet support some Markdown features like column alignment, so it currently only works for reStructuredText.

The parser is a quite verbatim reimplementation of the algorithm used in rst, with all its pros and cons. I'm going to put it into a new package `gridtables` in the hope that it should be reasonably easy to reuse the library in commonmark-extensions.

tarleb:
This should be good to go here (on the pandoc side), but I'll continue with more improvements in the *gridtables* package. See [this project](https://github.com/users/tarleb/projects/1/views/1) for my todo list.

jgm:
Cool! I wasn't quite clear from the above what the status of column alignments is in markdown grid tables.  We don't want to drop support for them if they're already supported.  Maybe they're now supported by your gridtables?  I guess we have tests already that test this feature?

Other question is about support in the markdown writer. I guess we still don't have that, correct?
Is this something you plan to add to gridtables?

tarleb:
The change is *almost* completely backwards compatible, with two exceptions:

1. If there are multiple blocks in a cell then `Para` are not converted to `Plain`; they remain as `Para` elements. There was one test case that I changed because of this.
2. Zero-width characters are treated like half-width chars. This is consistent with the behavior in docutils RST, but differs from pandoc's current parser, which treats those as having no width.

Side note on character widths: neither docutils nor pandoc handle the table below as one might expect, with docutils failing and pandoc producing spurious results.
```
+----+---+---+
|Fo👨‍🔬|Bar| 1 |
+----+---+---+
```

I do plan to add writer support as well. I'd like to do:

1. Add support for `doclayout`-based output to `gridtables`.
2. Create another separate package `text-tables` (or maybe `textual-tables`) that comes with a typeclass and unified interfaces for different table representations.
3. Write the code to connect pandoc to these new types.

My hope here is to reduce the number of table conversions we need to do in our code. Appart from pandoc's default Table constructor, we also have AnnotatedTable, the (unfortunately named) GridTable for docx output, OrgTable, and possibly more -- I hope to move some of these to external packages and find a common interface.

It may take quite a few weeks though, due to various circumstances.

jgm:
> Zero-width characters are treated like half-width chars. This is consistent with the behavior in docutils RST, but differs from pandoc's current parser, which treats those as having no width.

Can you clarify?  Is this because treating them as 0-width will mess up RST parsing, or for other reasons?

> ```
> +----+---+---+
> |Fo👨‍🔬|Bar| 1 |
> +----+---+---+
> ```

Can you explain why this isn't handled properly? Is it because of a bad width calculation for the emoji, and if so, maybe we can fix this?  (I know that we recently did a lot of work getting proper widths from the emojis library, but maybe we haven't integrated this into pandoc?)

ALthough it's not ideal to have reader but not writer support (round-trip will break, for example), it's probably okay for the moment to merge this.

tarleb:
The problem with zero-width characters is the cell-tracing algorithm used by rst: the input is treated as an array of characters, and the algorithm then walks the columns and rows of that array to trace out the table cells. Adding placeholders into the array for wide characters is easy, but handling zero-width chars isn't.

> Can you explain why this isn't handled properly? Is it because of a bad width calculation for the emoji, and if so, maybe we can
> fix this? (I know that we recently did a lot of work getting proper widths from the emojis library, but maybe we haven't integrated
> his into pandoc?)

Oh, I didn't know that was available! Yes, the problem is the width calculation, as the emoji consists of three characters (two emoji and the zero-width joiner). Emojis like 👨‍👩‍👦 contain five chars, etc. I'm not sure if its reasonable, but I was thinking about treating all characters connected with a zwj as a single character that has the width of the widest character that's being joined.

jgm:
Actually the changes I was thinking about are in doclayout -- which should provide accurate widths for emojis now.
See the changelog.
https://hackage.haskell.org/package/doclayout-0.4/changelog
There's a function in there to detect zero-width joiners.

jgm:
Let me know when you consider this ready to merge.

tarleb:
If the different behavior with regard to zero-width characters is acceptable for now, then I think we're good to go.

To be honest, I'm still a bit nervous as the *gridtables* package makes heavy use of array manipulations. It feels fragile.¹ But, given that all tests pass, I'm confident that it won't cause any serious problems.

¹ Maybe Liquid Haskell and dependent types can solve that.

jgm:
I am kind of nervous about the zero-width character issue.

They can be used not just with emojis, but, e.g., to defeat ligatures.

And it seems that there are some languages in which they are used more widely.
https://ptiglobal.com/2018/04/26/the-beauty-of-unicode-zero-width-characters/

There's also stuff like the [text presentation selector](http://www.unicode.org/reports/tr51/#def_text_presentation_selector), FE0E, which we use in pandoc's HTML writer to prevent macos from rendering the footnote backreferences as huge emojis.

Is it possible to modify the RST algorithm somehow to allow for zero-width code points?

Perhaps a simple modification would be to include each zero-width code point in a cell with the following (or maybe preceding) code point?  That's sort of like what you were suggesting for the emojis, above, but generalized.

tarleb:
:+1: I'll give that a try.

tarleb:
OK, I've pushed some new additional tests with wide and zero-width characters to master. The updated *gridtables* version passes these tests now.

jgm:
Excellent, thanks!

jgm:
Sorry, I didn't mean to squash your commits.

tarleb:
:partying_face: :tada: 

No worries about squashing, I only added a second commit to highlight the API change.

PR #8208, opened 2022-07-29
HTML: use CSS flex boxes for columns

tarleb:
This allows to render an arbitrary number of columns, while the previous
approach assumed exactly two columns.

### DIFF
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -163,7 +163,8 @@ header {
 $endif$
 code{white-space: pre-wrap;}
 span.smallcaps{font-variant: small-caps;}
-div.column{display: inline-block; vertical-align: top; width: 50%;}
+div.columns{display: flex; gap: min(4vw, 1.5em);}
+div.column{flex: 1;}
 div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
 ul.task-list{list-style: none;}
 $if(quotes)$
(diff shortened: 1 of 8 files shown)


tarleb:
I guess another downside is that the wrapping `columns` div would then be required, while it can currently be omitted. That's not documented though.

What do you think @mb21?

data/templates/styles.html
```diff
@@ -164,7 +164,8 @@
 code{white-space: pre-wrap;}
 span.smallcaps{font-variant: small-caps;}
 span.underline{text-decoration: underline;}
-div.column{display: inline-block; vertical-align: top; width: 50%;}
+div.columns{display: flex;}
+div.column{flex: 1;}
```
mb21, line 168: cool, yes, the previous CSS is probably ancient...

tho thougths:

instead of doing `div.column{flex: 1;}`, could also do `div > div {flex: 1;}`, then we wouldn't have to write the `.column` class in markdown... but yeah since the `:::` syntax needs a class anyway, probably just as well to keep it like this.

while we're at it, I'd probably add to the `.columns` some kind of gap, for example setting the gap to `2em` except on small screens where that would be too much set it to 4% of the viewportwidth:

    gap: min(4vw, 2em);

tarleb: Good idea! I've added the gap, but opted for `1.5em`. I didn't know CSS had a `min` function, that's extremely useful!

PR #8225, opened 2022-08-11
Lua: add function `pandoc.utils.citeproc`

tarleb:
The function runs the *citeproc* processor on a Pandoc document.
Exposing this functionality to Lua allows to make citation processing
part of a filter or writer, simplifies the creation of multiple
bibliographies, and enables the use of varying citation styles in
different parts of a document.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -3589,6 +3589,28 @@ Usage:
     --   pandoc.Emph{ pandoc.Str 'Paragraph2' }
     -- }
 
+### `citeproc (doc)` {#pandoc.utils.citeproc}
+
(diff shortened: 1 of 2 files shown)


tarleb:
This current way to implement this method is to call `pandoc.utils.run_json_filter(doc, 'pandoc', {'--citeproc'})`, but it has some downsides: Besides being slow, it also uses the first `pandoc` executable in the `PATH`, which may not be the same executable as the one running the filter.

I'm not sure if the `pandoc.utils` module is the best place for this function, but it seemed natural to place it next to the very related `pandoc.utils.references` function.

jgm:
Sounds good to me!

PR #8200, opened 2022-07-26
MediaWiki reader: Allow optional '|-' before Table caption as described in wikipedia document

chinapedia:
image

[Help:Table](https://en.wikipedia.org/wiki/Help:Table)

There are a lot of cases where `|-` appeared before the table caption in wikipedia, this PR would stop raising error for those cases. 

After this change, this kind of [error](https://github.com/chinapedia/wikipedia.zh/blob/master/Errors/2000年中華民國總統選舉.err.log) stopped:

> Error at "/var/folders/71/0dwsyc8x37bdb69jn73jfncw0000gn/T/pandoc62df9cc69a796" (line 194, column 3):
unexpected '+'
|+'''第10任[[中華民國總統|中華民國總統]][[中華民國副總統|副總統]]選舉各縣市投票結果''' 
  ^

Related [wikitext](https://github.com/chinapedia/wikipedia.zh/blob/master/Errors/12月24日.wikitext): 
```
 {| class="wikitable" style="margin: auto; border: 1px #aaa solid; border-collapse: collapse; font-size: 95%; line-height: 1em"
|-  
|+'''第10任[[中華民國總統|中華民國總統]][[中華民國副總統|副總統]]選舉各縣市投票結果''' 
|-
! rowspan=2 | 縣市 !! colspan=2 style="border-bottom:4px solid #FF6310;" | 宋楚瑜、張昭雄 !! colspan=2 style="border-bottom:4px solid #000099;" | 連戰、蕭萬長 !! colspan=2 style="border-bottom:4px solid #fdd731;" | 李敖、馮滬祥 !! colspan=2 colspan=2 style="border-bottom:4px solid #000000;" | 許信良、朱惠良 !! colspan=2  style="border-bottom:4px solid #1B9431;" | 陳水扁、呂秀蓮
! rowspan=27 |[[File:2000ROCPresident.svg|350px]]

|- bgcolor="#BEBEBE"
!|得票數!!|得票率!!|得票數!!|得票率!!|得票數!!|得票率!!得票數!!得票率!!|得票數!!|得票率
|-

```

PR #8231, opened 2022-08-17
Parse mediawiki table cell with attribs, to support rowspan, colspan

liruqi:
Currently mediawiki table cells are constructed by `simpleCell` which unable to pass rowspan & colspan attributes. But [wikitext table syntax supports rowspan & colspan attributes](https://en.wikipedia.org/wiki/Help:Table#Combined_use_of_COLSPAN_and_ROWSPAN) like HTML tables. This PR would parse mediawiki span attributes like HTML table parser.

Code changes from #8200 are included here too.

### DIFF
--- a/src/Text/Pandoc/Readers/MediaWiki.hs
+++ b/src/Text/Pandoc/Readers/MediaWiki.hs
@@ -35,7 +35,7 @@ import Text.Pandoc.Definition
 import Text.Pandoc.Logging
 import Text.Pandoc.Options
 import Text.Pandoc.Parsing hiding (nested, tableCaption)
-import Text.Pandoc.Readers.HTML (htmlTag, isBlockTag, isCommentTag)
+import Text.Pandoc.Readers.HTML (htmlTag, isBlockTag, isCommentTag, toAttr)
 import Text.Pandoc.Shared (safeRead, stringify, stripTrailingNewlines,
                            trim, splitTextBy, tshow, formatCode)
 import Text.Pandoc.XML (fromEntities)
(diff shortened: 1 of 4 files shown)


liruqi:
Hi, @jgm @tarleb 

While trying to convert wikitext to gfm, table would be converted to either html or gfm format. Could you point out the choosing logic here? Thanks.

Sample input: https://github.com/chinapedia/mediawiki-to-gfm/blob/master/tests/pandoc/1.mediawiki
Sample output: https://github.com/chinapedia/mediawiki-to-gfm/blob/master/tests/pandoc/1.md

jgm:
If the table can't be rendered as a pipe table, because the cells contain block-level content (e.g. multiple paragraphs or things like lists), then we have to fall back on HTML.

liruqi:
@jgm I found related logic in [Markdown writer](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Writers/Markdown.hs).

There's a design issue here, `hasSimpleCells` does not check for rowspan & colspan values. And `gfm` table does support rowspan & colspan (I'm not sure about other markdown formats), I suggest table with any rowspan>1 or colspan>1 be rendered as HTML table. 

Could I move this switch case to the top?
```
| isEnabled Ext_raw_html opts -> fmap (id,) $
       literal <$>
       writeHtml5String opts{ writerTemplate = Nothing } (Pandoc nullMeta [t])

```

jgm:
> There's a design issue here, hasSimpleCells does not check for rowspan & colspan values. And gfm table does support rowspan & colspan attributes (I'm not sure about other markdown formats), I suggest table with any cell rowspan>1 or colspan>1 be rendered as HTML table.

I'd be okay with that.  The system is designed so that the fallback you'll get with colspans and rowspans is not TERRIBLE, but an HTML table would always be more accurate.  The logic was designed before we had row/colspans.

> Could I move this switch case to the top, as a workaround for now?

Wouldn't that mean that HTML tables are ALWAYS used?  We wouldn't want that.

> Another design issue is, the table head could be not only in the first row. There could be 2 rows of table head, and 1st column of each row could also be table head. This kind of structure could not be presented in pandoc AST for now.

If I'm not mistaken, we can represent multiple header rows.
A `TableHead` can take any number of rows.

src/Text/Pandoc/Readers/MediaWiki.hs
```diff
@@ -222,37 +222,47 @@ table = do
...
+  let toRow = Row nullAttr
       toHeaderRow l = [toRow l | not (null l)]
   return $ B.table (B.simpleCaption $ B.plain caption)
                    cellspecs
                    (TableHead nullAttr $ toHeaderRow headers)
                    [TableBody nullAttr 0 [] $ map toRow rows]
                    (TableFoot nullAttr [])
 
+calculateAlignments :: [Cell] -> [Alignment]
+calculateAlignments cells =
```
tarleb, line 248: This could be shortened to

``` haskell
calculateAlignments = map cellAligns
  where
    ...
```

tarleb (approved):
Thank you for this! I left an inline comment, but LGTM otherwise!

src/Text/Pandoc/Readers/MediaWiki.hs
liruqi: @tarleb updated. thanks.

src/Text/Pandoc/Readers/MediaWiki.hs
```diff
@@ -289,37 +296,48 @@ cellsep = try $ do
...
+  let width = case lookup "width" attribs of
                     Just xs -> fromMaybe 0.0 $ parseWidth xs
                     Nothing -> 0.0
-  return ((align, width), bs)
+  let rowspan = RowSpan . fromMaybe 1 $
+                safeRead =<< lookup "rowspan" attribs
+  let colspan = ColSpan . fromMaybe 1 $
+                safeRead =<< lookup "colspan" attribs
+  let handledAttribs = ["align", "colspan", "rowspan"]
+      attribs' = foldr go [] attribs
```
tarleb, line 335: I had missed this the first time: I think this could be shortened to ``filter ((`notElem` handledAttribs) . fst) attribs``. Or, less compact but more readable, list comprehension:

``` haskell
attribs' = [ (k, v) | (k, v) <- attribs
                    , k `notElem` handledAttribs
           ]
```

I'll make a change after merging.

tarleb:
Thanks again!!

src/Text/Pandoc/Readers/MediaWiki.hs
liruqi: Thanks. Acutally I referred to [HTML table reader](https://github.com/jgm/pandoc/blob/master/src/Text/Pandoc/Readers/HTML/Table.hs) here, and this is a simplied version to filter attributes.

tarleb: Ah, that makes sense :+1:

PR #8237, opened 2022-08-21
HTML writer: revert to using `width` property for column widths

tarleb:
The default `flex` and `overflow-x` properties of a column are set to
`auto`. In combination, these changes allow to get good results when
using columns with or without explicit widths.

### DIFF
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -164,7 +164,7 @@ $endif$
 code{white-space: pre-wrap;}
 span.smallcaps{font-variant: small-caps;}
 div.columns{display: flex; gap: min(4vw, 1.5em);}
-div.column{flex: 1;}
+div.column{flex: auto; overflow-x: auto;}
 div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
 ul.task-list{list-style: none;}
 ul.task-list li input[type="checkbox"] {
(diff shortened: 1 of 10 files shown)


tarleb:
See quarto-dev/quarto-cli#1940 for the motivation behind this change. It also makes it easy to revert to the previous (pre 2.19) behavior by including this additional CSS:

``` css
/* the `body` prefix is used to make the rule more specific than the default */
body div.column{display: inline-block; vertical-align: top; width: 50%;}
body div.columns{display: block;}
```

Cc @mb21

jgm:
Based on my limited knowledge, this seems fine to me.
@tarleb, I assume you've confirmed this works in the major browsers?
@jjallaire do you have comments on this proposal?

tarleb:
Tested with all browsers I have around, i.e., Firefox, Chrome, and Chromium.

A word on the `overflow-x: auto` change: this isn't required to get the old non-flex behavior, but it gives much nicer output when the column contains a code block. Otherwise the content of the left column can run into and overlay the right column. Previous versions of reveal.js used `overflow-x: auto` for source code elements, so this didn't happen with old reveal.js/pandoc versions.

jjallaire:
Yes, this does indeed LGTM!

tarleb:
Great, thanks! I'm merging this then.

PR #8246, opened 2022-08-24
Fix link to fedora package site.

mmzx:
The current link on @master and on the _pandoc.org_ webpage brings to the site via a redirect: 
https://packages.fedoraproject.org//pandoc
which returns with a 404 site.

This PR fixes the link. I am not sure whether the pandoc.org site uses this very _INSTALL.md_ file or another one which I did not find; if so, please let me know so I could extend the current PR with updating the link there as well.


tarleb:
Thanks!

Issue #8195, opened 2022-07-22
Multiple --csl arguments give bad error message


mgajda:
**Explain the problem.**

If using `pandoc` within a build pipeline, one may mistakenly give two `--csl` arguments:

```
pandoc -s input.md --csl style1.csl --csl style2.csl -o output.pdf 
```

This will give a spurious error message when `pandoc` tries to find `style1.csl style2.csl` which is not a file.

The issue stems from the fact that [`Text.Pandoc.App.CommandLineOptions.addMeta`](https://github.com/jgm/pandoc/blob/e04ea99326261c1fa935971a864477ce4e4182b1/src/Text/Pandoc/App/CommandLineOptions.hs#L1090) tries to append data to the previous meta.
Unfortunately, [`--csl` option does not treat the input as a list of styles, instead merges it](https://github.com/jgm/pandoc/blob/e04ea99326261c1fa935971a864477ce4e4182b1/src/Text/Pandoc/App/CommandLineOptions.hs#L734).

Would it be more convenient to use separate `replaceMeta` (possibly with a warning when previous metadata was non-empty), or change parsing of meta only for CSL?

Currently [`Text.Pandoc.Citeproc.getStyle`](https://github.com/jgm/pandoc/blob/e04ea99326261c1fa935971a864477ce4e4182b1/src/Text/Pandoc/Citeproc.hs#L123) simply uses [`metaValueToText`]() before trying to read it as a full path.

**Pandoc version?**

```
$ pandoc --version
pandoc 2.14.0.3
Compiled with pandoc-types 1.22, texmath 0.12.3.1, skylighting 0.10.5.2,
citeproc 0.4.0.1, ipynb 0.1.0.1
User data directory: /home/m/.local/share/pandoc
Copyright (C) 2006-2021 John MacFarlane. Web:  https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.
```
This is Linux version.

mgajda:
The issue stems from the fact that `Text.Pandoc.App.CommandLineOptions.addMeta` tries to append data to the previous meta.
Unfortunately, `--csl` option does not treat the input as a list of styles, instead merges it.

jgm:
> Would it be more convenient to use separate replaceMeta (possibly with a warning when previous metadata was non-empty), or change parsing of meta only for CSL?

Yes, I think that's a good idea.  If you specify two `--csl` options, one should be ignored, with a warning.
[...]

PR #8252, opened 2022-08-27
Ignore unused CSL style with warning

pt2121:
re #8195 

WIP - I am unsure about the warning message.

### DIFF
--- a/src/Text/Pandoc/App/CommandLineOptions.hs
+++ b/src/Text/Pandoc/App/CommandLineOptions.hs
@@ -35,7 +35,7 @@ import Data.List (intercalate, sort, foldl')
 #ifdef _WINDOWS
 import Data.List (isPrefixOf)
 #endif
-import Data.Maybe (fromMaybe, isJust)
+import Data.Maybe (fromMaybe, isJust, fromJust)
 import Data.Text (Text)
 import HsLua (Exception, getglobal, openlibs, peek, run, top)
 import Safe (tailDef)
@@ -69,6 +69,7 @@ import qualified Data.ByteString.Lazy as B
 import qualified Data.Map as M
 import qualified Data.Text as T
 import qualified Text.Pandoc.UTF8 as UTF8
+import Text.Pandoc.Citeproc.MetaValue (metaValueToText)
 
 #ifdef NIGHTLY
 import qualified Language.Haskell.TH as TH
@@ -727,10 +728,14 @@ options =
 
      , Option "" ["csl"]
                  (ReqArg
-                  (\arg opt ->
-                     return opt{ optMetadata =
-                                   addMeta "csl" (normalizePath arg) $
-                                   optMetadata opt })
+                  (\arg opt -> do
+                     let (replaced, meta) = replaceMeta "csl" (normalizePath arg) $
+                                   optMetadata opt
+                     when (isJust replaced) $
+                      runIOorExplode $ report $ DuplicateMeta $
+                          "Mulitiple CSL styles provided. The last one is used"
+                            <> maybe "" (" - ignored " <>) (metaValueToText $ fromJust replaced)
+                     return opt{ optMetadata = meta})
                    "FILE")
                  ""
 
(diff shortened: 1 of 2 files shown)


jgm:
As for the message, perhaps it makes sense to mention the one that is being ignored, but NOT the other one. After all, we might encounter a later option that overrides that one!  It would be undesirable to get messages like:
Using A instead of B
Using C instead of B
I think better just something like:
Only one CSL file can be specified.  Ignoring old.csl.

Finally, instead of creating a new cosntructor for PandocError, why don't we just use PandocOptionError, which is what we use for other errors involving command-line options?

pt2121:
Thanks for the pointers. I'll look into them :+1:

pt2121:
@jgm
I updated the message:

```
> pandoc -s AUTHORS.md --csl style1.csl --csl style2.csl --csl style3.csl -o output.pdf
[WARNING] Only one CSL file can be specified. Ignoring style1.csl
[WARNING] Only one CSL file can be specified. Ignoring style2.csl
```
Does ^ look good to you?

About `PandocOptionError`, I think we only want to warn the user, not throw exception? I use the `report` function that we have. What do you think?

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -727,10 +728,14 @@ options =
...
-                                   addMeta "csl" (normalizePath arg) $
-                                   optMetadata opt })
+                  (\arg opt -> do
+                     let (replaced, meta) = replaceMeta "csl" (normalizePath arg) $
+                                   optMetadata opt
+                     when (isJust replaced) $
+                      runIOorExplode $ report $ DuplicateMeta $
+                          "Only one CSL file can be specified."
+                            <> maybe "" (" Ignoring " <>) (metaValueToText $ fromJust replaced)
+                     return opt{ optMetadata = meta})
```
jgm, line 738: This use of isJust and fromJust is better avoided (since fromJust is a partial function).
Better:
```
case replaced of
  Nothing -> ...
  Just f -> ...
```

jgm:
> About PandocOptionError, I think we only want to warn the user, not throw exception? I use the report function that we have. What do you think?

Oh right.  Well, I wonder whether an error might make more sense in this context?  I'm really not sure.  I suppose we could raise the question on the original issue.  I'd be okay with a warning, though.

As for the code, I would prefer to have something more straightforward that doesn't use `replaceMeta` (which I predict will confuse me some time in the future).  Check to see if `css` already has a value; if it does, issue the warning that it's being ignored.  Then (unconditionally) set the new value.

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -727,9 +728,11 @@ options =
...
      , Option "" ["csl"]
                  (ReqArg
-                  (\arg opt ->
-                     return opt{ optMetadata =
-                                   addMeta "csl" (normalizePath arg) $
+                  (\arg opt -> do
+                    case lookupMeta (T.pack "csl") $ optMetadata opt of
+                      Just meta -> E.throwIO $ PandocOptionError $
+                         "Only one CSL file can be specified." <> maybe "" (" Ignoring " <>) (metaValueToText meta)
+                      Nothing -> return opt{ optMetadata = addMeta "csl" (normalizePath arg) $
```
jgm, line 735: I think that if this becomes an error instead of a warning, then the "Ignoring ..." message is just going to be misleading.  (After all, the program stopped; it's ignoring both --csl arguments.)

So, I'd say, either we go with a warning and say what we're ignoring, or go with an error and just say "Only one CSL file can be specified." The user will figure it out from there.

pt2121: Thanks, good point. I went with the error and fixed the message.

jgm: ok, thanks.

jgm:
Thank you, I've merged this!  The CI failure on Windows was an unrelated issue.

PR #8261, opened 2022-08-30
[RFC] Docx writer: allow to supply a template for `word/document.xml`

tarleb:
It would be nice if we had Word templates. This PR is the result of an afternoon experiment and a crude attempt to make docx generation easier to customize.

Pandoc checks for a file `templates/docx/document.xml` and uses it as template if it's present.

Pros:
  - simple,
  - based on plain text files,
  - uses the familiar template syntax.

Cons:
  - Completely separate from `reference.docx`,
  - files in nested directories below `data/templates`,
  - unclear how this would be used with `--template` command line option.

Comments welcome.

### DIFF
--- a/data/templates/docx/document.xml
+++ b/data/templates/docx/document.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"
+xmlns:m="http://schemas.openxmlformats.org/officeDocument/2006/math"
+xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships"
+xmlns:o="urn:schemas-microsoft-com:office:office"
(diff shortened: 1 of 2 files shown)


tarleb:
Custom bytestring writers with a pandoc.zip package seem like a much better way forward. Closing in favor of #8275 (unfinished).

iandol:
@tarleb — I assume I am not understanding correctly, but it seems this closed solution and your favoured `pandoc.zip` are not equivalent. There remains a desire to enable templating for DOCX output, and moving this into a Lua filter means someone needs to learn Lua, which is more involved than the existing simple templating system used for text outputs? EPubs are well customised with CSS already, but DOCX, a format which sadly remains dominant in many fields, remains a challenge. 

Thank you for thinking about this working on solutions!

iandol:
OK. Thank you for you excellent contributions to the Pandoc community!

PR #8275, opened 2022-09-03
Lua: add new module `pandoc.zip`

tarleb:
Allows to handle docx and epub files.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -5281,3 +5281,148 @@ Parameters:
 Returns:
 
 -   A new [Version] object.
+
+# Module pandoc.zip
+
+Functions to create, modify, and extract files from zip archives.
+
+The module can be called as a function, in which case it behaves
+like the `zip` function described below.
+
+Zip options are optional; when defined, they must be a table with
+any of the following keys:
+
+  - `recursive`: recurse directories when set to `true`;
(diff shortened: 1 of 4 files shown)


tarleb:
Todo:

- [ ] Documentation
- [ ] Better module interface

The main use-case that I see for this module is in custom writers, to modify docx and epub files that were generated with `pandoc.write`. However, the module is probably not that useful yet, as custom writers are `TextWriter`s, not `ByteStringWriter`s, so the modified archive cannot be the result of a `Writer` function.

jgm:
What's the status of this?  Is it something we should aim to include in 3.0?
What is holding it up?

tarleb:
I based it on the next HsLua version and haven't backported it yet. Should happen later this week.

tarleb:
Good to go, I think.

Example use to write just the main body of a docx:

``` lua
local zip = require 'pandoc.zip'

function Writer (doc, opts)
  local docx = pandoc.write(doc, 'docx', opts)
  local archive = zip.Archive(docx)
  local document_entry = archive.entries:find_if(
    function (entry) return entry.path == 'word/document.xml' end
  )
  local body = document_entry:contents()
    :gsub('.*<w:body>(.*)<w:sectPr ?%/></w:body>.*', '%1')
  return body
end
```

PR #8277, opened 2022-09-04
T.P.Mime: Base module on package mime-types.

tarleb:
The *mime-types* module is a dependency of *http-client* and therefore
already a transitive dependency of pandoc.


tarleb:
The purpose of the *mime-types* package is very close to what is done in `Text.Pandoc.MIME`. However, there are some differences in the list of MIME types. E.g., while the list in *mime-types* has entries for extensions like `.json` and `.epub`, it is missing some TeX-related entries like `.bib`. Also, files with a `.c` extension are currently given the type `text/x-csrc`, they'd get `text/plain` in the future. I'm not sure how relevant these changes are for most users, who, I assume, mostly care about *font*, *image* and *video* types.

tarleb:
I'm keeping this as a draft until I've confirmed that this change does not reintroduce pandoc-specific problems that had already been fixed.

tarleb:
I've changed the PR to keep all MIME entries that would otherwise change. So the only difference for users will be that more file extensions are recognized.

PR #8282, opened 2022-09-07
Add prefixes to identifiers with `--file-scope`.

### LINKED ISSUE #6384: Broken links between multiple input Markdown files
We convert our documentation from Markdown to PDF using Pandoc. Usually several Markdown files are converted to one PDF file.

Links between Markdown files that are included in the same PDF are broken in the PDF.

A very similar bug report already existed in 2016 with #2719.

## Minimal example

File `test-1.md`:

```markdown
# Headline 1

some text

## Headline 2

more text
```

File `test-2.md`:

```markdown
# Headline 1

some other text

## Another headline

more other text

[link to #headline-1](#headline-1) **wrong link**
refers to Headĺine 1 in test-1, shoud refer to Headĺine 1 in test-2

[link to #another-headline](#another-headline) **works**
as expected because anchor is unique

[link to test-1.md](test-1.md) **broken link**

[link to test-1.md#headline-1](test-1.md#headline-1) **broken link**

[link to test-1.md#headline-2](test-1.md#headline-2) **broken link**

[link to test-2.md](test-2.md) **broken link**

[link to test-2.md#headline-1](test-2.md#headline-1) **broken link**

[link to test-2.md#another-headline](test-2.md#another-headline) **broken link**
```

Compile it to PDF:

```bash
docker run --rm --volume "$(pwd):/data" --user $(id -u):$(id -g) pandoc/latex:2.9.2.1 test-1.md test-2.md -o test.pdf
```

You get the same behavior with HTML which is simpler to debug:

```bash
docker run --rm --volume "$(pwd):/data" --user $(id -u):$(id -g) pandoc/latex:2.9.2.1 test-1.md test-2.md -o test.html
```

Here is the HTML output:

```html
<h1 id="headline-1">Headline 1</h1>
<p>some text</p>
<h2 id="headline-2">Headline 2</h2>
<p>more text</p>
<h1 id="headline-1-1">Headline 1</h1>
<p>some other text</p>
<h2 id="another-headline">Another headline</h2>
<p>more other text</p>
<p><a href="#headline-1">link to #headline-1</a> <strong>wrong link</strong> refers to Headĺine 1 in test-1, shoud refer to Headĺine 1 in test-2</p>
<p><a href="#another-headline">link to #another-headline</a> works as expected because anchor is unique</p>
<p><a href="test-1.md">link to test-1.md</a> <strong>broken link</strong></p>
<p><a href="test-1.md#headline-1">link to test-1.md#headline-1</a> <strong>broken link</strong></p>
<p><a href="test-1.md#headline-2">link to test-1.md#headline-2</a> <strong>broken link</strong></p>
<p><a href="test-2.md">link to test-2.md</a> <strong>broken link</strong></p>
<p><a href="test-2.md#headline-1">link to test-2.md#headline-1</a> <strong>broken link</strong></p>
<p><a href="test-2.md#another-headline">link to test-2.md#another-headline</a> <strong>broken link</strong></p>
```

## Expected behavior

1. links to anchors (without file name) should also refer in the converted file to the correct headline from the original Markdown file and not to the first identical headline from all Markdown files
2. links to files with anchors that belong to the list of markdown files passed to Pandoc should link to the corresponding anchor of the converted file and not to the original file itself
3. links to files without anchors that belong to the list of markdown files passed to Pandoc are a bit more difficult to resolve
    1. If the markdown file referred to starts with a headline (of any order), it should be linked to this headline
    2. Otherwise, an additional anchor must be inserted at this point in the target document to which the link is then made (alternatively, such links could be removed pragmatically, since such cases are likely to be very rare, but the first proposed solution would be preferable if it can be implemented, since very rare is not a never)

jgm:
This change only affects the case where `--file-scope` is used and more than one file is specified on the command line.

In this case, identifiers will be prefixed with a string derived from the file path, to disambiguate them. For example, an identifier `foo` in `contents/file1.txt` will become `contents__file1.txt__foo`.  Links will be adjusted accordingly: if `file2.txt` links to `file1.txt#foo`, then the link will be changed to point to `#file1.txt__foo`.  Similarly, a link to `file1.txt` will point to `#file1.txt`.  A Div with an identifier derived from the file path will be added around each file's content, so that links to files will still work.

Closes #6384.

[API change]: Text.Pandoc.Shared exports `textToIdentifier`.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -564,6 +564,17 @@ header when requesting a document from a URL:
     footnotes and links will not work across files. Reading binary
     files (docx, odt, epub) implies `--file-scope`.
 
+    If two or more files are processed using `--file-scope`,
+    prefixes based on the filenames will be added to identifiers
+    in order to disambiguate them, and internal links will
+    be adjusted accordingly.  For example, a header with
+    identifier `foo` in `subdir/file1.txt` will have its
+    identifier changed to `subdir__file1.txt__foo`.
+
+    In addition, a Div with an identifier based on the filename
+    will be added around the file's content, so that internal
+    links to the filename will point to this Div's identifier.
+
 `-F` *PROGRAM*, `--filter=`*PROGRAM*
 
 :   Specify an executable to be used as a filter transforming the
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -216,6 +216,8 @@ extra-source-files:
                  test/command/B.txt
                  test/command/C.txt
                  test/command/D.txt
+                 test/command/file1.txt
+                 test/command/file2.txt
                  test/command/three.txt
                  test/command/01.csv
                  test/command/chap1/spider.png
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -49,6 +49,7 @@ import System.FilePath ( takeBaseName, takeExtension)
 import System.IO (nativeNewline, stdout)
 import qualified System.IO as IO (Newline (..))
 import Text.Pandoc
+import Text.Pandoc.Walk (walk)
 import Text.Pandoc.Builder (setMeta)
 import Text.Pandoc.MediaBag (mediaItems)
 import Text.Pandoc.MIME (getCharset, MimeType)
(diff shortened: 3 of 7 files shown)


bebuch:
I have tested this on one of my biggest real world markdown to HTML projects. Works great!

jgm:
@jkr @tarleb any thoughts about this change?

tarleb:
I'm in favor. :+1:

kanban-heretic:
I think I must be misunderstanding how this is supposed to work. From what I can tell, a link between input files like this:
```
[link](inputfile1.md#foo)
```
 or 
```
[link](../inputs/inputfile1.md#foo)
```

Should work as an internal link in the output file. But when I run the following, the link just looks for the external file. I'm running version 2.19.2.

```
pandoc \
-s \
--file-scope \
inputfile1.md \
inputfile2.md \
-o ../outputs/newfile.html
```
Am I doing this wrong? I'm not 100% clear on whether this change is intended to convert external links to internal links, or just to solve the problem of possible duplicate headers. Perhaps I'm formatting my links incorrectly?

Apologies if this isn't the appropriate place for this question.

jgm:
I don't think the change made it into 2.19.2. Try a nightly.

kanban-heretic:
Thanks. I don't think a nightly will work for me, but I wrote a simple lua script that discards everything before the # in a link target. Not particularly robust but I think it will work until the next release.

PR #8298, opened 2022-09-13
Defines a command in the LaTeX template for Level 6 Markdown headers

### LINKED ISSUE #8069: Markdown to LaTeX: Level 6 Heading
As far as I understand Markdown defines headings upto 6 levels, to correspond to ```h1```-```h6``` in HTML.

When Pandoc is converting a level 5 heading it is converted to ```\subparagraph{}```, while a level 6 heading is just converted to plain text.

I think it would be better to have a level 6 heading wrapped in a ```\subsubparagraph{}``` macro, while having something like ```\newcommand\subsubparagraph[1]{#1}``` in __default.latex__. That way it would preserve the previous behavior, while also making it possible to write templates that modify the behavior of level 6 headings. This would help people like me who are writing templates for converting markdown to pdf.

dvnrussell:
Adds a "\subsubparagraph" command definition to the LaTeX template. This command defaults to the current behavior, which compiles level 6 headers (tagged h6 in HTML) to plain text.

Closes #8069.

A potential concern for this change: the command is defined in default.latex, but that's only helpful when the standalone option is used. I am new to this codebase, so I'm not sure if the LaTeX writer has access to the standalone condition. Actually, this is my first pull request on an open source project, so I would love any feedback or advice.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -277,6 +277,7 @@ $endif$
 \setlength{\emergencystretch}{3em} % prevent overfull lines
 \providecommand{\tightlist}{%
   \setlength{\itemsep}{0pt}\setlength{\parskip}{0pt}}
+\providecommand{\subsubparagraph}{} % template for Level 6 Markdown headers
 $if(numbersections)$
 \setcounter{secnumdepth}{$if(secnumdepth)$$secnumdepth$$else$5$endif$}
 $else$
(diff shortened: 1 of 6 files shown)


jgm:
> A potential concern for this change: the command is defined in default.latex, but that's only helpful when the standalone option is used.

That's exactly why I've hesitated on this before.

> I am new to this codebase, so I'm not sure if the LaTeX writer has access to the standalone condition. 

Yes, it does. (Check for a Just value in writerTemplate in WriterOptions).
In the LW monad this is in state, so
```
  isStandalone <- gets (isJust . writerTemplate . stOptions)
```
will get it for you.

However, I'm not sure I'd want to make the behavior vary depending on whether a template is used.  That could be confusing, I think.

Your PR looks good.  I just need to think more about whether it's too much of a drawback that fragments produced after this change may not compile without special macro definitions.

PR #8309, opened 2022-09-20
Split off pandoc-cli and pandoc-server

jgm:
I'm putting this up for comment @tarleb.
If you like the Lua changes, we could merge this part separately before deciding on the package-splitting.


tarleb:
I definitely like the Lua part, it ensures a much cleaner separation!

The only comments I have are about `Setup.hs` and `COPYING.md`: the former can be omitted needed when using `build-type: Simple`; the latter could be a symlink, as cabal is smart enough to use the linked file when running `cabal sdist`.

Looks really good to me.

jgm:
I've merged the Lua changes into master, so this PR now only contains the splitting off of pandoc-cli and pandoc-server.

One issue to consider is whether the pandoc man page (or perhaps a symlink to it) should go in pandoc-cli.  Maybe the pandoc-server man page should go there too?  But then, there's also a case to be made for including teh pandoc-server man page in pandoc-server.

PR #8311, opened 2022-09-22
Make pandoc behave like a Lua interpreter when called as `pandoc-lua`.

tarleb:
A program name of `pandoc-lua`, e.g., when the binary was invoked through a symlink with that name, will make pandoc behave like a Lua interpreter: the first command line argument is run as a Lua script, and the remaining arguments are passed to the interpreter in the global table `arg`. Lua options are not supported yet.

### DIFF
--- a/doc/pandoc-lua.md
+++ b/doc/pandoc-lua.md
@@ -11,7 +11,7 @@ date: September 22, 2022
 # DESCRIPTION
 
 `pandoc-lua` is a standalone Lua interpreter with behavior similar
-to that of the standard `lua` executable, but exposes all of
+to that of the standard `lua` executable, but exposing all of
 pandoc's Lua libraries. All `pandoc.*` packages, as well as the
 packages `re` and `lpeg`, are available via global variables.
 Furthermore, the globals `PANDOC_VERSION`, `PANDOC_STATE`, and
--- a/doc/pandoc-lua.md
+++ b/doc/pandoc-lua.md
@@ -1,0 +1,50 @@
+---
+title: pandoc-lua
+section: 1
+date: September 22, 2022
+---
+
+# SYNOPSIS
+
+`pandoc-lua` [*options*] [*script* [*args*]]
+
+# DESCRIPTION
+
+`pandoc-lua` is a standalone Lua interpreter with behavior similar
+to that of the standard `lua` executable, but exposes all of
+pandoc's Lua libraries. All `pandoc.*` packages, as well as the
+packages `re` and `lpeg`, are available via global variables.
+Furthermore, the globals `PANDOC_VERSION`, `PANDOC_STATE`, and
+`PANDOC_API_VERSION` are set at startup.
+
+If no script argument is given, then the script is assumed to be
+passed in via *stdin*. Interactive mode is not supported at this
+time.
+
+# OPTIONS
+
+`-e stat`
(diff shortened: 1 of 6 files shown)


tarleb:
Todo:

- [ ] docs
- [ ] possibly add support for more CLI options
- [ ] medium term: extract the `runScript` function into an HsLua package
- [ ] medium term: ensure signals like SIGTERM are handled in the way that the `lua` binary handles them.

jgm:
Nice. Is an interactive REPL possible?

jgm:
If we had a REPL, it might be worth adding some fancy pretty-printing of the userdata Haskell values, similar to https://github.com/wlupton/pandoc-lua-logging

tarleb:
I'd like to do a REPL-less version first to avoid the need for an additional library (like [haskeline](https://hackage.haskell.org/package/haskeline). But I'd like to add one if the extra dependencies turn out to be small.

jgm:
I think you'd add the following dependencies with haskeline:
Win32 nats unbuildable stm semigroups
That's not huge.  And assuming we split the package, these would only be dependencies of pandoc-cli, not pandoc library.

src/Text/Pandoc/Lua.hs
```diff
@@ -13,15 +15,49 @@ module Text.Pandoc.Lua
...
+import qualified Lua.Constants as Lua
+import qualified HsLua as Lua
+import qualified HsLua.Core.Types as Lua
+
+-- | Uses the first command line argument as the name of a script file
+-- and tries to run that script in Lua. Falls back to stdin if no file
+-- is given. Any remaining args are passed to Lua via the global table
+-- @arg@.
+runScripts :: IO ()
+runScripts = do
```
jgm, line 47: I wonder why this is called `runScripts` when it runs exactly one script.
`runScript` might make that clearer (otherwise people might think they can supply multiple scripts as arguments).

tarleb: Oh, right. That's from an earlier version, where it did in fact run multiple scripts.

tarleb:
The current state is quite usable already and ready to merge from my side.

I plan to write a "hslua-cli" package, with a proper REPL, in the not-too-distant future. But I'd like to finish some other things first (e.g., Markdown tables).

jgm:
Can you add a `doc/pandoc-lua.md` and adjust Makefile so it can be compiled to `man/pandoc-lua.1` just as we now handle `doc/pandoc-server.md`?

tarleb:
Done. :+1: 

BTW, to avoid complications, I'd like to limit support of the *hslua-cli* package to Lua 5.4. This would mean that we'd need to drop support for the `lua53` flag at that point.

doc/pandoc-lua.md
```diff
@@ -0,0 +1,50 @@
...
+---
+
+# SYNOPSIS
+
+`pandoc-lua` [*options*] [*script* [*args*]]
+
+# DESCRIPTION
+
+`pandoc-lua` is a standalone Lua interpreter with behavior similar
+to that of the standard `lua` executable, but exposes all of
```
jgm, line 14: exposes -> exposing

tarleb:
I've extracted most of the code to the new package *hslua-cli* and fixed some more issues along the way. I'm dabbling with haskeline and may try isocline as well.

PR #8312, opened 2022-09-22
Manual: Added short para about producing accessible PDFs

the-solipsist:
This is based on a a finding by Ignacio Jonatan Hernández López:
https://groups.google.com/g/pandoc-discuss/c/2lO357t5yQg/m/WJSfBEFvAwAJ

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1353,6 +1353,11 @@ header when requesting a document from a URL:
         see [print-css.rocks](https://print-css.rocks) for a good
         introduction to PDF generation from HTML/CSS)
     - `-t ms`:  `pdfroff`
+    
+    Note that LaTeX engines have difficulty producing structured PDFs
+    that are accessible for screen readers. If you need to produce
+    accessible PDFs, consider using `--pdf-engine=prince` with
+    `--pdf-engine-opt=--tagged-pdf`.
 
 `--pdf-engine-opt=`*STRING*
 


jgm:
Don't know if I want to recommend prince (not open-source, and rather expensive).
Is that the only alternative?
#7885 is a path towards getting good tagging support via ConTeXt.
What about things like pagedjs-cli, weasyprint, wkhtmltopdf?

the-solipsist:
Fair points. I haven't myself compared the outputs of pagedjs-cli, weasyprint, wkhtmltopdf, etc. I'll check with the participants on that thread, and try with some samples using the [Tingtun checker](https://checkers.eiii.eu/en/pdfcheck/).  But I think that will take a little bit of time. In the meanwhile should I close this pull-request or should I leave it open since eventually _some_ guidance ought to be added about accessible PDFs?

jgm:
You can leave it open, thanks.

jgm:
@tarleb I think we should add something about accessible PDFs, perhaps as a separate section after "reproducible builds."  The question is what advice to put there.  One route is prince, as noted above.  Another is `-f context+tagging` -- but in the docs you added for that, you specify that `pdfa` variable should be set, too. Is that necessary for accessibility? It seems to add a lot of additional requriements about colors etc., and maybe we'd need to give more instruction on this.  Also worth exploring:  create docx and then produce a PDF from Word (would this be accessible)?  Via HTML using weasyprint or pagedjs-cli or wkhtmltopdf?

PR #8316, opened 2022-09-23
latex template: set fonts after Beamer theme

### LINKED ISSUE #6260: latex template sets fonts before applying beamer theme
Using pandoc 2.5
Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.7.7
In an ubuntu/eoan container environment

`metadata.yaml`:
```yaml
---
theme: metropolis
monofont: JetBrainsMono
...
```

`cmdline`:
`pandoc -s -f markdown -t beamer --pdf-engine=lualatex -V classoption=notes --fail-if-warnings --metadata-file=metadata.yaml -o slides.pdf slides.md`

Results in a PDF that uses `FiraMono` as metropolis sets defaults for fonts (`FiraSans` and `FiraMono`), which is unexpected to say the least.

WIth a custom template I confirmed, that re-ordering the blocks that apply beamer themes and the font block fixes this problem for me. I am not very familiar with latex and therefor am not aware of any new problems this might cause for other latex-engines.

Would you accept a PR changing the default latex template?

knuesel:
Beamer themes such as metropolis and saintpetersburg change the default fonts. This PR gives precedence to the user font settings by moving them after the loading of the Beamer theme.

This fixes #6260

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -82,11 +82,6 @@ $if(beamerarticle)$
 \usepackage{beamerarticle} % needs to be loaded first
 $endif$
 \usepackage{amsmath,amssymb}
-$if(fontfamily)$
-\usepackage[$for(fontfamilyoptions)$$fontfamilyoptions$$sep$,$endfor$]{$fontfamily$}
-$else$
-\usepackage{lmodern}
-$endif$
 $if(linestretch)$
 \usepackage{setspace}
 $endif$
@@ -98,15 +93,48 @@ $endif$
 \else % if luatex or xetex
 $if(mathspec)$
   \ifXeTeX
-    \usepackage{mathspec}
+    \usepackage{mathspec} % this also loads fontspec
   \else
-    \usepackage{unicode-math}
+    \usepackage{unicode-math} % this also loads fontspec
   \fi
 $else$
-  \usepackage{unicode-math}
+  \usepackage{unicode-math} % this also loads fontspec
 $endif$
-  \defaultfontfeatures{Scale=MatchLowercase}
+  \defaultfontfeatures{Scale=MatchLowercase}$-- must come before Beamer theme
   \defaultfontfeatures[\rmfamily]{Ligatures=TeX,Scale=1}
+\fi
+$if(fontfamily)$
+$else$
+$-- Set default font before Beamer theme so the theme can override it
+\usepackage{lmodern}
+$endif$
+$-- Set Beamer theme before user font settings so they can override theme
+$if(beamer)$
+$if(theme)$
+\usetheme[$for(themeoptions)$$themeoptions$$sep$,$endfor$]{$theme$}
+$endif$
+$if(colortheme)$
+\usecolortheme{$colortheme$}
+$endif$
+$if(fonttheme)$
+\usefonttheme{$fonttheme$}
+$endif$
+$if(mainfont)$
+\usefonttheme{serif} % use mainfont rather than sansfont for slide text
+$endif$
+$if(innertheme)$
+\useinnertheme{$innertheme$}
+$endif$
+$if(outertheme)$
+\useoutertheme{$outertheme$}
+$endif$
+$endif$
+$-- User font settings (must come after default font and Beamer theme)
+$if(fontfamily)$
+\usepackage[$for(fontfamilyoptions)$$fontfamilyoptions$$sep$,$endfor$]{$fontfamily$}
+$endif$
+\ifPDFTeX\else
+  % xetex/luatex font selection
 $if(mainfont)$
   \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$]{$mainfont$}
 $endif$
@@ -166,26 +194,6 @@ $if(zero-width-non-joiner)$
   \protected\def ^^^^200c{\zerowidthnonjoiner}
 \fi
 %% End of ZWNJ support
-$endif$
-$if(beamer)$
-$if(theme)$
-\usetheme[$for(themeoptions)$$themeoptions$$sep$,$endfor$]{$theme$}
-$endif$
-$if(colortheme)$
-\usecolortheme{$colortheme$}
-$endif$
-$if(fonttheme)$
-\usefonttheme{$fonttheme$}
-$endif$
-$if(mainfont)$
-\usefonttheme{serif} % use mainfont rather than sansfont for slide text
-$endif$
-$if(innertheme)$
-\useinnertheme{$innertheme$}
-$endif$
-$if(outertheme)$
-\useoutertheme{$outertheme$}
-$endif$
 $endif$
 % Use upquote if available, for straight quotes in verbatim environments
 \IfFileExists{upquote.sty}{\usepackage{upquote}}{}
--- a/test/lhs-test.latex
+++ b/test/lhs-test.latex
@@ -5,16 +5,19 @@
 \documentclass[
 ]{article}
 \usepackage{amsmath,amssymb}
-\usepackage{lmodern}
 \usepackage{iftex}
 \ifPDFTeX
   \usepackage[T1]{fontenc}
   \usepackage[utf8]{inputenc}
   \usepackage{textcomp} % provide euro and other symbols
 \else % if luatex or xetex
-  \usepackage{unicode-math}
+  \usepackage{unicode-math} % this also loads fontspec
   \defaultfontfeatures{Scale=MatchLowercase}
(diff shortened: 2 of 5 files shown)


jgm:
I'm slightly nervous about the way in which this breaks up the current font settings, moving some of them after the beamer theme options and leaving some where they are. That adds complexity. Wouldn't it be simpler just to move the block concerning beamer themes before all the font settings?  Or is there some reason that won't work?

knuesel:
I agree splitting the font-related code is unfortunate. I did it this way to minimize the risk of introducing problems or changing current behavior: Moving some `\setXXXfont` calls later is unlikely to cause an issue. On the other hand, moving the `\useXXXtheme` calls earlier (for example at the end of the previous beamer-related stuff) might be more impactful: some theme files load several packages, and I wanted to avoid changing the package loading order.

Now looking in more details at what the themes actually do, I don't see anything obvious that should cause troubles. I also just tested a template with beamer theme loading moved upwards together with other beamer commands, and was able to compile a few presentations without issue. Shall I amend this PR accordingly?

jgm:
Yes, I think it's desirable if possible to keep the font stuff together in the template.
If you could amend accordingly and test with a number of theme/font combinations, that would be great.

knuesel:
I used the following document:

    ---
    theme: metropolis
    monofont: JuliaMono
    ...
    
    # Title
    
    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
    tempor incididunt ut labore et dolore magna aliqua.
    
    ```
    ẋ, x⃗, x⃗̇, x͍⃡, x↑, θ, θ̂
    ```

and compared the output between these two versions:

1. Current pandoc release
2. A variant of this PR with the beamer theme logic all moved before the font logic: https://github.com/knuesel/pandoc/commit/0aca25914082601d757cea2b219a3dd82002ea8b

This uncovers at least two issues with the second version:

* The default font settings of the Metropolis theme are lost because `\usepackage{lmodern}` now comes later and overwrites them.
* Commenting `\usepackage{lmodern}` fixes the font family but the sizes are different from the current pandoc release, because the `\defaultfontfeatures{Scale=MatchLowercase}` now come too late to have an effect. (It's not clear to me if this change is a bug or is fixing previously wrong behavior).

Another possibilty is to move the Beamer logic after the main font package loading and defaults, but before the font setting code: https://github.com/knuesel/pandoc/commit/e4d6a95d0f1e06cf107a90060c71fce626915dc3 . Compared to the current PR, this leaves the loading of packages `xeCJK`, `luatexja-preset` and `luatexja-fontxpec` after the Beamer stuff, together with other CJK code, so it looks a bit less messy. This may be a good trade off?

jgm:
> Another possibilty is to move the Beamer logic after the main font package loading and defaults, but before the font setting code: https://github.com/knuesel/pandoc/commit/e4d6a95d0f1e06cf107a90060c71fce626915dc3 . Compared to the current PR, this leaves the loading of packages xeCJK, luatexja-preset and luatexja-fontxpec after the Beamer stuff, together with other CJK code, so it looks a bit less messy. I haven't found any issue with this version. Maybe it's a good trade off?

Thanks for investigating. That approach seems cleaner to me.  But it would be worth checking with e.g. a Chinese document, with lang set to zh-CN, so that CJK code will be used.

knuesel:
I tried compiling a Beamer presentation from https://github.com/mba811/dev-study/tree/master/Markdown with metadata

```
---
theme: metropolis
monofont: JuliaMono
lang: zh-CN
CJKmainfont: SimSun
...
```

and it works fine (with and without the proposed changes) with xelatex and lualatex. (With lualatex there's an unrelated error that looks like https://tex.stackexchange.com/questions/609104/lualatex-ctex-and-microtype-together-produce-errors-when-using-scshape; it works when disabling microtype.)

I've updated the PR to the new approach.

jgm:
To fix the tests, you can do `make TESTARGS=--accept`.
This will replace the expected test values with what your modified pandoc produces.
Just be sure to look over the changes to the tests and make sure everything is correct.

knuesel:
Ah thanks, the tests should pass now.

jgm:
That is odd.
Maybe we could make it less odd by doing it this way:
```
\ifPDFTeX
\else
% Font selection: after Beamer theme selection to override e.g. metropolis or
% saintpetersburg font selection
$if(mainfont)$
  \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$]{$mainfont$}
$endif$
```
Then it will look like
```
\ifPDFTeX
\else
% Font selection: after Beamer theme selection to override e.g. metropolis or
% saintpetersburg font selection
\fi
```
and it least it will be clear what this section is for.  I don't know, what do you think?

The other thing that's a bit odd is that fonts can be selected to override beamer theme fonts ONLY in xelatex/lualatex.  Why should that be?

knuesel:
Good idea, and I also see it's odd for the comment to mention a Beamer theme, when using regular LaTeX. I pushed some changes, so now in the worst case it looks like this:

```latex
\usepackage{amsmath,amssymb}
\usepackage{lmodern}
\usepackage{iftex}
\ifPDFTeX
  \usepackage[T1]{fontenc}
  \usepackage[utf8]{inputenc}
  \usepackage{textcomp} % provide euro and other symbols
\else % if luatex or xetex
  \usepackage{unicode-math} % this also loads fontspec
  \defaultfontfeatures[\rmfamily]{Ligatures=TeX,Scale=1}
\fi
\ifPDFTeX\else
  % xetex/luatex font selection
\fi
```

It still looks a bit odd to have the empty "if" block split from the rest, but that's already an improvement I think?

Regarding the second point: another good point... in the PR I'm only moving the fontspec commands `\setmainfont`, `\setsansfont`, etc. which correspond to the pandoc settings `mainfont`, etc. Pdflatex doesn't support these commands since it cannot use truetype/opentype fonts. Currently the template does this very early:

```
$if(fontfamily)$
\usepackage[$for(fontfamilyoptions)$$fontfamilyoptions$$sep$,$endfor$]{$fontfamily$}
$else$
\usepackage{lmodern}
$endif$
```

This applies to all engines. I think the `\lmodern` default must stay there, but it would make sense to move the `fontfamily` case later with the xetex/luatex commands, at the risk of changing the behavior for some users... I can make some tests if that approach sounds good to you.

jgm:
```
\fi
\ifPDFTeX\else
```
We could just put these lines under the `$if(beamer)$` so that they only appear in beamer output, with the result that the font selection is included in the regular `else` branch for standard tex output.  That is, just move the endif at line 131 down a couple lines...

jgm:
>  I think the \lmodern default must stay there, but it would make sense to move the fontfamily case later with the xetex/luatex commands, at the risk of changing the behavior for some users... I can make some tests if that approach sounds good to you.

This does sound more consistent, so please do.

knuesel:
> We could just put these lines under the `$if(beamer)$` so that they only appear in beamer output, with the result that the font selection is included in the regular `else` branch for standard tex output. That is, just move the endif at line 131 down a couple lines...

That would work well, but it's kind of incompatible with the other change (moving the `fontfamily` setting between the Beamer theme and the xetex/luatex font selection) because `fontfamily` applies to all engines so that's another need for interrupting the xetex/luatex `\if` block.

However I think it actually makes perfect sense to move the `lmodern` default closer to the other font selection but just before the Beamer theme, and this also helps with the useless blocks. I pushed a new version implementing this and better documenting the template code. Now in the worst case it looks like this:

```
\ifPDFTeX
  \usepackage[T1]{fontenc}
  \usepackage[utf8]{inputenc}
  \usepackage{textcomp} % provide euro and other symbols
\else % if luatex or xetex
  \usepackage{unicode-math} % this also loads fontspec
  \defaultfontfeatures[\rmfamily]{Ligatures=TeX,Scale=1}
\fi
\usepackage{lmodern}  % Beamer theme if defined would be set *after* this line
\ifPDFTeX\else
  % xetex/luatex font selection
\fi
```

or if `fontfamily` is defined:

```
\ifPDFTeX
  \usepackage[T1]{fontenc}
  \usepackage[utf8]{inputenc}
  \usepackage{textcomp} % provide euro and other symbols
\else % if luatex or xetex
  \usepackage{unicode-math} % this also loads fontspec
  \defaultfontfeatures[\rmfamily]{Ligatures=TeX,Scale=1}
\fi
\usepackage[]{times}  % Beamer theme if defined would be set *before* this line
\ifPDFTeX\else
  % xetex/luatex font selection
\fi
```

I find this finally starts to make sense... What do you think?

knuesel:
Thanks for the comments!

PR #8317, opened 2022-09-23
doc: use cabal's --package-env more

ulysses4ever:
I noticed that you already use it on the same page and thought that it'd make it more consistent (not to mention, more fail-safe).


jgm:
Thanks.

PR #8318, opened 2022-09-24
Extract Lua engine into a separate package

tarleb:


tarleb:
This makes it easy to build pandoc without Lua. The effect on the resulting binary is unexpectedly *huge*, shrinking the (unoptimized) binary by about 40%.

jgm:
I tried this on an M1 mac.  I saw the following difference in stripped binary sizes:
with lua: 154 MB
without lua: 151 MB
Also, in the build with lua, filters and custom writers don't seem to work; it just hangs when I try to use them, and I have to press ^C twice to stop.

Separate thought: maybe a ScriptingEngine should have  a textual description field, which we can include in the `pandoc --version` output (e.g. "compiled with Lua 5.4 scripting support").

Another separate thought: if we have a `lua` flag on pandoc-cli to disable Lua support, we should also have a `server` flag so taht server support can be disabled.

tarleb:
> I tried this on an M1 mac. I saw the following difference in stripped binary sizes:
> with lua: 154 MB
> without lua: 151 MB

That actually helps me sleep much easier. I thought the general idea of "almost free" functionality through Lua was fundamentally flawed. But that means that there is something strange happening on Linux, because here I get
with lua: 194 MB
without lua: 116 MB

The issue of pandoc hanging appears to be related to the mediabag: using `pandoc-lua` works fine unless the script uses `pandoc.mediabag.fetch`. Not sure yet what's going on there.

jgm:
> The issue of pandoc hanging appears to be related to the mediabag: using pandoc-lua works fine unless the script uses pandoc.mediabag.fetch. Not sure yet what's going on there.

I'm not talking about `pandoc-lua`.  I'm talking about using `pandoc` with `--lua-filter` or a custom reader or writer. This hangs (even if you provide the name of a nonexistent filter).

tarleb:
Oh, interesting. I assumed that both bugs are very much related, and fixing one would fix the other. But the fact that it hangs on a nonexistent filter makes me doubt that assumption.

I kind of hope that this is one of those bugs whose fix will improve the whole system. My first thought is "lazyness", but I was only half-right about that the last time.

jgm:
I assume you're not seeing the same thing on linux, then?
Maybe worth adding a few sanity checks to the CI (at least temporarily), so we can see how it works on all architectures/compiler versions.

jgm:
PS. I'll add a `server` flag in the main branch.

tarleb:
Now I'm seeing it on Linux, too. Presumably I tested that with a different version without realizing.

It seems that pandoc hangs in Text.Pandoc.Class.IO.readFileStrict, but not always. I'm still at a loss so far.

tarleb:
Ah, found it. I created a loop by replacing `Text.Pandoc.Class.IO` with `Text.Pandoc.Class` in the definition of PandocLua. :man_facepalming:

tarleb:
I tried to clean this up, all API changes are now in a separate commit. The last remaining point is that the tests for custom writer must be moved to the new package.

pandoc.cabal
```diff
@@ -560,6 +538,8 @@ library
                    Text.Pandoc.MediaBag,
                    Text.Pandoc.Error,
                    Text.Pandoc.Filter,
+                   Text.Pandoc.Filter.Environment,
+                   Text.Pandoc.Filter.JSON,
```
tarleb, line 542: Maybe the relevant functions should be exported by `Text.Pandoc.Filter`, so these submodules can stay internal?

jgm: Yes, I think that makes sense.

tarleb:
Good to go, I think.

jgm:
Are you still seeing the large difference in binary size with +lua and -lua?

jgm:
PS. If the gridtables package is ready to release, it would be good to depend on a released version rather than git.

tarleb:
I found that the difference is significant only when linking the Lua library dynamically: there's still a large difference in that case. Compiling a static binary yields a mere 2MB difference.

jgm:
That's surprising that the difference would be *more* with dynamic linking. You'd expect the opposite, wouldn't you?

By the way, I noticed that the command you implemented is `pandoc serve` rather than `pandoc server`.
Was there a reason for this?  I'm not sure, myself, which is better.  But `lua` is a noun, so perhaps that favors `server`.

tarleb:
Gridtables is updated now.

I used `serve` because of this comment: https://github.com/jgm/pandoc/pull/8322#issuecomment-1258439073. I figured it made sense, but now that I think about it, `server` is probably more consitent?

tarleb:
I changed it to `server`.

PR #8320, opened 2022-09-24
rST writer: list tables rendering, closes #4564

### LINKED ISSUE #4564: Option to use list-tables in rST writer.

danse:
When a table is marked with a "list-table" attribute class, it will now be rendered using the list table syntax documented here http://docutils.sourceforge.net/docs/ref/rst/directives.html#list-table

### DIFF
--- a/src/Text/Pandoc/Writers/RST.hs
+++ b/src/Text/Pandoc/Writers/RST.hs
@@ -311,7 +311,7 @@ blockToRST (CodeBlock (_,classes,kvs) str) = do
 blockToRST (BlockQuote blocks) = do
   contents <- blockListToRST blocks
   return $ nest 3 contents <> blankline
-blockToRST (Table _ blkCapt specs thead tbody tfoot) = do
+blockToRST (Table attrs blkCapt specs thead tbody tfoot) = do
   let (caption, aligns, widths, headers, rows) = toLegacyTable blkCapt specs thead tbody tfoot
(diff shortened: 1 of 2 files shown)


danse:
I added `"list-tables"` as an attribute class, since there is no need to have a value in this case.

I found no suitable place to add documentation about the new attribute, but if you think that it could go somewhere just let me know where and i will write it.

jgm:
Thanks for reviving the PR. However, I'm coming around to the old approach of just having a new command-line option.

The main advantage of this PR's approach (using attributes) is that some tables can be formatted as list tables and others not. However, nobody has really said that they need that.

At the same time, there are several disadvantages to using attributes:

- people won't be able to access this without using a special filter
- nor will they easily discover that it is even possible
- `list-table` refers to how a table is formatted in a particular output format, so it's a bit weird to have this in the AST at all; it seems like an abuse of attributes

So it seems to me that a `--list-tables` cli option might be the more straightforward approach.  This does have the drawback that it would require a change to WriterOptions (a potentially breaking API change), but I think we can live with that (especially as it's a new field, not a change to an old one).

jgm:
I'm just going to merge this whole PR and modify it to use an option instead of an attribute.

danse:
Hi, sorry for coming late on this. I am glad that we finally enabled this feature, and i hope that user will appreciate the easiness coming from the direction you have chosen! Cheers

PR #8322, opened 2022-09-26
Allow to control executable behavior with PANDOC_PROGRAM_NAME variable

tarleb:
If the variable is set, then this value will be used as the program name to decide the behavior of the pandoc executable. Currently supported values are `pandoc-server.cgi` for the CGI server, `pandoc-server` for the HTTP(S) server, `pandoc-lua` for the Lua interpreter. Any other value will trigger the usual behavior.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -7094,6 +7094,10 @@ exposing the same API as `pandoc-server`.
 Haskell's type system to provide strong guarantees that no I/O
 will be performed on the server during pandoc conversions.
 
+An alternative to renaming or symlinking the binary is to set the
+`PANDOC_PROGRAM_NAME` environment variable. Its value, if set,
+overrides the name of the executable.
+
 [pandoc-server]: https://github.com/jgm/pandoc/blob/master/doc/pandoc-server.md
 
 # Running pandoc as a Lua interpreter
@@ -7104,6 +7108,11 @@ mostly identical to that of the [standalone `lua`
 executable][lua standalone], version 5.4. However, there is no
 REPL yet, and the options `-W`, `-E`, and `-i` currently don't
 have any effect.
+
+The name of the executable can be changed by renaming, by creating
+a symlink with a new name, or by setting the `PANDOC_PROGRAM_NAME`
+environment variable. The latter, if set, overrides the name of
+the executable.
 
 [lua standalone]: https://www.lua.org/manual/5.4/manual.html#7
 
--- a/pandoc-cli/src/pandoc.hs
+++ b/pandoc-cli/src/pandoc.hs
@@ -36,7 +36,7 @@ import System.Exit (exitWith, ExitCode(ExitFailure))
 
 main :: IO ()
 main = E.handle (handleError . Left) $ do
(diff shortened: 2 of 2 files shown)


tarleb:
From what I understand, symlinks are problematic on Windows. So this would offer an alternative method to use pandoc as a server or Lua interpreter.

jgm:
Not sure: I'm a little nervous about having an environment variable change what the executable does in this way.
Doesn't Windows support symlinks now?  Did somebody try this and have it fail?  Perhaps the real issue is the '.exe' that would probably be returned by `getProgName`?
Even if not, renaming/copying would still be an option on Windows, wouldn't it?

jjallaire:
Here is the deal with symlinks on Windows: https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/

So they are supported in Windows 10 if you enable a special "developer mode". That might be a bit too much to ask of an end user though.

Our use case here is that we'd like to be able to run Lua scripts directly for Quarto pre/post processing hooks or to run unit test frameworks. The renaming trick would work fine for Windows but would just increase the installer size (although perhaps this would compress-away so not be a big deal after all?).

jgm:
I see.  Well, maybe this is a good solution.  Another approach could be to make pandoc sensitive to an optional special first command line option, so that `pandoc --server` starts in server mode and `pandoc --lua` in lua mode.  This would be a bit awkward, because these special command-line options wouldn't be included in the regular `--help` output, and couldn't be re-ordered.

tarleb:
At least for Lua it might make sense to allow `--lua` in any position. Any extra arguments would then not be treated as input files, but as arguments to the Lua CLI. Then `pandoc-lua ...` would be equivalent to `pandoc --lua -- ...`.

tarleb:
The advantage of that would be that users could set the user-dir and similar variables. This is not possible otherwise, as the PANDOC_STATE variable is read-only in Lua.

jgm:
Hm, maybe we could do the same for `--server`.  If it is present as ANY of the arguments, we could parse the arguments as server arguments.

Still, it's a bit awkward I think.  Another model might be:

`pandoc serve --port=5000`

`pandoc lua -e 'print("HI")'`

tarleb:
> Another model might be:
>
> `pandoc serve --port=5000`
>
> `pandoc lua -e 'print("HI")'`

I think that would be my personal favorite, despite the drawbacks. Maybe a `convert` command for the default behavior should be added for symmetry?

Git has separate manpages for `git`, `git-commit`, `git-add`, etc., which makes me think that a we might get away with small updates the MANUAL.txt, with docs for the other programs in separate files.

jgm:
Yes, I agree, that seems the best approach. I don't think we need to add an optional `convert`; it might just confuse matters.  And I'd favor not even mentioning these special modes in the main part of the manual.  In the places where we mention server and lua modes, we can say that they can be triggered this way, and refer people to the pandoc-server and pandoc-lua manuals.

tarleb:
I've opened #8327 for this.

jgm:
Closing in favor of #8327.

Issue #8331, opened 2022-09-29
Expose AuthorOnly as CitationMode


rgaiacs:
Academic writing has two ways to provide citations: author-prominent and information-prominent.

Example of author-prominent:

> Knuth et al. (1997) wrote a famous book.

Example of information-prominent:

> A popular programming language is R (R Core Team, 2020).

Use of one over the other is a matter of field and writing style.

[Citation Style Language](https://citationstyles.org/) 1.0.2 only support information-prominent, see https://github.com/jgm/citeproc/issues/69#issuecomment-849753083. This means that write using author-prominent format is annoying when using a numeric citation style (for example, IEEE) because name of authors need to be hard coded. For example, the Markdown

```
@knuth wrote a famous book.
```

will be render by Pandoc as

```
[1] wrote a famous book.
```

instead of

```
Knuth et al. (1997) wrote a famous book [1].
```

or similar variation and the user need to have in the Markdown

```
Knuth et al. (1997) wrote a famous book [@knuth].
```

As mentioned before, Citation Style Language doesn't support information-prominent and will be hard for Pandoc to support all different journal styles. One way to compromise is to provide Markdown syntax for author-only citations (already proposed in https://github.com/jgm/pandoc/issues/7258). Until https://github.com/jgm/pandoc/issues/7258 is resolved, would be great to facilitate the use of filters by Pandoc users. For example, Pandoc implements author-prominent "[by emitting `AuthorOnly` and `SuppressAuthor` in sequence](https://github.com/jgm/citeproc/issues/69#issue-903284123)", see https://github.com/jgm/pandoc/blob/751383f33370928c7576f17d1a85d48cc9e8e30e/src/Text/Pandoc/Citeproc.hs#L335-L344

As a user, I would love to replace my use of `AuthorInText` with `AuthorOnly`. For example, the Markdown

```
@knuth [-@knuth] wrote a famous book [@knuth].
```

could be converted to

```
Knuth et al. (1997) wrote a famous book [1].
```

by calling

```
$ pandoc \
    --from markdown \
    --citeproc \
    --bibliography=references.bib \
    --csl=ieee.csl \
    mwe.md \
    --to json | \
    sed 's/AuthorInText/AuthorOnly/g' | \
    pandoc \
    --from json \
    --citeproc \
    --bibliography=references.bib \
    --csl=ieee.csl \
    --to html
```

# Limitations

Based on my experiments, `AuthorOnly` is empty for Citation Style Language's numeric citation style.

[...]

PR #8332, opened 2022-09-29
Expose AuthorOnly as CitationMode

rgaiacs:
Related to https://github.com/jgm/pandoc/issues/8331.

### DIFF
--- a/src/Text/Pandoc/Citeproc.hs
+++ b/src/Text/Pandoc/Citeproc.hs
@@ -333,6 +333,8 @@ fromPandocCitations locale otherIdsMap = concatMap go
            then []
            else
              case citationMode c of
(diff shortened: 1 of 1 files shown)


tarleb:
Thanks! The problem is that this would require changes to pandoc's AST, defined in the [pandoc-types](https://github.com/jgm/pandoc-types) package. Changes to these data structures usually have far-reaching consequences.

Is the suggested change supported by CSL or an extension thereof?

rgaiacs:
@tarleb Thanks for the reply.

> Is the suggested change supported by CSL or an extension thereof

Not supported by CSL (see https://discourse.citationstyles.org/t/pandocs-author-in-text-citation/1775/2). It would be an extension. I will work more and return with a longer minimal working example.

tarleb:
I'm closing this for now; please *do* reopen if there is progress.

Issue #6215, opened 2020-03-26
Split off core package


tarleb:
I'm wondering whether it would make sense to split off parts of pandoc into a separate `pandoc-core` package. This would make it easier to move other parts into separate packages as well.

My motivation here is the Lua system. It is growing quite large, but, with the exception of the `pandoc.read` function, is built only on a small part of pandoc the library. The pandoc core (i.e., T.P.Class etc) as well as the Lua system are relatively stable, so the overhead of having additional packages to maintain seems acceptable.

In a similar vein: while writing jira-wiki-markup, I would have liked to have a `pandoc-parsing` library. Depending on such library would make it easy to ensure that library uses the same parser as pandoc. It could include some of the fixes and convenience functions available in `Text.Pandoc.Parsing`.

jgm:
I'm open to exploring this, but what exactly would you conceive as being the core modules?

tarleb:
I'd considere everything required to define `PandocMonad` as "core", so the modules

- BCP47
- Class
- Data
- Error
- Logging
- MediaBag
- MIME
- Process
- Translations
- UTF8

plus the function `uriPathToPath` from T.P.Shared.

Additionally maybe Emoji, UUID, and XML, although those would increase the dependency footprint.

jgm:
I'm still not sure I understand the motivation.  This would allow creation of a package `pandoc-lua` with the lua system. This package would depend on `pandoc-core`.  But what would depend on this package, besides pandoc itself?  IF the answer is nothing, then I'm not sure it's worth the hassle of splitting.

tarleb:
A main motivation for me is compile time: e.g., switching branches to fix a bug while working on Lua frequently causes recompilation of *all* pandoc modules. Finding a way to reduce compile times would remove a huge bottleneck from my workflow. Splitting of smaller modules seems like a good option to achieve this, and should also reduce the frequency with which I'd have to switch branches.

jgm:
> switching branches to fix a bug while working on Lua frequently causes recompilation of all pandoc modules

One way to deal with this kind of thing is to clone the branch in a separate directory.

alerque:
> > switching branches to fix a bug while working on Lua frequently causes recompilation of all pandoc modules
> 
> One way to deal with this kind of thing is to clone the branch in a separate directory.

Along these lines `git subtree` is quite useful for this too because it shares the object store with the main repo and so is fast and easy on the file system.

[...]

PR #8340, opened 2022-09-30
Extract package pandoc-core

tarleb:
Closes: #6215


tarleb:
I'm not sure about this. Having this would allow to write readers and writers without having to depend on the full pandoc lib. It could be helpful for some projects like pandoc-crossref, but I'm worried about having to maintain too many packages.

pandoc-core/pandoc-core.cabal
```diff
@@ -0,0 +1,127 @@
...
+                     , Text.Pandoc.Parsing.Capabilities
+                     , Text.Pandoc.Parsing.Citations
+                     , Text.Pandoc.Parsing.General
+                     , Text.Pandoc.Parsing.GridTable
+                     , Text.Pandoc.Parsing.Lists
+                     , Text.Pandoc.Parsing.Math
+                     , Text.Pandoc.Parsing.Smart
+                     , Text.Pandoc.Parsing.State
+                     , Text.Pandoc.Parsing.Types
+                     , Text.Pandoc.Readers.LaTeX.Types
```
jgm, line 119: Why this?  It seems LaTeX-specific and thus not a candidate for core, but I assume something else here depends on it?

jgm:
I'm leaning to the view that we should not do this; I don't see clear advantages to doing it that would outweigh the disadvantages.  The principle of the split -- what goes in core and what doesn't -- isn't conceptually clean.  E.g., XML goes one way, emojis go the other; citeproc goes one way, highlighting another; MIME goes one way, image size another -- and it's not clear why one of these is more "core" than another.  (I don't mean to say this selection is arbitrary -- presumably it's a kind of minimum core that can be cleanly extracted -- but that the principle of selection won't be obvious.

I'm open to persuasion though.

pandoc-core/pandoc-core.cabal
tarleb: `Macro` from that module is uesd by Text.Pandoc.Parsing.{State,Capabilities}. There's probably a better solution, this is all very experimental.

tarleb:
I had this code stashed away in my local repo for some time. My main objective back then was to have `Text.Pandoc.Parsing` in a separate package, the selection of modules was driven by that. I think I explored this while writing the jira package, but not sure.

I agree that the downsides out-weight the benefits. Closing...

Issue #6338, opened 2020-05-04
Missings Instructions/Feature to add sub-heading to "Section Header" slide layout in PPTX


seesharprun:
> *I'm not sure if this is a missing instruction in the User Guide or a missing feature in the tool. I searched both GitHub and the guide and could not find an answer.*

If you generate the default reference PPTX file using ``pandoc -o custom-reference.pptx --print-default-data-file reference.pptx``, the file will have a Section Header layout with a subtitle:

![image](https://user-images.githubusercontent.com/5067401/80982619-92f31a00-8df9-11ea-8678-44b8a2f79cdd.png)

I was unable to find a way to provide content for the subtitle with a placeholder of ``"Some explanatory text"``.

I first tried to generate a PPTX file using the [sample presentation Markdown](https://pandoc.org/MANUAL.html#producing-slide-shows-with-pandoc) found on the User Guide and the simple command ``pandoc sample.md --output sample.pptx``:

![image](https://user-images.githubusercontent.com/5067401/80982941-faa96500-8df9-11ea-9d56-8d953b738856.png)

I even tried adding text after the header to see if it shows up as a subtitle but it only showed up as an extra slide:

```
...
# In the evening

Some subtitle here

## Dinner
...
```

```
pandoc sample.md --output sample.pptx --slide-level=2
```

![image](https://user-images.githubusercontent.com/5067401/80983177-48be6880-8dfa-11ea-8436-51845251204c.png)

Am I missing something basic here? I couldn't tell if it was an undocumented feature or unimplemented feature.

seesharprun:
Almost forgot, I am running pandoc 2.7.3 on Windows 10.

jgm:
I don't use pptx writer myself, but have you tried setting the metadata title and subtitle in your document?

seesharprun:
If you set the metadata title and subtitle at the top of the document, that only impacts the title slide.

For the section header slides, I didn't see an easy way to set the subtitle. I tried this syntax:

```
## Setting the scene { subTitle="Demo" }
```

jgm:
Try a YAML header:
```
---
title: My title
subtitle: My subtitle
author: Me
...

# First slide

etc.
```

jgm:
Oh, I see, you want something similar on slides other than the first one.

jgm:
@jkr can probably answer this.

jkr:
[...]

PR #8342, opened 2022-10-01
Remove unsupported element from pptx reference doc

LunkRat:
Addresses #6338 by removing the misleading text from the pptx reference doc.


jgm:
Can you explain more fully what you've removed and why?

LunkRat:
Sure, I removed a bit of text from Slide 3 of the pptx reference doc, which was a small bit that appeared above the H1 on the 'Section Header' slide, which said:

> Some explanatory text

As pointed out in #6338 this example slide is misleading because it suggests that there is some Markdown mechanism for getting such text to appear on your Section Heading slides. As there is no such mechanism, I removed the example text so that others are not led astray trying to get their Markdown slides to appear in that text box.

jgm:
OK, thanks!

StianBerger:
Not that it matters so much, but I can say that just the presence of that text box is enough for "someone" :sweat_smile:  to spend hours figuring out how to use it...

Issue #8313, opened 2022-09-23
Pandoc generates corrupt docx files with trivial custom writer


svenboulanger:
I wanted to write a custom writer for docx documents, and wrote the following:

```lua
local pandoc = require 'pandoc'

function Writer(doc, opts)
    return pandoc.write(doc, 'docx', opts)
end
```

As far as I understand from the documentation, this should be simply falling back to the original writer.

Note that the docx file is corrupt, but also changing the extension to ZIP results in a corrupt zip archive.

I am using pandoc v2.19.2.

tarleb:
The problem is that custom writers are currently expected to be text writers, i.e., that they return UTF-8 encoded text. When binary data is returned, it will contain non-UTF-8 characters, and those will get mangled.

I plan to add support for a `BinaryWriter` function, but it requires some refactoring to the way PDFs are produced.

tarleb:
The current development version allows to do this with

``` lua
function ByteStringWriter(doc, opts)
    return pandoc.write(doc, 'docx', opts)
end
```

PR #8343, opened 2022-10-01
[API Change] Support binary custom writers.

tarleb:
New-style custom Lua writers can now define an alternative entry function `ByteStringWriter`. If a function with that name is defined, then pandoc will treat the returned string as binary output. This allows to generate formats like docx and odt with custom writers.

`T.P.Lua.writeCustom` function changed to reflect the possibility of a ByteString writer being returned.

Closes: #8313

### DIFF
--- a/data/sample.lua
+++ b/data/sample.lua
@@ -13,19 +13,19 @@
 -- produce informative error messages if your code contains
 -- syntax errors.
 
+function Writer (doc, opts)
+  PANDOC_DOCUMENT = doc
+  PANDOC_WRITER_OPTIONS = opts
+  loadfile(PANDOC_SCRIPT_FILE)()
+  return pandoc.write_classic(doc, opts)
+end
+
 local pipe = pandoc.pipe
 local stringify = (require 'pandoc.utils').stringify
 
--- The global variable PANDOC_DOCUMENT contains the full AST of
--- the document which is going to be written. It can be used to
--- configure the writer.
-local meta = PANDOC_DOCUMENT.meta
-
 -- Choose the image format based on the value of the
--- `image_format` meta value.
-local image_format = meta.image_format
-  and stringify(meta.image_format)
-  or 'png'
+-- `image_format` environment variable.
+local image_format = os.getenv 'image_format' or 'png'
 local image_mime_type = ({
     jpeg = 'image/jpeg',
     jpg = 'image/jpeg',
--- a/doc/custom-writers.md
+++ b/doc/custom-writers.md
@@ -16,16 +16,61 @@ install any additional software to do this.
 [Lua]: https://www.lua.org
 
 A custom writer is a Lua file that defines how to render the
-document. Two styles of custom writers are supported: classic
-custom writers must define rendering functions for each AST
-element. New style writers, available since pandoc 2.17.2, must
-define just a single function `Writer`, which gets passed the
-document and writer options, and then does all rendering.
+document. Writers must define just a single function named
+`Writer`, which gets passed the document and writer options, and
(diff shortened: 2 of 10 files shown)


jgm:
I think the `CustomWriter` constructor is a bit awkward (since a custom writer is either a text or a bytestring writer, it's just that which it is depends on what function the writer defines).

Here's a thought:  what if `engineReadCustom` had the type:

```hs
forall m. (PandocMonad m, MonadIO m) => FilePath -> m (Reader m)
```

`engineReadCustom` could parse the Lua file and determine whether it defines `Writer` or `ByteStringWriter`, then return the appropriate kind of `Reader`.
Would that work?

tarleb:
At least theoretically, it would be possible to have something like the below, which is captured by the type.

``` lua
if PANDOC_DOCUMENT.meta.use_binary then
  function BinaryWriter (doc, opts) ... end
else
  function Writer (doc, opts) ... end
end
```

An alternative interface could be to check for a second return value of `Writer`, and to treat the result as binary if the second value is truthy. That might be a bit cleaner for the current type.

The interface was influenced by my attempts to create something similar to what you suggest above, but I couldn't make it work. I don't remember the details, the first try dates back a few weeks. Maybe I tried with readers, and problem was the rank-n type of TextReader? Let me just give it another go.

tarleb:
So I think this is possible, but we'd have to remove the `PANDOC_DOCUMENT` and `PANDOC_WRITER_OPTIONS` globals that we currently set before loading the script. I think that'd be completely acceptable for new-style writers, but might break some classic writers.

tarleb:
Related discussions: #8120 and #8141.

tarleb:
I've changed the `CustomWriter` constructor to type `FilePath -> Writer`; feels a little less awkward and is easier to extend.

tarleb:
The more I think about it, the less I like the `CustomWriter` constructor. Changing this back to a draft.

jgm:
There are a lot of issues here!

Just a few comments:  I don't like the idea of making this sensitive to a metadata field. I'd prefer to keep a clean separation between metadata and processing directives.  The idea of checking a second return value from Writer makes more sense to me (but maybe instead of a Boolean, we should just check for the string "binary", which might make it more obvious what this is doing).

I still think the CustomWriter constructor is awkward; even more so if it's a function `FilePath -> Writer`.  Why not just evaluate the function right away, since we have the FilePath from the start?  But I may be misunderstanding the motivation for that.

Since we're going to 3.0, we could in principle drop support for old-style writers. In this case we'd want to make the transition easy by providing a canned function that converts an old-style writer into a new-style one.

tarleb:
OK great, I think we actually agree on most things. The next draft won't have the CustomWriter constructor. Most of the classic writer functionality should be salvageable, just the `PANDOC_DOCUMENT` global will probably become unavailable.

tarleb:
I'm reasonably happy with this now.

Changes:

- Updates to `ScriptingEngine` as you suggested.
- Added a new function `pandoc.write_classic` that runs the current Lua environment as classic writer.
- Classic writers are deprecated; the docs are updated to show an easy way to update a classic  to a new-style writer.
- The `PANDOC_DOCUMENT` and `PANDOC_WRITER_OPTIONS` globals are set to placeholder values when writer scripts are loaded.
- Support for ByteStringWriter.

jgm:
Excellent!

jgm:
Do we have any tests for ByteStringWriters?

tarleb:
Uh-oh. Let me fix that. I just realized that I also forgot to include the documentation for `pandoc.write_custom`. Will do so now.

PR #8348, opened 2022-10-03
Separate out T.P.Data, T.P.Translations from T.P.Class.

jgm:
This makes T.P.Class more self-contained, and suitable for extraction into a separate package if desired.

[API changes]

- T.P.Data is now an exported module, providing `readDataFile`, `readDefaultDataFile` (both formerly provided by T.P.Class), and also `getDataFileNames` (formerly unexported in T.P.App.CommandLineOptions).

- T.P.Translations is now an exported module (along with T.P.Translations.Types), providing `readTranslations`, `getTranslations`, `setTranslations`, `translateTerm`, `lookupTerm`, `readTranslations`, `Term(..)`, and `Translations`.

- T.P.Class: `readDataFile`, `readDefaultDataFile`, `setTranslations`, and `translateTerm` are no longer exported. `checkUserDataDir` is now exported.

- Text.Pandoc now exports Text.Pandoc.Data and `setTranslations` and `translateTerm`.


jgm:
I'd also like to consider breaking up T.P.Shared, so T.P.Class doesn't depend on it.
But this change is pretty self-contained.

tarleb (approved):
This is beautiful!

PR #8353, opened 2022-10-05
Update Lua filter doc: modify Zerobrane instructions to use Lua 5.4

iandol:
Zerobrane IDE does not yet include Lua 5.4. I had tried to get 5.3 working with pandoc debugging but the debugger never triggered. Then I used a 5.4 install (I installed from homebrew) with mobdebug and luasocket installed. Setting this in Zerobrane's settings and configuring it in the IDE and remote debugging worked again. I've modified the instructions. Note I'm not a Lua person so I hope the details are clear...

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -413,16 +413,21 @@ run within the donation-ware Lua editor and IDE,
 REPL console and UI to step-through and view all variables and
 state.
 
-If you already have Lua 5.3 installed, you can add
+Zerobrane doesn't come with Lua 5.4 bundled, but it can debug it, 
+so you should install Lua 5.4, and then add
 [`mobdebug`](https://luarocks.org/modules/paulclinger/mobdebug)
 and its dependency
 [`luasocket`](https://luarocks.org/modules/luasocket/luasocket)
-using [`luarocks`](https://luarocks.org), which should then be
-available on the path. ZeroBrane also includes both of these in
-its package, so if you don't want to install Lua separately, you
-should add/modify your `LUA_PATH` and `LUA_CPATH` to include the
-correct locations; [see detailed instructions
-here](https://studio.zerobrane.com/doc-remote-debugging).
+using [`luarocks`](https://luarocks.org). ZeroBrane can use your 
+lua 5.4 install by adding `path.lua = "/path/to/your/lua"` 
+in your settings file. Open your lua filter in Zerobrane, 
+add `require('mobdebug').start()` at the line where you want 
+your breakpoint.  Then make sure the Project > Lua Intepreter 
+is set to "Lua" and enable "Start Debugger Server" 
+[see detailed instructions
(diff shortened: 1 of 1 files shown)


tarleb:
Thank you! I don't use ZeroBrane, but looks good to me. Just one request: could you check the capitalization of "Lua" and "ZeroBrane"? Lua is usually capitalized, except when referring to the program.

If you have time, could you check if you can get it to run with the current development version of pandoc? We now allow to use pandoc as a drop-in replacement for the `lua` binary: Just create a symlink to `pandoc` and name it `pandoc-lua`. That would remove the requirement of installing a separate Lua interpreter. See also https://github.com/jgm/pandoc/blob/9e3a2e0dde43cf42b12bb497a5ff854738d67097/MANUAL.txt#L7124

iandol:
Yes, I'm happy to test with `pandoc-lua` and see how that goes. @tarleb how can I add `luasocket` and `mobdebug`; should I add my existing Lua paths to `LUA_PATH` or `package.path`, or are they already bundled?

iandol:
When I set `pandoc-lua` in Zerobrane then for normal debugging it works fine:

```
Program starting as '"/Users/ian/bin/pandoc-lua" -e "io.stdout:setvbuf('no')" "/var/folders/gh/fj1zbxfj5g39pqmmssm9pt340000gn/T/.8SmWiM"'.
Program 'pandoc-lua' started in '/Users/ian/Code/dotpandoc/filters' (pid: 49987).
Debugging session started in '/Users/ian/Code/dotpandoc/filters/'.
```

So far so good, next I removed my existing Lua 5.4 and Luarocks install and tried to use remote debugging, but got

```
/Users/ian/.local/share/pandoc/filters/loggingTest.lua:10: module 'mobdebug' not found:
```

So it seems `pandoc-lua` alone is not enough for this. Now I think Mobdebug is just a `.lua` file, so if the `luasocket` dep is compiled in then we can just put Mobdebug somewhere in `LUA_PATH`?

iandol:
And adding `mobdebug.lua` to `package.path` and it seems `socket` is not available:

```
Error running filter /Users/ian/.local/share/pandoc/filters/loggingTest.lua:
/Users/ian/.local/share/pandoc/filters/mobdebug.lua:106: module 'socket' not found:
	no field package.preload['socket']
```

jgm:
I was not sure from the comments above whether this is ready to go or needs more work?

iandol:
I have made the modification requested by @tarleb about being consistent with uppercase for names. I didn't get `pandoc-lua` working, and perhaps this can be ignored for another pull request in the future?

Regarding `pandoc-lua` it could work with mobdebug if the binary package `luasocket` were available, or some way to get luarocks working with it, but this seems a lot of work when installing Lua itself is so easy...

jgm:
Thanks!

tarleb:
Thank you for testing @iandol!

PR #8357, opened 2022-10-06
Lua: allow table structure as format spec.

tarleb:
This allows to pass structured values as format specifiers to
`pandoc.write` and `pandoc.read`.


tarleb:
Depends on #8355.

jgm:
What's the status of this PR?

tarleb:
It's outdated after the recent changes to Extensions and needs updating. It also depends on the decision on #8378.

tarleb:
This is good to go, but will probably need a rebase if it's merged after #8431.

jgm:
good, thanks!

PR #8360, opened 2022-10-08
Make #+pandoc-emphasis-pre work as expected (#8059)

### LINKED ISSUE #8059: Org-mode reader: `#+pandoc-emphasis-pre` doesn't work as expected

adql:
The main issue with the original implementation is that it kept the assumption that no emphasis markers may occur directly after most non-whitespace chars, e.g. in the middle of the word (in Pandoc terms: following native `Str`). But with the possibility to modify which chars are allowed around emphasized text, Pandoc should remain indifferent in that respect.

These commits allow it and modify some tests accordingly. This is both:

1. in accordance with Org-mode's behavior when customizing `org-emphasis-regexp-components`.
2. reasonable in many languages[^1] (but rare in English).

Resolves #8059.

[^1]: which is the case in Hebrew and German, both of which I speak.

### DIFF
--- a/src/Text/Pandoc/Readers/Org/Inlines.hs
+++ b/src/Text/Pandoc/Readers/Org/Inlines.hs
@@ -130,7 +130,8 @@ str = return . B.str <$>
       ( many1Char (noneOf $ specialChars ++ "\n\r ") >>= updatePositions' )
       <* updateLastStrPos
   where
-    updatePositions' str' = updatePositions (T.last str') >> return str'
+    updatePositions' str' = str' <$
+      maybe mzero (updatePositions . snd) (T.unsnoc str')
 
 -- | An endline character that can be treated as a space, not a structural
 -- break.  This should reflect the values of the Emacs variable
--- a/src/Text/Pandoc/Readers/Org/Inlines.hs
+++ b/src/Text/Pandoc/Readers/Org/Inlines.hs
@@ -126,8 +126,11 @@ linebreak :: PandocMonad m => OrgParser m (F Inlines)
 linebreak = try $ pure B.linebreak <$ string "\\\\" <* skipSpaces <* newline
 
 str :: PandocMonad m => OrgParser m (F Inlines)
-str = return . B.str <$> many1Char (noneOf $ specialChars ++ "\n\r ")
+str = return . B.str <$>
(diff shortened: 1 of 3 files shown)


adql:
By the way, I originally used `*>` in `updatePositions'`, but `hlint` suggested to replace it with `$>`. This is very strange because it's wrong. Replacing with `>>` eliminated the suggestion. Any explanation to that?

tarleb:
Thank you, this is excellent. `hlint` sometimes suggests functions that'd need a separate import; this would work though: `updatePositions str' = str' <$ update...`.

The function `T.last` is partial and throws an error when passed `""`. While we know that this can never happen, but we have the general policy of avoiding partial functions if possible: could you change the code to use `T.unsnoc` to replace `T.last`?

adql:
> this would work though: `updatePositions str' = str' <$ update...`.

Yes, I see how this expression is useful. Makes it clear that we're here for the side effect.

tarleb:
Thank you!

adql:
My pleasure :)

PR #8370, opened 2022-10-12
Lua: add function `pandoc.template.meta_to_context`.

tarleb:
The functions converts Meta values to template contexts; the intended use is in combination with `pandoc.template.apply`.


tarleb:
Here it is in action in djot: https://github.com/tarleb/djot/commit/d5ad78536611606988af7b70a1ce458a8cdbc8d0

Issue #8135, opened 2022-06-19
Need a standard directory to store custom writers (both in user data and system wide)


chrisaga:
Custom writers are powerful, especially the new style. So as we don't code a writer only for one document it would be nice to be able to store them in a directory and call them with only their filenames (equivalent to lua filters).

In addition it would allow us to take advantage of this undocumented feature : 
If the name of the custom writer starts the same as a template, the template is used and you don't need to clone it. For instance Pandoc uses default latex template when you run `pandoc -t latex-custom-writer.lua doc.md`

tarleb:
We've added support for this just last week: https://github.com/jgm/pandoc/pull/8112
There is no official release with that code yet, but the latest [nightly build](https://github.com/jgm/pandoc/actions/runs/2523241734) should work.

I never realized that the template selector could be tricked like that. It should be considered a bug, even though I agree that it's quite convenient.

jgm:
> For instance Pandoc uses default latex template when you run pandoc -t latex-custom-writer.lua doc.md

Oh yes, I can see why; everything after the `-` is stripped off as an extension when we determine the template name. I agree with @tarleb that this is unintended behavior and should be fixed.  #8137 tracks this.

PR #8372, opened 2022-10-13
Lua: Allow custom writers to define a default template

tarleb:
The template is only used if no template file is found in the *templates* data directory. Custom template files in `data/templates/` are expected to have a `.tmpl` extension.

Example use (using an example from pandoc-discuss):

``` lua
local stringify = pandoc.utils.stringify

function Writer (doc, opts)
  assert(doc.blocks[1] and doc.blocks[1].t == 'Table', 'expected a table')
  local tbl = pandoc.utils.to_simple_table(doc.blocks[1])
  local result = pandoc.List{}
  for _, row in ipairs(tbl.rows) do
    local meta = pandoc.Meta(doc.meta)   -- clone
    for i, cell in ipairs(tbl.header) do
      meta[stringify(cell)] = pandoc.utils.blocks_to_inlines(row[i])
    end
    result:insert(pandoc.write(pandoc.Pandoc({}, meta), 'markdown', opts))
  end
  return table.concat(result, '\f\n') .. '\f\n'
end

Template = pandoc.template.compile [[
${name} ${surname}, born ${date} in ${location}
]]
```

`Template` should probably be a function that's invoked only when pandoc is called with `-s`.

---

I think I like the behavior we get with this. My biggest worry is that I over-engineer custom writers.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 test/fb2/reader/* -text
 pandoc-lua-engine/test/*.custom		-text
+pandoc-lua-engine/test/*.txt		-text
--- a/doc/custom-writers.md
+++ b/doc/custom-writers.md
@@ -83,6 +83,17 @@ function Writer (doc, opts)
 end
 ```
 
+## Default template
+
+The default template of a custom writer is defined by the return
+value of the global function `Template`. Pandoc uses the default
+template for rendering when the user has not specified a template,
+but invoked with the `-s`/`--standalone` flag.
+
+The `Template` global can be left undefined, in which case pandoc
+will throw an error when it would otherwise use the default
+template.
+
 ## Example: modified Markdown writer
 
 Writers have access to all modules described in the [Lua filters
@@ -105,6 +116,11 @@ function Writer (doc, opts)
     end
   }
   return pandoc.write(doc:walk(filter), 'gfm', opts)
+end
+
+function Template ()
+  local template = pandoc.template
+  return template.compile(template.default 'gfm')
 end
 ```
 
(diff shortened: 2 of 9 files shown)


jgm:
If we were to do this, I'd be tempted to remove the searching for templates in `templates/` for custom writers.
That way there would just be one canonical way to specify a custom template for a writer.  Do you think the search path is still necessary?

tarleb:
I'd be fully on board with that, but was worried it could be too radical a change. Searching for the default template could be recovered with

``` lua
function Template ()
  local filename = pandoc.path.filename(PANDOC_SCRIPT_FILE)
  return pandoc.template.compile(
    pandoc.template.default(filename .. ".tmpl")
  )
end
```

jgm:
If we made this change we could presumably roll back
3cac33bdd2b2b69ac585cc37498deea49763c665
and
4a97f5799a4bf9b672202fbd95cdef6cc49f62f8
as well.

I doubt that many people are using custom templates with custom writers.  There is #8135 from June which shows evidence of a nonstandard use (which won't work anymore anyway).  Indeed, no pre-existing uses would work with this proposed change anyway (`.tmpl` extension is new). Why don't we pose a question on pandoc-discuss to see if eliminating a standard search path for custom writer templates would cause problems?

tarleb:
OK will do. :+1: 

I'd like to keep 4a97f5799a4bf9b672202fbd95cdef6cc49f62f8, or otherwise the code in my previous comment wouldn't work. I agree that the second commit was a mistake.

jgm:
Great - do you want to roll back https://github.com/jgm/pandoc/commit/3cac33bdd2b2b69ac585cc37498deea49763c665?

tarleb:
:+1:

tarleb:
Just realized that the rollback was basically included in this PR; nothing else to do.

PR #8376, opened 2022-10-14
replaced # Bytestring readers paragraph 1

IlonaSilverwood:


doc/custom-readers.md
```diff
@@ -78,13 +78,7 @@ ensuring backwards compatibility.
...
 # Bytestring readers
 
-Pandoc expects text input to be UTF-8 encoded. However, formats
-like docx, odt, epub, etc. are not text but binary formats. To
-read them, pandoc supports `ByteStringReader` functions. These
-functions work just like the `Reader` function that process text
-input, but instead of a list of sources, `ByteStringReader`
-functions are passed a bytestring, i.e., a string that contains
-the binary input.
+Pandoc expects text input to be UTF-8 encoded. In order to read binary formats, including docx, odt, and epub, pandoc supports ByteStringReader functions. These functions are similar to the Reader function that processes text input. Instead of a list of sources, ByteStringReader functions are passed a bytestring, i.e., a string that contains the binary input.
```
tarleb, line 81: Thank you! There is one `Reader` or `ByteStringReader` function per file, would singular "function" make more sense?

IlonaSilverwood: maybe, which version is more readable?

>In order to read binary formats, including docx, odt, and epub, pandoc supports (the?) `ByteStringReader` function. A `ByteStringReader` function is similar to the `Reader` function that processes text input. Instead of a list of sources, the `ByteStringReader` function is passed a bytestring, i.e., a string that contains the binary input.

cesar803 (commented):
Nice

doc/custom-readers.md
tarleb: Sorry, this got lost. I like the above :+1:

jgm: I didn't want this to get lost.  I think the above looks good (maybe with `the`?).
If @IlonaSilverwood doesn't want to update it, maybe @tarleb or I could do a quick online edit; it also needs to be wrapped within 80 columns.

tarleb: Ilona and I talked briefly, an updated version is incoming.

tarleb:
I've picked the commit manually, fixed it up and pushed it to main. Closing.

Issue #8373, opened 2022-10-13
Simplify the creation of non-derivative custom writers


tarleb:
Classic writers performed actions that require additional boilerplate in new-style writers. It would be nice to offer a similar level of convenience.

Maybe instead of the new function `pandoc.write_classic(doc, wopts)` we could define a function `pandoc.make_writer(style[, fun_tbl])` that returns a callable table. If `style` is `classic`, then the functions should look like those in a classic writer. Otherwise the functions work like filter functions but must return `Doc` elements (or strings).

The current `pandoc.utils.write_classic(doc, opts)` could be recovered with `pandoc.make_writer('classic', _G)(doc, opts)`.

A typical writer would then look like this:

``` lua
Writer = pandoc.make_writer()
Writer.Inlines.Str = function (s) return escape(s.text) end
Writer.Inlines.Emph = functions (emph) return concat('<em>', Writer.Inlines(emph), '</em>') end
--- ...
```

The functions `Inlines` and `Blocks` would be auto-generated, and a template would be applied automatically.

I'm slightly fuzzy on the details, but I think it could result in a decent interface.

jgm:
So, is the idea that `pandoc.make_writer()` would return a "generic" writer that just outputs nothing for each type of element, unless you explicitly override these fields of the table?
(Or would cause an "unimplemented" warning to be issued in these cases?)
This does seem a nice idea to me.

tarleb:
Great, I'll give it a try then.

Lua 5.4 comes with support for a "warning system". I'd like to link it to pandoc's warnings and use it to notify users of undefined elements. (The warning system could also be plugged in so users could add pandoc log messages via the new standard Lua function `warn`; but I'm getting ahead of myself.)

jgm:
That all sounds good to me!

PR #8377, opened 2022-10-15
Lua: add function `pandoc.make_writer`.

tarleb:
This can be used to reduce boilerplate in custom writers.

Closes: #8373

### DIFF
--- a/doc/custom-writers.md
+++ b/doc/custom-writers.md
@@ -126,6 +126,62 @@ end
 
 [Lua filters documentation]: https://pandoc.org/lua-filters.html
 
+## Reducing boilerplate with `pandoc.make_writer`
+
+The module `pandoc.make_writer` is a custom writer scaffold that
+serves to avoid common boilerplate code when defining a custom
+writer. The function allows to define a writer by adding render
+functions for each element type.
+
+The output of `pandoc.make_writer` is a new function that should
+usually be assigned to the global `Writer`:
+
+``` lua
+Writer = pandoc.make_writer()
+```
+
+The render functions for Block and Inline values can then be added
+to `Writer.Block` and `Writer.Inline`, respectively. The functions
+are passed the element and the WriterOptions.
+
+``` lua
+Writer.Inline.Str = function (str)
+  return str.text
+end
+Writer.Inline.SoftBreak = function (_, opts)
+  return opts.wrap_text == "wrap-preserve"
+    and cr
+    or space
+end
+Writer.Inline.LineBreak = cr
+
+Writer.Block.Para = function (para)
+  return {Writer.Inlines(para.content), pandoc.layout.blankline}
+end
+```
+
+The render functions must return a string, a pandoc.layout *Doc*
+element, or a list of such elements. In the latter case, the
+values are concatenated as if they were passed to
+`pandoc.layout.concat`. If the value does not depend on the input,
+a constant can be used as well.
+
+The tables `Writer.Block` and `Writer.Inline` can be used as
+functions; they apply the right render function for an element of
+the respective type. E.g., `Writer.Block(pandoc.Para 'x')` will
+delegate to the `Writer.Para` render function and will the result
+of that call.
+
+Similarly, the functions `Writer.Blocks` and `Writer.Inlines` can
+be used to render lists of elements, and `Writer.Pandoc` renders
+the document's blocks.
+
+All predefined functions can be overwritten when needed.
+
+The resulting Writer uses the render functions to handle metadata
+values and converts them to template variables. The template is
+applied automatically if one is given.
+
 # Classic style
 
 A writer using the classic style defines rendering functions for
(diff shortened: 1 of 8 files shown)


tarleb:
Left to do:

- [ ] add default functions
- [ ] add warnings
- [ ] possibly rename items.
- [ ] more usability improvements, if any

Example use in djot: https://github.com/tarleb/djot/commit/d6a2d934d44bbe69b69c72fead6a995a4dfbb026
There's a minor difference in that the `Inline` rendering functions are collected in table `Inline`, vs. `Inlines` in the djot writer. The `Writer.Inline` function can then be used to render any Inline element, similar to what we do in filters.

A minor usability improvement: `return pandoc.layout.concat{...}` seems like such a common pattern that the function call to `concat` can now be omitted.

Performance is roughly unchanged to the manual version.

Happy to add more features were possible.

tarleb:
The filter currently doesn't enforce the division of `Inline` and `Block` functions: defining `Writer.Emph` is equivalent to setting `Writer.Inline.Emph`. Not sure I like that though.

tarleb:
Additional changes:

- We allow `Writer.Inline.Space` etc. to be plain `Doc` values; saves a little bit of typing and gives a small performance boost (it saves the marshaling step each time a Space is converted.
- All functions get `WriterOptions` as the second parameter. E.g., SoftBreak in `djot` would read
    ``` lua
    Inlines.SoftBreak = function (el, opts)
      if opts.wrap_text == "wrap-preserve" then
        return cr
      else
        return space
      end
    end
    ```

djot: https://github.com/tarleb/djot/commit/d682d018f8c8fd26e7bd40aef64b1bfdca2b8441

jgm:
This is looking good!

jgm:
I don't think I can be much help here, I don't know much about Lua stack tracing!

jgm:
Your HsLua suggestion sounds worth a try, but I don't know enough about it to offer an informed opinion!
What do the error messages above look like after your cleanup?

tarleb:
Error message when there's an error in `Str` for input `Emph [Str ""]`:

```
Error running Lua:
../testing/custom-writer.lua:18: intentional error
stack traceback:
        ../testing/custom-writer.lua:18: in function <../testing/custom-writer.lua:17>
        [C]: in field 'Inlines'
        ../testing/custom-writer.lua:15: in function <../testing/custom-writer.lua:14>
        [C]: in field 'Inlines'
        ../testing/custom-writer.lua:9: in function <../testing/custom-writer.lua:8>
        while rendering Inline `Str`
        while rendering Inlines
        while rendering Inline `Emph`
        while rendering Inlines
        while rendering Block `Para`
        while rendering Blocks
        while rendering Pandoc
stack traceback:
```

When there's no render function for an element:

```
Error running Lua:
No render function for Inline value 'Link';
define a function `Writer.Inline.Link` that returns a string or Doc.
stack traceback:
        ../testing/custom-writer.lua:9: in function <../testing/custom-writer.lua:8>
        while rendering Block `Para`
        while rendering Blocks
        while rendering Pandoc
stack traceback:
```

tarleb:
I went with `pandoc.make_writer` for now, but I'm still conflicted about this choice.

Other than that it should be good to go.

jgm:
To me they look equally clean.  So if the module approach gives better error messages, I'd say that gives it an edge.

tarleb:
I've renamed it to `pandoc.writer_scaffolding` to make the intend of this construct a little clearer. The error message is now

```
Error running Lua:
../testing/custom-writer.lua:19: intentional error
stack traceback:
        ../testing/custom-writer.lua:19: in function <../testing/custom-writer.lua:18>
        [C]: in function 'pandoc.writer_scaffolding.Inlines'
        ../testing/custom-writer.lua:10: in function <../testing/custom-writer.lua:8>
        [C]: in function 'pandoc.writer_scaffolding.Blocks'
        [C]: in function 'pandoc.writer_scaffolding.Pandoc'
        while rendering Inline `Str`
        while rendering Inlines
        while rendering Block `Header`
        while rendering Blocks
        while rendering Pandoc
stack traceback:
```

tarleb:
I'm thinking about naming it just `pandoc.scaffolding`; not sure if we'd want to add a different type of scaffolding at some point.

tarleb:
I went with `pandoc.scaffolding.Writer`, even though it's quite a mouthful.

The error message looks like this now:

```
Error running Lua:
../testing/custom-writer.lua:19: intentional error
stack traceback:
        ../testing/custom-writer.lua:19: in function 'pandoc.scaffolding.Writer.Inline.Str'
        [C]: in function 'pandoc.scaffolding.Writer.Inlines'
        ../testing/custom-writer.lua:10: in function 'pandoc.scaffolding.Writer.Block.Header'
        [C]: in function 'pandoc.scaffolding.Writer.Blocks'
        [C]: in function 'pandoc.scaffolding.Writer.Pandoc'
        while rendering Inline `Str`
        while rendering Inlines
        while rendering Block `Header`
        while rendering Blocks
        while rendering Pandoc
stack traceback:
```

jgm:
OK!  Should these commits all be squashed into one, or did you want to arrange them in some other way?

tarleb:
Just squashing is fine, the individual commits have no real relevance other than making it easier for me to experiment.

PR #8387, opened 2022-10-20
Add custom extensions.

jgm:
T.P.Extensions [API change]:

+ Add CustomExtension constructor to Extension.
+ Remove Bounded, Enum instances for Extension.
+ Add `extensionsToList` function.
+ Revise `readExtension` so it can handle CustomExtension.

T.P.Format:

+ Revise error checking to handle CustomExtension.


tarleb (approved):
Oh wow, that's clean! :sparkles:

jgm:
I think I made it even cleaner by adding `showExtension`.

Issue #8378, opened 2022-10-15
Extensions for custom readers/writers


jgm:
Provide a method for passing custom extensions to custom readers/writers.

One approach would be to add a `CustomExtension Text` constructor to `Extension`.

Another would be something like
```hs
data FlavoredFormat =
    Format Text Extensions
  | CustomFormat FilePath (Set CustomExtension)
```

My thought is that no extensions would be enabled by default for custom readers/writers (an extension means "non-default change"), and readers/writers could implement whatever extensions, with whatever names, they like.

One question here is whether custom readers/writers use standard library functions that are sensitive to ordinary extensions (e.g. functions that create auto identifiers and might be sensitive to `ascii_identifiers`).  Certainly `pandoc.write` and `pandoc.read` are sensitive to regular extensions, but the custom reader/writer could always set these appropriately before calling them, and could even set them based on the setting of a like-named custom extension.

tarleb:
> My thought is that no extensions would be enabled by default for custom readers/writers (an extension means "non-default change"), and readers/writers could implement whatever extensions, with whatever names, they like.

My main concern is that could make it more difficult or awkward to write and use derivative writers, i.e., one that's built on top of an existing writer. For example, if I understand correctly, then users could express "use this custom kramdown writer, but without 'smart' behavior" by calling `--to=kramdown.lua+no_smart`, which seems a little awkward.

I don't really like the `ExtensionsDiff` type, but it captures the relevant information nicely and allows to pass it around.

jgm:
I suppose that if we instead added a `CustomExtension Text` constructor to `Extensions`, we could have
```
  | CustomFormat FilePath ExtensionsDiff
```
and then the custom format would have full information about what was enabled and disabled on the command line. Would this work better?  (We'd still have to consider the costs of teh custom constructor.)

jgm:
@tarleb any thoughts on this proposal, as against the current status quo, in which custom readers/writers define their own extensions + default values, but must pick from the already defined extensions?

tarleb:
I was a bit short on time but will write up my thoughts tomorrow.

tarleb:
I'm completely in favor of the above. I assume we'd add some small helper functions to `pandoc.format` to make it easier to implement different behaviors in the custom readers/writers?

I've updated #8357, and the changes there might be helpful for the proposed changes.

jgm:
[...]

PR #8390, opened 2022-10-21
WIP: extensions in custom readers/writers

jgm:
For #8378

### DIFF
--- a/src/Text/Pandoc/Scripting.hs
+++ b/src/Text/Pandoc/Scripting.hs
@@ -22,7 +22,7 @@ import Text.Pandoc.Class.PandocMonad (PandocMonad)
 import Text.Pandoc.Definition (Pandoc)
 import Text.Pandoc.Error (PandocError (PandocNoScriptingEngine))
 import Text.Pandoc.Filter.Environment (Environment)
-import Text.Pandoc.Format (ExtensionsConfig)
+import Text.Pandoc.Format (ExtensionsDiff)
 import Text.Pandoc.Templates (Template)
 import Text.Pandoc.Readers (Reader)
 import Text.Pandoc.Writers (Writer)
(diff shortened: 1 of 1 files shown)


jgm:
@tarleb how do we make the ExtensionsDiff available in the custom reader/writer?
Is this a matter of setting a global variable? If so, what should it be?  Is there already code for marshalling an ExtensionsDiff?

tarleb:
I've been wondering about this, too. A few options:

1. `WriterOptions` could get a new field `writerFormat`, then `writerExtensions` in `WriterOptions` would become a function that calculates the value from the format. This would be quite clean, I think, but might be a little bothersome to work with in the non-custom writers.
2. We set `INPUT_FORMAT` and `OUTPUT_FORMAT` globals.
3. The format gets passed as a third parameter to the `Reader` and `Writer` functions..
4. We do 1., but only on the marshaled Lua object; the Haskell structure stays as it is.

My current order of preference is 1, 4, 3, 2.

tarleb:
Marshal functions would be something like

``` haskell
pushExtDiff = pushAsTable
  [ ("enable", pushList pushExtension . extsToEnable)
  , ("disable", pushList pushExtensions . extsToDisable)
  ]
```

which would get a structure very close to what we have in Haskell.

Alternatively, we could build a table with extensions as keys and a boolean value to indicate whether it should be enabled or disabled.

``` haskell
pushExtDiff diff = pushKeyValuePairs $ (map (,True) (extsToEnable diff)) ++ map (,False) (extsToDisable diff)
```

jgm:
We don't really need FORMAT; we just need ExtensionsDiff.
Why not make it available as a global, say,
```
EXTENSIONS = { 'foobar': true, 'smart': false }
```
where true and false indicate set/unset?  Or we could just use `'enabled'`/`'disabled'`, which might cause less confusion.

tarleb:
Sounds good. Boolean keys would be easier to use, I think. We could also add a module `pandoc.Set` (or, more general, `pandoc.Map`) that makes it easier to work with Lua sets.

jgm:
> Boolean keys would be easier to use, I think. 

My worry is that we need a threefold distinction:  enabled, disabled, not mentioned.  Boolean keys make sense with a twofold distinction, since `extensions.smart` will be treated as a false value whether `smart` is `false` or simply unspecified.  Of course, you can test for `== false` as opposed to `== nil` and get the distinction, but my worry is that this might be too subtle.

jgm:
Alternatively we could have something like
```
extensions = {
  enabled = {
    smart: true 
  },
  disabled = {
    foo: true
  }
}
```

jgm:
Another alternative would be to keep the original plan, in which the custom writer defines its own defaults for extension, returning ExtensionsConf.  If we did that, then the only change needed would be to add the ability to handle custom extensions (and maybe very little or nothing would be needed for that).

tarleb:
I'm ok with both; users have to learn about another global either way. I agree that `Extensions` would be a better name.

> However, if we go this way, we should also have custom readers define their own extensions and return ExtensionsConf.

I think that's covered by 253d2e768a43c8ab3ad8e1c46b2bc4a02acec946.

I keep wondering if it could make sense to implement JSON filters as a scripting engine. That would then also add JSON readers and writers. Maybe that would help us to make design decisions.

jgm:
> I keep wondering if it could make sense to implement JSON filters as a scripting engine. That would then also add JSON readers and writers. Maybe that would help us to make design decisions.

Interesting idea.  What sorts of design constraints would this impose, do you think?

tarleb:
Not really sure, but I assume that passing data to the readers/writers is easier than retrieving config data from the writers. It would be interesting to see how easy or difficult it would be to support custom default templates: e.g., would pandoc call the writer twice, once for the template and once for the conversion, or would it try to communicate via a pipe, or would we rather define a default location where the template would be expected. Likewise for readers and extensions. There would also be the challenge of making the mediabag content available to the writer and to retrieve it from the reader, but that's orthogonal to what we're discussing here. 

I believe that we already arrived at a good abstraction, and that the small changes that we are discussing would not change the implementation in any significant way. The idea may therefore be more trouble than it's worth. It might increase the usefulness of panflute and the like, but not sure.

jgm:
I changed the names to `Extensions`.

jgm:
Our present abstraction:
```hs
  , engineWriteCustom :: forall m. (PandocMonad m, MonadIO m)
                      => FilePath
                      -> m (Writer m, ExtensionsConfig, m (Template Text))
```
One thing I wonder about here: for the template, we return an action that will fetch the template if run.  For the ExtensionsConfig, we just return the ExtensionsConfig directly.  Why the difference?

tarleb:
With the current implementation, the ExtensionsConfig is always required, while the Template is retrieved on demand. Not sure if this is a good reason though.

PR #8395, opened 2022-10-22
Support passthrough for literate agda

isovector:
Agda supports literate markdown and tex files, and is capable of performing semantic syntax highlighting of these literate documents. The result is significantly better highlighting than can be achieved without compiler support.

Unfortunately, Agda can only produce highlighted latex files if the source is literate latex. A nice thought would be to preprocess literate markdown files into literate latex, and then run Agda's latex backend to get the final highlighting. This almost works in pandoc today, except that pandoc transforms agda code fences into `\begin{verbatim}` blocks in the outputted latex.

This is a quick PR that adds a new `Ext_literate_agda` extension which emits `\begin{code}` in the latex writer for agda code blocks. The result is that agda can now happily consume the resulting `latex` file.

### DIFF
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -699,7 +699,7 @@ codeBlockFenced = try $ do
               Just (elementId, classes, attrs) -> (elementId, maybe classes (: classes) languageId, attrs)))
   blankline
   contents <- T.intercalate "\n" <$>
-                 manyTill (gobbleAtMostSpaces indentLevel >> anyLine)
+                 manyTill (anyLine)
                           (try $ do
                             blockDelimiter (== c) (Just size)
                             blanklines)
--- a/src/Text/Pandoc/Extensions.hs
+++ b/src/Text/Pandoc/Extensions.hs
@@ -96,6 +96,7 @@ data Extension =
     | Ext_link_attributes         -- ^ link and image attributes
     | Ext_lists_without_preceding_blankline -- ^ Allow lists without preceding blank
     | Ext_literate_haskell    -- ^ Enable literate Haskell conventions
+    | Ext_literate_agda    -- ^ Enable literate Agda conventions
     | Ext_markdown_attribute      -- ^ Interpret text inside HTML as markdown iff
                                   --   container has attribute 'markdown'
     | Ext_markdown_in_html_blocks -- ^ Interpret as markdown inside HTML blocks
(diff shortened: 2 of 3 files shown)


jgm:
I'm open to adding literate Agda support, but it would need to be more complete and consistent with existing literate Haskell support.

At the very least, we need support in both the LaTeX reader and LaTeX writer. Also, the convention of sensitivity to a `literate` class should be preserved in the writer.  (The point of this is that you might have a literate Agda document that contains both Agda code samples that are not part of the literate program and code that is part of the literate program; the `literate` class is used to mark out the latter from the former.)

Since you're just modifying the LaTeX writer, I assume you're reading the content as Markdown or some other format?  In that case, a code block could be marked with the `literate` class using the following syntax (in pandoc markdown):

````
``` {.agda .literate}
foobar
```
````

If your source format is gfm, then this won't work, but you could use a small Lua filter to inject the `literate` class on every code block that has `agda`.

Note also that if you just want a "quick and dirty" solution, you can use a Lua filter to format every agda code block using `\begin{code}...\end{code}` in LaTeX output:

```lua
function CodeBlock(el)
  if FORMAT == "latex" and el.classes:includes("agda") then
    return pandoc.RawBlock("latex",
      "\\begin{code}\n" .. el.text .. "\n\\end{code}")
  end
end
```

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -699,7 +699,7 @@ codeBlockFenced = try $ do
               Just (elementId, classes, attrs) -> (elementId, maybe classes (: classes) languageId, attrs)))
   blankline
   contents <- T.intercalate "\n" <$>
-                 manyTill (gobbleAtMostSpaces indentLevel >> anyLine)
+                 manyTill (anyLine)
```
jgm, line 702: Why do you consider this a bug?  It is meant to ensure that whitespace that only serves to align the code with the start of the fence will not be counted as part of the code.
e.g. in
````
1.  ``` ruby
    x = y
    ````
````
The contents of the code block should be `"x=y"` not `"    x=y"`.

Comment (isovector) [reply src/Text/Pandoc/Readers/Markdown.hs:702]: I think this PR is necessarily more destructive to pandoc than is feasible to get merged. The issue is that Agda's literate mode doesn't attempt to do any whitespace reindenting, so 

~~~
```agda
record Foo where
  field
```

some prose

1. ```agda
    someField : Type
    ```
~~~

gets mangled in the output to be

```agda
record Foo where
  field
someField : Type
```

which breaks the indent rules.

jgm: I see.  Well, let's close this PR, then.  I think the Lua filter approach I outline above could be helpful, but you may run across the indentation problem with it too.

PR #8406, opened 2022-10-31
Text.Pandoc.App: Change `parseOptionsFromArgs` and `parseOptions`...

jgm:
They now return `Either OptInfo Opt`.

Add `OptInfo` type.

Add `handleOptInfo` function.  This performs the IO actions for things like `--version` that were previously done in `parseOptionsFromArgs`. An argument for a `ScriptingEngine` has been added, to facilitate printing custom templates and custom extensions for Lua filters. (However, at this stage nothing is yet done with it.)

[API change]

### DIFF
--- a/pandoc-cli/src/pandoc.hs
+++ b/pandoc-cli/src/pandoc.hs
@@ -16,7 +16,7 @@ module Main where
 import qualified Control.Exception as E
 import System.Environment (getArgs, getProgName)
 import Text.Pandoc.App ( convertWithOpts, defaultOpts, options
-                       , parseOptionsFromArgs )
+                       , parseOptionsFromArgs, handleOptInfo )
 import Text.Pandoc.Error (handleError)
 import qualified Text.Pandoc.UTF8 as UTF8
 import System.Exit (exitSuccess)
@@ -61,10 +61,12 @@ main = E.handle (handleError . Left) $ do
       case rawArgs of
         "lua" : args   -> runLuaInterpreter "pandoc lua" args
         "server": args -> runServer args
-        _              -> do
+        args           -> do
           engine <- getEngine
-          opts <- parseOptionsFromArgs options defaultOpts prg rawArgs
-          convertWithOpts engine opts
+          res <- parseOptionsFromArgs options defaultOpts prg args
+          case res of
+            Left e -> handleOptInfo engine e
+            Right opts -> convertWithOpts engine opts
 
 copyrightMessage :: String
 copyrightMessage =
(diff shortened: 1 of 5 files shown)


jgm:
Still not sure about this. Maybe it would be better simply to add the ScriptingEngine option to parseOptionsWithArgs, so users don't have to worry about handling both alternatives and running `handleOptInfo`.  Indeed, then we could hide `OptInfo`.

tarleb:
I really like this (especially the use of `ExceptT`, which makes the change surprisingly small).

jgm:
What do you think about the alternative of not exposing `handleInfo`, and instead adding a `ScriptingEngine` parameter to `parseOptions`?  That seems simpler for the user, and I'm having trouble thinking of a  reason why to go the other way.

tarleb:
Both appear equally good to me. Adding the extra parameters seems a bit weird conceptually, but, as the Zen of Python states, practicality beats purity.

jgm:
They're going to have to specify a ScriptingEngine one way or another, either in parseOptions or in handleOptInfo; both seem equally weird, so there's an argument for simplifying the user experience.

The only reason I could see for keeping `handleOptInfo` separate would be, e.g., allowing a web application or GUI application to parse options and handle things in a different way (e.g. by displaying something in a box on the screen instead of printing to stdout).  Don't know if such applications would ever need/want to parse options, though, so this seems pretty speculative.

jgm:
@tarleb  I've been implementing `--list-extensions` on custom readers and writers, and I ran into an issue.  Even though `engineReadCustom` / `enginWriteCustom` are in PandocMonad, they seem to throw errors that can't be caught as pandoc exceptions.  e.g.
```lua
       getglobal "ByteStringReader" >>= \case
          TypeNil -> failLua $ "No reader function found: either 'Reader' or "
                     <> "'ByteStringReader' must be defined."
```
This prevents me from doing e.g.
```hs
   catchError (snd <$> engineReadCustom ...) (\_ -> snd3 <$> engineWriteCustom ...)
```
in order to try the file argument to `--list-extensions` first as a custom reader and then as a custom writer.
Can we change `failLua` here to a proper `throwError ` with a pandoc exception?

tarleb:
That's weird, `failLua` should raise a `PandocLuaError` in this context. Does it work with `catch` from `Control.Monad.Catch`?

jgm:
The question is not which error is being raised, but how it is being raised.  Can it be raised with `throwError`?  Then we could catch it in the regular way we catch pandoc errors (`catchError` from Control.Monad.Except).

Issue #6992, opened 2020-12-30
Handle col/rowspans in mediawiki tables


Dolu89:
Hello,
I discovered a problem on the converter mediawiki to gfm (and all markdown destination)
It cuts off everything after the colspan
Pandoc version (ubuntu 20.04 LTS) 
```
pandoc 2.5
Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.7.7
```

**Source mediawiki**
``` mediawiki
Block template creation can be influenced by various parameters:
{| class="wikitable"
!colspan=4|template request
|-
! Key !! Required !! Type !! Description
|-
| capabilities || No || Array of Strings || SHOULD contain a list of the following, to indicate client-side support: [[#Optional: Long Polling|"longpoll"]], "coinbasetxn", "coinbasevalue", [[bip-0023.mediawiki#Block Proposal|"proposal"]], [[bip-0023.mediawiki#Logical Services|"serverlist"]], "workid", and any of the [[bip-0023.mediawiki#Mutations|mutations]]
|-
| mode || No || String || MUST be "template" or omitted
|}
```
**Result of mediawiki**
![Capture d’écran de 2020-12-30 10-26-05](https://user-images.githubusercontent.com/4184590/103342349-95416d00-4a89-11eb-8deb-4e88cac09fa4.png)

I run this command to transform my mediawiki file to a md file
`pandoc -f mediawiki -t gfm ./mediawiki-file.mediawiki -o ./gfm-file.md`

**Result provided by pandoc (gfm)**
``` markdown
Block template creation can be influenced by various parameters:

| template request |
| ---------------- |
| Key              |
| capabilities     |
| mode             |
```
**Result of gfm**
![Capture d’écran de 2020-12-30 10-29-44](https://user-images.githubusercontent.com/4184590/103342538-06812000-4a8a-11eb-8fb3-7bb1a8ade6ee.png)

jgm:
Until recently pandoc could not represent col/rowspans in tables.
This has changed in the basic types, but we haven't updated all the readers and writers yet.

Dolu89:
Hi, I was just wondering if there are any updates on this issue and its resolution timeline.
Thanks!

PR #8418, opened 2022-11-03
Rewrite the MediaWiki Writer to use the new tables

wgevaert:
~~Fixes issue #6992~~ (on closer reading, that issue seems to be about the mediawiki reader, this fixes the writer.)

Now the MediaWiki tables can have colspan and rowspan.
The align-attribute has been replaced with a styling option.

See https://www.mediawiki.org/wiki/Help:Tables for more info on formatting mediawiki tables. As can be read there, mw tables do not have the possibility of having `thead`/multiple `tbody`s/`tfoot` or `colgroup`, so I ignored quite a lot of information regarding those.

Most of the code is similar to the html writer, which is why I moved some html-functions to the Shared file, such that it could also be used by mediawiki. 

I am very unexperienced with Haskell, and eager to learn, so please do not hesitate to give codestyle tips or other feedback.

Edit: I also added the `wikitable` class to all tables that are made in this way. It will give the tables the default mediawiki style, which I believe is preferred in most cases. ofc. I could remove that if you prefer the writer to be "pure".

### DIFF
--- a/src/Text/Pandoc/Writers/Shared.hs
+++ b/src/Text/Pandoc/Writers/Shared.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE LambdaCase #-}
 {- |
    Module      : Text.Pandoc.Writers.Shared
    Copyright   : Copyright (C) 2013-2022 John MacFarlane
@@ -23,6 +24,9 @@ module Text.Pandoc.Writers.Shared (
                      , defField
                      , getLang
                      , tagWithAttrs
+                     , htmlAddStyle
+                     , htmlAlignmentToString
+                     , htmlAttrs
                      , isDisplayMath
                      , fixDisplayMath
                      , unsmartify
@@ -53,6 +57,7 @@ import qualified Data.Map as M
 import qualified Data.Text as T
 import Data.Text (Text)
 import qualified Text.Pandoc.Builder as Builder
+import Text.Pandoc.CSS (cssAttributes)
 import Text.Pandoc.Definition
 import Text.Pandoc.Options
 import Text.DocLayout
@@ -168,9 +173,12 @@ getLang opts meta =
 
 -- | Produce an HTML tag with the given pandoc attributes.
 tagWithAttrs :: HasChars a => Text -> Attr -> Doc a
-tagWithAttrs tag (ident,classes,kvs) = hsep
-  ["<" <> text (T.unpack tag)
-  ,if T.null ident
+tagWithAttrs tag attr = "<" <> text (T.unpack tag) <> (htmlAttrs attr) <> ">"
+
+-- | Produce HTML for the given pandoc attributes, to be used in HTML tags
+htmlAttrs :: HasChars a => Attr -> Doc a
+htmlAttrs (ident, classes, kvs) = addSpaceIfNotEmpty (hsep [
+  if T.null ident
       then empty
       else "id=" <> doubleQuotes (text $ T.unpack ident)
   ,if null classes
@@ -178,7 +186,35 @@ tagWithAttrs tag (ident,classes,kvs) = hsep
       else "class=" <> doubleQuotes (text $ T.unpack (T.unwords classes))
   ,hsep (map (\(k,v) -> text (T.unpack k) <> "=" <>
                 doubleQuotes (text $ T.unpack (escapeStringForXML v))) kvs)
-  ] <> ">"
+  ])
+
+addSpaceIfNotEmpty :: HasChars a => Doc a -> Doc a
+addSpaceIfNotEmpty f = if isEmpty f then f else " " <> f
+
+-- | Adds a key-value pair to the @style@ attribute.
+htmlAddStyle :: (Text, Text) -> [(Text, Text)] -> [(Text, Text)]
+htmlAddStyle (key, value) kvs =
+  let cssToStyle = T.intercalate " " . map (\(k, v) -> k <> ": " <> v <> ";")
+  in case break ((== "style") . fst) kvs of
+    (_, []) ->
+      -- no style attribute yet, add new one
+      ("style", cssToStyle [(key, value)]) : kvs
+    (xs, (_,cssStyles):rest) ->
+      -- modify the style attribute
+      xs ++ ("style", cssToStyle modifiedCssStyles) : rest
+      where
+        modifiedCssStyles =
+          case break ((== key) . fst) $ cssAttributes cssStyles of
+            (cssAttribs, []) -> (key, value) : cssAttribs
+            (pre, _:post)    -> pre ++ (key, value) : post
+
+-- | Get the html representation of an alignment key
+htmlAlignmentToString :: Alignment -> Maybe Text
+htmlAlignmentToString = \case
+  AlignLeft    -> Just "left"
+  AlignRight   -> Just "right"
+  AlignCenter  -> Just "center"
+  AlignDefault -> Nothing
 
 -- | Returns 'True' iff the argument is an inline 'Math' element of type
 -- 'DisplayMath'.
--- a/src/Text/Pandoc/Writers/Shared.hs
+++ b/src/Text/Pandoc/Writers/Shared.hs
@@ -172,8 +172,8 @@ getLang opts meta =
                _                                 -> Nothing
 
 -- | Produce an HTML tag with the given pandoc attributes.
-tagWithAttrs :: HasChars a => Text -> Attr -> Doc a
-tagWithAttrs tag attr = "<" <> text (T.unpack tag) <> (htmlAttrs attr) <> ">"
+tagWithAttrs :: HasChars a => a -> Attr -> Doc a
+tagWithAttrs tag attr = "<" <> literal tag <> (htmlAttrs attr) <> ">"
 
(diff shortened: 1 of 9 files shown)


wgevaert:
I believe all the failing tests are testing changed behaviour, so the associated test should be changed to reflect that.

I have no experience with this testing, how do I change the tests such that they pass?

jgm:
If you have make, then `make TESTARGS=--accept` will change the golden tests.
(It won't change tests in Test.Writers.MediaWiki, if there are any.)
Just be sure to carefully inspect the changes to make sure everything looks right.

jgm:
Or if you don't want to use make, pass `--accept` into the test options via stack or cabal.
`--test-options=--accept` on cabal.

wgevaert:
I did adjust the tests like you said, thank you for your help!

In the tests, there are some table cells with text like "Example of a row that spans multiple lines." or "Here’s another one. Note the blank line between rows." that do not actually span multiple lines or have a blank line. However, the [html output of the same test](https://github.com/jgm/pandoc/blob/7a9832166e36f77402d5e0259647e9f5c7ba4e58/test/tables.html5) also has this behaviour, so I think that is ok.

Please tell me if I should do anything else before this will be merged.

Do you have an estimate when this will be included in a new release?

Thanks a lot for your help!

jgm:
> In the tests, there are some table cells with text like "Example of a row that spans multiple lines."

Those made sense in the original Markdown test suite, but not here. No need to worry about it though.

wgevaert:
Any idea if/when this will be merged? I'd help me a lot if these features are included in a stable release 😄

jgm:
Could you split this into two commits, one moving the functions from the HTML writer to T.P.Writers.Shared, and the other with the MediaWiki writer changes?

The commit that adds exports to T.P.Writers.Shared should contain `[API change]` in its comment, and describe the API change, so we will not forget to document this in the release.

I see no reason why this couldn't be in the next release.

wgevaert:
I've done that now :)

test/tables.mediawiki
```diff
@@ -1,145 +1,152 @@
...
+{| class="wikitable"
 |+ Demonstration of simple table syntax.
-!align="right"| Right
-! Left
-!align="center"| Center
+|-
+! style="text-align: right;"| Right
+! style="text-align: left;"| Left
+! style="text-align: center;"| Center
 ! Default
```
jgm, line 9: Can you comment on the change with align?
Is the use of `style: text-align` now recommended for wikipedia tables?

test/tables.mediawiki
```diff
@@ -1,145 +1,152 @@
...
 |-
-|align="right"| 12
-| 12
-|align="center"| 12
+| style="text-align: right;"| 12
+| style="text-align: left;"| 12
```
jgm, line 12: Can we leave off a style tag if it's default-aligned?

src/Text/Pandoc/Writers/Shared.hs
```diff
@@ -166,17 +171,48 @@ getLang opts meta =
 
 -- | Produce an HTML tag with the given pandoc attributes.
 tagWithAttrs :: HasChars a => Text -> Attr -> Doc a
-tagWithAttrs tag (ident,classes,kvs) = hsep
-  ["<" <> text (T.unpack tag)
-  ,if T.null ident
+tagWithAttrs tag attr = "<" <> text (T.unpack tag) <> (htmlAttrs attr) <> ">"
```
jgm, line 174: I'm pretty sure `literal tag` will do the same thing and avoid the unpack. (But I know this was in the original code, so no worries if you don't want to change it.)

jgm:
How about some tests of the new features?  I see changes to the old tests, but shouldn't we have tests for complex tables with rowspans and colspans and any other new features that are now possible?

The simplest way to do this would be to add a "command test" `test/command/8418.md`.  The format should be obvious from the other examples.

test/tables.mediawiki
wgevaert: Yes, this is recommended. [This edit from 2013](https://www.mediawiki.org/w/index.php?title=Help:Tables&diff=842660&oldid=842627) states "killing deprecated align= attributes", so it has been recommended to use `style: text=align` instead of the `align` attributes since then.

In general, mediawiki markup is very close to html markup, and the discussion on whether `(text-)align` on table elements should be an attribute or style tag for html is not very clear from google, but it seems like `style` tags for align have support since ~2000, and `align` attributes have been deprecated for some time, depending on the source (at least for html5).

I like to follow the [guide in my first comment](https://www.mediawiki.org/w/index.php?title=Help:Tables), which uses style tags since 2013. It seems to be an official source.

test/tables.mediawiki
wgevaert: I could do that. However, this might be a bit messy, as the default lay-out (defined by the `class=wikitable`) gives headers (starting with `!`) a default center-align, and cells (starting with `|`) a default left-align, so I should do a switch on this to determine what is the default. (See also the [wikitable css page](https://en.wikipedia.org/wiki/Wikipedia:Catalogue_of_CSS_classes/Wikitable), which even seems to state that the default align for the caption depends on whether you view it on mobile or desktop.)

For now I went with the easy solution of "When an align is given I add it and when it is default I ignore it" (I'm not that much of a front-end/accessibility guy).

I'll see what I can do to implement this kind of switch.

src/Text/Pandoc/Writers/Shared.hs
wgevaert: thanks, will do!

wgevaert: I don't understand it, it doesn't seem to compile:

using `text (literal tag)` instead of `text (T.Unpack tag)`:
```
src/Text/Pandoc/Writers/Shared.hs:174:38: error:
    • Couldn't match type: Doc Text
                     with: [Char]
      Expected: String
        Actual: Doc Text
    • In the first argument of ‘text’, namely ‘(literal tag)’
      In the first argument of ‘(<>)’, namely ‘text (literal tag)’
      In the second argument of ‘(<>)’, namely
        ‘text (literal tag) <> (htmlAttrs attr) <> ">"’
    |
174 | tagWithAttrs tag attr = "<" <> text (literal tag) <> (htmlAttrs attr) <> ">"
    |                                      ^^^^^^^^^^^
cabal: Failed to build pandoc-3.0 (which is required by exe:pandoc from
pandoc-cli-0.1).
```

Using `(literal tag)` instead of `text (T.unpack tag)`: 
```
src/Text/Pandoc/Writers/Shared.hs:174:41: error:
    • Couldn't match expected type ‘a’ with actual type ‘Text’
      ‘a’ is a rigid type variable bound by
        the type signature for:
          tagWithAttrs :: forall a. HasChars a => Text -> Attr -> Doc a
        at src/Text/Pandoc/Writers/Shared.hs:173:1-51
    • In the first argument of ‘literal’, namely ‘tag’
      In the first argument of ‘(<>)’, namely ‘(literal tag)’
      In the second argument of ‘(<>)’, namely
        ‘(literal tag) <> (htmlAttrs attr) <> ">"’
    • Relevant bindings include
        tagWithAttrs :: Text -> Attr -> Doc a
          (bound at src/Text/Pandoc/Writers/Shared.hs:174:1)
    |
174 | tagWithAttrs tag attr = "<" <> (literal tag) <> (htmlAttrs attr) <> ">"
    |                                         ^^^
cabal: Failed to build pandoc-3.0 (which is required by exe:pandoc from
pandoc-cli-0.1).
```

I do not know enough about haskell to solve this.

test/tables.mediawiki
wgevaert: Furthermore, the default depends on whether the language is rtl or ltr, I don't see a way to infer what writing direction the language will be. Any tips on this?

wgevaert:
I added some tests for the `planets`, `nordic` and `students` tables. When rendered by mediawiki, they look like this:

Students table
Nordic table
Planets table

I think this is what they should look like?

src/Text/Pandoc/Writers/Shared.hs
jgm: I see. Here's the type of `literal`

literal :: [HasChars](https://hackage.haskell.org/package/doclayout-0.4/docs/Text-DocLayout.html#t:HasChars) a => a -> [Doc](https://hackage.haskell.org/package/doclayout-0.4/docs/Text-DocLayout.html#t:Doc) a

So if `tag` is a `Text`, then `literal tag` will be a `Doc Text`. However, we've specified this function as polymorphic, it can produce `Doc a` for any `a` that is an instance of the `HasChars` class.  That's why it doesn't work, my mistake!

We could change the type of the function to make `tag` `a` rather than `Text`.  You could try that (and use `literal tag`), but I don't know if there are other places in the code that presuppose that tag is Text. I guess the compiler would let you know.

test/tables.mediawiki
jgm: Yes, it sounds as if what you're doing is what I meant: if we have AlignDefault, then don't print any special alignment.

I'm just surprised by the difference in the tests -- I had assumed that the reason the alignments were left off before was that we had AlignDefault -- but maybe the old code just treated AlignLeft as AlignDefault?

wgevaert: Yes it did

src/Text/Pandoc/Writers/Shared.hs
wgevaert: Thank you for the explanation, I am starting to understand Haskell a lot more!
Changing the type definition of `tagWithAttrs` worked (in hindsight it is quite obvious haha).

jgm:
Looks good! Thansk!

PR #8420, opened 2022-11-04
Use ghc4pandoc 9.4.3 docker image

benz0li:
@jgm Pandoc builds with GHC v9.4.3, too.

ℹ️ Tested for both `linux/amd64` and `linux/arm64/v8`.


jgm:
Good , thanks!

Issue #8405, opened 2022-10-28
export to pdf with -ms, extension of sub/superscript indices


nbehrnd:
Caveat:  I'm not sure if this can be adjusted in pandoc, or if addressing ghostcript were better.

For areas and volumes, I'm used to reach out for Alt + 2, or Alt + 3 to write e.g., 2 m², or 3 m³, respectively.  There is no apparent problem for pandoc to use such an input when generating a .pdf *via* groff

```shell
pandoc -Vpapersize=a4 --pdf-engine-opt=-P-pa4 -t ms -o result.pdf minimal.md
```

and want to suggest to extend the range available to `0123456789+-`.

It is observed the very same markdown input, e.g. as `test_unicode.md`

```
row 1: ? ? x² x³ ? ? ? ? ? ? ? ? ? ? ? ?

row 2: x⁰ xⁱ ? ? x⁴ x⁵ x⁶ x⁷ x⁸ x⁹ x⁺ x⁻ x⁼ x⁽ x⁾ xⁿ

row 3: x₀ x₁ x₂ x₃ x₄ x₅ x₆ x₇ x₈ x₉ x₊ x₋ x₌ x₍ x₎ ?

row 4: xₐ xₑ xₒ xₓ xₔ xₕ xₖ xₗ xₘ xₙ xₚ xₛ xₜ ? ? ?
```

yields a complete .pdf including the super/sub scripts when running 

```shell
$ pandoc -s test_unicode.md -o unicode_via_html.pdf -t html
[WARNING] This document format requires a nonempty <title> element.
  Defaulting to 'test_unicode' as the title.
  To specify a title, use 'title' in metadata or --metadata title="...".
Loading page (1/2)
Printing pages (2/2)                                               
Done                                                     
```

This contrasts to the initially intended approach with `-t ms`, where -- despite correct rendering of e.g., `äöüâé` -- only `x²` and `x³` pass well by the analogue request:

```shell
$ pandoc -Vpapersize=a4 --pdf-engine-opt=-P-pa4 -t ms test_unicode.md -o unicode_via_ms.pdf 
troff: <standard input>:108: warning: can't find special character 'u2070'
troff: <standard input>:108: warning: can't find special character 'u2071'
troff: <standard input>:110: warning: can't find special character 'u2074'
troff: <standard input>:110: warning: can't find special character 'u2075'
troff: <standard input>:110: warning: can't find special character 'u2076'
troff: <standard input>:110: warning: can't find special character 'u2077'
[...]

PR #8421, opened 2022-11-04
clarify font selection for pdf -t ms

nbehrnd:
Previous versions of the documentation did not clarify much which options of font selection are available for the generation of .pdf via `-t ms`.  Nor was there a description how to extend the set of of fonts provided by ghostscript/pdfgroff.  This commit equally hints to Peter Schaffter's script install-fonts.sh to facilitate greatly the later, too.

This PR addresses some of the findings relevant to the issue report filed earlier ([issue 8405](https://github.com/jgm/pandoc/issues/8405)).

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3088,7 +3088,23 @@ The `--css` option also affects the output.
 ### Variables for ms
 
 `fontfamily`
-:    font family (e.g. `T` or `P`)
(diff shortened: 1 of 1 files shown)


jgm:
Thank you. I am glad to get instructions for installing fonts for ms. I've edited this to make it more concise (since the manual is already quite long).  Any objections?

PR #8422, opened 2022-11-04
Change default background color of PDFs generated via HTML

marcin-serwin:
When creating PDF output via HTML the background of `body` element is `transparent`, so it uses the color of `html` which is set to `#fdfdfd` by default. This may potentially lead to wasted printer ink. The introduced fix changes the default color for printing to `white` while respecting the case of user setting a custom background color.

## Steps to reproduce

Run
```sh
$ echo foo | pandoc -o foo.pdf -t html
```

### Before and after
before after

### The above pictures with increased contrast and lowered brightness
before after

### DIFF
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -28,6 +28,9 @@ body {
   }
 }
 @media print {
+  html {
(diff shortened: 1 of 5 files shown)


jgm:
Good improvement!

PR #8423, opened 2022-11-05
Don't mess with user's font-size or line-height

snan:
Commit changes the template expansion so that unless fontsize or linestretch are set, browser defaults are used.

I realize that this isn't gonna be an "sure, NP, that's a small fix" automatic merge and that it's gonna see some resistance, and it's an immediately visible change from how the template currently looks, but I think it's ultimately the right call.

I don't think huge letters and enormous leading is necessarily good, readable typography for prose text. Even if you do, there are four places prose text font-size can be set:

1. On the client side, users setting their own preferred font sizes.
2. On the site side, every web page pushing their own preference.
3. On the browser implementation side, browsers deciding on a new default (which users then can change).
4. On the spec / W3C side.

Which of those four places do you think is the absolute worst and most pessimal to mess with this setting? That’s right, 2. On the site side. Making this a site-specific tweak, making the new, partially adopted "de facto default" 20px, is gonna lead to the web becoming even more of a gross hodge-podge of styles.

Even if you disagree with that, the commit still respects the pandoc user's $fontsize and $linestretch variables. (Those settings are not good for the open web but maybe people are generating styled HTML for other things.)

### DIFF
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -1,8 +1,8 @@
 $if(document-css)$
-html {
-  line-height: $if(linestretch)$$linestretch$$else$1.5$endif$;
-  font-family: $if(mainfont)$$mainfont$$else$Georgia, serif$endif$;
-  font-size: $if(fontsize)$$fontsize$$else$20px$endif$;
+html {$if(linestretch)$
+  line-height: $linestretch$;$endif$
+  font-family: $if(mainfont)$$mainfont$$else$Georgia, serif$endif$;$if(fontsize)$
+  font-size: $fontsize$;$endif$
   color: $if(fontcolor)$$fontcolor$$else$#1a1a1a$endif$;
   background-color: $if(backgroundcolor)$$backgroundcolor$$else$#fdfdfd$endif$;
 }
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -1,7 +1,9 @@
 $if(document-css)$
-html {$if(linestretch)$
+html {
+$if(linestretch)$
   line-height: $linestretch$;$endif$
-  font-family: $if(mainfont)$$mainfont$$else$Georgia, serif$endif$;$if(fontsize)$
+  font-family: $if(mainfont)$$mainfont$$else$Georgia, serif$endif$;
+$if(fontsize)$
   font-size: $fontsize$;$endif$
   color: $if(fontcolor)$$fontcolor$$else$#1a1a1a$endif$;
   background-color: $if(backgroundcolor)$$backgroundcolor$$else$#fdfdfd$endif$;
--- a/test/lhs-test.html
+++ b/test/lhs-test.html
@@ -7,9 +7,7 @@
   <title>lhs-test</title>
   <style>
     html {
-      line-height: 1.5;
       font-family: Georgia, serif;
-      font-size: 20px;
       color: #1a1a1a;
       background-color: #fdfdfd;
     }
--- a/test/lhs-test.html+lhs
+++ b/test/lhs-test.html+lhs
@@ -7,9 +7,7 @@
   <title>lhs-test</title>
   <style>
     html {
-      line-height: 1.5;
       font-family: Georgia, serif;
-      font-size: 20px;
       color: #1a1a1a;
       background-color: #fdfdfd;
     }
--- a/test/writer.html4
+++ b/test/writer.html4
@@ -10,9 +10,7 @@
   <title>Pandoc Test Suite</title>
   <style type="text/css">
     html {
-      line-height: 1.5;
       font-family: Georgia, serif;
-      font-size: 20px;
       color: #1a1a1a;
       background-color: #fdfdfd;
     }
--- a/test/writer.html5
+++ b/test/writer.html5
@@ -10,9 +10,7 @@
   <title>Pandoc Test Suite</title>
   <style>
     html {
-      line-height: 1.5;
       font-family: Georgia, serif;
-      font-size: 20px;
       color: #1a1a1a;
       background-color: #fdfdfd;
     }


masukomi:
I recognize that the following isn't quite the same thing, but the principle applies. 

I'm technically a "low vision" user. I have my font size set to 20 _and_ my _minimum_ font size set to 20.  This breaks countless sites and tools, because they've made assumptions about font sizing and layout that don't play well with accessibility tools.

Any one person's idea of a "just right" size is too small for some and too big for others. Pandoc should honor what's its input, and never insert its own opinions about "good" size. Browsers do a spectacular job of handling unspecified sizing in such a way as to accommodate users with various accessibility needs.

tarleb:
LGTM; pinging @mb21 to get a second (well, third) opinion.

mb21:
This PR is changing exactly two things:

1. Removing the `line-height: 1.5` style (browser default is `1`)
2. Removing the `font-size: 20px` style (browser default is `16px`)

Obviously, the exact values are a matter of taste, but they were discussed at length here: https://groups.google.com/g/pandoc-discuss/c/1oCCbois1G4/m/mazyChM-AwAJ

Note that these settings make the text _more_ legible than the browser defaults for the majority of users. Also, all modern browsers support zooming out or into the page, so users are not at all left to the mercy of what's set by the CSS sent from the server.

@snan Is there a specific use-case I cannot think of where this CSS takes precedence over a user supplied setting? If so, can you describe and exactly what browser/setting etc?

snan:
There are [problems with setting font-size on web pages](https://idiomdrottning.org/font-size), and the most commonly recommended font-size is the default (which, as of this writing, is 16px, which of course device and browser vendors can change to something that makes more sense for their customers).

I submitted this patch immediately after talking to a blind guy who uses pandoc to generate his web page (which he can't see himself, only listen to). He trusts in the defaults, that the defaults selected by pandoc should be good. "User-supplied settings" are almost meaningless to him.

mb21:
But can you describe a specific problem? The page you linked to seems outdated.

From the pandoc-discuss thread I previously linked to:

> Regarding user intention: There are actually two different kinds of browser settings the user can toggle:
> 
> 1. zoom in the whole page
> 2. increase just the font size
> 
> see https://support.mozilla.org/en-US/kb/font-size-and-zoom-increase-size-of-web-pages
> 
> In modern browsers, both of these work just fine when use set the base unit on the document root in pixels. AFAIK, the advice of using em on the root is outdated, and most people are more familiar with setting fontsizes in px than em.

---

>  talking to a blind guy who uses pandoc to generate his web page (which he can't see himself, only listen to)

I don't see how CSS should affect him at all? He'll use a screen-reader that basically reads out the HTML...

snan:
> But can you describe a specific problem? The page you linked to seems outdated.

I just wrote it the other day. 🤦🏻‍♀️ 

> setting fontsizes in px than em

Completely irrelevant to what I wrote, which is don't set font-size.

The widely recommended default is (equivalent to) 16px. Not touching font size at all accomplishes that in a more future-proof way.

> I don't see how CSS should affect him at all? He'll use a screen-reader that basically reads out the HTML...

Blind writers are publishing HTML for sighted readers and if those writers trust in the Pandoc defaults, that'll lead them to publish a non-standard large font size. My patch fixes that issue.

snan:
> 2. Setting a fixed pixel-based font size is not good from an accessibility point of view.
> 
> So my understanding is that we can reconcile those two points by setting the default font size to `1.25 rem`. 

No, pixel-based is not the issue. Leave the font at the browser default.

tarleb:
I don't understand then. What are the downsides of setting slightly opinionated defaults, as long as accessibility is not affected. Please explain in more detail (and, if you want to do me a favor, with less swearing / drastic wording than the article above).

tarleb:
OK, upon re-reading everything that was said here and elsewhere on the web, I now understand a little better. The choice is really between respecting user settings and making the site look better for the majority of users. I think I'm in favor of not setting the font-size, but line-height should be set regardless.

data/templates/styles.html
```diff
@@ -1,7 +1,9 @@
 $if(document-css)$
-html {$if(linestretch)$
+html {
+$if(linestretch)$
   line-height: $linestretch$;$endif$
```
jgm, line 4: I think this `$endif$` can go on the next line without messing up spacing (might need testing).  If so, this looks better:
```
$if(linestretch)$
   line-height: $linestretch$;
$endif$
```
Same approach for fontsize and the others.

snan:
> with less swearing / drastic wording than the article above

What swearing? There was no swearing.

mb21:
> I think I'd be in favor of making the default styles for HTML less opinionated and more minimalistic.

I agree with the principle.

> For LaTeX, for example, we just stick with the default styles

True, although one could argue that the "default" for browsers is no custom CSS, which is what pandoc had for a long time. Until we decided that browser's defaults are just a historical artifact that they need to keep for backwards-compatibility, but nobody actually likes anymore, and that we could improve the lives of most users by loading some minimal CSS with pandoc.

> But if we go in that direction there are a number of other things that should probably be reconsidered

Exactly, it's quite a slippery slope and ultimately mostly subject to personal opinions.

---

> One thing I don't like about our current setup is that it doesn't distinguish between things that are really essential for pandoc [...] and things that are more stylistic preferences [...]
>
> One option would be to have a more minimal default CSS, and change the behavior of --css so that instead of replacing this, it augments it. 

We have the debate whether things should be overriden or somehow merged (and if so what should take precedence) already with metadata arrays and it's not always easy for the user to understand what's going on...

Wouldn't it be perhaps enough to split the current `styles.html` file into two, and load the other one from the first one? So people tracking the repo can see into which category changes fall?

---

> Setting a fixed pixel-based font size is not good from an accessibility point of view.

I would like to question this. There is a lot of outdated information floating around the web, both regarding px vs. relative units like em, and regarding the zoom behaviour of browsers, which has improved a lot over the last five or ten years. I could not reproduce any problem with setting the font-size (even setting it in pixels) in a any modern browser. Even after toggling the zoom behviour (in Firefox under menu ``View -> Zoom -> Zoom Text Only`) zooming still works. That's why I've been asking for concrete, reproducible steps for the perceived problem.

snan:
> > Setting a fixed pixel-based font size is not good from an accessibility point of view.
> 
> I would like to question this. There is a lot of outdated information floating around the web, both regarding px vs. relative units like em, and regarding the zoom behaviour of browsers, which has improved a lot over the last five or ten years. I could not reproduce any problem with setting the font-size (even setting it in pixels) in a any modern browser. Even after toggling the zoom behviour (in Firefox under menu `View -> Zoom -> Zoom Text Only`) zooming still works. That's why I've been asking for concrete, reproducible steps for the perceived problem.

For the sixth time:

# It's not about px vs em or rem! The method of setting it isn't the issue.

Just leave font-size alone.

snan:
> > I think I'd be in favor of making the default styles for HTML less opinionated and more minimalistic. For LaTeX, for example, we just stick with the default styles (even though for my own output I might prefer different settings).
> 
> Personally I think everything except font-size and line-height is fine. I can make a better case for font-size whereas for line-height it's more subjective.

@jgm 

I think max-width is especially necessary to keep in, while font-size is especially important to remove, and I'd prefer to remove line-height also. Those three are what I care about. Anything else, like the block quote line or font choice, is fine either way, gone or remaining, either is fine.

tarleb:
@snan I think that was directed at me; it may not be relevant for the PR, but it is for general design decisions.

@mb21 I based that statement on a note that I found on MDN. I don't know enough about this topic and current practices to judge the extend to which this holds nowadays.

> Note: To maximize accessibility, it is generally best to use values that are relative to the user's default font size.

snan:
> Note: To maximize accessibility, it is generally best to use values that are relative to the user's default font size.

That's vague but correct, for things _other than the prose font-size itself_. Being able to zoom text was an accessibility feature back then (before smartphones and such made it something that concerns everyone) and you'd want some elements to be sized relative to the user's font size. (This was in the day when scaling images did not look very good and was not particularly fast, which is why the current method of setting a zoom factor for the entire page wasn't used.)

PR #8429, opened 2022-11-09
Add LaTeX option `urlstyle`

amarakon:
It would be neat if there was an option to change the URL style for LaTeX documents with an option. By default, Pandoc sets the `urlstyle` to `same` which I do not like because it results in me being unable to differenciate URLs and regular text. If you still want to set the default to `same`, feel free to make that change in my pull request.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -384,7 +384,9 @@ $if(csquotes)$
 $endif$
 \IfFileExists{bookmark.sty}{\usepackage{bookmark}}{\usepackage{hyperref}}
 \IfFileExists{xurl.sty}{\usepackage{xurl}}{} % add URL line breaks if available
-\urlstyle{same} % disable monospaced font for URLs
+$if(urlstyle)$
+\urlstyle{$urlstyle$}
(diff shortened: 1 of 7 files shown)


jgm:
Can you change it to keep `same` as the default?
Also, tests need to be adjusted.  `make TESTARGS=--accept` will do this if you're using the Makefile.  Or pass in `--accept` as an argument to the test suite.  Then examine changes carefully.

amarakon:
I made the changes you requested. Let me know if you want me to squash them into one commit.

data/templates/default.latex
```diff
@@ -386,6 +386,8 @@ $endif$
 \IfFileExists{xurl.sty}{\usepackage{xurl}}{} % add URL line breaks if available
 $if(urlstyle)$
 \urlstyle{$urlstyle$}
+$else$
+\urlstyle{same}
```
jgm, line 390: Better would be:
```
\urlstyle{$if(urlstyle)$$urlstyle$$else$same$endif$}
```

Issue #8417, opened 2022-11-03
Architectural change in ScriptingEngine


jgm:
Suggestion:  Instead of the current interface
```hs
  , engineReadCustom :: forall m. (PandocMonad m, MonadIO m)
                     => FilePath -> m (Reader m, ExtensionsConfig)
    -- ^ Function to parse input into a 'Pandoc' document.

  , engineWriteCustom :: forall m. (PandocMonad m, MonadIO m)
                      => FilePath
                      -> m (Writer m, ExtensionsConfig, m (Template Text))
```
what about something like
```hs
  , engineLoadScript :: forall m . (PandocMonad m, MonadIO m)
       => FilePath -> m (ScriptComponents m)
```
where
```hs
data ScriptComponents m =
  ScriptComponents
  { scriptCustomReader :: Maybe (Reader m)
  , scriptCustomWriter :: Maybe (Writer m)
  , scriptTemplate :: Maybe Text  -- or Maybe (Template Text), depending on that decision
  , scriptExtensions :: Maybe ExtensionsConfig
  }
```

This would make things nicer for implementing `-D script.lua`, since we wouldn't need to try to read it as a reader and then again as a writer if that fails.  Similarly for implementing `--list-extensions script.lua`.

Also, I just realized that the current architecture for custom readers and writers makes it possible to combine a reader and writer in a single script (or am I wrong)?  Maybe this should be noted in the documentation.

@tarleb comments?

tarleb:
That's elegant. I'm not sure yet, but I think I like this.

jgm:
Alternatively we could have
```hs
data ScriptComponents =
  ScriptComponents
  { scriptCustomReader :: forall m. (PandocMonad m, MonadIO m) => Maybe (Reader m)
  , scriptCustomWriter :: forall m. (PandocMonad m, MonadIO m) => Maybe (Writer m)
  , scriptTemplate :: Maybe Text  -- or Maybe (Template Text), depending on that decision
  , scriptExtensions :: Maybe ExtensionsConfig
  }
```
Are there reasons to go one way or the other here?

tarleb:
The parameterized version `data ScriptComponents m` seems a little nicer to me, but don't feel strongly about it.

PR #8431, opened 2022-11-10
Implement changes in ScriptingEngine

jgm:
See #8417.

@tarleb I left the changes to pandoc-lua-engine to you (I made the most obvious ones).
The second commit in this draft PR outlines what remains to be done.

### DIFF
--- a/pandoc-lua-engine/pandoc-lua-engine.cabal
+++ b/pandoc-lua-engine/pandoc-lua-engine.cabal
@@ -86,8 +86,7 @@ library
                      , Text.Pandoc.Lua.Module.Utils
                      , Text.Pandoc.Lua.Orphans
                      , Text.Pandoc.Lua.PandocLua
-                     , Text.Pandoc.Lua.Reader
-                     , Text.Pandoc.Lua.Writer
+                     , Text.Pandoc.Lua.Custom
                      , Text.Pandoc.Lua.Writer.Classic
                      , Text.Pandoc.Lua.Writer.Scaffolding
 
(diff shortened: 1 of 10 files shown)


tarleb:
Will do. I'm currently knee-deep in figure handling; trying to finish that first.

tarleb:
I started to work on this. One question I have is how we should handle files in the `$DATADIR/readers` and `$DATADIR/writers` directories. Should we simply check both for suitable scripts? Also should we start to search `$DATADIR/custom` for .lua files when running `customLoadScript`?

jgm:
>  One question I have is how we should handle files in the $DATADIR/readers and $DATADIR/writers directories. Should we simply check both for suitable scripts? Also should we start to search $DATADIR/custom for .lua files when running customLoadScript?

Yes, I wondered about this too.  How long ago did we add support for `readers` and `writers` directories?  Was it in a released version?  It is tempting to consolidate and just have a `custom` directory that could contain either a reader or a writer or a reader/writer.

tarleb:
Checking the changelog it seems that this feature was released in pandoc 2.19. Probably fine if we phase that out over the next year or so.

jgm:
For now we could search in both places and deprecate use of readers/writers?
But actually I'd be tempted just to make a clean break (3.0 after all) and document the change prominently.  I doubt too many people are using it yet.

tarleb:
Sounds good. Hope it's ok that I've pushed a rebased and updated version to your branch.

I've dropped the fallback handling of custom readers that expect string input instead of the actual Sources list, but could also put that code back.

PR #8456, opened 2022-11-27
Fix epub-embed-font documentation

### LINKED ISSUE #8455: Documentation Bug - epub-embed-font
**Explain the problem.**

The documentation for `--epub-embed-font` is incorrect.

It gives as an example CSS:

```css
@font-face {
font-family: DejaVuSans;
font-style: normal;
font-weight: normal;
src:url("DejaVuSans-Regular.ttf");
}
```

The `src:url` is incorrect. It needs to be a relative path.  e.g.

```css
src:url("../fonts/DejaVuSans-Regular.ttf");
```

I've tested this in KOReader and Calibre. Both require the relative path.

As discussed in https://groups.google.com/g/pandoc-discuss/c/3VrefhiuVmM/m/I-NjsG_zBgAJ

**Pandoc version?**

* pandoc 2.19.2
* Ubuntu

edent:
Fixes #8455

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1282,28 +1282,28 @@ header when requesting a document from a URL:
     to your CSS (see `--css`):
 
         @font-face {
-        font-family: DejaVuSans;
-        font-style: normal;
-        font-weight: normal;
-        src:url("DejaVuSans-Regular.ttf");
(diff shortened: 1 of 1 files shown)


jgm:
Thanks!

PR #8481, opened 2022-12-11
Store "unnumbered" class in DocBook role attribute

### LINKED ISSUE #1402: convert unnumbered sections to docbook's <simplesect>
Right now an unnumbered section: # section {-}

is just converted to a `<sectX>`. I think it makes sense to convert it to a `<simplesect>` or maybe a `<section>` to distinguish it from normal section in the docbook output.

lifeunleaded:
Markdown allows marking a heading as unnumbered, which is stored as a class token internally. This change will recognize this particular class token and append it to the role attribute, or create a role attribute with it if needed. This does not imply any processing in DocBook but is intended to let customized stylesheets identify these sections and act accordingly.

The enrichRole function is intentionally designed to be able to take more class tokens in the future by extending the `cand` list

fixes #1402

### DIFF
--- a/src/Text/Pandoc/Writers/DocBook.hs
+++ b/src/Text/Pandoc/Writers/DocBook.hs
@@ -14,7 +14,7 @@ Conversion of 'Pandoc' documents to DocBook XML.
 module Text.Pandoc.Writers.DocBook ( writeDocBook4, writeDocBook5 ) where
 import Control.Monad.Reader
 import Data.Generics (everywhere, mkT)
-import Data.Maybe (isNothing)
+import Data.Maybe (isNothing, maybeToList)
 import Data.Monoid (Any (..))
 import Data.Text (Text)
 import qualified Data.Text as T
@@ -169,7 +169,7 @@ blockToDocBook :: PandocMonad m => WriterOptions -> Block -> DB m (Doc Text)
 blockToDocBook _ Null = return empty
 -- Add ids to paragraphs in divs with ids - this is needed for
 -- pandoc-citeproc to get link anchors in bibliographies:
-blockToDocBook opts (Div (id',"section":_,_) (Header lvl (_,_,attrs) ils : xs)) = do
+blockToDocBook opts (Div (id',"section":_,_) (Header lvl (_,classes,attrs) ils : xs)) = do
   version <- ask
   -- DocBook doesn't allow sections with no content, so insert some if needed
   let bs = if null xs
@@ -191,7 +191,8 @@ blockToDocBook opts (Div (id',"section":_,_) (Header lvl (_,_,attrs) ils : xs)) 
                  else []
 
       -- Populate miscAttr with Header.Attr.attributes, filtering out non-valid DocBook section attributes, id, and xml:id
-      miscAttr = filter (isSectionAttr version) attrs
+      -- Also enrich the role attribute with certain class tokens
+      miscAttr = enrichRole (filter (isSectionAttr version) attrs) classes      
       attribs = nsAttr <> idAttr <> miscAttr
   title' <- inlinesToDocBook opts ils
   contents <- blocksToDocBook opts bs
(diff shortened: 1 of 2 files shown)


tarleb:
Thank you!

I think the code in `enrichRole` could possibly be cleaned up a little. Checkout the functions in the *Data.List* module, esp. [lookup](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-List.html#v:lookup) and [partition](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-List.html#v:partition), those could help to keep the function succinct.

jgm:
Is the use of `role="unnumbered"` standard (i.e., supported in any standard xslt stylesheets)?  Or is it that there's no standard way to do this, and you just want to provide a way that it *can* be done?

lifeunleaded:
@tarleb Thank you. I did look for an import of Data.List and couldn't find one, so I thought it would be too intrusive to add. It can certainly be made more readable with Data.List functions.

@jgm No, there is no standard method (that I know of). In the issue descriptions people mentioned "workarounds" just to be able to find these sections, so that's the intention of this: The XPath selector to identify the sections that are supposed to be unnumbered is deterministic, but the work to then act accordingly with section labelling has to be done in the stylesheets.

jgm:
No problem to use Data.List.

lifeunleaded:
@tarleb @jgm On second thought, this is a trivial filter to create since the class token already comes in from the MD reader, and we can put attributes pairs that create XML attributes. Maybe just write it and add to the filters repo and point to it? That would allow anyone to tailor which attribute and token to use, if they already have a stylesheet. My Lua is not stellar, but something like below would do it? That would remove the need to hardcode it in entirely.

```
local function contains(table, val)
   for i=1,#table do
      if table[i] == val then 
         return true
      end
   end
   return false
end

function Header (hdr)
    if contains(hdr.classes, 'unnumbered'
    then
        hdr.attributes = {{'role','unnumbered'}}
    end
    return hdr
end
```

tarleb:
Pandoc already does quite a few slightly opinionated things in the writer so I think that making the change in the writer would be fine.

Slightly shorter filter:

``` lua
function Header (hdr)
  if hdr.classes:includes 'unnumbered' then
    hdr.attributes.role = 'unnumbered'
    return hdr
  end
end
```

jgm:
I agree, I think it's fine to change this in the writer.

lifeunleaded:
Thank you. I will rewrite the enrichRole function and update.

lifeunleaded:
@tarleb It is now more succinct. The Prelude lookup sufficed for this. To be honest I'm not sure it's more readable. I'll gladly take further pointers on changing it.

src/Text/Pandoc/Writers/DocBook.hs
```diff
@@ -464,6 +465,14 @@ idAndRole (id',cls,_) = ident <> role
     ident = [("id", id') | not (T.null id')]
     role  = [("role", T.unwords cls) | not (null cls)]
 
+-- Used in blockToDocBook for Header (section) to create or extend
+-- the role attribute with candidate class tokens
+enrichRole :: [(Text, Text)] -> [Text] -> [(Text, Text)]
+enrichRole mattrs cls = [("role",rolevals) | rolevals /= ""]<>(filter (\x -> (fst x) /= "role") mattrs)
+  where
+    rolevals = T.unwords((filter (`elem` cand) cls)<>(maybeToList(lookup "role" mattrs)))
+    cand = ["unnumbered"]
```
tarleb, line 474: Nice, thanks! This is good to merge from my POV.

Your comments on readability made me wonder, and I ended up writing an (untested!!) version trying to strike a balance between readability and and conciseness. Not sure if I succeeded, but here we go:

``` haskell
enrichRole :: [(Text, Text)] -> [Text] -> [(Text, Text)]
enrichRole mattrs cls = [("role", T.unword roles) | not (null roles)] <> nonRole
  where
    (roleAttr, nonRole) = partition (\(key, _v) -> k == "role") mattrs
    roles = nub $ ["unnumbered" | "unnumbered" `elem` cls] <> map snd roleAttr
```

lifeunleaded: @tarleb Thank you for the suggestion! I'm fine either way, although having `cand` separate was something I hoped to keep, as I can envision there are more class tokens that would make sense to add as role tokens in DocBook, and adding to that list is (perhaps) less intimidating to future contributors. Let me know if you would prefer it changed, and if not, if there's more to do before merge (rebase on current master?)

tarleb: Agreed, keeping the list of "role classes" separate seems sensible.

I just noticed that some lines break the 80 chars limit, we generally want to keep lines below that length. Once that's fixed we can squash-merge everything into a single commit; no need for a rebase.

lifeunleaded: @tarleb Sorry, didn't get to the 80char fix before it merged. Will keep it in mind in the future.

tarleb: No problem at all. Thanks again, looking forward to the next one!

PR #8485, opened 2022-12-13
Lua: add basic support for chunks handling

tarleb:


tarleb:
I'm not sure if we should include this and whether the interface of chunks is likely to change.

jgm:
Not sure if it will change -- I think it's a decent interface, do you see improvements?

I had meant to try to implement chunked HTML output before the 3.0 release, but haven't gotten around to it yet.

tarleb:
I agree, it's a solid interface Haskell-wise. I think I'm a little worried to expose things not intended to be used by third parties, but the main issue is that I dislike parts of my implementation.

For example, I'm slightly unhappy with my current choice of placing this in the `pandoc.utils` module: it would be nicer to have this as a method on `Pandoc` objects, as in `doc:split {path_template = '%s-%n.xhtml'}`, but this isn't easy to do right now. What do you think about putting this into a new Lua module `pandoc.sections`, possibly together with `make_sections` currently in `pandoc.utils`? Maybe we can add a toc-generating function in there as well.

tarleb:
I'm happy with this now. Still open to suggestions of course, esp. about naming.

The `split_into_chunks` function is now in a new `pandoc.structure` module. I've also added `table_of_contents` and `make_sections` to that module. The latter is a reworked, friendlier version of the now deprecated function `pandoc.utils.make_sections`.

tarleb:
I took the new module for the proverbial test ride, and I'm quite pleased with the result. I had to add some slide-related functionality though, causing an additional API change. The result is here: <https://github.com/tarleb/psht>. It's still very much WIP.

jgm:
The Chunks API has changed, so this will need to be updated.

tarleb:
Done

doc/lua-filters.md
```diff
@@ -2403,6 +2403,59 @@ Usage:
...
+:   target filepath for this chunk (string)
+
+`up`
+:   link to the enclosing section, if any (string|nil)
+
+`prev`
+:   link to the previous section, if any (string|nil)
+
+`next`
+:   link to the next section, if any (string|nil)
```
jgm: Are these comments still accurate (also see the module below)?  `chunkNext` is now a `Maybe Chunk`, not just a link.

jgm: Oh, it looks like you changed this in the last commit.
I wonder if it would be worth integrating those changes into this commit, so that the change isn't spread between two commits that aren't obviously related?

tarleb: I must have made a mistake in one of my rebases, these were not intended to be separate. I'll fix it.

tarleb: Done

PR #8488, opened 2022-12-16
changed dash character

BebeSparkelSparkel:
Caused a runtime error `hGetContents: invalid argument (invalid character)`

### DIFF
--- a/data/templates/default.html5
+++ b/data/templates/default.html5
@@ -16,7 +16,7 @@ $endif$
 $if(description-meta)$
   <meta name="description" content="$description-meta$" />
 $endif$
-  <title>$if(title-prefix)$$title-prefix$ – $endif$$pagetitle$</title>
+  <title>$if(title-prefix)$$title-prefix$ - $endif$$pagetitle$</title>
   <style>
     $styles.html()$
   </style>


jgm:
This sounds like an encoding problem.  As long as the template is UTF-8 encoded, pandoc should handle the en dash just fine.  My guess is that you've saved your template in a non-UTF-8 encoding.  What OS?

If we change this at all, I'd suggest changing it to `&#8211;`.  The only argument I can see for a change is that someone might want to use pandoc with `--ascii` and would not expect a non-ascii character from the template.

jgm:
Also what pandoc version?

BebeSparkelSparkel:
I used wget to get the file from the pandoc-template repo which says pandoc 2.19.2 .

BebeSparkelSparkel:
OS: OpenBSD 7.2

BebeSparkelSparkel:
I tried `iconv -t UTF-8 templates/default.html5` and it too failed on that character.

jgm:
I'm not sure what is going on with your system, but the file in the repository is definitely UTF-8 encoded.
```
$ file default.html5
default.html5: HTML document text, Unicode text, UTF-8 text
```
You can also get the template using pandoc itself: `pandoc -D html5 -o mytemplate.html`.

BebeSparkelSparkel:
Thanks

PR #8490, opened 2022-12-16
added Wrapper type documentation to Text.Pandoc.Templates

### LINKED ISSUE #8489: `compileTemplate` not compatible with Pandoc* monads
https://github.com/renjianxiongqi/pandoc/commit/f36b5b8a5db8479bcf98827c267840304403bdf0 exported a few functions from doctemplates,  including `Text.Pandoc.Templates.compileTemplate`, that were not made compatible with the pandoc monads PandocIO and PandocPure.

The commit message suggests:
```text
 To compile a template and allow partials to be found locally
    (either on the file system or via HTTP, in the event that
    the main template has an absolute URL), ue
    `runWithPartials (compileTemplate ...)`
```
but this was not explained in the documentation and seems to break form with the other exported template utility functions.

I suggest to update the api so that `compileTemplate` and ... so that they can be used without wrappers with the Pandoc* monads.

#5979

BebeSparkelSparkel:
Added more information on the wrappers from the commit https://github.com/renjianxiongqi/pandoc/commit/f36b5b8a5db8479bcf98827c267840304403bdf0

closes #8489

### DIFF
--- a/src/Text/Pandoc/Templates.hs
+++ b/src/Text/Pandoc/Templates.hs
@@ -12,6 +12,24 @@
    Portability : portable
 
 Utility functions for working with pandoc templates.
+
+`WithDefaultPartials` and `WithPartials` are Monad wrappers.
+Wrapping these around an instance of `PandocMonad` gives
+different instances of `TemplateMonad`, with different
+search behaviors when retrieving partials.
+
+To compile a template and limit partial search to pandoc's
+data files, use `runWithDefaultPartials (compileTemplate ...)`.
+
+To compile a template and allow partials to be found locally
+(either on the file system or via HTTP, in the event that
+the main template has an absolute URL), ue
+`runWithPartials (compileTemplate ...)`.
+
+`getTemplate` seeks a template locally,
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Templates.hs
```diff
@@ -12,6 +12,24 @@
...
+To compile a template and allow partials to be found locally
+(either on the file system or via HTTP, in the event that
+the main template has an absolute URL), ue
+`runWithPartials (compileTemplate ...)`.
+
+`getTemplate` seeks a template locally,
+ or via HTTP if the template has an absolute URL,
+ falling back to the data files if not found.
+
 -}
```
jgm, line 33: Remember that Haskell comments use haddock markup, not markdown.
Fortunately you should be able to use pandoc itself to translate.

jgm: A hand translation would be even better, e.g. using
```
'getTemplate'
```
instead of
```
@getTemplate@
```
will give a link in the code to the definition.

PR #8493, opened 2022-12-17
ConTeXt writer: support syntax highlighting for code.

tarleb:


jgm:
I published new versions of skylighting, skylighting-core, and skylighting-format-context on Hackage, so you can depend on those.

PR #8494, opened 2022-12-18
Refactor scripting engine

tarleb:
Supersedes, closes: #8431.


jgm:
This looks good. Are there any changes to the public API?  If so, they should be clearly documented in the commit message.

tarleb:
OK done.

Now that custom readers are both searched in `DATADIR/custom`, should we check that directory for filters as well, or stick with the old `filters` dir? I think checking both would make sense.

jgm:
I think for now we can just check `filters`.  I'd like to keep the conceptual separation between filters and custom readers/writers, even though they're getting closer together.

PR #8495, opened 2022-12-18
add required variables template error to Text.Pandoc.Writers.HTML

BebeSparkelSparkel:
Allows an error to be thrown if a required variable is missing from the context. This has only been applied to the HTML writer but could be added to all the writers. I do not think this is a breaking change since existing templates do not have required variables. It requires the new features in doctemplates pull request https://github.com/jgm/doctemplates/pull/19

cabal.project with the required variable doctemplate https://github.com/BebeSparkelSparkel/pandoc/blob/required-var-html-my-build/cabal.project

### DIFF
--- a/src/Text/Pandoc/Templates.hs
+++ b/src/Text/Pandoc/Templates.hs
@@ -19,13 +19,15 @@ module Text.Pandoc.Templates ( Template
                              , WithPartials(..)
                              , compileTemplate
                              , renderTemplate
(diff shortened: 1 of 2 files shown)


jgm:
Can you add something to cabal.project and stack.yaml so that they will use the modified version of doctemplates?  Otherwise the tests won't run here.
There are examples in both files already of the sort of thing that is needed.

PR #8497, opened 2022-12-19
Opt: include more fields in JSON representation of Opt values

tarleb:
I'm not sure if those were left out intentionally?

### DIFF
--- a/src/Text/Pandoc/App/Opt.hs
+++ b/src/Text/Pandoc/App/Opt.hs
@@ -184,16 +184,16 @@ instance FromJSON Opt where
        <$> o .:? "tab-stop" .!= optTabStop defaultOpts
        <*> o .:? "preserve-tabs" .!= optPreserveTabs defaultOpts
        <*> o .:? "standalone" .!= optStandalone defaultOpts
-       <*> o .:? "from"
-       <*> o .:? "to"
+       <*> o .:? "from" .!= optFrom defaultOpts
+       <*> o .:? "to" .!= optTo defaultOpts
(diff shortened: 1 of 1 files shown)


jgm:
I think they were; the idea is that if these get Nothing values, they'll be treated as defaults anyway.
Given that, is there any reason to fill in the defaults in JSON parsing?

tarleb:
I experimented with a `pandoc.cli.convert` Lua function that would receive an `Opt` value, using `peekViaJSON` to retrieve the value. I was confused when not all values made it through.

I'm not sure yet what to think of that experiment, so this change might not be necessary.

jgm:
It shouldn't be necessary if you're just passing the Opt back into convertWithOpts.

jgm:
Well, it is odd that some fields have the default inserted here, and others don't. I don't really remember why and will have to have a closer look.

src/Text/Pandoc/App/Opt.hs
```diff
@@ -184,41 +184,41 @@ instance FromJSON Opt where
        <$> o .:? "tab-stop" .!= optTabStop defaultOpts
        <*> o .:? "preserve-tabs" .!= optPreserveTabs defaultOpts
        <*> o .:? "standalone" .!= optStandalone defaultOpts
-       <*> o .:? "from"
-       <*> o .:? "to"
+       <*> o .:? "from" .!= optFrom defaultOpts
+       <*> o .:? "to" .!= optTo defaultOpts
```
jgm, line 188: optForm and optTo defaultOpts are both Nothing, so adding this has no effect. Perhaps that's why I left it off.
Similarly for optTemplate.
Haven't checked all the others.

Issue #8499, opened 2022-12-19
Using U+2029 character as separator causes error with PDF conversion. 


cderv:
Take this document with special `title`

````markdown
---
title: "\\vspace{-1in} Untitled"
author: "Johann Gutenberg"
date: "1455-12-16"
---

## Document begins

Lorem ipsum dolor sit amet
````

Using nightly version of Pandoc, we have the following issue 
````
> pandoc --to pdf -f markdown --standalone -o test.pdf test.md
Error producing PDF.
! LaTeX Error: Unicode character ÔÇ® (U+2029)
               not set up for use with LaTeX.

> pandoc --version
pandoc.exe 3.0-nightly-2022-12-19
````

With Pandoc 2.18, it was working correctly. Since Pandoc 2.19, there is a pilcrow character added (which was reported in https://github.com/rstudio/rmarkdown/issues/2437 and in Pandoc at https://github.com/jgm/pandoc/issues/8475

Report at #8475 lead to a change by @tarleb to use PARAGRAPH SEPARATOR to separate block in inlines (https://github.com/jgm/pandoc/commit/e0e60871458329679c81cf8c589199d4b52922f8). 

This now create the error above when converting with LaTeX . Unicode character U+2029 does not seem to be correctly handle in LaTeX.

**Comment by tarleb** [comment]
We could add this to the LaTeX template:

````
```{=latex}
\ifPDFTeX
  \DeclareUnicodeCharacter{2029}{ }
\else
  \catcode`^^^^2029=\active
  \protected\def ^^^^2029{ }
\fi
```

cderv:
Yes this works ok 
````markdown
---
title: "\\vspace{-1in} Untitled"
author: "Johann Gutenberg"
date: "1455-12-16"
header-includes: |
  ```{=latex}
  \ifPDFTeX
    \DeclareUnicodeCharacter{2029}{ }
  \else
    \catcode`^^^^2029=\active
    \protected\def ^^^^2029{ }
  \fi
  ```
---

## Document begins

Lorem ipsum dolor sit amet
````

jgm:
Alternatively, we could go back to the pilcrow, or use something else (just a newline)?
Note that in this case, the pilcrow is undesirable even if it can render as LaTeX; we don't want a literal pilcrow here.

Another workaround---tell pandoc explicitly that you want the vspace in an inline context. (Untested)
````
title: >
  `\vspace{-1in}`{=latex}Untitled
````

[...]

PR #8501, opened 2022-12-20
Shared: use LineBreak as default block sep in blocksToInlines

tarleb:
This change also affects the `pandoc.utils.blocks_to_inlines` Lua
function.

Closes: #8499


jgm:
Thanks, I think I like this one better -- it should give better out of box results in more formats, and we don't have to add a lot of extra stuff to templates.

PR #8517, opened 2023-01-03
INSTALL: mention nightly builds

mustafa0x:


jgm:
I intended not to mention these, because I don't want to encourage people to use nightlies.
So, I'm not sure about this.

mustafa0x:
Ah, I see, sounds good!

PR #8525, opened 2023-01-05
Fix MediaWiki reader on internalLink for JCK languages

liruqi:
This PR would fix wrong link text for Chinese or JCK languages which words are not seperated by blank space. Could anyone share the purpose for this `linktrail` variable?

For example:

`是[[台灣|台式]]的傳統[[月餅]]`

would be parsed as below, which is abviously unreasonable.

```
        {
          "t": "Str",
          "c": "是"
        },
        {
          "t": "Link",
          "c": [
            [
              "",
              [],
              []
            ],
            [
              {
                "t": "Str",
                "c": "台式的傳統"
              }
            ],
            [
              "台灣",
              "wikilink"
            ]
          ]
        },
        {
          "t": "Link",
          "c": [
            [
              "",
              [],
              []
            ],
            [
              {
                "t": "Str",
                "c": "月餅"
              }
            ],
            [
              "月餅",
              "wikilink"
            ]
          ]
        },

```

After this fix, json output would be:

```
        {
          "t": "Str",
          "c": "是"
        },
        {
          "t": "Link",
          "c": [
            [
              "",
              [],
              []
            ],
            [
              {
                "t": "Str",
                "c": "台式"
              }
            ],
            [
              "台灣",
              "wikilink"
            ]
          ]
        },
        {
          "t": "Str",
          "c": "的傳統"
        },
        {
          "t": "Link",
          "c": [
            [
              "",
              [],
              []
            ],
            [
              {
                "t": "Str",
                "c": "月餅"
              }
            ],
            [
              "月餅",
              "wikilink"
            ]
          ]
        },

```

Original data: 
https://github.com/chinapedia/mediawiki-to-gfm/commit/40b17e7232fecc7c3d3e9f8ad066ab2ca2383830

Json output with fix:
https://github.com/chinapedia/mediawiki-to-gfm/commit/63add8cb8680247a581bbdb3be3184aacdad8002

### DIFF
--- a/src/Text/Pandoc/Readers/MediaWiki.hs
+++ b/src/Text/Pandoc/Readers/MediaWiki.hs
@@ -664,8 +664,7 @@ internalLink = try $ do
              -- [[Help:Contents|] -> "Contents"
              <|> return (B.text $ T.drop 1 $ T.dropWhile (/=':') pagename) )
   sym "]]"
-  linktrail <- B.text <$> manyChar letter
-  let link = B.link (addUnderscores pagename) "wikilink" (label <> linktrail)
+  let link = B.link (addUnderscores pagename) "wikilink" label
   if "Category:" `T.isPrefixOf` pagename
      then do
        updateState $ \st -> st{ mwCategoryLinks = link : mwCategoryLinks st }
--- a/src/Text/Pandoc/Readers/MediaWiki.hs
+++ b/src/Text/Pandoc/Readers/MediaWiki.hs
@@ -19,7 +19,7 @@ module Text.Pandoc.Readers.MediaWiki ( readMediaWiki ) where
 
 import Control.Monad
 import Control.Monad.Except (throwError)
-import Data.Char (isDigit, isSpace)
+import Data.Char (isAscii, isDigit, isLetter, isSpace)
 import qualified Data.Foldable as F
 import Data.List (intersperse)
 import Data.Maybe (fromMaybe, maybeToList)
@@ -664,7 +664,8 @@ internalLink = try $ do
              -- [[Help:Contents|] -> "Contents"
              <|> return (B.text $ T.drop 1 $ T.dropWhile (/=':') pagename) )
   sym "]]"
-  let link = B.link (addUnderscores pagename) "wikilink" label
+  linktrail <- B.text <$> manyChar (satisfy (\c -> isAscii c && isLetter c)) 
+  let link = B.link (addUnderscores pagename) "wikilink" (label <> linktrail)
   if "Category:" `T.isPrefixOf` pagename
      then do
        updateState $ \st -> st{ mwCategoryLinks = link : mwCategoryLinks st }


liruqi:
I checked the wikipedia rendering rules for English, for example: https://en.wikipedia.org/wiki/100-year_flood

`[[drainage basin]]s
`

would be rendered as 

`<a href="/wiki/Drainage_basin" title="Drainage basin">drainage basins</a>
`

This is a special treatment on wikipedia wikitext parser, which would apply to English (or Latin languages) wikipedia, but not to CJK languages. Probably we should not apply this treatment if pandoc is not aware of input language.

jgm:
Here's the link to their syntax description for "blending" links:

https://en.wikipedia.org/wiki/Help:Wikitext#Blend_link

Note the exception, which we currently don't implement:

> Exception: a trailing [apostrophe](https://en.wikipedia.org/wiki/Apostrophe) (') and any characters following the apostrophe are not blended.

jgm:
But I did some experiments on Wikipedia's sandbox, and their actual implementation of blending seems to be much simpler.  Only ASCII letters are taken to be part of the "link trail."  The blended content stops as soon as you get a punctuation mark or a non-ASCII character, even an accented latin character.

I think we should try to match this behavior -- and it will handle your case well.
Do you want to update your PR?

liruqi:
Yes. It looks like a appropriate solution.

liruqi:
Hi @jgm, please help to review my code changes.

jgm:
Looks great!

waldyrious:
> I did some experiments on Wikipedia's sandbox, and their actual implementation of blending seems to be much simpler. Only ASCII letters are taken to be part of the "link trail." The blended content stops as soon as you get a punctuation mark or a non-ASCII character, even an accented latin character.

For the record, the [mediawiki documentation](https://www.mediawiki.org/wiki/Help:Links#linktrail) indicates that the definition of what content after a wikilink is considered part of it is indeed language-specific. What this means, in practice, is that the specification is encoded in the language's `MessagesXx.php` file, namely in a variable called [`$linkTrail`](https://doc.wikimedia.org/mediawiki-core/REL1_39/php/classLanguage.html#a0bf1c2ad954fd06f5ac1389331dc4640). For example, [here](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesEn.php#528)'s the English configuration file (which should match the experience at the English Wikipedia Sandbox), and [here](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesZh_hans.php#415)'s the Chinese configuration file (which should match the experience at the [Chinese Wikipedia Sandbox](https://zh.wikipedia.org/wiki/Wikipedia:%E6%B2%99%E7%9B%92)).

I'm not 100% sure, but the issue this PR raised (that CJK languages don't separate words with whitespace) may be addressed in MediaWiki's [segmentByWord()](https://doc.wikimedia.org/mediawiki-core/REL1_39/php/classLanguage.html#af32a61cc2d3a15a3128a088c4ebfdc26) method —which, according to the docs, is (re)implemented in the [zh_hans](https://doc.wikimedia.org/mediawiki-core/REL1_39/php/classLanguageZh__hans.html#a69d65ed369716fc2d802cdbe278c1367), [yue](https://doc.wikimedia.org/mediawiki-core/REL1_39/php/classLanguageYue.html#a6c402a51790a52a6fc0f32893432d2c4), and [ja](https://doc.wikimedia.org/mediawiki-core/REL1_39/php/classLanguageJa.html#a5715c986dcc600ee4308f4428b9f3f12) configuration files— in addition to any language-specific `$linkTrail` configuration.

lwolfsonkin:
This issue has just come to my attention. While it is laudable that this fixed an issue parsing documents from Chinese Wikipedia, it seems to have also broken tons of other languages, as partially pointed out by @waldyrious. That is, this likely fixed links for languages using non-segmented scripts, such as Chinese, Japanese, Thai, while simultaneously breaking similar cases in every other language (except English which was specifically tested and given exception).

While on English wikipedia, [A through Z is the acceptable "link trail"](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesEn.php#528), on Spanish Wikipedia [this also includes the letters in the Spanish alphabet which can carry diacritics, namely á, é, í, ó, ú, and ñ](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesEs.php#307). Similar cases exist on [Italian](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesIt.php#227), [French](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesFr.php#302), [Portuguese](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesPt.php#112), and [German](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesDe.php#355) Wikipedias. This is then even more egregious in languages using a segmented script that isn't Latin, such as on [Arabic](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesAr.php#470), [Russian](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesRu.php#428), [Hebrew](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesHe.php#10), [Hindi](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesHi.php#298), [Tamil](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesTa.php#82), and [Urdu](https://gerrit.wikimedia.org/g/mediawiki/core/+/549cb9583f36d6d3c1aa1acd4b2c93c1248694fe/languages/messages/MessagesUr.php#368) Wikipedias.

In short, while it seems to me that the "proper" solution to this would be to have to have rules for each sub-dialect of Wikitext, a much simpler and likely sufficient solution which would likely fix nearly all cases would be to default-allow this blended link behavior for all trailing characters, unless the trailing characters belong to one of the unsegmented scripts (Chinese characters, Thai, Lao, etc.). This can be easily tested by checking the script of the following characters (Each character in Unicode belongs to a script, [each of which are defined by ISO 15924](https://www.unicode.org/iso15924/iso15924-codes.html)). While I'm not sure off-hand how to access this information in Haskell, if one can do this easily via some package, this would likely be the easiest and most inclusive solution (with the caveat that more scripts should be added the this list of unsegmented scripts, such as Khmer and Balinese for example).

I hope this is helpful and hope we can see this solved. Thank you!

Pinging @liruqi and also @jgm.

jgm:
@waldyrious @lwolfsonkin thanks for bringing this to my attention. Could you open a new issue about this, so we don't lose track?

jgm:
I've added a commit along the lines of @lwolfsonkin 's suggestion. No need to open a new issue if this is adequate.

PR #8548, opened 2023-01-14
Remove Null constructor on Block.

tarleb:
Co-authored-by: Albert Krewinkel <albert@zeitkraut.de>


tarleb:
Whoops, that happened already. :tada:

jgm:
I don't think I caught the lua documentation one, so you may need a separate commit for that.

PR #8554, opened 2023-01-17
Report more reference diagnostics in markdown and textile

mgsloan:
* Markdown reader now reports unused link reference definitions and broken link/header references

* Textile reader now reports diagnostics instead of failing


mgsloan:
Note that this is a breaking change of the `LogMessage` datatype and json format.  Is that ok?

mgsloan:
Quite a few test failures!  Looking into it, makes sense to wait on reviewing

mgsloan:
Ready for review!  There is one thing unresolved here, there is a single reference with key `also broken` but it is reported twice and the 2nd report `SourcePos` involves `_chunk` indicating something to do with `parseFromString`.  Generally it seems like `parseFromString` is messing up `SourcePos` which seems quite nonideal for diagnostics.

Also note that this does not improve diagnostics for RST.  I have WIP changes for that, but they aren't ready yet.

jgm:
I can explain the `parseFromString` stuff.  The way the markdown parser deals with certain constructions (here I focus on block quotes, but it's similar for list items) is to parse the lines beginning with `>`, then strip off the prefix, and parse the string content that results.  You're right, this isn't good for source positions.  The commonmark parser, which I wrote later in light of experience, uses a different approach and doesn't need `parseFromString`, giving us more accurate positions.

jgm:
Sorry I've let this languish so long.  Sometimes I can't keep up with the pace.
Can you rebase to fix the conflict in the RST reader?

PR #8555, opened 2023-01-17
Add runPureWithLogMessages for also getting log messages

mgsloan:


jgm:
Is this necessary?  With the current interface you can always do
``` hs
runPure $ do
  result <- yourAction
  messages <- getLog
  return (result, messages)
```
or more concisely
```
runPure $ (,) <$> yourAction <*> getLog
```

mgsloan:
Wow, can't believe I didn't think of that. Thanks!

PR #8585, opened 2023-01-29
Add support for Textile footnote backlinks

alts:
Backlinking from footnotes is already the behavior of `Note` values when writing HTML. This adds parsing support for the explicit backlinking syntax in Textile as described here:
https://textile-lang.com/doc/footnotes

Previously, these lines would be interpreted as a new paragraph beginning with "fn2^." in text.

### DIFF
--- a/src/Text/Pandoc/Readers/Textile.hs
+++ b/src/Text/Pandoc/Readers/Textile.hs
@@ -90,7 +90,10 @@ parseTextile = do
   Pandoc nullMeta . B.toList <$> parseBlocks -- FIXME
 
 noteMarker :: PandocMonad m => TextileParser m Text
-noteMarker = skipMany spaceChar >> string "fn" >> T.pack <$> manyTill digit (char '.')
+noteMarker = do
+  skipMany spaceChar
+  string "fn"
+  T.pack <$> manyTill digit (string "." <|> string "^.")
 
 noteBlock :: PandocMonad m => TextileParser m Text
 noteBlock = try $ do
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Readers/Textile.hs
```diff
@@ -90,7 +90,10 @@ parseTextile = do
   Pandoc nullMeta . B.toList <$> parseBlocks -- FIXME
 
 noteMarker :: PandocMonad m => TextileParser m Text
-noteMarker = skipMany spaceChar >> string "fn" >> T.pack <$> manyTill digit (char '.')
+noteMarker = do
+  skipMany spaceChar
+  string "fn"
+  T.pack <$> manyTill digit (string "." <|> string "^.")
```
jgm, line 96: Probably you want `try (string "^.")`?  What if the string contains a `^` not followed by `.`?

jgm:
The Windows build failure is not your problem; I'm not sure what the issue is there.

src/Text/Pandoc/Readers/Textile.hs
alts: Sorry about that! I'm used to attoparsec's non-consuming `<|>`. I've pushed a commit to wrap in `try`.

I wasn't able to produce any inputs (e.g. `"fn3^ aaa"`) that differed between the `try` and non-`try` versions when parsing a whole doc, but the suggestion expresses the syntax rule more accurately. Thank you!

PR #8589, opened 2023-01-29
ODT reader: fix blockquote indent detection

### LINKED ISSUE #3437: Pandoc does not detect quotes in odt files
I'm using `pandoc 1.19.1`. 

Consider a file called `quote.md`:

```
# Title

> this is a quote

This is the body.
``` 

now, convert it to `.odt`:

```
# pandoc quote.md -o quote.odt
```

Looking at the newly generated odt file, pandoc has correctly applied the "Quotations" style to the "this is a quote" part.

Now, let's convert this new `odt` file back to `.md`:

```
# pandoc quote.odt -o quote2.md
```

The `quote2.md` file has lost the information about which part of the text is quotes:

```
Title
=====

this is a quote

This is the body.
```

dkess:
The ODT reader is supposed to detect blockquotes by checking a paragraph style's indentation level. But it's broken for two reasons:

* The parser fails on non-integers. So "1in" will get read as 25mm, but "1.0in" fails. By default, the Quotations style is "0.3937in".
* The reader doesn't check indentation levels of parent styles. In my test documents, LibreOffice often creates child styles for individual paragraphs, so it's important to check parents (ODT files created by the Pandoc ODT writer don't have this issue though).

I added a new test "blockquote2" whose ODT file is generated directly from the corresponding Markdown file with pandoc.

Fixes #3437.

The issue in the first bullet point is why the roundtrip `blockquote2.md` -> `blockquote2.odt` -> `blockquote2.md` doesn't currently work (the existing `blockquote.odt` test must have been created manually with a "2cm" margin).

The issue in the second bullet point (`ContentReader.hs`) can be split into a different commit if necessary. I can also add another test for it, using an ODT file created from LibreOffice.

I think the discussion about style names and localization in #3437 was a red herring. The indentation check is fine in theory, it just wasn't implemented correctly.


jgm:
Excellent, thanks!

PR #8608, opened 2023-02-09
DocBook writer: output identifiers of figure blocks

Flupp:
@reviewer: I am not sure if the id needs to be prefixed with `writerIdentifierPrefix opts` as done in line 190 (this is the only place where the prefix is added AFAICT).

### DIFF
--- a/src/Text/Pandoc/Writers/DocBook.hs
+++ b/src/Text/Pandoc/Writers/DocBook.hs
@@ -320,7 +320,7 @@ blockToDocBook opts (Table _ blkCapt specs thead tbody tfoot) = do
   return $ inTagsIndented tableType $ captionDoc $$
         inTags True "tgroup" [("cols", tshow (length aligns))] (
          coltags $$ head' $$ body')
-blockToDocBook opts (Figure attr capt@(Caption _ caption) body) = do
+blockToDocBook opts (Figure attr@(id', _, _) capt@(Caption _ caption) body) = do
   -- TODO: probably better to handle nested figures as mediaobject
   let isAcceptable = \case
         Table {}  -> All False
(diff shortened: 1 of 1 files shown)


Flupp:
Note: In contrast to the DocBook writer, the DocBook reader already supports the identifier.

Behavior before the change:
```
$ pandoc -f Markdown -t docbook <<<'![caption text](file.svg){#testid}' | pandoc -f docbook -t markdown 
![caption text](file.svg "fig:")
```
Behavior after the change:
```
$ pandoc -f Markdown -t docbook <<<'![caption text](file.svg){#testid}' | pandoc -f docbook -t markdown
![caption text](file.svg "fig:"){#testid}
```

jgm:
Docs for `--id-prefix` say "Specify a prefix to be added to all identifiers and internal links in HTML and DocBook output," so yes, I'd say it should be added in this case too.

Flupp:
I’ve added the prefix. However, now I seriously wonder if the prefix is maybe also missing at other places in the DocBook writer. The following
```md
:::{#bar}
bar div
:::

:::{#sec .section}
sec div
:::

[foo link](#foo)

[span]{#spanid}

![caption text](file.svg){#testid}
```
translates via `pandoc --id-prefix=prefix- -f Markdown -t docbook` (including this change) to
```xml
<para xml:id="bar">
  bar div
</para>
<para xml:id="sec">
  sec div
</para>
<para>
  <link linkend="prefix-foo">foo link</link>
</para>
<para>
  <anchor xml:id="spanid" />span
</para>
<figure xml:id="prefix-testid">
  <title>caption text</title>
  <mediaobject>
    <imageobject>
      <imagedata fileref="file.svg" />
    </imageobject>
    <textobject><phrase>caption text</phrase></textobject>
  </mediaobject>
</figure>
```
Note that `prefix-` is not added to every id.

jgm:
Hi, I'm sorry I let this languish so long.

You're right that the prefix isn't added everywhere it should be.  Perhaps that could be addressed in a separate PR.  Or, feel free to address it here.

I'll add one substantive comment to your code above.

src/Text/Pandoc/Writers/DocBook.hs
```diff
@@ -344,10 +344,13 @@ blockToDocBook opts (Figure attr capt@(Caption _ caption) body) = do
                    DocBook4 -> pure mempty -- docbook4 requires media
                    DocBook5 -> blocksToDocBook opts body
       mediaobjects <- mapM toMediaobject body
+      version <- ask
+      let idAttr = [ (idName version, writerIdentifierPrefix opts <> id')
+                   | not (T.null id') ]
```
jgm, line 349: We have the `idAndRole` function, and I think it would make sense to use it here (and anywhere else where we'd manually be constructing an `id` attribute).

Note also that the `idAndRole` function doesn't use the version-dependent `idName`. That should be fixed.

jgm:
It would also be great to have a test. Simplest way is a "command test" called `test/command/8608.md`. See the other files in that directory for examples.

PR #8609, opened 2023-02-09
Add chromium pdf engine

arsalan0c:
Hi. This PR adds Chromium as a supported HTML pdf engine. 
Chromium's headless mode and pdf printing switch is used.

Example usage:
`cabal run pandoc-cli -- --pdf-engine=chromium -o test.pdf test.html`

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1342,7 +1342,7 @@ header when requesting a document from a URL:
 :   Use the specified engine when producing PDF output.
     Valid values are `pdflatex`, `lualatex`, `xelatex`, `latexmk`,
     `tectonic`, `wkhtmltopdf`, `weasyprint`, `pagedjs-cli`,
-    `prince`, `context`, and `pdfroff`. If the engine is not in
+    `prince`, `context`, `pdfroff` and `chromium`. If the engine is not in
     your PATH, the full path of the engine may be specified here.
     If this option is not specified, pandoc uses the following
     defaults depending on the output format specified using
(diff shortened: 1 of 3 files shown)


jgm:
Thanks! Can you say a bit about the comparative advantages/disadvantages of chromium compared to the other html->pdf engines?

jgm:
Two more places that need updating:

- MANUAL.txt (under `--pdf-engine`)
- src/Text/Pandoc/App/CommandLineOptions.hs htmlEngines line 218

arsalan0c:
> Two issues here:
> 
> * Chromium binaries go by a lot of different names on different systems, and the FOSS builds go by different names than the upstream-built binary Chrome builds. Any attempt to assume one name is bound to bring up questions about why they aren't all supported, and this has no configuration mechanism.
> * Specifically why would it by advantageous to use a headless chromium over and above `paged-js` which is already an option which also uses headless Chromium under the hood but wraps it in a bunch of typesetting related features and style improvements targeting PDFs as a print or at least paged medium? Stock chromium print output leaves a lot to be desired along those lines.

Thanks for pointing these out. 

For the first, I've added a commit ([597515b](https://github.com/jgm/pandoc/pull/8609/commits/597515b677b0209e5adef2a9303f8705ed9e16e7)) to allow the path to the binary to be specified. 

I wasn't aware of `pagedjs-cli` offering improvements over chromium, it does seem to be a better option in every situation. I see motivation for adding chromium as a convenient alternative, in case one already has chromium installed would prefer not to install another engine. Or in the unlikely event there are installation issues with the other engines. This perhaps may not be a compelling reason for the addition although I think the changes are minor and will not add burden for maintenance.

jgm:
I'll close this because paged-js just seems a better option.

PR #8612, opened 2023-02-09
URL style in ConTeXt

twsh:
I have added `\setupurl[style=]` to the ConTeXt template, and updated the relevant tests. This gives the same default behaviour as the LaTeX template, which is that URLs are the same as the main text.

### DIFF
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -23,7 +23,7 @@ $endif$
   style=$linkstyle$,
   color=$linkcolor$,
   contrastcolor=$linkcontrastcolor$]
-\setupurl[style=]
+\setupurl[style=$urlstyle$]
 
 % make chapter, section bookmarks visible when opening document
 \placebookmarks[chapter, section, subsection, subsubsection, subsubsubsection, subsubsubsubsection][chapter, section]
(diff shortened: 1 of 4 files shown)


data/templates/default.context
```diff
@@ -23,6 +23,7 @@ $endif$
   style=$linkstyle$,
   color=$linkcolor$,
   contrastcolor=$linkcontrastcolor$]
+\setupurl[style=]
```
jgm, line 26: shouldn't this have a variable in it?

jgm: That is, `\setupurl[style=$urlstyle$]`?

twsh: It depends what we want. If we just want to make the smallest possible change, and have all links be styled by `$linkstyle$` then we don't need a variable there.

Alternatively, If we want to add support for styling links without link text, like automatic link URLs, then I think that would be the way to do it. I would need to add the `$urlstyle$` variable to the manual too.

jgm: I think adding the variable would be consistent with what we do in the latex template:
```
\urlstyle{$if(urlstyle)$$urlstyle$$else$same$endif$}
```
`urlstyle` is documented in the manual under "Variables for LaTeX."  One could put something similar under the ConTeXt heading.

twsh: OK, I have done that.

jgm:
Great, thanks

PR #8617, opened 2023-02-13
Markdown reader: use title of implicit figure as short caption.

### LINKED ISSUE #7915: Use markdown image title text when generating list of figures
Currently if pandoc is called with `--lof` or configured with a `format.yml` containing:

```yaml
---
lof: yes
```

when converting a markdown source, each figure is listed in the LoF using its markdown alt text. This is a problem as others have pointed out in a [separate issue](https://github.com/jgm/pandoc/issues/2417) because the alt text is also used for image captions, and the text length desired for image captions is very often longer than a short identifying text needed in a list of figures.

Fortunately markdown offers a _title text_ attribute for images which offers a practical and natural solution. The proposal here is to use the mardown image title when generating the list of figures. Usage would look like this:

```markdown
[My image **caption** can be long with _formatting_.](figures/my_figure.jpg "Figure title for LoF entry")
```

The title attribute is simple text only (no formatting or math, etc.) which makes it perfect to use in an LoF entry where you don't want that anyway.

This solution allows for a semantic separation between image caption (markdown alt text) and LoF entry text (markdown title text)

tarleb:
The title of an implicit figure, if set, is used as the short caption of
a figure. The short caption of a figure replaces the full caption in the
list of figures.

Closes: #7915

### DIFF
--- a/src/Text/Pandoc/Readers/Markdown.hs
+++ b/src/Text/Pandoc/Readers/Markdown.hs
@@ -1092,7 +1092,9 @@ implicitFigure (ident, classes, attribs) capt url title =
               _               -> capt
       attribs' = filter ((/= "alt") . fst) attribs
       figattr = (ident, mempty, mempty)
-      caption = B.simpleCaption $ B.plain capt
+      caption = if T.null title
+                then B.simpleCaption $ B.plain capt
+                else B.caption (Just . B.toList $ B.text title) $ B.plain capt
       figbody = B.plain $ B.imageWith ("", classes, attribs') url title alt
   in B.figureWith figattr caption figbody
 
--- a/src/Text/Pandoc/Writers/Markdown.hs
+++ b/src/Text/Pandoc/Writers/Markdown.hs
@@ -672,21 +672,22 @@ blockToMarkdown' opts (Figure figattr capt body) = do
   let combinedAttr imgattr = case imgattr of
         ("", cls, kv) | (figid, [], []) <- figattr -> Just (figid, cls, kv)
         _ -> Nothing
-  let combinedAlt alt = case capt of
+  let combinedAlt alt title = case capt of
         Caption Nothing [] -> if null alt
                               then Just [Str "image"]
                               else Just alt
-        Caption Nothing [Plain captInlines]
-          | captInlines == alt -> Just captInlines
+        Caption short [Plain captInlines]
+          | captInlines == alt
+          , title == maybe "" stringify short -> Just captInlines
         _ -> Nothing
   case body of
     [Plain [Image imgAttr alt (src, ttl)]]
       | isEnabled Ext_implicit_figures opts
-      , Just descr    <- combinedAlt alt
+      , Just descr    <- combinedAlt alt ttl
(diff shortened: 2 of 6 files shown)


jgm:
I think my main question is still the one raised in #7915 -- what if formatting or math is needed in the short caption?

One approach to this would be to parse the title as markdown.

tarleb:
Another thought: The title could also be used as the image's `alt` text.  This would be more consistent with the fact that titles cannot contain markup. I'd give accessibility higher priority than convenience for short captions, and this could be a good way to specify a description that differs from the figure caption. OTOH, it might be confusing to have implicit figures and normal images behave so differently.

jgm:
Yes, I agree both that this makes sense and that it's probably too confusing to have implicit figures and normal images behave differently.

We can still specify alt text using an attribute, no?
```
![the caption](file.jpg){alt="My alt text"}
```

jgm:
I suppose one alternative to overloading the title would be to allow a short caption to be specified via an attribute:
```
![the caption](file.jpg){shortcaption="My short caption"}
```

Drawback: English keyword.  Drawback: no formatting.

tarleb:
Am I right in thinking that we currently only special treatment for English keywords in the writers, but not the Markdown reader? It might be a dangerous precedent.

Would it make sense to support the table caption syntax on images?

``` markdown
![Close-up shot of the colorful face of a male mandrill.](mandrill.jpg)

: A male mandrill
```

Then the problem would become the more general issue of how to add short captions with that syntax. I guess the downside is that this is ambiguous syntax that could also be intended to mark a description list with an image term?

jgm:
Hm, that does make a lot of sense.  I agree that it's ambiguous, and that's a strike against it, but it is still tempting, and how common is it to use a definition list with an image term?

I would suggest that if we did this, then when this explicit caption is present, the image description should be the short caption, and the thing after `:` should be the long caption (since here we can include block-level content).

tarleb:
If we could find a way to specify the short caption by extending the long caption syntax, then could use that with tables as well. E.g., something like

```
![](a.png)

:short caption:
    long caption.
```

or, inspired by org-mode description lists,

```
![](a.png)

: short caption :: long caption.
```

Fuller example:

``` markdown
![The colorful face of a male mandrill.](mandrill.jpg)

: "Mandrill" test image :: The "Mandrill" standard test image, sometimes
  erroneously called "Baboon", is a popular sample photo used in image
  processing research.
```

Maybe this could also open a path towards more general, multi-image figures, but I don't know yet how that might look like.

jgm:
```
: "Mandrill" test image :: The "Mandrill" standard test image, sometimes
  erroneously called "Baboon", is a popular sample photo used in image
  processing research.
```
That's not bad, actually!

acxz:
Hello, I've taken an interest in the short/long caption issue. (Although I might not be able to devote time in solving it...)

> ```
> : "Mandrill" test image :: The "Mandrill" standard test image, sometimes
>   erroneously called "Baboon", is a popular sample photo used in image
>   processing research.
> ```

Would this solve the problem of having specific formatting in the short caption?

PR #8628, opened 2023-02-15
fixed <xref> references to tables in DocBook files

### LINKED ISSUE #8626: docbook xref to table
This bug is very similar to an already solved bug #8064.

When an _xref_ in a _docbook_  format source document points to a _table_, the text of the link is `table_title`, not the actual table title. The referenced closed issue solves the same problem, but only for a _figure_ element.

Affected versions: all

The solution is very similar to the referenced. Actually I have already prepared it, I can create a PR when you confirm it.

ciklysta:
_xref_ elements can also be used to link to _table_ elements. The DocBook reader did not correctly select the content of the referenced _table_'s _title_. The problem of link name selection is rather complex, as described in #8065. But it seems that the most used target elements have been covered so far. 

Closes #8626.


jgm:
Thanks!

PR #8634, opened 2023-02-18
Update TableBody RowHeadColumns caculation: change from max to min

liruqi:
Recently I noticed that if a HTML table have different number of leading `th` tag in each row, colspan would be lost and some `td` cells would be converted to `th` cells. I looked into the HTML table parsing code, and currently the `TableBody` data struct assume each row have the same number of leading `th` tags.

This PR will set `RowHeadColumns` to the minimum value of each row, to reduce such abnormal colspan cases. Actually I believe this `RowHeadColumns` should be discarded, and we should move `CellType` to `Cell` attribute instead.

For example the Infobox in this page: https://en.wikipedia.org/w/index.php?title=Anna_Sorokin&oldid=1139601226 , the image is in a `td` cell with `colspan="2"`, but for the rows below, each row got one leading `th` cell.

### DIFF
--- a/src/Text/Pandoc/Readers/HTML/Table.hs
+++ b/src/Text/Pandoc/Readers/HTML/Table.hs
@@ -191,10 +191,12 @@ pTableBody block = try $ do
   optional $ pSatisfy (matchTagClose "tbody")
   guard $ isJust mbattribs || not (null bodyheads && null rows)
   let attribs = fromMaybe [] mbattribs
-  return $ TableBody (toAttr attribs) (foldr max 0 rowheads) bodyheads rows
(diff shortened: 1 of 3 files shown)


jgm:
That's a good point about `RowHeadColumns`.  Unfortunately this is something that is very awkward to change now.  But if you'd like to open an issue about that on jgm/pandoc-types, that would be the place.

jgm:
This change sets `RowHeadColumns` to the minimum value of each row,
which gives better results in cases where rows have different numbers of leading th tags.

jgm:
thanks and sorry for late merge!

PR #8641, opened 2023-02-20
asciidoc{,tor} template: fix revision date when author is unset

### LINKED ISSUE #8637: asciidoc writer emits invalid titleblocks when author is missing
**Explain the problem.**

When a document contains a date but no author:
```markdown
---
date: 2020-01-01
title: hello
---
hi
```

conversion `-f` marktown `-t` asciidoc{,tor} [generates](https://pandoc.org/try/?params=%7B%22text%22%3A%22---%5Cndate%3A+2020-01-01%5Cntitle%3A+hello%5Cn---%5Cnhi%5Cn%22%2C%22to%22%3A%22asciidoctor%22%2C%22from%22%3A%22markdown%22%2C%22standalone%22%3Atrue%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A%7B%7D%2C%22template%22%3Anull%7D) a header without an author line:

```asciidoc
= hello
2020-01-01

hi
```

[According to the docs](https://docs.asciidoctor.org/asciidoc/latest/document/revision-line/#when-can-i-use-the-revision-line), the revision syntax is only valid when an author line is present. Thus, this gets parsed incorrectly, where `2020-01-01` is considered to be the author.

The correct output in this case might instead be:

```asciidoc
= hello
:revdate: 2020-01-01

hi
```

**Pandoc version?**
3.1 according to the website, but also tested on pandoc 2.19.2.

arcnmx:
Revision line syntax [is only valid in combination with an author line](https://docs.asciidoctor.org/asciidoc/latest/document/revision-line/#when-can-i-use-the-revision-line). The documentation indicates that [the metadata attributes must be set explicitly](https://docs.asciidoctor.org/asciidoc/latest/document/revision-attribute-entries/) when the author line is missing, so the template has been updated to match.

Fixes #8637

### DIFF
--- a/data/templates/default.asciidoc
+++ b/data/templates/default.asciidoc
@@ -2,9 +2,11 @@ $if(titleblock)$
 = $title$
 $if(author)$
 $for(author)$$author$$sep$; $endfor$
-$endif$
 $if(date)$
 $date$
+$endif$
+$elseif(date)$
+:revdate: $date$
 $endif$
 $if(keywords)$
 :keywords: $for(keywords)$$keywords$$sep$, $endfor$
--- a/data/templates/default.asciidoctor
+++ b/data/templates/default.asciidoctor
@@ -2,9 +2,11 @@ $if(titleblock)$
 = $title$
 $if(author)$
 $for(author)$$author$$sep$; $endfor$
-$endif$
 $if(date)$
 $date$
+$endif$
+$elseif(date)$
+:revdate: $date$
 $endif$
 $if(keywords)$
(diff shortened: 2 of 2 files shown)


jgm:
thank you!

PR #8656, opened 2023-02-24
feat: Add CJKsansfont and CJKmonofont for XeLatex

krahets:
The XeLaTex supports `CJKsansfont` and `CJKmonofont` which is useful to change the CJK fonts for titles and codes.

The following args were tested and worked well.

```
---

mainfont: "Noto Serif SC"
sansfont: "Noto Sans SC"
monofont: "Ubuntu Mono"
CJKmainfont: "Noto Serif SC"
CJKsansfont: "Noto Sans SC"
CJKmonofont: "Noto Sans SC"

...
```

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -176,6 +176,18 @@ $if(CJKmainfont)$
   \fi
 $endif$
 \fi
+$if(CJKsansfont)$
+  \ifXeTeX
+    \usepackage{xeCJK}
+    \setCJKsansfont[$for(CJKoptions)$$CJKoptions$$sep$,$endfor$]{$CJKsansfont$}
+  \fi
+$endif$
+$if(CJKmonofont)$
+  \ifXeTeX
+    \usepackage{xeCJK}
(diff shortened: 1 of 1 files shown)


jgm:
This looks like a great addition. I'd suggest adding these to the existing conditional
```
$if(CJKmainfont)$
  \ifXeTeX
    \usepackage{xeCJK}
    \setCJKmainfont[$for(CJKoptions)$$CJKoptions$$sep$,$endfor$]{$CJKmainfont$}
  \fi
$endif$
```
This way we only usepackage xeCJK once. Of course, it wouldn't work if we specified a CJK mono font but not a CJK main font, but I think it's reasonable to assume that this won't reasonably happen?

Another question is whether we should have something similar for the lualatex path?

krahets:
> This way we only usepackage xeCJK once. Of course, it wouldn't work if we specified a CJK mono font but not a CJK main font, but I think it's reasonable to assume that this won't reasonably happen?

Done. I agree it is safe to assume the users use `CJKmonofont` and `CJKsansfont` with `CJKmainfont`.

> Another question is whether we should have something similar for the lualatex path?

Sorry that I'm not familiar with lualatex. I'm glad to get help from you or other contributors.

Issue #5583, opened 2019-06-14
Adjust EPUB footnotes with backlink before text.


jgm:
See #4491 which has been abandoned but has several good ideas in it, which I don't want to get lost.

[EPUB3 Accessibility Guidelines](http://kb.daisy.org/publishing/docs/html/notes.html) say

- Identify footnotes within the text using the aside element. [WCAG 1.3.1]
- Group footnotes or endnotes at the end of a section or work in lists. [WCAG 1.3.1]
- Link note references to their notes. [WCAG 1.3.1]
- Backlink notes at the end of a section or work to their references. [WCAG 1.3.1]
- For footnotes, the backlink precedes the note text so that the user will be aware that such a link is available.

All we really need to do is change the backlink to precede the text, and make sure all the links and cross-references and containing sections conform to guidelines.

See also:

- [Amazon kindle guidelines](https://kindlegen.s3.amazonaws.com/AmazonKindlePublishingGuidelines.pdf)
- [iBooks Asset Guide](https://help.apple.com/itc/booksassetguide/en.lproj/static.html#itccf8ecf5c8)
- [DPUB-Aria](https://www.w3.org/TR/dpub-aria-1.0/#doc-footnote)

jgm:
Note however that under "backlink with endnote" it recommends putting the backlink at the end, so as not to interfere with note numbering.
```
<li role="doc-endnote">
   <p>According to the usual nomenclature, the …</p>
   <p><a href="#ref" role="doc-backlink">Go to note reference</a></p>
</li>
```

lewer:
Currently, with epub3 writer, the `#footnotes` section has `epub:type="footnotes"` and every footnote has `<li epub:type="footnote">`

As I understand https://kb.daisy.org/publishing/docs/html/notes.html, "footnotes" are what is put in an aside block in the body, while "endnotes" are grouped at the end in a proper section.

So currently pandoc kind of mixes endnotes and foonotes: it uses endnotes but tags them with `epub:type="footnotes"`

To follow https://kb.daisy.org/publishing/docs/html/notes.html with minimal changes, what do you think of replacing `epub:type="footnotes"` with `epub:type="endnotes"` in the `#footnotes` section, add `role="doc-endnotes"` to this same section, and removing `epub:type="footnote"` on each individual footnote?

jgm:
> @Porges concluded that backlink should be placed at end for footnotes, and at start for endnotes.

[...]

PR #8672, opened 2023-03-06
Adjust EPUB footnotes with backlink before text

lewer:
Currently, pandoc footnotes leads to buggy display with kindle (see https://github.com/jgm/pandoc/pull/4491)

This PR aims at fixing https://github.com/jgm/pandoc/issues/5583.

I have taken @Porges code.

The PR makes pandoc's footnotes compliant with https://kindlegen.s3.amazonaws.com/AmazonKindlePublishingGuidelines.pdf (see p. 34, Method 2)

I don't code in haskell, so please review carefully. In particular, the code introduces an empty line between the `hrtag` and the first footnote, empty line that should not be there but that I don't know how to remove. Any help with this? Thanks.


src/Text/Pandoc/Writers/HTML.hs
```diff
@@ -557,13 +557,7 @@ footnoteSection refLocation startCounter notes = do
          container $ do
            nl
            hrtag
-           -- Keep the previous output exactly the same if we don't
-           -- have multiple notes sections
-           if startCounter == 1
```
lewer, line 562: I've assumed that this can be dropped as long as ordered lists are not used anymore, is it really the case?

test/Tests/Writers/HTML.hs
```diff
@@ -127,10 +127,9 @@ tests =
           , "<p>Some more text.</p>"
           , "<div class=\"footnotes footnotes-end-of-document\">"
           , "<hr />"
-          , "<ol>"
-          , "<li id=\"fn1\"><p>Down here.<a href=\"#fnref1\" class=\"footnote-back\">↩︎</a></p></li>"
-          , "<li id=\"fn2\"><p>The second note.<a href=\"#fnref2\" class=\"footnote-back\">↩︎</a></p></li>"
-          , "</ol>"
+          , ""
+          , "<div id=\"fn1\"><p><a href=\"#fnref1\" class=\"footnote-back\">1.</a> Down here.</p></div>"
+          , "<div id=\"fn2\"><p><a href=\"#fnref2\" class=\"footnote-back\">2.</a> The second note.</p></div>"
```
jgm, line 132: I don't think I like this change.  Using ordered lists gives us much better formatting (content of a note is aligned), and it will work well even if the note is, say, a code block.

src/Text/Pandoc/Writers/HTML.hs
jgm: No, it can't be dropped. This plays a significant role when you use `--reference-location=block` (or `section`), which causes the footnotes to be printed at the end of each block (or section).  In that case we need to make sure that the numbering continues appropriately in each new footnote section.

jgm:
Pandoc's current system complies with http://kb.daisy.org/publishing/docs/html/notes.html which says that, when endnotes are used (as with pandoc), the backlink should come at the *end* of the note.  I gather this still causes a problem on Kindle? Can you explain what the problem is, exactly?

I'm willing to consider a system that moves the backlink to the front, although that seems inferior from a reader's perspective.  (After reading the note, you don't want to have to scroll to the beginning of the note to find the backlink.)  Perhaps this could be configurable.

Making the backlink the note number itself requires moving away from the use of `<ol>`, and that has other disadvantages.  (You lose the nice alignment that ols give you, and you have to figure out how to insert the note number into, potentially, any type of block -- what if the note starts with a code block, for example?)  If we wanted to pursue this, I'd say the first step would be to work out what the HTML (and CSS) should look like (in general, not just for a simple paragraph note).

Porges:
> Pandoc's current system complies with http://kb.daisy.org/publishing/docs/html/notes.html which says that, when endnotes are used (as with pandoc), the backlink should come at the _end_ of the note. I gather this still causes a problem on Kindle? Can you explain what the problem is, exactly?
> 
> I'm willing to consider a system that moves the backlink to the front, although that seems inferior from a reader's perspective. (After reading the note, you don't want to have to scroll to the beginning of the note to find the backlink.) Perhaps this could be configurable.
> 
> Making the backlink the note number itself requires moving away from the use of `<ol>`, and that has other disadvantages. (You lose the nice alignment that ols give you, and you have to figure out how to insert the note number into, potentially, any type of block -- what if the note starts with a code block, for example?) If we wanted to pursue this, I'd say the first step would be to work out what the HTML (and CSS) should look like (in general, not just for a simple paragraph note).

The problems with Kindle are described in #4491

jgm:
Yes, but the guidelines recommend backlinks at the *end* for endnotes...like the ones we provide by default.  (See the link above.)

lewer:
> I'm willing to consider a system that moves the backlink to the front, although that seems inferior from a reader's perspective. (After reading the note, you don't want to have to scroll to the beginning of the note to find the backlink.) 

It's also true the other way around: when you click on a note and figure out that you don't want to read it, you don't want to have to scroll to the end of the note to find the backlink ;-)

I don't have an opinion about `div` rather than `ol` or `aside` for footnotes (and it seems that guidelines don't have a clear opinion about that neither...), but kindle is not the only one who struggles displaying correctly pandoc's footnotes. Look at this screenshot taken from Google Play Books: the second endnote appears with note number 1. in the popup



I agree that `ol` offers the advantage of native nice alignment. But epub readers offer popup endnotes/footnotes, which do not seem to go well with `ol`. Even if guidelines differ between kb.daisy.org, amazon, ibooks, they all seem to converge towards using the `aside` tag, as pointed out by @Porges [here](https://github.com/jgm/pandoc/pull/4491#issuecomment-378802283)

What about an option that selects between endnote (default, current behavior) and footnote (`aside` instead of `ol`, backlink at the beginning)?

jgm:
> What about an option that selects between endnote (default, current behavior) and footnote (aside instead of ol, backlink at the beginning)?

That's certainly possible. We could also make the "footnote" behavior the default when generating HTML for epub.  (And maybe if we did that we wouldn't need to make it user-configurable -- though perhaps some people would enjoy having the footnote behavior for web pages too?)

We'd still need to figure out a good way of constructing the footnote div and inserting the number (again, keeping in mind that it won't necessarily begin with paragraph text).  I suppose the simplest thing would be to put the number in a paragraph by itself if the note begins with something other than paragraph text.

lewer:
> We could also make the "footnote" behavior the default when generating HTML for epub. (And maybe if we did that we wouldn't need to make it user-configurable -- though perhaps some people would enjoy having the footnote behavior for web pages too?)

I was thinking the same. Maybe for a first PR, not user-configurable, to keep it simple. And later if needed, an option could be added for that.

> We'd still need to figure out a good way of constructing the footnote div and inserting the number (again, keeping in mind that it won't necessarily begin with paragraph text). I suppose the simplest thing would be to put the number in a paragraph by itself if the note begins with something other than paragraph text.

I agree.

I don't think there's a lot to do, the code written by @Porges already builds the footnote div and putting the note in a separate paragraph isn't complicated. What still needs to be done I think, is adding a "behavior" option to the function which generates the footnote section, and calling it with "behavior = footnotes" from epub writer. Maybe also a bit of CSS to display it nicely.

I can try, but I'm not familiar with Haskell so I don't promise anything. @Porges, do you have time to tackle this it? Otherwise I'll do it.

jgm:
The HTML writer has a state variable `stEPUBVersion` that will have a `Just` value if the writer is being used to produce EPUB. So that's the thing to look at to switch between the behaviors.

jgm:
I'll have a go at this.

jgm:
See new PR #8676 - please test if you can.

jgm:
Now that I've merged the other PR, this one is obsolete.  Thank you!

PR #8676, opened 2023-03-08
HTML writer: Use different structure for epub footnotes.

### LINKED ISSUE #5583: Adjust EPUB footnotes with backlink before text.

jgm:
Many EPUB readers are thrown off by pandoc's current footnote output.  Both the ol and the fact that the footnote backlink is at the end of the note seem to pose problems.

With this commit, we now create a list of aside (or div) elements, instead of an ordered list. Each element begins with a note number that is linked back to the note reference.  (So, the backlink occurs at the beginning rather than the end.)

References:

- https://kindlegen.s3.amazonaws.com/AmazonKindlePublishingGuidelines.pdf
- http://kb.daisy.org/publishing/docs/html/notes.html

Obsoletes #8672, closes #5583.

Thanks to @Porges and @lewer.

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -564,7 +564,7 @@ footnoteSection refLocation startCounter notes = do
              Nothing | startCounter == 1 ->
                (H.ol (nl >> mconcat notes)) >> nl
              Nothing -> (H.ol ! A.start (fromString (show startCounter)) $
-                         mconcat notes) >> nl
+                         nl >> mconcat notes) >> nl
 
 -- | Parse a mailto link; return Just (name, domain) or Nothing.
 parseMailto :: Text -> Maybe (Text, Text)
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -560,12 +560,11 @@ footnoteSection refLocation startCounter notes = do
            -- Keep the previous output exactly the same if we don't
            -- have multiple notes sections
            case epubVersion of
-             Just _ -> mconcat notes >> nl
+             Just _ -> mconcat notes
              Nothing | startCounter == 1 ->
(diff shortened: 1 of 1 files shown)


jgm:
One note: I tried this on macos Books and got a duplicated note number:

Screen Shot 2023-03-07 at 9 49 02 PM

The big note number is apparently being added by Books app.  Let me know what you see on Kindle and other readers.

lewer:
Thanks a lot for this @jgm.

Here's what I have with Kindle (tested with Kindle previewer) and Google Play Books.

The result is great, footnotes appear individually in the popup. Kindle and Google Play Books are able to retrieve the note's number and a title is added to the popup: Footnote number ...
I guess this is what Books also does, but with no "Footnote" prefix, only the number.

Just to let you know: in the epub source, there's an empty line between the `<hr />` that opens the footnote section and the first `<aside>`, and also an empty line between the last `<aside>` and the closing `</section>` tag.

PR #8685, opened 2023-03-13
Integreate Lua warnings in pandoc's logging system.

tarleb:
Lua 5.4 added the option to raise warnings with the `warn` function. This PR adds these warnings to the list of log messages and reports them together with the other logging events.

E.g.

``` lua
warn('This is the last warning!')
print(pandoc.json.encode(PANDOC_STATE.log[#PANDOC_STATE.log]))
```

```
[WARNING] Lua warning at ../testing/test.lua:1:  This is the last warning!
{"message":"This is the last warning!","position":"../testing/test.lua:1: ","type":"LuaWarning","verbosity":"WARNING"}
```

### DIFF
--- a/pandoc-lua-engine/src/Text/Pandoc/Lua/Init.hs
+++ b/pandoc-lua-engine/src/Text/Pandoc/Lua/Init.hs
@@ -23,12 +23,15 @@ import Control.Monad.Trans (MonadIO (..))
 import Data.Maybe (catMaybes)
 import Data.Version (makeVersion)
 import HsLua as Lua hiding (status, try)
-import Text.Pandoc.Class (PandocMonad (..))
+import Text.Pandoc.Class (PandocMonad (..), report)
 import Text.Pandoc.Data (readDataFile)
 import Text.Pandoc.Error (PandocError (PandocLuaError))
+import Text.Pandoc.Logging (LogMessage (ScriptingWarning))
 import Text.Pandoc.Lua.Global (Global (..), setGlobals)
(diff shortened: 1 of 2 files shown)


jgm:
Sounds good.  The parameter is just a text.  Where does the "position" come from in the structured JSON output?
Is it somehow parsing that out of the text?  Would it be better to have a SourcePos as a parameter of the constructor?

tarleb:
The `position` is created with (a wrapper for) the [`luaL_where`](https://www.lua.org/manual/5.4/manual.html#luaL_where) function, hence the slightly weird format. A more fine-grained position would require access to Lua's debugging interface, but I haven't written any wrappers for that yet. But we could parse the line number from the `position` string.

jgm:
I'm not so much concerned with the JSON message but with the pandoc log message constructor -- just asking whether it should have a SourcePos parameter in addition to the Text.  I don't really understand any of the details here, just raising the question.

tarleb:
I'll explore that some more. The one downside would be that the SourcePos wouldn't be entirely precise, as Lua debugging info does not contain any column info. We'd have to just set the column to `1` for all Lua warnings, which seems fine.

tarleb:
The above script would now generate

```
[WARNING] Scripting warning at ../testing/test.lua line 1 column 1: This is the last warning!
{"column":1,"line":1,"message":"This is the last warning!","source":"../testing/test.lua","type":"ScriptingWarning","verbosity":"WARNING"}
```

PR #8687, opened 2023-03-13
use Windows `%APPDATA%` variable (might fix #8686)

### LINKED ISSUE #8686: `%APPDATA%\pandoc` contains default `--data-dir` in Windows
From https://pandoc.org/MANUAL.html#option--data-dir:

> On Windows the default user data directory is `C:\Users\USERNAME\AppData\Roaming\pandoc`.

I think replacing the path with `%APPDATA%\pandoc` is more accurate, since it might cover other `%USERPROFILE%` values.

ousia:


MANUAL.txt
```diff
@@ -376,7 +376,7 @@ header when requesting a document from a URL:
     environment variable).  If that directory does not exist and
     `$HOME/.pandoc` exists, it will be used (for backwards compatibility).
     On Windows the default user data directory is
-    `C:\Users\USERNAME\AppData\Roaming\pandoc`.
+    `%APPDATA%\pandoc.
```
jgm, line 379: Let's add the closing backtick after pandoc, and I think this would now fit on the previous line.

Issue #8688, opened 2023-03-13
[org reader, regression] standard attributes not emitted as metadata


VladimirAlexiev:
**Explain the problem.**
According to https://pandoc.org/org.html#export-options, these org options:
```
# cat test.org
#+DATE: 2023-03-14
#+TITLE: Title
#+SUBTITLE: Subtitle
#+AUTHOR: Me, Myself
#+LANGUAGE: en

Start of text.
```
Should be emitted as markdown metadata. 
But they are not, even if I ask for `+yaml_metadata_block` (which is default with `markdown`):
```
# pandoc -t markdown+yaml_metadata_block test.org
Start of text.
```
Guess I can work-around with https://pandoc.org/org#directives-as-metadata, but these are standard options that should be emitted.

**Pandoc version?**
```
# pandoc -v
pandoc 3.1.1
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: C:\Users\Vladimir Alexiev\AppData\Roaming\pandoc

# uname
CYGWIN_NT-10.0
```

jgm:
Try adding `-s`.  The metadata is emitted for markdown only if "standalone" is selected.

VladimirAlexiev:
@jgm thanks! Made a PR for the documentation.

PR #8690, opened 2023-03-13
clarify org.md re #8688

VladimirAlexiev:


jgm:
Thanks - I went with somewhat different wording.

Issue #8691, opened 2023-03-14
[markdown writer] excessive/unnecessary backslash escaping of parens and quotes


VladimirAlexiev:
**Explain the problem.**
The markdown writer uses backslash escapes on all quotes `"` and some parens `(..)`.
- Why is this needed? AFAIK, these symbols are not used for markdown formatting
- Parens are escaped inconsistently inside tables (NO) vs outside tables (YES)
- Parens are escaped inconsistently depending on what's between the parens
- Asking for `-all_symbols_escapable` (https://pandoc.org/MANUAL.html#extension-all_symbols_escapable) doesn't change this behavior

The org source doesn't use any escaping of quotes or parens. 
The markdown result has the escaping problems described above.
```
# cat test3.org
| field | comment                            |
|-------+------------------------------------|
| foo   | See important "remark" below (1)   |
| bar   | See important "note" below (2)     |
| baz   | See important "notation" below (3) |

Important remarks
- (1) This "remark" is remarkable
- (2,3) These "notes" are notable

# pandoc -t markdown-simple_tables-all_symbols_escapable test3.org
| field | comment                              |
|-------|--------------------------------------|
| foo   | See important \"remark\" below (1)   |
| bar   | See important \"note\" below (2)     |
| baz   | See important \"notation\" below (3) |

Important remarks

-   \(1\) This \"remark\" is remarkable
-   (2,3) These \"notes\" are notable
```

**Pandoc version?**
```
# pandoc -v
pandoc 3.1.1
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: C:\Users\Vladimir Alexiev\AppData\Roaming\pandoc

# uname
CYGWIN_NT-10.0
```

tarleb:
Pandoc's Markdown comes with many extensions. The one relevant here are `smart` and `fancy_lists`.

The `smart` extension can also be enabled for org; doing so will prevent the backlash-escaping of quote chars: 

```
% pandoc -f org+smart -t markdown-simple_tables test.org
| field | comment                            |
|-------|------------------------------------|
| foo   | See important "remark" below (1)   |
| bar   | See important "note" below (2)     |
| baz   | See important "notation" below (3) |

Important remarks

-   \(1\) This "remark" is remarkable
-   (2,3) These "notes" are notable
```

The remaining backslashes are necessary, as one can test by removing them manually and comparing the conversion results.

VladimirAlexiev:
Thanks @tarleb ! 
[...]

PR #8694, opened 2023-03-14
Clarification re #8691

VladimirAlexiev:


tarleb:
This is helpful info, but I'm not sure if it belongs in the manual. The manual is already quite long, and the sections for the other extensions are kept brief. Maybe this can be shortened to a note similar to the one below the `smart` extension?

PR #8700, opened 2023-03-15
pandoc-cli: add a Lua REPL

tarleb:


tarleb:
Open questions and remaining work:

- It seems that isocline will only enable the command history if there's a history file. The current version takes that value from `PANDOC_REPL_HISTORY`, but maybe it should automatically create a history file in the user directory?
- I didn't get tab completion to work yet.
- Is it enough to have the REPL as part of `pandoc lua`, or should it be possible to start a REPL at arbitrary points in the code?

alerque:
Starting a REPL at arbitrary points in the document processing process would be far more useful to me at least that just a blank entry point.

jgm:
One option could be to create a history file in TMPDIR (if there is none already).
This would give you local history that wouldn't persist.  I feel like this doesn't belong in the user data dir.

tarleb:
About starting a REPL in arbitrary places: what would be a nice way to define the return value of a REPL? E.g.

1. Accept an extra string that's read after the REPL terminates, e.g. `return repl('foo')` to return the value of the global `foo`.
2. Always return the result(s) of the last evaluated expression.
3. Introduce a special `exit` function that exits the REPL and return the given values (`> exit(pandoc.Str 'test')`).
4. ...

I'm leaning towards option 2 because it's simple and not too difficult to implement.

jgm:
2 seems fine to me!

alerque:
2 would have been my first pick too unless `return val, ...` was somehow wired up, but that isn't how other Lua REPLs seem to work. Just remember that the "last evaluated expression" might return more than one value.

tarleb:
OK, this is ready I think. I'll handle the `pandoc.cli.repl`(?) function in another PR.

jgm:
Are any documentation changes needed?  Does `pandoc lua -i` start this repl?  If so, then `doc/pandoc-lua.md` needs updating, as it says that this is not supported.

tarleb:
Docs have been updated now.

jgm:
Excellent, this will be a nice feature!

PR #8701, opened 2023-03-15
Use ghc-musl 9.6.2 docker image

benz0li:
Some issues need to be resolved first.


benz0li:
Next: https://github.com/yesodweb/wai/issues/925

benz0li:
Next: https://github.com/haskell-servant/servant/issues/1659

benz0li:
> Build is failing at `Checking that the binary is statically linked...`, because it is not:
> 
> ```bash
> file "linux-arm64/pandoc"
> ```
> 
> ```
> linux-arm64/pandoc: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-aarch64.so.1, stripped
> ```

– https://github.com/haskell-servant/servant/issues/1659#issuecomment-1490555294

@jgm Did I miss a specific flag? I took the ones according to the [Makefile](/jgm/pandoc/blob/main/Makefile) and added `--allow-newer` to `CABALOPTS`.

jgm:
Great!  Let's wait for it to work with released packages before merging.

benz0li:
Some more output from the `docker run` command that might be of interest:

```
src/Text/Pandoc/Writers/EPUB.hs:254:5: warning: [GHC-53633] [-Woverlapping-patterns]
    Pattern match is redundant
    In an equation for ‘addMetadataFromXML’:
        addMetadataFromXML e@(Element (QName name _ (Just "dc")) attrs _ _)
                           md | name == "type" = ...
    |
254 |   | name == "type" = md { epubType = Just $ strContent e }
    |     ^^^^^^^^^^^^^^
```

Cross references:

* https://errors.haskell.org
* https://errors.haskell.org/messages/GHC-53633/

jgm:
The compiler has probably gotten smarter about detecting redundancies...
But if we remove this we may get a warning about incomplete pattern match from earlier compilers...would have to check.

jgm (commented):
You probably want to change l. 55 of .cirrus.yml as well.

benz0li:
Waiting for https://github.com/haskell-servant/servant/pull/1680

benz0li:
Waiting for https://github.com/haskell-servant/servant/pull/1679

benz0li:
@jgm Pandoc builds with GHC v9.6.2, too.

ℹ️ Tested for both `linux/amd64` and `linux/arm64/v8`.

jgm:
excellent!

PR #8703, opened 2023-03-18
Lua: add pandoc.cli.repl function

tarleb:


jgm:
It works!  Very useful.

tarleb:
It was reported to me that this does not work on Windows, but I'm clueless as to why.

jgm:
It would be worth adding that fact to the documentation until we figure out why.
(Is it that the repl doesn't work when you do `pandoc repl -i`, or just the function that causes the repl to appear?)

PR #8706, opened 2023-03-19
Textile reader: Add support for link references

alts:
Textile supports what it calls a "link alias", which is analogous to Markdown's reference-style links.

During the parser's first-pass, we can collect link references in addition to footnotes. Textile differs from Markdown in that reference definitions don't have titles nor attributes, so the stored values in `stateKeys` is relatively sparse.

I modeled the parser after the Markdown, and the new test cases are passing, but let me know if there's anything essential from Markdown's reference parsing that I missed.


jgm:
thanks and sorry for the belated merge!

PR #8710, opened 2023-03-20
T.P.Format: add new function `formatFromFilePaths` [API Change]

tarleb:
The module `Text.Pandoc.App.FormatHeuristics` is removed.


tarleb:
This is an alternative to #8693. Cc @mhwombat, who wrote that PR.

doc/lua-filters.md
```diff
@@ -3703,6 +3703,9 @@ are explicitly passed via the `env` parameter; e.g.
       return pandoc.cli.repl{ doc = doc }
     end
 
+**Note**: it seems that the function exists immediately on
```
jgm, line 3706: exists -> exits

jgm:
This looks like a good improvement to me!  I noted one typo.

tarleb:
Thanks!
Large parts of `lua-filters.md` are now generated from the docs in the Haskell sources: the idea is that we can build an online help system without our other docs getting out of sync. That's why the *pandoc.cli.repl* note was in this PR: I had added the note to the Haskell sources, but forgotten to run `make -B doc/lua-filters.md` afterwards. The command updates that file with the docs from the sources. The *repl* note was added here when I ran it after adding the new from_path function. 

There are still some API docs of lua-filters.md that are not generated from the sources, as the documentation system in HsLua is not powerful enough to capture the details adequately. I'm working on that.

Issue #8716, opened 2023-03-22
Org-mode reader: zero width space not recognised as an escape character


cchristiansen:
Org-mode uses the zero width space `U+200B` as an escape character, [documentation here](https://orgmode.org/manual/Escape-Character.html). This means that when we convert org-mode to pandoc, the following org-mode input
``` org
/emph/​asis.
/emph/​asis in​/correct/.
/try/​/​/except/​/​/else/​/​/finally/
/try/​/​/except/​/else/​/finally/
```
(again but with zero width spaces replaced with `X` for clarity),
``` org
/emph/Xasis.
/emph/Xasis inX/correct/.
/try/X/X/except/X/X/else/X/X/finally/
/try/X/X/except/X/else/X/finally/
```
should be rendered as
``` pandoc
para (emph "emph" <> "Xasis")
para (emph "emph" <> "Xasis" <> space <> "inX" <> emph "correct")
para (emph "try" <> "X/X" <> emph "except" <> "X/X" <> emph "else" <> "X/X" <> emph "finally")
para (emph "try" <> "X/X" <> emph "except" <> "X" <> emph "else" <> "X" <> emph "finally")
```
however it is currently rendered as
``` pandoc
para "/emph/Xasis"
para (emph "/emph/Xasis inXcorrect")
para (emph "/try/X/X/except/X/X/else/X/X/finally/")
para (emph "/try/X/X/except/X/else/X/finally/")
```

This problem appears to be related to #6070 and #7135. #8059 did not solve this issue.

**Pandoc version**
pandoc 3.0.1

jgm:
Cloesd by fe720dd

PR #8717, opened 2023-03-22
Org reader: Allow zero width space as an escape character

cchristiansen:
Allow the character U+200B to be used as an escape character as described in the Org-mode documentation https://orgmode.org/manual/Escape-Character.html

Closes issue #8716.


tarleb:
Thank you!

I guess we should find a way to add *all* Unicode space characters to that list, but zero width space is definitely the most important.

PR #8723, opened 2023-03-26
Typst: fix output for syntax-highlighted inline code.

tarleb:


src/Text/Pandoc/Writers/Typst.hs
```diff
@@ -208,8 +208,8 @@ inlineToTypst inline =
               DisplayMath -> return $ "$ " <> literal r <> " $"
     Code (_,cls,_) code -> return $
       case cls of
-        (lang:_) -> "#raw(lang=" <> doubleQuotes (literal lang) <>
-                        ", " <> doubleQuotes (literal code) <> ")"
+        (lang:_) -> "#raw(lang:" <> doubleQuotes (literal lang) <>
+                    ", " <> doubleQuotes (literal $ T.replace "\"" "\\\"" code) <> ")"
```
jgm, line 212: Oh right!  This makes me think we need to worry about other escapes too.
Let me check on the site.

jgm: Yeah, we also need to backslash escape any backslashes in the code.

jgm: I think it would be good to add a general function that handles this for everything we double quote -- I will do this, so don't worry about updating here.

tarleb:
One other issue I noticed is with dots: the dot in `#emph[2.0]` is rendered as a sentence-ending period unless it is escaped.

jgm:
The sentence-ending period issue is more difficult. I guess we can probably safely escape all word-internal periods?.

jgm:
Hopefully that's a bug in typst that will be fixed...

jgm:
Update: Looks like it's not a sentence-ending space. Rather, when at beginning of line, this is parsed as an ordered list item, despite the lack of space after the `.`.

jgm:
I think this is obsoleted by the commit I just pushed,
cd6b3bde622c93eea763e981692b4bc53dd5e5df

PR #8724, opened 2023-03-27
Add rowspan, colspan and alignment to cells in jats table reader

noahmalmed:
### Description 
I'm looking for some guidance on completing this code change. This is my first time writing Haskell code and I'm getting stuck on some code and I can't tell if my design is wrong/bad or if I'm just missing something syntactically or conceptually.

My issue is arising from trying to parse the xml `Element` attributes at the same time as when we parse the content of the `Element` itself, so when we create a `Cell`, we have all of the information we need.

The previous design of the code just parsed the xml row `Element`s into `Blocks` and then made `simpleCell`s (which don't accept rowspan, colspan or alignment attributes) from those `Blocks`.

So, I'm getting stuck with a typing/monad issue. I'm able to parse out the `colspan`, `rowspan` and `alignment` attributes from the xml and push those values into cells, but I'm having a heard time using `parseMixed` to parse an `Element` into `Blocks`. `parseMixed`  returns the type `StateT JATSState Blocks`. What would be an idiomatic way to get the `Blocks` value out of the `StateT JATSState` monad that is returned by this function?

I also realize that because I'm new to Haskell and Pandoc that my approach to this problem may be way off base, so any direction there would be welcome as well!

### DIFF
--- a/src/Text/Pandoc/Readers/JATS.hs
+++ b/src/Text/Pandoc/Readers/JATS.hs
@@ -273,26 +273,27 @@ parseBlock (Elem e) =
                                            Just c -> filterChildren isColspec c
                                            _      -> filterChildren isColspec e'
                       let isRow x = named "row" x || named "tr" x
-                      headrows <- case filterChild (named "thead") e' of
-                                       Just h  -> case filterChild isRow h of
-                                                       Just x  -> parseRow x
-                                                       Nothing -> return []
-                                       Nothing -> return []
-                      bodyrows <- case filterChild (named "tbody") e' of
-                                       Just b  -> mapM parseRow
-                                                  $ filterChildren isRow b
-                                       Nothing -> mapM parseRow
-                                                  $ filterChildren isRow e'
+
+                      -- list of header cell elements
+                      let headRowElements = case filterChild (named "thead") e' of
+                                      Just h -> maybe [] parseElement (filterChild isRow h)
+                                      Nothing -> []
+                      -- list of list of body cell elements 
+                      let bodyRowElements = case filterChild (named "tbody") e' of
+                                      Just b -> map parseElement $ filterChildren isRow b
+                                      Nothing -> map parseElement $ filterChildren isRow e'
                       let toAlignment c = case findAttr (unqual "align") c of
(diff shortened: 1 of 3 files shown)


src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -303,15 +304,22 @@ parseBlock (Elem e) =
                                                 Just ws' -> let tot = sum ws'
                                                             in  ColWidth . (/ tot) <$> ws'
                                                 Nothing  -> replicate numrows ColWidthDefault
-                      let toRow = Row nullAttr . map simpleCell
+
+                      -- How do I parse an Element into Blocks?
+                      -- parseCell takes an element and returns a StateT JATSState Blocks, but I need it to return the unwrapped value
+                      let parseCell = parseMixed plain . elContent
+
+                      let elementToCell element = cell (toAlignment element) (RowSpan $ toRowSpan element) (ColSpan $ toColSpan element) mempty
```
noahmalmed, line 312: Here is where I am having trouble. Right now, I'm creating an empty `Cell`, but I really want to be able to use `parseCell` to parse the content of the xml cell into the Pandoc cell

tarleb: All parsing happens ins the `StateT JATSState m` type, a.k.a. `JATS m`. It is an instance of `Monad`, so we can use `cellContent <- parseCell element` while we are in a `do` block, as is the case here.
I could probably free an hour tomorrow if you want to pair on this.

noahmalmed: Thanks for the quick response @tarleb! Running `cellContent <- parseCell element` makes sense, but I think what's tripping me up is that I would want to run it in the context of `elementToCell`, which is not in a do block. 

I'd love to do a quick pair! I'm normally available between 8-5 CST (GMT -6), but also wouldn't mind stretching on either end. Let me know when works for you!

tarleb: Would [14:30 UTC](https://time.is/1430_28_Mar_2023_in_UTC) work for you? If so, please send me an invite via mail.

Issue #8408, opened 2022-11-01
JATS reader - add support for new table features


Opussci:
Hi @tarleb The table with colspan and rowspan is not transformed correctly from JATS to HTML or Docx. Maybe the JATS reader needs to be adjusted to support these table features.

jgm:
Please give pandoc version, exact command line used, and a small input sample sufficient to reproduce the issue.
If you're not using the latest version, please try with that or on https://pandoc.org/try.

coryschires:
Hi! I'd like to revive this issue, as I've been able to recreate the problem using the latest version of Pandoc.

## Problem
Table styles – specifically `colspan` and `rowspan` – are not retained when using the JATS reader.

## Goal
I'd like to retain at least `colspan` and `rowspan` as I know there is increasingly support for this feature in various reader and writers. 

Depending on the difficulty, we may be willing to make a pull request to add this functionality. So, if y'all are not interested in coding this feature in the near future, I'd appreciate a little nudge in right direction, so we can take a crack at it ourselves. 

Thanks!

---

## Pandoc Version

```bash
pandoc --version
pandoc 3.1
Features: +server +lua
Scripting engine: Lua 5.4
```

## CLI command to reproduce

```
pandoc sample.xml -f jats -t html -o sample.html
```

### sample.xml

```xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Publishing DTD v1.1 20151215//EN" "https://jats.nlm.nih.gov/publishing/1.1/JATS-journalpublishing1-mathml3.dtd">
<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" article-type="research-article" dtd-version="1.1" xml:lang="en">
  <front>
    <journal-meta>
      <journal-title-group>
        <journal-title>Survey Practice</journal-title>
      </journal-title-group>
    </journal-meta>
    <article-meta>
      <title-group>
        <article-title>How Weighting by Past Vote Can Improve Estimates of Voting Intentions</article-title>
      </title-group>
    </article-meta>
  </front>
  <body>
    <table-wrap id="attachment-147394">
      <table>
        <thead>
          <tr>
            <td>
              <bold>Benchmark</bold>
            </td>
            <td>
              <bold>Population %</bold>
[...]

PR #8726, opened 2023-03-28
Add rowspan, colspan and alignment to cells in jats table reader

noahmalmed:
### Background 
See [this issue](https://github.com/jgm/pandoc/issues/8408) for background

### Description 
Parse the colspan, rowspan and alignment attributes of Jats table into the Cell node

### DIFF
--- a/src/Text/Pandoc/Readers/JATS.hs
+++ b/src/Text/Pandoc/Readers/JATS.hs
@@ -287,8 +287,10 @@ parseBlock (Elem e) =
                                                 Just "right"  -> AlignRight
                                                 Just "center" -> AlignCenter
                                                 _             -> AlignDefault
-                      let toColSpan element =  fromMaybe 1 $ safeRead $ fromMaybe "1" $ findAttr (unqual "colspan") element
-                      let toRowSpan element =  fromMaybe 1 $ safeRead $ fromMaybe "1" $ findAttr (unqual "rowspan") element
+                      let toColSpan element = fromMaybe 1 $ 
+                            findAttr (unqual "colspan") element >>= safeRead
+                      let toRowSpan element =  fromMaybe 1 $ 
+                            findAttr (unqual "rowspan") element >>= safeRead
                       let toWidth c = do
                             w <- findAttr (unqual "colwidth") c
                             n <- safeRead $ "0" <> T.filter (\x -> isDigit x || x == '.') w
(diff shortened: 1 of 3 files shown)


src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -273,26 +273,27 @@ parseBlock (Elem e) =
...
+                      let bodyRowElements = case filterChild (named "tbody") e' of
+                                      Just b -> map parseElement $ filterChildren isRow b
+                                      Nothing -> map parseElement $ filterChildren isRow e'
                       let toAlignment c = case findAttr (unqual "align") c of
                                                 Just "left"   -> AlignLeft
                                                 Just "right"  -> AlignRight
                                                 Just "center" -> AlignCenter
                                                 _             -> AlignDefault
+                      let toColSpan element =  fromMaybe 1 $ safeRead $ fromMaybe "1" $ findAttr (unqual "colspan") element
+                      let toRowSpan element =  fromMaybe 1 $ safeRead $ fromMaybe "1" $ findAttr (unqual "rowspan") element
```
jgm, line 291: We try to keep lines to 80 columns or less.
In this case, you could probably avoid having two `fromMaybe`s:
```
fromMaybe 1 $ findAttr (unqual "colspan") element >>= safeRead
```
(untested). similarly on next line.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -303,15 +304,22 @@ parseBlock (Elem e) =
                                                 Just ws' -> let tot = sum ws'
                                                             in  ColWidth . (/ tot) <$> ws'
                                                 Nothing  -> replicate numrows ColWidthDefault
-                      let toRow = Row nullAttr . map simpleCell
-                          toHeaderRow l = [toRow l | not (null l)]
+
+                      let parseCell = parseMixed plain . elContent
+                      let elementToCell element = cell (toAlignment element) (RowSpan $ toRowSpan element) (ColSpan $ toColSpan element) <$> (parseCell element)
```
jgm, line 309: line length

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -303,15 +304,22 @@ parseBlock (Elem e) =
...
+                      let rowElementsToCells elements = mapM elementToCell elements
+                      let toRow = fmap (Row nullAttr) . rowElementsToCells
+                          toHeaderRow element = sequence $ [toRow element | not (null element)]
+
+                      parsedRow <- toHeaderRow headRowElements
+                      parsedRows <- mapM toRow bodyRowElements
```
jgm, line 315: I think this would be clearer to read if it were `headerRow` and `bodyRows`.

jgm (commented):
Looks good, with some minor comments!

noahmalmed:
@jgm Thanks for the review! Made those changes for you 👍

jgm:
Great!  I'm wondering if this completely addresses #8408 or only partially.
(It can be merged now either way, I just want to know whether to indicate that it closes #8408 .)

Does JATS have a concept of a table foot, or multiple table bodies, or row heads?  I don't think these are handled yet.

noahmalmed:
Good point @jgm. 

Taking a look at the [jats docs](https://jats.nlm.nih.gov/archiving/tag-library/1.1d1/n-by90.html), it looks like a table can have more than one body and zero or more footer.
> [<tfoot>](https://jats.nlm.nih.gov/archiving/tag-library/1.1d1/n-8az0.html) [Table Footer (NISO JATS table model)](https://jats.nlm.nih.gov/archiving/tag-library/1.1d1/n-8az0.html), zero or one
[<tbody>](https://jats.nlm.nih.gov/archiving/tag-library/1.1d1/n-g4m0.html) [Table Body (NISO JATS table model)](https://jats.nlm.nih.gov/archiving/tag-library/1.1d1/n-g4m0.html), one or more

Which the pandoc code partially handles. [Here](https://github.com/noahmalmed/pandoc/blob/f19c2054636a949247c4ce4ae634e30cd9bf5ce4/src/Text/Pandoc/Readers/JATS.hs#L325) we can see that the code assumes there is only one body in the table and [here](https://github.com/noahmalmed/pandoc/blob/f19c2054636a949247c4ce4ae634e30cd9bf5ce4/src/Text/Pandoc/Readers/JATS.hs#L326) we can see that the code skips parsing the footer elements. 


It may make sense to just create a new issue to fix the above two issues and include col/row span and alignment parsing in that issue. I'd be happy to write that up if that makes sense to you.

jgm:
Yes, a new issue/PR would be great.
I'll merge this meanwhile!

Issue #8729, opened 2023-03-29
Associate text with checkboxes as `<label>`s for accessible HTML


batpigandme:
## Explain the problem.

For accessibility/WCAG standards, checkboxes need labels that are programmatically determinable for assistive technologies. Currently, the text for checkbox `<input>`s are in the same `<li>`, but aren't associated as labels.

### Input
```md
-  [ ] Item 1
-  [ ] Item 2
```

### Current output
```
pandoc::pandoc_convert(text = c("-  [ ] Item 1","-  [ ] Item 2"), to = "html")
#> <ul class="task-list">
#> <li><input type="checkbox" />Item 1</li>
#> <li><input type="checkbox" />Item 2</li>
#> </ul>
```

Using Try pandoc, or `pandoc input.md -o output.html`
```html
<ul class="task-list">
<li><input type="checkbox" />Item 1</li>
<li><input type="checkbox" />Item 2</li>
</ul>
```

### Desired/expected output
The easiest way to make these accessible is [Using label elements to associate text labels with form controls (WCAG H44)](https://www.w3.org/WAI/WCAG21/Techniques/html/H44.html). The text should become `<label>`s, and the `<input>`s need `id`s so that the labels can be associated with the `for` attribute.

For this example, the output might be:
```html
<ul class="task-list">
    <li><input id="cb1" type="checkbox" /><label for="cb1">Item 1</label></li>
    <li><input id="cb2" type="checkbox" /><label for="cb2">Item 2</label></li>
</ul>
``` 

**Pandoc version?**
MacOS 13.2.1 Pandoc 3.1.2 and/or https://pandoc.org/try/

tarleb:
[MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) states that the `<label>` element can also be used to wrap both the `<input>` element and the description. This doesn't require us to generate ID and is much easier to implement, so I went with that. Leaving the issue open for now in case that accessibility software doesn't work well with that.

tarleb:
This requires more though, I'm reverting the commit: the `<label>` element cannot contain lists, so nested task lists like the below lead to invalid HTML:

``` markdown
 - [x] foo
  - [ ] bar
  - [x] baz
- [ ] bim
```

tarleb:
Pandoc requires task lists to start with a paragraph or plain text. One possible solution would be to only add that text to the label.

PR #8730, opened 2023-03-29
HTML writer: use first paragraph in task item as checkbox label.

tarleb:
Closes: #8729


tarleb:
The change doesn't affect the layout AFAIK. A side effect of this change is that clicking the task description will toggle the checkbox, which is another accessibility improvement. Thus merging.

Issue #8731, opened 2023-03-29
Task lists styling doesn't work in reveal.js


tarleb:
Tasks lists like the below are rendered in a way that requires extra styling in HTML. The styles that pandoc adds are overridden by the reveal.js style, leading to an unpleasant look of task lists when placed on reveal.js slides.

``` markdown
$ pandoc -o tasklist.html -s -t revealjs
- [ ] this
- [ ] that
^D
```

See quarto-dev/quarto-cli#2834 for more details.

jgm:
Would it work to put reveal styling before pandoc styling in the template?

tarleb:
I forgot most of what I knew about CSS rule precedence, but I seem to remember that rules with a class always take precedent of those without a class, so it won't work, probably.

PR #8733, opened 2023-03-29
styles.html: fix task-list styling in reveal.js

tarleb:
Fixes: #8731


tarleb:
This increases selector specificity by adding the `[class]` selector, which is otherwise superfluous. This way our rule takes precedence over the `.reveal ul` rule.

An alternative would be to nest a reveal-specific rule below a conditional.

PR #8734, opened 2023-03-29
Typo

perrotuerto:
"spacesto" -> "spaces to"
"theline" -> "the line"

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -5663,7 +5663,7 @@ Returns
 
 `render (doc[, colwidth])`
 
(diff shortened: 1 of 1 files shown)


jgm:
@tarleb - just wondering if these are in the automatically generated parts of the document?

perrotuerto:
It could be, I just noted that it appears since 49f1e7608ead7a9bc0e73e99b24e47a9a8cf0400 (where `Module pandoc.layout` started to appear in the docs).

tarleb:
The docs were generated, and the error *did* exist in the source files, but this part of the lua-filters.md is **not** automatically updated yet.

Thank you for the patch!

PR #8772, opened 2023-04-10
MANUAL: Note that fancy_lists don't work with commonmark

wlupton:
This PR appears also to have deleted some trailing spaces in other parts of the file. Hope that's OK.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -4168,7 +4168,7 @@ ordered list marker in place of a numeral:
     #. one
     #. two
 
-Note:  the `fancy_lists` extension doesn't work with `commonmark`.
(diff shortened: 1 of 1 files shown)


MANUAL.txt
```diff
@@ -4168,6 +4168,8 @@ ordered list marker in place of a numeral:
     #. one
     #. two
 
+Note:  the `fancy_lists` extension doesn't work with `commonmark`.
```
tarleb, line 4171: I think this is a bit too general, as only the `#.` syntax doesn't work. Other features, like using `(a)`, `(b)` as item markers, *do* work.

wlupton: Thanks. I updated the note.

Issue #8590, opened 2023-01-30
commonmark_x fenced divs can't have trailing colons


wlupton:
With commonmark_x, fenced divs appear not to support trailing colons (as in "::: note :::"). [Here's a pandoc try example.](https://pandoc.org/try/?params=%7B%22text%22%3A%22%23+with+commonmark_x%2C+this+doesn%27t+work%5Cn%5Cn%3A%3A%3A+note+%3A%3A%3A%5Cn%5CnThis+is+a+note.%5Cn%5Cn%3A%3A%3A%5Cn%5Cn%23+but+this+does%5Cn%5Cn%3A%3A%3A+note%5Cn%5CnThis+is+a+note.%5Cn%5Cn%3A%3A%3A%22%2C%22to%22%3A%22html5%22%2C%22from%22%3A%22commonmark_x%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A%7B%7D%2C%22template%22%3Anull%7D)

I think this is just a documentation issue, and that a note could be added to the manual (modeled on the note that already explains the YAML metadata block restrictions). I'd be happy to create a PR.

jgm:
Sure, go ahead.  I am tempted to modify the fenced_div extension for commonmark so it allows this. But in the mean time, documentation should reflect reality.

wlupton:
OK. I won't do it immediately!

I wondered whether it was deliberate because of potential ambiguity with definition lists. But I only had this thought because this:

```
::: note :::

This is a note.

:::
```

...gives this HTML.

```
<p>::: note :::</p>
<dl>
<dt>This is a note.</dt>
<dd>
<p>::</p>
</dd>
</dl>
```

jgm:
No, it wasn't for that reason.  I think either (a) I was trying to keep it simple and regular (a good goal) or (b) this just implemented before we allowed `::: note :::` in pandoc.

wlupton:
@jgm, you said:

> I am tempted to modify the fenced_div extension for commonmark so it allows this

That would be great! Currently this is the main markdown / commonmark_x incompatibility that I'm seeing.

tarleb:
[...]

PR #8773, opened 2023-04-10
MANUAL: add commonmark fenced_div note

wlupton:
Addresses #8590.


tarleb:
This seems like an important addition, so I'm merging the PR. I'll add a note to #8590 to revert the commit once the commonmark reader has been updated.

PR #8800, opened 2023-04-22
Fix apparent inaccuracy in MANUAL.txt

seifferth:
The section about pandoc's template syntax used to state that
spaces and tabs following the closing delimiter around variables
and control structures will be ignored. Some quick testing with
pandoc version 2.9.2.1 suggests that this is not the case. Spaces
and tabs _preceding_ the closing delimiter, however, are indeed
ignored. I therefore assume that this is simply a slight inaccuracy
in the manual; possibly produced by copying the sentence about the
opening delimiter and forgetting to adjust that word.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2179,7 +2179,7 @@ The styles may also be mixed in the same template, but the
 opening and closing delimiter must match in each case.  The
 opening delimiter may be followed by one or more spaces
 or tabs, which will be ignored. The closing delimiter may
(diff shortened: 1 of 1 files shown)


jgm:
Can you test with a recent version of pandoc? 2.9.2.1 is quite old, and the manual describes the current version (3.1.2).

jgm:
However, looking at this, I believe you are right (also for current version). Thanks.

Issue #8803, opened 2023-04-25
Highlighting (`mark` filter) does not work with beamer


mh4ckt3mh4ckt1c4s:
**Explain the problem.**

When using the `mark` extension, output is correctly highlighted in pdf output (using LaTeX), but not when using beamer. This problem is also present without using the extension, when using the `mark` class described [here](https://pandoc.org/MANUAL.html#highlighting).

Content of the `test.md` file : 

```md
# Please highlight my test

This following text is highlighted : ==wow so shiny==

[Mark]{.mark}
```

Command and output using LaTeX : 

```bash
pandoc -i test.md -o test.pdf -f markdown+mark -t latex
```

Output : 

![image](https://user-images.githubusercontent.com/76600167/234225938-d2970fee-d27a-4dc4-a118-e09f4f882f7e.png)

Command and output using beamer : 

```bash
pandoc -i test.md -o test.pdf -f markdown+mark -t beamer
```

![image](https://user-images.githubusercontent.com/76600167/234226303-425236ad-1d59-49ec-8759-ae0f51bbb27f.png)

A potential solution for this could be here : https://tex.stackexchange.com/questions/41683/why-is-it-that-coloring-in-soul-in-beamer-is-not-visible 

**Pandoc version?**

Pandoc 3.0.1 on Archlinux

jgm:
That looks like the right sort of solution. Do you want to propose a specific revision to the default template?

mh4ckt3mh4ckt1c4s:
Yep, I can do that.

Issue #8707, opened 2023-03-20
soul.sty for strikeout does't work with lualatex


mako09:
When using pandoc with `--pdf-engine=lualatex`, an error occurs if strikeout is included.

```
$ echo '~~打ち消し線~~、それから[下線]{.underline}、さらに==ハイライト==' | pandoc -f markdown+mark --pdf-engine=lualatex -V documentclass=jlreq -o sample.pdf
Error producing PDF.
! Package soul Error: Reconstruction failed.

See the soul package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.59 \st{打ち消し線}
```

This is because the package `soul` cannot be used with lualatex.

Would it be possible to modify the `default.latex` as follows to address this issue?

```
$if(strikeout)$
$-- also used for underline
\ifLuaTeX
  \usepackage{luacolor}
  \usepackage[soul]{lua-ul}
\else
  \usepackage{soul}
\fi
$endif$
```

tarleb:
Minor note: *soul* has worked fine for me in the past, even with lualatex.
I support the change though, as it seems to improve CJK support.

PR #8806, opened 2023-04-26
Fix beamer highlights

mh4ckt3mh4ckt1c4s:
Fixing #8803 . The changes introduced in #8707 makes the highlighting works with lualatex, but changes in the template are still needed for pdflatex and xelatex. I tested these changes with pdflatex and lualatex both with standard and beamer output and it works.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -297,6 +297,15 @@ $-- also used for underline
   \usepackage[soul]{lua-ul}
 \else
   \usepackage{soul}
+$if(beamer)$
+  \makeatletter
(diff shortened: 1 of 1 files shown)


mh4ckt3mh4ckt1c4s:
Fixed merge conflict.

jgm:
Note the test failure -- probably you need to change something to do with indentation in a template.

mh4ckt3mh4ckt1c4s:
Tests are finally fixed. Took me more time than expected.

jgm:
Thanks!

Issue #8256, opened 2022-08-29
Multiline titles do not display in OpenDocument/ODT files


andrewheiss:
Using pandoc 2.19.2 on macOS 12.5.1, a multiline title will not display correctly in an .odt file because of nested paragraphs.

---

Ordinarily it is possible to add a linebreak to a title using YAML multiline block scalars, like so:

**testing.md:**

``` text
---
title: |
  This *is* \
  a test
author: Someone
---

Testing.
```

When using a multiline entry like this, metadata entry for `title` is treated as a block instead of an inline. This means that when rendering to HTML, the title content is wrapped in `<p>` tags:

``` bash
pandoc testing.md -o testing.html --standalone
```

``` html
<h1 class="title"><p>This <em>is</em><br />
a test</p></h1>
```

When using a single line, like

``` text
---
title: This *is* a test
author: Someone
---

Testing.
```

...the content isn't wrapped in `<p>`s:

``` html
<h1 class="title">This <em>is</em> a test</h1>
```

With HTML output, it is valid and legal to have a `<p>` inside a `<h1>`, so there are no problems.

However, with ODT output, the document title will not show up when using linebreaks in a YAML key.

For instance, running

**testing.md**

``` text
---
title: |
  This *is* \
  a test
author: Someone
---

Testing.
```

``` bash
pandoc testing.md -o testing.odt
```

creates a document that looks like this:

<img width="942" alt="image" src="https://user-images.githubusercontent.com/73663/187250827-a0a50ace-9017-4149-941e-a6856d8dd81a.png">

There's an empty paragraph where the title should be.

Inspecting the `content.xml` file inside the zipped `.odt` file shows that the title really is there, but it's wrapped in a `<text:p>` tag, as with HTML

``` xml
<text:p text:style-name="Title">
  <text:p text:style-name="Text_20_body">This <text:span text:style-name="T1">is</text:span><text:line-break/>a test</text:p>
</text:p>
```

Placing a `<text:p>` tag inside another `<text:p>` tag is invalid OOXML syntax, and both LibreOffice and Word are unable to display it.

This is similar to these issues:

-   Multiline titles don't display when rendering to EPUB: <https://github.com/jgm/pandoc/issues/8091> and <https://github.com/jgm/pandoc/issues/8095>. One solution there is to use line-block style values, like:

    ``` yaml
    title: |
        | This *is*
        | a test
    ```

    but doing this with ODT still creates the nested `<text:p>` issue.

[...]

PR #8807, opened 2023-04-26
    Writers.OpenDocument: fix invalid text:p inside text:p from meta

mistmist:
The only element in meta for which data/templates/default.opendocument
    allows block element content appears to be "abstract", which is already
    filtered out of meta' at this point, so simply convert all blocks to
    linebreak-separated inlines.
    
    Fixes: #8256

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -255,7 +255,7 @@ writeOpenDocument opts (Pandoc meta blocks) = do
                collectBlockIdent _                             = []
            modify $ \s -> s{ stIdentTypes = query collectBlockIdent blocks }
            m <- metaToContext opts
-                  (blocksToOpenDocument opts)
+                  (inlinesToOpenDocument opts . blocksToInlines)
                   (fmap chomp . inlinesToOpenDocument opts)
                   meta'
            b <- blocksToOpenDocument opts blocks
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Writers/OpenDocument.hs
```diff
@@ -255,7 +255,7 @@ writeOpenDocument opts (Pandoc meta blocks) = do
                collectBlockIdent _                             = []
            modify $ \s -> s{ stIdentTypes = query collectBlockIdent blocks }
            m <- metaToContext opts
-                  (blocksToOpenDocument opts)
+                  (inlinesToOpenDocument opts . B.toList . blocksToInlinesWithSep (B.str "<text:line-break/>"))
```
jgm, line 258: This doesn't seem quite right -- have you tested with multiparagraph content?
I think what you want is not `B.str "<text:linebreak/>"` but `B.rawInline (Format "opendocument") "<text:linebreak/>"`.

mistmist: oh, i didn't notice that this exists.
another thing i didn't notice is B.lineBreak which is easier.
another thing i didn't notice is "blocksToInlines" uses B.lineBreak by default.

hmm, for the input format used here multiple paragraphs in title result in a SoftBreak and not Para elements, so i'll just use blocksToLines.

btw i had to rebase this on the other (new) PR due to the version number change, which affects the test result - the top commit is the only one that "belongs" to this PR.

jgm:
I've merged the other commits, so you should be able to rebase now.

PR #8810, opened 2023-04-27
Add lang attribute to html tag in epub tpl

lewer:
Hi,

in default.epub2 and default.epub3 templates, the `lang` attribute is missing in the `<html>` tag (it is there in the default.html4 and default.html5 templates)

This has the effect of raising USAGE(HTM_021) [Content file doesn't contain lang attribute.] when checking the epub with epubcheck. When setting the `lang` attribute, the usage message disappears.

Following https://kb.daisy.org/publishing/docs/html/lang.html
>Is it necessary to specify both the lang and xml:lang attributes?
>
>For optimal accessibility both should be specified when creating XHTML content. Assistive technologies may only recognize one or the other attribute.


jgm:
thanks!

PR #8812, opened 2023-04-28
fix some issues in Writers.ODT

mistmist:
- switch to ODF 1.3 because of attribute that doesnt exist in 1.3
- fix manifest:version on entries
- remove LibreOffice-generated cruft


jgm:
Thank you!

PR #8816, opened 2023-05-01
8764 Writers.OpenDocument: don't lose row header column

mistmist:
not sure if the 2nd commit is a good idea, guess it depends on how "common" various different shapes of HTML (or other) tables are, please read the commit message...


jgm:
Great!  I think the second commit is a good idea, personally.
Thank you.

PR #8822, opened 2023-05-04
PandocMonad: allow nested data files

### LINKED ISSUE #6635: [Enhancement] Search for filters/lua-filters within sub-directories of $DATADIR/filters
Problem:
Currently, if a lua-filter (say `wordcount.lua`) is at `$DATADIR/filters/wordcount.lua`, things work fine. But if it is in `$DATADIR/filters/wordcount/wordcount.lua`, then the `pandoc` command is unable to locate it.

Commands:
```
$ cd $HOME/.local/share
$ git clone https://github.com/pandoc/lua-filters.git pandoc/filters
$ pandoc --lua-filter=wordcount.lua ~/example.md
```

Expected:
```
311 words in body
1741 characters in body
2014 characters in body (including spaces)
```

Actual Output:
```
Error running filter wordcount.lua:
wordcount.lua: openBinaryFile: does not exist (No such file or directory)
```

tarleb:
Data file fallbacks are now also searched in the directory that has the same name as the file minus the extension. For example, a filter with name `frob.lua` will be searched in `$DATADIR/filters/frob.lua` as well as in `$DATADIR/filters/frob/frob.lua`.

This allows to place full git repositories in the data directory.

Closes: #6635

### DIFF
--- a/src/Text/Pandoc/Class/PandocMonad.hs
+++ b/src/Text/Pandoc/Class/PandocMonad.hs
@@ -417,16 +417,19 @@ findFileWithDataFallback subdir fp = do
   existsInWorkingDir <- fileExists fp
   if existsInWorkingDir
      then return $ Just fp
-     else do
-       mbDataDir <- checkUserDataDir fp
-       case mbDataDir of
-         Nothing -> return Nothing
-         Just datadir -> do
-           let datafp = datadir </> subdir </> fp
-           existsInDataDir <- fileExists datafp
-           return $ if existsInDataDir
-                    then Just datafp
-                    else Nothing
+     else checkUserDataDir fp >>= findInDataDir
+ where
+  findInDataDir Nothing = pure Nothing
+  findInDataDir (Just datadir) = do
+    let datafp = datadir </> subdir </> fp
+    let nested = datadir </> subdir </> dropExtension fp </> fp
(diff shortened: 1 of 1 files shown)


jgm:
I'm on the fence about this.  I see how it meets a need in the case of filters, but (a) it's slightly "magic" in the way it derives the directory name and (b) it applies to all data files, not just filters.

A narrower intervention would do this just for filters.

I'm just not sure what's best, hence the delay in merging this!

0xdevalias:
> A narrower intervention would do this just for filters.
> 
> I'm just not sure what's best, hence the delay in merging this!

I wonder if we could downscope it to just the filters for now, and then if a need arises for other things in future, then it could be expanded to those when an appropriate use case is presented?

tarleb:
I think my thoughts back when I opened this were that this would also enable (or rather, simulate) "packages" in the sense that templates, filter, etc. could all be placed in a directory together with a defaults file. It would be easier to call `-d package` than `-d package/package`.

Now I tend to think that it might be better to invest the time into a proper package system to bundle files into a package. I think that would mostly serve the same use-case as this PR.

rnwst:
See also the discussion in #10874. The consensus there seemed to be that `frob/init.lua` would be most idiomatic. I still plan to submit a PR for that at some stage.

PR #8840, opened 2023-05-09
Improve title and label parsing in the Jats reader

### LINKED ISSUE #8718: Output formats fails to retain JATS `<title>` and `<label>` when nested inside `<statement>`
## Problem
When converting from JATS to Markdown (or HTML and probably other outputs), both `<label>` and `<title>` tags are ignored when nested within `<statement>`.

Given the following (abridged) JATS XML:

```xml
<statement>
  <label>Hypothesis 1</label>
  <p>We hypothesize the following</p>
</statement>
```

The following command:

```bash
pandoc statement-with-label.xml -f jats -t markdown -o statement-with-label.md
```

Will produce to following Markdown:

```markdown
#

We hypothesize the following
```

## Solution / Expected Behavior
Instead, I would expect the label (or title) to be retained:

```markdown
# Hypothesis 1

We hypothesize the following
```

### Pandoc Version
```
pandoc 3.1
Features: +server +lua
Scripting engine: Lua 5.4
```

#### statement-with-label.xml
```xml
<!DOCTYPE article PUBLIC "-//NLM//DTD Journal Publishing DTD v3.0 20080202//EN" "journalpublishing3.dtd">
<article article-type="research-article" dtd-version="3.0" xml:lang="en" xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink">
  <body>
    <sec>
      <statement>
        <label>Hypothesis 1</label>
        <p>We hypothesize the following</p>
      </statement>
    </sec>
  </body>
</article>
```

#### statement-with-title.xml
```xml
<!DOCTYPE article PUBLIC "-//NLM//DTD Journal Publishing DTD v3.0 20080202//EN" "journalpublishing3.dtd">
<article article-type="research-article" dtd-version="3.0" xml:lang="en" xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink">
  <body>
    <sec>
      <statement>
        <title>Hypothesis 1</title>
        <p>We hypothesize the following</p>
      </statement>
    </sec>
  </body>
</article>
```

### Commands to reproduce
```bash
# Markdown as output
pandoc statement-with-label.xml -f jats -t markdown -o statement-with-label.md
pandoc statement-with-title.xml -f jats -t markdown -o statement-with-title.md

# HTML as output
pandoc statement-with-label.xml -f jats -t html -o statement-with-label.html
pandoc statement-with-title.xml -f jats -t html -o statement-with-title.html
```

---

Thanks for the help! If the fix is easy, we might be able to contribute a PR. Just point us in the right direction.

noahmalmed:
Fixes #8718

## Description
There were a number of elements where the jats reader did not parse their `title` and `label` into headers. This PR fixes that

### DIFF
--- a/src/Text/Pandoc/Readers/JATS.hs
+++ b/src/Text/Pandoc/Readers/JATS.hs
@@ -170,19 +170,7 @@ parseBlock (Elem e) = do
         "code" -> codeBlockWithLang
         "preformat" -> codeBlockWithLang
         "disp-quote" -> parseWithHeader (sectionLevel+1) parseBlockquote
-        "list" ->  parseWithHeader (sectionLevel+1)
-          ( case attrValue "list-type" e of
-              "bullet" -> bulletList <$> listitems
-              listType -> do
-                let start =
-                      fromMaybe 1 $
-                        ( filterElement (named "list-item") e
-                            >>= filterElement (named "label")
-                        )
-                          >>= safeRead . textContent
-                orderedListWith (start, parseListStyleType listType, DefaultDelim)
-                  <$> listitems
-          )
+        "list" ->  parseWithHeader (sectionLevel+1) parseList
         "def-list" -> parseWithHeader (sectionLevel+1) (definitionList <$> deflistitems)
         "sec" -> parseBlockWithHeader
         "abstract" -> parseBlockWithHeader
@@ -254,6 +242,18 @@ parseBlock (Elem e) = do
                                               mapM parseInline (elContent z)
             contents <- getBlocks e
             return $ blockQuote (contents <> attrib)
+         parseList = do
+            case attrValue "list-type" e of
+              "bullet" -> bulletList <$> listitems
+              listType -> do
+                let start =
+                      fromMaybe 1 $
+                        ( filterElement (named "list-item") e
+                            >>= filterElement (named "label")
+                        )
+                          >>= safeRead . textContent
+                orderedListWith (start, parseListStyleType listType, DefaultDelim)
+                  <$> listitems
          parseListStyleType "roman-lower" = LowerRoman
          parseListStyleType "roman-upper" = UpperRoman
          parseListStyleType "alpha-lower" = LowerAlpha
--- a/src/Text/Pandoc/Readers/JATS.hs
+++ b/src/Text/Pandoc/Readers/JATS.hs
@@ -161,23 +161,54 @@ parseBlock (Text (CData _ s _)) = if T.all isSpace s
                                      then return mempty
                                      else return $ plain $ trimInlines $ text s
 parseBlock (CRef x) = return $ plain $ str $ T.toUpper x
-parseBlock (Elem e) =
+parseBlock (Elem e) = do
+  sectionLevel <- gets jatsSectionLevel 
+  let parseBlockWithHeader = parseWithHeader (sectionLevel+1) (getBlocks e)
+
   case qName (elName e) of
         "p" -> parseMixed para (elContent e)
         "code" -> codeBlockWithLang
         "preformat" -> codeBlockWithLang
-        "disp-quote" -> parseBlockquote
-        "list" -> case attrValue "list-type" e of
-                    "bullet" -> bulletList <$> listitems
-                    listType -> do
-                      let start = fromMaybe 1 $
-                                  (filterElement (named "list-item") e
-                                               >>= filterElement (named "label"))
-                                   >>= safeRead . textContent
-                      orderedListWith (start, parseListStyleType listType, DefaultDelim)
-                        <$> listitems
-        "def-list" -> definitionList <$> deflistitems
-        "sec" -> gets jatsSectionLevel >>= sect . (+1)
+        "disp-quote" -> parseWithHeader (sectionLevel+1) parseBlockquote
+        "list" ->  parseWithHeader (sectionLevel+1)
+          ( case attrValue "list-type" e of
+              "bullet" -> bulletList <$> listitems
+              listType -> do
+                let start =
+                      fromMaybe 1 $
+                        ( filterElement (named "list-item") e
+                            >>= filterElement (named "label")
+                        )
+                          >>= safeRead . textContent
+                orderedListWith (start, parseListStyleType listType, DefaultDelim)
+                  <$> listitems
+          )
+        "def-list" -> parseWithHeader (sectionLevel+1) (definitionList <$> deflistitems)
+        "sec" -> parseBlockWithHeader
+        "abstract" -> parseBlockWithHeader
+        "ack" -> parseBlockWithHeader
+        "answer" -> parseBlockWithHeader
+        "answer-set" -> parseBlockWithHeader
+        "app" -> parseBlockWithHeader
+        "app-group" -> parseBlockWithHeader
+        "author-comment" -> parseBlockWithHeader
+        "author-notes" -> parseBlockWithHeader
+        "back" -> parseBlockWithHeader
+        "bio" -> parseBlockWithHeader
+        "explanation" -> parseBlockWithHeader
+        "glossary" -> parseBlockWithHeader
+        "kwd-group" -> parseBlockWithHeader
+        "list-item" -> parseBlockWithHeader
+        "notes" -> parseBlockWithHeader
+        "option" -> parseBlockWithHeader
+        "question" -> parseBlockWithHeader
+        "question-preamble" -> parseBlockWithHeader
+        "question-wrap-group" -> parseBlockWithHeader
+        "statement" -> parseBlockWithHeader
+        "supplement" -> parseBlockWithHeader
+        "table-wrap-foot" -> parseBlockWithHeader
+        "trans-abstract" -> parseBlockWithHeader
+        "verse-group" -> parseBlockWithHeader
         "graphic" -> para <$> getGraphic Nothing e
         "journal-meta" -> parseMetadata e
         "article-meta" -> parseMetadata e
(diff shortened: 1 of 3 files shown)


src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -321,24 +352,28 @@ parseBlock (Elem e) =
...
+                            labelText <- case filterChild (named "label") e of
+                                            Just t -> (<> ("." <> space)) <$>
+                                                      getInlines t
+                                            Nothing -> return mempty
+                            headerText <- case filterChild (named "title") e `mplus`
+                                                (filterChild (named "info") e >>=
+                                                    filterChild (named "title")) of
+                                              Just t  -> (labelText <>) <$>
+                                                          getInlines t
+                                              Nothing -> return labelText
```
noahmalmed, line 367: Not sure if this logic is correct: 

Previously, if there was a label and no title, we would not include the label in the AST. That seems wrong to me, so I switched it to include the label text (if there is any) in the AST.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -321,24 +352,28 @@ parseBlock (Elem e) =
...
+                            oldN <- gets jatsSectionLevel
+                            modify $ \st -> st{ jatsSectionLevel = n }
+                            blocks <- mBlocks
+                            let ident = attrValue "id" e
+                            modify $ \st -> st{ jatsSectionLevel = oldN }
+                            return $ (if 
+                              headerText == mempty 
+                            then mempty 
+                            else headerWith (ident,[],[]) n' headerText) <> blocks
```
noahmalmed, line 376: This is also something I wasn't sure about, but seemed wrong. Previously, we would include a header block even if there was no content to put inside of it (`headerText == mempty`). Without this `if` check, every instance of all the tags that use this function will include a header block even if there isn't any content to put inside of it

src/Text/Pandoc/Readers/JATS.hs
kamoe: I agree, this looks good to me.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -161,23 +161,54 @@ parseBlock (Text (CData _ s _)) = if T.all isSpace s
...
-                                  (filterElement (named "list-item") e
-                                               >>= filterElement (named "label"))
-                                   >>= safeRead . textContent
-                      orderedListWith (start, parseListStyleType listType, DefaultDelim)
-                        <$> listitems
-        "def-list" -> definitionList <$> deflistitems
-        "sec" -> gets jatsSectionLevel >>= sect . (+1)
+        "disp-quote" -> parseWithHeader (sectionLevel+1) parseBlockquote
+        "list" ->  parseWithHeader (sectionLevel+1)
+          ( case attrValue "list-type" e of
```
kamoe, line 174: This is the point where I would just define a separate function (e.g. parseList), for readability.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -321,24 +352,28 @@ parseBlock (Elem e) =
...
+                                            Nothing -> return mempty
+                            headerText <- case filterChild (named "title") e `mplus`
+                                                (filterChild (named "info") e >>=
+                                                    filterChild (named "title")) of
+                                              Just t  -> (labelText <>) <$>
+                                                          getInlines t
+                                              Nothing -> return labelText
+                            oldN <- gets jatsSectionLevel
+                            modify $ \st -> st{ jatsSectionLevel = n }
+                            blocks <- mBlocks
```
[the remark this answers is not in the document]
noahmalmed: Ah @kamoe , I'm relatively new to haskell so apologies if my understanding/vocabulary is a bit off: 

My understanding is that the monadic action wouldn't run until we invoke it using the `<-`, so by passing `blocks` wrapped in a monad 
```
parseWithHeader (sectionLevel+1) (getBlocks e)
```
`(getBlocks e)` wouldn't actually evaluate until execution reach L:370, which is basically how this function worked before. I feel relatively confident this is correct because the jats-reader test has a bunch of nested headers that evaluate correctly

src/Text/Pandoc/Readers/JATS.hs
noahmalmed: fixed!

src/Text/Pandoc/Readers/JATS.hs
kamoe: Looks correct to me. I do not know if there is any rule in ATL for keeping empty headers, if there is not, then I would say this is a good solution.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -161,23 +161,42 @@ parseBlock (Text (CData _ s _)) = if T.all isSpace s
...
-                    listType -> do
-                      let start = fromMaybe 1 $
-                                  (filterElement (named "list-item") e
-                                               >>= filterElement (named "label"))
-                                   >>= safeRead . textContent
-                      orderedListWith (start, parseListStyleType listType, DefaultDelim)
-                        <$> listitems
-        "def-list" -> definitionList <$> deflistitems
-        "sec" -> gets jatsSectionLevel >>= sect . (+1)
+        "disp-quote" -> parseWithHeader (sectionLevel+1) parseBlockquote
```
kamoe, line 172: I think this logic is correct and well thought out, but jus a suggestion, since this had me wondering for a while: Would you call the new function something like  `wrapWithHeader` instead of `parseWithHeader`? 

This is because you now cleverly outsource the actual block parsing out of the function, and it now expects the (would-be) parsed block as a parameter. I know we discussed the fact that the execution of `getBlocks` actually happens inside the `parseWithHeader` function, but calling it `parseWithHeader` is a bit confusing, specially when you look at these three cases, `disp-quote`, `list`, `def-list`,  where you seem to have to parse the blocks twice: one to prepare the parameter, and a second time inside the actual function. 

Again, this is really just a suggestion, and this won't have an impact on output, but for the sake of clarity might help others understand what is happening. Up to you :)

kamoe:
@jgm What would be next steps here? I do think this is a good solution for the issue, and I have added my comments. I do not think I can -or should- approve this. How do we proceed? Thanks.

src/Text/Pandoc/Readers/JATS.hs
noahmalmed: That makes sense! I'll make that change 👍

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -321,24 +352,28 @@ parseBlock (Elem e) =
...
-                                                  getInlines t
-                                      Nothing -> return mempty
-                     oldN <- gets jatsSectionLevel
-                     modify $ \st -> st{ jatsSectionLevel = n }
-                     b <- getBlocks e
-                     let ident = attrValue "id" e
-                     modify $ \st -> st{ jatsSectionLevel = oldN }
-                     return $ headerWith (ident,[],[]) n' headerText <> b
+         wrapWithHeader n mBlocks = do 
+                            isbook <- gets jatsBook
```
jgm, line 356: Here I'd indent the content under the `do` less, so that it all fits in 80 columns and is more readable.

test/jats-reader.native
```diff
@@ -226,6 +226,37 @@ Pandoc
...
+      , Str "for"
+      , Space
+      , Str "a"
+      , Space
+      , Str "statment."
+      , Space
+      , Str "CAP"
+      , Space
+      , Str "TITLE"
+      ]
```
jgm, line 246: I'm ignorant of JATS, so I don't know what a document with both a label and title is supposed to look like, but this seems rather odd, just appending the two.  Is that right?  OR is that combination unlikely to come up in practice?

kamoe: There is a thorough explanation by @coryschires  [here](https://github.com/jgm/pandoc/issues/8718#issuecomment-1522320635), under "_What's the difference between label and title?_", although it is quite lengthy. 

TLDR; A label is the type of content you are displaying, including an enumerator, whilst the title is the actual name of the content you are displaying.  (See official usage from the JATS website [here](https://jats.nlm.nih.gov/archiving/tag-library/1.3/element/label.html)). A simple example is 

```
<supplement>
    <label>Table 1.</label>
    <title>Average yearly growth</title>
</supplement>
```
Or, also, from @coryschires's own post:
```
<sec>
  <label>3.</label>
  <title>Conclusions</title>
</sec>
```
So, you do indeed want title and label appended together in the Header, and it is not uncommon. 

That is the reason I suggested [here](https://github.com/jgm/pandoc/issues/8718#issuecomment-1532223808) that we could not split the parsing of "title" and "label" as separate cases, since they both need to be displayed side by side when co-ocurring.

jgm: Thanks for the explanation. It seems to me that maybe for *pandoc*'s purposes, the label should simply be suppressed.  E.g., suppose you use pandoc to convert a JATS document to PDF via LaTeX. LaTeX will supply its own label for sections and figures, so if we include the label we'll get doubles...

kamoe: Ohhh I see. Well, if @coryschires does not have any objection, I don't mind suppressing the label in that case. @coryschires, what do you think?

coryschires: Yeah, I agree that JATS distinction between `label` and `title` is unusual and causes needless confusion. Unfortunately, I don't think there's an obviously correct answer to this question.

With that in mind, I think @jgm's plan is good. Let's suppress `<label>` and see how people respond. It seems like that would be the cleanest fit with Pandoc's AST, and we can always revisit it later.

noahmalmed:
Thanks for the review @jgm . I've just updated the PR

jgm:
thank you!

PR #8857, opened 2023-05-17
doc/lua-filters: use full field name in example

MattDodsonEnglish:
Remove old `.c` alias that mimics the JSON representation:
- Since 2.15ish, the lua structure no longer closely resembles that JSON representation
- even if it did, introductory examples should be explicit. Using an undocumented alias creates ambiguity.


tarleb:
Thanks!

Issue #8584, opened 2023-01-28
Fix RST raw directive to allow file includes?


spollard:
The reStructuredText raw directive can take an option named :file: with a filename value, in which case the contents of the filename are passed in as the body of the directive (https://docutils.sourceforge.io/docs/ref/rst/directives.html#raw-data-pass-through), but only if the output format on the raw directive is the same as the output format that the document is actually being converted to. 

At the moment, I don't think the RST parser's raw directive takes into account the output format (See https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/Readers/RST.hs#L665). I agree with this comment (https://github.com/jgm/pandoc/issues/2716#issuecomment-275157082) that the readers shouldn't know the output format or behave differently for a given output format, so I think this is fine. 

I think the RST parser only works with raw content directly in the body of the raw directive, but not read in from a separate file.  I don't think it would be too difficult to implement this, but my Haskell is rusty and I bet somebody else knows how to 1. detect if the :file: option is part of the raw directive 2.  read the contents of a file and 3. include the contents as the body of the raw directive. 

My guess is the solution might be using `insertIncludedFile` or a stripped out version of the `includeDirective` (https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/Readers/RST.hs#L456). Looks like you can detect the :file: field similarly to this `let startLine = lookup "start-line" fields >>= safeRead` from the `includeDirective.`

Steps to reproduce:
1. Run `rst2html including.rst` and see that Included has been included (docutils must be installed)
2. Try running `pandoc including.rst` and see that Included is not there

Files
include_me.html
```
<div>Included</div>
```

including.rst
```

Including
============

.. raw:: html
   :file: include_me.html
```

pandoc 2.7.3 on Ubuntu 20.04
Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.8.1

jgm:
You're using a version of pandoc that was released over 3 years ago. Please try with the latest before reporting a bug.

spollard:
I just got the same result using version 3.0.1.

spollard:
I may have time to take a crack at it soon.Thanks for the reminder! I'll start by trying to implement what I described in the paragraph starting with "My guess...", unless you have a better idea.

spollard:
[...]

PR #8858, opened 2023-05-18
Sketched out fix for letting raw directive in RST include files

spollard:
Started adding functionality to reStructuredText's raw directive to allow it to include files.

Some missing features for now. 

See https://github.com/jgm/pandoc/issues/8584


frasertweedale:
Please see my comment about security in the corresponding issue: https://github.com/jgm/pandoc/issues/8584#issuecomment-1652690237.

PR #8860, opened 2023-05-18
MANUAL: fix typo

sdhoward:
It should have the same labels as the table that comes 10 lines later. I didn't file an issue for this since it's so small.


jgm:
thanks!

Issue #8869, opened 2023-05-24
Usage of "" in header-args of Org src-block breaks export to Jupyter notebook


torfjelde:
**Problem**
org -> ipynb is seemingly unable to handle usage of `""` in the header args of a src-block, e.g. the following works just fine:

```org
#+begin_src jupyter-julia :exports both
1 + 2
#+end_src
```

resulting in 

```json
{
 "cells": [
  {
   "cell_type": "code",
   "execution_count": null,
   "metadata": {
    "exports": "both"
   },
   "outputs": [],
   "source": [
    "1 + 2\n"
   ],
   "id": "a2b36dc3-128b-4f98-a97f-0ff4b4a9ba0f"
  }
 ],
 "nbformat": 4,
 "nbformat_minor": 5,
 "metadata": {}
}
```

But if I do

```org
#+begin_src jupyter-julia :exports "both"
1 + 2
#+end_src
```

then we get an empty notebook:

```json
{
 "cells": [],
 "nbformat": 4,
 "nbformat_minor": 5,
 "metadata": {}
}
```

So it seems like something strange is going on when parsing the header args. Just for the record, usage of "" is completely valid Org format.

**Pandoc version?**
```sh
pandoc 3.1.2
Features: +server +lua
Scripting engine: Lua 5.4
```

PR #8870, opened 2023-05-24
Fix for #8869 in addition to proper handling of quoted header args in Org

torfjelde:
In addition to fixing #8869, this should also properly handle cases such as

```org
#+begin_src python :session "my :session"
...
#+end_src
```

which is valid Org but is currently not supported by pandoc.

Note that this is, I believe, the first time I've actually touched Haskell, so feel free to make changes:) But at least my test cases seem to pass nicely now :+1:

### DIFF
--- a/src/Text/Pandoc/Readers/Org/Blocks.hs
+++ b/src/Text/Pandoc/Readers/Org/Blocks.hs
@@ -415,7 +415,7 @@ blockOption = try $ do
   return (argKey, paramValue)
 
 orgParamValue :: Monad m => OrgParser m Text
-orgParamValue = try $ fmap T.pack $
+orgParamValue = try $ fmap (stripQuotes . T.pack) $
   skipSpaces
     *> notFollowedBy orgArgKey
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Readers/Org/Blocks.hs
```diff
@@ -418,12 +418,15 @@ orgParamValue :: Monad m => OrgParser m Text
...
-    *> noneOf "\n\r" `many1Till` endOfValue
+    *> (quotedValue <|> regularValue)
     <* skipSpaces
  where
   endOfValue = lookAhead $  try (skipSpaces <* oneOf "\n\r")
                         <|> try (skipSpaces1 <* orgArgKey)
 
+  quotedValue = char '\"' *> manyTill anyChar (char '\"')
+
+  regularValue = noneOf "\n\r" `many1Till` endOfValue
```
jgm, line 429: This will parse spaces; is that intended?
The old parser would stop if it gets to spaces followed by an orgArgKey, while this one won't, if I understand correctly.

PR #8875, opened 2023-05-28
docs: clarify returning nil in lua filters

CodeSandwich:
This is what I was able to observe, but it wasn't documented.


CodeSandwich:
It's not actually always true, at least not for filters working on `Pandoc` objects. You probably should always return the modified input object.

PR #8879, opened 2023-05-30
Fix for #8788: update boolean flags to take an optional 'true' or 'false' value

### LINKED ISSUE #8788: Allow option flags to accept False value
# Context

I have a project where `pandoc` is used to generate a Reveal.JS presentation, and this project aims to be a presentation generator.

In order to allow for offline usage and proper archival, I use the `--embed-resources --standalone` option flags, and I want this to be the default behavior so I set them in the defaults parameters file.

# Observed Problem

During writing, it is much faster not to embedd the resources. But my defaults file activates the `--embed-resources`, and there is no way to overload the flag for deactivation. Providing `--embed-resources=false` is refused.

For the moment, I therefore deactivated the `embed-resources` in the defaults file, but shall take care to activate it in the rendering pipeline.

# Proposal

Allow for overloading the `embed-resources` option from defaults with a command-line value requesting no embedding.

In #8362, in particular with [one answer](https://github.com/jgm/pandoc/issues/8362#issuecomment-1501007776), there is a proposal to extend accepted values for `--embed-resources` flag with `none` value: this would be perfect.

samahri:
This pull request introduces an optional value to the --embed-resources flag. To ensure backward compatibility, the default value is set to true unless false is explicitly provided.

Resolves #8788

### DIFF
--- a/src/Text/Pandoc/App/CommandLineOptions.hs
+++ b/src/Text/Pandoc/App/CommandLineOptions.hs
@@ -297,7 +297,7 @@ options =
                   (\arg opt -> do
                         boolValue <- readBoolFromOptArg arg
                         return opt { optFileScope = boolValue })
-                  "[true|false]")
+                  "true|false")
                  "" -- "Parse input files before combining"
 
     , Option "" ["sandbox"]
@@ -305,7 +305,7 @@ options =
                   (\arg opt -> do
                         boolValue <- readBoolFromOptArg arg
                         return opt { optSandbox = boolValue })
-                  "[true|false]")
+                  "true|false")
                  ""
 
     , Option "s" ["standalone"]
@@ -313,7 +313,7 @@ options =
                   (\arg opt -> do
                         boolValue <- readBoolFromOptArg arg
                         return opt { optStandalone = boolValue })
-                  "[true|false]")
+                  "true|false")
                  "" -- "Include needed header and footer on output"
 
     , Option "" ["template"]
@@ -350,7 +350,7 @@ options =
                   (\arg opt -> do
                         boolValue <- readBoolFromOptArg arg
                         return opt { optAscii = boolValue })
-                  "[true|false]")
+                  "true|false")
                  ""  -- "Prefer ASCII output"
 
     , Option "" ["toc", "table-of-contents"]
@@ -358,7 +358,7 @@ options =
                  (\arg opt -> do
                         boolValue <- readBoolFromOptArg arg
                         return opt { optTableOfContents = boolValue })
-                 "[true|false]")
+                 "true|false")
                "" -- "Include table of contents"
 
     , Option "" ["toc-depth"]
@@ -505,7 +505,7 @@ options =
                   (\arg opt -> do
                         boolValue <- readBoolFromOptArg arg
                         return opt { optPreserveTabs = boolValue })
-                  "[true|false]")
+                  "true|false")
                  "" -- "Preserve tabs instead of converting to spaces"
 
     , Option "" ["tab-stop"]
@@ -552,7 +552,7 @@ options =
                         deprecatedOption "--self-contained" "use --embed-resources --standalone"
                         boolValue <- readBoolFromOptArg arg
                         return opt { optSelfContained = boolValue })
-                    "[true|false]")
+                    "true|false")
                  "" -- "Make slide shows include all the needed js and css (deprecated)"
 
     , Option "" ["embed-resources"] -- maybe True (\argStr -> argStr == "true") arg
@@ -560,7 +560,7 @@ options =
                   (\arg opt -> do
                         boolValue <- readBoolFromOptArg arg
                         return opt { optEmbedResources =  boolValue })
-                  "[true|false]")
+                  "true|false")
                  "" -- "Make slide shows include all the needed js and css"
 
     , Option "" ["request-header"]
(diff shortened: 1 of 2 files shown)


jgm:
Seems to me we should make all the boolean flags take an optional `[=true|false]` if we do one.
Do you agree?

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -536,8 +536,9 @@ options =
                  "" -- "Make slide shows include all the needed js and css (deprecated)"
 
     , Option "" ["embed-resources"]
-                 (NoArg
-                  (\opt -> return opt { optEmbedResources = True }))
+                 (OptArg
+                  (\arg opt -> return opt { optEmbedResources = maybe True (\argStr -> argStr == "true") arg })
```
jgm, line 540: This just checks for the string `"true"`.  So, `--embed-resources=1`, `--embed-resources=yes`, `--embed-resources=True`, etc. will all be *silently* interpreted as `--embed-resources=false`.

I think a better behavior would raise an error if the optional argument is present and is not `"true"` or `"false"`.

For examples of this kind of checking and error raising, see e.g. the code for the option `eol`.

jgm: I also think that, either in this PR or in a later change, we should uniformly change all boolean options so they take this optional parameter.  So the validation code should be factored out into a separate function that can be used elsewhere.

samahri: I've update the PR to include the suggested changes.

samahri:
Update: The PR now enables the user to add optional boolean values to the following flags:
- `--file-scope`
- `--sandbox`
- `-s`, `--standalone`
- `--ascii`
- `--toc`, `--table-of-contents`
- `-p`, `--preserve-tabs`
- `--self-contained`
- `--embed-resources`
- `--no-check-certificate`
- `--strip-comments`
- `--reference-links`
- `--list-tables`
- `--listings`
- `--incremental`
- `--section-divs`
- `--html-q-tags`
- `--trace`
- `--dump-args`
- `--ignore-args`
- `--fail-if-warnings`

Only `true` or `false` are accepted. The boolean value is optional to ensure backwards compatibility. If the flag is passed without a value, it will default to `true`.

jgm:
That's great. The CI failure isn't due to anything in your changes -- can you rebase again against main so we can get a clean CI?

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -529,15 +553,22 @@ options =
...
-                 (NoArg
-                  (\opt -> do
+                 (OptArg
+                  (\arg opt -> do
                     deprecatedOption "--self-contained" "use --embed-resources --standalone"
-                    return opt { optSelfContained = True }))
+                    case readBoolFromOptArg arg of 
+                           Just boolValue -> return opt { optSelfContained = boolValue }
+                           Nothing -> optError $ PandocOptionError "value must be either true or false")
+                    "[true|false]")
```
jgm, line 562: There's a lot of code repetition here.  Why not have
```
readBoolFromOptArg :: String -> ExceptT OptInfo IO Bool
```
(so that it can raise the error directly instead of just returning a Maybe value) and then this code can be simplified to
```hs
boolValue <- readBoolFromOptArg arg
return opt { opthSelfContained = boolValue }
```

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -1098,6 +1175,13 @@ readMetaValue s
   | s == "FALSE" = MetaBool False
   | otherwise    = MetaString $ T.pack s
 
+readBoolFromOptArg :: Maybe String -> Maybe Bool
+readBoolFromOptArg arg = case arg of 
+   Nothing      -> return True
+   Just "true"  -> return True
+   Just "false" -> return False
+   _            -> Nothing
+
```
jgm, line 1184: Do you think it would make sense to do `case map toLower arg of` so that `TRUE` and `True` will work also?

hftf:
I think this change is resulting in some confusing error messages. For example, I've been using `pandoc -so` (intending as both `-s` + `-o`) for a while, but recently it produces the unclear error message `value must be either true or false`. It took me a while to figure out I needed to express `-s -o` instead. Perhaps the error message should include the option name and the non-boolean value that was received. Or, a backwards compatible design that allows multiple shorthand options like `-so` to continue to be formatted that way.

jgm:
@hftf Oh yes, I think this is an undesirable regression!  You've always been able to collapse short flags like this and I would rather not change that.

@samahri I wonder if it's possible to enable the `=(true|false)` behavior only for long option names?  This may be a limitation of the option parsing library.

jgm:
@hftf if you could open a new issue for this it will ensure we don't lose track.

jgm:
As far as I can see, the only way to enable this only for long options would be to separate the clauses. Instead of
```hs
    , Option "s" ["standalone"]
```
we'd need
```hs
    , Option "s" []
```
(without the optional argument) and then
```hs
    , Option "" ["standalone"]
```
with the optional argument.  But this would cause these options to be listed on separate lines in `--help`, which is bad.

jgm:
Having already pushed out this backwards-incompatible change, we might just have to accept that `-so` &c. are no longer possible. But improving the error messages should be feasible.

jgm:
I've improved the message so that you'd now get:

    Argument of --standalone/-s must be either true or false

PR #8885, opened 2023-06-02
INSTALL.md: Use more relevant link to NetBSD/pkgsrc entry

dressupgeekout:
Just a minor detail, but pandoc hasn't been a 'work in progress' in pkgsrc for a while now.


jgm:
thanks!

PR #8887, opened 2023-06-03
Retain image query parameters in dokuwiki reader

ech0-de:
While converting dokuwiki syntax to gfm, I noticed that the query parameters of images were stripped from the output. In general this makes sense, as the parameters' semantics are specific to dokuwiki. However, I wanted to use a custom filter to convert these parameters properly. So this pull request retains the query parameters of image urls in the dokuwiki reader, by adding it as an extra `query` attribute.

As far as I can see, this wont have any (negative) effect on existing writers, but allows users to access them in a filter and do with them what they want.


jgm:
thanks

PR #8896, opened 2023-06-08
Use crypton-connection instead of connection

felixonmars:
Follows the change introduced in tls 1.7.0. Fixes building.

Should be merged after https://github.com/snoyberg/http-client/pull/506


jgm:
I want to wait for a release of http-client first on Hackage.

felixonmars:
http-client-tls 0.3.6.2 has been released with the change.

jgm:
Probably also need to explicitly set versions for tls, http-client, connection-crypton, and anything else that departs from the lts release we're targeting in stack.yaml.

jgm:
Still one failing build.
```
       In the dependencies for pandoc-3.1.3(+embed_data_files):
           crypton-connection must match >=0.3.1, but the Stack configuration has no specified
                              version (latest matching version is 0.3.1)
```
Add this to stack.yaml?

jgm:
I'll fix the stack build after merging.

PR #8914, opened 2023-06-19
Update Linux install links

harabat:
- Updated Arch link (previous one was 404'ing)
- Standardised some other links that weren't pointing to the package or the distro-specific search results


jgm:
thanks!

jgm:
Website will be updated on next release.

Issue #8933, opened 2023-06-30
docx to asciidoc conversion of strikethrough (line-through) formatted text incorrect.


kbroch-rivosinc:
Apparently strikethrough formatting used to be a special designation (that the AST recognized?) https://stackoverflow.com/questions/38804306/how-do-i-do-strikethrough-line-through-in-asciidoc

Now it it handle in css https://github.com/asciidoctor/asciidoctor/blob/be7356dafd3bdd82d9b515a59997da64f300eaf1/src/stylesheets/asciidoctor.css#L1808  using https://docs.asciidoctor.org/asciidoc/latest/text/custom-inline-styles/

Currently the pandoc conversion below  generates:

```
[line-through]*This should be striked-through*
```

It should instead generate:

```
[line-through]#This should be striked-through#
```

Details of running pandoc:

```
(venv)kbroch@penguin:~/chromeos-downloads $ pandoc --version
pandoc 3.1.4
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: /home/kbroch/.local/share/pandoc
Copyright (C) 2006-2023 John MacFarlane. Web: https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.
(venv)kbroch@penguin:~/chromeos-downloads $ pandoc -o strke-through.adoc strike-through.docx 
(venv)kbroch@penguin:~/chromeos-downloads $ cat strke-through.adoc 
[line-through]*This should be striked-through*
(venv)kbroch@penguin:~/chromeos-downloads $ lsb_release -irs
Debian
11
```

[strike-through.docx](https://github.com/jgm/pandoc/files/11920072/strike-through.docx)

PR #8934, opened 2023-06-30
update line-through for asciidoc writer to custom inline style

kbroch-rivosinc:
issue:  #8933

### DIFF
--- a/src/Text/Pandoc/Writers/AsciiDoc.hs
+++ b/src/Text/Pandoc/Writers/AsciiDoc.hs
@@ -517,7 +517,7 @@ inlineToAsciiDoc opts (Strong lst) = do
   return $ marker <> contents <> marker
 inlineToAsciiDoc opts (Strikeout lst) = do
   contents <- inlineListToAsciiDoc opts lst
-  return $ "[line-through]*" <> contents <> "*"
+  return $ "[line-through]#" <> contents <> "#"
 inlineToAsciiDoc opts (Superscript lst) = do
   contents <- inlineListToAsciiDoc opts lst
   return $ "^" <> contents <> "^"
--- a/test/writer.asciidoc
+++ b/test/writer.asciidoc
@@ -425,7 +425,7 @@ So is *_this_* word.
 
 This is code: `>`, `$`, `\`, `\$`, `<html>`.
 
-[line-through]*This is _strikeout_.*
+[line-through]#This is _strikeout_.#
 
 Superscripts: a^bc^d a^_hello_^ a^hello there^.
 
--- a/test/writer.asciidoctor
+++ b/test/writer.asciidoctor
@@ -426,7 +426,7 @@ So is *_this_* word.
 
 This is code: `+>+`, `+$+`, `+\+`, `+\$+`, `+<html>+`.
 
-[line-through]*This is _strikeout_.*
+[line-through]#This is _strikeout_.#
 
 Superscripts: a^bc^d a^_hello_^ a^hello there^.
 


jgm:
Q: Is this an asciidoctor-only feature?
If it only works with asciidoctor and not asciidoc, we need to put this under a conditional.

kbroch-rivosinc:
> Q: Is this an asciidoctor-only feature? If it only works with asciidoctor and not asciidoc, we need to put this under a conditional.

I thiink it is an asciidoc feature, but here's detailed pros/cons:

* pro: in asciidoc spec: https://docs.asciidoctor.org/asciidoc/latest/text/custom-inline-styles/
* con: it is css controlled at end of the day
* pro: works with asciidoctor-pdf out of the box: `asciidoctor-pdf strke-through.adoc`
[strke-through.pdf](https://github.com/jgm/pandoc/files/11924982/strke-through.pdf)

Wondering of @mojavelinux could add any more insights?

mojavelinux:
line-through has always been recognized as a built-in role in AsciiDoc. The default stylesheet/theme will apply the line-through text decoration to this span of text.

(In the planned semantic HTML converter, we will likely be switching to the `<s>` tag in the HTML output. However, that's a presentation concern and thus won't change the AsciiDoc source).

mojavelinux:
> pro: in asciidoc spec: docs.asciidoctor.org/asciidoc/latest/text/custom-inline-styles

FYI: This is not the AsciiDoc spec. This is the initial contribution to the AsciiDoc spec based on the defacto implementation of AsciiDoc (how it is implemented in Asciidoctor). Once the first version of the AsciiDoc spec is completed, the documentation at that URL will become the user-facing guide for AsciiDoc (and may be a mirror of the official guide).

Please note that there is no "asciidoc vs asciidoctor". There is only one AsciiDoc. It's currently in the process of being formalized as a specification. Asciidoctor is the reference for that specification and will become an implementation once there is a specification to implement. It's possible there will be changes during this time, but whatever is decided on becomes AsciiDoc and there is nothing else.

jgm:
When I say `asciidoc` I'm referring to the older Python implementation, which last I checked was still being used (e.g. by git for its documentation), and which has somewhat different syntax in places. Pandoc distinguishes between `asciidoc` and `asciidoctor` because of this.

If this is no longer the case, and the old implementation is not being used by anyone, we could adjust this and unify the syntax so that `asciidoc` = `asciidoctor`.

Remember that one important use of pandoc is converting legacy documentation, so even if the old python `asciidoc` isn't much used any more, it's valuable to be able to support conversion from old documentation bases that used it.

kbroch-rivosinc:
@jgm IMO there's not a lot of value in supporting a doc conversion workflow that still wants to use the old asciidoc (py) format as output (there isn't any support for asciidoc(tor) in input, right?): 

* New doc workflows would choose, to the new supported format. 
*  If for some reason I had a workflow that converted to old asciidoc format in the middle to support some other tooling, I'd be worried this would be deprecated and eventually EOL'ed.

mojavelinux:
I interact with the AsciiDoc community on a daily basis, and I'm not aware of anyone who's still actively using AsciiDoc.py to convert content being written today. Naturally, older documents still exist that were written during the time of AsciiDoc.py and, as a result, may still rely on it to be converted. However, no one should be generating new documents with AsciiDoc.py as a target. This is even acknowledged on that project's website: "AsciiDoc.py is a legacy processor for this syntax".

As the creator and maintainer of Asciidoctor, I would prefer not to see an asciidoctor option in pandoc. As the lead of the AsciiDoc specification, I'm especially keen on not letting the impression to take hold that there are flavors of AsciiDoc. In fact, it's written into the charter of our group to prevent that from happening.

For trademark reasons, the AsciiDoc.py project has be renamed to asciidoc-py. What that means is that AsciiDoc.py is no longer AsciiDoc. To reiterate, there's only one AsciiDoc, and it's actively being formalized as a specification based on the syntax supported by Asciidoctor. For the past decade leading up to current specification effort, Asciidoctor has been the maintainer of the definition of AsciiDoc.

If you feel strongly that it's important to retain an output format that targets AsciiDoc.py (which is understandable), then I would encourage you to rename it to asciidoc-py. That would mean the output format named asciidoctor should become renamed to (or aliased to) asciidoc. Given this is not my project, I acknowledge the the choice is not mine to make. But I would hope you would take my recommendation under consideration.

If you have questions or concerns about what I've stated here, I'm glad to provide clarifications. You're also invited to ask those question in the project chat for AsciiDoc at https://chat.asciidoc.org.

jgm:
> As the creator and maintainer of Asciidoctor, I would prefer not to see an asciidoctor option in pandoc. 

There already is one.  That's because pandoc started supporting asciidoc before asciidoctor existed. When asciidoctor came along with some incompatible syntax, we differentiated by adding `asciidoctor` as a separate writer name.

We could rename things so that `asciidoc` does what `asciidoctor` used to and `asciidoc_legacy` or `asciidoc_py` does what `asciidoc` used to.  I will think about this, but of course there is a cost: possibly breaking workflows that currently work.  In any event, there should probably be a separate issue to track this; I'll set one up.

mojavelinux:
Thanks John.

jgm:
Now that we've gotten clear about terminology, let me rephrase my original question.  Is the output for strikethrough produced by this PR suitable for *both* new asciidoc (asciidoctor) and legacy python asciidoc?

If not, then we need some conditionals so we can give good results for both.

You can determine which variant you're using by doing
```hs
  isAsciidoctor <- gets asciidoctorVariant                         
```
Of course, if we implement the renaming discussed above, these internal names may also change, but for purposes of this PR let's go with what we have.

mojavelinux:
For the purpose of this issue, there is no difference in any event. The `line-through` role for strikethrough has been supported in the AsciiDoc syntax since it's inception.

jgm:
The build failures are not related to this PR. I think rebasing against main should give you a clean build and then we can merge.

PR #8940, opened 2023-07-04
Correct Russell's ethnicity in example

frabjous:
It's very important that pandoc not spread untruths about important philosophers.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -4151,7 +4151,7 @@ capital letter with a period, by at least two spaces.[^2]
 [^2]:  The point of this rule is to ensure that normal paragraphs
     starting with people's initials, like
 
-        B. Russell was an English philosopher.
+        B. Russell was a Welsh philosopher.
 
     do not get treated as list items.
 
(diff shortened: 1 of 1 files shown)


MANUAL.txt
```diff
@@ -4151,7 +4151,7 @@ capital letter with a period, by at least two spaces.[^2]
 [^2]:  The point of this rule is to ensure that normal paragraphs
     starting with people's initials, like
 
-        B. Russell was an English philosopher.
+        B. Russell was a Welsh philosopher.
```
jgm, line 4154: I wonder if this might be a complicated question. Wikipedia says: "Bertrand Russell was born at Trellech, Monmouthshire, now in Wales, into an aristocratic English family..."  Does being born in Wales make you ethnically Welsh?

What if we say "British"?

frabjous: "British" is fine, though his family roots are both Welsh and English,.

jgm:
Are you okay with the change I made?

frabjous:
Sure, that's a good sentence too!

PR #8945, opened 2023-07-07
Fix build (transitive dep not in lts)

hseg:
Commit 921b0949f4400af2e602982806e48222b7611bd9 bumped the skylighting-core version, which requires a newer version of xml-conduit than is available in lts-21.0.
Bump the version in stack.yaml


jgm:
thanks!

Issue #8889, opened 2023-06-04
Functions isBlockElement and parseMixed cancel each other out; JATS body blocks are treated as inlines


kamoe:
**Background**

In JATS, there exists a number of paragraph-level or [body blocks](https://jats.nlm.nih.gov/archiving/tag-library/1.3/chapter/nfd-blocks.html), and other structurally similar elements, that sit at the same level of a `<p>`. 

By definition from the JATS specification, they are _"elements, such as tables and figures, that are content units separated from other content visually and logically, typically with whitespace before and after them. These elements are typically used in the same places a paragraph may be used, for example, inside a section after the section title."_

Now, in JATS, the element `<p>` can contain some of such block-like elements, e.g. `<code>`:

```
<p>
   Start of a paragraph
    <code>Some code</code>
   End of a paragraph
</p>
```

Because `<code>` is supposed to sit at paragraph-level, the above should look like three independent paragraphs separated by a whitespace, although in JATS it is all just one paragraph. 

Pandoc seems to have taken this into account, with the `parseMixed` function of the JATS reader, where every element is parsed either as a block, or as an inline, as appropriate:

https://github.com/jgm/pandoc/blob/16f28ef5e945f3be14e05afb7d91f8adca18e49a/src/Text/Pandoc/Readers/JATS.hs#L202-L211

According tho the above, when a block-like element is found, the reader parses it as a block, creating a separate paragraph for it. Otherwise, the reader parses the element as an inline.

**The problem**
The output of the `isBlockElement` function determines if the element gets to be parsed as a block or as an inline (see L203 in code extract above). However, the `isBlockElement` function actually only returns TRUE if the input element is a `<p>` (See discussion [here](https://groups.google.com/g/pandoc-discuss/c/QDLNtNdObXU/)). Since `isBlockElement` is only ever called over _children_ of `<p>`*, `isBlockElement` is never TRUE**, and as a result, all children of `<p>` are always and systematically parsed as inlines (L208-L2011 above are never reached). 

So the whole purpose of the `parsedMixed` function is defeated. This is causing issue https://github.com/jgm/pandoc/issues/8804, for instance.

\*This is because `isBlockElement` is only ever called from `parseMixed`, and `parseMixed` is only ever called from the case of parsing `<p>`.
\** This is because in the JATS specification `<p>` cannot contain a `<p>`, in other words, no input element of `isBlockElement` is ever a `<p>` that could yield a value TRUE. 

**The root cause**
[...]

PR #8971, opened 2023-07-27
Fix display of block elements in JATS reader - Issue 8889

kamoe:
A number of block elements, like `disp-quote`, `list`, and `disp-formula`, were always treated as inlines if appearing inside paragraphs, even if their usage granted a separate block. The function `isElementBlock` has been refined to prevent this, and a number of specific parse cases have been added to `parseBlock`. Also some general cleanup and removal of redundant lines.

### DIFF
--- a/src/Text/Pandoc/Readers/JATS.hs
+++ b/src/Text/Pandoc/Readers/JATS.hs
@@ -221,7 +221,8 @@ parseBlock (Elem e) = do
                             else getBlocks e
         "disp-formula" -> if hasFormulaChild e
                             then blockFormula displayMath e
-                            else getBlocks e
+                            else divWith (attrValue "id" e, ["disp-formula"], []) 
+                                    <$> getBlocks e
         "?xml"  -> return mempty
         _       -> getBlocks e
    where parseMixed container conts = do
--- a/src/Text/Pandoc/Readers/JATS.hs
+++ b/src/Text/Pandoc/Readers/JATS.hs
@@ -105,17 +105,30 @@ instance HasMeta JATSState where
   deleteMeta field s = s {jatsMeta = deleteMeta field (jatsMeta s)}
 
 isBlockElement :: Content -> Bool
-isBlockElement (Elem e) = qName (elName e) `S.member` blocktags
-  where blocktags = S.fromList (paragraphLevel ++ lists ++ mathML ++ other) \\ S.fromList inlinetags
+isBlockElement (Elem e) = case qName (elName e) of
+            "disp-formula" -> if onlyOneChild e
+                                  then if hasFormulaChild e 
+                                          then False 
+                                          else case filterChild (named "alternatives") e of
+                                            Just a -> if hasFormulaChild a then False else True
+                                            Nothing -> True
+                                  else True
+            "alternatives" -> if hasFormulaChild e then False else True
+            _ -> qName (elName e) `S.member` blocktags
+
+  where blocktags = S.fromList (paragraphLevel ++ lists ++ formulae ++ other) \\ S.fromList canBeInline
         paragraphLevel = ["address", "answer", "answer-set", "array", "boxed-text", "chem-struct-wrap",
-            "code", "explanation", "fig", "fig-group", "graphic", "media", "preformat", "question", "question-wrap", "question-wrap-group"
+            "code", "explanation", "fig", "fig-group", "graphic", "media", "preformat", "question", "question-wrap", "question-wrap-group",
             "supplementary-material", "table-wrap", "table-wrap-group",
             "alternatives", "disp-formula", "disp-formula-group"]
         lists = ["def-list", "list"]
-        mathML = ["tex-math", "mml:math"]
+        formulae = ["tex-math", "mml:math"]
         other = ["p", "related-article", "related-object", "ack", "disp-quote",
             "speech", "statement", "verse-group", "x"]
-        inlinetags = ["alternatives", "tex-math", "mml:math", "x"]
+        canBeInline = ["tex-math", "mml:math", "related-object", "x"] 
+        onlyOneChild x = length (allChildren x) == 1
+        allChildren x = filterChildren (const True) x
+        
 isBlockElement _ = False
 
 -- Trim leading and trailing newline characters
@@ -203,6 +216,12 @@ parseBlock (Elem e) = do
              else divWith (attrValue "id" e, ["caption"], []) <$> wrapWithHeader 6 (getBlocks e)
         "fn-group" -> parseFootnoteGroup
         "ref-list" -> parseRefList e
+        "alternatives" -> if hasFormulaChild e
+                            then blockFormula displayMath e
+                            else getBlocks e
+        "disp-formula" -> if hasFormulaChild e
+                            then blockFormula displayMath e
+                            else getBlocks e
         "?xml"  -> return mempty
         _       -> getBlocks e
    where parseMixed container conts = do
@@ -586,8 +605,11 @@ parseInline (Elem e) =
              let attr = (attrValue "id" e, [], [])
              return $ linkWith attr href title ils'
 
-        "disp-formula" -> formula displayMath
-        "inline-formula" -> formula math
+        "alternatives" -> if hasFormulaChild e
+                            then inlineFormula math e
+                            else innerInlines id
+        "disp-formula" -> inlineFormula displayMath e
+        "inline-formula" -> inlineFormula math e
         "math" | qURI (elName e) == Just "http://www.w3.org/1998/Math/MathML"
                    -> return . math $ mathML e
         "tex-math" -> return . math $ textContent e
@@ -600,11 +622,14 @@ parseInline (Elem e) =
         _          -> innerInlines id
    where innerInlines f = extractSpaces f . mconcat <$>
                           mapM parseInline (elContent e)
-         mathML x =
-            case readMathML . showElement $ everywhere (mkT removePrefix) x of
-                Left _ -> mempty
-                Right m -> writeTeX m
-         formula constructor = do
+         codeWithLang = do
+           let classes' = case attrValue "language" e of
+                               "" -> []
+                               l  -> [l]
+           return $ codeWith (attrValue "id" e,classes',[]) $ strContentRecursive e
+
+inlineFormula ::  PandocMonad m => (Text->Inlines) -> Element -> JATS m Inlines
+inlineFormula constructor e = do
             let whereToLook = fromMaybe e $ filterElement (named "alternatives") e
                 texMaths = map textContent $
                             filterChildren (named  "tex-math") whereToLook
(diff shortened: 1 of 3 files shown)


src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -363,9 +369,7 @@ parseBlock (Elem e) = do
          wrapWithHeader n mBlocks = do
                       isBook <- gets jatsBook
                       let n' = if isBook || n == 0 then n + 1 else n
-                      headerText <- case filterChild (named "title") e `mplus`
```
kamoe, line 366: There is no `info` element in JATS. This is legacy from the DocBook reader. As such this line is redundant.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -15,7 +15,7 @@ Conversion of JATS XML to 'Pandoc' document.
 
 module Text.Pandoc.Readers.JATS ( readJATS ) where
 import Control.Monad.State.Strict ( StateT(runStateT), gets, modify )
-import Control.Monad (forM_,  when, unless, MonadPlus(mplus))
```
kamoe, line 18: `mplus` is now redundant, since line dealing with the non-existing`info` element is redundant and was removed.

test/jats-reader.xml
```diff
@@ -289,8 +379,16 @@ These should not be escaped:  \$ \\ \&gt; \[ \{</preformat>
       </list-item>
     </list>
     <p>List styles:</p>
-    <list list-type="order"></list>
```
kamoe, line 292: `<list>` cannot be empty. Dummy elements added.

test/jats-reader.xml
```diff
@@ -406,33 +504,50 @@ These should not be escaped:  \$ \\ \&gt; \[ \{</preformat>
...
+      </list>
+      End of paragraph.
+    </p>
+  </sec>
 </sec>
 <sec id="inline-markup">
   <title>Inline Markup</title>
   <p>This is <italic>emphasized</italic>, and so <italic>is
     this</italic>.</p>
-  <p>This is <bold role="strong">strong</bold>, and so
```
kamoe, line 422: Attribute `@role` does not exist for `<bold>`

test/jats-reader.xml
```diff
@@ -406,33 +504,50 @@ These should not be escaped:  \$ \\ \&gt; \[ \{</preformat>
...
-  <p><bold role="strong"><italic>This is strong and em.</italic></bold></p>
-  <p>So is <bold role="strong"><italic>this</italic></bold> word.</p>
+  <p><bold><italic>This is strong and em.</italic></bold></p>
+  <p>So is <bold><italic>this</italic></bold> word.</p>
+  <p><bold><italic>This is strong and em.</italic></bold></p>
+  <p>So is <bold><italic>this</italic></bold> word.</p>
   <p>This is code: <monospace>&gt;</monospace>, <monospace>$</monospace>,
     <monospace>\</monospace>, <monospace>\$</monospace>,
     <monospace>&lt;html&gt;</monospace>.</p>
-  <p>This is <sc role="smallcaps">small caps</sc>.</p>
```
kamoe, line 435: Attribute `@role` does not exist for `<sc>`

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -105,31 +105,30 @@ instance HasMeta JATSState where
...
+            "alternatives" -> if hasFormulaChild e then False else True
+            _ -> qName (elName e) `S.member` blocktags
+
+  where blocktags = S.fromList (paragraphLevel ++ lists ++ formulae ++ other) \\ S.fromList canBeInline
+        paragraphLevel = ["address", "answer", "answer-set", "array", "boxed-text", "chem-struct-wrap",
+            "code", "explanation", "fig", "fig-group", "graphic", "media", "preformat", "question", "question-wrap", "question-wrap-group",
             "supplementary-material", "table-wrap", "table-wrap-group",
             "alternatives", "disp-formula", "disp-formula-group"]
         lists = ["def-list", "list"]
-        mathML = ["tex-math", "mml:math"]
```
kamoe, line 115: Renamed as to not conflict with `mathML` function at the end of the reader.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -105,31 +105,30 @@ instance HasMeta JATSState where
...
+        formulae = ["tex-math", "mml:math"]
         other = ["p", "related-article", "related-object", "ack", "disp-quote",
             "speech", "statement", "verse-group", "x"]
-        inlinetags = ["email", "ext-link", "uri", "inline-supplementary-material",
```
kamoe, line 118: Refined as discussed in https://github.com/jgm/pandoc/issues/8889#issuecomment-1645076773

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -217,6 +216,13 @@ parseBlock (Elem e) = do
              else divWith (attrValue "id" e, ["caption"], []) <$> wrapWithHeader 6 (getBlocks e)
         "fn-group" -> parseFootnoteGroup
         "ref-list" -> parseRefList e
+        "alternatives" -> if hasFormulaChild e
```
kamoe, line 219: Given these can now be considered as block elements, as per the refinement of `isBlockElement`, the respective parse cases needed adding.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -363,9 +369,7 @@ parseBlock (Elem e) = do
          wrapWithHeader n mBlocks = do
                       isBook <- gets jatsBook
                       let n' = if isBook || n == 0 then n + 1 else n
-                      headerText <- case filterChild (named "title") e `mplus`
-                                          (filterChild (named "info") e >>=
```
kamoe, line 367: There is no `info` element in JATS. This is legacy from the DocBook reader. As such these lines are redundant.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -616,24 +623,44 @@ parseInline (Elem e) =
         _          -> innerInlines id
    where innerInlines f = extractSpaces f . mconcat <$>
                           mapM parseInline (elContent e)
-         mathML x =
```
kamoe, line 619: Created independent `mathML` function, as code repeated in three instances.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -616,24 +623,44 @@ parseInline (Elem e) =
...
-            case readMathML . showElement $ everywhere (mkT removePrefix) x of
-                Left _ -> mempty
-                Right m -> writeTeX m
-         formula constructor = do
```
kamoe, line 623: Split `formula` function into `inlineFormula` and `blockFormula` to deal with each representation in the native document. (Could further recycle common code, but that is scope for a further PR, for clarity leaving as is for now).

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -616,24 +623,44 @@ parseInline (Elem e) =
...
                       qURI  (elName x) ==
                                       Just "http://www.w3.org/1998/Math/MathML"
-         removePrefix elname = elname { qPrefix = Nothing }
-         codeWithLang = do
-           let classes' = case attrValue "language" e of
-                               "" -> []
-                               l  -> [l]
-           return $ codeWith (attrValue "id" e,classes',[]) $ strContentRecursive e
+
+formulaChildren :: Element -> [Element]
```
kamoe, line 662: Helper functions to `isBlockElement` and the two new parse cases.

test/jats-reader.xml
```diff
@@ -660,7 +775,7 @@ These should not be escaped:  \$ \\ \&gt; \[ \{</preformat>
   <title>Tables</title>
   <sec id="tables-with-headers">
     <title>Tables with Headers</title>
-    <table>
```
kamoe, line 663: Table can only occur inside a `<table-wrap>`

test/jats-reader.xml
```diff
@@ -1211,27 +1326,14 @@ These should not be escaped:  \$ \\ \&gt; \[ \{</preformat>
...
+    </table></table-wrap>
   </sec>
   <sec id="table-with-footer">
     <title>Table with footer</title>
-    <table>
+    <table-wrap><table>
       <col align="left" />
       <col align="left" />
       <col align="left" />
-      <tbody>
```
kamoe, line 1222: For some strange reason, `<tfoot>` has to appear before `<tbody>`. Swapped them around.

jgm:
The failing test is
```
    JATS
      inline code
        basic:                                                       OK
        lang:                                                        FAIL
          test/Tests/Readers/JATS.hs:30:
          
          <p>
            <code language="c">@&amp;</code>
          </p>
          - [ Para [ Code ( "" , [ "c" ] , [] ) "@&" ] ]
          + [ CodeBlock ( "" , [ "c" ] , [] ) "@&" ]
          Use -p '/inline code.lang/' to rerun this test only.
```
I think we decided in discussion that `<code>` is meant only for code blocks, not inline code.
So it might make sense to simply remove this test.  Alternatively, the test output could be changed.

kamoe:
But.Of.Course.

In line with the test for the writer, I have simply removed that line from the reader's test. Passed for me.

test/Tests/Readers/JATS.hs
```diff
@@ -27,7 +27,6 @@ jats = purely $ readJATS def
 tests :: [TestTree]
 tests = [ testGroup "inline code"
           [ test jats "basic" $ "<p>\n  <monospace>@&amp;</monospace>\n</p>" =?> para (code "@&")
-          , test jats "lang" $ "<p>\n  <code language=\"c\">@&amp;</code>\n</p>" =?> para (codeWith ("", ["c"], []) "@&")
```
kamoe, line 30: Removed as code no longer used for inline code, as per discussion

test/jats-reader.native
```diff
@@ -366,6 +375,428 @@ Pandoc
...
+  , Para
+      [ Str "Einstein"
+      , Space
+      , Str "showed"
+      , Space
+      , Str "that"
+      ]
+  , Div
+      ( "" , [ "disp-formula" ] , [] )
+      [ LineBlock [ [ Math DisplayMath "e=mc^2" ] ] ]
```
jgm, line 567: Why the LineBlock?
I wonder whether the Div with class "disp-formula" is really needed.  Why not just a LineBlock or even just a Para with the Math DisplayMath inside it?

jgm (approved):
This looks good to me.  I've made one query about the treatment of formulas outside paragraphs.  This is always something that could be refined later, but we should probably dicuss now before merging.

test/jats-reader.native
kamoe: This is because in this case the `disp-formula` is not the inmediata parent of the math elements, but the grandparent.  So the `disp-formula` builds the Div, then the `alternatives` parent builds the LineBlock. 

Also, I suppose it's a balance between a clean output and a clean code. The only way of getting rid of the Div is to start adding what can become convoluted conditions to the parse cases for `disp-formula` and `alternatives`, while right now they are quite simple, the recursion does the rest. 

But I see the point. I could definitely give it more thought and see in the following days if I find a way to write the added conditions nicely.

jgm: Simple is good.  But why 

> the alternatives parent builds the LineBlock.

What does LineBlock have to do with alternatives?

kamoe: LineBlock has to do with whoever is the direct parent of math elements. `alternatives` does not always build a LineBlock, it does only when it is the parent of `mml:math` or `tex-math`.

Conversely, `disp-math` doesn't always build a Div. It does when it is not the direct parent of math elements; but if it is, it too builds a LineBlock.

jgm: Okay, but then my question is: what does LineBlock have to do with mml:math or tex-math?

kamoe: There is no specific or intrinsic relationship between LineBlock and math elements. We could have chosen another type of block, like Div. 

LineBlock fitted nicely with the code for  `blockFormula`, a new function derived from `inlineFormula` (it has the same input and output type as `mconcat`, the function it replaces in the block version of the formula builder).

jgm: I think Para is the normal default choice for display math that is in a block by itself.  Unless there's a good specific reason to use LineBlock, we should use Para.

kamoe: OK, fair enough. Just replaced LineBlock by Para.

jgm:
Thank you!

PR #8978, opened 2023-08-01
Documentation - FAQs - Links

diogoalmiro:
Fixes the link to the manual that wasn't creating an anchor tag in HTML on line 119.

I also suggest that the link to the issue had a text like the all other links in this page.


jgm:
Great, thanks.

Issue #8118, opened 2022-06-09
Add capability to bypass beamer frame in section slides not just for "notes" slides


tchevri:
I would like to follow up with the resolution provided in <https://github.com/jgm/pandoc/issues/7857> and this issue I raised please: <https://github.com/jgm/pandoc/issues/4317#issuecomment-524169706>

At the time, [jgm](https://github.com/jgm) responded [on Aug 23, 2019](https://github.com/jgm/pandoc/issues/4317#issuecomment-524169706):

> @tchevri I'm sorry about this -- it appears one can't please everyone. Unfortunately your particular use wasn't represented in the discussion. I had assumed that people who didn't want the content to appear on slides could simply hide the content (using a filter or whatever), but your case is different; you don't want to hide the content; you want it to appear in the document outside the slide environment

and concluded with:

>But I can't think of a good workaround right now

It's been a couple of years now and a solution was found for another user in a nearly identical situation: [here](https://github.com/jgm/pandoc/issues/7856) and [here](https://github.com/jgm/pandoc/issues/7857).  
Basically the [src/Text/Pandoc/Writers/LaTeX.hs](https://github.com/jgm/pandoc/commit/52b78b10c81e99926d560bc383938006be3b0458#diff-3352fb3c002000fc6dd779f959bff3a83ab3b18abe57bc65ad07166faf6a2168) code was enhanced to include:

``` case titleBs of
              [] -> Div (ident,"section":dclasses,dkvs) xs
              [Div (_,"notes":_,_) _] ->  -- see #7857, don't create frame
                    -- just for speaker notes after section heading
```

Now any above slide level content starting with `::: notes` will not be added onto a slide.
Is it possible to add a couple more lines to entertain my use case as well please?  In line with [the beamer userguide section 21.3 Details on Modes](http://tug.ctan.org/macros/latex/contrib/beamer/doc/beameruserguide.pdf#subsection.21.3), the `\mode` command controls what part of your text belongs to which mode. It is to be specified outside frames. This is a key feature for controlling how `beamer` behaves.

**Therefore, my hope is to be able to escape a command such as `\mode<presentation>` (just like `::: notes`) and not have it forced onto a slide please. To achieve this, would the code need to be wrapped in a ` ```{=beamer} ` command or would it be possible to escape it directly please?**

Many thanks in advance. thomas

tchevri:
Thank you so much for such prompt follow up despite the current heavy flow on the list and the many other tasks you will be busy with.

[...]

Issue #4317, opened 2018-01-28
Missing vertical slides in revealjs


Wenzel:
Hi !

I am trying to generate slides from `Markdown` to `RevealJS` with pandoc.

The problem is that i don't have **vertical slides** , like revealjs can provide. instead everything is packed into the first level slide.

Here is a short example to reproduce:

Makefile:
~~~Makefile
all:
	pandoc \
		-t revealjs \
		--standalone \
		--self-contained \
		-o prez.html \
		prez.md

download_revealjs:
	wget https://github.com/hakimel/reveal.js/archive/master.tar.gz
	tar -xzvf master.tar.gz
	mv reveal.js-master reveal.js
	rm master.tar.gz
~~~

prez.md
~~~Markdown
---
title:  'Test RevealJS'
author:
- Tarral Mathieu
---

# Introspection ?

- hello
- world

## sub

- hello world
~~~

- `make download_revealjs`
- `make all`

The output:
![capture d ecran de 2018-01-28 16-37-13](https://user-images.githubusercontent.com/964610/35483358-8d901f9a-0449-11e8-99ae-2249fe737983.png)

My pandoc version is:
~~~
pandoc 1.19.1
Compiled with pandoc-types 1.17.0.5, texmath 0.9.1, highlighting-kate 0.6.4
Default user data directory: /home/wenzel/.pandoc
Copyright (C) 2006-2016 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
~~~

Do you know what's wrong with my Markdown ?

Thanks !

tarleb:
Please try running pandoc with `--slide-level=2`. This will tell pandoc to turn second-level headers into slides. Note that only the header of first-level slides will be shown, the first list in your example will not be visible.

If you have further questions, please direct them to the [pandoc-discuss mailing list](https://groups.google.com/forum/#!forum/pandoc-discuss); we try to keep this issue tracker reserved for bugs and enhancement proposals.

HDembinski:
Please reopen this issue, because it does not solve OPs problem. I am having the same issue.

tarleb:
The original issue seems to be resolved. @HDembinski, it is difficult to assess your issue without more information about your specific problem. Please also see <http://pandoc.org/help.html>.

HDembinski:
[...]

PR #8983, opened 2023-08-04
Solve #8118 and #4317 by adding 1 line to LaTeX.hs

tchevri:
Before Pandoc 2.7, choosing slide_level: 3 meant that ### was needed to create a slide. This commit https://github.com/jgm/pandoc/commit/5990f14ad497999141a5b975651f83d751ec421f caused this desired behavior to no longer be true and content above slide level is automatically included onto a slide, to accommodate compatibility with reveal.js format. At JGM's request, who could not think of a quick fix at the time, I posted a request on https://groups.google.com/g/pandoc-discuss/c/joqekZQRpq8, but no resolution was suggested.

This suggested simple 1 extra line change fixes these issues #8118 and #4317. Thank you in advance for entertaining this suggestion. FYI, I ran the tests to make sure all ok. Test suite test-pandoc: RUNNING... Test suite test-pandoc: PASS (I also compiled and used this pandoc.exe to make sure it works as expected and it does).

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -259,6 +259,9 @@ elementToBeamer slideLevel (Div (ident,"section":dclasses,dkvs)
               [Div (_,"notes":_,_) _] ->  -- see #7857, don't create frame
                     -- just for speaker notes after section heading
                     Div (ident,"section":dclasses,dkvs) xs
+                    -- See #8118 and #4317, don't create frame to change mode
+              [RawBlock (Format "latex") str] | "\\mode<presentation>" `T.isInfixOf` str ->
+                    Div (ident,"section":dclasses,dkvs) xs
               _  -> Div (ident,"section":dclasses,dkvs)
                      (h : Div ("","slide":dclasses,dkvs) (h:titleBs) : slideBs)
   | otherwise


jgm:
could you explain a bit more fully how this works?  You're matching on `\mode<presentation>` at the beginning of each slide?  So that people would need to add this to each slide?

tchevri:
Apologies, I used `ChatGPT` to write this code, and thanks to your query, I found out the suggested code does not work as I expected. The filter does not seem to pick up my command and see it as an exception. I am so sorry for having wasted your time (and mine!). I will close this pull request. A pity - it's been 4 years, since I can no longer use `pandoc` as you produce it and must compile my own version after "hacking" the `LaTeX.hs` file....

My use case is the same as in #4317, but I am more than happy to re-explain.

- I use `bookdown` to write lecture notes / slides. 
- The same project produces both lecture notes and lecture slides. This makes sure teaching slides are in sync with lecture notes (reproducibility) and allows to put extra content in the lecture notes that is only discussed orally if not just skipped during lectures. `Beamer` is a very powerful slide maker that allows precisely anyone to achieve this goal. 
- There is no vertical / horizontal slides though in `beamer`, as it produces a `.pdf`. Therefore, #4317 really messed up my workflow, since in `beamer`, a main switch is the `\mode` command. You switch between slides / notes modes by using `\mode<presentation>{}`, `\mode<article>{}`, etc... [For example, `\mode<presentation>{...}` specifies that the content within the curly braces should only be shown in presentation mode](https://tex.stackexchange.com/questions/263440/what-are-the-available-modes-in-beamer). #4317 causes my command `\mode<presentation>` to be slapped onto a spurious slide.
Specifically, the following code in an `.Rmd` file:
````
# Data: plots and summaries {#data}
```{r Lecture1Front, eval = (out_type=="beamer"), echo=FALSE, results='asis'}
```
````

and 

```
## ---- Lecture1Front ----
cat(
  "\n\n\\mode<presentation>{ \n",
  "\\title[\\CourseName{} Lecture 1]{\\CourseName{} \\newline \\, Lecture 1}\n",
  "\\date{\\DayOne{}\\Yr{}}}\n\n",
  "###\n\n",
  "\\titlepage\n\n",
  sep = "")
```
in the matching `.R` file produced 
```
\hypertarget{data}{%
\section{Data: plots and summaries}\label{data}}
\hypertarget{data-1}{%
\subsection{Data}\label{data-1}}
\mode<presentation> 
{ 
\title[\CourseName{} Lecture 1]{\CourseName{} \newline \, Lecture 1}
\date{\DayOne{}\Yr{}}
}
\begin{frame}{}
\protect\hypertarget{section}{}
\titlepage
\end{frame}
```
before `Pandoc-2.7`, but 
```
\section{Data: plots and summaries}\label{data}}
\begin{frame}{Data: plots and summaries}
\mode<presentation>{ 
\title[\CourseName{} Lecture 1]{\CourseName{} \newline \, Lecture 1}
\date{\DayOne{}\Yr{}}}
\end{frame}

```
starting with `pandoc-2.7` "thanks" to #4317. We're not just getting a meaningless section title slide, the `LaTeX` code is no longer being executed and refreshed, so that the added frame is stale and showing the first title slide over and over.

jgm:
yes, I understand -- you want to inject latex code outside of a frame environment.  I still don't have much to add to what I said in #4317 comments...

PR #8995, opened 2023-08-12
Add support for verse lang with --listings

nutti:
Verse lang is a new language introduced in Unreal Engine for Fortnite.
https://dev.epicgames.com/documentation/en-us/uefn/learn-programming-with-verse-in-unreal-editor-for-fortnite

This PR adds a verse lang support of listing.

### DIFF
--- a/src/Text/Pandoc/Highlighting.hs
+++ b/src/Text/Pandoc/Highlighting.hs
@@ -215,6 +215,7 @@ langsList =
   ("tex","TeX"),
   ("latex","TeX"),
   ("vbscript","VBScript"),
+  ("verse","Verse"),
   ("verilog","Verilog"),
   ("vhdl","VHDL"),
   ("vrml","VRML"),


jgm:
Is it actually supported by the LaTeX listings package?

chdemko:
Currently: verse is not supported by the listings package (http://mirrors.ctan.org/macros/latex/contrib/listings/listings.pdf#table.1)

nutti:
Actually not yet.
Should we support by the LaTeX listing package at first?

jgm:
The intent of the piece of code you're changing here is to filter out languages actually supported by listings (and change their names when listings has a distinct name).  We don't want to add `language=foo` if `foo` is not a language supported by listings (or not a language at all).  So, the omission of languages not currently supported by listings (such as `verse`) is by design.

DeanGadberry:
Please close this PR

nutti:
Sorry, I forgot to reply this PR.
Yes, I think this PR should be closed until tex itself supports Verse language.

Issue #8979, opened 2023-08-01
`.odt` Reader does not handle attribute `text:continue-numbering` on lists.


StephanMeijer:
# Explain the problem.

We have identified an issue where the conversion process from `.odt` to `.html` using both Microsoft Word and Google Docs does not correctly preserve the `start` attribute on ordered lists. This problem seems to be specific to the conversion from `.odt`, as the same conversion from `.docx` to `.html` correctly maintains the `start`-attribute when performed on identical documents.

## Expected Result:

The `start`-attribute on ordered lists should be preserved during the conversion from `.odt` to `.html`, similar to how it is preserved in the `.docx` to `.html` conversion.

## Actual Result:

The `start`-attribute on ordered lists is not preserved when converting from `.odt` to `.html`.

# Pandoc version

```
pandoc 3.1.6
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: /Users/steve/.local/share/pandoc
Copyright (C) 2006-2023 John MacFarlane. Web: https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.
```

# Examples

<table>
<thead>
<tr>
<th>Testcase</th>
<th>Folder including input and output for <code>.docx</code></th>
<th>Output from <code>.docx</code></th>
<th>Folder including input and output for <code>.odt</code></th>
<th>Output from <code>.odt</code></th>
</tr>
</thead>
<tbody>

<tr>
<td>
Google Docs
</td>
<td>
<a href="https://gitlab.com/toegang-voor-iedereen/pandoc-filters/-/tree/master/test/NLDOC-765-google-docs-docx-lists-several-types?ref_type=heads">Here.</a>
</td>
<td>
<a href="https://gitlab.com/toegang-voor-iedereen/pandoc-filters/-/blob/master/test/NLDOC-765-google-docs-docx-lists-several-types/none/expected.html?ref_type=heads">Here.</a>
</td>
<td>
<a href="https://gitlab.com/toegang-voor-iedereen/pandoc-filters/-/tree/master/test/NLDOC-765-google-docs-odt-lists-several-types?ref_type=heads">Here.</a>
</td>
<td>
<a href="https://gitlab.com/toegang-voor-iedereen/pandoc-filters/-/blob/master/test/NLDOC-765-google-docs-odt-lists-several-types/none/expected.html?ref_type=heads">Here.</a>
</td>
</tr>

<tr>
<td>
[...]

PR #8998, opened 2023-08-13
#8979 Support attr `text:continue-numbering` on ODT Reader

StephanMeijer:
**Please keep the following in mind:** I am a beginner in Haskell. So in case I wrote garbage, please don't judge me too harshly. Please also check **thoroughly** before merging. Thanks!

See #8979 

---

- [x] Update test `.odt` to be from LibreOffice not Google Docs, so `.native`-output isn't cluttered with `BlockQuote`.
- [x] Fix typo @ZwaarContrast caught.


src/Text/Pandoc/Readers/ODT/ContentReader.hs
```diff
@@ -455,6 +474,12 @@ getListConstructor ListLevelStyle{..} =
     toListNumberDelim (Just "(") (Just ")") = TwoParens
     toListNumberDelim     _          _      = DefaultDelim
 
+--
+startingNumber :: Bool  -> Int -> Maybe ListLevelStyle -> Int
+startingNumber continueNumbering prevStartCounter mListLevelStyle
+    | continueNumbering                = prevStartCounter + 1
```
jgm, line 480: Suppose we have a list
```
1. a
2. b
```
and then another list with "continue numbering."  The new list should begin with 3, but I think your code will make it begin with 2 (previous list's start number + 1).

StephanMeijer: > The new list should begin with 3, but I think your code will make it begin with 2 (previous list's start number + 1).`

No it won't.

```hs
  where
    constructOrderedList startNum listLevel listItemCount =
          reader
      >>> modifyExtraState (shiftListLevel (-1))
      >>> modifyExtraState (modifyPreviousListStartCounter listLevel (startNum + listItemCount - 1))
      >>> arr (orderedListWith (startNum, DefaultStyle, DefaultDelim))
    constructListWith listLevelStyle startNum listLevel listItemCount =
          reader
      >>> getListConstructor listLevelStyle startNum
      ^>> modifyExtraState (shiftListLevel (-1))
      >>> modifyExtraState (modifyPreviousListStartCounter listLevel (startNum + listItemCount - 1))
```

The previousListStartCounter for that `listLevel` will be seto to `(startNum + listItemCount - 1)`. 

I need to give it better naming, as it is basically the last number on the list.

e.g. with

```md
1. a
2. b
3. c
```

the value will be set to 3.

I can probably refactor it to the value being set to `4`, renaming the counter to `listContinuationStartCounter` or something.

StephanMeijer:
@jgm I refactored the code to make the functionality more clear.

src/Text/Pandoc/Readers/ODT/ContentReader.hs
```diff
@@ -75,6 +75,10 @@ data ReaderState
                  , styleTrace       :: [Style]
                    -- | Keeps track of the current depth in nested lists
                  , currentListLevel :: ListLevel
+                   -- | Keeps track of the previous list start counters,
+                   -- so whenever a new list want to continue numbering,
```
ZwaarContrast, line 79: typo, "wants to continue"

StephanMeijer:
@jgm Updated PR with cleaner example for test. Would you be willing to review once again? 🙏

src/Text/Pandoc/Readers/ODT/ContentReader.hs
```diff
@@ -75,6 +75,10 @@ data ReaderState
                  , styleTrace       :: [Style]
                    -- | Keeps track of the current depth in nested lists
                  , currentListLevel :: ListLevel
+                   -- | Keeps track of the previous list start counters,
+                   -- so whenever a new list wants to continue numbering,
+                   -- we know what number to start from.
+                 , listContinuationStartCounters :: M.Map ListLevel Int
```
jgm, line 81: I think this comment may be a bit misleading ; isn't this keeping track of what the *next* start number should be if another list in the same level is marked as continuing?

test/odt/native/listContinueNumbering.native
```diff
@@ -0,0 +1,89 @@
+[ OrderedList
```
jgm, line 1: I may have missed it, but do you have a test where the first list has one item, containing a sublist with two items, and then a second list continues the first list? That kind of case would be good to test explicitly.

jgm (changes requested):
This looks good; I've just made a couple small suggestions.

src/Text/Pandoc/Readers/ODT/ContentReader.hs
StephanMeijer: Will improve.

test/odt/native/listContinueNumbering.native
StephanMeijer: Will create additional testcase.

StephanMeijer:
@jgm Feedback taken into account and changed things appropriately. I didn't test locally so I will test on CI/CD.

PR #8999, opened 2023-08-13
Change semicolon to colon in inline CSS in HTML template

jtschuster:
Converting to standalone html uses a template with a type with a semicolon where there should be a colon in the inline CSS.


jgm:
Argh!  Thanks for noticing that.
Tests need to be updated but I can do that for you.

PR #9004, opened 2023-08-15
Formats section of README shifted to dedicated file, along lines of INSTALL.md

arighorbani:
Moved the information for input and output formats to a dedicated template-markdown set. Makes the README more friendly to new users, in the same fashion that the installation details are cordoned off into their own markdown file, and simplifies the process of changing the documentation to account for more features slightly.


jgm:
No, thanks! I prefer it the way it is.

Issue #9025, opened 2023-08-26
comma not escaped in Org reader macro


natto1784:
Org reader currently does not escape comma (and backslashes) when parsing macro invocations.
See footnote: https://orgmode.org/manual/Macro-Replacement.html#FOOT127

#### Minimal example
```org
#+MACRO: HELLO $1
{{{HELLO(foo\,bar)}}}
```
#### Output
```html
<p>foo\</p>
```
#### Expected
```html
<p>foo,bar</p>
```

PR #9026, opened 2023-08-26
Org reader: allow escaping commas in macro arguments

natto1784:
Fixes https://github.com/jgm/pandoc/issues/9025


jgm:
looks good to me, thanks!

PR #9029, opened 2023-08-26
Correct name of JATS element <attrib>

kamoe:
Replacement of wrong element name "attribution" by the correct one "attrib"


jgm:
thanks!

PR #9044, opened 2023-09-01
HTML writer: use the ID prefix in the ID for the footnotes section.

bdesham:
In general, the ID prefix makes it possible to combine multiple pieces of Pandoc-generated HTML with no possibility that IDs will conflict. One exception to this was that the footnotes were always put into an element like

    <aside id="foonotes" ...>

This commit applies the ID prefix to this ID as well.

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -299,7 +299,7 @@ pandocToHtml opts (Pandoc meta blocks) = do
     if null (stNotes st)
       then return mempty
       else do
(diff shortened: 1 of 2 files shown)


bdesham:
This is in the same vein as #8770, but that PR focuses on the case where there are multiple footnotes sections within the same document. This one focuses on the case where there is (probably) only one footnotes section, but the user expressed a desire (through `--id-prefix` or `writerIdentifierPrefix`) that all `id` values use a certain prefix.

jgm:
looks good, thanks!

Issue #9032, opened 2023-08-28
Markdown size attributes: `{style="width:2.0in;... ` instead of `{width="2.0in"`


andrzejQ:
**Markdown size attributes: `{style="width:2.0in;... ` instead of `{width="2.0in"`**

Pandoc is a great program - especially for fans of simple files with markup format (which is me).

I often use the markdown format, e.g. in a jekyll blog or with a library [Markdig](https://github.com/lunet-io/markdig) (MarkdownPanel for Notepad++).

The curly brackets `{}` in a markdown expression `![](){}` or `[](){}` are generally interpreted as a set of attributes. The proposal as above would lead to even greater universality of "*.md" files produced by pandoc.

For example, after creating the file "width in inch.docx" with the command
`pandoc "tst0.md" -o "width in inch.docx"`

where "tst0.md":
````md
Pandoc
======

![CommonMark ico](https://avatars.githubusercontent.com/u/8661225?v=4)
````
And re-convert to *.md:  
`pandoc "width in inch.docx" -o "width in inch.docx.md" --extract-media=./img/`

we get the file "width in inch.docx.md":

````md
# Pandoc

![CommonMark ico](./img/media/rId20.png){width="2.0in" height="2.0in"}

CommonMark ico
````

If this ".md" file was converted using pandoc to ".html"  
`pandoc "width in inch.docx.md" -o "width in inch.html`  
then we get valid HTML with image attributes:
````html
<img src="./img/media/rId20.png" style="width:2in;height:2in" alt="CommonMark ico" />
````

But in other libraries image attributes are often explicitly rewritten, e.g. Markdig:
````html
<img src="./img/media/rId20.png" width="2.0in" height="2.0in" alt="CommonMark ico" />
````

And in this context, width and height can only be the number of pixels. Browsers simply ignore everything except the integer part, i.e. in this example, they display an image in the size of 2px x 2px.

If pandoc generated attributes ready for insertion into HTML without conversion, it would extend the universality of the markdown file, i.e. in the 3rd line "width in inch.docx.md" it could be:

````md
![CommonMark ico](./img/media/rId20.png){style="width:2.0in;height:2.0in"}
````

jgm:
I don't want to do the conversion to a `style` attribute at the AST level, because that's HTML specific. This belongs in the HTML writer (where it is).
The problem you describe sounds like a problem with Markdig's HTML generation and could be reported there.

andrzejQ:
[...]

Issue #9053, opened 2023-09-03
pandoc 3.1.7 with `citeproc` clobbers beamer lists


jpcirrus:
In pandoc 3.1.7 with the `citeproc` filter and output to beamer, because the `CSLReferences` environment is now provided as a new itemize list using the `enumitem` package in the resolution of #9031 by 353177f, the beamer theme lists are clobbered due to issues using  `enumitem` with beamer: https://tex.stackexchange.com/search?q=beamer+enumitem

This can be seen with this input file (`csl-beamer.md`):

````markdown
---
aspectratio: 169
references:
- id: moore2000
  type: article-journal
  title: Diabetes in New Zealand
  container-title: Diabetes
  page: 1-2
  volume: 50
  issue: 2
  author:
    - family: Moore
      given: M.Peter
    - family: Lunt
      given: Helen
  issued:
    date-parts:
      - - 2000
- id: powers2015
  type: book
  title: Exercise physiology
  publisher: McGraw-Hill
  author:
    - family: Powers
      given: Scott K.
    - family: Howley
      given: Edward T.
  issued:
    date-parts:
      - - 2015
  publisher-place: 'New York, NY'
nocite: |
  @*
header-includes: |
  ```{=latex}
  \setlength{\cslentryspacing}{0ex}  % Use vertical not horizontal length
  \AtBeginEnvironment{CSLReferences}{% Format bibliography
    \bigskip \scriptsize \setlength{\parskip}{0ex}}
  ```
---

## Itemise {.t}

- apple
- banana
- cherry

## Bibliography {.t}
````

and compiling with:

```bash
pandoc -t beamer -o csl-beamer.pdf csl-beamer.md
```

gives:

<img width="827" alt="image" src="https://github.com/jgm/pandoc/assets/39587/709f5cc0-32bd-4a5e-89e3-5aa467b57459">

but, when adding `--citeproc` the labels (and more?) get clobbered:

```bash
pandoc -t beamer -o csl-beamer.pdf --citeproc csl-beamer.md
```

<img width="828" alt="image" src="https://github.com/jgm/pandoc/assets/39587/855d93fe-bd9b-4a76-816b-ed72a86fe667">

<img width="827" alt="image" src="https://github.com/jgm/pandoc/assets/39587/52cfe146-9860-4a8a-b3c1-2224a84ee07f">

jgm:
[...]

Issue #7296, opened 2021-05-17
CSL entry-spacing="0" not removing vertical space


rnhmjoj:
At least with the Latex writer, setting the bibliography property `entry-spacing="0"` in the CSL doesn't completely removes the vertical space between references.

Example document to reproduce the problem:

```
---
title: Issue \#335 reproducer
citation-style: https://www.zotero.org/styles/elsevier-with-titles
reference-section-title: References
references:
  - id: trapani15
    type: article-journal
    author:
      family: Trapani
      given: Lorenzo
    title: Testing for (in)finite moments
    container-title: Journal of Econometrics
    issued:
      - year: 2015
  - id: silver86
    type: book
    author:
      family: Silverman
      given: Bernard W.
    title: Density Estimation for Statistics and Data Analysis
    issued:
      - year: 1986
  - id: robertson74
    type: article-journal
    author:
    - family: Robertson
      given: Tim
    - family: Cryer
      given: Jonathan D
    issued:
      - year: 1974
    title: An iterative procedure for estimating the mode
    container-title: Journal of the American Statistical Association
    volume: 69
    issue: 348
...

# Testing

This is a test[@trapani15] for CSL[@silver86] display options[@robertson74].

# Compiling

Run this to compile: `pandoc --citeproc test.md -o test.pdf`
```

jgm:
This change to the default template seems to help:
```
\newlength{\cslentryspacing}
\newenvironment{CSLReferences}[2] % #1 hanging-ident, #2 entry spacing
 {% don't indent paragraphs
  \setlength{\parindent}{0pt}
  % set entry spacing
  \setlength{\cslentryspacing}{#2\baselineskip - \parskip}
  % turn on hanging indent if param 1 is 1
  \let\oldpar\par
  \ifodd #1
  \def\par{\hangindent=\cslhangindent\oldpar\vspace{\cslentryspacing}}
  \else
  \def\par{\hangindent=\cslhangindent\oldpar\vspace{\cslentryspacing}}
  \fi
 }%
 {}
```
I don't know if that's the best approach.

jgm:
Note: if you want a closely spaced bibliography, you probably also don't want gaps between paragraphs.  With released pandoc and default template, try with
```
pandoc test.md --citeproc -o test.pdf -Vindent=false
```
This will remove the paragraph space (which was causing the extra spacing in the bibliography) throughout the document, and your bibliography will look the way you want it.

jgm:
[...]

PR #9058, opened 2023-09-04
LaTeX template: fix `\CSLRightInline` vertical space

jpcirrus:
When using a CSL stylesheet with the `citation-number` attribute e.g. `nature`, when `\CSLRightInline` is only one line currently extraneous vertical space is added.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -362,7 +362,7 @@ $if(csl-refs)$
 \usepackage{calc}
 \newcommand{\CSLBlock}[1]{#1\hfill\break}
 \newcommand{\CSLLeftMargin}[1]{\parbox[t]{\csllabelwidth}{#1}}
-\newcommand{\CSLRightInline}[1]{\parbox[t]{\linewidth - \csllabelwidth}{#1}\break}
+\newcommand{\CSLRightInline}[1]{\parbox[t]{\linewidth - \csllabelwidth}{#1\\[-1.7ex]}}
 \newcommand{\CSLIndent}[1]{\hspace{\cslhangindent}#1}
 $endif$
 $if(lang)$


jpcirrus:
Currently, with `indent: false`, when `\CSLRightInline` is only one line, then the output has extraneous vertical space below that bibitem:

image

this change matches the document paragraph spacing:

image

and currently with `indent: true`:

20230905T101927-PDF Expert@2x

which this change corrects to:

20230905T102024-PDF Expert@2x

I have tested with the standard classes as well as with the KOMA-Script and APA classes.

data/templates/default.latex
```diff
@@ -362,7 +362,7 @@ $if(csl-refs)$
 \usepackage{calc}
 \newcommand{\CSLBlock}[1]{#1\hfill\break}
 \newcommand{\CSLLeftMargin}[1]{\parbox[t]{\csllabelwidth}{#1}}
-\newcommand{\CSLRightInline}[1]{\parbox[t]{\linewidth - \csllabelwidth}{#1}\break}
+\newcommand{\CSLRightInline}[1]{\parbox[t]{\linewidth - \csllabelwidth}{#1\\[-1.7ex]}}
```
jgm, line 365: Can you explain why this change fixes the problem?

jpcirrus: To determine where the additional vertical space was coming from I wrapped the `\parbox{}` in an `\fbox{}`. This indicated that it was from the `\break` outside the `\parbox`. By moving that inside the `\parbox` results were consistent regardless of bibitem length, but without negative adjustment were too large. By iteration arrived at the value of `-1.7ex`. I had previously been using a custom template with `\break` replaced by `\\[0.7ex]` to get spacing that matched the document. Technically why, I don't know.

jgm: It just seems a bit "magic" and unprincipled...I wonder if there's a better way?

jgm: What about just removing the `\break`?  Is it needed?
It seems to work fine without it and I don't get the extra space.

jpcirrus: During my testing I had initially tried that, but when using a more complex document using `article` and `scrartcl` the linespacing was removed between bibitems:

---

20230905T141943-PDF Expert@2x

---   

20230905T141542-PDF Expert@2x

---

but having used the above input file with `article`, `scrartcl`, `apa6` (with the `man` class option) and `beamer` it works correctly, so there must be something (not obvious) in that document which is causing the issue which I will sort out.

jpcirrus: Right, the problem I had experienced when only removing `\break` was caused by me using a `bibliography` with multiple references for testing. In that case the spacing between bibitems is only maintained for items that fit on one line. If in the MWE above you change the order of the two references you will see the linespacing is gobbled between the two items. The opposite of the original issue.

jgm: Can you upload a test case? I tried to produce one myself but didn't see the problem.
Screen Shot 2023-09-04 at 10 05 32 PM

jpcirrus:
````markdown
---
fontsize: 12pt
indent: true
csl: nature
references:
- id: francois2014
  type: article-journal
  title: >-
    Lorem ipsum dolor sit amet, consectetur ullamco laboris nisi ut aliquip ex ea commodo
  container-title: Diabetologia
  author:
    - family: Williams
      given: Michael J. A.
    - family: Francois
      given: Monique E.
    - family: Baldi
      given: James C.
    - family: Manning
      given: Patrick J.
    - family: Lucas
      given: Samuel J. E.
    - family: Hawley
      given: John A.
    - family: Cotter
      given: James D.
  issued:
    date-parts:
      - - 2014
- id: WatsonCrick1953
  type: article-journal
  author:
  - family: Watson
    given: J. D.
  issued:
    date-parts:
    - - 1953
  title: Molecular aliquip ex ea commodo
  container-title: Nature
- id: zfrancois2014
  type: article-journal
  title: >-
    Lorem ipsum dolor sit amet, consectetur ullamco laboris nisi ut aliquip ex ea commodo
  container-title: Diabetologia
  author:
    - family: Zilliams
      given: Michael J. A.
    - family: Francois
      given: Monique E.
    - family: Baldi
      given: James C.
    - family: Manning
      given: Patrick J.
    - family: Lucas
      given: Samuel J. E.
    - family: Hawley
      given: John A.
    - family: Cotter
      given: James D.
  issued:
    date-parts:
      - - 2014
- id: zatsonCrick1953
  type: article-journal
  author:
  - family: Zatson
    given: J. D.
  issued:
    date-parts:
    - - 1953
  title: Molecular aliquip ex ea commodo
  container-title: Nature
nocite: |
  @*
...

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
ullamco laboris nisi ut aliquip ex ea commodo.

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
ullamco laboris nisi ut aliquip ex ea commodo.
````

I have increased `fontsize` to 12pt to make the difference more apparent. The difference is the same regardless of the value of `indent`, but easier to discern when `true`. It is subtle, and the font makes a difference to perception (I was using Minion Pro earlier, as that's what I use for reports, and probably because it's more condensed, makes it easier to notice). I have been taking a screenshot of the pdf output, reducing the transparency to 50% and superimposing the body text over the bibliography. The vertical space is correct after the single line bibitem, but reduced when not. I compile in a separate step with `latexmk`,  which is configured to use `xelatex`.

Comment (jgm) [reply data/templates/default.latex:365]: OK, I can see it now (with pdflatex too).  So, in this example the vertical space between items 2 and 3 is what we want between all items?

Comment (jgm) [reply data/templates/default.latex:365]: I notice that changing the setting of `\cslentryspacing` makes no difference whatsoever to the output.

Comment (jgm) [reply data/templates/default.latex:365]: A bit ugly though because of the slightly larger line skip between internal lines and between entries.

Comment (jpcirrus) [reply data/templates/default.latex:365]: It isn't. My results, which match exactly, before I go to sleep:

```markdown
indent: false: entry-spacing="0": \setlength{\cslentryspacing}{\lineskip - \parskip}
indent: false: entry-spacing="1": \setlength{\cslentryspacing}{\lineskip}
indent: false: entry-spacing="2": \setlength{\cslentryspacing}{\lineskip + 1\baselineskip}
indent: false: entry-spacing="3": \setlength{\cslentryspacing}{\lineskip + 2\baselineskip}
indent: true: entry-spacing="0":  \setlength{\cslentryspacing}{\lineskip}
TODO: indent: true: entry-spacing="1":
TODO: indent: true: entry-spacing="2":
TODO: indent: true: entry-spacing="3":
```

I see you have closed #9032. These results are not based on that. Will look at that tomorrow, had a busy day.

jgm: Yes, I think #9053 changes may make a difference here - I haven't had a chance to test either.

jpcirrus: I have tested your changes with the pandoc nightly, and with the changes from b7e1ce4 to `default.latex`. Everything is working correctly (in my opinion) when the following changes are implemented:

1.   When `entry-spacing="0"`, the correct spacing is obtained with ` 1\baselineskip - font size`. In latex the font size is held in `\f@size`. The line `\setlength{\itemsep}{#2\baselineskip}}}` accordingly needs to be able to cater for this special case. This is to comply with the [CSL specification](https://docs.citationstyles.org/en/stable/specification.html#whitespace) for `entry-spacing`:

      > Specifies vertical distance between bibliographic entries. By default (with a value of “1”), entries are separated by a single additional line-height (as set by the line-spacing attribute). Can be set to any non-negative integer to specify a multiple of this amount.

      which suggests that `0` means no *additional* spacing, not zero spacing. Also, as stated, the default is `1`, not `0`, which is currently set in `src/Text/Pandoc/Writers/LaTeX.hs:349` by `Nothing -> "0"` (can be tested with `vancouver.csl`).

2.   When using a CSL style sheet with `display="block"`, the current configuration for `CSLBlock` adds extra spacing when the block item, say `note`, is the last reference item, as in `apa-annotated-bibliography.csl`. But, if this change is implemented:

      ```patch
      - \newcommand{\CSLBlock}[1]{\hfill\break#1\hfill\break}
      + \newcommand{\CSLBlock}[1]{\hfill\break#1}
      ```

      then if the block item isn't the last item, it's not possible to compile. I'm not sure of a general solution which covers both cases. For my personal use I will use a custom template with this change.

9.   Your suggestion, which fixes the title of this PR:

   ```patch
   - \newcommand{\CSLRightInline}[1]{\parbox[t]{\linewidth - \csllabelwidth}{#1}\break}
   + \newcommand{\CSLRightInline}[1]{\parbox[t]{\linewidth - \csllabelwidth}{#1}}
   ```

Does pandoc need to consider adding `line-spacing`, or is this just assumed to match document line spacing?

jgm: Thanks for this. I'll make these changes (except for removing the final break from CSLBlock).

I'm wondering whether, when entry spacing is set to 0, we should just use `\parskip` instead of actually using 0?  In a document that has a `\parskip` of 6pt (default for pandoc), the bibliography may look odd with 0 space between entries.  But maybe that's too opinionated a divergence.

jgm: Currently we're ignoring `line-spacing` (though citeproc does record this).  It seems to me that nobody is really going to want a line spacing in the bibliography that differs from that of the document itself.  If people say otherwise, we could consider it.

jgm: I've pushed some more changes; further testing welcome!

jpcirrus: In `default.latex`, fix typo, and use CSL name for both arguments:

```patch
- \newenvironment{CSLReferences}[2] % #1 hanging-ident, #2 entry spacing
+ \newenvironment{CSLReferences}[2] % #1 hanging-indent, #2 entry-spacing
```

jgm: Thanks, fixed.

jpcirrus: In `default.latex` fix rare issue that causes an extra line to be a added inside a bibitem's parbox. Occurs in the situation where the last line of a bibitem extends up to the right margin, but not quite far enough to wrap.

20230908T092159-PDF Expert@2x

I initially never had the spaces, but added them after reading on TeX Stack Exchange (David Carlisle) that without them wrapping could be affected if the parboxes contained hard line breaks, so thought better to cater for possible future changes:(

```patch
- \newcommand{\CSLLeftMargin}[1]{\parbox[t]{\csllabelwidth}{\strut #1 \strut}}
- \newcommand{\CSLRightInline}[1]{\parbox[t]{\linewidth - \csllabelwidth}{\strut #1 \strut}}
+ \newcommand{\CSLLeftMargin}[1]{\parbox[t]{\csllabelwidth}{\strut#1\strut}}
+ \newcommand{\CSLRightInline}[1]{\parbox[t]{\linewidth - \csllabelwidth}{\strut#1\strut}}
```

jgm: OK, done.

jpcirrus: > I'm wondering whether, when entry spacing is set to 0, we should just use `\parskip` instead of actually using 0? In a document that has a `\parskip` of 6pt (default for pandoc), the bibliography may look odd with 0 space between entries. But maybe that's too opinionated a divergence.

No. I think it would be incorrect in terms of the CSL spec. I believe the current implementation is accurate.

With `entry-spacing="0"` and `\setlength{\itemsep}{#2\baselineskip}}}`:

20230908T103846-PDF Expert@2x

and `entry-spacing="0"` and `\setlength{\itemsep}{\parskip}}}`:

20230908T103936-PDF Expert@2x

The spacing is approaching what it is when `entry-spacing="1"`. The paragraph spacing has nothing to do with the entry spacing, which had me confused until I read the spec more carefully

jgm: Yes, I agree it's more accurate this way.  I just think it's a bit awkward that, with the default 6pt parskip that pandoc gives you without `indent`, you have a choice between a bigger or a smaller space between bib entries, but no way to get the same...

jpcirrus: I actually prefer the way it is :smiley:. 

20230908T160345-PDF Expert@2x

This is going to be the  case with all documents where `entry-spacing="0"` and `indent: false`, regardless if using a KOMA class, the `parskip` package, or the pandoc default of `6pt plus 2pt minus 1pt`. In my custom template I used to have the `entry-spacing` depend on `indent`. But, I was wrong, due to my ignorance from not having read the spec, assuming 0 meant zero spacing (not zero *additional* spacing), 1 was line spacing (actually 0)... and which clearly others do too e.g. #7296. But, I can appreciate the visual appeal:

20230908T160629-PDF Expert@2x

rather than moving up to `entry-spacing="1"`:

20230908T160659-PDF Expert@2x

However, if you decide to implement this please don't make it the default, but rather an option, or better yet as a filter, so that at least the default is compliant with the spec. Of course, for those wanting something special, it is always possible to use `header-includes`:

```latex
\usepackage{etoolbox}
\AtBeginEnvironment{CSLReferences}[\setlength{\itemsep}{ANYTHING YOU WISH}
```

I have completed testing the latex CSL functionality. I have tested with the KOMA, APA, and standard classes, together with beamer. All combinations of linespacing (single, onehalf and double) work correctly with `entry-spacing` values of 0, 1 and 2, using bibliographies of approx 50–200 items. Thanks very much for your efforts with this, I am really happy with the results.

jgm: Thanks very much for the extensive testing!  I'll close this.

Issue #9047, opened 2023-09-01
HTML reader and HTML writer out of sync on tasks


RobTrew:
**Problem**

In the case of task list items, the HTML reader fails to  parse the output of the HTML writer.

`- [ ]` and `- [x]` prefixes are read from `markdown` to native:
- `[Plain [Str "\9744", Space ...`
- `[Plain [Str "\9746", Space ...` 

respectively.

The HTML *writer* renders these successfully in the pattern:

```html
<li><label><input type="checkbox" />an unchecked item</label></li>
<li><label><input type="checkbox" checked="" />a checked
```

but the HTML *reader* fails to recognise that `<label>` pattern, and loses the task status, interpreting such lines as plain unadorned `<li>`.

For an `HTML -> Native -> Markdown` return to `- [ ]` and `- [x]` the HTML reader only recognises direct use of the Unicode _Ballot Box_ and _Ballot Box with X_ pattern:

```html
<li>&#9744; an unchecked item</li>
<li>&#9746; a checked item</li>
```

It looks as if the HTML reader has not yet caught up with the HTML writer, and doesn't yet understand its task item output.

**Pandoc version**
Pandoc 3.1.6.2
macOS 13.5.1

jgm:
Yes, there were changes in the writer's output, and the reader should be updated to interpret this.

PR #9066, opened 2023-09-07
HTML reader can parse checkboxes

sspeaks:
Fixed an issue specified by bug found at https://github.com/jgm/pandoc/issues/9047

I am very new to haskell so there are probably many better choices I could have made. Though I'm not asking for guidance, it'd be much appreciated!


src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -342,7 +342,20 @@ pListItem = do
...
+    pLabel' = try $ pInTag TagsRequired "label" pInput'
+    pInput' :: PandocMonad m => TagParser m Blocks
+    pInput' = do
+          TagOpen _ attr' <- lookAhead $ pSatisfy $ matchTagOpen "input" [("type","checkbox")]
+          let attr = toStringAttr attr'
+          let isChecked = isJust $ lookup "checked" attr
+          let escapeSequence = B.Str $ if isChecked then "\9746" else "\9744"
+          _ <- pSelfClosing (== "input") (const True)
+          bs <- inline
+          return $ B.Many . Seq.singleton $ B.Plain ([escapeSequence, B.Space ] <> B.toList bs)
```
jgm, line 357: Here you can just use the function `plain` from Text.Pandoc.Builder to avoid these conversions:
```hs
return $ B.plain (escapeSequence <> B.space) <> bs
```
You'll also need to use B.str for escapeSequence:
```hs
let escapeSequence = B.str $ if isChecked ...etc.
```
Basically `B.str` = `B.Many . Seq.singleton . B.Str`.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -342,7 +342,20 @@ pListItem = do
...
+    pLabel' :: PandocMonad m => TagParser m Blocks
+    pLabel' = try $ pInTag TagsRequired "label" pInput'
+    pInput' :: PandocMonad m => TagParser m Blocks
+    pInput' = do
+          TagOpen _ attr' <- lookAhead $ pSatisfy $ matchTagOpen "input" [("type","checkbox")]
+          let attr = toStringAttr attr'
+          let isChecked = isJust $ lookup "checked" attr
+          let escapeSequence = B.Str $ if isChecked then "\9746" else "\9744"
+          _ <- pSelfClosing (== "input") (const True)
+          bs <- inline
```
jgm, line 356: Typically we use `bs` for blocks and `ils` for inlines.  Also, you'll want to parse many inlines, not just one.  (I think there's a parser `inlines`? Or maybe you just use many and concat -- check for other examples in the code.)

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -342,7 +342,20 @@ pListItem = do
...
                            _ -> B.divWith (ident, [], []) bs
   maybe id addId (lookup "id" attr) <$>
-    pInTags "li" block
+    pInTags "li" (pLabel' <|> block)
+  where 
+    pLabel' :: PandocMonad m => TagParser m Blocks
+    pLabel' = try $ pInTag TagsRequired "label" pInput'
+    pInput' :: PandocMonad m => TagParser m Blocks
+    pInput' = do
+          TagOpen _ attr' <- lookAhead $ pSatisfy $ matchTagOpen "input" [("type","checkbox")]
```
jgm, line 351: I wasn't sure why you wanted `lookAhead` here?

sspeaks: I wasn't sure if I needed to parse the closing tag too using the approach I took, so I just looked ahead and then parsed the whole thing later.

But it does make much better sense to consume the opening tag as I read it.

I'll change it, thanks!

src/Text/Pandoc/Readers/HTML.hs
sspeaks: Sure! Just cleaned the code up and it looks a lot better!

src/Text/Pandoc/Readers/HTML.hs
sspeaks: Makes sense! Not sure the `ils` bit is relevant anymore with my newest changes but should be parsing multiple inlines now

sspeaks:
One thing I'm curious about is how we should handle scenarios where the checkbox follows the label.

From a markdown perspective, that never happens and doesn't really make sense, so the current behaviour just ignores trailing checkboxes.

I added a unit test to check for this scenario.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -341,20 +341,20 @@ pListItem = do
...
-          let escapeSequence = B.Str $ if isChecked then "\9746" else "\9744"
-          _ <- pSelfClosing (== "input") (const True)
-          bs <- inline
-          return $ B.Many . Seq.singleton $ B.Plain ([escapeSequence, B.Space ] <> B.toList bs)
+  maybe id addId (lookup "id" attr) <$> 
+    pInTags "li" (blockWithPotentialCheckbox)
+  where
+    blockWithPotentialCheckbox :: PandocMonad m => TagParser m Blocks
+    blockWithPotentialCheckbox = let p = (<>) <$> parseCheckbox <*> inline
+          in B.plain . trimInlines . mconcat <$> many1 (try p <|> inline)
```
jgm, line 349: Could this be simplified to `many1 (try parseCheckbox <|> inline)`?  (And remove `p`)
Note:  this (and the current code) does not require that the checkbox be at the beginning; is that okay? Probably?

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -341,20 +341,20 @@ pListItem = do
...
-    pInput' = do
-          TagOpen _ attr' <- lookAhead $ pSatisfy $ matchTagOpen "input" [("type","checkbox")]
-          let attr = toStringAttr attr'
-          let isChecked = isJust $ lookup "checked" attr
-          let escapeSequence = B.Str $ if isChecked then "\9746" else "\9744"
-          _ <- pSelfClosing (== "input") (const True)
-          bs <- inline
-          return $ B.Many . Seq.singleton $ B.Plain ([escapeSequence, B.Space ] <> B.toList bs)
+  maybe id addId (lookup "id" attr) <$> 
+    pInTags "li" (blockWithPotentialCheckbox)
```
jgm, line 345: Parens are not needed here around blockWithPotentialCheckbox

src/Text/Pandoc/Readers/HTML.hs
sspeaks: I don't think it can be simplified to that particular code because my current assumption is if we see a checkbox we want to see at least 1 inline after it.
`many1 (try parseCheckbox <|> inline)` would potentially parse only a checkbox

sspeaks:
My current issue with the tests is that the old code would parse a `block` within the `li` tags.

That included in nested lists or other blocks. 

This worked because the `block` function falls through to `sPlain` which then parses `many1 inline`.... very similar to what I'm doing.

I need to figure out how to still parse `ul` and similar blocks while also looking for checkboxes.. 

Solutions I can think of :
- Maybe I could modify the block function?
- Or I could explicitly add the blocks we'd might expect to see in a list item into my function.
- modify the `sPlain` function to account for potential checkBoxes?
- modify the 'inline' function to parse checkboxes.

I'm leaning toward the latter since then we can keep almost all of the semantics as they exist. It feels like the best option.

The only downside is that approach would allow checkboxes at the END of `Inlines` or checkboxes by themselves

src/Text/Pandoc/Readers/HTML.hs
jgm: Can't there be empty task list items?

jgm:
One very simple approach would be to leave the list parsing code alone, and just add an inline parser that reads an `<input type="checkbox"... />` and replaces it with the appropriate Unicode character.  Then everything would work, I think, without any further changes, and the code change would be quite simple.

Possible drawback: one couldn't include an `<input type="checkbox".../>` in arbitrary places and expect it to be parsed as raw HTML (with the `raw_html` extension set).

Possible solution: start the parser with `guardDisabled Ext_raw_html`.  Possible drawback: someone might want to enable this extension AND have task lists work.

Second possible solution: keep track of when you're in a list item (maybe via a new field in HTMLState) and have this parser check this.

src/Text/Pandoc/Readers/HTML.hs
sspeaks: Decided to remove this code in favor of the old approach w/ added inline parser

src/Text/Pandoc/Readers/HTML.hs
sspeaks: Decided to remove this code in favor of the old approach w/ added inline parser

sspeaks:
Hmm I'll look into the raw_html extension you mentioned.. Don't currently grok it tbh.

Just adding a parser to the `inline` function for checkboxes feels so much cleaner so that's what's currently pushed.

sspeaks:
I noticed a new issue when adding tests that I don't like..

The last test in 9047.md was originally

```
% pandoc -f html+raw_html -t html+raw_html
<ul>
<li><label><input type="checkbox" checked/>foo</label></li>
</ul>
^D
<ul class="task-list">
<li><label><input type="checkbox" checked="" />foo</label></li>
</ul>
```
(notice the added `<label>` in the input)

When using raw_html, the reader saw the `label` and then just skipped over the `input` since it was part of the label.

I don't know if this should be accounted for? I think it's probably fine but am curious of your opinion

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -334,16 +338,26 @@ pBulletList = try $ do
...
-pListItem = do
+pListItem = setInListItem $ do
   TagOpen _ attr' <- lookAhead $ pSatisfy (matchTagOpen "li" [])
   let attr = toStringAttr attr'
   let addId ident bs = case B.toList bs of
                            (Plain ils:xs) -> B.fromList (Plain
                                 [Span (ident, [], []) ils] : xs)
                            _ -> B.divWith (ident, [], []) bs
-  maybe id addId (lookup "id" attr) <$>
+  maybe id addId (lookup "id" attr) <$>
```
jgm, line 348: looks like an extra space was introduced here at the end of the line

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -673,6 +687,9 @@ inline = pTagText <|> do
         "var" -> pCodeWithClass "var" "variable"
         "span" -> pSpan
         "math" -> pMath False
+        "input" 
+          | Just x <- lookup "type" attr,
+            x == "checkbox" -> asks inListItem >>= guard >> pCheckbox
```
jgm, line 692: Will this work?
```hs
| Just "checkbox" <- lookup "type" attr
```
if so it is cleaner

test/command/9047.md
```diff
@@ -0,0 +1,47 @@
...
+<ul>
+<li><input type="checkbox" checked/>foo</li>
+</ul>
+^D
+<ul class="task-list">
+<li><label><input type="checkbox" checked="" />foo</label></li>
+</ul>
+```
+
+w/o rawHTML extension, a checkbox is properly ignored
```
jgm, line 41: to be clearer, "a checkbox outside of a list item"?

jgm (changes requested):
This looks good to me, I've made made a few very small comments/queries.

src/Text/Pandoc/Readers/HTML.hs
sspeaks: Hmmm that'd be neat, I'll check. I'm new to guards in case statements (didn't even know it could be done!)

jgm: or maybe just `| lookup "type" attr == Just "checkbox"`?

sspeaks:
Should be good to go now. My first guthub pr, so not sure how to indicate that.

jgm:
Thank you and congratulations!

PR #9072, opened 2023-09-10
LaTeX template: fix \CSLBlock vertical space

jpcirrus:
When using a CSL style which has an item with the `display="block"` attribute, the vertical space after the block depends on if the block item is the last item of the bibitem, or otherwise. This fixes that. Examples with the block item bordered by an `\fbox` for illustration:

---

20230910T172042-PDF Expert@2x

---

20230910T175215-PDF Expert@2x

---

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -364,7 +364,7 @@ $if(csl-refs)$
   \setlength{\itemsep}{#2\baselineskip}}}
  {\end{list}}
 \usepackage{calc}
-\newcommand{\CSLBlock}[1]{\hfill\break#1\hfill\break}
+\newcommand{\CSLBlock}[1]{\hfill\break\parbox[t]{\linewidth}{\strut\ignorespaces#1\strut}}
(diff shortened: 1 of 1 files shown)


jpcirrus:
An issue with this proposal, would be if the block item`s contents exceeded a page, as it wouldn't break.

jgm:
I don't see the problem. Can you tell me what I should be looking at?

jpcirrus:
With the current `\newcommand{\CSLBlock}[1]{\hfill\break#1\hfill\break}`, if a block item is not the last item of the bibitem the spacing is correct, but if it is the last item, the trailing `\hfill\break` gives an additional vertical space before the next bibitem, which doesn't match the value of `entry-spacing=`:

---

20230911T121052-PDF Expert@2x

jgm:
Thanks, that makes sense. I see the point about page breaks, but I'm guessing this would very rarely be needed, so on balance this is still probably the thing to do.

PR #9095, opened 2023-09-22
Added translation nb.yaml

sda030:
Added nb.yaml, which is a duplicate of no.yaml.
no=Norwegian major language consists in reality of two microlanguages nn=Nynorsk and nb=Bokmål. Downstream software (Quarto) requires consistency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,4 +68,3 @@
 .*sw?
 .DS_Store
 .stack-work
-.Rproj.user
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@
 .*sw?
 .DS_Store
 .stack-work
+.Rproj.user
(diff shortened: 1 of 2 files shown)


jgm:
Thanks. I'd prefer no unrelated changes (e.g. .gitignore) in the PR.

Also -- would it be better to make nb.yaml a symlink to no.yaml (or the reverse), rather than duplicating contents?

sda030:
Ah, sorry about .gitignore. Also, I did not know/think about the symlink possibility. Is hard or soft link preferred? Let me try once more please, I am a bit new to PRs and pandoc dev.

jgm:
I think that making it a symlink is the best approach.  I think git handles that fine as long as it's a relative link to a file in the same repository.

PR #9100, opened 2023-09-24
Clean up epub CSS

samuel-weinhardt:
This PR contains mostly cosmetic changes, namely:
1. Combine identical styles
2. Wrap to 80 characters
3. Use consistent whitespace for brackets and indentation

It also includes one meaningful change:
1. Apply styles to `h6`. It looks like there was a typo in the previous rules, with two sets being applied to `h5` and none to `h6`.

Thanks for this incredible piece of software!


jgm:
Great, thanks!

PR #9125, opened 2023-10-08
Fix abbreviation for astronomical unit

mcclurgm:
In the siunitx implementation, the abbreviation for "astronomicalunit" is currently "ua". It should be "au" (see table 3 of the [official documentation](https://ctan.mirrors.hoobly.com/macros/latex/contrib/siunitx/siunitx.pdf)). This PR corrects that.


jgm:
thanks!

Issue #8763, opened 2023-04-07
Citations do not work with Typst writer


benjamineskola:
**Explain the problem.**
I'm trying to generate a PDF, including citations, using as minimal a change as possible from `pdf-engine: latex` to `pdf-engine: typst`. When simply changing the engine to `typst` failed, I tried changing the final output format to typst so I could see the output and try to make sense of the issue.

Firstly, an absolute path to the bibliography fails. If I generate with `pandoc -s --bibliography "/Users/ben/Documents/Study/bibliography.yaml" …` then I get a typst file containing the following:

```typst
…
#bibliography("/Users/ben/Documents/Study/bibliography.yaml")
```

When I tried processing this the error was as follows:

```
$ typst TMA04.typ
error: file not found (searched at Users/ben/Documents/Study/bibliography.yaml)
    ┌─ TMA04.typ:135:14
    │
135 │ #bibliography("/Users/ben/Documents/Study/bibliography.yaml")
    │               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Note the lack of leading `/` in the error message: Typst seems to be stripping this and treating it as a relative path.

Passing a relative path to Pandoc (e.g. `pandoc -s --bibliography=../bibliography.yaml …`) seems to work — at least, it finds the bibliography file.

However this then produces the second error:

```
$ typst TMA04.typ
error: unknown bibliography format
    ┌─ TMA04.typ:135:14
    │
135 │ #bibliography("../bibliography.yaml")
    │               ^^^^^^^^^^^^^^^^^^^^^^
```

My expectation was that a citeproc yaml file would be processed by pandoc to generate the output (as with the other output formats, which as far as I know don't support citeproc). However it seems that the typst writer needs the yaml file to be in a different format (typst-specific?).

I don't know if the right thing to do is to alter Pandoc's behaviour or just to document it — I can completely understand if it's a limitation of the output format.

**Pandoc version?**
pandoc 3.1.2 on macOS 13.3

jgm:
> My expectation was that a citeproc yaml file would be processed by pandoc to generate the output (as with the other output formats, which as far as I know don't support citeproc).

This is correct, *if* you specify `--citeproc` on the command line. Did you?  Even if you did, I think there may be a bug; pandoc should not be passing through the `bibliography` to the template in this case.

jgm:
[...]

PR #9127, opened 2023-10-10
Document typst-citations to enable citeproc for typst output

iandol:
See https://github.com/jgm/pandoc/issues/8763#issuecomment-1753234008 for context, basically the `citations` extension applies to docx and org readers but has been overloaded to also affect the typst writer to allow citeproc to work instead of typst citation system. Not so easy to document so this seems the best place?

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1422,8 +1422,9 @@ header when requesting a document from a URL:
     The citation processing transformation may be applied before
     or after filters or Lua filters (see `--filter`,
     `--lua-filter`): these transformations are applied in the
-    order they appear on the command line.  For more
-    information, see the section on [Citations].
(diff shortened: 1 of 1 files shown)


salim-b:
Maybe also add a separate ``### Extension: `citations` (typst) {#typst-citations}`` section like we already have for [`citations` (org)](https://pandoc.org/MANUAL.html#org-citations) and [`citations` (docs)](https://pandoc.org/MANUAL.html#docx-citations) (introduced in https://github.com/jgm/pandoc/commit/95bff94ae55a2b748353181083c1738d907ed1bb)?

jgm:
I've superseded this with my own changes, thanks.

iandol:
Thanks Jon!

PR #9131, opened 2023-10-12
Make suppress-bibliography work with --biblatex

bgrenet:
Currently, setting the metadata variable suppress-bibliography has no effect with --biblatex nor --natbib. I add a simple conditional to make it work.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -553,7 +553,6 @@ $body$
 $if(has-frontmatter)$
 \backmatter
 $endif$
-$if(suppress-bibliography)$$else$
 $if(natbib)$
 $if(bibliography)$
 $if(biblio-title)$
@@ -574,6 +573,7 @@ $endif$
 
 $endif$
 $endif$
+$if(suppress-bibliography)$$else$
 $if(biblatex)$
 $if(beamer)$
 \begin{frame}[allowframebreaks]{$biblio-title$}
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -553,6 +553,7 @@ $body$
 $if(has-frontmatter)$
 \backmatter
 $endif$
+$if(suppress-bibliography)$$else$
 $if(natbib)$
 $if(bibliography)$
 $if(biblio-title)$
@@ -583,6 +584,7 @@ $else$
 \printbibliography$if(biblio-title)$[title=$biblio-title$]$endif$
 $endif$
 
+$endif$
 $endif$
 $for(include-after)$
 $include-after$
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -573,7 +573,8 @@ $endif$
 
 $endif$
 $endif$
-$if(suppress-bibliography)$$else$
+$if(suppress-bibliography)$
+$else$
 $if(biblatex)$
 $if(beamer)$
 \begin{frame}[allowframebreaks]{$biblio-title$}


jgm:
Looks like this would only affect `--natbib`, not `--biblatex`...

bgrenet:
> Looks like this would only affect `--natbib`, not `--biblatex`...

The conditional encompasses the two versions, it opens on line 556 and closes on line 588. I've tested it and seems to work. 

For the CI, from what I understand there are some failure because my changes introduce some new extra blank lines, right?

jgm:
OK, that helps. The template's structure is hard to see -- I've added some tabs to help myself:
```
$if(natbib)$
  $if(bibliography)$
    $if(biblio-title)$
      $if(has-chapters)$
      \renewcommand\bibname{$biblio-title$}
      $else$
      \renewcommand\refname{$biblio-title$}
      $endif$
    $endif$
    $if(beamer)$
      \begin{frame}[allowframebreaks]{$biblio-title$}
        \bibliographytrue
    $endif$
    \bibliography{$for(bibliography)$$bibliography$$sep$,$endfor$}
    $if(beamer)$
      \end{frame}
    $endif$
  $endif$
$endif$
$if(biblatex)$
  $if(beamer)$
    \begin{frame}[allowframebreaks]{$biblio-title$}
      \bibliographytrue
      \printbibliography[heading=none]
    \end{frame}
  $else$
    \printbibliography$if(biblio-title)$[title=$biblio-title$]$endif$
  $endif$
$endif$
```
If I understand correctly, you're adding a conditional that omits this whole section when `suppressbibliography` is used.

My understanding was that in natbib, `\bibliography` is needed not only to print the bibliography but to tell the compiler what bibtex file to look at.  In biblatex, these two functions are separated out into `\addbibresource` and `\printbibliography`.  So, does it really work with natbib if there's nothing in the document telling latex what the bibliography file is?

On the test suite:  yes, if your changes introduced blank lines, then either you should modify the template or the test suite accordingly.
You could try changing
```
$if(suppress-bibliography)$$else$
```
to
```
$if(suppress-bibliography)$
$else$
```
I know there's some fancy stuff in doctemplates regarding blank lines, but I don't recall the details.

bgrenet:
> If I understand correctly, you're adding a conditional that omits this whole section when `suppressbibliography` is used.

Right.

> My understanding was that in natbib, `\bibliography` is needed not only to print the bibliography but to tell the compiler what bibtex file to look at. In biblatex, these two functions are separated out into `\addbibresource` and `\printbibliography`. So, does it really work with natbib if there's nothing in the document telling latex what the bibliography file is?

I had the impression that it works, but I have to run more extensive tests to be sure. I'll do that and do accordingly. (I am not used to natbib so I have to be more careful I guess.)

> On the test suite: yes, if your changes introduced blank lines, then either you should modify the template or the test suite accordingly.

Ok I'll do that.

bgrenet:
> My understanding was that in natbib, `\bibliography` is needed not only to print the bibliography but to tell the compiler what bibtex file to look at. In biblatex, these two functions are separated out into `\addbibresource` and `\printbibliography`. So, does it really work with natbib if there's nothing in the document telling latex what the bibliography file is?

I did some tests with natbib. The point would be to allow the user to print the bibliography at the place she wants. Yet the TeX result would be valid (would compile) only if the user has issued a `\bibliography{...}` somewhere in the text. Since I am not sure my change meets some necessity, I removed it. 

So currently (I am waiting for CI reports) I implement only suppress-bibliography for --biblatex where it is very useful, for instance to issue manually one or several `\printbibliography`.

jgm:
There may be a way to make it work with natbib too.
https://tex.stackexchange.com/questions/229895/is-there-a-way-to-hide-the-bibliography-references-section-altogether
Another interesting approach using savebox:
https://tex.stackexchange.com/questions/31373/citations-with-no-bibliography-at-the-end

bgrenet:
I've seen and tested these solutions. Actually, my problem with them is that there are several seemingly mutually exclusive solutions, that actually depend on what the user wants and does:

- Either one issues `\nobibliography*` (or even nothing): The user must provide a `\bibliography{...}`
- Or one issues `\nobibliography{biblio.bib}` or uses the savebox method: The user cannot provide`\bibliography{...}` anymore and can only issue some `\bibentry{...}`'s manually.

My understanding is that the first solution is ok for users that want to print the bibliography but not at the end, while the second solution is ok for users that want not to print any bibliography, or to print it by hand entry per entry.

Actually, the `\bibliography{...}` is issued only if the user provide a `--bibliography=...`: It seems that `suppress-bibliography` is redundant then… Do you agree?

bgrenet:
Now I see one possible solution for natbib:

- If a user wants to provide `\bibliography{...}` by themself, they can simply use `--natbib` without `--bibliography=...`;
- If a user wants to only have citations but no bibliography, or to print some bib entries only, they can use `--natbib` with `--bibliography=...` and `--suppress-bibliography`.

If this makes sense, I can implement it. Not sure it's worth it though.

jgm:
> If a user wants to only have citations but no bibliography, or to print some bib entries only, they can use `--natbib` with `--bibliography=...` and `--suppress-bibliography`.

That would be the intended behavior, yes, but the question we're asking is how is this to be implemented?  That is, what will `--suppress-bibliography` do, in terms of the template? If it removes `\bibliography{}` then citeproc won't generate citations either.  That's the whole problem we've been discussing, so I must not be understanding you.

bgrenet:
> That would be the intended behavior, yes, but the question we're asking is how is this to be implemented? That is, what will `--suppress-bibliography` do, in terms of the template? If it removes `\bibliography{}` then citeproc won't generate citations either. That's the whole problem we've been discussing, so I must not be understanding you.

I don't see the problem since, as the pandoc user's guide states, the `--natbib` option "is not for use with the `--citeproc` option or with PDF output." Therefore one can implement something for `--natbib` that won't have any effect with `--citeproc`. I just tested with my initial implementation (which was buggy for other reasons): `pandoc MWE.md --citeproc --metadata=suppress-bibliography --bibliography=biblio.bib -o MWE.pdf` works perfectly, the citation is correctly typeset and there is no bibliography at the end of the document.

jgm:
> I don't see the problem since, as pandoc user's guide states, the `--natbib` option "is not for use with the `--citeproc` option or with PDF output."

That's right, but using `--bibliography` with `--natbib` doesn't trigger citeproc, it only sets the content of `\bibliography{..}` in the template, which is something you *need* for natbib.

PR #9134, opened 2023-10-13
Correct location of alt-text content

kamoe:
Fix JATS reader erroneously attempting to fetch alt-text content from non-existent @alt-text attribute, to optional child element <alt-text>

### DIFF
--- a/src/Text/Pandoc/Readers/JATS.hs
+++ b/src/Text/Pandoc/Readers/JATS.hs
@@ -142,11 +142,13 @@ getGraphic :: PandocMonad m
            => Maybe (Inlines, Text) -> Element -> JATS m Inlines
 getGraphic mbfigdata e = do
   let atVal a = attrValue a e
+  let altText = filterElement (named "alt-text") e of
+    Just alt = alt
+    Nothing = mempty
       (ident, title, capt) =
          case mbfigdata of
            Just (capt', i) -> (i, "fig:" <> atVal "title", capt')
-           Nothing        -> (atVal "id", atVal "title",
-                              text (atVal "alt-text"))
+           Nothing        -> (atVal "id", atVal "title", altText)
       attr = (ident, T.words $ atVal "role", [])
       imageUrl = atVal "href"
   return $ imageWith attr imageUrl title capt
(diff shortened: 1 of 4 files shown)


src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -138,18 +138,16 @@ trimNl = T.dropAround (== '\n')
 
 -- function that is used by both graphic (in parseBlock)
 -- and inline-graphic (in parseInline)
-getGraphic :: PandocMonad m
-           => Maybe (Inlines, Text) -> Element -> JATS m Inlines
```
kamoe, line 142: Noticed this function was only ever called with Nothing as first argument, which means it never actually uses it. Got rid of it.

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -138,18 +138,16 @@ trimNl = T.dropAround (== '\n')
...
   let atVal a = attrValue a e
-      (ident, title, capt) =
-         case mbfigdata of
-           Just (capt', i) -> (i, "fig:" <> atVal "title", capt')
-           Nothing        -> (atVal "id", atVal "title",
-                              text (atVal "alt-text"))
+  let altText = case filterElement (named "alt-text") e of
+         Just alt -> textContent alt
+         Nothing -> mempty
+      (ident, title, altText') = (atVal "id", atVal "title", text altText)
```
kamoe, line 147: third argument is actually capturing alt-text, not caption, renamed as such

src/Text/Pandoc/Readers/JATS.hs
```diff
@@ -138,18 +138,16 @@ trimNl = T.dropAround (== '\n')
...
-getGraphic mbfigdata e = do
+getGraphic :: PandocMonad m => Element -> JATS m Inlines
+getGraphic e = do
   let atVal a = attrValue a e
-      (ident, title, capt) =
-         case mbfigdata of
-           Just (capt', i) -> (i, "fig:" <> atVal "title", capt')
-           Nothing        -> (atVal "id", atVal "title",
-                              text (atVal "alt-text"))
+  let altText = case filterElement (named "alt-text") e of
```
kamoe, line 144: this is where we fetch content from a child, rather than an argument

jgm:
great, thanks!

PR #9142, opened 2023-10-16
Handle empty cells correctly in creole reader. Fixes #9141

### LINKED ISSUE #9141: Creole reader does not handle empty table cells correctly
Empty table cells are only recognized by the creole parser if they contain at least one space.

Example:
```
echo -e "| a | bbb |||\n| c | ddd |||" >x.creole
pandoc --standalone --from creole x.creole | grep ddd
  <p>| a | b ||| | c | ddd |||</p>

echo -e "| a | bbb | | |\n| c | ddd | | |" >y.creole
pandoc --standalone --from creole y.creole | grep 'ddd'
<td>ddd</td>
```

Kudos to @bernhardreiter for reporting this to me.

swilde:
Handle empty table cells of the form `|||` correctly.  Without this fix empty cells must at least contain one space like so `| | |` which shouldn't be necessary.


jgm:
thanks!

Issue #8986, opened 2023-08-07
Issue with `.docx` reader's handling of top styles in international documents


StephanMeijer:
# Explain the problem.

How does the `.docx` reader in Pandoc determine the top style, such as _Title_, and what implications does this approach have for international documents? Specifically, in Dutch (NL) documents, the top style for _Title_ is often named _Title_ but has a Style ID of _Titel_ (the Dutch translation for _Title_). 

I believe this might result in the Title being converted to a regular paragraph.

# Pandoc version?

```
pandoc 3.1.6
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: /Users/steve/.local/share/pandoc
Copyright (C) 2006-2023 John MacFarlane. Web: https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.
```

# Examples

_Coming soon..._

StephanMeijer:
I'm currently unable to provide examples but I think this `HeadingInPairs` attribute might have something to do with this..

<img width="602" alt="image" src="https://github.com/jgm/pandoc/assets/3169264/58bb84e5-606b-4df1-8da2-756182ae1e2d">

StephanMeijer:
ChatGPT guesses this Word-file to be created with Microsoft Word 2016
<img width="851" alt="image" src="https://github.com/jgm/pandoc/assets/3169264/6701f457-b23d-46f7-8c4a-7ad4dc41eff6">

StephanMeijer:
Example added.

jgm:
You are right: we just look at the style.
```hs
metaStyles :: M.Map ParaStyleName T.Text
metaStyles = M.fromList [ ("Title", "title")
                        , ("Subtitle", "subtitle")
                        , ("Author", "author")
                        , ("Date", "date")
                        , ("Abstract", "abstract")]
```
Paragraphs with these styles turn into metadata values.
I'm not sure how to deal with the full variety of style names in other languages.

jgm:
I would have assumed that the style ID would stay the same in localizations, while the style name changes, but you are reporting the reverse.  It would be good to have more information here from others using localized versions of Word.

StephanMeijer:
@jgm Is this an issue that ought to be resolved in Pandoc itself, or is it better to so some preprocessing on our side? 

This anonymized example is based on a real-life document we got. So I assume it might be best to fix it in Pandoc? Or is the setup like this invalid per standards?

[...]

PR #9162, opened 2023-10-31
#8986 Image before Title still respects Title metadata

StephanMeijer:
The way I implemented this:

1. First comes regular stuff, like images or anything else
2. Then a block of meta comes including empty paragraphs
3. Then regular stuff like images and paragraphs again

Then it's processed in this order:

1. Meta without any empty paragraphs
2. Non-meta
    a. First block of regular stuff
    b. All empty paragraphs from within Meta
    c. Second block of regular stuff

Visualisation of processing logic

-----

_Our internal ID: [NLDOC-837](https://digitoegankelijk.atlassian.net/browse/NLDOC-837)_


test/docx/0_level_headers.native
```diff
@@ -1,52 +1,182 @@
...
-,Para [Str "SECTION",Space,Str "1",Space,Str "Introduction",Space,Link ("",[],[]) [Str "2"] ("#introduction","")]
-,Header 1 ("figures",["Heading-0"],[]) [Str "FIGURES"]
-,Para [Strong [Str "Figure",Space,Str "Page"]]
-,Para [Strong [Str "No",Space,Str "table",Space,Str "of",Space,Str "figures",Space,Str "entries",Space,Str "found."]]
-,Header 1 ("tables",["Heading-0"],[]) [Str "TABLES"]
-,Para [Strong [Str "Table",Space,Str "Page"]]
-,Para [Strong [Str "No",Space,Str "table",Space,Str "of",Space,Str "figures",Space,Str "entries",Space,Str "found."]]
-,Header 1 ("introduction",[],[]) [Str "Introduction"]
-,Para [Str "Nothing",Space,Str "to",Space,Str "introduce,",Space,Str "yet."]]
+Pandoc
```
StephanMeijer, line 1: Screenshot 2023-10-31 at 14 27 39

StephanMeijer: Not sure what to do with this. It is entirely logical this is being converted to a `Title` (meta) so I would prefer letting that meaning just sit in there and maybe change the example.

pkrijgsman (approved):
Looks great, this should be a fix for the issues that we have.

StephanMeijer:
@jgm responded at https://github.com/jgm/pandoc/pull/9170#issuecomment-1795353801

> This approach (skipping images) feels a bit ad hoc to me.
> 
> I'm thinking it might be preferable to allow items with class Title, Author, or Date to be counted as metadata no matter whether they occur in the document. In the case of Title or Date, only the first would be counted as part of metadata if there are multiple. In the case of multiple Author, they could all be included in a list.

StephanMeijer:
@jgm I changed the PR based on your feedback in my other PR.

test/docx/0_level_headers.native
jgm: Can you explain more fully?  Are you saying that the Title style is used for both the document title and the title of the table of contents section?

jgm:
Actually, my suggestion was to replace all the "splitting" code with the simple behavior I mentioned.  Was trying to make it simpler, not more complicated.

StephanMeijer:
@jgm I think I misunderstood what you mean then, as I implemented (I think) exactly what you described. Could you elaborate?

src/Text/Pandoc/Readers/Docx.hs
```diff
@@ -169,30 +169,54 @@ spansToKeep = []
...
+    (singleMetas, restWithoutRelevantMeta) = foldr extractSingle ([], restWithoutMulti) (M.keys singleMetaStyles)
+
+    extractSingle :: ParaStyleName -> ([BodyPart], [BodyPart]) -> ([BodyPart], [BodyPart])
+    extractSingle styleName (accSingleMetas, remainingBPs) =
+      let (found, rest) = extractFirst (isSingleMetaPar styleName) remainingBPs
+      in (maybeToList found ++ accSingleMetas, rest)
+
+    maybeToList :: Maybe a -> [a]
+    maybeToList Nothing = []
+    maybeToList (Just x) = [x]
```
jgm, line 201: I think there's a function like this in Data.Maybe

jgm:
I think you do implement what I suggested.  But I guess I'd had in mind what might be a simpler approach.  Instead of going through all the bp's and splitting them into metadata and non-metadata, I thought it would be simpler to simply keep a state variable for metadata and populate it as we go through the bp's, sequentially.  Wouldn't that allow us to avoid all the splitting code at the beginning, while producing the same result?  Let me know if that makes sense.

StephanMeijer:
@jgm it absolutely makes sense!

PR #9165, opened 2023-11-02
Center tables in HTML

remysucre:
Modern computer screens are quite wide, so a center-aligned table looks much better than a left-aligned one. For example, compare the two below. This can be achieved by simplifying the CSS template as done in this PR. 

Screenshot 2023-11-01 at 8 49 41 PM
Screenshot 2023-11-01 at 8 49 33 PM

### DIFF
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -136,11 +136,9 @@ hr {
   margin: 1em 0;
 }
 table {
-  margin: 1em 0;
+  margin: auto;
   border-collapse: collapse;
-  width: 100%;
   overflow-x: auto;
-  display: block;
   font-variant-numeric: lining-nums tabular-nums;
 }
 table caption {


jgm:
This should be tested with a much wider selection of tables, including tables with relative column widths (e.g. some of the later tables in test/tables.native or test/tables-rstsubset.native).  There may be reasons for the things that were deleted.

mb21:
> Modern computer screens are quite wide

yes but the width of the default CSS actually constrains the text width to 80 characters or so.

> a center-aligned table looks much better than a left-aligned one.

for a very small table like the one in the screenshot, I agree. but for even just a bit wider tables, I think it's less clear and may be subjective.

Either way, some things to look out for when changing this is to test with big (very wide) tables, and test on mobile phones: the `width: 100%` is there so that the table is horizontally scrollable if it's wider than the screen.

remysucre:
> the table is horizontally scrollable if it's wider than the screen. 

I see. I couldn't come up with a way to both center narrow tables and make wide tables scroll. IMO the default template should optimize for tables that are not too wide, and people should provide custom styles for wide tables. After all that's the default behavior for tables in CSS. 

I do agree this is a subjective issue so feel free to close.

Issue #8982, opened 2023-08-04
PDF compilation fails on italic text containing multiblock footnote


brandondyck:
My markdown input, `test.markdown`, has an italicized sentence with a multiblock footnote in it:
```
_This sentence[^1] has a footnote._

[^1]: Here's a paragraph.
    
    Here's another paragraph.
```

When I convert it with `pandoc -o test.pdf test.markdown`, pdftex chokes on its input:
```
Error producing PDF.
! Paragraph ended before \text@command was complete.
<to be read again>
                   \par
l.60 ...e’s another paragraph.} has a footnote.}
```

Converting it to a .tex file with `pandoc -o test.tex test.markdown` gives the following:
```
\emph{This sentence}\footnote{Here's a paragraph.

  Here's another paragraph.} \emph{has a footnote.}
```

Pandoc creates the PDF properly if I remove either the italics or the second block of the footnote. For now I'm working around it by writing the first part as `_This sentence_[^1] _has a footnote._`

I'm using the following software:
- Pandoc 3.1.6
- Ubuntu 22.04.1 LTS on WSL2
- pdfTeX 3.141592653-2.6-1.40.22 (TeX Live 2022/dev/Debian)

ibayashi-hikaru:
I fixed this bug in this PR #9168. 
Can you take a look please?

Note: I'm new to the pandoc project.

PR #9168, opened 2023-11-05
Bug fix for the Latex writer issue #8982

ibayashi-hikaru:
Latex Writer converts
`emph (str "This sentence" <> note (para (str "paragraph1") <> para (str "paragraph2")) <> str " has footnote.")` to
`"\\emph{This sentence}\\footnote{paragraph1\n\n  paragraph2}\\emph{ has footnote.}"` instead of
`"\\emph{This sentence\\footnote{paragraph1\n\n  paragraph2} has footnote.}"`

### DIFF
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -127,6 +127,41 @@ convertWithOpts scriptingEngine opts = do
                Left e -> E.throwIO $ PandocShouldNeverHappenError $ T.pack e
           | otherwise -> writeFnBinary outputFile bs
 
+isolateBigNote :: ([Inline] -> Inline) -> [Inline] -> [Inline]
+isolateBigNote constructor xs
+  | any isBigNote xs = let (before, after) = break isBigNote xs
+              in case after of
+                   (noteInline:rest) -> constructor before : noteInline : isolateBigNote constructor rest
+                   _ -> error "break should always return a non-empty 'after' when 'any isNote xs' is True."
+  | otherwise = [constructor xs]
+  where
+    isBigNote :: Inline -> Bool
+    isBigNote (Note blocks) = countParas blocks > 1
+      where
+        countParas :: [Block] -> Int
+        countParas = length . filter isPara
+
+        isPara :: Block -> Bool
+        isPara (Para _) = True
+        isPara _        = False
+    isBigNote _ = False
+
+liftBigNotes :: [Inline] -> [Inline]
+liftBigNotes [] = []
+liftBigNotes (x:xs) =
+    case x of
+        Emph inner    -> isolateBigNote Emph (liftBigNotes inner) ++ liftBigNotes xs
+        Strong inner  -> isolateBigNote Strong (liftBigNotes inner) ++ liftBigNotes xs
+        _             -> x : liftBigNotes xs
+
+liftBigNotesInBody :: Pandoc -> Pandoc
+liftBigNotesInBody (Pandoc meta bs) = Pandoc meta (fmap liftBigNotesInBlock bs)
+  where
+    liftBigNotesInBlock :: Block -> Block
+    liftBigNotesInBlock (Para xs)  = Para (liftBigNotes xs)
+    liftBigNotesInBlock (Plain xs) = Plain (liftBigNotes xs)
+    liftBigNotesInBlock others     = others
+
 convertWithOpts' :: (PandocMonad m, MonadIO m, MonadMask m)
                  => ScriptingEngine
                  -> Bool
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -127,41 +127,6 @@ convertWithOpts scriptingEngine opts = do
                Left e -> E.throwIO $ PandocShouldNeverHappenError $ T.pack e
           | otherwise -> writeFnBinary outputFile bs
 
-isolateBigNote :: ([Inline] -> Inline) -> [Inline] -> [Inline]
-isolateBigNote constructor xs
-  | any isBigNote xs = let (before, after) = break isBigNote xs
-              in case after of
-                   (noteInline:rest) -> constructor before : noteInline : isolateBigNote constructor rest
-                   _ -> error "break should always return a non-empty 'after' when 'any isNote xs' is True."
-  | otherwise = [constructor xs]
-  where
-    isBigNote :: Inline -> Bool
-    isBigNote (Note blocks) = countParas blocks > 1
-      where
-        countParas :: [Block] -> Int
-        countParas = length . filter isPara
-
-        isPara :: Block -> Bool
-        isPara (Para _) = True
-        isPara _        = False
-    isBigNote _ = False
-
-liftBigNotes :: [Inline] -> [Inline]
-liftBigNotes [] = []
-liftBigNotes (x:xs) =
-    case x of
-        Emph inner    -> isolateBigNote Emph (liftBigNotes inner) ++ liftBigNotes xs
-        Strong inner  -> isolateBigNote Strong (liftBigNotes inner) ++ liftBigNotes xs
-        _             -> x : liftBigNotes xs
-
-liftBigNotesInBody :: Pandoc -> Pandoc
-liftBigNotesInBody (Pandoc meta bs) = Pandoc meta (fmap liftBigNotesInBlock bs)
-  where
-    liftBigNotesInBlock :: Block -> Block
-    liftBigNotesInBlock (Para xs)  = Para (liftBigNotes xs)
-    liftBigNotesInBlock (Plain xs) = Plain (liftBigNotes xs)
-    liftBigNotesInBlock others     = others
-
 convertWithOpts' :: (PandocMonad m, MonadIO m, MonadMask m)
                  => ScriptingEngine
                  -> Bool
@@ -276,8 +241,7 @@ convertWithOpts' scriptingEngine istty datadir opts = do
   metadataFromFile <- getMetadataFromFiles readerNameBase readerOpts
                          (optMetadataFiles opts)
 
-  let transforms = (liftBigNotesInBody :) .
-                   (case optShiftHeadingLevelBy opts of
+  let transforms = (case optShiftHeadingLevelBy opts of
                         0             -> id
                         x             -> (headerShift x :)) .
                    (if extensionEnabled Ext_east_asian_line_breaks
--- a/src/Text/Pandoc/App.hs
+++ b/src/Text/Pandoc/App.hs
@@ -276,7 +276,8 @@ convertWithOpts' scriptingEngine istty datadir opts = do
   metadataFromFile <- getMetadataFromFiles readerNameBase readerOpts
                          (optMetadataFiles opts)
 
-  let transforms = (case optShiftHeadingLevelBy opts of
+  let transforms = (liftBigNotesInBody :) .
+                   (case optShiftHeadingLevelBy opts of
                         0             -> id
                         x             -> (headerShift x :)) .
                    (if extensionEnabled Ext_east_asian_line_breaks
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -64,6 +64,7 @@ import Text.Pandoc.Writers.LaTeX.Util (stringToLaTeX, StringContext(..),
                                        getListingsLanguage, mbBraced)
 import Text.Pandoc.Writers.Shared
 import qualified Text.Pandoc.Writers.AnnotatedTable as Ann
+-- Work around problems with notes inside emphasis (see #8982)
 
 isolateBigNotes :: ([Inline] -> Inline) -> [Inline] -> [Inline]
 isolateBigNotes constructor xs =
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -65,10 +65,37 @@ import Text.Pandoc.Writers.LaTeX.Util (stringToLaTeX, StringContext(..),
 import Text.Pandoc.Writers.Shared
 import qualified Text.Pandoc.Writers.AnnotatedTable as Ann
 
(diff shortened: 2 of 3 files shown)


jgm:
Thanks, and welcome to pandoc!

This patch looks good, it will do the job. We could merge it .But I was thinking that a more general solution might be worth considering. After all, the same problem that comes up here with Emph will arise with Strong, and probably also Underline and other inline containers.

Another issue is that I don't think this patch will handle cases of double nesting, e.g. a footnote inside an Emph inside another Emph.

One more general solution would be an AST transformation that would cause the notes to "bubble up" to the top level:

```
[Emph [Str "a", Emph [Str "b", Note [Plain [Str "note"]], Str "c"], Str "d"]]
->
[Emph [Str "a", Emph [Str "b"], Note [Plain [Str "note"]], Emph [Str "c"], Str "d"]]
->
[Emph [Str "a", Emph [Str "b"]], Note [Plain [Str "note"]], Emph [Emph [Str "c"], Str "d"]]
```

ibayashi-hikaru:
Hi @jgm,
Thank you for the warm welcome! Your comment makes total sense.
I will apply further change based on that.
I will get back to you soon.

jgm:
Another note:  I think we only need to do this in cases where the note contains more than a single Para or Plain element.  Of course, we could just do it for all notes.  Not sure if there's an advantage to being selective -- but in some cases people might expect e.g. that the note reference would be in italics or boldface, and we could at least preserve that for notes that don't contain multiple blocks (which is most notes).

ibayashi-hikaru:
Hi @jgm,
As you suggested, I fixed this bug more generally [by lifting up `Note` when transforming AST](https://github.com/jgm/pandoc/pull/9168/files#diff-443ac0cb5dc534fd8342ce0ba3c3f7a1e90b0c2b3499914f6787d3b832a9144aR279)

Also, I designed this lift-up selective, i.e., [`Note`s are lifted up when it has more than one paragraph](https://github.com/jgm/pandoc/pull/9168/files#diff-443ac0cb5dc534fd8342ce0ba3c3f7a1e90b0c2b3499914f6787d3b832a9144aR138)

Can you take a look please?

Thank you

src/Text/Pandoc/App.hs
```diff
@@ -127,6 +127,41 @@ convertWithOpts scriptingEngine opts = do
...
+    isBigNote _ = False
+
+liftBigNotes :: [Inline] -> [Inline]
+liftBigNotes [] = []
+liftBigNotes (x:xs) =
+    case x of
+        Emph inner    -> isolateBigNote Emph (liftBigNotes inner) ++ liftBigNotes xs
+        Strong inner  -> isolateBigNote Strong (liftBigNotes inner) ++ liftBigNotes xs
+        _             -> x : liftBigNotes xs
+
```
jgm, line 156: I think this is unnecessary.  If you define
```
liftBigNotes :: Inline -> Inline
liftBigNotes (Emph inner) = ...
etc.
```
then you can lift this `Inline -> Inline` transformation to a `Pandoc -> Pandoc` transformation using `walk`.
```
walk liftBigNotes
```
can be made to have the type Pandoc -> Pandoc, and everything should work (of course, needs testing).

src/Text/Pandoc/App.hs
```diff
@@ -127,6 +127,41 @@ convertWithOpts scriptingEngine opts = do
...
+liftBigNotesInBody :: Pandoc -> Pandoc
+liftBigNotesInBody (Pandoc meta bs) = Pandoc meta (fmap liftBigNotesInBlock bs)
+  where
+    liftBigNotesInBlock :: Block -> Block
+    liftBigNotesInBlock (Para xs)  = Para (liftBigNotes xs)
+    liftBigNotesInBlock (Plain xs) = Plain (liftBigNotes xs)
+    liftBigNotesInBlock others     = others
+
```
jgm, line 164: See above.  All you need is an Inline -> Inline transformation; `walk` will then magically turn this into a `Pandoc -> Pandoc` transformation.

jgm (changes requested):
In addition to the comments I made on the code, this needs some tests to make sure it's handling all the cases we described.

These could be "command tests" (see test/command) or you could work them into test/Test/Writers/LaTeX.hs.

src/Text/Pandoc/App.hs
ibayashi-hikaru: Thank you for the feedback.

I'm not sure how to translate this `liftBigNotes :: [Inline] -> [Inline]` to `liftBigNotes :: Inline -> Inline`.
I feel the result values will need to have `[Inline]` type. See the following wrong code.


```haskell
liftBigNotes :: Inline -> Inline
liftBigNotes (Emph inner)    = isolateBigNote Emph (liftBigNotes inner) ++ liftBigNotes xs
liftBigNotes (Strong inner)  = isolateBigNote Strong (liftBigNotes inner) ++ liftBigNotes xs
liftBigNotes others          = [others]
```
This code is wrong because 
1. `liftBigNotes`'s output has the `Inline` type, while the actual function body outputs the `[Inline]` type
2. `liftBigNotes inner`’s inner has the `[Inline]`  type while `liftBigNotes` expects the `Inline` type

Can you elaborate a little more on how to modify this function?

jgm: Sorry, right.  What about the following, with type `[Inline] -> [Inline]`?
``` hs
liftBigNotes :: [Inline] -> [Inline]
liftBigNotes (Emph inner : xs)    = isolateBigNote Emph (liftBigNotes inner) ++ liftBigNotes xs
liftBigNotes (Strong inner : xs)  = isolateBigNote Strong (liftBigNotes inner) ++ liftBigNotes xs
liftBigNotes (x : xs)          = x : liftBigNotes xs
liftBigNotes [] = []
```
I believe that `walk liftBigNotes` will lift this to a `Pandoc -> Pandoc` transformation.

jgm: I worry a bit about performance implications, but we can test that.

ibayashi-hikaru:
Hi @jgm,
I updated the patch. Here is the quick summary.
- Moved all the transformation functions to `src/Text/Pandoc/Writers/LaTeX.hs`
- Simplifed function with `walk`
- Renamed `liftBigNotes` to `raiseBigNotes` as `lift` has a confusing connotation in Haskell.
- Made this transformation cover `Underline` and `Strikeout` (I think this is debatable. see below)
  1. Is it better to conver more or less?
  2. The current implementation will aplit the `Underline` (or`Strikeout`) with a big note. But I feel this wouldn't be what user would expect.
  image

Thanks!

src/Text/Pandoc/App.hs
```diff
@@ -127,6 +127,41 @@ convertWithOpts scriptingEngine opts = do
...
+    isBigNote :: Inline -> Bool
+    isBigNote (Note blocks) = countParas blocks > 1
+      where
+        countParas :: [Block] -> Int
+        countParas = length . filter isPara
+
+        isPara :: Block -> Bool
+        isPara (Para _) = True
+        isPara _        = False
+    isBigNote _ = False
```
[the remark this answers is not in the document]
ibayashi-hikaru: That makes sense. Fixed.

src/Text/Pandoc/App.hs
ibayashi-hikaru: Function is updated. But I might need to address the performance implications later.

src/Text/Pandoc/App.hs
ibayashi-hikaru: Updated accordingly.

src/Text/Pandoc/App.hs
```diff
@@ -127,6 +127,41 @@ convertWithOpts scriptingEngine opts = do
                Left e -> E.throwIO $ PandocShouldNeverHappenError $ T.pack e
           | otherwise -> writeFnBinary outputFile bs
 
+isolateBigNote :: ([Inline] -> Inline) -> [Inline] -> [Inline]
+isolateBigNote constructor xs
+  | any isBigNote xs = let (before, after) = break isBigNote xs
+              in case after of
+                   (noteInline:rest) -> constructor before : noteInline : isolateBigNote constructor rest
+                   _ -> error "break should always return a non-empty 'after' when 'any isNote xs' is True."
```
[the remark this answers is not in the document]
ibayashi-hikaru: Updated accordingly.

src/Text/Pandoc/App.hs
```diff
@@ -127,6 +127,41 @@ convertWithOpts scriptingEngine opts = do
                Left e -> E.throwIO $ PandocShouldNeverHappenError $ T.pack e
           | otherwise -> writeFnBinary outputFile bs
 
+isolateBigNote :: ([Inline] -> Inline) -> [Inline] -> [Inline]
```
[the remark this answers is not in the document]
ibayashi-hikaru: Moved.

src/Text/Pandoc/App.hs
```diff
@@ -127,6 +127,41 @@ convertWithOpts scriptingEngine opts = do
...
+isolateBigNote constructor xs
+  | any isBigNote xs = let (before, after) = break isBigNote xs
+              in case after of
+                   (noteInline:rest) -> constructor before : noteInline : isolateBigNote constructor rest
```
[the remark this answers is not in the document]
ibayashi-hikaru: Updated.

jgm:
> The current implementation will split the Underline (orStrikeout) with a big note. But I feel this wouldn't be what user would expect.

You're absolutely right -- nobody would expect the gap this produces.  But the question is whether this is needed for the code to compile.  I haven't tested with underline and strikeout, but I imagine that the error noted in the original issue could be reproduced for them, too?  That would be the thing to check.

ibayashi-hikaru:
> I imagine that the error noted in the original issue could be reproduced for them, too? 

Yes, correct. The original pandoc cannot compile the following .tex code,

```latex
\ul{Footnote \footnote{paragraph 1


paragraph 2} separates an inline.}

\ul{Footnote does not \footnote{paragraph 1} separate an inline.}


\st{Footnote \footnote{paragraph 1


paragraph 2} separates an inline.}

\st{Footnote does not \footnote{paragraph 1} separate an inline.}

```
outputting the following error message.
```bash
! Paragraph ended before \SOUL@@ was complete.
<to be read again> 
                   \par 
l.66   paragraph 2} separates an inline.}
```

With my patch, pandoc converts it to [this pdf file](https://github.com/jgm/pandoc/files/13343373/example.pdf) (same as the screenshot above).

ibayashi-hikaru:
I think one way to move forward is to merge this PR as it is and open a separate issue for the broken `Underline` and `Strikeout`. Current patch does resolve the #8982 issue and I believe it needs another non-trivial change to fix the broken `Underline` and `Strikeout` issue.

How does it sound to you?

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -65,10 +65,37 @@ import Text.Pandoc.Writers.LaTeX.Util (stringToLaTeX, StringContext(..),
...
+  = isolateBigNotes Emph (raiseBigNotes inner) ++ raiseBigNotes xs
+raiseBigNotes (Strong inner : xs)
+  = isolateBigNotes Strong (raiseBigNotes inner) ++ raiseBigNotes xs
+raiseBigNotes (Underline inner : xs)
+  = isolateBigNotes Underline (raiseBigNotes inner) ++ raiseBigNotes xs
+raiseBigNotes (Strikeout inner : xs)
+  = isolateBigNotes Strikeout (raiseBigNotes inner) ++ raiseBigNotes xs
+raiseBigNotes (x : xs)          = x : raiseBigNotes xs
+raiseBigNotes [] = []
+
```
jgm, line 94: I'm getting a bit worried about the performance implications of this.
Can you run `make bench` with and without your PR to see what change we see in the LaTeX writer results?

src/Text/Pandoc/Writers/LaTeX.hs
```diff
@@ -65,10 +65,37 @@ import Text.Pandoc.Writers.LaTeX.Util (stringToLaTeX, StringContext(..),
 import Text.Pandoc.Writers.Shared
 import qualified Text.Pandoc.Writers.AnnotatedTable as Ann
```
jgm, line 67: It would be useful to add a comment here like
```
-- Work around problems with notes inside emphasis (see #8982)
```

src/Text/Pandoc/Writers/LaTeX.hs
ibayashi-hikaru: Below are the results of the benchmark tests. It looks like there is no performance regression. What do you think?
- Before my change
    - 1 st trial (I don’t know why it was slow.)
        
        ```jsx
        latex:                 OK (4.77s)
        5.10 ms ± 4.5 ms,  13 MB allocated, 435 KB copied, 122 MB peak memory,       same as baseline
        ```
        
    - 2nd trial
        
        ```jsx
        latex:                 OK (2.48s)
        2.49 ms ± 318 μs,  13 MB allocated, 442 KB copied, 128 MB peak memory,       same as baseline
        ```
        
- After my change
    - 1st trial
        
        ```jsx
        latex:                 OK (2.77s)
        2.80 ms ± 253 μs,  14 MB allocated, 538 KB copied, 124 MB peak memory,       same as baseline
        ```
        
    - 2nd trial
        
        ```jsx
        latex:                 OK (2.72s)
        2.79 ms ± 136 μs,  14 MB allocated, 541 KB copied, 124 MB peak memory,       same as baseline
        ```

src/Text/Pandoc/Writers/LaTeX.hs
ibayashi-hikaru: Added!

src/Text/Pandoc/Writers/LaTeX.hs
jgm: Assuming we throw out the first result as anomalous, it looks like there's a slight performance regression (on order of 10%).

If that's right, then it's a question of weighing a slight performance hit on all uses vs better correctness for a rare case (a note inside an emphasized section).

jgm:
looks good, thanks!

PR #9170, opened 2023-11-05
#8986 filter elements that only contain visuals like images from normal meta

StephanMeijer:
Prevents images as Title to be interpreted as Title.

## Visualisation

Untitled drawio (1)

## Test Document

Screenshot 2023-11-05 at 18 02 19


StephanMeijer:
Related to https://github.com/jgm/pandoc/issues/8986

jgm:
This approach (skipping images) feels a bit ad hoc to me.

I'm thinking it might be preferable to allow items with class Title, Author, or Date to be counted as metadata no matter whether they occur in the document.  In the case of Title or Date, only the first would be counted as part of metadata if there are multiple.  In the case of multiple Author, they could all be included in a list.

StephanMeijer:
@jgm I do agree that would work in the ideal world. We have users who markup their images as being `Title` when they are not a title but just incorrectly styled.

Is there any valid case for a situation where an image to be correctly marked as being a `Title`?

jgm:
I see now that my comment above wasn't really relevant; I think I was confused because you referenced #8986.  That was about skipping an image *not* marked with Title style and parsing metadata in following paragraphs.  This is really a different issue entirely.  

> Is there any valid case for a situation where an image to be correctly marked as being a Title?

Yes, although it's not common, I think that there are cases where an inline image might be part of a title.  I wouldn't want to make this impossible.

Issue #9181, opened 2023-11-13
Should document different commonmark behaviour re blank lines before lists


wlupton:
See [this pandoc-discuss thread](https://groups.google.com/g/pandoc-discuss/c/1vuWchFMg2c/m/cQECVgE1BwAJ) for context. I've since realised that the issue isn't really about definition lists; it's about whether or not blank lines are required before lists.

With the default markdown reader, there must be a blank line before a list: 
```
para
- item
```
gives:
```
<p>para - item</p>
```

But with the commonmark reader, such blank lines are optional and the above input (correctly; per the [principle of uniformity](https://spec.commonmark.org/0.30/#principle-of-uniformity)) gives this:
```
<p>para</p>
<ul>
<li>item</li>
</ul>
```

However, I don't believe that this commonmark difference is documented in the pandoc manual.

I can (and will) create a short PR to add a note to the manual, but I'm wondering whether it would be a good idea to capture this difference via a named extension.

(pandoc 3.1.9 on macOS)

jgm:
As noted in the PR, I don't think it's the job of the pandoc manual to describe all the differences between pandoc's markdown and commonmark.

wlupton:
@jgm wrote (in #9182):
> I'm not sure this belongs here. The pandoc manual tries to give a description of pandoc's markdown syntax (and extensions). For commonmark one should consult the commonmark spec.

OK, I get the point, but please note that there are already several notes such as the one that I suggested, although I do recognise that several of them came from me!

I just realised that there's already a [lists_without_preceding_blankline](https://pandoc.org/MANUAL.html#extension-lists_without_preceding_blankline) non-default extension, but (as explained in [Markdown Variants](https://pandoc.org/MANUAL.html#markdown-variants)) it doesn't show up as a `commonmark` extension.

I do feel that perhaps `commonmark`, `commonmark_x` and `gfm` are a bit of a special case, because their differences from pandoc markdown aren't exposed in terms of lists of named extensions.

Anyway, would you consider, rather than my proposed change, instead pointing to the [lists_without_preceding_blankline](https://pandoc.org/MANUAL.html#extension-lists_without_preceding_blankline) extension?

jgm:
The problem is that *this* wouldn't be the end of documenting differences btw pandoc's markdown and commonmark. There'd be many, many things that one would have to document. People who want to use commonmark should consult its own reference materials.

PR #9182, opened 2023-11-13
MANUAL: Add note about commonmark lists (#9181)

wlupton:


jgm:
I'm not sure this belongs here.  The pandoc manual tries to give a description of pandoc's markdown syntax (and extensions). For commonmark one should consult the commonmark spec.

Issue #8872, opened 2023-05-27
latex parsing: final item label in theorem environment applied to theorem rather than item


ultronozm:
pandoc version 3.1.2 on macOS 12.6, "pandoc test.tex -o test.md"

Input:

```
\documentclass{amsart}
\newtheorem{theorem}{Theorem}
\begin{document}
\begin{theorem}\label{thm}
\begin{enumerate}
\item \label{item1} text1
\item \label{item2} text2
\item text3
\end{enumerate}
\end{theorem}
\end{document}
```

Output:

```
::: {#item2 .theorem}
**Theorem 1**. *[]{#thm label="thm"}*

1.  *[]{#item1 label="item1"} text1*

2.  *text2*

3.  *text3*
:::
```

Expected output:

```
::: {#thm .theorem}
**Theorem 1**.

1.  *[]{#item1 label="item1"} text1*

2.  *[]{#item2 label="item2"} text2*

3.  *text3*
:::
```

That is to say, the final item label is applied to the enclosing block rather than to the item.

I wasn't able to find related issues in github or pandoc-discuss.  I would be happy to try to understand this issue better.  Would Text/Pandoc/Readers/LaTeX.hs be the place to look?

jgm:
T/P/Readers/LaTeX/Math.hs theoremEnvironment.

We currently do something fairly primitive.  When we encounter a label, we modify sLastLabel in state.  We then check this after parsing a theorem environment.

PR #9198, opened 2023-11-17
Fixed theorem label issue in Latex

ibayashi-hikaru:
This patch addresses the bug raised in #8872.

In the current implementation , theorem labels in `theoremEnvironment` are determined by the `LastLabel` in the current state. This approach works well when the `\label{thm}` is placed at the end of the theorem body (, which is a standard place to put a label.)

However, when a label is placed at the beginning of the theorem (another common practice) and additional labels follow in the theorem body, `theoremEnvironment` incorrectly picks the last label (e.g., `\label{item2}` in #8872).

This patch addresses the issue by extracting the label extraction independently of the `LastLabel` state.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX/Math.hs
+++ b/src/Text/Pandoc/Readers/LaTeX/Math.hs
@@ -11,7 +11,7 @@ module Text.Pandoc.Readers.LaTeX.Math
   , proof
   )
 where
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, mapMaybe, listToMaybe)
 import Text.Pandoc.Walk (walk)
 import Text.Pandoc.Builder as B
 import qualified Data.Sequence as Seq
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Readers/LaTeX/Math.hs
```diff
@@ -140,6 +140,15 @@ newtheorem inline = do
                             M.insert name spec tmap }
   return mempty
 
+extractLabelFromBlock :: Block -> Maybe Text
+extractLabelFromBlock (Para inlines) = extractLabel inlines
+  where
+    extractLabel :: [Inline] -> Maybe Text
+    extractLabel (Span (_, _, attrs) _:_) = lookup "label" attrs
+    extractLabel (_:xs) = extractLabel xs
+    extractLabel [] = Nothing
```
jgm, line 149: Won't this return Nothing if there is a span that isn't a label, even if there's a ltr span that *is* a label?
I think a better approach would be a fold, something like
```hs
extractLabel = foldl' go
go :: Maybe Text -> Inline -> Maybe Text
go (Just t) _ = Just t
go Nothing (Span...) = ... lookup as before ...
go Nothing _ = Nothing
````

ibayashi-hikaru: Very true. Updated.

jgm:
Thanks!

PR #9204, opened 2023-11-20
LaTeX template: support font fallback

lawcho:
This allows low-effort PDF generation from Unicode source.

e.g. I can write `src.md`:

```markdown
---
title: Demo Document
mainfont: Latin Modern Sans
mainfontfallback:
  - "FreeSans:"
  - "NotoColorEmoji:mode=harf"
---

Look, the empty set! → ∅ ←

Emoji work too 😎!
```

Then convert with `pandoc -s src.md -o out.pdf --pdf-engine=lualatex`

The PDF file then contains the Unicode as expected:

Screenshot from 2023-11-20 12-41-42
Notes:

* Only the `lualatex` PDF engine is supported
  * Since this patch calls `luaotfload`
  * Other PDF engines are unaffected
* `mainfont`, `sansfont`, and `monofont` can each be given a separate fallback chain
* For each `XXXfontfallback` to work, the YAML option `XXXfont` must also be set
  * I could not work out how to control `fontspec` without knowing the names of fonts

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -139,13 +139,37 @@ $endif$
 \ifPDFTeX\else
   % xetex/luatex font selection
 $if(mainfont)$
-  \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$]{$mainfont$}
+  $if(mainfontfallback)$
+    \ifLuaTeX
+      \usepackage{luaotfload}
+      \directlua{luaotfload.add_fallback("mainfontfallback",{
+        $for(mainfontfallback)$"$mainfontfallback$"$sep$,$endfor$
+      })}
+    \fi
+  $endif$
+  \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$$if(mainfontfallback)$,RawFeature={fallback=mainfontfallback}$endif$]{$mainfont$}
 $endif$
 $if(sansfont)$
-  \setsansfont[$for(sansfontoptions)$$sansfontoptions$$sep$,$endfor$]{$sansfont$}
+  $if(sansfontfallback)$
+    \ifLuaTeX
+      \usepackage{luaotfload}
+      \directlua{luaotfload.add_fallback("sansfontfallback",{
+        $for(sansfontfallback)$"$sansfontfallback$"$sep$,$endfor$
+      })}
+    \fi
+  $endif$
+  \setsansfont[$for(sansfontoptions)$$sansfontoptions$$sep$,$endfor$$if(sansfontfallback)$,RawFeature={fallback=sansfontfallback}$endif$]{$sansfont$}
 $endif$
 $if(monofont)$
-  \setmonofont[$for(monofontoptions)$$monofontoptions$$sep$,$endfor$]{$monofont$}
+  $if(monofontfallback)$
+    \ifLuaTeX
+      \usepackage{luaotfload}
+      \directlua{luaotfload.add_fallback("monofontfallback",{
+        $for(monofontfallback)$"$monofontfallback$"$sep$,$endfor$
+      })}
+    \fi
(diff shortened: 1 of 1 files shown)


data/templates/default.latex
```diff
@@ -139,13 +139,37 @@ $endif$
...
-  \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$]{$mainfont$}
+  $if(mainfontfallback)$
+    \ifLuaTeX
+      \usepackage{luaotfload}
+      \directlua{luaotfload.add_fallback("mainfontfallback",{
+        $for(mainfontfallback)$"$mainfontfallback$"$sep$,$endfor$
+      })}
+    \fi
+  $endif$
+  \setmainfont[$for(mainfontoptions)$$mainfontoptions$$sep$,$endfor$$if(mainfontfallback)$,RawFeature={fallback=mainfontfallback}$endif$]{$mainfont$}
```
jgm, line 150: Shouldn't this be `fallback=$mainfontfallback$`?  What if $mainfontfallback$ is a list?

Has this code been tested?

lawcho: `mainfontfallback` is supposed to be a list:
 * Each item is a font.
 * `luaotfload` tries each font in turn, until it finds one containing the missing character.
 * This allows setting up a *chain* of fallback fonts for greater coverage 
     (e.g. try the main font, then `StixTwoText`, then `NotoColorEmoji`, then `LastResort`)

RE testing:

* I successfully used the template in a presentation today ([slides](https://ics.uu.nl/docs/vakken/b3tc/downloads-2018/TC-03-final.pdf))
* I've used similar code several times before (via `header-includes`) for typesetting literate Agda ([documentation](https://agda.readthedocs.io/en/v2.6.4.1-rc1/tools/generating-latex.html#known-pitfalls-and-issues))
* The code closely follows the [example in the `luaotfload` docs](https://tug.ctan.org/macros/luatex/generic/luaotfload/luaotfload-latex.pdf#page=19), and the similar code on [TeX SE](https://tex.stackexchange.com/a/572220)

lawcho: RE `fallback=mainfontfallback`, the lack of `$...$` is deliberate:
* `luaotfload` requires each fallback chain to have a name
* I chose the name 'mainfontfallback' for the fallback chain set by the `mainfontfallback:` YAML option
* So the literal string `mainfontfallback` (without `$...$`) is correct here

jgm:
Thanks for the clarification. I'm not sure about this one -- only working with luatex is a drawback.  (That sort of thing tends to lead to a lot of confusion, even if it's documented.)  Looks like [font fallback isn't yet possible with xelatex](https://github.com/latex3/fontspec/issues/304).  In any case, if this were merged, the new variable would have to be documented in MANUAL.txt.

Would be glad to get feedback from others about the proposed change.

lmcnulty:
I would very much appreciate this feature. If you use a lot of unicode characters, it's hard to find one font that covers them all, and all the other solutions I've found for this are very hacky and laborious.

You already have to specify a non-default pdf engine to use unicode or non-Tex fonts in the first place, so I don't think this would be much more confusing than it is now. Maybe a warning should be shown if fallback fonts are selected but the PDF engine isn't LuaLatex?

max-heller:
I would also appreciate this feature. I've tried to implement the same functionality using `{main,sans,mono}fontoptions` and `header-includes`, but that doesn't work because the fallback must be defined earlier in the header than `header-includes` places it.

tarleb:
Adding my voice to speak out in favor of this feature. It's usefulness outweighs the downsides in my opinion.

@lawcho, do you happen to know if something similar is available for ConTeXt as well?

jgm:
I merged this and documented the relevant variables in MANUAL.txt.
Thanks!

PR #9207, opened 2023-11-21
LaTeX template: add titlegraphicoptions

nim65s:
Hi,

Including a title graphic on a beamer template might require some options, like e.g. `height=1.5cm` to ensure some image will fit at its place in a given template.

I suggest to add an option for this use case.

Without it, my current workaround is:
```yaml
header-includes:
- \titlegraphic{\includegraphics[height=1.5cm]{media/cdl23.pdf}}
```

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -2789,6 +2789,9 @@ These variables change the appearance of PDF slides using [`beamer`].
 `titlegraphic`
 :   image for title slide
 
+`titlegraphicoptions`
+:   options for title slide image
+
 ### Variables for PowerPoint
 
 These variables control the visual aspects of a slide show that
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -487,7 +487,7 @@ $if(institute)$
 \institute{$for(institute)$$institute$$sep$ \and $endfor$}
 $endif$
 $if(titlegraphic)$
-\titlegraphic{\includegraphics{$titlegraphic$}}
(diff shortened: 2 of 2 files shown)


tarleb (approved):
LGTM!

The only possible issue is that we already have a lot of variables in the default template, and we try to be very conservative when it comes to adding more.

jgm:
Maybe it should be
```
[$for(titlegraphicsoptions)$it$$sep$,$endfor$]
```
or something like that (please test).  That would allow one to use a list for `titlegraphicsoptions`, which I think is consistent with e.g. `mainfontoptions`, `biblatexoptions`, etc.

nim65s:
I had to use `standalone` in the test to check the output for title graphic, but it seems  too verbose now, and will probably fail if anything else unrelated change. 

I did try `pandoc -st beamer | grep titlegraphic`, but grep is not available.

NB: Nix CI fail with unrelated ` error: Package ‘connection-0.3.1’ in /nix/store/2arfgnb82dbh6s3h9fv05rxpr4f6z930-source/pkgs/development/haskell-modules/hackage-packages.nix:72925 is marked as broken, refusing to evaluate.`

test/command/9207-beamer-titlegraphicoptions.md
```diff
@@ -0,0 +1,223 @@
+```
+% pandoc -st beamer
+---
+titlegraphic: media/cdl23.pdf
```
jgm, line 4: Let's leave out this test.  The problem is that it would need to be updated every time there's any change in the default LaTeX template, which gets to be a hassle.

jgm:
thanks!

PR #9263, opened 2023-12-17
Make reference.docx pass validation

edwintorok:
I've used 2 docx validators:
* [docx-validator](https://github.com/devoidfury/docx-validator) with a [slightly tweaked script](https://gist.github.com/edwintorok/c1dee6fdba1932312e38dcf8020050b5) to also validate styles and settings https://gist.github.com/edwintorok
* [OOXML-Validator](https://github.com/mikeebowen/OOXML-Validator) built on Fedora39 using [this script](https://gist.github.com/edwintorok/46407d2fc8d1dd53fa5af194f3897279)

They spotted some genuine errors (an extra `>` after a close tag, I assume a typo?), but there are also a lot of annoying errors about  "Element is not expected". Took me a while to figure out that they complain about the order of XML tags!
Both the XSD and RELAXNG schemas for ISO/IEC 29500 use `xs:sequence` in a lot of places which demands a particular ordering for the XML tags. I don't know whether any application actually cares about this order, but it is better to fix them, otherwise the real validation errors are difficult to see due to all the noise.

With the changes in this PR the reference.docx now validates with both of the above validators, and at least LibreOffice and Google Docs can still open the .docx files.

Bugs fixed:
* various XML tag ordering issues in reference.docx
* duplicate XML tags that should be unique (pStyle)
* value of tblW (the format of this value changed between various revisions of the ISO standard)
* a missing required attribute on `cnfStyle`
* a trailing `>` after a `w:color` (i.e. `>>`)

The actual output of `pandoc` doesn't always validate, and I haven't looked at validation anything else than docx, but lets start by fixing the reference docx in this PR.

The changes here are 1 commit / error fixed with one testfile regeneration commit at the end, but if you prefer I can squash them into a single commit or rebase&regen at each step, whichever you prefer (to retain bisectability post merge).


jgm:
Many thanks!  This is great. I had no idea these elements had to go in a certain order.
Please let us know if there are other validation issues affecting pandoc-generated documents.

edwintorok:
Thanks, I've opened a separate issue https://github.com/jgm/pandoc/issues/9264, the reference doc validates now, but an empty doc created by pandoc does not. I guess the xml order is lost, but I don't know enough about how Haskell's XML library works (it'll probably need to sort the tags to keep them in the original order?)

Issue #9269, opened 2023-12-18
docx-validator comments and fonts


edwintorok:
**Explain the problem.**

These files fail validation with a [modified](https://gist.github.com/edwintorok/0938a14eba775e32975e2027af15c29e) docx-validator script that looks are more XML files:

pandoc/test/docx/golden/comments.docx
pandoc/test/docx/golden/custom_style_reference.docx
pandoc/test/docx/golden/track_changes_scrubbed_metadata.docx

```
pandoc/test/docx/golden/comments.docx.out:./tmp/comments-pretty.xml:11: element annotationRef: Schemas validity error : Element '{http://schemas.openxmlformats.org/wordprocessingml/2006/main}annotationRef': This element is not expected. Expected is ( {http://schemas.openxmlformats.org/wordprocessingml/2006/main}rPrChange ).
pandoc/test/docx/golden/comments.docx.out:./tmp/comments-pretty.xml:27: element annotationRef: Schemas validity error : Element '{http://schemas.openxmlformats.org/wordprocessingml/2006/main}annotationRef': This element is not expected. Expected is ( {http://schemas.openxmlformats.org/wordprocessingml/2006/main}rPrChange ).
pandoc/test/docx/golden/comments.docx.out:./tmp/comments-pretty.xml:43: element annotationRef: Schemas validity error : Element '{http://schemas.openxmlformats.org/wordprocessingml/2006/main}annotationRef': This element is not expected. Expected is ( {http://schemas.openxmlformats.org/wordprocessingml/2006/main}rPrChange ).
pandoc/test/docx/golden/comments.docx.out:./tmp/comments-pretty.xml:65: element annotationRef: Schemas validity error : Element '{http://schemas.openxmlformats.org/wordprocessingml/2006/main}annotationRef': This element is not expected. Expected is ( {http://schemas.openxmlformats.org/wordprocessingml/2006/main}rPrChange ).
pandoc/test/docx/golden/comments.docx.out:./tmp/comments-pretty.xml:81: element annotationRef: Schemas validity error : Element '{http://schemas.openxmlformats.org/wordprocessingml/2006/main}annotationRef': This element is not expected. Expected is ( {http://schemas.openxmlformats.org/wordprocessingml/2006/main}rPrChange ).
[...]

PR #9270, opened 2023-12-18
fix(docx): fix validation error in w:nsid

edwintorok:
The length here seems to refer to length in bytes (so twice as long in hex):
```
./tmp/numbering-pretty.xml:4: element nsid: Schemas validity error : Element '{http://schemas.openxmlformats.org/wordprocessingml/2006/main}nsid', attribute '{http://schemas.openxmlformats.org/wordprocessingml/2006/main}val': [facet 'length'] The value 'A990' has a length of '2'; this differs from the allowed length of '4'.
```

[This](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.nsid?view=openxml-2.8.1) also documents the longer values.

I've used [this modified script](https://gist.github.com/edwintorok/0938a14eba775e32975e2027af15c29e) to validate more XML files (there are other files beyond 'document' and 'styles' )

There are 3 files which fail validation now (with docx-validator, with OOXMLValidator there are more): https://github.com/jgm/pandoc/issues/9269

I'm not sure whether the actual numbering value here is meaningful, so I just did the minimal change to pad and fix up the length, although a nicer way to solve this would be to keep the numbering id as integer and use printf with %08x maybe?


edwintorok:
Also this PR does actually change some ids in the styles that pandoc outputs, should this be mentioned in the changelog (e.g. could a Lua or other filter rely on the actual `A990` value?)

jgm:
That took us from 185 errors to 150 in `make validate-docx-golden-tests`. Progress!

PR #9271, opened 2023-12-18
fix(docx): fix validation error on w:annotationRef

edwintorok:
annotationRef is not valid for `w:rPr`, only for `w:r` according to wml.xsd.

See https://github.com/jgm/pandoc/issues/9269.

Manually running this on comments.native does seem to fix the problem, but I can't figure out why this doesn't trigger a regeneration of the golden files, are changes in comments.xml ignored somewhere?


edwintorok:
Got it: compareXml doesn't compare the length of lists, fix coming up!

edwintorok:
Ok, down to 4 validation failures now.

Makefile
```diff
@@ -60,7 +60,7 @@ quick-stack: ## unoptimized build and tests with stack
 	  --system-ghc --flag 'pandoc:embed_data_files' \
 	  --fast \
 	  --test \
-	  --test-arguments='-j --hide-successes --ansi-tricks=false $(TESTARGS)'
+	  --test-arguments='-j32 --hide-successes --ansi-tricks=false $(TESTARGS)'
```
edwintorok, line 63: I keep forgetting to remove this, but apparently my version of cabal/stack/ghc/etc. doesn't accept a `-j` here, it insists on having a number.

edwintorok: I changed this to `-j4`, it is what other similar flags in this Makefile use, and moved it into a separate commit

PR #9272, opened 2023-12-19
docx: fix more validation errors noticed by OOXMLValidator

edwintorok:
It is a bit more strict than the .xsd from docx-validator, and it does some checks that the .xsd one cannot (e.g. it found a dangling reference from `endnotePr` to a part of the document that didn't exist).

I don't know why xmllint didn't complain about ordering of `w:b` vs `w:i` (maybe because they're empty tags?).

I added a makefile rule similar to the docx-validator one, except this tool can recursively check all files, so I used that form.
It requires dotnet, but in total cloning, building and running the tool takes ~2s:
```
time make validate-docx-golden-tests2
[...]
make validate-docx-golden-tests2  1.99s user 0.27s system 81% cpu 2.761 total
```

Only one validation error remains now: https://github.com/jgm/pandoc/issues/9273


jgm:
Terrific!

jgm:
Can you figure out how we install the necessary tools in CI to run this makefile target?

edwintorok:
Done, also fixed the last validation error https://github.com/jgm/pandoc/pull/9277

PR #9277, opened 2023-12-19
docx: fix validation error reported by OOXMLValidator

edwintorok:
And add a CI rule that runs xmllint + the dotnet OOXMLValidator

You can see a successful run of that here (completes in ~42s in that case): https://github.com/edwintorok/pandoc/actions/runs/7267632699

Everything passes validation now :)


edwintorok:
Turns out I still remember enough Haskell to figure out how to sort a list of XML elements, but please feel free to nitpick at code style issues if any.
I usually write code in OCaml, not Haskell, so there might be a more idiomatic way to implement this.

jgm:
Super! Thank you.

jgm:
@edwintorok do you see how to fix the errors here?
https://github.com/jgm/pandoc/actions/runs/7772323364/job/21194608607#step:5:33

edwintorok:
@jgm opened a PR and the docx step is green now https://github.com/jgm/pandoc/actions/runs/7790229330/job/21243632727?pr=9414

PR #9291, opened 2023-12-27
ci(commit-validation): use git to walk commits

edwintorok:
The PR event doesn't include the actual commits, just a count and a URL to fetch it.
But we've already checked out the entire git history, so we might as well use `git` itself to walk the commit history.
(Using the remote commits URL would be useful only if we'd want to do a shallow clone because we have too many commits)

We can revisit this once pandoc's commit history grows large enough that this takes too long. For now it should unblock the CI.


jgm:
Thank you!

PR #9340, opened 2024-01-16
Add `readMan` to `Text.Pandoc.Readers` exports

georgestagg:
It looks like `readMan` is missing in the list of exports of the module `Text.Pandoc.Readers`. I couldn't see why, so I've added it here.


jgm:
Thank you!

PR #9343, opened 2024-01-17
Lua image size

tarleb:
Adds a new Lua module `pandoc.image` that provides basic image querying functions.


tarleb:
This is based on suggestions by @cscheid.

jgm:
This looks fine to me -- do you want to make it a non-draft so I can merge it?

My main reservation about such changes is just a general one -- it's more code that has to be changed if we ever change the ImageSize API. But if there are sufficient tests, we will be reminded about the need to make these changes if/when necessary.

tarleb:
I've fixed the docs and added another test. Should be good to go now.

PR #9359, opened 2024-01-23
Add @var support in Texinfo Writer #8534

### LINKED ISSUE #8534: HTML → texinfo: `<var>` within `<code>` not taken into account
[version 2.19.2]

Consider the following minimum HTML input.

```html
<html>
foo <code>bar <var>baz</var> bar</code> foo
</html>
```

If processed with `pandoc -f html -t texinfo` I get the following output.

```texinfo
foo @code{bar }@code{baz}@code{ bar} foo
```

This should rather be

```texinfo
foo @code{bar @var{baz} bar} foo
```

instead.  Both Firefox and Chrome display 'baz' in code italic for the above HTML code, and Texinfo actually supports `@var` in `@code` just fine, too.

gok99:
Fixes #8534

Implements @jgm's suggestion, checking if the "variable" class is contained in the attributes, and without consolidation of adjacent code elements.

### DIFF
--- a/src/Text/Pandoc/Writers/Texinfo.hs
+++ b/src/Text/Pandoc/Writers/Texinfo.hs
@@ -446,6 +446,9 @@ inlineToTexinfo (Subscript lst) = do
 inlineToTexinfo (SmallCaps lst) =
   inCmd "sc" <$> inlineListToTexinfo lst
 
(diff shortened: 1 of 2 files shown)


jgm:
Looks good, thanks!

PR #9361, opened 2024-01-23
ConTeXt template: support font fallback

lawcho:
Follow up to https://github.com/jgm/pandoc/pull/9204#issuecomment-1900741386.

You write:

```markdown
---
title: Demo Document
mainfontfallback:
  - EmojiOne Color
  - OpenSymbol
  - XITS Math
---

This empty set comes from OpenSymbol → ∅ ←

But these [stange operators](https://tex.stackexchange.com/a/397674)
aren't in OpenSymbol, so are fetched from XITS Math:

X × ⎰Y ⨿ Z⎱ → ⎰X × Y⎱ ⨿ ⎰X × Z⎱.

Emoji fallback doesn't work 🙁 .
```

You run:

```bash
pandoc src.md -o out.pdf --pdf-engine=context
```

You get:

Screenshot from 2024-01-23 17-19-38

Notes:

* You don't need to set mainfont (unlike in the `LaTeX` template)
* ConTeXt provdes two competing commands for fallback setup:
  * [`\definefontfallback`](https://wiki.contextgarden.net/Command/definefontfallback) requires a later call to `\definefontsynonym`
  * 👉 [`\definefallbackfamily`](https://wiki.contextgarden.net/Command/definefallbackfamily) allows setting fallbacks without knowing the default font
* I couldn't get emoji fallback to work
  * `NotoColorEmoji` fails to load (with a `loca table not foundmtx-context` error)
  * `EmojiOne Color` loads but doesn't substitute anything
* Testing `ConTeXt` with new fonts is fiddly
  * Must run `mtxrun --script fonts --reload`
* More tests needed (cc @tarleb)

### DIFF
--- a/data/templates/default.context
+++ b/data/templates/default.context
@@ -59,6 +59,17 @@ $endif$
 
 \setupbodyfontenvironment[default][em=italic] % use italic as em, not slanted
 
+% Set up font fallbacks
+$for(mainfontfallback)$
+\definefallbackfamily[mainface][rm][$mainfontfallback$][range=0x0000-0xFFFF, check=yes, force=no]
+$endfor$
+$for(sansfontfallback)$
+\definefallbackfamily[mainface][ss][$sansfontfallback$][range=0x0000-0xFFFF, check=yes, force=no]
+$endfor$
+$for(monofontfallback)$
+\definefallbackfamily[mainface][tt][$monofontfallback$][range=0x0000-0xFFFF, check=yes, force=no]
+$endfor$
+
 \definefallbackfamily[mainface][rm][CMU Serif][preset=range:greek, force=yes]
 \definefontfamily[mainface][rm][$if(mainfont)$$mainfont$$else$Latin Modern Roman$endif$]
 \definefontfamily[mainface][mm][$if(mathfont)$$mathfont$$else$Latin Modern Math$endif$]


tarleb:
Thanks for this. I tried it and didn't succeed in getting emojis working either, and it [seems](https://www.mail-archive.com/ntg-context@ntg.nl/msg102621.html) that this is not an easy issue to solve. Apart from that it seems to work really well.

You can avoid the test failures by using a template comment instead of a TeX comment (syntax: `$-- comment text`). LGTM once that's fixed.

jgm:
MANUAL.txt should be updated as well.

lawcho:
> You can avoid the test failures by using a template comment instead of a TeX comment (syntax: `$-- comment text`).

I tried this (in [9384ac3](https://github.com/jgm/pandoc/pull/9361/commits/9384ac3af81f32fbec698b0c8df357b97587e2b3)), but the CI still fails.

Probably the comment isn't the root cause, since there are `% Ordinary LaTeX comments` nearby in the same file.

Could somebody else please fix this, I don't know what I'm doing with pandoc's CI.

tarleb:
The error comes from the empty line that's added in the output. Removing line 72 should fix the test.

Issue #9325, opened 2024-01-10
Don't emit unnecessary classes in HTML tables


ThomasSoeiro:
It is currently not possible to prevent `pandoc` from adding attributes to the HTML output from a Markdown input (e.g. `.header`, `.odd`, `.even` in the ReprEx below). It is only possible to [drop attributes using filters](https://stackoverflow.com/q/53403041).

Since both the CommonMark and GitHub Flavored Markdown specs do not mention default attributes in HTML output, shouldn't this be opt-in by default? Or possible to opt-out at least?

# ReprEx

Using e.g. this input:

```
| foo | bar |
| --- | --- |
| baz | bim |
| baz | bim |
```

And converting to HTML using `pandoc --from gfm --to html5`, we get:

```html
<table>
<thead>
<tr class="header">
<th>foo</th>
<th>bar</th>
</tr>
</thead>
<tbody>
<tr class="odd">
<td>baz</td>
<td>bim</td>
</tr>
<tr class="even">
<td>baz</td>
<td>bim</td>
</tr>
</tbody>
</table>
```

jgm:
These are harmless; just don't add CSS rules that do things with them.

I don't think adding the option to avoid these is worth the increase in complexity.

jgm:
Agreed -- we could use something like table-header, even-row, odd-row.
Of course, it would be a backwards-incompatible change, so I'm not sure it's a good idea.

ThomasSoeiro:
I just noticed the following in the html output:

```css
    header {
      margin-bottom: 4em;
      text-align: center;
    }
```

Shouldn't `header` be `.header`?

jgm:
> My concern was for the sake of dropping "unecessary" classes to prevent name clashes since we can easely select .header using thead and .odd/even using variations of tbody tr:nth-child(2n).

This is true, though it wasn't in earlier versions of pandoc (when nth-child wasn't supported and we didn't put the header in thead!).

I think it might be worth stopping using these classes, and using the alternative you suggest instead.

tarleb:
Could this be a "good first issue"?

jgm:
Yes, it would be an easy one -- just have to change the HTML writer, the styles.html template, and some tests I think.

ThomasSoeiro:
Could you point me to the HTML writer please?
(I'll have a look but I don't know Pandoc internals nor haskell...)

jgm:
[...]

PR #9376, opened 2024-01-26
Don't emit unnecessary classes in HTML tables (#9325)

ThomasSoeiro:
Here is a try for #9325 according to @jgm guidance :

- The change in the writer needs to be refined but I guess this is the idea.
- I found no use of `header`/`odd`/`even` classes in `data/templates/styles.html`.
- I removed `header`/`odd`/`even` accross the tests; I'm not sure about changes in `update test/command/*.md` so I put it in a separate commit.

I also found occurence of `odd`/`even` in:

- `pandoc-lua-engine/test/sample.lua` and `pandoc-lua-engine/test/tables.custom`
- and `/src/Text/Pandoc/Writers/Textile.hs` and `test/tables.textile`

Do they need updating too?

_PS: This is my first pull request. It is only WIP but I hope it helps. Otherwise, feel free to close!_

### DIFF
--- a/src/Text/Pandoc/Writers/HTML.hs
+++ b/src/Text/Pandoc/Writers/HTML.hs
@@ -1253,12 +1253,6 @@ tableRowToHtml :: PandocMonad m
                -> TableRow
                -> StateT WriterState m Html
 tableRowToHtml opts (TableRow tblpart attr rownum rowhead rowbody) = do
-  let rowclass = case rownum of
-        Ann.RowNumber x | x `rem` 2 == 1   -> "odd"
-        _               | tblpart /= Thead -> "even"
-        _                                  -> "header"
-  let attr' = case attr of
-                (id', classes, rest) -> (id', rowclass:classes, rest)
   let celltype = case tblpart of
                    Thead -> HeaderCell
                    _     -> BodyCell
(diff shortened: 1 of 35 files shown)


jgm:
And yes, I'd say those other things should also be updated for consistency.

ThomasSoeiro:
hi @jgm!

> (...) But this all looks good to me, I think you could "undraft" it.

Done, thank you for having a look!

> And yes, I'd say those other things should also be updated for consistency.

Sorry, I won't be able to take care of it for several weeks.

jgm:
Thanks, I'll just merge this and take care of the other thing separately.

Malcolmnixon:
The removal of odd/even is a bit of a breaking change for us. We were relying on that to produce pretty HTML tables with alternating lines.

It may be possible to use the CSS `nth-child(even)` trick, but that may not work when the tables are split across pages using something like pagedjs.

tarleb:
@Malcolmnixon: [Lua filter](https://pandoc.org/lua-filters) to restore the old behavior:

``` lua
local function add_odd_even (rows)
  for rownum, row in ipairs(rows) do
    row.classes:insert((rownum % 2) == 0 and 'even' or 'odd')
  end
  return rows
end

function Table (tbl)
  add_odd_even(tbl.head.rows)
  for _, tblbody in ipairs(tbl.bodies) do
    add_odd_even(tblbody.body)
  end
  add_odd_even(tbl.foot.rows)
  return tbl
end
```

Issue #5706, opened 2019-08-24
Add ability to customize tables in powerpoint output 


Neurrone:
Using Pandoc 2.7.3

1. Save the following as tables.md:

```markdown
% test slides for table rendering

## Sprint backlog - Done!

story | status
----------|-:
Evaluate third-party integration | done
this is a very very very very very very very very very long story | Done
```

2. Generate the presentation with `pandoc "tables.md" -o "tables.pptx"`

I expect to be able to customize the font used by the table, by using a reference template, but this seems to be impossible. 

The number of dashes under the column headers should cause the first column to take most of the space, but each column is currently taking up equal width.

mb21:
So you're using [pipe tables](https://pandoc.org/MANUAL.html#extension-pipe_tables), then the column widths only go into effect when the line is long enough. Try:

```
story | status
----------|-:
Evaluate third-party integration | done
this is a very very very very very very very very very long stor asdf asdf asdf asf asdf y | Done

```

But I don't know about the powerpoint writer's ability to customize tables...

Neurrone:
@mb21 in case I don't have text that would be long enough to force it, which tables would be better for what I want to do? The intent is basically for the second column to only take up as much space as needed (in this case just for 4 characters), with the rest of the space allocated for the left column. In my actual use of this, right now the table is a bit too long to fit into one slide, but would fit if it allocated more space to the left column.

mb21:
I think both multiline_tables and grid_tables always compute widths, see https://pandoc.org/MANUAL.html#tables

Neurrone:
I can't get widths to work at all, even when using multiline tables which are supposed to compute widths. This sounds like a bug.

Neurrone:
CC: @jkr

P-N-L:
I ran into this as well. I am using pipe tables in Markdown, and the columns (including the `-` separators under the headers) are different sizes, but the widths in the PowerPoint output are the same size.

PR #9392, opened 2024-01-31
Using internal column widths in pptx writer tables

tomasdahlqvist:
The table writer used to only divide all available width evenly for all columns. In this update the code uses the incoming widths if they are available. If they are not set the earlier even distribution is used. Some of the golden templates are adjusted slightly because of different rounding when using the new calculation model.


src/Text/Pandoc/Writers/Powerpoint/Output.hs
```diff
@@ -1411,19 +1411,24 @@ getDefaultTableStyle = do
...
+  let colWidths = if any (== 0.0) widths
+                  then if null hdrCells
+                      then case rows of
                          r : _ | not (null r) -> replicate (length r) $
                                                  tableWidth `div` toInteger (length r)
                          -- satisfy the compiler. This is the same as
                          -- saying that rows is empty, but the compiler
                          -- won't understand that `[]` exhausts the
                          -- alternatives.
                          _ -> []
```
jgm, line 1427: Maybe more elegant to do
```
  r@(_:_) : _ ->
  [] ->
```
I think the compiler would be happy with that.

tomasdahlqvist: Almost. I also needed to match empty list in list.
Sorry about the additional checkin; I had a hard time understanding when the widths actually were set so I did some debug printouts which are now removed.

tomasdahlqvist:
I have forgotten to add that this fixes https://github.com/jgm/pandoc/issues/5706

jgm:
Great, thanks!

PR #9400, opened 2024-02-01
EPUB accessibility additions

Iacobus1983:
AccessModeSufficient and AccessibilitySummary are added.

These metadata are optional according to the 
[EPUB Accessibility 1.1](https://www.w3.org/TR/epub-a11y-11/#sec-disc-package) 
standard. Though, ACE Daisy checker flags ebooks for missing them
(cf https://github.com/daisy/ace/issues/389 and https://github.com/daisy/ace/issues/400)

With this commit:
- AccessModeSufficient is defaulted to "textual".
- AccessibilitySummary can be included via the metadata block.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -6848,6 +6848,10 @@ The following fields are recognized:
   ~ An array of strings ([schema][accessibility schema]).
     Defaults to `["textual"]`.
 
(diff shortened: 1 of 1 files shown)


jgm:
thanks!

PR #9417, opened 2024-02-06
Manual: clarify scope of option --resource-path

cagix:
Add a note to the documentation of the `--resource-path` option that it is only used by Pandoc itself. This should clarify its scope for users, even if the text is listed in the "General Writer Options" section. 

The relevant discussion on this was at https://github.com/jgm/pandoc/discussions/9360#discussioncomment-8373190.


cagix:
updated commit message to comply with suggested line length

jgm:
Thanks - I went with somewhat different wording.

cagix:
> Thanks - I went with somewhat different wording.

that's much better, thx!

PR #9593, opened 2024-03-20
Typst writer: support Typst 0.11 table features.

### LINKED ISSUE #9588: Support new typst table features (Typst 0.11)
rowspans, colspans, header, footer, column widths, alignments:
https://typst.app/docs/guides/table-guide/

jgm:
- colspans
- rowspans
- cell alignment overrides
- relative column widths
- header and footer
- multiple table bodies with intermediate headers

Row heads are not yet supported.

The default typst template has been modified so that tables don't have lines by default. As is standard with pandoc, we only add a line under a header or over a footer. However, a different default stroke pattern can easily be added in a template.

Closes #9588.

### DIFF
--- a/data/templates/default.typst
+++ b/data/templates/default.typst
@@ -10,7 +10,8 @@ $definitions.typst()$
 }
 
 #set table(
-  inset: 6pt
+  inset: 6pt,
+  stroke: none
 )
 
 $if(template)$
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -326,6 +326,7 @@ extra-source-files:
                  test/tables/*.html4
                  test/tables/*.html5
                  test/tables/*.latex
+                 test/tables/*.typst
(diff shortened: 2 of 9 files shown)


gordonwoodhull:
Nice! I did basic testing of colspans, rowspans, cell alignment, header and footer via Quarto, mostly by including HTML. 

Couldn't figure out how to test intermediate headers - guess I am not familiar with this.

We should remove the table lines and keep only the header/footer lines in Quarto as well.

jgm:
I think the main question I have at this point is whether there's any natural way to support "row heads" in typst.  I don't see one, so maybe I'll leave this unimplemented?

gordonwoodhull:
I haven’t heard of “row heads” nor am I finding anything in a brief search

jgm:
https://stackoverflow.com/questions/28047264/tables-with-row-and-column-headers

Issue #9626, opened 2024-04-03
Docx reader: bookmark in header not working when last parPart was a bookmark


mbrackeantidot:
**Explain the problem.**
When transforming this document: [bookmark_simplified.docx](https://github.com/jgm/pandoc/files/14850165/bookmark_simplified.docx) with pandoc using `pandoc bookmark_simplified.docx`, the link at the start of the file is not correctly transformed.

*Expected ouptut*:
```
<p><a href="#referenced-title">A link to the bookmark</a></p>
<p><strong>Accomplices:</strong></p>
<p>There are bookmarks around here.</p>
<h1 id="referenced-title">Referenced title</h1>
<p>Content</p>
```

*Actual output*:
```
<p><a href="#_APPENDIX_5:_FINANCIAL">A link to the bookmark</a></p>
<p><strong>Accomplices:</strong></p>
<p>There are bookmarks around here.</p>
<h1 id="referenced-title">Referenced title</h1>
<p>Content</p>
```

I found out that this is because the last paragraph part before the header is a bookmark. In this case, `parPartToInlines'` doesn't create a span for the bookmark because there is a bookmark immediately before the header (represented by `docxImmedPrevAnchor`) and doesn't add the bookmark to the anchor map `docxAnchorMap` because it's in a header, which is handled by `makeHeaderAnchor'`. But `makeHeaderAnchor'` assumes that a span with an anchor was added by `parPartToInlines'`.

To verify my understanding of the problem, I implemented a fix and tested it manually. If you're curious, the code can be found here: https://github.com/jgm/pandoc/commit/09fc2040fea5901df2c00255660e742ada3d76f3

I would like to create a proper commit containing that fix and make a pull request. Do I have your blessing?

**Pandoc version?**
Latest development version, compiled from source on macOS. Older versions on Linux and macOS had the same bug.

jgm:
> I would like to create a proper commit containing that fix and make a pull request. Do I have your blessing?

Go for it!

mbrackeantidot:
Thanks!

tarleb:
I believe that the merged PR #9626 fixed this.

PR #9629, opened 2024-04-04
Docx reader: fix anchor in header after anchor (#9626)

mbrackeantidot:
When the last parPart before a header was a bookmark, no span with an anchor was added for a bookmark in the header. But the function that adds header anchors to the anchor map, needs a span with an anchor. So this commit adds that span.


tarleb (approved):
LGTM!

jgm:
thanks!

PR #9648, opened 2024-04-10
Typst property output

gordonwoodhull:
Here is my work in progress on Typst property output. Supercedes #9623.

https://github.com/gordonwoodhull/pandoc/tree/typst-property-output

It handles 8 cases of attribute-bearing AST nodes:
* span text (no element properties because `text()` is the closest element to span)
* div and div text
* table and table text
* td and td text (3 cases)

Now that I have tests I would be comfortable implementing the rest, but I haven't needed them yet. Thought I'd check in to ask if I'm on track here. All suggestions are welcome!

I think the code is more idiomatic Haskell (but not perfect) and the usage is terse (except for the align case, see below). Some details of whitespace/line breaks probably need more work.

There aren't row elements in Typst; tr attributes would have to be underlaid on cell properties.

#### align is messy

There is one case that is very bad - the Typst reader extracts (horizontal) `align`, but this needs to be coalesced with `vertical-align` for the Typst cell `align`. An alternative design would be to not remove the style in the reader, and allow Lua to override the whole property. Technically I should also parse `typst:align` as addition and then dedupe, because Typst will error if there are duplicate terms in the align expression, but Lua only sees `vertical-align` currently, so it works out.

### DIFF
--- a/src/Text/Pandoc/Writers/Typst.hs
+++ b/src/Text/Pandoc/Writers/Typst.hs
@@ -88,6 +88,18 @@ pandocToTypst options (Pandoc meta blocks) = do
        Nothing  -> main
        Just tpl -> renderTemplate tpl context
 
+pickTypstAttrs :: [(Text, Text)] -> [(Text, Text)]
+pickTypstAttrs = filter (T.isPrefixOf "typst:element:" . fst)
+
+pickTypstTextAttrs :: [(Text, Text)] -> [(Text, Text)]
+pickTypstTextAttrs = filter (T.isPrefixOf "typst:text:" . fst)
+
+formatAttrs :: [(Text, Text)] -> [Text]
+formatAttrs =
+  map (\(k,v) -> let prop = T.splitOn (T.pack ":") k !! 2 in
+        prop <> ": " <> v)
+
+
 blocksToTypst :: PandocMonad m => [Block] -> TW m (Doc Text)
 blocksToTypst blocks = vcat <$> mapM blockToTypst blocks
 
@@ -163,7 +175,7 @@ blockToTypst block =
                    else vsep items') $$ blankline
     DefinitionList items ->
       ($$ blankline) . vsep <$> mapM defListItemToTypst items
-    Table (ident,_,_) (Caption _ caption) colspecs thead tbodies tfoot -> do
+    Table (ident,_,tabkvs) (Caption _ caption) colspecs thead tbodies tfoot -> do
       let lab = toLabel FreestandingLabel ident
       capt' <- if null caption
                   then return mempty
@@ -185,7 +197,8 @@ blockToTypst block =
           formatalign AlignCenter = "center,"
           formatalign AlignDefault = "auto,"
       let alignarray = parens $ mconcat $ map formatalign aligns
-      let fromCell (Cell _attr alignment rowspan colspan bs) = do
+
+      let fromCell (Cell (_,_,kvs) alignment rowspan colspan bs) = do
             let cellattrs =
                   (case alignment of
                      AlignDefault -> []
@@ -197,7 +210,8 @@ blockToTypst block =
                      RowSpan n -> [ "rowspan: " <> tshow n ]) ++
                   (case colspan of
                      ColSpan 1 -> []
-                     ColSpan n -> [ "colspan: " <> tshow n ])
+                     ColSpan n -> [ "colspan: " <> tshow n ]) ++
+                  formatAttrs (pickTypstAttrs kvs)
             cellContents <- blocksToTypst bs
             pure $ if null cellattrs
                       then brackets cellContents
@@ -223,6 +237,10 @@ blockToTypst block =
             hrows <- mapM fromRow headRows
             brows <- mapM fromRow bodyRows
             pure $ vcat (hrows ++ ["table.hline()," | not (null hrows)] ++ brows)
+      let (textstart, textend) =
+            (case formatAttrs $ pickTypstTextAttrs tabkvs of
+              [] -> ("", "")
+              tkvs -> ("#text" <> parens (literal (T.intercalate ", " tkvs)) <> "[", "]"))
       header <- fromHead thead
       footer <- fromFoot tfoot
       body <- vcat <$> mapM fromTableBody tbodies
@@ -230,7 +248,7 @@ blockToTypst block =
         "#figure("
         $$
         nest 2
-         ("align(center)[#table("
+         ("align(center)[" <> textstart <> "#table("
           $$ nest 2
              (  "columns: " <> columns <> ","
              $$ "align: " <> alignarray <> ","
@@ -238,7 +256,7 @@ blockToTypst block =
              $$ body
              $$ footer
              )
-          $$ ")]"
+          $$ ")]" <> textend
           $$ capt'
           $$ ", kind: table"
           $$ ")")
@@ -261,10 +279,13 @@ blockToTypst block =
                           $$ ")" $$ lab $$ blankline
     Div (ident,_,_) (Header lev ("",cls,kvs) ils:rest) ->
       blocksToTypst (Header lev (ident,cls,kvs) ils:rest)
-    Div (ident,_,_) blocks -> do
+    Div (ident,_,kvs) blocks -> do
       let lab = toLabel FreestandingLabel ident
+      let props = case formatAttrs $ pickTypstAttrs kvs of
+                    [] -> ""
+                    tkvs -> parens $ literal (T.intercalate ", " tkvs)
       contents <- blocksToTypst blocks
-      return $ "#block[" $$ contents $$ ("]" <+> lab)
+      return $ "#block" <> props <> "[" $$ contents $$ ("]" <+> lab)
 
 defListItemToTypst :: PandocMonad m => ([Inline], [[Block]]) -> TW m (Doc Text)
 defListItemToTypst (term, defns) = do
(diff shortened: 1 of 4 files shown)


src/Text/Pandoc/Writers/Typst.hs
```diff
@@ -88,6 +88,52 @@ pandocToTypst options (Pandoc meta blocks) = do
        Nothing  -> main
        Just tpl -> renderTemplate tpl context
 
+pickTypstAttrs :: [(Text, Text)] -> ([(Text, Text)],[(Text, Text)])
+pickTypstAttrs =
+  partition (not . T.isPrefixOf "typst:text:" . fst) . filter (T.isPrefixOf "typst:" . fst)
+
+formatTypstProps :: [(Text, Text)] -> [Text]
+formatTypstProps =
+  map (\(k,v) -> last (T.splitOn (T.pack ":") k) <> ": " <> v)
```
jgm, line 97: Avoid `last`, as it's a partial function.
Also you don't need `T.pack ":"`; `":"` will work by itself because we have OverloadedStrings.

test/Tests/Writers/Typst.hs
```diff
@@ -0,0 +1,178 @@
...
+      , ""
+      , "    ], [Bar"
+      , ""
+      , "    ],"
+      , "  )]"
+      , "  , kind: table"
+      , "  )"
+      ]
+    ]
+  ]
```
jgm, line 178: I'd suggest just using a "command test" (see `test/commands/*.md`).
They are easier to create and much easier to maintain (since we can just do `make test TESTARGS=--accept` to update the expected output instead of doing it manually.

src/Text/Pandoc/Writers/Typst.hs
```diff
@@ -88,6 +88,52 @@ pandocToTypst options (Pandoc meta blocks) = do
...
+toTypstSetTextUnchecked typstTextAttrs =
+  "#set text" <> parens (toTypstProps typstTextAttrs) <> "; " -- newline?
+
+-- there is probably a more idiomatic way to do this
+ifNotEmpty ::  ([a2] -> Doc Text) -> [a2] -> Doc Text
+ifNotEmpty f list =
+  case list of
+    [] -> ""
+    _ -> f list
+
```
jgm, line 121: I think it would be more elegant to use a fold.  E.g. `foldl' addTypstProperty mempty props`.
If the list `props` is empty, you get `mempty`, so you wouldn't need this ugly function.
It should be possible to make this code simpler.

src/Text/Pandoc/Writers/Typst.hs
gordonwoodhull: Good points. I had to improve validation to drop keys like `typst:foo:bar`, so I took the opportunity to pull the last element at input time and not worry about invalid input during output.

src/Text/Pandoc/Writers/Typst.hs
gordonwoodhull: It's a little bit more complicated, because we have this conditional text we want to add around the lists, like we want to emit either `#set text(prop:value); ` or the empty string at the top of a block where we are applying `typst:text:*` attributes. We don't want to emit a useless `#set text(); ` if that's even valid.

And likewise everywhere else: if a block has no attributes, we want to emit `#block[content]` and not `#block()[content]`

@cscheid and I looked at it together and agreed that most clear thing to do is not to try to abstract it, but just write functions that both check for the empty list and generate text. It's about half as long now.

test/Tests/Writers/Typst.hs
gordonwoodhull: Aha, much simpler to write tests that way!

I threw in negative cases: elements that don't have `typst:*` attributes should not affect the output. This was covered by previous tests but worth it to be somewhat complete.

gordonwoodhull:
Okay, I think I have addressed your concerns, as well as adding solid validation of `typst:*` keys.

Glad to make any other changes! Should I add documentation somewhere?

I think I'd prefer to wait to implement output of properties from other elements until we get more experience, unless you have any requests.

`<span>`, `<div>`, `<table>` and `<td>` are the most important elements as far as Quarto is concerned.

I can do another PR later with other elements as our users inevitably test the boundaries of this, or we find more uses for Typst properties.

Thanks for all your helpful comments so far.

jgm:
As for documentation, one possibility would be to have a separate document along the lines of
https://pandoc.org/org.html
whose source is `doc/org.md`.  Alternatively it could be a section in the manual like the current section for EPUBs, but my thinking is that this concerns a pretty niche feature that is only accessible through filters, so it probably is best to document it separately...

src/Text/Pandoc/Writers/Typst.hs
```diff
@@ -88,6 +90,42 @@ pandocToTypst options (Pandoc meta blocks) = do
        Nothing  -> main
        Just tpl -> renderTemplate tpl context
 
+pickTypstAttrs :: [(Text, Text)] -> ([(Text, Text)],[(Text, Text)])
+pickTypstAttrs =
+  partitionEithers . catMaybes . map (\(k,v) ->
+    let spl = T.splitOn ":" k in
+    if length(spl) < 2 || length(spl) > 3
+      || head spl /= "typst"
+      || (length(spl) == 3 && spl !! 1 /= "text")
```
jgm, line 99: We like to avoid using partial functions like `head`. I think it's cleaner to use pattern matching for this kind of thing. Could you do something like
```hs

case T.splitOn ":" k of
  "typst":"text":x -> Right (x,v)
  "typst":x -> Left (x, v)
  _ -> Nothing
```

Second thought: instead of using `partitionEithers . catMaybes . map`, why not just use a fold that directly constructs the tuple?  This way you also avoid needing to use Maybe or Either.

```hs
pickTypstAttrs = foldr go ([],[])
  where
    go (k,v) (eltProps, textProps) =
      case T.splitOn ":" k of
        "typst":"text":x -> (eltProps, (x,v):textProps)
        "typst":x -> ((x,v):eltProps, textProps)
        _ -> (eltProps, textProps)
```
Or even more compactly with Data.Bifunctor.(first,second)
```hs
pickTypstAttrs = foldr go ([],[])
  where
    go (k,v)
      case T.splitOn ":" k of
        "typst":"text":x -> second ((x,v):)
        "typst":x -> first ((x,v):)
        _ -> id
```

src/Text/Pandoc/Writers/Typst.hs
```diff
@@ -88,6 +90,42 @@ pandocToTypst options (Pandoc meta blocks) = do
...
+       else Just $ Right (spl !! 2, v))
+
+formatTypstProps :: [(Text, Text)] -> [Text]
+formatTypstProps =
+  map (\(k,v) -> k <> ": " <> v)
+
+toTypstPropsListSep :: [(Text, Text)] -> Doc Text
+toTypstPropsListSep typstAttrs =
+   literal (T.intercalate ", " $ formatTypstProps typstAttrs)
+
```
jgm, line 112: With this approach the attribute lists will be unbreakable.
That's probably not what we want.  Better:
```hs
formatTypstProps = map (\(k,v) -> literal (k <> ": " <> v))
toTypstPropsListSep = hsep . intersperse "," . formatTypstProps
```

src/Text/Pandoc/Writers/Typst.hs
```diff
@@ -88,6 +90,42 @@ pandocToTypst options (Pandoc meta blocks) = do
...
+toTypstPropsListTerm [] = ""
+toTypstPropsListTerm typstAttrs = toTypstPropsListSep typstAttrs <> ","
+
+toTypstPropsListParens :: [(Text, Text)] -> Doc Text
+toTypstPropsListParens [] = ""
+toTypstPropsListParens typstAttrs = parens $ toTypstPropsListSep typstAttrs
+
+toTypstTextElement :: [(Text, Text)] -> Doc Text -> Doc Text
+toTypstTextElement [] content = content
+toTypstTextElement typstTextAttrs content = "#text" <> parens (toTypstPropsListSep typstTextAttrs) <> brackets content
```
jgm, line 123: Could simplify by using `toTypstPropsListParens` here instead of the spelled out version

src/Text/Pandoc/Writers/Typst.hs
gordonwoodhull: This is very nice!

All five of my logical conditions covered by three patterns, and the meaning is very clear with the bifunctor approach.

src/Text/Pandoc/Writers/Typst.hs
gordonwoodhull: Great, knew I was missing something about line breaks. Thanks!

gordonwoodhull:
Excellent feedback - I think it is very clean and simple now.

Working on docs.

iandol:
Amazingly useful contribution @gordonwoodhull!!! 🥳

doc/typst-property-output.md
```diff
@@ -0,0 +1,155 @@
...
+Given the HTML input
+
+```html
+<p>Here is some <span style="color:orange">orange text</span>.</p>
+```
+
+the command
+
+```sh
+quarto pandoc -f html -t typst --lua-filter ./typst-property-example.lua
```
jgm, line 146: Since this is pandoc's documentation, not quarto, shouldn't this just be `pandoc`?

gordonwoodhull: Yep!

gordonwoodhull:
I fixed the example, did a few more copy edits, and squashed everything into one commit. 

I think this is ready! Please let me know if you want any further changes.

test/command/typst-property-output.md
```diff
@@ -0,0 +1,153 @@
...
+    columns: 2,
+    align: (auto,center,),
+    [A], table.cell(align: center)[#set text(fill: maroon); B],
+  )]
+  , kind: table
+  )
+```
+
+```
+% pandoc -f html -t typst
```
jgm, line 140: Is there a difference between this test and the one at line 76?

jgm: Did you mean to have a test of an attribute on `table`?

gordonwoodhull: Hmm, I guess I added negative tests for both table and cell without realizing they are the same. Removed.

gordonwoodhull:
I found that I had failed to update the ugly special case for `align` when I improved validation, so I pushed a fix and associated test.

Also I noticed that we can use a set-text rule instead of text element for table, since the table is inside the figure & align scopes and the set-text rule won't leak out.

jgm:
I think it's good to go! Thanks.

gordonwoodhull:
Thank you!

gordonwoodhull:
This is working great and we'd like to include "Typst CSS" in Quarto 1.5.

When you're around and have a minute, please release 3.1.14!

PR #9655, opened 2024-04-13
Update MANUAL.txt to mention custom-style works with ODT

iandol:
This commit https://github.com/jgm/pandoc/commit/bde3d7622ddd2cda7adabc309335bb2865c1d520 allowed `custom-style` to work with ODT but this is not mentioned in the manual. I tested what happens when the Style is not present and updated the text accordingly. I edited via Github web interface so hard wrapping is approximate. I hope the anchor to the reference doc works as expected (`#option--reference-doc` is the anchor as far as I could determine)...

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -7199,8 +7199,8 @@ would style the two contained lines with the `Poetry` paragraph style.
 
 Styles will be defined in the output file as inheriting
 from normal text (docx) or Default Paragraph Style (odt), if the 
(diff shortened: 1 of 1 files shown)


jgm:
Perhaps best to wait til custom-style works for inline (span) content as well before adding this.

iandol:
Correct, this would be dependent on fixing #9657 / #2106, so those should be closed first. Thank you!

Issue #8249, opened 2022-08-26
Feature Request: Simplify Output of Conversion to LaTeX


ocelotStride:
Is it possible to add a command which outputs LaTeX without the additional "flare"

For example, I don't want `\tightlist`, I don't want `\def\labelenumi{\arabic{enumi}.}`, I don't want `\hypertarget{modelling-data}{%` and it's matching `\label{modelling-data}}`.

Is it possible to just output very simple latex with simple sections, subsection, itemise, and enumerate sections?

I've tried finding a way to do this but I'm starting to think there isn't. I even created my own custom.latex template and these options still show up

jgm:
Changes in the template wouldn't affect this; the writer would have to be changed.
I'm sorry, it would add too much complexity (and therefore too many bugs) to the code to add configuration options for simpler output.
I'd suggest creating a custom Lua writer to produce the simple LaTeX you desire.
https://pandoc.org/custom-writers.html

PR #9666, opened 2024-04-18
New approach to preventing image overflow in LaTeX (fixes #9660)

### LINKED ISSUE #9660: Large SVGs are rendered incorrectly in latex output
**Explain the problem.**

Firstly, I just want to mention this is *not* an issue with pandoc's internal translation of svgs to pdf. That works fine! The issue here is when svgs are outputted into 'latex'.

Consider the following one line markdown document, using this svg:

bubbleup

```
![](bubbleup.svg)
```

If we build this by doing:

`pandoc pan.md --to latex -o pan.tex -s`

`pdflatex --shell-escape pan.tex` (--shell-escape is required as we have an svg inside the tex file. The latex svg package will invoke inkscape to render the svg).

Then the output is badly corrupted.

This happens (I think) because the SVG needs to be resized, and the resizing doesn't work correctly due to the setting of `Gin` in the  `\setkeys{Gin}{...}` line of the latex template. I raised this issue with the `svg` latex package here: https://github.com/mrpiggi/svg/issues/60  , where I was told setting `Gin` in this way isn't supported.

The reason the bad rendering happens at all is by default the latex `svg` package extracts text from the svg and renders it using the latex engine. The resizing is messing up the placement of this text.

This could be fixed by setting `inkscapelatex` in the `svg` package to false. This makes the svg latex package just render the svg to a single PDF and everything works fine. This would stop text in SVGs being rendered as latex, if anyone was using that functionality (they can of course get it back by editing the template).

The other option would be to disable auto-resizing, but that would effect all images (not just svg), and be much worse (as large pictures would overflow the page).

This problem was originally found as an issue in quarto, which does this two-step pass of using pandoc to turn markdown -> tex, then compiling the resulting tex.

**Pandoc version?**
Ubuntu 23.04

```
pandoc 3.1.13
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: /home/caj/.local/share/pandoc
```

jgm:
Instead of relying on graphicx internals, we now define a macro that gets used with images when explicit size information is not provided. This macro prevents the image from overflowing vertically or horizontally.

Templates will have to be updated to include the new macro.

Closes #9660.

The macro should be credited to @mrpiggi. See https://github.com/mrpiggi/svg/issues/60 for discussion.

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -302,15 +302,17 @@ $endif$
 $if(graphics)$
 \usepackage{graphicx}
 \makeatletter
-\def\maxwidth{\ifdim\Gin@nat@width>\linewidth\linewidth\else\Gin@nat@width\fi}
-\def\maxheight{\ifdim\Gin@nat@height>\textheight\textheight\else\Gin@nat@height\fi}
-\makeatother
-% Scale images if necessary, so that they will not overflow the page
-% margins by default, and it is still possible to overwrite the defaults
-% using explicit options in \includegraphics[width, height, ...]{}
-\setkeys{Gin}{width=\maxwidth,height=\maxheight,keepaspectratio}
+\newsavebox\pandoc@box
+\newcommand*\pandocbounded[1]{% scales image to fit in text height/width
+  \sbox\pandoc@box{#1}%
+  \Gscale@div\@tempa{\textheight}{\dimexpr\ht\pandoc@box+\dp\pandoc@box\relax}%
+  \Gscale@div\@tempb{\linewidth}{\wd\pandoc@box}%
+  \ifdim\@tempb\p@<\@tempa\p@\let\@tempa\@tempb\fi% select the smaller of both
+  \ifdim\@tempa\p@<\p@\scalebox{\@tempa}{\usebox\pandoc@box}%
+  \else\usebox{\pandoc@box}%
+  \fi%
+}
 % Set default figure placement to htbp
-\makeatletter
 \def\fps@figure{htbp}
 \makeatother
 $endif$
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -1087,6 +1087,9 @@ inlineToLaTeX (Image attr@(_,_,kvs) _ (source, _)) = do
                                   [d <> "\\textheight"]
                                 _ -> []
       optList = showDim Width <> showDim Height <>
+                (case (dimension Height attr, dimension Width attr) of
+                  (Just _, Just _) -> []
+                  _ -> ["keepaspectratio"]) <>
                 maybe [] (\x -> ["page=" <> literal x]) (lookup "page" kvs) <>
                 maybe [] (\x -> ["trim=" <> literal x]) (lookup "trim" kvs) <>
                 maybe [] (const ["clip"]) (lookup "clip" kvs)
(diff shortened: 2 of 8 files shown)


tarleb:
An approach like this might make it necessary to reconsider #8249. Or maybe to publish a `pandoc` CTAN package, as the growing number of custom commands makes it increasingly difficult to maintain custom LaTeX templates.

jgm:
@tarleb I think I'd rather not publish a pandoc CTAN package; that's one more maintenance step required in every release.  Big changes like this are relatively rare, but in general one does have to keep an eye on template changes if one uses custom templates.

I don't remember why I haven't merged this yet.  I suppose it's because I'm worried about people who will use custom templates with the new pandoc and get errors about undefined `\pandocbounded`.  Do you think we should hold off or merge it?  It does solve a real problem.

#8249 is still a cool idea and would largely avoid issues like this; I just haven't gotten around to exploring it.

tarleb:
I'm torn. On one hand I worry about how much this would impact the usefulness of LaTeX snippets, and how the snippets are becoming more pandoc-specific. On the other hand, I fully agree that this solves a real problem, and I'm aware that some people, all of whom know much more about LaTeX than I do, have been complaining about the current image scaling method.

Adding a `plain` flag will certainly lead to discussions about what *exactly* clean LaTeX output looks like, so that's a risky step, too. But I think it would currently be my favored approach.

jgm:
In principle, we could omit the `\pandocbounded` when `writerTemplate` is Nothing.
This would make fragments more "plain," at the cost of introducing a potentially confusing mismatch between fragment and standalone output.

jgm:
It's also a possibility to patch `\includegraphics` to add `\pandocbounded` implicitly, so that we can emit LaTeX that would at least work without special macro definitions.  (Against this see the comment https://github.com/mrpiggi/svg/issues/60#issuecomment-2062399458 .) This would still require a sizable bit of code in the default template, but it would improve backwards-compatibility with older templates. It's still a bit of a mixed bag, though.  With a patched `\includegraphics`, pandoc would still work with older templates, but we'd lose the ability to entirely defeat the size upper bound by specifying sizes explicitly (unless we made the new definition very complicated and actually parsed options).

Maybe it's okay to do something slightly painful and start using `\pandocbounded`. I'm just not sure.

CGMossa:
I've just been facing issues with the pandoc and SVG files, so this would be a welcome change. Thanks for the hardwork on this!

jgm:
Note that up to 2014 we did redefine includegraphics in the template: see commit a3cce8d3594fd5da7a83c9926e39eb3ec210b83b

tarleb:
I agree that this is preferable over the current system. I can open a new issue to discuss something like a `plain` extension for LaTeX.

pklausner:
So where and how should this `\pandoxbounded` be defined?

I'm not able to insert the snippet https://github.com/mrpiggi/svg/issues/60  at the right place, whatever it's supposed to do.

jgm:
This is the code (defined in data/templates/common.latex):
```
\makeatletter
\newsavebox\pandoc@box
\newcommand*\pandocbounded[1]{% scales image to fit in text height/width
  \sbox\pandoc@box{#1}%
  \Gscale@div\@tempa{\textheight}{\dimexpr\ht\pandoc@box+\dp\pandoc@box\relax}%
  \Gscale@div\@tempb{\linewidth}{\wd\pandoc@box}%
  \ifdim\@tempb\p@<\@tempa\p@\let\@tempa\@tempb\fi% select the smaller of both
  \ifdim\@tempa\p@<\p@\scalebox{\@tempa}{\usebox\pandoc@box}%
  \else\usebox{\pandoc@box}%
  \fi%
}
% Set default figure placement to htbp
\def\fps@figure{htbp}
\makeatother
```

pklausner:
Thanks, this saved the day - the markdown to formally correct pdf pipeline was broken.

PR #9668, opened 2024-04-18
Lua: add a `pandoc.log` module. [API change]

tarleb:


jgm:
The idea seems sound to me.

PR #9673, opened 2024-04-20
Update MANUAL.txt to at least introduce term before shorten viersion

jidanni:
Should be in opposite order:
```
$ 2>&- man pandoc|grep -i odt|fgrep \(
              • odt (ODT)
              • odt (OpenOffice text document)
```
I'm saying that the detailed name should be shown first, or even better, both places.

There are several similar cases in this file of the same phenomena.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -259,7 +259,7 @@ header when requesting a document from a URL:
     - `man` ([roff man])
     - `muse` ([Muse])
     - `native` (native Haskell)
-    - `odt` ([ODT])
+    - `odt` ([OpenOffice text document])
     - `opml` ([OPML])
     - `org` ([Emacs Org mode])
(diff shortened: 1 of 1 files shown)


MANUAL.txt
```diff
@@ -259,7 +259,7 @@ header when requesting a document from a URL:
     - `man` ([roff man])
     - `muse` ([Muse])
     - `native` (native Haskell)
-    - `odt` ([ODT])
+    - `odt` ([OpenOffice text document])
```
jgm, line 262: This will be a broken link, because the reference is not defined.
It would have to be `[OpenOffice text document][ODT]` as in the other case.

jidanni: Please fix it your way. All I know is how it looks on the man page. Thanks.

PR #9688, opened 2024-04-23
Harmonize typographic dashes

salim-b:
Harmonizes the usage of typographic dashes to ~~the most frequently used type: *spaced en dash*~~ *unspaced em dashes*

I systematically checked all occurences (with some simple regex).

cf. https://en.wikipedia.org/wiki/Dash#Parenthetic_and_other_uses_at_the_sentence_level

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -555,7 +555,7 @@ header when requesting a document from a URL:
 
 `--indented-code-classes=`*CLASSES*
 
-:   Specify classes to use for indented code blocks -- for example,
+:   Specify classes to use for indented code blocks---for example,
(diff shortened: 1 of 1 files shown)


jgm:
I'd rather standardize on `---` without spaces.
Except in cases like the list under `lang` and the `--` used in the attribution of a quote to John Gruber.

salim-b:
> I'd rather standardize on `---` without spaces. Except in cases like the list under `lang` and the `--` used in the attribution of a quote to John Gruber.

Aight, done.

I've also fixed the quotation (it was missing a hard line break).

jgm:
perfect, thanks

PR #9738, opened 2024-05-10
Collect transforms in new module T.P.Transforms. [API change]

tarleb:
Bundles all document transformations, i.e., functions that run predefined filters on Pandoc documents.


tarleb:
The reason for this PR is that I considered to expose `adjustLinksAndIds` to Lua. I decided against it, and this PR is what remains of that experiment. Honestly not sure if it's all that helpful.

jgm:
This makes a lot of sense from the point of view of clean organization.
It's an API change but probably worth it.

PR #9739, opened 2024-05-10
Remove Text.Pandoc.Shared.stripEmptyParagraphs [API change]

tarleb:
This is no longer used.


jgm:
I think it would be best to deprecate first and then remove it in a later release.

jgm:
Since this is part of the public API.

jgm:
Since we're changing the API anyway, I'm just going to merge this.

jgm:
Actually now it needs rebasing because of changes to T.P.Shared from the other PR.

PR #9744, opened 2024-05-10
RST reader: fix `figclass` and `align` annotations for figures

### LINKED ISSUE #7473: rst figure options are applied to image only
**Explain the problem.**
When converting a reStructuredText `figure` directive to HTML, options like `figclass` and `align` are only applied to the image and not the entire figure in the output. Based on the [reference documentation](https://docutils.sourceforge.io/docs/ref/rst/directives.html#figure) for `figure`, I think these options are intended to apply to the entire figure.

For example, if I run the following command:
```
$ pandoc --from=rst --to=html5 <<EOF
.. figure:: example.img
   :figclass: foobar

   This is a caption.
EOF
```
I would expect the output to be:
```
<figure class="foobar">
<img src="example.img" alt="This is a caption." /><figcaption aria-hidden="true">This is a caption.</figcaption>
</figure>
```
Instead, I get the following, where the `class=foobar` attribute is part of the `img` tag.
```
<figure>
<img src="example.img" class="foobar" alt="This is a caption." /><figcaption aria-hidden="true">This is a caption.</figcaption>
</figure>
```

This issue causes the image and caption to get separated if the CSS stylesheet tries using something like `float` to align the figure.

**Pandoc version?**
I can reproduce this issue with the current development version (git) and pandoc 2.14.0.2 (pacman version: 2.14.0.2-2).
My OS is Arch Linux.

gok99:
Fixes #7473 

This PR implements the following:
1. It seems `figclass`es are currently not being extracted correctly since the wrong key-values are used - this PR extracts them from `keyvals` instead of the key-value output from `imgAttr`
2. `figclass` and `align` annotations are implemented following the [rst docs](https://docutils.sourceforge.io/docs/ref/rst/directives.html#figure): these options apply to the figure while other options are passed onto the underlying image.

### DIFF
--- a/src/Text/Pandoc/Readers/RST.hs
+++ b/src/Text/Pandoc/Readers/RST.hs
@@ -634,7 +634,7 @@ directive' = do
       name = trim $ fromMaybe "" (lookup "name" fields)
       classes = T.words $ maybe "" trim (lookup "class" fields)
       keyvals = [(k, trim v) | (k, v) <- fields, k /= "name", k /= "class"]
-      imgAttr cl = (name, classes ++ alignClasses, widthAttr ++ heightAttr)
+      imgAttr cl = (name, classes, alignClasses, widthAttr ++ heightAttr)
         where
           alignClasses = T.words $ maybe "" trim (lookup cl fields) <>
                           maybe "" (\x -> "align-" <> trim x)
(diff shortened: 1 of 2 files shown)


jgm:
Thanks!

PR #9757, opened 2024-05-14
Update MANUAL.txt to remove false info on Lua mode not supporting -i

iandol:
Current pandoc-lua does support REPL and `-i` as far as I understand.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -7455,7 +7455,6 @@ Calling the pandoc executable under the name `pandoc-lua` or with
 `lua` as the first argument will make it function as a standalone
 Lua interpreter. The behavior is mostly identical to that of the
 [standalone `lua` executable][lua standalone], version 5.4.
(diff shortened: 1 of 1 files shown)


jgm:
I also note
```
% pandoc lua -h
pandoc: user error (unrecognized option `-h'
Usage: pandoc lua [options] [script [args]]
  -e stat    execute string 'stat'
  -i         interactive mode -- currently not supported
  -l mod     require library 'mod' into global 'mod';
             if 'mod' has the pattern 'g=module', then
             require library 'module' into global 'g'
  -v         show version information
  -E         ignore environment variables -- partially supported
  -W         turn warnings on -- currently not supported
)
```
@tarleb where is that message "interactive mode -- currently not supported" defined? We should change that.

tarleb:
Oh, thanks. I fixed it here: https://github.com/hslua/hslua/blob/main/hslua-cli/src/HsLua/CLI.hs#L248

PR #9826, opened 2024-05-30
Update MANUAL.txt to fix alerts example

iandol:
There is no [!INFO] alert and Pandoc does not trigger the alerts syntax if you try to use it, replace it with [!TIP] which does exist and does work.


jgm:
thank you

PR #9862, opened 2024-06-09
[WIP]: Lua: Add function `pandoc.init`

tarleb:
This function allows to initialize the state in `PANDOC_STATE`, but only if it hasn't been initialized yet. The function has no effect in filters and custom readers/writers, but only in standalone scripts.

Any function that queries the state will cause it to be auto-initialized to the default value, so there should be no behavior changes.

The Haskell API probably needs more work, I tried to keep changes small for now.


jgm:
Could you say a bit more about the kind of things that would require this, by way of motivation?

tarleb:
Sure, sorry. The background is that I'm extending the [Inara](https://github.com/openjournals/inara) system used by [JOSS](https://joss.theoj.org), because I need to create files with names that are built from metadata values. I don't want to use too much shell scripting for that, so I'm trying to write it as a standalone Lua script. [WIP](https://github.com/openjournals/inara/blob/inara.lua/inara.lua).

There script always runs with the default CommonState, so without user data directory and `.` as resource path. It's not a big problem, because it's fairly easy to work around this, but some code would be shorter if I could configure the CommonState. The goal would be to write something like

``` lua
pandoc.init {
  user_data_dir = '/usr/share/openjournals/data',
}
```

so I could do things like

```
local template = pandoc.template.default 'html5'
```

to get templates from that data directory (if they exist).

So it's less that I *need* it, and more that it seems to be a missing functionality that might help in some cases.

jgm:
No objections on my part!

tarleb:
Superseded by #11371.

Issue #6534, opened 2020-07-14
Selectively disable typographic ligatures in LuaLaTeX


fkohrt:
[English](https://github.com/mrkline/modern-latex/blob/ac485ba/fonts.tex#L219-L225) and [German](http://mirrors.ctan.org/macros/latex/contrib/tudscr/doc/tutorials/treatise.pdf#subsection.11.3) introductions to typography mention that some words must not contain ligatures at certain points, such as

- shel**ff**ul
- Kau**fl**eute

(illegal ligature in **strong emphasis**)

In LaTeX, one would either have to manually disable them with `\kern0pt` or `\hspace{0pt}`(such as `shelf{\kern0pt}ful`), or use `microtype` to [disable them globally](https://texfragen.de/ligatur#global).

However, there's a LuaLaTeX package that does this automatically for an included selection of words: `selnolig`. According to [`selnolig`'s documentation](http://mirrors.ctan.org/macros/luatex/latex/selnolig/selnolig.pdf#subsection.2.3), it has to be loaded after the `babel` package. Therefore, just using `header-includes` is not feasible, as that's located [before](https://github.com/jgm/pandoc-templates/blob/70a4509/default.latex#L324-L336).

Hence I suggest loading the `selnolig` package in case the LuaLaTeX engine is used, probably somewhere after [line 341](https://github.com/jgm/pandoc-templates/blob/70a4509/default.latex#L341) of the LaTeX template:

```diff
--- pandoc-templates/default.latex	2020-07-15 00:28:58.124000000 +0200
+++ -	2020-07-15 00:31:00.189391730 +0200
@@ -339,6 +339,9 @@
 $endif$
 \fi
 $endif$
+\ifluatex
+  \usepackage{selnolig} $-- @template
+\fi
 $if(dir)$
 \ifxetex
   % Load bidi as late as possible as it modifies e.g. graphicx
```

mb21:
you can probably use a [custom template](https://pandoc.org/MANUAL.html#templates)?

fkohrt:
Oh yeah, I'm doing that. But I thought this might improve the default template; in a similar way everybody could load `microtype` themself, but instead it's been added to the default template.

jgm:
My only worry here is about depending on packages that aren't widely available.  We try to stick with the most common ones so the default template will work with a more minimal install.

jgm:
Still, I think this is probably a good idea.

jgm:
[...]

PR #9864, opened 2024-06-10
Provide selnolig with babel language

### LINKED ISSUE #9863: Regression for LuaLaTeX: ligatures not suppressed anymore
**Explain the problem.**

The `selnolig` package has previously been [integrated](https://github.com/jgm/pandoc/issues/6534) into the default LaTeX template to automatically disable incorrect ligatures.

With #7604, the document's (babel) language is not passed as classoption anymore, on which `selnolig` relied on. Citing from its [documentation](http://mirrors.ctan.org/macros/luatex/latex/selnolig/selnolig.pdf#subsection.2.1):

> - If you load the package without an explicit language option, i.e., as
>   
>       \usepackage{selnolig}
>   
>   but if one or more of the language options noted above are specified as options in the `\documentclass` instruction, LaTeX will pass these options on to the `selnolig` package.

Therefore, currently `selnolig` does not work for English or German documents.

Now, the alternative is to pass the (babel) language directly to `selnolig`:

> - If your document is in English and you want to enable a “basic” set of rules to suppress f-ligatures selectively, load the package by issuing the following instruction in the preamble of your document:
>   
>       \usepackage[english]{selnolig}
>   
>   Synonymous options are `UKenglish`, `british`, `USenglish`, `american`, `canadian`, `australian`, and `newzealand`.
> 
> [...]
> 
> - If your document is written in German, load the package as follows:
>   
>       \usepackage[ngerman]{selnolig}
>   
>   Synonymous language options are `german`, `austrian`, `naustrian`, `swissgerman`, and `swiss`.

**Pandoc version?**

pandoc 3.2

fkohrt:
Since 3d8f011, the babel language is not
a classoption anymore, on which selnolig
relied. Now the babel language is directly
passed to selnolig.

Fixes #9863

### DIFF
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -444,9 +444,11 @@ $endif$
 $for(header-includes)$
 $header-includes$
 $endfor$
+$if(babel-lang)$
 \ifLuaTeX
-  \usepackage{selnolig}  % disable illegal ligatures
+  \usepackage[$babel-lang$]{selnolig}  % disable illegal ligatures
 \fi
+$endif$
 $if(dir)$
 \ifPDFTeX
   \TeXXeTstate=1
--- a/data/templates/default.latex
+++ b/data/templates/default.latex
@@ -446,7 +446,7 @@ $header-includes$
 $endfor$
 $if(babel-lang)$
 \ifLuaTeX
-  \usepackage[$babel-lang$]{selnolig}  % disable illegal ligatures
+  \usepackage[$babel-lang$$for(babel-otherlangs)$,$babel-otherlangs$$endfor$]{selnolig}  % disable illegal ligatures
 \fi
 $endif$
 $if(dir)$


jgm:
Great, thank you.
We also have a `babel-otherlangs` variable (a list).  Can multiple languages be passed in this way to selnolig?

fkohrt:
It appears to me that works, at least that's what I see for a sample document with both English and German words. For other languages no ligature suppression rules exist in `selnolig`.

fkohrt:
Alright, I accounted for `babel-otherlangs`. For reference, this is the document `doc.md` I tested it with:

```md
---
title: Test
lang: de
---

Auflage

::: {lang=en}
shelfful
:::
```

Run with...

```sh
pandoc --standalone --from=markdown --to=latex --output=doc.tex --template=patched-default.latex test2.md
lualatex doc.tex
```

...where `patched-default.latex` is [`default.latex`](https://github.com/jgm/pandoc-templates/blob/ff51cf4/default.latex) patched as follows:

```diff
--- -
+++ default.latex
@@ -444,9 +444,11 @@
 $for(header-includes)$
 $header-includes$
 $endfor$
+$if(babel-lang)$
 \ifLuaTeX
-  \usepackage{selnolig}  % disable illegal ligatures
+  \usepackage[$babel-lang$$for(babel-otherlangs)$,$babel-otherlangs$$endfor$]{selnolig}  % disable illegal ligatures
 \fi
+$endif$
 $if(dir)$
 \ifPDFTeX
   \TeXXeTstate=1
```

fkohrt:
@jgm I do wonder, however, whether the second commit was a smart move:

Since `selnolig`s creation, another package for ligature suppression has emerged that improves on `selnolig` for German texts: [`ligtype`](https://www.ctan.org/pkg/ligtype). I wonder whether it might be smarter to use `ligtype` for German-only texts and `selnolig` for everything else?

fkohrt:
Also, I just noticed that my second commit doesn't work anyways, because `selnolig` only accepts `german` and `english` (or their aliases). A document with french text, therefore, results in compilation errors. How about the following:

1. If `babel-lang` is set to a German variant and `babel-otherlangs` is empty, use `ligtype`.
2. Else, if `babel-lang` or `babel-otherlangs` contain `english` or `german` (or one of their supported aliases), pass them on to `selnolig`.

jgm:
The logic you suggest seems right.  We will have to calculate these things in the LaTeX writer and set a special variable, e.g. `selnolig-langs`, that the template can be sensitive to.

On `ligtype` -- in what ways is it better than `selnolig`?  If there isn't a big difference I'd be happy to reduce the dependency footprint and just use the one package.

fkohrt:
Hm, thinking about this: `selnolig` has to be loaded after the `babel` package, which, at the time of its inclusion to Pandoc, was not possible (hence #6534). The order of `header-includes` and `\usepackage{babel}` in the default LaTeX template has reversed since, making it entirely possible to use it with the default template via `header-includes`.

Given that `selnolig` (probably) hasn't been doing anything for the majority of Pandoc runs in the last three years (since 3d8f011) anyways and it's easy nowadays to include it on a custom basis, maybe consider removing it from the default LaTeX template and leaving up to the user what package to use for hyphenation suppression (`selnolig`, `ligtype`, `autotype` etc.)?

> On `ligtype` -- in what ways is it better than `selnolig`?

Quoting from `ligtype`'s documentation:

> There’s already the `selnolig` package by Mico Loretan providing such capability. However, it has
some significant shortcomings that render it barely applicable in many use cases. The `ligtype`
package steps in to address these deficiencies. The main improvements are:
> 
> __Kerning__ The `ligtype` package applies kerning for the glyphs of the suppressed ligatures. Both
font and user kerning are applied in the usual way.
> 
> __Short-armed f__ If available, it automatically replaces the f-glyphs of suppressed ligatures with their short-arm variant.
> 
> __Speed__ It is between ten and twenty times faster on a typical document, thanks to its completely different architecture.
> 
> Another, rather minor, difference is that it does not require fontspec.

fkohrt:
Good idea! `selnolig` does provide the command `\selnoligoff` to turn off suppression of ligatures. Loading `ligtype` before or after `selnolig` leads to an error, however:

```
! LaTeX Error: Command \nolig already defined.
               Or name \end... illegal, see p.192 of the manual.
```

How about providing the capability that `selnolig` never loads, for example, by setting `selnolig-langs` to be empty? If it's empty most users won't need `selnolig` anyways (provided they don't bring their own ligature suppression rules, in which case they could still manually load `selnolig`).

jgm:
`selnolig-langs` would be populated automatically by the writer; is your thought that a German user who didn't want `selnolig` could specify `selnolig-langs: []` in metadata and then load `ligtype`?

jgm:
Actually, a cleaner solution, with German, might be to load `ligtype` if available and otherwise `selnolig`. This wouldn't introduce a new (hard) dependency.  This would require a bit more fancy stuff in the writer and templates, but maybe worth it; this is the kind of thing users should be freed from worrying about.

fkohrt:
> `selnolig-langs` would be populated automatically by the writer; is your thought that a German user who didn't want `selnolig` could specify `selnolig-langs: []` in metadata and then load `ligtype`?

Yeah, that was my thought.

> Actually, a cleaner solution, with German, might be to load `ligtype` if available and otherwise `selnolig`.

But this one I like most! I am currently trying to figure out how `ligtype` compares to `autotype`, another recent package for ligature suppression, i.e. if one is to be preferred over the other.

jgm:
I pushed a fix that supersedes this, so closing.

jgm:
In the end I went for the simple approach that just uses `selnolig`.
Users who want to use a fancier package can always manually set `selnolig-langs: []` and add their own thing to `header-includes`....

PR #9925, opened 2024-06-28
pandoc-cli: Behave like a Lua interpreter when progname starts with lua

tarleb:
This allows to use pandoc as a drop-in replacement for the default Lua interpreter.

### DIFF
--- a/pandoc-cli/src/pandoc.hs
+++ b/pandoc-cli/src/pandoc.hs
@@ -56,6 +56,7 @@ main = E.handle (handleError . Left) $ do
     "pandoc-server.cgi" -> versionOr runCGI
     "pandoc-server"     -> versionOr $ runServer rawArgs
     "pandoc-lua"        -> runLuaInterpreter prg rawArgs
+    ('l':'u':'a':_)     -> runLuaInterpreter prg rawArgs
     _ ->
       case rawArgs of
         "lua" : args   -> runLuaInterpreter "pandoc lua" args


tarleb:
The alternative is to use a script like

``` sh
#!/bin/sh
pandoc-lua "$@"
```

The change would simplify that and would allow something like `ln -s /usr/local/bin/pandoc /usr/local/bin/lua`.

alerque:
I can see why this might be useful in some scenarios, but a comment on your "the alternative". It seems to me the most likely alternative would actually be to just use one of the current hashbang options in any scripts wanting to use the Lua interpreter:

```
#!/usr/local/bin/pandoc lua
```

or

```
#!/usr/bin/env -S pandoc lua
```

In addition (or maybe in place of) to what this PR does I would add the `$0` name being `pandoc-lua` as a way to invoke the lua subcommand, so `ln -s /usr/local/bin/pandoc  /usr/local/bin/pandoc-lua`. This is more is line with other programs that provide subcommands as linked binaries, and what Pandoc already does with `pandoc-server`.

tarleb:
> In addition (or maybe in place of) to what this PR does I would add the `$0` name being `pandoc-lua` as a way to invoke the lua subcommand, so `ln -s /usr/local/bin/pandoc /usr/local/bin/pandoc-lua`.

That already works (and I recently updated the Docker images to include that symlink).

``` sh
docker run --rm -it --entrypoing=pandoc-lua pandoc/core
```

jgm:
What is the motivation for this?  Aren't there differences between what pandoc-lua does and what stock lua does?  Wouldn't it cause confusion to have a system on which `lua` is symlinked to `pandoc`?

alerque:
This isn't placing such a symlink, only making it possible if somebody manually were to do that. A possible use case is to use as a drop in replacement Lua VM inside some other app that expects to find Lua available in the $PATH at `lua` and isn't easy to change — any also where you want to exactly match Pandoc's Lua VM version for whatever reason. I have certainly run into scenarios like this in the past. Of course there are differences between system default VM and Pandoc's built in one, but since the user would have to have setup the scenario to begin with I would err on the side of assuming they have a good reason. Really it isn't much different that the utility of `pandoc-lua` (which serves my own use cases well enough) but where the user doesn't happen to have an easy a time changing the called name.

jgm:
> This isn't placing such a symlink, only making it possible if somebody manually were to do that.

I know...but making it possible only makes sense if it would be a good idea.

With stock Lua, for example,

```
print(pandoc.Str("hi"))
```

will produce an error.  With pandoc-lua it will print `Str "hi"`.  So, IMO it's very misleading, and the source of difficult to find bugs, to make `lua` run pandoc's lua interpreter.

tarleb:
> making it possible only makes sense if it would be a good idea.

It would be handy, but that doesn't make it a good idea. I'll stick to my little helper script.

PR #9930, opened 2024-06-30
Add Japanese translations

stephen-huan:
This is copied directly from [Polyglossia](https://github.com/reutenauer/polyglossia/blob/702eb276dc9fee7039872e310b08ba44a3f7b1d9/tex/gloss-japanese.ldf#L37-L59) rather than [Babel](https://github.com/latex3/babel/blob/4d2c814307b919e7b865b068236c2fa6f495ce55/locale/ja/babel-ja.ini#L30-L56) because the Polyglossia version is more complete. The only differences are that Polyglossia uses 「端書き」 for `Preface` (Babel uses 「前書き」) and Polyglossia uses 「文献目録」 for `Bibliography` while Babel uses 「参考文献」 (which is the same word both use for `References`). These changes seem reasonable so I defer to the Polyglossia keys for consistency; happy to switch to Babel if necessary.

Is there interest in automatically copying translation data from Polyglossia/Babel with some sort of ad-hoc script? Happy to contribute to something to `tools/` if so.


jgm:
Sure, I think some kind of automatic system could be helpful.  There have been some minor (manual) tweaks to some of the translations over the years (could `git log data/translations` to see).

PR #9968, opened 2024-07-10
Add gemini URI scheme

prodrigestivill:
In order to detect properly `gemini://` links I propose to add it as an unofficial URI scheme.

More info regarding this scheme at [wikipedia](https://en.wikipedia.org/wiki/Gemini_(protocol)) and their [official website](https://geminiprotocol.net/).


prodrigestivill:
If somebody requires it, as a workarround this lua filter or similar can be used using `--lua-filter=filter.lua`:

```lua
-- Fix gemini:// links
function Link(el)
  protopos = string.find(el.target, "/gemini://")
  if protopos then
    el.target = string.sub(el.target, protopos + 1);
  end
  return el
end
```

Issue #9956, opened 2024-07-05
Improving the Typst base template to match equivalent PDF options


christopherkenny:
## Proposed Improvement:

The default template for Typst currently supports many fewer base features than the PDF template. This, of course makes sense, as Typst is much newer. Many features included in PDF templates could be added to the Typst base template without substantial amounts of code. Below, I've shortened a list from a discussion in Quarto:

### Title & Author

- `thanks`: To the call to `#title`, we could add:

```typst
#if thanks != none {
    footnote(thanks, numbering: "*")
    counter(footnote).update(n => n - 1)
}
````

### Fonts

- `linestretch`: 1 = single spaced, 2 = double spaced, and so on

```typst
set par(leading: linestretch * 0.65em)
```

- `mathfont`: 

```typst
#show math.equation: set text(font: mathfont)
```

- `codefont`:

```typst
#show raw: set text(font: codefont)
```

- `linkcolor`: Each color setting below require converting content to string or running a `.replace("\\#", "#")` on `linkcolor` due to pandoc's escaping

```typst
#show link: set text(fill: rgb(linkcolor))
```

- `filecolor`, `urlcolor`: these have to be handled together as far as I know

```typst
#show link: this => {
    if type(this.dest) != label {
        text(this, fill: rgb(filecolor))
    }
}
```

- `citecolor`:

```typst
#show ref: this => {
    text(this, fill: rgb(citecolor))
}
```

-  `toccolor`

```typst
#show link: this => {
    if type(this.dest) == label {
        text(this, fill: rgb(to-string(filecolor)))
    }
}
```

### Metadata

- `keywords`

```typst
#set document(keywords: keywords)
```

- `title-meta`

```typst
#set document(title: title-meta)
```

- `author-meta`

```typst
#set document(author: author-meta)
```

I'm happy to PR in any of these, if you are interested.

## Alternatives:

- This started as a downstream discussion for Quarto (https://github.com/quarto-dev/quarto-cli/discussions/10223), but it was suggested to look here for editing template-specific pieces.

- These could also be in custom templates, of course. However, at least some set of these options are probably good to have in the default template.

jgm:
I'd certainly consider a PR with changes like this.
(It would need careful testing, though.)

Note that `author-meta` and `title-meta` are just used in formats that require special treatment for these fields (e.g. formats where these can't contain formatting); we don't use them for typst.

PR #9970, opened 2024-07-10
Adding new features to the Typst template

christopherkenny:
This implements the changes suggested in #9956, with the exception of the `filecolor`/`urlcolor` one. These would require adding some regex to guess the link types. This is theoretically possible to do, but it wasn't clear to me that this is a good thing to put in a default template. Happy to adjust if you have thoughts on this.

Some things to note: 

- I'm converting colors by passing them as content, as I was seeing pandoc escape `#` if that was included. An alternative here is to just do something like: `rgb(linkcolor.replace("\\#", "#"))`.
- I set the default fonts for math and code ("raw") to fonts that are bundled with Typst. These need not be those fonts if there are more familiar pandoc preferences.

### DIFF
--- a/data/templates/default.typst
+++ b/data/templates/default.typst
@@ -72,6 +72,12 @@ $endif$
 $if(fontsize)$
   fontsize: $fontsize$,
 $endif$
+$if(mathfont)$
+  mathfont: ($for(mathfont)$"$mathfont$",$endfor$),
+$endif$
+$if(codefont)$
+  codefont: ($for(codefont)$"$codefont$",$endfor$),
+$endif$
 $if(linestretch)$
   linestretch: $linestretch$,
 $endif$
--- a/data/templates/default.typst
+++ b/data/templates/default.typst
@@ -84,6 +84,15 @@ $endif$
 $if(section-numbering)$
   sectionnumbering: "$section-numbering$",
 $endif$
+$if(linkcolor)$
+  linkcolor: [$linkcolor$],
+$endif$
+$if(citecolor)$
+  citecolor: [$citecolor$],
+$endif$
+$if(toccolor)$
+  toccolor: [$toccolor$],
+$endif$
   cols: $if(columns)$$columns$$else$1$endif$,
   doc,
 )
--- a/data/templates/default.typst
+++ b/data/templates/default.typst
@@ -57,6 +57,9 @@ $endif$
 $if(abstract)$
   abstract: [$abstract$],
 $endif$
+$if(thanks)$
+  thanks: [$thanks$],
+$endif$
 $if(margin)$
   margin: ($for(margin/pairs)$$margin.key$: $margin.value$,$endfor$),
 $endif$
--- a/data/templates/default.typst
+++ b/data/templates/default.typst
@@ -72,6 +72,9 @@ $endif$
 $if(fontsize)$
   fontsize: $fontsize$,
 $endif$
+$if(linestretch)$
+  linestretch: $linestretch$,
+$endif$
 $if(section-numbering)$
   sectionnumbering: "$section-numbering$",
 $endif$
--- a/data/templates/template.typst
+++ b/data/templates/template.typst
@@ -28,11 +28,14 @@
   codefont: "DejaVu Sans Mono",
   linestretch: 1,
   sectionnumbering: none,
+  linkcolor: ["#800000"],
+  citecolor: ["#0000FF"],
+  toccolor: ["#800000"],
   doc,
 ) = {
   set document(
     title: title,
-    author: authors.map(author => content-to-string(author.name)),
+    author: authors.map(author => content-to-string(author.name)).join(", ", last: ", and "),
     keywords: keywords,
   )
   set page(
@@ -51,6 +54,17 @@
   show math.equation: set text(font: mathfont)
   show raw: set text(font: codefont)
   set heading(numbering: sectionnumbering)
+
+  show link: set text(fill: rgb(content-to-string(linkcolor)))
+
+  show link: this => {
+    if type(this.dest) == label {
+        text(this, fill: rgb(content-to-string(filecolor)))
+    }
+  }
+  show ref: this => {
+    text(this, fill: rgb(content-to-string(citecolor)))
+  }
 
   if title != none {
     align(center)[#block(inset: 2em)[
--- a/data/templates/template.typst
+++ b/data/templates/template.typst
@@ -16,6 +16,7 @@
   keywords: (),
   date: none,
   abstract: none,
+  thanks: none,
   cols: 1,
   margin: (x: 1.25in, y: 1.25in),
   paper: "us-letter",
@@ -45,7 +46,10 @@
 
   if title != none {
     align(center)[#block(inset: 2em)[
-      #text(weight: "bold", size: 1.5em)[#title]
+      #text(weight: "bold", size: 1.5em)[#title #if thanks != none {
+            footnote(thanks, numbering: "*")
+            counter(footnote).update(n => n - 1)
+          }]
       #(if subtitle != none {
         parbreak()
         text(weight: "bold", size: 1.25em)[#subtitle]
--- a/data/templates/template.typst
+++ b/data/templates/template.typst
@@ -24,6 +24,7 @@
   region: "US",
   font: (),
   fontsize: 11pt,
+  linestretch: 1,
   sectionnumbering: none,
   doc,
 ) = {
@@ -37,7 +38,10 @@
     margin: margin,
     numbering: "1",
   )
-  set par(justify: true)
+  set par(
+    justify: true,
+    leading: linestretch * 0.65em
+  )
   set text(lang: lang,
            region: region,
            font: font,
--- a/data/templates/template.typst
+++ b/data/templates/template.typst
@@ -24,6 +24,8 @@
   region: "US",
   font: (),
   fontsize: 11pt,
+  mathfont: "New Computer Modern Math",
+  codefont: "DejaVu Sans Mono",
   linestretch: 1,
   sectionnumbering: none,
   doc,
@@ -46,6 +48,8 @@
            region: region,
            font: font,
            size: fontsize)
+  show math.equation: set text(font: mathfont)
+  show raw: set text(font: codefont)
   set heading(numbering: sectionnumbering)
 
   if title != none {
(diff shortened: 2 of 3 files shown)


agbeans24 (approved):

christopherkenny:
Hi @jgm, I don't want to bother you if this is already on your radar, but should I add you as a reviewer (for when you have time)? I'm not 100% sure of the norms for contributions here.

jgm:
Sorry for the delay on this!

data/templates/default.typst
```diff
@@ -69,8 +72,26 @@ $endif$
 $if(fontsize)$
   fontsize: $fontsize$,
 $endif$
+$if(mathfont)$
+  mathfont: ($for(mathfont)$"$mathfont$",$endfor$),
+$endif$
+$if(codefont)$
+  codefont: ($for(codefont)$"$codefont$",$endfor$),
+$endif$
```
jgm, line 80: Curious why you used `$for$` here.  `mathfont` would normally be expected to be a string in pandoc templates.

data/templates/default.typst
```diff
@@ -69,8 +72,26 @@ $endif$
...
 $if(section-numbering)$
   sectionnumbering: "$section-numbering$",
+$endif$
+$if(linkcolor)$
+  linkcolor: [$linkcolor$],
+$endif$
+$if(citecolor)$
+  citecolor: [$citecolor$],
+$endif$
```
jgm, line 92: It works in content like this, with the escaped `#`?
```
citecolor: [\#ffffff]
```
You've tested I assume?

data/templates/template.typst
```diff
@@ -16,36 +16,62 @@
...
+  thanks: none,
   cols: 1,
   margin: (x: 1.25in, y: 1.25in),
   paper: "us-letter",
   lang: "en",
   region: "US",
   font: (),
   fontsize: 11pt,
+  mathfont: "New Computer Modern Math",
+  codefont: "DejaVu Sans Mono",
```
jgm, line 28: Are these different from the fonts Typst uses by default?  Would it not be better to allow these to be specified but stay with Typst's normal defaults by default?

data/templates/template.typst
```diff
@@ -16,36 +16,62 @@
...
+  linestretch: 1,
   sectionnumbering: none,
+  linkcolor: ["#800000"],
+  citecolor: ["#0000FF"],
+  toccolor: ["#800000"],
```
jgm, line 33: Again, is this the same as Typst's default behavior?  If not, why these colors?  Would it perhaps be better to make Typst use its standard defaults unless these are specified?

data/templates/template.typst
```diff
@@ -16,36 +16,62 @@
...
   doc,
 ) = {
   set document(
     title: title,
-    author: authors.map(author => content-to-string(author.name)),
+    author: authors.map(author => content-to-string(author.name)).join(", ", last: ", and "),
```
jgm, line 38: I don't like this because `and` is an English word; anyone writing in another language would need to change the template.  We could consider using `&`, or just use commas.

data/templates/template.typst
```diff
@@ -16,36 +16,62 @@
...
   set page(
     paper: paper,
     margin: margin,
     numbering: "1",
   )
-  set par(justify: true)
+  set par(
+    justify: true,
+    leading: linestretch * 0.65em
+  )
```
jgm, line 49: What does the `leading` do here, and why is it needed with linestretch?  Is the result the same as Typst's defaults when the default linestretch is used?

data/templates/template.typst
```diff
@@ -16,36 +16,62 @@
...
+  show ref: this => {
+    text(this, fill: rgb(content-to-string(citecolor)))
+  }
+
   if title != none {
     align(center)[#block(inset: 2em)[
-      #text(weight: "bold", size: 1.5em)[#title]
+      #text(weight: "bold", size: 1.5em)[#title #if thanks != none {
+            footnote(thanks, numbering: "*")
+            counter(footnote).update(n => n - 1)
```
jgm, line 73: I assume this is needed because otherwise the first note would be `2`?  That's a bit unexpected when `numbering` is explicitly specified, but maybe Typst works that way?  (I'd consider it a bug if so.)

jgm (changes requested):
Thanks for this PR, and sorry I took so long to review it.  I made a number of queries and comments, which you can look over.

data/templates/template.typst
christopherkenny: Yes, so the numbering is a wrapper on top of a counter for footnotes. The numbers for footnotes are the counting numbers (1, 2, ...) and don't update even when you change the style. So this simply says: use * for the title, then set it back to 1 so that it can use 1, 2, ... for future footnotes.

data/templates/template.typst
christopherkenny: `leading` sets the space between lines. Here, mixing it with `linestretch` ensures that a value of `1 * 0.65em = 0.65em` or Typst's default single-spaced. For double spacing, then `1.3 = 2 * 0.65em`.  Here, this setting just makes it explicit so that setting `1`  for single, etc, all map onto what is expected.

data/templates/template.typst
christopherkenny: Happy to go either way here. I pushed a change to make it ` & `. I can make it anything that you think is least problematic, even removing the `last` special case.

data/templates/template.typst
christopherkenny: This matches the LaTeX default. (Many months later, I can see that I was focused a bit on matching LaTeX rather than keeping a pure Typst-focused set of defaults.) I suppose that making these editable but only using if set would make the most sense? If you think that's best, let me know and I can easily update it.

data/templates/template.typst
christopherkenny: These are the fonts that Typst uses by default, just set explicitly.

I can make these not set explicitly and only set if passed directly, if that's preferable.

data/templates/default.typst
christopherkenny: This is a bit of an odd one, but Typst uses a vector of fonts because it automatically falls back to the next font if one is not found. This use of `$for$` keeps that idea, where you might set (for example) `EB Garamond` and then `Garamond`. If `EB Garamond` is downloaded, then the template uses that. If not, it falls back to the (less featured, but more common) `Garamond`. This goes to the design choice that Typst will cycle through all the fonts chosen until it finds one available or ends up back at the default.

data/templates/default.typst
christopherkenny: I can confirm that this works, as I'm currently using it in a custom template. If there's something that you *don't* like about it, happy to brainstorm alternatives.

christopherkenny:
d05ac30 handles the merging with the pagenumber change

data/templates/template.typst
christopherkenny: Okay, I've further refined this to only be set if passed.

data/templates/template.typst
christopherkenny: This is also now simplified to not change the color unless the relevant color is specified.

christopherkenny:
Hi @jgm. I've rerequested your review for when you have a chance. These are all now set to be less opinionated and better fit with the default typst settings, especially removing colors and fonts unless specified.

christopherkenny:
Apologies, spoke too soon on the last message. Just went through and triple checked things and *now* it is ready for re-review.

data/templates/template.typst
```diff
@@ -16,73 +16,99 @@
...
+    }
 
-  #if abstract != none {
-    block(inset: 2em)[
-    #text(weight: "semibold")[Abstract] #h(1em) #abstract
-    ]
-  }
+    #if abstract != none {
+      block(inset: 2em)[
+        #text(weight: "semibold")[Abstract] #h(1em) #abstract
```
jgm, line 109: It would be nice not to force the use of an English word "Abstract" here. People write in other languages, too.  If typst doesn't have a way of retrieving a localization of this, we colud introduce `$abstract-title$`, and I could set it to a localized default in the Tyspt writer.

jgm:
looks pretty good -- I made one comment about "Abstract."

jgm:
The other thing that needs to be done is modifying MANUAL.txt so that the new variables are documented.
There should be a section "Variables for Typst."

jgm:
TODO (for me):  set `abstract-title` variable to a localized default in the typst writer.

jgm:
```hs
  abstractTitle <- translateTerm Abstract
  ```
and then when setting template variables,
```hs
                  defField "abstract-title" abstractTitle .
```

christopherkenny:
The MANUAL.txt and abstract-title defaults are now updated. Thanks for the review!

jgm:
Looks like the branch has to be rebased on to main in order to be merged cleanly.

christopherkenny:
So, I just rebased onto main, but I have a strong suspicion that this is not what you wanted. Before doing this, I made a `backup` branch which is just a copy prior to me rebasing. I can open a clean PR from that if this is not what you wanted.

rnwst:
@christopherkenny -- I had a look at what may have gone wrong with your rebase. My theory is as follows (please correct me if any of this is wrong!):
- During your work on this branch, you used the 'Update branch' button a few times to merge `jgm:main` into `christopherkenny:typst-template`.
- Your pandoc fork hasn't been synced since June 2024 and so the latest commit on `christopherkenny:main` is correspondingly old.
- When you rebased onto `main`, you therefore rebased onto a commit from June 2024 rather than the latest `main`. In addition, you ran `rebase` without the `--rebase-merges` option, so your merge commits (created previously via the 'Update branch' button) were replaced with all of the respective commits from `jgm:main`, which is how you ended up with 662 commits on your branch. If the `--rebase-merges` option had been used, I think the rebase would have had no effect (since the rebase would've been done on the same commit on `jgm:main` that your branch was branching out from anyway).

To fix this, you could restore your previous version of this branch (via your `backup` branch), sync your fork (and do a `git pull` to make sure your local `main` is up to date as well), and perform an interactive rebase on `main`, during which you would drop all merge commits.

christopherkenny:
Yeah, that sounds exactly like what happened. Thanks for the help @rnwst. I will get this cleaned up tomorrow.

christopherkenny:
Sorry for the delay and thank you for your patience. It is now rebased into 1 commit and all check pass @jgm.

jgm:
I rebased this from the command line and committed it. Thank you!

christopherkenny:
Thanks @jgm and @rnwst for the help and patience.

PR #9982, opened 2024-07-13
Move resource embedding into HTML writer [API change]

tarleb:


jgm:
I think this makes good sense, but I want to mull it over. Can you think of any arguments against doing this?

Issue #9603, opened 2024-03-25
DOCX Reader to support DOCX table custom style


rgaiacs:
Microsoft Word allows user to define custom style for tables.

![Screenshot of Microsoft Word Table Style Editor](https://github.com/jgm/pandoc/assets/1506457/2373dd24-2abc-4b8e-a1e2-e5d706f457d0)

Would be great if the DOCX reader would support table custom style as it support paragraph custom style.

# Minimal Working Example

[mwe-table-custom-style.docx](https://github.com/jgm/pandoc/files/14741864/mwe-table-custom-style.docx)

# XML structure

```
<w:tbl>
  <w:tblPr>
    <w:tblStyle w:val="Example"/>
    <w:tblW w:w="0" w:type="auto"/>
    <w:tblLook w:val="0620" w:firstRow="1" w:lastRow="0" w:firstColumn="0" w:lastColumn="0" w:noHBand="1" w:noVBand="1"/>
  </w:tblPr>
  ...
</w:tbl>
```

DOCX reader would have to read the value of `w:val` in `<w:tblStyle />` that is a child of `<w:tblPr>` which is a child of `<w:tbl>`.

# Expected Output

`<table>` has attribute `data-custom-style`. For example, `pandoc --from docx+styles --to html mwe-table-custom-style.docx` would return

```
<table data-custom-style="Example">
<colgroup>
<col style="width: 50%" />
<col style="width: 50%" />
</colgroup>
<thead>
<tr class="header">
<th>A</th>
<th>B</th>
</tr>
</thead>
<tbody>
<tr class="odd">
<td>1</td>
<td>2</td>
</tr>
</tbody>
</table>
```

# Observed Output

`pandoc --version` returns

```
pandoc 3.1.11.1
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: /home/raniere/.local/share/pandoc
Copyright (C) 2006-2023 John MacFarlane. Web: https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.
```

`<table>` is missing the attribute `data-custom-style`. For example, `pandoc --from docx+styles --to html mwe-table-custom-style.docx` returns

```
<table>
<colgroup>
<col style="width: 50%" />
<col style="width: 50%" />
</colgroup>
<thead>
<tr class="header">
<th>A</th>
<th>B</th>
</tr>
</thead>
<tbody>
<tr class="odd">
<td>1</td>
[...]

Issue #7549, opened 2021-09-05
docx writer : Changeable Table style (not hard-coded)


shutingrz:
I would like to use multiple styles of Table in a single docx, but the docx writer only allows the hard-coded "Table" style.

https://github.com/jgm/pandoc/blob/4340bd52c49b987087e123de2eae4464b300332f/src/Text/Pandoc/Writers/Docx/Table.hs#L79

There is an idea to achieve this with custom-styles in fenced_divs. It would be great to be able to use multiple Styles with this kind of syntax.

```
::: {custom-style="other-table-style"}
| Right | Left | Default | Center |
|------:|:-----|---------|:------:|
|   12  |  12  |    12   |    12  |
|  123  |  123 |   123   |   123  |
|    1  |    1 |     1   |     1  |
:::
```

There are a lot of issues related to docx writer tables, but none of them exist independently for this issue. I'm sorry if this is a duplicate.
Thanks

danblaker:
I would love to see support for `custom-style` in Table nodes. Being able to specify "NormalTable" or "GridTable" seems safe, given that they're part of the default Normal template in recent (>2016) versions of Word.

jgm:
Since we have attributes on Table now in the AST, we could implement custom-style.
This would require changing `tableToOpenXML` to take a parameter for a table style (to use instead of hardcoded `"Table"`).  This parameter could be extracted from `custom-style` attribute in T.P.Writers.Docx around line 954 and
passed to `tableToOpenXML`.

This would be expecting the `custom-style` attribute to be on the Table element itself, not an enclosing Div, so your example above wouldn't work out of the box. However, it would only take a Lua filter to migrate the custom-style to the Table itself.

jgm:
The docx writer has supported `custom-style` attributes on tables for some time now, so this can be closed.

Issue #6496, opened 2020-07-01
How to set a custom-style for a table?


kugelblitz:
Hello,

My task is to allow to set a Word table style (e. g. table without borders) in Pandoc Markdown.

In some previous version of pandoc (2.1.1) I used the following trick to apply a custom table style in the generated docx:
1. Put the Markdown table into a paragraph with `:::{custom-style=MyParagraphStyle}`, where `MyParagraphStyle` does not have any importance of its own -- it just serves as a marker for the postprocessing.
2. Generate docx with Pandoc. Table cells will have `MyParagraphStyle` as their paragraph style.
3. Do some postprocessing in Powershell where I search for such tables and then apply `MyTableStyle` to them. MyTableStyle is a table style properly defined in the reference docx. 

In Pandoc 2.9.2 and probably in 2.10, it no longer works. Pandoc no longer lets the style of the paragraph the table is in leak into the table cells.

Is this an intended change? Do I have any other ways to achieve my goal? Thanks.

jgm:
IT might be this change in 2.8:

```
    + Preserve built-in styles in DOCX with custom style (Ben Steinberg,
      #5670).  This change prevents custom styles on divs and spans
      from overriding styles on certain elements inside them, like
      headings, blockquotes, and links. On those elements, the
      "native" style is required for the element to display correctly.
      This change also allows nesting of custom styles; in order to do so,
      it removes the default "Compact" style applied to Plain blocks,
      except when inside a table.
```

jgm:
If you're postprocessing anyway, you could do something like put the word "SPECIALTABLE" before the table.  Postprocessing script would have to find this and strip it out.

kugelblitz:
I tackled the problem by moving the custom-style into the first cell of the table, and having the postprocessing script checking the style of the first cell. I see that some work is planned on table styles in #4697, so maybe some day the postprocessing hack will be gone.

metorm:
I want this, too.

`pandoc-crossref` uses a table to format equation numbers.

If we cannot used different table styles for different tables, all equations would be formated like other tables.

jgm:
#4697 is open and that's the place to track this.

Issue #4697, opened 2018-06-11
DOCX custom-style for lists and tables


bpj:
There have been several questions on pandoc-discuss lately on styling lists and tables in DOCX. This is currently not possible: If you wrap a list or table in a div with `custom-style=Foo` this results in applying the *paragraph* style `Foo` to all the items in the list or cells in the table, which probably seldom is what the user expects. What would it take to apply *list* style `Foo` or *table* style `Foo` in these cases? I guess @jkr is the person who could answer this.

agusmba:
@bpj see related issues: https://github.com/jgm/pandoc/issues/2667, https://github.com/jgm/pandoc/issues/2033, https://github.com/jgm/pandoc/issues/774

mb21:
Looks like we should close this issue (and others?) in favour of a single one?

agusmba:
I agree with not having multiple duplicate issues. Two of them have already been closed. I see two relevant ones open (maybe not as concise as this one, but with more discussion).

 #2667 deals with grid table formatting but indirectly also with table styles

#1024 has been mentioned elsewhere as a way of introducing necessary changes to support adding custom styles to tables (different from the official "Table") Note that this issue does not deal with custom-style for tables directly

jsommer1738:
Is there any progress on this?

tarleb:
> Is there any progress on this?

I don't know what kind of answer you expect. If this is a way of saying "go work on it", then please refrain from asking any such question in the future.

jsommer1738:
It certainly wasn't meant as a way of passively aggressively asking someone to work on it.  Maybe a better way to ask would have been, "Is this ticket scheduled for a specific release?"

In general, I can see the ticket is open but I have no visibility into when it might get worked.  And, no, I am in no way trying to pressure anyone to do work.

mb21:
it's all volunteers here :-) so there is no schedule. either somebody comes along and makes a pull request, or at some point one of the core devs finds a day to burn... but no telling when that day arrives :)

jsommer1738:
Understood.  Thanks!

PR #10009, opened 2024-07-19
Support custom-style attribute for docx table

Sebbones:
This is a suggested fix to support the `custom-style` attribute for docx tables. I'm referring to the issues https://github.com/jgm/pandoc/issues/9603, https://github.com/jgm/pandoc/issues/7549, https://github.com/jgm/pandoc/issues/6496, https://github.com/jgm/pandoc/issues/4697. I needed this feature to support multiple table styles in one document and saw a lot of issues addressing this feature. With absolutely no prior experience in Haskell, I just gave it a shot :) 

The conversion from html to docx with and without it, worked fine. With the `table[custom-style="CustomStyle"]` attribute, it used the CustomStyle for the table and if omitted, the fallback/default style is "Table".

I hope the code is okayish, but happy to change it if needed.

### DIFF
--- a/src/Text/Pandoc/Writers/Docx/Table.hs
+++ b/src/Text/Pandoc/Writers/Docx/Table.hs
@@ -22,7 +22,7 @@ import Control.Monad ( unless , zipWithM )
 import Control.Monad.Except ( throwError )
 import Data.Array ( elems, (!), assocs, indices )
 import Data.Text (Text)
-import Data.Maybe (catMaybes)
+import Data.Maybe (catMaybes, fromMaybe)
 import Text.Pandoc.Definition
     ( ColSpec,
       Caption(Caption),
@@ -71,7 +71,7 @@ tableToOpenXML :: PandocMonad m
                -> WS m [Content]
 tableToOpenXML opts blocksToOpenXML gridTable = do
   setFirstPara
-  let (Grid.Table (ident,_,_) caption colspecs _rowheads thead tbodies tfoot) =
+  let (Grid.Table (ident,_,tableAttr) caption colspecs _rowheads thead tbodies tfoot) =
         gridTable
   let (Caption _maybeShortCaption captionBlocks) = caption
   tablenum <- gets stNextTableNum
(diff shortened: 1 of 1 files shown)


jgm:
Perfect!  Thanks.

tdewin:
FINALLY, thank you soo soo soo much for those 4 golden lines of code

R3myG:
Would it be possible to have an example of how to use this new parameter? 
I'm using Quarto 1.6.32 which uses Pandoc 3.4 and the following didn't work:

```
::: {custom-style="TableCustom"}
| Header 1 | Header 2 |
|----------|----------|
| Val1         |    Val2      |
:::
```

```
| Header 1 | Header 2 |
|----------|----------|
| Val1         |    Val2      |
: {custom-style="TableCustom"}
````

**Comment by tarleb** [comment]
Pandoc Markdown doesn't yet have built-in syntax for table attributes. With `--from=commonmark_x` one would write

``` markdown
{custom-style="TableCustom"}
| Header 1 | Header 2 |
|----------|----------|
| Val1     |    Val2      |
```

For pandoc Markdown you'll need the first syntax you give, but use an additional Lua filter like the one below (untested)

``` lua
function Div (div)
  -- do nothing if the div doesn't have a custom-style attribute
  -- or if the div doesn't have exactly one element
  if not div.attributes['custom-style'] or #div.content == 1 then
    return div
  end
  local tbl = div.content[1]
  if tbl.t == 'Table' then
    tbl.attributes['custom-style'] = div.attributes['custom-style']
    return tbl
  end
end
```

R3myG:
@tarleb Thank you for the suggestion, I've tested it in quarto and couldn't make it work unfortunately.

I had to use the following notation because of quarto:
```
::: {custom-style="TableCustom"}
| Header 1 | Header 2 |
|----------|----------|
| Val1     |    Val2      |
:::
```

I also had to adapt the lua code to:

```
function Div (div)
    -- First, check if we have the custom-style attribute
    if not div.attributes['custom-style'] then
        return div
    end

    -- Check if div has content
    if not div.content or #div.content == 0 then
        return div
    end

    -- Get the first element
    local first_elem = div.content[1]
    
    -- Check if it's a table
    if first_elem and first_elem.t == 'Table' then
        -- Copy the custom-style attribute to the table
        if not first_elem.attributes then
            first_elem.attributes = {}
        end
        first_elem.attributes['custom-style'] = div.attributes['custom-style']
        return first_elem
    end

    -- If we get here, return the original div
    return div
end
```

Unfortunately, afterwards the tables were still using the `Table` style. 

If you have any suggestions on how to make it work, don't hesitate.

Issue #8245, opened 2022-08-24
Add a lof and lot options to docx output


nicola-lunghi:
see also https://github.com/jgm/pandoc/issues/458

Hi,
I was wondering if it would be complicated to add support for lot, lof in docx / odt
the toc works nice in the current version.

Something similar to this should work
```
makeLOF :: (PandocMonad m) => WriterOptions -> WS m [Element]
makeLOF opts = do
  let lofCmd = "TOC \\h \\z \\c \"Figure\""
  return
    [
      mknode "w:p" [] [
        mknode "w:pPr" [] [
          mknode "w:pStyle" [("w:val","TableofFigures")] (),
        ], -- w:pPr
        mknode "w:r" [] [
          mknode "w:fldChar" [("w:fldCharType","begin"),("w:dirty","true")] (),
          mknode "w:instrText" [("xml:space","preserve")] lofCmd,
          mknode "w:fldChar" [("w:fldCharType","separate")] (),
          mknode "w:fldChar" [("w:fldCharType","end")] ()
          ] -- w:r
      ] -- w:p
    ]
```
Now I need someone that knows haskell to verify it and add the missing bits :-) anyone?

I've takenthe xml from the examples [here](http://www.ericwhite.com/blog/exploring-tables-of-contents-in-open-xml-wordprocessingml-documents-part-2/
) and adapted the TOC code

```
        public static void AddTof(WordprocessingDocument doc, XElement addBefore, string switches, int? rightTabPos)

            string xmlString =
@"<w:p xmlns:w='http://schemas.openxmlformats.org/wordprocessingml/2006/main'>
  <w:pPr>
    <w:pStyle w:val='TableofFigures'/>
    <w:tabs>
      <w:tab w:val='right' w:leader='dot' w:pos='{0}'/>
    </w:tabs>
    <w:rPr>
      <w:noProof/>
    </w:rPr>
  </w:pPr>
  <w:r>
    <w:fldChar w:fldCharType='begin' dirty='true'/>
    <w:instrText xml:space='preserve'> {1} </w:instrText>
    <w:fldChar w:fldCharType='separate'/>
    <w:fldChar w:fldCharType='end'/>
  </w:r>
</w:p>";
```
Note that to get a List of Tables word seems to do the same but with

```
  let lotCmd = "TOC \\h \\z \\c \"Table\""
```

instead of
```
  let lofCmd = "TOC \\h \\z \\c \"Figure\""
```

[...]

PR #10029, opened 2024-07-25
implement lof and lot for docx

acxz:
lof and lot have been implemented for LaTeX for a while now.
This PR adds support for lof and lot in docx, closing #8245 

In doing so, lof and lot are also exposed to the command line and corresponding documentation is added.

Please if you could review any instances where I have missed documentation regarding lof and lot, I would much appreciate it, @jgm.  I did my best, but I would appreciate another pair of eyes.

A specific question:
Should I be adding in `lof`/`lot` and `lof-title`/`lot-title` here based on the changes that this PR has?
https://github.com/jgm/pandoc/blob/2f36df65e6b8aa9b12551b3773d55a855c24a03d/MANUAL.txt#L3337-L3344

Note: the latex writer prob needs to be modified to use these command line args. This wasn't obvious for me to do. I'll need some help if this change needs to be done in this PR. See: https://github.com/jgm/pandoc/commit/64c7a0ab765eced232c84a9e86064518f786a78c for more info.

Also note that the default.openxml template needs to be modified as well.
(Edit: has been modified)

Example:

Pandoc Markdown File:
```pandoc
---
title: hello
author: Jane Doe
toc: true
lof: true
lot: true
---

# A

Blah

![Image A](placeholder.png)

|a|b|
|-|-|
|c|d|

: Table A

# B

Blah

![Image B](placeholder.png)

|a|b|
|-|-|
|c|d|

: Table B
```

Output Docx File:
image
image


acxz:
@iandol if you get the chance I would love for you to try this PR out.

iandol:
I don't have a build environment for haskell, so I'd need a build for macOS to test...

src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -175,9 +175,19 @@ writeDocx opts doc = do
...
                    ls -> ls
 
+  let lofTitle = case lookupMetaInlines "lof-title" meta of
+                   [] -> stLofTitle defaultWriterState
+                   ls -> ls
+
+  let lotTitle = case lookupMetaInlines "lot-title" meta of
+                   [] -> stLotTitle defaultWriterState
+                   ls -> ls
+
```
jgm, line 185: Instead of this, I'd recommend just using the Translations interface.
```
lofTitle <- B.text <$> translateTerm Translations.ListOfFigures
lotTitle <- B.text <$> translateTerm Translations.ListOfTables
```

src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -175,9 +175,19 @@ writeDocx opts doc = do
...
   let initialSt = defaultWriterState {
           stStyleMaps  = styleMaps
         , stTocTitle   = tocTitle
+        , stLofTitle   = lofTitle
+        , stLotTitle   = lotTitle
         , stCurId      = 20
```
jgm, line 191: Indeed, if we recover these from Translations, then you wouldn't really need to carry them around in state.

src/Text/Pandoc/Writers/Docx/OpenXML.hs
```diff
@@ -171,6 +171,58 @@ makeTOC opts = do
       ])
     ]] -- w:sdt
 
+makeLOF :: (PandocMonad m) => WriterOptions -> WS m [Element]
+makeLOF opts = do
+  let lofCmd = "TOC \\h \\z \\t \"Image Caption\" \\c" :: Text
```
jgm, line 176: Why "Image Caption"?

src/Text/Pandoc/Writers/Docx/OpenXML.hs
```diff
@@ -171,6 +171,58 @@ makeTOC opts = do
...
+  lofTitle <- gets stLofTitle
+  -- Create a separate style for LOF Heading?
```
jgm, line 178: Can we use the same style that is used for the TOC heading?
Otherwise, maybe a separate style makes sense.

src/Text/Pandoc/Writers/Docx.hs
jgm: I note that we currently hard-code  the string "Table of Contents."  We should also use Translations for that.

src/Text/Pandoc/Writers/Docx/OpenXML.hs
acxz: See: https://github.com/jgm/pandoc/blob/edb9e049cae72084cf2cf5bdb6e982ee25a34f24/src/Text/Pandoc/Writers/Docx/OpenXML.hs#L440

tldr, the "Image Caption" style is used for figure captions in docx.

src/Text/Pandoc/Writers/Docx/OpenXML.hs
acxz: We can use the same style.

src/Text/Pandoc/Writers/Docx.hs
acxz: I ended up getting the following error:
```
src/Text/Pandoc/Writers/Docx/OpenXML.hs:151:15: error:
    Not in scope: ‘B.text’
    No module named ‘B’ is imported.
    |
151 |   tocTitle <- B.text <$> translateTerm Translations.TableOfContents
    |               ^^^^^^

src/Text/Pandoc/Writers/Docx/OpenXML.hs:151:40: error:
    Not in scope: data constructor ‘Translations.TableOfContents’
    No module named ‘Translations’ is imported.
    |
151 |   tocTitle <- B.text <$> translateTerm Translations.TableOfContents
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

acxz: I'm sorry @jgm if you could guide me on the API for `Translations` I'd much appreciate it.

jgm: Maybe `B.` isn't needed.  It's Text.Pandoc.Builder; some modules import this qualified as `B`, but maybe not the Docx writer. Check the imports.

https://hackage.haskell.org/package/pandoc-3.3/docs/Text-Pandoc-Translations.html
https://hackage.haskell.org/package/pandoc-3.3/docs/Text-Pandoc-Translations-Types.html

jgm: We already do `setTranslations` at the beginning of the docx writer.
All you need is `translateTerm` and the various terms (in T.P.Translations.Types).

jgm: Actually, you probably don't need `B.text` at all.
Eventually you just want a string `Text` with "Table of Figures" or whatever, to put into the relevant XML attribute.  `translateTerm` will give you that.  `B.text` is only needed if you want to convert that to an `Inlines`.  (I used it because you had used `lookupMetaInlines` which gives an `Inlines`.)

jgm:
```hs
lotTitle <- T.text <$> translateTerm Term.ListOfTables
```
is raising an error.  `translateTerm` will return a Text.  If you want `lotTitle` to be a Text, you can just do
```hs
lotTitle <- translateTerm Term.ListOfTables
```
If you want `lotTitle` to be an Inlines, you'll have to do
```hs
import qualified Text.Pandoc.Builder as B
...
lotTitle <- B.text <$> translateTerm Term.ListOfTables
```
But I think it will work for you if it's just a Text, because you just need to stuff it into an XML attribute.

src/Text/Pandoc/Writers/Docx.hs
acxz: Tried without `Text.Pandoc.Builder` and it gave an error mentioning requiring it to be a `Inline` instead of a `Text`.

acxz:
@jgm Again thanks for your help on this.

Neither of those options will work as we don't want a `Text` or `Inlines`, we need an `Inline`. See the above conversation in your PR review for the kind of errors we are facing. Can you provide more feedback on how to get an `Inline`?

acxz:
@jgm Do you have any ideas? Otherwise I'll revert back to my original solution.

src/Text/Pandoc/Writers/Docx.hs
jgm: yes, sorry, this was an incorrect suggestion on my part; I hadn't seen the whole context.

jgm: To convert an Inlines to [Inline], you just need B.toList.

acxz: Looks like that did the trick! Thanks so much @jgm !

acxz:
Now that we've iterated on the suggestions. Is there anything else left in this PR, before we add the snippet mentioned in the OP regarding `lof` and `lot` in the template repo?

Also would like a confirmation if any more changes are required in the docs/manual due to this change?

There is also the latex issue (i.e. latex using the command line options) as well, but that is prob best handled in a follow up PR.

Again thanks for reviewing this work @jgm

src/Text/Pandoc/Writers/Docx/OpenXML.hs
```diff
@@ -225,13 +278,23 @@ writeOpenXML opts (Pandoc meta blocks) = do
...
                  (fmap (vcat . map (literal . showContent)) . blocksToOpenXML opts)
                  (fmap (hcat . map (literal . showContent)) . inlinesToOpenXML opts)
                  meta
   let context = resetField "body" body
               . resetField "toc"
                    (vcat (map (literal . showElement) toc))
+              . resetField "lof"
+                   (vcat (map (literal . showElement) lof))
+              . resetField "lot"
+                   (vcat (map (literal . showElement) lot))
```
jgm, line 297: You are setting these variables, but I don't see a change to the default openxml template that adds these variables.  So how do they have any effect?

acxz: Just found out that the separate `pandoc-templates` repo is a subtree in this repo. Added a commit to address this.

jgm:
You've tested this and the lot and lof appear correctly in Word documents?

jgm:
I tried to check this out locally, but it seems it needs to be rebased on main.

acxz:
@jgm rebased! The output you see in the OP is the result of this PR. I'd like for you to try it locally (if you can) before we merge it in however, just so we have another user as a datapoint.

jgm:
Thanks. Tested it locally and it works well.  But we do need to make sure that it works the same for LaTeX and ConTeXt, too.  Here's how to do it:

``` diff
diff --git a/MANUAL.txt b/MANUAL.txt
index ba965cf37..2fd2ce152 100644
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -872,18 +872,18 @@ header when requesting a document from a URL:
 `--lof[=true|false]`, `--list-of-figures[=true|false]`
 
 :   Include an automatically generated list of figures (or, in
-    the case of `latex` or `docx`, an instruction to create
-    one) in the output document. This option has no effect
-    unless `-s/--standalone` is used, and it only has an effect
-    on `latex` or `docx` output.
+    some formats, an instruction to create one) in the output
+    document. This option has no effect unless `-s/--standalone`
+    is used, and it only has an effect on `latex`, `context`, and
+    `docx` output.
 
 `--lot[=true|false]`, `--list-of-tables[=true|false]`
 
 :   Include an automatically generated list of tables (or, in
-    the case of `latex` or `docx`, an instruction to create
-    one) in the output document. This option has no effect
-    unless `-s/--standalone` is used, and it only has an effect
-    on `latex` or `docx` output.
+    some formats, an instruction to create one) in the output
+    document. This option has no effect unless `-s/--standalone`
+    is used, and it only has an effect on `latex`, `context`, and
+    `docx` output.
 
 `--strip-comments[=true|false]`
 
diff --git a/src/Text/Pandoc/Writers/ConTeXt.hs b/src/Text/Pandoc/Writers/ConTeXt.hs
index 6c4bab355..9e35c803b 100644
--- a/src/Text/Pandoc/Writers/ConTeXt.hs
+++ b/src/Text/Pandoc/Writers/ConTeXt.hs
@@ -104,6 +104,8 @@ pandocToConTeXt options (Pandoc meta blocks) = do
   mblang <- fromBCP47 (getLang options meta)
   st <- get
   let context =   defField "toc" (writerTableOfContents options)
+                $ defField "lof" (writerListOfFigures options)
+                $ defField "lot" (writerListOfTables options)
                 $ defField "placelist"
                    (mconcat . intersperse ("," :: Doc Text) $
                      take (writerTOCDepth options +
diff --git a/src/Text/Pandoc/Writers/LaTeX.hs b/src/Text/Pandoc/Writers/LaTeX.hs
index 04e0ab3db..edbc40aaf 100644
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -194,6 +194,8 @@ pandocToLaTeX options (Pandoc meta blocks) = do
                     $ lookupMetaInlines "nocite" meta
 
   let context  =  defField "toc" (writerTableOfContents options) $
+                  defField "lof" (writerListOfFigures options) $
+                  defField "lot" (writerListOfTables options) $
                   defField "toc-depth" (tshow
                                         (writerTOCDepth options -
                                               if stHasChapters st
```

Also, I think that before merging it might make sense to squash everything into a single commit.

acxz:
@jgm done!

jgm (changes requested):
I think it looks great!  I'm sorry, one more request.  Can you update your commit message so that it indicates clearly what has changed?  I will need this later to add these items to the changelog.  It is especially important to note the new command-line options and any API changes (new fields in WriterOptions and Opt for example).
API chages should be marked as `[API change]`.

acxz:
Of course, I want to make sure this PR is done properly.

Changed the commit message. Let me know if there is anything else.

jgm:
Thanks!  I rewrote the commit message a bit to fit what I like for the changelog.

acxz:
Tried looking at your previous commits to get a feel for it, but what you wrote is cleaner.

Thanks for taking the cycles working with me to get this merged in, @jgm! I really appreciate it.
Whenever I have a feature I'd like in pandoc, I'd love to come back and contribute again!

jgm:
> Whenever I have a feature I'd like in pandoc, I'd love to come back and contribute again!

I hope you will.

PR #10047, opened 2024-08-01
Fix asciidoc figure with empty title

Maneren:
I ran into this when converting from docbook's `informalfigure` (that is a figure without a title) to asciidoc.

For this input presented as example in docbook docs
```xml
<informalfigure>
  <mediaobject>
    <imageobject>
      <imagedata fileref="watarun.eps"/>
    </imageobject>
  </mediaobject>
</informalfigure>
```
the output includes extra leading `.` that would precede a title but is invalid alone.
```adoc
.
image::watarun.eps[watarun]
```

In this PR I added a simple if check for empty title so it correctly outputs
```adoc
image::watarun.eps[watarun]
```


jgm:
Apologies!  I didn't notice that this was a pull request -- I took it to be an issue, and I fixed it myself.  Your PR was perfectly good, though. I'm sorry for not noticing that.

Maneren:
LMAO. No problem at all.

I started writing an issue but looked at the code and figured it's this simple, so the description very much reads like an issue at the first glace.

Anyway, thank you for such quick response and glad it's fixed :D

Issue #10053, opened 2024-08-04
ODT reference.odt does not contain TableCaption style


iandol:
**Explain the problem.**

When Pandoc has a table with a caption:

```
> pandoc -s -o out.odt

  Right     Left     Center     Default
-------     ------ ----------   -------
     12     12        12            12
    123     123       123          123
      1     1          1             1

Table:  Demonstration of simple table syntax.

Text
```

<img width="652" alt="image" src="https://github.com/user-attachments/assets/3f399859-f7b8-4424-8a9e-8a5c81cdd274">

The ODT does not have a styled caption (it is Default Paragraph Style). The paragraph XML source is styled with `TableCaption`:

```xml
<text:p text:style-name="TableCaption">Demonstration of simple table syntax.</text:p>
```

…however this style is not present in the reference.odt — I checked in Pandoc V3.3

The fix is simple, add this named style. I attach a `custom-reference.odt` where this is done and using this reference the table caption is styled. Here is what I added:

<img width="230" alt="image" src="https://github.com/user-attachments/assets/92a587f8-3a8d-4c48-9bfe-f1f9783e2592">

This is the latest pandoc, generating using: `pandoc -o custom-reference.odt --print-default-data-file reference.odt`

[custom-reference.odt](https://github.com/user-attachments/files/16486988/custom-reference.odt)

I edited custom-reference.odt using Libreoffice:

Version: 24.2.5.2 (AARCH64) / LibreOffice Community
Build ID: bffef4ea93e59bebbeaf7f431bb02b1a39ee8a59
CPU threads: 8; OS: macOS 14.6; UI render: Skia/Metal; VCL: osx
Locale: en-GB (en_GB.UTF-8); UI: en-US
Calc: threaded

I can make this a pull request once I find where reference.odt is on github.

**Pandoc version?**
ε ➪ pandoc -v
pandoc 3.3
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: /Users/ian/.local/share/pandoc
Copyright (C) 2006-2024 John MacFarlane. Web: https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.

iandol:
The added style in styles.xml is this:

```
<style:style style:name="TableCaption" style:family="paragraph" style:parent-style-name="Caption"/>
```

So perhaps adding this line to https://github.com/jgm/pandoc/blob/main/data/odt/styles.xml is enough?

[...]

PR #10055, opened 2024-08-04
Add TableCaption to ODT styles.xml

iandol:
See #10053 -- ODT output currently has no styling for table captions, and yet the writer is assigning the `TableCaption` style to table captions. But the reference.odt doesn't contain that style. We add `TableCaption` to the styles.xml as this is what Pandoc uses in the generated content. My current Libreoffice writes the style xml a bit differently, but I tried to follow the current style.xml formatting.


jgm:
Thanks! the only problem is the overlong line in the commit message (wrap to 80 cols, please).

iandol:
Hm, I cannot rewrite the history of the first commit message (I'm using github's interface), I'll redo the pull from a new branch and close this one...

iandol:
See #10058

Issue #9996, opened 2024-07-17
Move `header-includes` earlier in Typst template?


gordonwoodhull:
Currently `header-includes` in `default.typst` are included after

```typst
#show: doc => conf(
 // ...
)
```

They are also directly before `include-before`:

https://github.com/jgm/pandoc/blob/8613aa39e8b5b8282f30b9fd48b04ec2595ec8eb/data/templates/default.typst#L79-L86

causing these two kinds of include to have the same effect.

Typst doesn't really have "headers", but I argue it would be more more consistent with the HTML behavior to move `header-includes` before the show rule.

Currently, attempting to set the page color or background in `header-includes` will cause the title and abstract to be rendered on a default black-on-white page, followed by a page break, before the rest of the content is rendered with the desired page settings.

Here is an example:

````
---
title: header-includes is too late in Typst template
abstract: |
  The title and abstract will be printed in black on a white page,
  and the rest of the content will be printed in white on a navy page.
header-includes: |
  ```{=typst}
  #set page(fill: navy)
  #set text(fill: white)
  ```
---

# Text on a new page

Because we did

```typst
#set page(fill: navy)
```

after

```typst
#show: doc => conf(
 // ...
)
```

we started a new page with the new background color.
````

Happy to open a pull request with this change, but I figured I'd open a issue to discuss first.

cc @cscheid

jgm:
sounds reasonable to me!

PR #10070, opened 2024-08-08
typst template: move header-includes before show doc

gordonwoodhull:
Please see #9996 for rationale.

The behavior of `header-includes` is untested, at least for Typst. Apparently one test uses `-s` and I saw it fail due to a line break issue (now fixed).

I have tested this for the example in #9996 and would be glad to add tests if appropriate.

cc @cscheid


gordonwoodhull:
Thanks!

Issue #10069, opened 2024-08-08
pandoc-lua can't require binary packages when installed with Conda


iandol:
**Explain the problem.**

This is a weird issue. I've been trying to switch to more isolated workflow, using local environments of tools, and in this case used a conda manager [pixi](https://pixi.sh) to install `lua` `luarocks` and `pandoc` local to a working folder. I used `luarocks` to install a lua debugging package `mobdebug`. Given a minimal `test.lua` script:

```lua
md = require("mobdebug")

print("Hello World")
```

Run with lua it is ok:

```
ζ lua test.lua 
Hello World
```

but run with pandoc-lua, a `mapping process and mapped file (non-platform) have different Team IDs` error occurs:

```
ζ ➪ pandoc-lua test.lua
Error running Lua:
error loading module 'socket.core' from file '/Users/ian/Code/dotpandoc/.pixi/envs/default/lib/lua/5.4/socket/core.so':
	dlopen(/Users/ian/Code/dotpandoc/.pixi/envs/default/lib/lua/5.4/socket/core.so, 0x0006): tried: '/Users/ian/Code/dotpandoc/.pixi/envs/default/lib/lua/5.4/socket/core.so' (code signature in <605C052E-EA8E-36D3-9CC1-598BA8D1AC1E> '/Users/ian/Code/dotpandoc/.pixi/envs/default/lib/lua/5.4/socket/core.so' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs), '/System/Volumes/Preboot/Cryptexes/OS/Users/ian/Code/dotpandoc/.pixi/envs/default/lib/lua/5.4/socket/core.so' (no such file), '/Users/ian/Code/dotpandoc/.pixi/envs/default/lib/lua/5.4/socket/core.so' (code signature in <605C052E-EA8E-36D3-9CC1-598BA8D1AC1E> '/Users/ian/Code/dotpandoc/.pixi/envs/default/lib/lua/5.4/socket/core.so' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs)
stack traceback:
	[C]: in function 'require'
	...de/dotpandoc/.pixi/envs/default/share/lua/5.4/socket.lua:12: in main chunk
	[C]: in function 'require'
	.../dotpandoc/.pixi/envs/default/share/lua/5.4/mobdebug.lua:99: in main chunk
	[C]: in function 'require'
	test.lua:2: in main chunk
```

[...]

PR #10098, opened 2024-08-19
INSTALL.md — reorganise info on static binaries and add conda-forge install options

iandol:
See #10069 -- the static linked pandoc on other OSes also have a limitation to use binary lua packages. So added this point at the top of the document. In addition conda-forge is cross-platform way to install pandoc and other tools, added to Windows, macOS and Linux sections. I wrapped at 75 characters, is that right?

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,8 +7,10 @@ The simplest way to get the latest pandoc release is to use the installer.
   Download the latest installer
 </a>
 
-For alternative ways to install pandoc, see below
-under the heading for your operating system.
+For alternative ways to install pandoc, see below under the heading for
+your operating system. In cases where a statically linked executable is
+installed, the pandoc binary from this package cannot use lua filters that
+require external lua modules written in C.
 
 ## Windows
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -7,8 +7,10 @@ The simplest way to get the latest pandoc release is to use the installer.
   Download the latest installer
 </a>
 
-For alternative ways to install pandoc, see below
-under the heading for your operating system.
+For alternative ways to install pandoc, see below under the heading for
+your operating system. In cases where a statically linked executable is
+installed, the pandoc binary from this package cannot use lua filters that
+require external lua modules written in C.
```
jgm, line 13: I'd suggest breaking this up.  First sentence in its own paragraph. Then a new paragraph:
```
Note: the statically linked pandoc binaries we provide cannot use Lua filters that
require external Lua modules written in C.  If you need to use such filters, choose
another installation method.
```

INSTALL.md
```diff
@@ -70,6 +84,18 @@ Note: On unsupported versions of macOS (more than three releases old),
...
+[`conda`](https://conda.pydata.org/docs/intro.html),
+[`[micro]mamba`](https://mamba.readthedocs.io/en/latest/index.html) or
+[`pixi`](https://prefix.dev). Conda forge contains multiple LaTeX and other
+packages (including `pandoc-citeproc`, `pandoc-plot`, `rsvg-convert` via
+`librsvg` etc.). Note this is a statically-linked executable:
+
+    conda install -c conda-forge pandoc
+    pixi global install pandoc
+    micromamba install pandoc
+
```
jgm, line 98: Rather than repeating this three times, maybe we should have a separate section on installing via Conda.  These repeated paragraphs could then be replaced by something like:
```
Or, you can install pandoc using a [Conda forge][Installing via Conda].
```

INSTALL.md
iandol: I've separated that paragraph. I slightly tweaked wording as previously INSTALL.md only stated the Linux binary packages are statically linked, and it wasn't clear if this limitation applies to windows etc.

INSTALL.md
iandol: Done…

jgm:
thanks!

PR #10123, opened 2024-08-27
Lint Lua using Luacheck

alerque:
This is dependent on the upstream PR adding builtin support for Pandoc to Luacheck, see https://github.com/lunarmodules/luacheck/pull/115. I'll take it out of draft mode when that is in a release ... which could be sooner rather than later if feedback here/there is reasonable. The main motivation for adding the presets is using it on my own filters ... running it on the Pandoc repo was just a handy way to give it a shaking out.

### DIFF
--- a/pandoc-lua-engine/test/bytestring-reader.lua
+++ b/pandoc-lua-engine/test/bytestring-reader.lua
@@ -1,4 +1,4 @@
-function ByteStringReader (input, opts)
+function ByteStringReader (input, _opts)
   local chars = pandoc.List{}
   for i = 1, #input do
     chars:insert(utf8.char(input:byte(i,i)))
--- a/.github/workflows/luacheck.yml
+++ b/.github/workflows/luacheck.yml
@@ -1,0 +1,13 @@
+name: Luacheck
+
+on: [ push, pull_request ]
+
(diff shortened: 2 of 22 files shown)


pandoc-lua-engine/test/bytestring-reader.lua
```diff
@@ -1,4 +1,4 @@
-function ByteStringReader (input, opts)
+function ByteStringReader (input, _)
```
alerque, line 1: Having reviewed and dealt with 100% of the lint errors thrown my `luacheck` on the entire code base, this is the one default lint I'm not so sure is helpful for this project. Since much of the Lua code in this repository is examples that people will be expected to extend rather than use as-is in production, the unused variable and unused argument lints might be counter productive. Thoughts?

jgm: Using `_opts` might be another alternative.

alerque: In my book that's a reasonable convention, and `luacheck` supports it. Some other Lua linters do not (e.g. `lua-language-server` flags it) but I'm still down. I just refactored all the lint handling using that method.

PR #10153, opened 2024-09-04
Add CRediT roles to JATS

### LINKED ISSUE #10152: Add support for CRediT roles in JATS export
Here's some JATS XML that has author roles in it. It would be cool if pandoc supported this!

```xml
<article-meta>
<article-id pub-id-type="doi">10.1002/leap.1210</article-id>
…
<contrib-group>
<contrib contrib-type="author">
    <contrib-id content-type="orcid">0000-0002-9298-3168</contrib-id>
    <name>
        <surname>Allen</surname>
        <given-names>Liz</given-names>
    </name>
    <role vocab="credit" vocab-identifier="https://credit.niso.org/"
        vocab-term-identifier="https://credit.niso.org/contributor-roles/conceptualization/"
        vocab-term="Conceptualization">Conceptualization</role>
    <role vocab="credit" vocab-identifier="https://credit.niso.org/"
        vocab-term-identifier="https://credit.niso.org/contributor-roles/writing-original-draft/"
        vocab-term="Writing &#x2013; original draft">Writing &#x2013; original
        draft</role>
</contrib>
```

This is the template where this would need to get handled:

https://github.com/jgm/pandoc/blob/180d2b5a025362f7fdf20c68a8bc595e0938cb86/data/templates/article.jats_publishing#L92-L129

This is where this needs to get documented:

https://github.com/jgm/pandoc/blob/180d2b5a025362f7fdf20c68a8bc595e0938cb86/doc/jats.md?plain=1#L15

I am happy to try getting this working myself!

cthoyt:
Closes #10152

Enable annotating author roles using the [Contribution Role Taxonomy (CRediT)](https://credit.niso.org/contributor-roles) and export this information in conformant JATS (#10153, Charles Tapley Hoyt and Jez Cope).

## Motivation

I'm motivated to add this to Pandoc since I want the Journal of Open Source Software (JOSS), which is built on top of Pandoc, to be able to create compliant JATS. We're already adding support for encoding this information in article metadata in parallel in https://github.com/openjournals/inara/pull/87.

## Demo

Now, you can specify an authors' role(s) in the metadata like this:

```yaml
title: CRediT Test
author:
  - name: Max Mustermann
    roles:
      - credit: software
        credit-name: Software    # optional, looked up automatically
        degree: Lead             # optional
        name: Programas          # optional, useful for internationalization
```

where the only required key inside each roles dictionary is `credit`. If that's there, then the primary English label can be automatically looked up and put into `credit-name`. If you want to override for internationalization purposes, you can use `name`.

The `degree` field is optional, and can be one of `Lead`, `Supporting`, or `Equal` as defined in the JATS standard.

## Testing

The tests (with narrative documentation) can be found in https://github.com/cthoyt/pandoc/blob/patch-1/test/command/10152.md. Run these tests specifically with `cabal test --test-options="-p 10152.md"`

I also put the 7 outputs in `10152.md` as XML files inside [examples.zip](https://github.com/user-attachments/files/18642767/examples.zip). I tested all of them are valid/conformant JATS using the web-based tool at https://jats4r-validator.niso.org. The zip archive can be downloaded and you can run these tests yourself, if desired.

## Acknowledgements

huge shout-out to John for many rounds of helping me get comfortable with Haskell and the Pandoc codebase, and @jezcope for helping me get across the finish line. This PR would have only been a half-finished idea without them.

### DIFF
--- a/data/templates/article.jats_publishing
+++ b/data/templates/article.jats_publishing
@@ -109,6 +109,17 @@ $elseif(author.name)$
 $else$
 <string-name>$author$</string-name>
 $endif$
+$for(author.roles)$
+$if(it.credit)$
+<role vocab="credit"$if(it.degree)$ degree-contribution="$it.degree$"$endif$
+      vocab-identifier="https://credit.niso.org/"
+      vocab-term-identifier="https://credit.niso.org/contributor-roles/$it.credit$/"
+      vocab-term="$it.credit-name$"
+>$if(it.name)$$it.name$$else$$it.credit-name$$endif$</role>
+$elseif(it.name)$
+<role>$it.name$</role>
+$endif$
+$endfor$
 $if(author.email)$
 <email>$author.email$</email>
 $endif$
--- a/src/Text/Pandoc/Writers/JATS.hs
+++ b/src/Text/Pandoc/Writers/JATS.hs
@@ -27,7 +27,7 @@ import Control.Monad.Reader
 import Control.Monad.State
 import Data.Generics (everywhere, mkT)
 import qualified Data.Map as M
-import Data.Maybe (fromMaybe, listToMaybe)
+import Data.Maybe (fromMaybe, listToMaybe, isNothing)
 import Data.Time (toGregorian, Day, parseTimeM, defaultTimeLocale, formatTime)
 import qualified Data.Text as T
 import Data.Text (Text)
@@ -43,7 +43,7 @@ import Text.DocLayout
 import Text.Pandoc.Shared
 import Text.Pandoc.URI
 import Text.Pandoc.Templates (renderTemplate)
-import Text.DocTemplates (Context(..), Val(..))
+import Text.DocTemplates (Context(..), Val(..), toVal)
 import Text.Pandoc.Writers.JATS.References (referencesToJATS)
 import Text.Pandoc.Writers.JATS.Table (tableToJATS)
 import Text.Pandoc.Writers.JATS.Types
@@ -53,6 +53,54 @@ import Text.Pandoc.XML
 import Text.TeXMath
 import qualified Text.Pandoc.Writers.AnnotatedTable as Ann
 import qualified Text.XML.Light as Xml
+
+-- | Default human-readable names for roles in the Contributor Role
+-- Taxonomy (CRediT). This is useful for generating JATS that annotate
+-- contributor roles
+creditNames :: M.Map Text Text
+creditNames = M.fromList [
+    ("conceptualization", "Conceptualization"),
+    ("data-curation", "Data curation"),
+    ("formal-analysis", "Formal analysis"),
+    ("funding-acquisition", "Funding acquisition"),
+    ("investigation", "Investigation"),
+    ("methodology", "Methodology"),
+    ("project-administration", "Project administration"),
+    ("resources", "Resources"),
+    ("software", "Software"),
+    ("supervision", "Supervision"),
+    ("validation", "Validation"),
+    ("visualization", "Visualization"),
+    ("writing-original-draft", "Writing – original draft"),
+    ("writing-review-editing", "Writing – review & editing")]
+
+-- | Ensure every role with a `credit` key also has a `credit-name`,
+-- using a default value if necessary
+addCreditNames :: Context Text -> Context Text
+addCreditNames context =
+  case getField "author" context of
+    -- If there is an "authors" key, then we replace the existing value
+    -- with one we mutate by running the addCreditNamesToAuthor helper
+    -- function on each
+    Just (ListVal authors) ->
+      resetField "author" (map addCreditNamesToAuthor authors) context
+    -- If there is no "authors" key in the context, then we don't have to do
+    -- anything, and just return the context as is
+    _ -> context
+  where
+    addCreditNamesToAuthor :: Val Text -> Val Text
+    addCreditNamesToAuthor val = fromMaybe val $ do
+      MapVal authorCtx <- pure val
+      ListVal roles <- getField "roles" authorCtx
+      return $ toVal $ resetField "roles" (map addCreditNameToRole roles) authorCtx
+
+    addCreditNameToRole :: Val Text -> Val Text
+    addCreditNameToRole val = fromMaybe val $ do
+      MapVal roleCtx <- pure val
+      guard $ isNothing (getField "credit-name" roleCtx :: Maybe (Val Text))
+      creditId <- getField "credit" roleCtx
+      creditName <- M.lookup creditId creditNames
+      return $ toVal $ resetField "credit-name" creditName roleCtx
 
 -- | Convert a @'Pandoc'@ document to JATS (Archiving and Interchange
 -- Tag Set.)
@@ -159,6 +207,7 @@ docToJATS opts (Pandoc meta blocks') = do
                (lookupMetaInlines "title" meta)
   let context = defField "body" main
               $ defField "back" back
+              $ addCreditNames
               $ resetField "title" title'
               $ resetField "date" date
               $ defField "mathml" (case writerHTMLMathMethod opts of
--- a/test/command/10152.md
+++ b/test/command/10152.md
@@ -1,0 +1,446 @@
+# CRediT Roles
+
+This document contains tests and examples for enabling
+export of roles to JATS. It was added for
+[Issue #10152](https://github.com/jgm/pandoc/issues/10152)
+and corresponding [Pull Request #10153](https://github.com/jgm/pandoc/pull/10153).
+
+In the first example, we show a fully qualified CRediT role.
+An explicit name isn't given, so the CRediT name is used.
+
+```
+% pandoc -s -t jats
+---
+title: CRediT Test
+author:
+  - name: Max Mustermann
+    affiliation: [ 1 ]
+    roles:
+      - credit: software
+        credit-name: Software
+        degree: Lead
+affiliation:
+- id: 1
+  name: Silverlight University
+---
+^D
+<?xml version="1.0" encoding="utf-8" ?>
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2 20190208//EN"
+                  "JATS-archivearticle1.dtd">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.2" article-type="other">
+<front>
+<journal-meta>
+<journal-id></journal-id>
+<journal-title-group>
+</journal-title-group>
+<issn></issn>
+<publisher>
+<publisher-name></publisher-name>
+</publisher>
+</journal-meta>
+<article-meta>
+<title-group>
+<article-title>CRediT Test</article-title>
+</title-group>
+<contrib-group>
+<contrib contrib-type="author">
+<string-name>Max Mustermann</string-name>
+<role vocab="credit" degree-contribution="Lead"
+      vocab-identifier="https://credit.niso.org/"
+      vocab-term-identifier="https://credit.niso.org/contributor-roles/software/"
+      vocab-term="Software"
+>Software</role>
+<xref ref-type="aff" rid="aff-1"/>
+</contrib>
+<aff id="aff-1">
+<institution-wrap>
+<institution>Silverlight University</institution>
+</institution-wrap>
+</aff>
+</contrib-group>
+<permissions>
+</permissions>
+</article-meta>
+</front>
+<body>
+
+</body>
+<back>
+</back>
+</article>
+```
+
+In the second example, we show a fully qualified CRediT role.
+An explicit name is given in a different language.
+
+```
+% pandoc -s -t jats
+---
+title: CRediT Test
+author:
+  - name: Max Mustermann
+    affiliation: [ 1 ]
+    roles:
+      - credit: software
+        credit-name: Software
+        degree: Lead
+        name: Programas
+affiliation:
+- id: 1
+  name: Silverlight University
+---
+^D
+<?xml version="1.0" encoding="utf-8" ?>
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2 20190208//EN"
+                  "JATS-archivearticle1.dtd">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.2" article-type="other">
+<front>
+<journal-meta>
+<journal-id></journal-id>
+<journal-title-group>
+</journal-title-group>
+<issn></issn>
+<publisher>
+<publisher-name></publisher-name>
+</publisher>
+</journal-meta>
+<article-meta>
+<title-group>
+<article-title>CRediT Test</article-title>
+</title-group>
+<contrib-group>
+<contrib contrib-type="author">
+<string-name>Max Mustermann</string-name>
+<role vocab="credit" degree-contribution="Lead"
+      vocab-identifier="https://credit.niso.org/"
+      vocab-term-identifier="https://credit.niso.org/contributor-roles/software/"
+      vocab-term="Software"
+>Programas</role>
+<xref ref-type="aff" rid="aff-1"/>
+</contrib>
+<aff id="aff-1">
+<institution-wrap>
+<institution>Silverlight University</institution>
+</institution-wrap>
+</aff>
+</contrib-group>
+<permissions>
+</permissions>
+</article-meta>
+</front>
+<body>
+
+</body>
+<back>
+</back>
+</article>
+```
+
+In this example, we show a partially qualified CRediT role
+that does not have a `degree`:
+
+```
+% pandoc -s -t jats
+---
+title: CRediT Test
+author:
+  - name: Max Mustermann
+    affiliation: [ 1 ]
+    roles:
+      - credit: software
+        credit-name: Software
+affiliation:
+- id: 1
+  name: Silverlight University
+---
+^D
+<?xml version="1.0" encoding="utf-8" ?>
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2 20190208//EN"
+                  "JATS-archivearticle1.dtd">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.2" article-type="other">
+<front>
+<journal-meta>
+<journal-id></journal-id>
+<journal-title-group>
+</journal-title-group>
+<issn></issn>
+<publisher>
+<publisher-name></publisher-name>
+</publisher>
+</journal-meta>
+<article-meta>
+<title-group>
+<article-title>CRediT Test</article-title>
+</title-group>
+<contrib-group>
+<contrib contrib-type="author">
+<string-name>Max Mustermann</string-name>
+<role vocab="credit"
+      vocab-identifier="https://credit.niso.org/"
+      vocab-term-identifier="https://credit.niso.org/contributor-roles/software/"
+      vocab-term="Software"
+>Software</role>
+<xref ref-type="aff" rid="aff-1"/>
+</contrib>
+<aff id="aff-1">
+<institution-wrap>
+<institution>Silverlight University</institution>
+</institution-wrap>
+</aff>
+</contrib-group>
+<permissions>
+</permissions>
+</article-meta>
+</front>
+<body>
+
+</body>
+<back>
+</back>
+</article>
+```
+
+In this example, we show a more stripped-down data that requires automatic lookup of the
+`credit-name`.
+
+```
+% pandoc -s -t jats
+---
+title: CRediT Test
+author:
+  - name: Max Mustermann
+    affiliation: [ 1 ]
+    roles:
+      - credit: software
+affiliation:
+- id: 1
+  name: Silverlight University
+---
+^D
+<?xml version="1.0" encoding="utf-8" ?>
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Archiving and Interchange DTD v1.2 20190208//EN"
+                  "JATS-archivearticle1.dtd">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.2" article-type="other">
+<front>
+<journal-meta>
+<journal-id></journal-id>
+<journal-title-group>
+</journal-title-group>
+<issn></issn>
+<publisher>
+<publisher-name></publisher-name>
+</publisher>
+</journal-meta>
+<article-meta>
+<title-group>
+<article-title>CRediT Test</article-title>
+</title-group>
+<contrib-group>
+<contrib contrib-type="author">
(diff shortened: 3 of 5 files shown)


cthoyt:
Hi @jgm , I think I figured out making a test file. I have a few questions I bet you could answer very quickly but would take me a long time to figure out:

1. Is there a specific naming convention I should use for my test command file?
2. I created a Credit.hs that contains a dictionary that I want to inject into the JATS output process. Where should I look to do that?

jgm:
> Is there a specific naming convention I should use for my test command file?

Usually we try to name them after the associated issue.  So, if there's an open issue that this fixes you could use that number; otherwise you could use the PR's number, 10153.md.  That makes it easy to figure out later why the test was added.

> I created a Credit.hs that contains a dictionary that I want to inject into the JATS output process. Where should I look to do that? you can see in the diff what I am currently trying in the JATS writer

Instead of creating a separate module, I'd prefer if you just included this map in the JATS writer module itself.

cthoyt:
@jgm thanks for that feedback, I addressed both of your points. Now, the only thing left is how to make this dictionary (renamed to `creditNames`) available inside the template `article.jats_publishing`.


My first try was `$ defField "creditNames" creditNames` because this group is also where some other things that appear in the template are referenced, but I am unfortunately not familiar with Haskell and this is a very complex codebase! Any help appreciated :)

jgm:
> My first try was $ defField "creditNames" creditNames

what happens when you try that?

By the way, our usual convention for variable names would favor `credit-names`.

cthoyt:
It seems the issue was the definition of `creditNames :: M.Map String String` - it didn't seem to have a way to handle strings and gave this error:

```
src/Text/Pandoc/Writers/JATS.hs:182:17: error: [GHC-39999]
    • Could not deduce ‘Text.DocTemplates.Internal.ToContext
                          Text (M.Map String String)’
        arising from a use of ‘defField’
      from the context: PandocMonad m
        bound by the type signature for:
                   docToJATS :: forall (m :: * -> *).
                                PandocMonad m =>
                                WriterOptions -> Pandoc -> JATS m Text
        at src/Text/Pandoc/Writers/JATS.hs:124:1-68
```

Switching the signature to use `Text` instead of `String` in `creditNames :: M.Map Text Text` and finally got it to compile, but it seems like the Text values in the dictionary don't get displayed in the template and instead only evaluate to `true`. Is there either:

1. A way to add handling for regular dictionaries
2. Unpack the Pandoc Text to be a regular string inside the template?

jgm:
I don't understand why you'd be getting this behavior.  Nested Maps seem to work fine in ToContext:  here I explore using `cabal repl` in jgm/doctemplates:
```
ghci> toContext $ fromList [("a"::Text, fromList [("b"::Text, fromList [("c"::Text,"d"::Text)])])] :: Context Text
Context {unContext = fromList [("a",MapVal (Context {unContext = fromList [("b",MapVal (Context {unContext = fromList [("c",SimpleVal (Text 1 "d"))]}))]}))]}
ghci> Right (tpl :: Template Text) <- compileTemplate "." "$a.b.c$" 
ghci> renderTemplate tpl context
Text 1 "d"
```

jgm:
Btw, you can see what instances are defined for ToContext here:
https://hackage.haskell.org/package/doctemplates-0.11.0.1/docs/Text-DocTemplates.html

ToContext typeclass is relevant because of the type of `defField`:
```hs
defField   :: ToContext a b => Text -> b -> Context a -> Context a
```

jgm:
Are you saying that `$credit-names.conceptualization$` is coming out as `true`?

cthoyt:
> Are you saying that `$credit-names.conceptualization$` is coming out as `true`?

Yes! This is what I'm experiencing

jgm:
I'm a bit confused because originally you had a *nested* map and now it seems not to be nested.  With a nested map, in which the value of `$credit-names.conceptualization$` is itself a map, it would render as true in the template.  But with the non-nested map it should render as a string value.

cthoyt:
I thought that the nested map was a bit over-complicated and unnecessary so I simplified it to a regular map. This means that the `$credit-names[it.type]$` becomes `true` and the testing suite gives the error:

```
--- test/command/10152.md
+++ pandoc -s -t jats
+  25       vocab-term="true"
-  25       vocab-term="Software"
+  28     true
-  28     Software
+  33       vocab-term="true"
+  34       
-  33       vocab-term="Methodology"
+  36     true
-  35     Methodology
+  45       vocab-term="true"
-  44       vocab-term="Software"
+  48     true
-  47     Software
```

ref: https://github.com/jgm/pandoc/actions/runs/11021333152/job/30608170051?pr=10153#step:9:1926

jgm:
Ah, I see what you are trying to do.  I don't think `variable[field]` is a valid syntax for doctemplates.

The way to do this is to resolve these names in the writer.  Probably the simplest approach would be to use your Map to add a `name` field to the role in the Context, before you pass it to the template.  (You could do this only if a `name` field isn't already present, which would give the user the ability to override the name, or just do it always.)

estedeahora:
I believe that it may be useful to re-iterate this [comment](https://github.com/jgm/pandoc/issues/10152#issuecomment-2332073367) regarding this issue.

jgm:
Yes, overriding is actually important for people who don't want it to be in English!

cthoyt:
Thanks @jgm and @estedeahora for the feedback. I've opted to make the input YAML format a bit more explicit, so you have to say exactly what the CRediT name is. You also have the option to override, to support the multi-language things.

I also improved the docs to give all of the different possibilities and added to the tests one for each.

estedeahora:
I see that in the last commits you opted to generate the changes by incorporating everything manually in the yaml header and then these in the template (leaving aside the modification of JATS.hs). This is a “correct” solution. However, as you pointed out in #10152 it is an inefficient solution because writing three times variations of the same information can generate errors. I think it would be a contribution to generate this in Haskell, but I don't know how to do it yet. On the other hand, I don't think it is advisable to add this functionality to Pandoc if this point is not solved, because it generates too many fields making it not very concise.

As an alternative and provisional solution I think that users can use a lua filter and “inject” these fields through a dictionary. Then this can be placed in a custom template. I think our discussion provides elements for others to do this if they need to.

jgm:
It should be completely feasible to add the `credit-name` field automatically, from a standard map, if it is missing.  @cthoyt if you can't see how to do this I may be able to help.

cthoyt:
I am feeling like figuring out haskell and the specifics of this codebase might be too much of a lift for me at the moment. @jgm if you are willing to give a hand, then I would be very grateful.

I had squashed the commits earlier, but just added back the name lookup dictionary in `JATS.hs`

jgm:
Here are some tips.  In Text.Pandoc.Writers.Shared you'll find
```hs
-- | Retrieve a field value from a template context.
getField   :: FromContext a b => Text -> Context a -> Maybe b
```
and
```hs
-- | Reset a field of a template context.  If the field already has a
-- value, the new value replaces it.
-- This is a utility function to be used in preparing template contexts.
resetField :: ToContext a b => Text -> b -> Context a -> Context a
```
So the basic template for adding these fields to the `roles` variable would be something like:
```hs
addCreditNames :: Context a -> Context a
addCreditNames context =
  case getField "roles" context of
    Nothing -> context
    Just roles -> resetField "roles" (map addCreditName roles) context

addCreditName :: M.Map Text Text -> M.Map Text Text
addCreditName rolemap =
  case M.lookup "credit-name" rolemap of
    Just _ -> rolemap
    Nothing -> maybe id (M.insert "credit-name")
      (M.lookup "credit-id" rolemap >>= flip M.lookup creditNames)
```
Then you'd put `addCreditNames` in the pipeline that currently has stuff like `defField "blah" blah`.

jezcope:
I'm happy to give this a look, since I'm 100% in favour of getting more useful metadata into the journal ecosystem and getting CRediT roles into more JATS seems like a good thing. Probably won't have time to get properly into it until the weekend though!

cthoyt:
> Then you'd put `addCreditNames` in the pipeline that currently has stuff like `defField "blah" blah`.

Do you mean this section?

https://github.com/jgm/pandoc/blob/56fc7d5cfce2d3a7cb9084533f0579d9c506b21c/src/Text/Pandoc/Writers/JATS.hs#L160-L166

I tried sticking it in the middle like 

```haskell
let context = defField "body" main
              $ defField "back" back
              $ addCreditNames
              $ resetField "title" title'
              $ resetField "date" date
              $ defField "mathml" (case writerHTMLMathMethod opts of
                                        MathML -> True
                                        _      -> False) metadata
```

but this doesn't appear to work correctly - what's the right way to apply it?

jgm:
The thing you said doesn't appear to work should work.

Each line in this pipeline has type `Context Text -> Context Text`.

src/Text/Pandoc/Writers/JATS.hs
```diff
@@ -54,6 +54,64 @@ import Text.TeXMath
...
+    ("project-administration", "Project administration"),
+    ("resources", "Resources"),
+    ("software", "Software"),
+    ("supervision", "Supervision"),
+    ("validation", "Validation"),
+    ("visualization", "Visualization"),
+    ("writing-original-draft", "Writing – original draft"),
+    ("writing-review-editing", "Writing – review & editing")]
+
+addCreditNames :: Context a -> Context a
```
jgm, line 77: This type might need to be changed to `Context Text -> Context Text`.

cthoyt: Okay I think this helped. Now the error is inside the other functions:

```haskell
src/Text/Pandoc/Writers/JATS.hs:79:8: error: [GHC-39999]
    • No instance for ‘Text.DocTemplates.Internal.FromContext
                         Text (M.Map Text Text)’
        arising from a use of ‘getField’
    • In the expression: getField "roles" context
      In the expression:
        case getField "roles" context of
          Nothing -> context
          Just roles -> resetField "roles" (map addCreditName roles) context
      In an equation for ‘addCreditNames’:
          addCreditNames context
            = case getField "roles" context of
                Nothing -> context
                Just roles -> resetField "roles" (map addCreditName roles) context
   |
79 |   case getField "roles" context of
   |        ^^^^^^^^

```

jgm: Look at https://hackage.haskell.org/package/doctemplates-0.11.0.1/docs/Text-DocTemplates.html

You can see that there is a ToContext instance for ToContext b => Map Text b,
but there isn't a similar FromContext instance!  Argh.

Probably I should just add one.  Stand by.

jgm: Well actually we may be able to work around this by operating directly with a Val type instead of a Map.
```hs
addCreditName :: Val Text -> Val Text
addCreditName val =
  case val of
    MapVal ctx ->
      case getField "credit-name" ctx of
        Just _ -> val
        Nothing -> maybe val (MapVal . resetField "credit-name")
          (getFiled "credit-id" ctx >>= flip M.lookup creditNames)
    _ -> val
```
(UNTESTED! But I think something along these lines will work, without the need for changes in doctemplates.)

cthoyt: I tried a couple variations on this. The code in https://github.com/jgm/pandoc/pull/10153#discussion_r1776077467 gave this error:

```haskell
src/Text/Pandoc/Writers/JATS.hs:95:40: error: [GHC-83865]
    • Couldn't match type: Context a0 -> Context a0
                     with: Context Text
      Expected: Text -> Context Text
        Actual: Text -> Context a0 -> Context a0
    • Probable cause: ‘resetField’ is applied to too few arguments
      In the second argument of ‘(.)’, namely ‘resetField "credit-name"’
      In the second argument of ‘maybe’, namely
        ‘(MapVal . resetField "credit-name")’
      In the expression:
        maybe
          val (MapVal . resetField "credit-name")
          (getField "credit-id" ctx >>= flip M.lookup creditNames)
   |
95 |         Nothing -> maybe val (MapVal . resetField "credit-name")
   |     
```

Unfortunately, I don't think it's very effective working this way and I'm not sure how to push this forward. If we agree on the base ideas (extending the documented data model, how the templating for JATS changes, and how the tests look) then I would say we should split this PR into two parts - one that introduces those changes, then a second as a follow-up that uses some haskell magic to help when data is incomplete (i.e. the changes we're talking about in this thread)

@jgm do you think that would work? Otherwise I fear that I might not be able to get the job done and am sucking up too much of your open source time, and/or this thread becomes impossible to follow

jgm: As the message indicates, `resetField` doesn't have enough parameters.  Just a dumb mistake on my part.
```hs
        Nothing -> maybe val (MapVal . resetField "credit-name")
          (getField "credit-id" ctx >>= flip M.lookup creditNames)
```
should be
```hs
        Nothing -> maybe val (MapVal . (\x -> resetField "credit-name" x val)
          (getField "credit-id" ctx >>= flip M.lookup creditNames)
```
Or, more verbose but easier to follow:
```hs
        Nothing ->
           case getField "credit-id" ctx of
             Nothing -> val
             Just creditId ->
               case M.lookup creditId creditNames of
                 Nothing -> val
                 Just creditName -> resetField "credit-name" creditName val
```

cthoyt: alright, I think we're making progress, but now it seems that the way getField is defined is causing the issue:

```haskell
src/Text/Pandoc/Writers/JATS.hs:93:12: error: [GHC-39999]
    • Ambiguous type variable ‘b0’ arising from a use of ‘getField’
      prevents the constraint ‘(Text.DocTemplates.Internal.FromContext
                                  Text b0)’ from being solved.
      Probable fix: use a type annotation to specify what ‘b0’ should be.
      Potentially matching instances:
        instance Text.DocTemplates.Internal.TemplateTarget a =>
                 Text.DocTemplates.Internal.FromContext a a
          -- Defined in ‘Text.DocTemplates.Internal’
        instance Text.DocTemplates.Internal.TemplateTarget a =>
                 Text.DocTemplates.Internal.FromContext a (Doc a)
          -- Defined in ‘Text.DocTemplates.Internal’
        ...plus two others
        (use -fprint-potential-instances to see them all)
    • In the expression: getField "credit-name" ctx
      In the expression:
        case getField "credit-name" ctx of
          Just _ -> val
          Nothing
            -> case getField "credit-id" ctx of
                 Nothing -> val
                 Just creditId
                   -> case M.lookup creditId creditNames of
                        Nothing -> ...
                        Just creditName -> ...
      In a case alternative:
          MapVal ctx
            -> case getField "credit-name" ctx of
                 Just _ -> val
                 Nothing
                   -> case getField "credit-id" ctx of
                        Nothing -> val
                        Just creditId -> ...
   |
93 |       case getField "credit-name" ctx of
   |            ^^
```

jgm: >       Probable fix: use a type annotation to specify what ‘b0’ should be.

This is your hint.  `getField` is polymorphic, so we may need to explicitly specify the type of the intended result, if it's not pinned down by other uses.

in this case, `getField "credit-name" ctx` is ambiguous because we don't do anythnig with the `Just _` return value ,so it could be any type.  Thus, we need to say what the type is that we expect:

```
Just (_ :: Text) -> val
```
if we want to leave it alone if it's a text value, or
```
Just (_ :: Val Text) -> val
```
if we want to leave it alone if it's any kind of value.

cthoyt: Woohoo! It has a few warnings but managed to compile! I added a corresponding test to make sure that it works (i.e., if there's a `credit-id` and no `credit-name`, add the right `credit-name`.) This isn't passing yet, so this needs some more invesigation

jgm: At line 92 you match Just (MapVal _) but you don't match anything else.
You need a case for the catch-all "other" clause: `_ -> val`.
See my original example at https://github.com/jgm/pandoc/pull/10153#discussion_r1776077467

cthoyt: hi @jgm sorry I had to step away from this because I felt like I had hit a wall, but I realized an issue with the way my filtering was working was it didn't match the data structure (I was looking for a roles list, but the only thing that was available was an authors list, and there was a roles list in each)

I have (I hope) a final issue with the compiler, again related to me not understanding the typing system enough to annotate the cases correctly. Could you please give me another hint?

```
src/Text/Pandoc/Writers/JATS.hs:79:8: error: [GHC-39999]
    • No instance for ‘Text.DocTemplates.Internal.FromContext
                         Text (Context Text)’
        arising from a use of ‘getField’
    • In the expression: getField "authors" context
      In the expression:
        case getField "authors" context of
          Nothing -> error "No Authors!"
          Just authors
            -> resetField
                 "authors" (map addCreditNamesToAuthor authors) context
      In an equation for ‘addCreditNames’:
          addCreditNames context
            = case getField "authors" context of
                Nothing -> error "No Authors!"
                Just authors
                  -> resetField
                       "authors" (map addCreditNamesToAuthor authors) context
   |
79 |   case getField "authors" context of
   |        ^^
```

jezcope:
@cthoyt @jgm What's the status of this patch now? Can I help in any way?

cthoyt:
@jezcope hello! That's a very kind offer of you. The PR is nearly finished but I am caught on the last part of navigating how to add the default CRediT name when the ID is available - unfortunately this is because of my lack of comfortability with haskell.

You can check out my branch and run `cabal test --test-options="-p 10152.md"` to see which tests pass, and exactly what is left to do. I think someone with a bit more knowledge than me can easily fix this. If it's not possible to understand the state of things, I am happy to hop on a call and maybe we can work it out together.

src/Text/Pandoc/Writers/JATS.hs
jgm: I'm sorry, I guess I never answered your question here.

The FromContext typeclass is from doctemplates:
https://hackage.haskell.org/package/doctemplates-0.11.0.1/docs/Text-DocTemplates.html#t:FromContext

A `FromContext Text (Context Text)` instance would allow you to do
```
lookupContext :: Text -> Context Text -> Maybe (Context Text)
```
The compiler thinks you're trying to use this instance, but it doesn't exist.

So let's go back.  In `resetField "authors" (map creditNamseToAuthor authors) context`, what is the type of `addCreditNamesToAuthor`?  Knowing that would tell us what type the compiler is inferring for `authors`, and that can help explain why the compiler is assuming you're using this instance of FromContext. Hope that's intelligible, feel free to ask if not.

One thing that can be helpful is to add explicit type annotations to make sure that you're not making false assumptions about the types.  E.g.
```
Just (authors :: NameOfTypeHere) ->
```

jgm: Sorry I didn't follow up.
```hs
-- | Retrieve a field value from a template context.
getField   :: FromContext a b => Text -> Context a -> Maybe b
```
So when you do `getField "authors" context`, what is the compiler assuming `b` will be here?
`context` should be `Context Text`.  Judging from the compiler's message, the compiler thinks `b` should be `Context Text`...since it is complaining that there isn't an instance `FromContext Text (Context Text)`.

Is that what you want `authors` to be?  The compiler would be inferring this from the use you make of `authors`, i.e. `map addCreditNamesToAuthor authors`.  What is the type of `addCreditNamesToAuthor`?

jgm: I often debug this sort of issue by inserting explicit type annotations, e.g. `authors :: TheTypeIThinkItShouldBe`, which forces the compiler to give a more useful error message.

jezcope:
Ok, cool, I will try and get pandoc building probably this weekend and see if I can figure out what needs to be done. I'm on NixOS so I might end up addressing #10509 first... 😅

jezcope:
Ok, I have it compiling and passing the tests with only fairly minor changes, and I've taken the liberty of rebasing it onto current `main`. You can take a look at it on my fork https://github.com/jezcope/pandoc/tree/feature/jats-credit

I don't think I can push to this PR without @cthoyt giving me access, so do I need to open a new PR?

Also I tidied up the new code a little bit with the `Maybe` monad while I was figuring out how everything fits together, but I'm still not familiar enough with the codebase to know if this fits in style-wise. The 3 new functions are now almost exactly the same shape so are crying out for a refactor too.

cthoyt:
@jezcope wow, thank you so much! I have rebased your branch back into mine, and waiting for CI to pass now. After, I think this is ready for a final review

jezcope:
Don't know if it's useful but I've now pushed up another commit with a small refactor and some doc comments

cthoyt:
@jezcope it looks like the recent commit you pushed broke CI, can you check that please?

jezcope:
@cthoyt Oh yeah, I missed that warning, fixed now (and it makes the code clearer too, good compiler). I should have remembered to build with `-Werror` before committing. Funny that it only breaks the Windows build though.

data/templates/article.jats_publishing
```diff
@@ -109,6 +109,22 @@ $elseif(author.name)$
 $else$
 <string-name>$author$</string-name>
 $endif$
+$if(author.roles)$
```
jgm, line 112: I don't think you need this outer `if`.  If `author.roles` is not defined, `$for(author.roles)$...$endfor$` just won't do anything (iirc).

test/command/10152.md
```diff
@@ -0,0 +1,403 @@
...
+<contrib contrib-type="author">
+<string-name>Max Mustermann</string-name>
+<role vocab="credit"
+      vocab-identifier="https://credit.niso.org/"
+      vocab-term-identifier="https://credit.niso.org/contributor-roles/software/"
+      vocab-term="Software"
+      degree-contribution="Lead"
+>
+    Software
+</role>
```
jgm, line 55: Just checking: is it okay to have this whitespace around "Software"?

test/command/10152.md
```diff
@@ -0,0 +1,403 @@
...
+<article-title>CRediT Test</article-title>
+</title-group>
+<contrib-group>
+<contrib contrib-type="author">
+<string-name>Max Mustermann</string-name>
+<role vocab="credit"
+      vocab-identifier="https://credit.niso.org/"
+      vocab-term-identifier="https://credit.niso.org/contributor-roles/software/"
+      vocab-term="Software"
+
```
jgm, line 188: It would be nice if this blank line weren't produced.

jgm:
It looks good to me now. @tarleb do you want to have a look?  You know more about JATS than I.

cthoyt:
@jgm I am working through getting the whitespace fixed and tests passing, but overall the idea is done!

data/templates/article.jats_publishing
cthoyt: Thanks, I addressed this!

test/command/10152.md
cthoyt: I think the answer is yes, since I ran the output through a JATS validator, but just to be safe, I got rid of this whitespace

test/command/10152.md
cthoyt: Agreed, I had some difficulty figuring out how to make this work nicely with leading whitespace and newlines inside the if/else, so I reorganized the order to make sure we don't get ugly new lines

cthoyt:
Whitespace issues addressed, and commits have all been squashed.

jezcope:
Cool, thanks @tarleb. I've been wanting to contribute more to `pandoc` in any case, so I'm happy to take a look at those suggestions as further pull requests. :)

jgm:
One more request: can you amend your commit message so that it includes a brief description of the change (wrapped to 80 columns) and `Closes #10152.`?
This is needed when I create the changelog.

cthoyt:
@jgm done in b9e3d0da7

jgm:
Thanks!

cthoyt:
I'm so excited for this to be merged! Thank you again @jgm and @jezcope for all of the excellent feedback and assistance and @estedeahora for the helpful discussion

Issue #6704, opened 2020-09-27
Using custom highlight style (.theme file) not working


pck1980:
pandoc exits with error when using a custom theme file, even when it's the unmodified output of --print-highlight

```
PS C:\test> pandoc --print-highlight kate > k.theme
PS C:\test> pandoc --highlight-style k.theme
Could not read highlighting theme k.theme
```

There's a different error message if the file does not exist, so it seems pandoc is unable to correctly parse it.

```
PS C:\test> pandoc --version
pandoc.exe 2.10.1
Compiled with pandoc-types 1.21, texmath 0.12.0.2, skylighting 0.8.5
[...]
```

Perhaps similar to issue #4133?

jgm:
Hm, I cannot reproduce this. Maybe it works on non-windows platforms but not on windows?

jgm:
One thing to check would be whether k.theme has CRLF line endings.
If so, does it help to change to LF line endings?
If not, does it help to change to CRLF line endings?

jgm:
We just use aeson ToJSOn/FromJSON instances to parse and render the styles as themes, so it's hard to see how line endings could be mishandled.  But I'm at a loss as to what else could be going wrong, without being able to reproduce. Does k.theme look okay? Can you upload it?

pck1980:
I just tried to use LF endings instead of CRLF, but with no success. Attached are both versions of the exported theme.
[k.zip](https://github.com/jgm/pandoc/files/5290451/k.zip)

tarleb:
The files seem to use the UTF-16 encoding, which is the default on Windows, while pandoc expects all input to use UTF-8. Can you try to convert your files with the methods described [in this answer](https://superuser.com/a/1163765)?

pck1980:
Converting the theme file to UTF-8 does indeed solve the problem. Thank you for your help.

Perhaps pandoc could print out a more helpful error message when it is provided with UTF-16 encoded input.

tarleb:
Maybe we could also support the `-o`/`--output` option to redirect the output of `--print-highlighting-style`. This should mostly avoid the problem for Windows users.

jgm:
The problem is that we are passing a bytestring to aeson's `eitherDecode` function.

It does return an error message in this case, but it's so useless that we don't print it: "Error in $: Failed reading: not a valid json value."  It would be nice if aeson reported that the input was not properly encoded, but it doesn't  work that way.

@tarleb it does already work with `--output`; however, note the following from the manual:
[...]

PR #10192, opened 2024-09-18
Suggest documentation for syntax highlighting issues on Windows

yakatz:
Suggest documentation to preemptively address user issues like #6704

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1,4 +1,5 @@
----
+---byte-or
+
 title: Pandoc User's Guide
 author: John MacFarlane
 date: September 9, 2024
@@ -7262,7 +7263,7 @@ use `--print-highlight-style` to generate a JSON `.theme` file which
 can be modified and used as the argument to `--highlight-style`. To
 get a JSON version of the `pygments` style, for example:
 
-    pandoc --print-highlight-style pygments > my.theme
+    pandoc -o my.theme --print-highlight-style pygments
 
 Then edit `my.theme` and use it like this:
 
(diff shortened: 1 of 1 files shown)


jgm:
Instead of this, I'd suggest changing
```
    pandoc --print-highlight-style pygments > my.theme
```
to
```
    pandoc -o my.theme --print-highlight-style pygments
```

(Indeed, if there is anywhere else in the manual where we recommend shell redirection (`>`), we might want to consider a similar change.)

MANUAL.txt
```diff
@@ -7277,6 +7277,9 @@ definitions](https://github.com/KDE/syntax-highlighting/tree/master/data/syntax)
 
 To disable highlighting, use the `--no-highlight` option.
 
+If you receive an error that pandoc `Could not read highlighting theme`,
+check that the JSON file is encoded with UTF-8 and has no Byte-Order Mark (BOM).
+
```
jgm, line 7282: This paragraph may no longer be necessary (given the change to the example), but if it is kept it should go before the previous paragraph, because it concerns `--syntax-definition`.   Also I'd suggest using double quotes:
```
"Could not read highlighting theme"
```

Are you sure that a BOM causes a problem?  That would surprise me; generally we have set things up so that a BOM is ignored, and if there's somewhere it isn't, we might want to take a look at that.

yakatz: I have tested and BOM is still an issue in a theme file.

You mentioned it in this 2021 comment: https://github.com/jgm/pandoc/issues/6704#issuecomment-832075056

Paragraph order changed

jgm: Oh yes, I see now "the BOM is illegal in JSON."  So, this comment is worth making.

jgm:
thanks!

PR #10244, opened 2024-10-01
Support short forms in Beamer

twsh:
I added variables for short forms of the:

* title
* subtitle
* author
* institute
* date


twsh:
I realise now that empty short authors (and other short forms?) have been OK in beamer only quite recently: https://github.com/jolars/moloch/issues/11

Is that a reason to move the brackets inside the conditionals?

twsh:
OK, this is done.

PR #10245, opened 2024-10-01
Move section numbering code and enable for beamer

twsh:
This means that `--number-sections` will now number sections in beamer.


twsh:
I'm not sure why this fails some test. I tried to update the relevant 'writer' files.

jgm:
Look at the test output; it will show you what is expected (looks like a line moved to a different location).

twsh:
I fixed the test, I just need to work out how to shorten the commit message that's too long.

jgm:
Don't worry, I'll squash and edit the commit msg.

Issue #10263, opened 2024-10-06
The background of a epub book on kindle is a grey color that makes it annoying to read the book


thewhitetulip:
**Explain the problem.**
Include the **exact command line** you used and **all inputs necessary to reproduce the issue**.  Please create as minimal an example as possible, to help the maintainers isolate the problem.  Explain the output you received and how it differs from what you expected.

The problem is that for each paragraph page we are having a grayish background color that makes it really ugly to read books. I looked over the internet and certain people have found out a work around - to remove the background colour property. Hence I thought of raising this bug and doing a code fix myself. Even for my novel that I am generating via pandoc I have to give a custom css just for removing the background colour.

**Pandoc version?**
What version of pandoc are you using, on what OS? (If it's not the latest release, please try with the latest release before reporting the issue.)
this applies to all pandoc version because the issue is in the style.css in the data folder.

thewhitetulip:
This is fixed in https://github.com/jgm/pandoc/pull/10264

There are a bunch of other formatting issues, will submit a PR if you are open to accepting it.

tarleb:
Could you say a bit more about the effect of this? A link to the resources you mentioned would be quite useful.

thewhitetulip:
![image](https://github.com/user-attachments/assets/51d6e5e1-09cc-4096-83a0-7939c26716ad)
If you look at this image, we get a grey box in the background due to the link that I removed.

thewhitetulip:
Will you be open to merging this? Also the default css stylesheet makes the epub book almost unreadable on Kindle. I use KOREADER. I was planning to edit the css file so that the default epub looks pretty

jgm:
I agree that this looks bad, and in view of that removing the default background color in the CSS seems a good idea.  People can always add it if they want it.  But we shouldn't have default CSS that looks bad on the most popular e-reader.  (Note: my understanding is that the Kindle converts EPUBs to its own native format, and perhaps there is a problem in that conversion, but it really doesn't matter for our purposes if that's where the problem lies.)

thewhitetulip:
Yep. What happens is the epub generated by pandoc is giving a really bad styling the to book. Amazon converts epub to azw3 format and the default css mattwrs quite a bit.

I have seen that for android or Linux ebook readers this doesn't matter because they outright ignore the css 😂

If you are willing to accept another PR. I would love to send one by doing some R&D on a decent default css that would be great to read on Kindle.

tarleb:
[...]

PR #10264, opened 2024-10-06
Update epub.css

thewhitetulip:
removed grayish background that made it irritating to read on kindle.


thewhitetulip:
solves #10263

PR #10265, opened 2024-10-06
MANUAL.txt: fix typo in template syntax

Wandmalfarbe:
That seems to me to be a typo:

```
${months[, ]}$

${articles:bibentry()[; ]$
```

> To mark variables and control structures in the template,
either `$`...`$` or `${`...`}` may be used as delimiters.
The styles may also be mixed in the same template, but the
opening and closing delimiter must match in each case.


jgm:
thanks

PR #10280, opened 2024-10-09
adding support for creating letters with template

jasonpearcy:
adding support for creating letters with template

This change:
1. Adds the latex.letter template to data/templates
2. Updates pandoc.cabal to include the new template
3. Updates MANUAL.txt to include documentation on how to use pandoc to create letters

The latex.letter template supports creating letters with the "letter" class and KOMA-Script classes (both the scrlttr2 class or using the scrletter package).

Detailed examples are published here: https://github.com/jasonpearcy/pandoc_letters

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3154,6 +3154,56 @@ These variables function when using BibLaTeX for [citation rendering].
 `natbiboptions`
 :   list of options for natbib
 
+#### LaTeX Letters
+
+Pandoc uses these optional varibles when creating a letter.
+
+`cc`
+:   a list of carbon copy recipients
+
+`closing`
+:   the closing of the letter, such as `Sincerely,`
+
+`closing-indentation`
+:   a length value used to indent the closing and the signature
+
+`encl`
+:   a list of enclosures
+
+`letterhead-file`
+:   the filename of the letterhead image
+
+`letteroption`
+:   a list of lco files to use for formatting when using one of the
+    [KOMA-Script] classes
+
+`opening`
+:   the opening of the letter, such as `Dear John,`
+
+`ps`
+:   the post script text to include
+
+`return-address`
+:   the address of the letter writer
+
+`signature-after`
+:   the amount of vertical space to add after the printed signature
+
+`signature-before`
+:   the amount of vertical space to add before the signature image
+
+`signature-file`
+:   the filename of the signature image
+
+`scrletter`
+:   a variable that when defined will load the scrletter package   
+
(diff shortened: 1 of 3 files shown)


twsh:
I noticed one thing that I think is a problem, and I have a few other thoughts.

The problem is that the template still has code for `\maketitle` and an abstract environment. Those don't work with scrlttr2, and don't make sense for a letter anyway, in my view.

My thoughts:

* I think some people will want to set the author and signature independently, e.g., author is 'Jane Smith' and signature is 'Professor J. Smith'.
* Should some more things, like email address and phone number, have variables? Maybe the template could also support those two for letter.cls?
* Is there a way to have the documentclass default to a letter class? I don't know how to do that, but it would be nice.
* The KOMA letter classes have fold marks and a return address that are based on the assumption that the letter is going in a windowed envelope. Should they be turned off by default? (I think of them as like the navigation symbols in Beamer or the monospaced URLs in LaTeX: not wanted by most people.)

twsh:
Regarding adding more variables, I think this would allow the user to set any KOMA variable/option in a neat way:

```
$for(komavar)$
\setkomavar{$it.key$}{$for(it.value)$$it$$sep$\\$endfor$}
$endfor$

$for(komaoption)$
\KOMAoption{$it.key$}{$it.value$}
$endfor$

$if(komaoptions)$
\KOMAoptions{$for(komaoptions)$$it$$sep$,$endfor$}
$endif$

$for(letteroption)$
\LoadLetterOption{$it$}
$endfor$
```

I don't think we need both the `KOMAoption` and `KOMAoptions` version (and if so they should have different names). I think that they could be passed as class options anyway, so maybe we need neither?

We could also do things like this for the common things people want in a letter:

```
$if(email)$
\setkomavar{fromemail}{$email$}
\KOMAoption{fromemail}{true}
$endif$
```

It also occurred to me that, as well as setting the signature KOMA variable, `$author$` could set the 'fromname' KOMA variable.

tarleb:
Hi @jasonpearcy! My apologies for letting this sit around for so long.

One of the reasons that this hasn't been merged is that it's unclear if this belongs in the main pandoc code base. My gut feeling is that this should live as an extension and should therefore its own project in a separate directory.

A well maintained letter template that uses KOMA would be a useful addition to the pandoc ecosystem. Hopefully, the recent changes to the default LaTeX template, and the creation of more template partials, will hopefully make it easier to maintain the template.

Please let us know if we can help in the process, and please don't miss posting to the discussion board after the template has been published.

Issue #10294, opened 2024-10-16
typst 0.12 two column layout changes


jarnowic:
See https://github.com/typst/typst/issues/5215 for background.

I'm about to propose two pulls to fix the issue, but I don't know if that may be premature; in any event, an expert's opinion is surely welcome.

jgm:
My question would just be whether `#set page(columns: 2)` works only on typst 0.12 or also on earlier versions.  Currently 0.11 is our official target...I haven't even looked at the 0.12 changelog yet.

jarnowic:
> My question would just be whether `#set page(columns: 2)` works only on typst 0.12 or also on earlier versions. Currently 0.11 is our official target...I haven't even looked at the 0.12 changelog yet.

let me check...

jarnowic:
> > My question would just be whether `#set page(columns: 2)` works only on typst 0.12 or also on earlier versions. Currently 0.11 is our official target...I haven't even looked at the 0.12 changelog yet.
> 
> let me check...

Update: one parameter (`scope`) of the `place` function is not yet properly documented, and it does not work in typst 0.11; therefore, it is better not merging any change until 0.12 becomes stable (it's still rc 2).

I'll be doing tests in the 0.12.rc series and keep on reporting.

Cheers!

jarnowic:
I have confirmation (in the original thread) that this 

> > > My question would just be whether `#set page(columns: 2)` works only on typst 0.12 or also on earlier versions. Currently 0.11 is our official target...I haven't even looked at the 0.12 changelog yet.
> > 
> > 
> > let me check...
> 
> Update: one parameter (`scope`) of the `place` function is not yet properly documented, and it does not work in typst 0.11; therefore, it is better not merging any change until 0.12 becomes stable (it's still rc 2).
> 
> I'll be doing tests in the 0.12.rc series and keep on reporting.
> 
> Cheers!

I have confirmation (in the typst thread) that this change is upcoming with 0.12; so there is no need to make any change on the pandoc code thus far.

jarnowic:
Typsty 0.12 was released today. I am updating my patches for the templates to accomodate to the new layout commands, awaiting for release for upcoming Pandoc 3.5 series.

PR #10324, opened 2024-10-23
Issue10294 new code for two column layout on typst 0.12.x 

jarnowic:
Typst 0.12.0 introduces some changes for two column layout, preferring `#page(columns: <N>)` for global two column output, and supplementing `#place` with new parameters `scope` and `clearance` to layout one column content (such as titleplage) on two column documents. This patch tries to upgrade the `typst.template` to comply with the newer requirements. Check [issue 10294](https://github.com/jgm/pandoc/issues/10294).


jgm:
template.typst has a `set page` with other attributes. Why not put the `columns` there rather than in default.typst?

jarnowic:
> template.typst has a `set page` with other attributes. Why not put the `columns` there rather than in default.typst?

because `default.typst` is the piece of code where the variables are accessed to from the YAML headers or the command line, iiuc. 

to tell the truth, i am still not quite sure why so many defaults are set in `template.typst`; it seems to me that only the variables to be used in customizations in the typst documents should be declared in the typst template, while the document layout decisions might be delayed to the document itself. looks to me as if `template.typst` should work not much unlike `latex.fmt`, just declaring the variables available, while all layout defaults should be moved into `default.typst` to give room for customization, not much unlike LaTeX classes and packages. my apologies for the comparison, but that's the model i am more familiar with; `groff_ms` is still, not much unlike `typst`, work in progress on my side.

so, if i were to suggest something, i'd say that the layout code should be moved into `default.typst` from `template.typst` to allow for more customizations, and not the other way around.

jgm:
> template.typst is out of reach, so pandoc variables cannot affect it.

Not so.  You can put variables in partials!

jarnowic:
> > template.typst is out of reach, so pandoc variables cannot affect it.
> 
> Not so. You can put variables in partials!

you're right; and doing so keeps the change less intrusive than the other one. proceeding.

jarnowic:
> > > template.typst is out of reach, so pandoc variables cannot affect it.
> > 
> > 
> > Not so. You can put variables in partials!
> 
> you're right; and doing so keeps the change less intrusive than the other one. proceed

> > > template.typst is out of reach, so pandoc variables cannot affect it.
> > 
> > 
> > Not so. You can put variables in partials!
> 
> you're right; and doing so keeps the change less intrusive than the other one. proceeding.

I think it is ready now...

data/templates/default.typst
```diff
@@ -31,12 +31,12 @@ $endif$
 $if(smart)$
 $else$
 #set smartquote(enabled: false)
-
 $endif$
+
```
jgm, line 35: Why did you move the endif here?  Putting the space before the endif is a feature: it means that we'll only get the additional blank line if we get the `#set smartquote` line.  Putting it after will give us a blank line even if there is nothing.

data/templates/default.typst
```diff
@@ -31,12 +31,12 @@ $endif$
...
 $for(header-includes)$
 $header-includes$
-
 $endfor$
+
```
jgm, line 39: Ditto for endfor.

data/templates/template.typst
```diff
@@ -35,55 +35,50 @@
     paper: paper,
     margin: margin,
     numbering: "1",
-  )
+    columns: cols,
+    )
+
```
jgm, line 40: why the blank line?  we don't have one elsewhere, e.g. btw 33 and 34

data/templates/template.typst
```diff
@@ -35,55 +35,50 @@
...
            size: fontsize)
   set heading(numbering: sectionnumbering)
 
-  if title != none {
-    align(center)[#block(inset: 2em)[
-      #text(weight: "bold", size: 1.5em)[#title]
-      #(if subtitle != none {
+  place(top + center, float: true, scope: "parent", clearance: 4mm)[
+    #if title != none {
+      text(weight: "bold", size: 1.5em)[#title]
```
jgm, line 50: Can you say something about the rationale for this change in the way the title is formatted?

data/templates/template.typst
```diff
@@ -35,55 +35,50 @@
...
 
-  if cols == 1 {
-    doc
-  } else {
-    columns(cols, doc)
-  }
+    #if abstract != none {
+      text(weight: "semibold")[Abstract]
+      h(1em) 
+      abstract
```
jgm, line 80: Can you confirm that this gives the same visual appearance as the previous template?  Before we had a block with 2em inset.  The `#block` has been removed.  What effect does that have?

data/templates/default.typst
jarnowic: fair enough: reverted

data/templates/default.typst
jarnowic: reverted, as for smartquote

data/templates/template.typst
jarnowic: it was somewhat of an aid for me to debug the code; didn't seem to do any harm, but now I see the underlying rationales.

jarnowic: it was there, in the previous version, between the page settings and the titlepage code.

jarnowic: actually, it was there in the first place, it was just somewhat hidden from view in the change file...

data/templates/template.typst
jarnowic: sure: in versions <0.12 column placement was somewhat assumed to be global on layout; with typst >=0.12 it is explicitly controlled by page settings, with an explicit "scope", so the previous code placed only the main text in two columns, while the footnotes were still single column. Changing the page layout to double column had the side effect to set the title page also in double column (it was moved to the first, left column in my test). This new code is the current standard way to set a block of text single spaced, outside the scope of the double column page layout.

data/templates/template.typst
jarnowic: I believe it's all cleaned up now...

jgm:
looks good - I'll merge this and update the tests.

Issue #10298, opened 2024-10-17
Issue setting a variable of type Dictionary from the command line


ZoomRmc:
I have a trouble setting up a variable of a type Dictionary from the command line.
From what I see in the docs, there's only [one such variable](https://pandoc.org/MANUAL.html#variables-for-typst), namely the `margin` for typst.

There's no problem setting it with the defaults YAML, but I'm not sure if pandoc currently can parse a dictionary from the command line. 
The issue is exacerbated by a total lack of any error feedback on incorrectly formatted options from the CLI.

The following syntaxes should be equal:
* CLI:   ```-V margin={left:1cm, right:1cm, bottom:1cm, top:1cm, x:0cm, y:0cm}```
* `defaults.yaml`:

  ```yaml
  variables:
    margin:
      left: 1cm
      right: 1cm
      bottom: 1cm
      top: 1cm
      x: 0cm
      y: 0cm
  ```

Tested on latest release (3.5).

ZoomRmc:
>  This needs to be made clearer in the manual under --variable

That's more or less clear from the docs and from the code. But having a variable of a type that cannot be set is a a disconnect between the API as documented and the implementation and requires at least a note "can only be set from the YAML" for the specific variable.

PR #10342, opened 2024-10-30
CLI: treat variables with value `true` or `false` as boolean

tarleb:
Variables with boolean-looking values are now converted to booleans when
defined via the command line. This matches the existing behavior for
metadata values.

Closes: #10340

### DIFF
--- a/src/Text/Pandoc/App/CommandLineOptions.hs
+++ b/src/Text/Pandoc/App/CommandLineOptions.hs
@@ -364,9 +364,9 @@ options =
     , Option "V" ["variable"]
                  (ReqArg
                   (\arg opt -> do
-                     let (key, val) = splitField arg
+                     let (key, val) = second readBoolOrString $ splitField arg
                      return opt{ optVariables =
-                                  setVariable (T.pack key) (T.pack val) $
+                                  setVariable (T.pack key) val $
                                     optVariables opt })
                   "KEY[:VALUE]")
                  ""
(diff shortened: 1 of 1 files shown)


jgm:
I suppose this issue is also related to #10298. One could imagine a special syntax that says: interpret this variable value as JSON.

iandol:
> Being able to give a boolean value of `false` could be useful to override the same value in the metadata, but I admit that it's not a particularly likely scenario.

In fact this is really really useful IMO. I often have a defaults file or other metadata for general use, but then I want to quickly override the defaults for a one-off document. See 

https://github.com/jgm/pandoc/discussions/10301 

for some examples, e.g. I usually set `colorlinks` to true, but occasionally want to override this. At present I have to keep making new defaults files for each possible permutation of metadata state as command-line values cannot override them.

I wonder if we can have some other workaround to set either a string or a boolean, how about @ to identify a string: `pandoc -M colorlinks=@true -t native`?

tarleb:
We could allow types to be specified with `/type`, e.g. `-V colorlinks/bool:true`. This wouldn't interfere with any current syntax, is reminiscent of template "pipes", and would allow to also solve #10298 with `-V 'complexvar/json:{ "this is json": true}'.

jgm:
Well, if we had `/json`, we wouldn't need `/bool`, because `false` is a perfectly good JSON value.

alerque:
The first of those makes way more sense to me and I dare say would be a lot easier to cope with generating commands programmatically (something I do a lot of): `--variable/json foo=false`.

jgm:
Closing this as I have now added `--variable-json`.

Issue #6622, opened 2020-08-19
Pandoc failing to run filters on Mac OS 10.13


samfearn:
I've written a couple of filters which I've shared with some colleagues. The filters run fine on my machine (Mac OS 10.15), and I had a colleague test them on their machine too. However, one colleague is now reporting problems running pandoc with the config file I gave them. They get the following error:
~~~
dyld: lazy symbol binding failed: Symbol not found: ____chkstk_darwin 
  Referenced from: /usr/local/bin/pandoc (which was built for Mac OS X 
10.15) 
  Expected in: /usr/lib/libSystem.B.dylib 

dyld: Symbol not found: ____chkstk_darwin 
  Referenced from: /usr/local/bin/pandoc (which was built for Mac OS X 
10.15) 
  Expected in: /usr/lib/libSystem.B.dylib 

Abort trap: 6 
~~~
Modifying the config `.yaml` file I gave them, we've tracked the problem down to the use of the lua filters I wrote. If we remove the `filters: ...` definition in the `.yaml` defaults file, the error doesn't occur.

Has anyone else come across this problem? Does anyone have any suggestions?

Edit: FWIW, I'm running pandoc version 2.10 and they're running version 2.10.1.

tarleb:
Can you check if the error occurs for all Lua filters? Does it fail for a filter file which just contains `print 'Hello!'`, or even an empty file?

It would also be important to know which method you and you colleague used to install pandoc, i.e., whether you installed via brew or downloaded the installer from the "releases" site.

samfearn:
I've contact my colleague to ask him to check. My config `.yaml` was loading four lua filters, and I already got him to check each one individually, and it failed with any one of these, but I didn't test a blank filter, or one with only a print statement, so I'll get him to try that too and report back.

samfearn:
He installed pandoc via the installer from the 'releases' site. It fails for any filter, even a single print statement or an empty file.
~~~
pandoc -L filters/empty.lua test.tex -o file.html
dyld: lazy symbol binding failed: Symbol not found: ____chkstk_darwin
  Referenced from: /usr/local/bin/pandoc (which was built for Mac OS X 10.15)
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: ____chkstk_darwin
  Referenced from: /usr/local/bin/pandoc (which was built for Mac OS X 10.15)
  Expected in: /usr/lib/libSystem.B.dylib

Abort trap: 6
~~~

[...]

PR #10348, opened 2024-11-03
GitHub workflow 'nightly': bump Mac OS version

deining:
The macOS 12 Actions runner image [will be fully unsupported](https://github.com/actions/runner-images/issues/10721) by 12/3/24. This PR addresses this by bumping os version to latest version 15.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -69,7 +69,7 @@ jobs:
 
   macos:
 
-    runs-on: macos-12
+    runs-on: macos-15
 
     steps:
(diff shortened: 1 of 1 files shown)


tarleb (commented):
LGTM, but maybe we should consider #6622 before merging.

jgm:
It's best to build on the oldest version we can (for compatibility), but in this case we don't have much of a choice.

tarleb:
The oldest supported version is now macos-13, should we use that?

jgm:
Yes, I'll change it to 13.
Besides, ghcup is missing on 15!

PR #10350, opened 2024-11-03
JATS: correct suppress attribute

deining:
For `title` element, JATS defines the attribute [`suppress`](https://jats.nlm.nih.gov/extensions/bits/tag-library/2.0/attribute/suppress.html). In the source code, `suppress` is misspelled (one single `p`only) therefore code will not work as intended. This PR fixes that issue.


jgm:
Thanks!

PR #10376, opened 2024-11-11
Respect empty LineBlock lines in plain and ansi writers

silby:
The plain writer behaved as a markdown variant with Ext_line_blocks turned off, and so empty lines in a line block would get eliminated.  This is surprising, since if there's anything where the intent can be preserved in plain text output it's empty lines.

It's still a bit surprising to have nbsps in plain text output, as in the test, where the distinction doesn't really matter, but that'd be an orthogonal change.

As presented here the change ignores whether the `line_blocks` extension is enabled. As a write-only format, I'm not sure whether it makes sense for the `plain` writer to behave differently in the presence or absence of this extension.

Ditto ANSI writer.

### DIFF
--- a/src/Text/Pandoc/Writers/ANSI.hs
+++ b/src/Text/Pandoc/Writers/ANSI.hs
@@ -124,7 +124,11 @@ blockToANSI opts (Plain inlines) = inlineListToANSI opts inlines
 
 blockToANSI opts (Para inlines) = inlineListToANSI opts inlines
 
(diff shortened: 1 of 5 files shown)


silby:
An alternative here is that possibly `T.P.Shared.linesToPara` should become `linesToParas` and split paragraphs on empties in the list of lines, but I haven't studied that.

jgm:
Good idea.

Issue #10311, opened 2024-10-20
EXIF orientation is ignored


xi:
To reproduce this issue, I used the images from https://github.com/recurser/exif-orientation-examples/ (see exact markdown below).
In the [generated pdf](https://github.com/user-attachments/files/17449480/test.pdf), the images do not have the correct orientation. I get the same result when converting to odt or docx.

Tested with both pandoc 3.1.11.1 (debian trixie) and pandoc 3.5 (docker)

```md
![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_0.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_1.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_2.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_3.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_4.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_5.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_6.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_7.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Landscape_8.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Portrait_0.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Portrait_1.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Portrait_2.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Portrait_3.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Portrait_4.jpg)

![](https://raw.githubusercontent.com/recurser/exif-orientation-examples/refs/heads/master/Portrait_5.jpg)

[...]

PR #10386, opened 2024-11-14
Respect Exif orientation in LaTeX, Docx, ODT output

silby:
This branch uses EXIF orientation metadata in image files to affect the output of images in Docx, ODT, and LaTeX writers. Addresses #10311 for these three writers only.

Having worked on this for a few days I am a bit dubious that this in scope for Pandoc. On the one hand, it's probably somewhat common for users to encounter images with a flipped/rotated orientation, given the behavior smartphone cameras and so on, and common to not know what's going on or how to fix it when you generate a PDF with Pandoc. On the other hand, the problem has to be addressed in a specific way for each output format and the results can questionable.

The LaTeX case is particularly bad here, I think. For one thing, it introduces a need to read the image bytes in the LaTeX writer where it was previously unnecessary, and so I'm just ignoring all `fetchItem` exceptions and defaulting to no rotation. There's also no _necessary_ relationship between the environment where `pandoc` is being run and the environment where LaTeX is being run: if someone has a rotated `foo.jpg` in their working directory when they run pandoc, then they execute LaTeX in a directory with an unrotated `foo.jpg`, have we done the right thing or not?

In the ODT case, there's no good-looking solution available without actually rotating the pixels of the image data. See the commit message for cc73c01.

I think the approach here is more or less reasonable but I don't wholeheartedly endorse it.

Incidentally, typst natively respects EXIF orientation, seemingly by actually rotating and flipping the pixels, [see here](https://github.com/typst/typst/blob/main/crates/typst-library/src/visualize/image/raster.rs#L158-L176)

### DIFF
--- a/src/Text/Pandoc/ImageSize.hs
+++ b/src/Text/Pandoc/ImageSize.hs
@@ -21,6 +21,7 @@ module Text.Pandoc.ImageSize ( ImageType(..)
                              , sizeInPixels
                              , sizeInPoints
                              , desiredSizeInPoints
+                             , rotatedDesiredSizeInPoints
                              , Dimension(..)
                              , Direction(..)
                              , Flip(..)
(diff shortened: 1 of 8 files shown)


src/Text/Pandoc/ImageSize.hs
```diff
@@ -451,3 +495,32 @@ webpSize opts img =
   case AW.parseOnly pWebpSize img of
     Left _   -> Nothing
     Right sz -> Just sz { dpiX = fromIntegral $ writerDpi opts, dpiY = fromIntegral $ writerDpi opts}
+
+imageTransform :: ByteString -> ImageTransform
```
jgm, line 499: `imageTransform` seems too general a name for a function that just operates on exif.
Also, all exported functions and types should have Haddock comments.

silby:
Closing this. I didn't really believe in this too strongly by the time I pushed it up and that hasn't changed in the time I've allowed it to age.

PR #10388, opened 2024-11-15
Text.Pandoc.Format: remove duplicate typst entry

alerque:
If .typ is typst, but .typ is also typst, which one wins?


tarleb:
:+1:

PR #10397, opened 2024-11-21
Add WebP support to ImageSize

### LINKED ISSUE #9561: pandoc metadata "cover-image" doesn't recognize WebP image
**Explain the problem.**
Problem:
pandoc metadata "cover-image" doesn't recognize WebP image. It should.

The command line

```
	pandoc test_webp.html -o test_webp.epub \
       --metadata cover-image="front_cover.webp"

```
generates the warning message

```
	[WARNING] Could not determine image size for front_cover.webp: could not determine image type

```
It'd be nice if the problem causing this warning could be fixed.

**Pandoc version?**
This warning is seen when using pandoc 3.1.12.2
under Windows 10 Pro x64 22H2 build 19045.4046
with Cygwin 3.5.1

The .webp image file was created using ImageMagick 7.0.10-61
with the command 
```
convert front_cover.jpg -resize 300x300 front_cover.webp

```
so I think the format of the WebP file itself is OK.

The resulting .epub document does show the cover page image when displayed using the epub reader provided with Calibre 7.6. which again suggests that the file format is OK.

The attached zip file contains the following files which can be used to reproduce the warning message.

test_webp.sh
test_webp.html
front_cover.webp

test_webp.epub
readme.txt

[test_webp.zip](https://github.com/jgm/pandoc/files/14564270/test_webp.zip)

silby:
My code feels a bit messy and I've tested it with like precisely one WebP image per variant (lossy, lossless, and extended). I found myself having/wanting to combine attoparsec (to get `Alternative` and backtracking) and Data.Binary.Get (to get numbers); maybe there's a less awkward looking way to do this.

This is an API change because the constructors of ImageType are public.

Closes #9561

### DIFF
--- a/src/Text/Pandoc/ImageSize.hs
+++ b/src/Text/Pandoc/ImageSize.hs
@@ -39,7 +39,7 @@ import qualified Data.ByteString.Lazy as BL
 import Data.Binary.Get
 import Data.Bits ((.&.), shiftR, shiftL)
 import Data.Word (bitReverse32)
-import Data.Maybe (isJust, fromMaybe)
+import Data.Maybe (isJust, fromJust)
 import Data.Char (isDigit)
 import Control.Monad
 import Text.Pandoc.Shared (safeRead)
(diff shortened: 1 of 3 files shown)


tarleb:
This code is quite low-level, which makes me think that it would better be placed in a different package. Would extending JuicyPixels be an option?

silby:
It doesn’t really make sense to drive-by contribute this code to JuicyPixels. It’s hardly the foundation for a full WebP codec. If one were to create a library that just extracts the dimensions of various image sizes it would make sense there but I have no particular interest in doing that. Bitwise operators aside, it’s no more low-level than emfSize and pdfSize. On Nov 22, 2024, at 3:07 AM, Albert Krewinkel ***@***.***> wrote:﻿
This code is quite low-level, which makes me think that it would better be placed in a different package. Would extending JuicyPixels be an option?

jgm:
I think this is fine; in the past we used all kinds of low-level code to find image sizes; JuicyPixels allowed us to remove a lot of this low-level stuff, but if we still need it for webp, okay.

PR #10434, opened 2024-12-04
Parse base64-encoded data URIs more efficiently

silby:
(in some places)

Very long data: URIs in source documents are causing outsized memory usage due to various parsing inefficiencies, for instance in Network.URI, TagSoup, and T.P.R.Markdown.source. See e.g. #10075.

This change improves the situation in a couple places we can control relatively easily by using an attoparsec text-specialized parser to consume base64-encoded strings. Attoparsec's takeWhile + inClass functions are designed to chew through long strings like this without doing unnecessary allocation, and the improvements in peak heap allocation are significant.

One of the observations here is that if you parse something as a valid data: uri it shouldn't need any further escaping so we can short-circuit various processing steps that may unpack/iterate over the chars in the URI.

The code here is organized a little bit randomly and I expect it needs more work to go in.

Good improvements though. The peak heap allocation for html -> markdown for the example provided by @ebeigarts in #10075 goes down from ~2900MB to ~2500 MB on my computer, and markdown -> json goes from 1577 MB to 73 MB! As discussed in the issue comments HTML reading has, at least, a TagSoup issue.

### DIFF
--- a/src/Text/Pandoc/Parsing/Base64.hs
+++ b/src/Text/Pandoc/Parsing/Base64.hs
@@ -1,0 +1,39 @@
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{- |
+Module      : Text.Pandoc.Parsing.Base64
+Copyright   : © 2024 Evan Silberman
+License     : GPL-2.0-or-later
+Maintainer  : John MacFarlane <jgm@berkeley.edu>
+
+Parse large base64 strings efficiently within Pandoc's
+normal parsing environment
+-}
+
+module Text.Pandoc.Parsing.Base64
+  ( parseBase64String )
+
+where
+
+import Data.Text as T
+import Data.Attoparsec.Text as A
+import Text.Parsec (ParsecT, getInput, setInput, incSourceColumn)
+import Text.Pandoc.Sources
+import Control.Monad (mzero)
+
+parseBase64String :: Monad m => ParsecT Sources u m Text
+parseBase64String = do
+  Sources ((pos, txt):rest) <- getInput
+  let r = A.parse pBase64 txt
+  case r of
+    Done remaining consumed -> do
+      let pos' = incSourceColumn pos (T.length consumed)
+      setInput $ Sources ((pos', remaining):rest)
+      return consumed
+    _ -> mzero
+
+pBase64 :: A.Parser Text
+pBase64 = do
+  most <- A.takeWhile1 (A.inClass "A-Za-z0-9+/")
+  rest <- A.takeWhile (== '=')
+  return $ most <> rest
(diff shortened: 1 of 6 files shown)


src/Text/Pandoc/Parsing/Base64.hs
```diff
@@ -0,0 +1,39 @@
...
+      let pos' = incSourceColumn pos (T.length consumed)
+      setInput $ Sources ((pos', remaining):rest)
+      return consumed
+    _ -> mzero
+
+pBase64 :: A.Parser Text
+pBase64 = do
+  most <- A.takeWhile1 (A.inClass "A-Za-z0-9+/")
+  rest <- A.takeWhile (== '=')
+  return $ most <> rest
```
jgm, line 39: Two thoughts on this:

1.  Is attoparsec really necessary?  Why not just `Data.Text.takeWhile`?
2.  My experience is that parsers like this, which just manipulate the input directly using `getInput` and `setInput`, are problematic in parsec because parsec doesn't realize that input has been consumed.  I've had to use a regular parsec parser somewhere in there to make it realize this.  One option is just something like `count charsConsumed anyChar`, and then you don't need to compute the end position manually...

silby: 1. In this spot I can possibly just borrow the fast `inClass` function from attoparsec and use it with regular text `takeWhile`, will have to fiddle with it.
2. will investigate, I took it for granted that I would make the parsec state happy by fiddling with the input as seen here but that was not based on deep understanding or rigorous analysis.

jgm: It's probably fine to use attoparsec, but there might be a slight speedup if you can avoid it.

On 2: you could try putting this parser under `many` and see if parsec complains.

src/Text/Pandoc/Readers/Markdown.hs
```diff
@@ -1822,13 +1822,35 @@ parenthesizedChars = do
...
+pBase64DataURI = mconcat <$> sequence
+  [ (textStr "data:")
+  , (T.singleton <$> (letter <|> digit))
+  , restrictedName
+  , (T.singleton <$> char '/')
+  , restrictedName
+  , textStr ";" <* trace "cool"
+  , (mconcat <$> many mediaParam)
+  , textStr "base64," <* trace "fine"
+  , parseBase64String
```
jgm, line 1835: What if we replace l.1835 with
```
   many1Char (satisfy (A.inClass "A-Za-z0-9+/"))
,  manyChar (char '=')
```
and get rid of the special T.P.Parsing.Base64 module.  What is the impact on performance on the sorts of files that were problematic before?  I'd like to try the simplest possible thing before worrying about possible optimizations.

jgm:
I borrowed some code from this in a more minimal commit, 1a8da4f4dbf1a7327fa9abf1331ab506cbe9b3e8
which leads to big improvements.  Still need to investigate the html rendering side of things.

silby:
Harvested and improved by @jgm in other commits; closing

PR #10457, opened 2024-12-14
Miscellaneous refactors

josephcsible:


src/Text/Pandoc/Writers/ANSI.hs
```diff
@@ -82,7 +82,7 @@ pandocToANSI opts (Pandoc meta blocks) = do
   body <- blockListToANSI opts blocks'
   notes <- gets $ reverse . stNotes
   let notemark x = D.literal (tshow (x :: Int) <> ".") <+> D.space
-  let marks = take (length notes) $ map notemark [1..]
+  let marks = map notemark [1..length notes + 1]
```
jgm, line 85: Is the `+ 1` here right?  I don't see why.  Previously `marks` was a list of the same length as `notes`.  With this change it will be 1 longer than `notes`.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -927,25 +927,21 @@ pMath inCase = try $ do
   let attr = toStringAttr attr'
   unless inCase $
     guard (maybe True (== mathMLNamespace) (lookup "xmlns" attr))
-  let isDisplay = case lookup "display" attr of
-                     Just "block" -> True
-                     _ -> False
+  let possiblyDisplayMath = case lookup "display" attr of
```
jgm, line 930: Instead of `possiblyDisplayMath`, I think `constructor` would be better.

jgm (changes requested):
Looks good except for the two things noted!

src/Text/Pandoc/Writers/ANSI.hs
josephcsible: Whoops, you're right. Fixed.

src/Text/Pandoc/Readers/HTML.hs
josephcsible: Done.

jgm:
thanks!

PR #10471, opened 2024-12-18
Move nocite inside frame

twsh:
I should have put the nocite code inside a frame. Otherwise, beamer ignores it.


jgm:
I rebased it for you and merged it.

PR #10478, opened 2024-12-20
Mention typst in PandocUnknownWriterError for pdf

silby:


silby:
come to think of it tho I'm not sure I can actually get this message to be printed?

jgm:
This might actually be something that used to have a function but doesn't any more.
Pandoc seems happy with `-t pdf` now...

jgm:
See https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/App/OutputSettings.hs#L74-L76
https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/App/OutputSettings.hs#L308-L309

Issue #8416, opened 2022-11-03
Better error message when trying to use output-only format for input


lassik:
**Describe your proposed improvement and the problem it solves.**

```
$ pandoc hello.adoc -t hello.html
Unknown input format asciidoc
```

Pandoc can write asciidoc, but cannot read it. It's confusing that the error message says the asciidoc format is "unknown" to pandoc. Could the message be altered to say "asciidoc is supported only as an output format, not for input" or something similar?

**Describe alternatives you've considered.**

Leave the error message as is.

foursixnine:
I just created: https://github.com/jgm/pandoc/pull/10492 to improve the error message

PR #10492, opened 2024-12-23
Improve message for asciidoc input error #8416

foursixnine:
I noticed that asciidoct is not supported as input format, so in this case its
better to be clear about what the failure is about.

```
fursixnine@pakhet cndk8 % pandoc -t markdown  ./CNDK8.adoc $OUTPUT.md
Unknown input format asciidoc
```

This should improve the error being presented to the user.

### DIFF
--- a/src/Text/Pandoc/Error.hs
+++ b/src/Text/Pandoc/Error.hs
@@ -124,6 +124,7 @@ renderError e =
                  "\nTry using Word to save your DOC file as DOCX," <>
                  " and convert that with pandoc."
         "pdf" -> "\nPandoc can convert to PDF, but not from PDF."
+        "asciidoc" -> "\nPandoc can convert to asciidoc, but not from asciidoc."
         _     -> ""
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Error.hs
```diff
@@ -124,6 +124,8 @@ renderError e =
                  "\nTry using Word to save your DOC file as DOCX," <>
                  " and convert that with pandoc."
         "pdf" -> "\nPandoc can convert to PDF, but not from PDF."
+        "asciidoc" -> "\nPandoc can convert to asciidoc, but not from " <>
+                      "\n asciidoc (sometimes adoc also)."
```
jgm, line 128: I'd prefer a simpler one liner "Pandoc can convert to asciidoc, but not from asciidoc."

foursixnine: @jgm done

jgm:
thanks!

PR #10498, opened 2024-12-26
Add Pod reader

silby:
Pod ("Plain old documentation") is a markup languaged used principally to document Perl modules and programs. Since it was originally meant to be translated pretty directly to man, the semantics are fairly simple.

This Pod reader was developed with reference to the canonical user and implementer documentation of Pod: https://perldoc.perl.org/perlpod and https://perldoc.perl.org/perlpodspec.

There are 1490 .pod, .pl, and .pm in the Perl 5.34 distribution found in /System/Library/Perl on my mac. Of those, this reader dies with a parse error on 7 of them. All of them seem to be cases where pod commands are found within a non-colon-prefixed =begin/=end. perlpodspec says I may treat this as an error.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -269,6 +269,7 @@ header when requesting a document from a URL:
     - `odt` ([OpenOffice text document][ODT])
     - `opml` ([OPML])
     - `org` ([Emacs Org mode])
+    - `pod` (Perl's [Plain Old Documentation])
     - `ris` ([RIS] bibliography)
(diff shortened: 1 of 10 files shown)


jgm:
Looks good, as far as I can see, the only thing missing is a mention of `pod` in the appropriate section of MANUAL.txt.

silby:
gah I love forgetting to do that

jgm:
thanks!

PR #10532, opened 2025-01-13
fix changelog: pod is reader, not writer

mjgardner:
I was excited to read in the latest changelog that pandoc could now write Perl's POD format. But upon further examination, it looks like v3.6.2 can only read POD, not write it.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,7 @@
 
 ## pandoc 3.6.2 (2025-01-12)
 
-  * New output format: `pod` (Evan Silberman). Pod ("Plain old documentation")
+  * New input format: `pod` (Evan Silberman). Pod ("Plain old documentation")
     is a markup languaged used principally to document Perl modules and
     programs.
 


silby:
Indeed. If you have an existing workflow that would benefit from a Pod writer I’d be glad to hear about it though!

mjgardner:
> Indeed. If you have an existing workflow that would benefit from a Pod writer I’d be glad to hear about it though!

I don’t personally have an existing workflow, but I can imagine teams collaborating on documentation using a wiki or Google Docs and then exporting to one of Pandoc’s readers for inclusion in a Perl module or distribution.

It might also prove helpful with incorporating benchmarks, test coverage, and source code profiling. ([Devel::Cover](https://metacpan.org/dist/Devel-Cover/view/bin/cover#REPORT-FORMATS) and [Devel::NYTProf](https://metacpan.org/pod/Devel::NYTProf#nytprofhtml) both report to HTML.)

mjgardner:
Thanks for merging. Would you mind also updating the changelog for GitHub’s [release 3.6.2](https://github.com/jgm/pandoc/releases/tag/3.6.2)? It still describes Pod as an output format.

PR #10544, opened 2025-01-15
Consume blanks after =encoding in pod reader

### LINKED ISSUE #10537: POD reader: A directive immediately following an `=encoding` directive is treated as text
Given a minimal input POD document like this

```pod
=encoding utf8

=head1 NAME

Test document
```

the recently released pandoc 3.6.2, run with `pandoc -f pod Minimal.pod` (on Linux), outputs

```html
<p>=head1 NAME</p>
<p>Test document</p>
```

instead of the expected

```html
<h1>NAME</h1>
<p>Test document</p>
```

Placing a normal paragraph after the `=encoding`  yields an expected result, i.e.

```pod
=encoding utf8

Dummy paragraph.

=head1 NAME

Test document
```

outputs

```html
<p>Dummy paragraph.</p>
<h1>NAME</h1>
<p>Test document</p>
```

However, adding an extra blank line instead, as in

```pod
=encoding utf8


=head1 NAME

Test document
```

leads to a parsing error:

```shell
Error at "Minimal.pod" (line 4, column 1):
unexpected '\n'
expecting Pod paragraph

^
```

which does not seem reasonable.

silby:
I neglected to properly consume empty lines after =encoding commands, which produced various incorrect parses depending on the content between there and the next command. If I had actually looked closely at the golden test I originally generated I might have noticed the problem!

Fixes #10537

### DIFF
--- a/src/Text/Pandoc/Readers/Pod.hs
+++ b/src/Text/Pandoc/Readers/Pod.hs
@@ -112,6 +112,7 @@ encoding :: PandocMonad m => PodParser m Blocks
 encoding = do
   cmd "encoding"
   anyLine
+  optional blanklines
   logMessage $ IgnoredElement "=encoding; Pandoc requires UTF-8 input"
   return mempty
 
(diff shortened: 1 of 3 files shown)


jgm:
thanks!

PR #10557, opened 2025-01-21
Prefer MIME type when determining extensions for MediaBag items

### LINKED ISSUE #10553: LaTeX writer: remote SVG images with non-`.svg` URI extensions are not treated as SVGs
**Explain the problem.**
The LaTeX writer relies on the path extension of an image to determine whether to use `\includegraphics` or `\includesvg`:
https://github.com/jgm/pandoc/blob/ba04a9987dd8b2251f93231dbf9aa378fd7ac001/src/Text/Pandoc/Writers/LaTeX.hs#L1125

This breaks in instances where the URI of an image does not reference `.svg`. Consider, for instance, the following image:
```md
![](https://img.shields.io/github/actions/workflow/status/jgm/pandoc/ci.yml)
```
(Rendered: )

Passing this to pandoc yields an error:
```sh
$ echo '![](https://img.shields.io/github/actions/workflow/status/jgm/pandoc/ci.yml)' | pandoc -o test.pdf
[WARNING] Could not convert image /private/var/folders/_n/dwfw0_316w554fyv3q9t7m1w0000gn/T/tex2pdf.-7a6e25991b71d6c9/6e49fa4c4ffe43659ba3f4187ea0145ff5768d7a.yml: Cannot load file
  Jpeg Invalid marker used
  PNG Invalid PNG file, signature broken
  Bitmap Invalid Bitmap magic identifier
  GIF Invalid Gif signature : <svg x
  HDR Invalid radiance file signature
  Tiff Invalid endian tag value
  TGA not enough bytes
Error producing PDF.
! LaTeX Error: Unknown graphics extension: .yml.

 ...

l.69 ...fa4c4ffe43659ba3f4187ea0145ff5768d7a.yml}}
```

Relevant code and possible solutions:
- When the image is added to the media bag, the `.yml` extension is preferred over the mimetype.
   https://github.com/jgm/pandoc/blob/ba04a9987dd8b2251f93231dbf9aa378fd7ac001/src/Text/Pandoc/MediaBag.hs#L110-L112
   For remote images, this could be changed to prefer mimetype over any extension contained in the URI, at least if the extension in the URI does not correspond to an image. Then, the image would be stored as `...8d7a.svg` and the extension check would be sufficient.
- The LaTeX writer could check the mimetype of the image in the mediabag instead of checking the extension of the URI: https://github.com/jgm/pandoc/blob/ba04a9987dd8b2251f93231dbf9aa378fd7ac001/src/Text/Pandoc/Writers/LaTeX.hs#L1082

**Pandoc version?**
pandoc 3.6.1 MacOS

max-heller:
Currently, remote images added to the MediaBag are stored at paths with extensions determined based on the external URI. For instance, an image from https://example.com/image.png is stored as <hash>.png. If the URI does not contain an extension (e.g., https://example.com/image), then the content-type of the downloaded image is used to determine the extension.

This change switches the precedence such that content-type is preferred over extensions contained in the URI. This is necessary because some images are located at URIs with misleading extensions -- shields.io, for instance, serves SVGs from URIs with .yml extensions. With this change, the image/svg+xml content-type is now preferred over the .yml URI extension. This fixes a bug in the PDF writer in which such an image would be mishandled due to not being identified as an SVG.

Closes #10553

### DIFF
--- a/src/Text/Pandoc/MediaBag.hs
+++ b/src/Text/Pandoc/MediaBag.hs
@@ -107,9 +107,11 @@ insertMedia fp mbMime contents (MediaBag mediamap)
                   _     -> getMimeTypeDef fp''
   mt = fromMaybe fallback mbMime
   path = maybe fp'' (unEscapeString . uriPath) uri
-  ext = case takeExtension path of
-          '.':e | '%' `notElem` e -> '.':e
-          _ -> maybe "" (\x -> '.':T.unpack x) $ extensionFromMimeType mt
+  ext = case extensionFromMimeType mt of
+             Just e -> '.':T.unpack e
+             Nothing -> case takeExtension path of
+                             '.':e | '%' `notElem` e -> '.':e
+                             _ -> ""
 
 -- | Lookup a media item in a 'MediaBag', returning mime type and contents.
 lookupMedia :: FilePath
--- a/test/Tests/MediaBag.hs
+++ b/test/Tests/MediaBag.hs
@@ -29,7 +29,7 @@ tests = [
         assertBool "file in directory is not extracted with original name" exists1
         exists2 <- doesFileExist ("foo" </> "f9d88c3dbe18f6a7f5670e994a947d51216cdf0e.jpg")
         assertBool "file above directory is not extracted with hashed name" exists2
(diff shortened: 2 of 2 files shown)


jgm:
thanks!

PR #10563, opened 2025-01-24
bracket tables with typst:no-figure and typst:text attributes

gordonwoodhull:
The combination of #9648 `typst:text:*` property output and #9778 `typst:no-figure` can cause fonts, text sizes, etc. to spill out of tables.

This is because setting Typst text properties across a table requires `#set text(...)` outside the table, and previously we were relying on the figure to provide a scope.

Example:

```html
<p>Paragraph before.</p>

<table typst:text:size="3em">
  <tr>
    <td>A</td>
    <td>B</td>
    <td>C</td>
  </tr>
</table>

<p>Paragraph after. Oops.</p>
```

produces

```typst
Paragraph before.

#set text(size: 3em); #table(
  columns: 3,
  align: (auto,auto,auto,),
  [A], [B], [C],
)
Paragraph after. Oops.
```
image

This PR provides the minimal fix, only adding an extra `#[...]` when the table has class `typst:no-figure` and also has `typst:text:*` attributes.

We work around this bug in Quarto in quarto-dev/quarto-cli#11113, but it would be better to have the fix in Pandoc.

@cscheid


jgm:
This looks good.  But instead of
```
#[#set text(size: 3em); #table(
  columns: 3,
  align: (auto,auto,auto,),
  [A], [B], [C],
)]
```
why not
```
#{set text(size: 3em); table(
  columns: 3,
  align: (auto,auto,auto,),
  [A], [B], [C],
)}
```
Isn't this a bit cleaner?

gordonwoodhull:
Sure! The rest of the output uses content mode as a reasonable convention, but I'm a big fan of code mode in Typst.

Done.

PR #10572, opened 2025-01-27
DOCX reader: do not issue warning for comments with ext. `+styles`

### LINKED ISSUE #10571: DOCX reader with `docx+styles` generate warning on commends
**Explain the problem.**
When using the DOCX-reader with comments included, the extension `+styles` creates warnings with every comment:

```
~/src/mybook$ pandoc text/custom-reference.docx --from=docx+styles --to=native --track-changes=all --output=test.debug --verbose 
[WARNING] Docx parser warning: Docx comment 9 will not retain formatting
~/src/mybook$
```

Converting without the styles-extension, no warning is given:

```
$ pandoc text/custom-reference.docx --from=docx --to=native --track-changes=all --output=test.debug --verbose 
~/src/mybook$ 
```

**Pandoc version?**
3.6.2

Looking at the source::
https://github.com/jgm/pandoc/blob/220dde406b6ed835b8ac7a267807bbeea53a190d/src/Text/Pandoc/Readers/Docx.hs#L358-L366
only `Para` and `Plain` are considered as *save* block elements.

If using `+styles`, all block elements are pretended with a `Div` element, even comment texts.

Therefore the fix should be very straightforward:
```haskell
  blocksToInlinesWarn cmtId blks = do
  let paraOrPlain :: Block -> Bool
      paraOrPlain (Para _)  = True
      paraOrPlain (Plain _) = True
      paraOrPlain (Div _ _) = True
      paraOrPlain _         = False
  unless (all paraOrPlain blks) $
    lift $ P.report $ DocxParserWarning $
      "Docx comment " <> cmtId <> " will not retain formatting"
  return $ blocksToInlines' (toList blks)
```

**Example document**
[custom-reference.docx](https://github.com/user-attachments/files/18562995/custom-reference.docx)

steven-r:
This fixes #10571 by also allowing `Div` as a block element

### DIFF
--- a/src/Text/Pandoc/Readers/Docx.hs
+++ b/src/Text/Pandoc/Readers/Docx.hs
@@ -359,7 +359,7 @@ blocksToInlinesWarn cmtId blks = do
   let paraOrPlain :: Block -> Bool
       paraOrPlain (Para _)       = True
       paraOrPlain (Plain _)      = True
-      paraOrPlain (Div _ nested) = (all paraOrPlain nested)
+      paraOrPlain (Div _ nested) = all paraOrPlain nested
       paraOrPlain _              = False
   unless (all paraOrPlain blks) $
     lift $ P.report $ DocxParserWarning $
--- a/src/Text/Pandoc/Readers/Docx.hs
+++ b/src/Text/Pandoc/Readers/Docx.hs
@@ -58,7 +58,6 @@ implemented, [-] means partially implemented):
 module Text.Pandoc.Readers.Docx
        ( readDocx
        ) where
-
 import Codec.Archive.Zip
 import Control.Monad ( liftM, unless )
 import Control.Monad.Reader
@@ -359,6 +358,7 @@ blocksToInlinesWarn cmtId blks = do
   let paraOrPlain :: Block -> Bool
       paraOrPlain (Para _)  = True
       paraOrPlain (Plain _) = True
+      paraOrPlain (Div _ _) = True
       paraOrPlain _         = False
   unless (all paraOrPlain blks) $
     lift $ P.report $ DocxParserWarning $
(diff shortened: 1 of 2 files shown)


tarleb:
A different approach could be to "flatten" the style divs before checking that the content is safe. I'd assume that there are still cases where we'd want to show this warning even for Div elements.

steven-r:
> A different approach could be to "flatten" the style divs before checking that the content is safe. I'd assume that there are still cases where we'd want to show this warning even for Div elements.

Like this?
```haskell
  let paraOrPlain :: Block -> Bool
      paraOrPlain (Para _)       = True
      paraOrPlain (Plain _)      = True
      paraOrPlain (Div _ nested) = (all paraOrPlain nested)
      paraOrPlain _              = False
```

src/Text/Pandoc/Readers/Docx.hs
```diff
@@ -356,10 +357,10 @@ extentToAttr _ = nullAttr
 blocksToInlinesWarn :: PandocMonad m => T.Text -> Blocks -> DocxContext m Inlines
 blocksToInlinesWarn cmtId blks = do
   let paraOrPlain :: Block -> Bool
-      paraOrPlain (Para _)  = True
-      paraOrPlain (Plain _) = True
-      paraOrPlain (Div _ _) = True
-      paraOrPlain _         = False
+      paraOrPlain (Para _)       = True
+      paraOrPlain (Plain _)      = True
+      paraOrPlain (Div _ nested) = (all paraOrPlain nested)
```
jgm, line 362: This is the right approach.  But you don't need the `()` around `all paraOrPlain nested`.

steven-r: Adjusted the style accordingly

PR #10573, opened 2025-01-28
Test \{,re}newcommand arguments

### LINKED ISSUE #4470: \newcommand-like definitions with 'bare' command name
When defining a LaTeX macro, subsequent occurrences of `\renewcommand` or `\providecommand` with 'bare' command name (i.e. not enclosed in `{ }`) fails because the (bare) command name gets wrongly expanded.

Input:
```latex
\newcommand{\foo}{123}\renewcommand\foo{456}\foo
```

Expected output:
```
456
```

Actual output (with 2.1.3 from git master):
```sh
$ echo '\newcommand{\foo}{123}\renewcommand\foo{456}\foo' | stack exec pandoc -- -f latex -t native

Error at "source" (line 1, column 36):
unexpected 123
\newcommand{\foo}{123}\renewcommand\foo{456}\foo
                                   ^
```

silby:
Closes #4470

(Or you could just close that and this if you think LaTeX reader macro expansion is adequately tested elsewhere, I didn't scrutinize the other existing tests that thoroughly.)

### DIFF
--- a/test/command/4470.md
+++ b/test/command/4470.md
@@ -1,0 +1,27 @@
+```
+% pandoc -r latex -w plain
+\newcommand{\foo}{123}\renewcommand\foo{456}\foo
+^D
+456
+```
+
+```
+% pandoc -r latex -w plain
+\newcommand\foo{123}\renewcommand\foo{456}\foo
+^D
+456
+```
+
+```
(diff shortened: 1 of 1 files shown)


jgm:
thanks!

PR #10574, opened 2025-01-28
Handle <abbr> as a span-like inline

### LINKED ISSUE #5793: Better support for <abbr> in HTML
Currently we drop the attribute:
```
% pandoc -f html -t markdown
You can use <abbr title="Cascading Style Sheets">CSS</abbr>
^D
You can use CSS
```
It would be better, perhaps, to emit
```
You can use [CSS]{title="Cascading Style Sheets"}
```
though I'm not completely sure.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr

silby:
Closes #5793


jgm:
Great.

Issue #3044, opened 2016-07-28
Mediawiki reader: Unclosed formatting markup should automatically get closed at the end of line


Bobo1239:
Input:
```
aa'''aa
aa
```
Wikimedia output (e.g. [test here](https://en.wikipedia.org/w/index.php?title=Wikipedia:Sandbox&action=submit)):
```
<p>aa<b>aa</b> aa</p>
```
[Pandoc output](http://pandoc.org/try/?text=aa%27%27%27aa%0Aaa%0A&from=mediawiki&to=html):
```
<p>aa'''aa aa</p>
```

PR #10599, opened 2025-02-06
Track wikilinks with a class instead of a title

silby:
Once upon a time the only metadata element for links in Pandoc's AST was a title, and it was hijacked to track certain links as having originated in the wikilink syntax. Now we have Attrs and we can use a class to handle wikilinks instead.

Requires coordinated changes to commonmark-hs: https://github.com/jgm/commonmark-hs/pull/160

### DIFF
--- a/src/Text/Pandoc/Readers/MediaWiki.hs
+++ b/src/Text/Pandoc/Readers/MediaWiki.hs
@@ -667,7 +667,7 @@ internalLink = try $ do
   sym "]]"
   -- see #8525:
   linktrail <- B.text <$> manyChar (satisfy (\c -> isLetter c && not (isCJK c)))
-  let link = B.link (addUnderscores pagename) "wikilink" (label <> linktrail)
+  let link = B.linkWith (mempty, ["wikilink"], mempty) (addUnderscores pagename) (stringify label) (label <> linktrail)
   if "Category:" `T.isPrefixOf` pagename
      then do
        updateState $ \st -> st{ mwCategoryLinks = link : mwCategoryLinks st }
--- a/src/Text/Pandoc/Readers/Vimwiki.hs
+++ b/src/Text/Pandoc/Readers/Vimwiki.hs
@@ -56,15 +56,7 @@ import Data.Text (Text)
 import qualified Data.Text as T
 import Safe (lastMay)
 import Text.Pandoc.Builder (Blocks, Inlines, fromList, toList, trimInlines)
-import qualified Text.Pandoc.Builder as B (blockQuote, bulletList, code,
-                                           codeBlockWith, definitionList,
-                                           displayMath, divWith, emph,
-                                           headerWith, horizontalRule, image,
-                                           imageWith, link, math, orderedList,
-                                           para, plain, setMeta, simpleTable,
-                                           softbreak, space, spanWith, str,
-                                           strikeout, strong, subscript,
-                                           superscript)
+import qualified Text.Pandoc.Builder as B
 import Text.Pandoc.Class.PandocMonad (PandocMonad (..))
 import Text.Pandoc.Definition (Attr, Block (BulletList, OrderedList),
                                Inline (Space), ListNumberDelim (..),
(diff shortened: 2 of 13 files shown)


src/Text/Pandoc/Readers/MediaWiki.hs
```diff
@@ -667,7 +667,7 @@ internalLink = try $ do
   sym "]]"
   -- see #8525:
   linktrail <- B.text <$> manyChar (satisfy (\c -> isLetter c && not (isCJK c)))
-  let link = B.link (addUnderscores pagename) "wikilink" (label <> linktrail)
+  let link = B.linkWith (mempty, ["wikilink"], mempty) (addUnderscores pagename) (stringify label) (label <> linktrail)
```
jgm, line 670: Can you comment on the reason for putting `stringify label` in the title?

src/Text/Pandoc/Readers/MediaWiki.hs
```diff
@@ -699,7 +699,7 @@ inlinesBetween start end =
 emph :: PandocMonad m => MWParser m Inlines
 emph = B.emph <$> inlinesBetween start end
     where start = sym "''"
-          end   = try $ notFollowedBy' (() <$ strong) >> sym "''"
+          end   = try $ notFollowedBy' (() <$ strong) >> ((sym "''") <|> void newline)
```
jgm, line 702: Can you explain why you added the `<|> void newline` ?

src/Text/Pandoc/Readers/Vimwiki.hs
```diff
@@ -555,17 +547,14 @@ link = try $ do
...
-                    return $ B.link (procLink url) tit lab
+                    return $ B.linkWith (attr url) (procLink url) "" lab
                   else do
                     manyTill anyChar (string "]]")
 -- not using try here because [[hell]o]] is not rendered as a link in vimwiki
-                    let tit = if isURI contents
-                                 then ""
-                                 else "wikilink"
-                    return $ B.link (procLink contents) tit (B.str contents)
+                    return $ B.linkWith (attr contents) (procLink contents) "" (B.str contents)
```
jgm, line 554: Can you comment on the decision to add a substantive title here?

src/Text/Pandoc/Readers/MediaWiki.hs
silby: This (approximately) aligns with what mediawiki does, and I tossed in this change since I freed up the `title` on the mediawiki reader. I don't mind leaving it out of this change as orthogonal.

src/Text/Pandoc/Readers/MediaWiki.hs
silby: hah, this crept in because what I was _originally_ trying to do was address #3044 but then I noticed the `"wikilinks"` thing nearby and was distracted.

src/Text/Pandoc/Readers/Vimwiki.hs
silby: I'm afraid you've misread, title is still `""` here.

src/Text/Pandoc/Readers/MediaWiki.hs
jgm: OK, let's leave this out of this PR and address it separately.

src/Text/Pandoc/Readers/MediaWiki.hs
jgm: I'm okay leaving it in if it's what mediawiki does.

src/Text/Pandoc/Readers/Vimwiki.hs
jgm: oh, right.

src/Text/Pandoc/Readers/MediaWiki.hs
silby: OK, this is ready to go in now then!

PR #10607, opened 2025-02-09
Adding support for sidebars to Asciidoc writer

GHyman83:
This PR adds support for [sidebars](https://docs.asciidoctor.org/asciidoc/latest/blocks/sidebars/) to the Asciidoc writer. The proposed changes are as follows:

1. Sidebars are handled alongside admonitions in the Asciidoc writer, since the patterns they follow are very similar.
2. A simple test for sidebar block writing has been added to the Asciidoc writer tests.

Any feedback is appreciated, and many thanks for considering this request!


jgm:
looks good to me. Thank you!

PR #10628, opened 2025-02-19
doc/lua-filters.md: Add missing html_math_method 'katex'

rnwst:


jgm:
Looks good to be (unless this is generated from somewhere else? @tarleb ?)

PR #10654, opened 2025-03-02
MANUAL: correct missing 'date' as possible block entry in metadata

blankoworld:
As metadata suggests 3 possible values I correct a 'doublon': "just a title, a title and an author" to take 'a date' in account.

This avoids repetition.


jgm:
thanks!

Issue #9089, opened 2023-09-20
Parse DocBook "role" attributes as "class" whenever possible, including block elements


de103252:
Docbook has a "role" attribute that is common to all DocBook elements and is  intended to "provides additional, user-specified classification for an element".
For a few selected elements (e.g., phrase, indexterm, emphasis – all of them inline if I'm not mistaken), the Pandoc DocBook reader does examine the "role" attribute and handles it in a special way.
Would it be possible to handle "role" in a generic way for other DocBook elements as well so that a custom filter could examine it? (I hope I am phrasing this at least semi-correctly.)

What I want to achieve:

My toolchain converts from AsciiDoc to DocBook and from DocBook via Pandoc to docx. In AsciiDoc, I have block level elements that I give a custom "Finding" role [(Custom role)](https://docs.asciidoctor.org/pdf-converter/latest/theme/custom-role/):
```
[.Finding]
Lorem ipsum dolor sit amet
```
which results in a `<simpara role="Finding">Lorem ipsum dolor sit amet</simpara>`. In the final Word output, I would then like to have a paragraph with the docx custom paragraph style "finding" (that I would define in the reference docx) applied.

As suggested in https://github.com/jgm/pandoc/issues/7195, this could be done with a filter that acts on the respective elements, if only the information was carried over into the filter input. This would also address https://github.com/jgm/pandoc/issues/3657.

jgm:
Pandoc Para elements don't take attributes.  The only way to handle this case would be to generate a Div with the attributes, wrapping the Para.

de103252:
Okay, thanks—seems I'm at a dead end then, if I understand you correctly. To summarize, my goal is to carry an AsciiDoc custom role label over to a custom paragraph style name in Word, and that seems not to be possible in an elegant way with the current toolchain. What I could and will try to do is create a custom filter that recognizes text fragments (such as "Finding" and "Recommendation"). Or swallow the pill and change the Word output manually.

jgm:
My comment wasn't meant to close this discussion.  It would be possible to add a wrapper Div with the class.  That would be enough to allow you to do what you need in filters, no?

de103252:
[...]

PR #10183, opened 2024-09-13
Parse id, class, and tabstyle on tables in DocBook Reader

lifeunleaded:
Add parsing of id (xml:id), class, and tabstyle XML attributes for the table and informaltable elements in the DocBook reader. The tabstyle value is put in the 'custom-style' attribute.

fixes #10181

PR #10665, opened 2025-03-04
Adding support for "role" attributes for the DocBook reader

yanntrividic:
As the title says, this PR adds support for the `role` attributes to be parsed by the DocBook reader. This issue has been outlined in https://github.com/jgm/pandoc/issues/9089. This PR is implementing the solution that was discussed in the issue by @jgm and @lifeunleaded in particular.

I haven't written any unit test for this, as I couldn't find a unit test file for the DocBook reader, and as I wasn't ready to start one... I hope it won't be blocking for this PR.

This change didn't feel like it required a modification to the `MANUAL.txt` file.


jgm:
Thanks for the PR!  I wonder if we couldn't simplify things considerably by using the function
https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/Shared.hs#L293-L302
Using this, you could just `addAttributes [("role", therole)]` on *any* pandoc Inlines or Blocks.
This might allow you to just have one place where the role is checked, instead of separate places for each type of element.

lifeunleaded:
Excellent addition, thank you!  https://github.com/jgm/pandoc/pull/10183/commits/61ff730aad6d4d24da2e69ba32e7eae105680315 Should point you to DocBook reader tests, but I could take a stab at adding some if you don’t feel comfortable doing it.

yanntrividic:
Hello, thank you both for your help!

@jgm, thank you for pointing me in this direction. Indeed it seems to be exactly what we need. I updated the code to work in this direction. Is it what you expected?

@lifeunleaded,  I don't know how I missed those files... I'm not sure how we should tackle this though. Should we add in each "zone" of the code a test with a new node that has a `role` attribute? Tbh, if you feel like you are fine taking care of it, I wouldn't mind _at all_.

lifeunleaded:
@yanntrividic I can take a look within a few days, as soon as I know what the changes do. I'll let @jgm determine whether that should block the PR. Otherwise I can submit a separate PR for tests.

lifeunleaded:
@yanntrividic I think it would be a good idea to just add a role attribute to some intended supported elements in `test/docbook-reader.docbook` and then run e.g.  `stack test --test-arguments='-p docbook -j4 --hide-successes'` to get a view of what it expects. I just added a quick test role to the `sect1` and got unexpected results. Not sure if my setup has issues or if the code needs adjusting. I'll try to look deeper soon.

lifeunleaded:
Building the branch and running `~/.local/bin/pandoc -f docbook -t native ~/sandbox/test.xml` with this input:
```
<?xml version="1.0" encoding="utf-8" ?>                                                                                                                                                 
<article xmlns="http://docbook.org/ns/docbook"
         xmlns:xlink="http://www.w3.org/1999/xlink"
         xmlns:mml="http://www.w3.org/1998/Math/MathML" version="5.0">
    <title>Pandoc Test Suite</title>                                                                                                                                                    

    <section id="headers" role="sect1role">
      <title>title1</title>
      <para>Test</para>
      <section><title>title2</title>                                                                                                                                                    
      <para>Test2</para>                                                                                                                                                                
      </section>
    </section>
</article>
``` 

Gives me this:
```
[ Header
    1
    ( "headers"
    , []
    , [ ( "role" , "sect1role" ) , ( "role" , "sect1role" ) ]
    )
    [ Str "title1" ]
, Div
    ( ""
    , []
    , [ ( "wrapper" , "1" ) , ( "role" , "sect1role" ) ]
    )
    [ Para [ Str "Test" ] ]
, Header
    2
    ( "" , [] , [ ( "role" , "sect1role" ) ] )
    [ Str "title2" ]
, Div
    ( ""
    , []
    , [ ( "wrapper" , "1" ) , ( "role" , "sect1role" ) ]
    )
    [ Para [ Str "Test2" ] ]
]
```

It seems to put role twice in the actual element, and then on all the child elements. I assume that's not intended?

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1025,8 +1030,8 @@ parseBlock (Elem e) =
          parseTable = do
                       let elId = attrValue "id" e
                       let attrs = case attrValue "tabstyle" e of
-                                    "" -> []
-                                    x  -> [("custom-style", x)]
+                                    "" -> getRoleAttr e
+                                    x  -> ("custom-style", x) : getRoleAttr e
```
jgm, line 1034: Do you need this given that we're adding the attribute in l. 980?

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1099,7 +1104,8 @@ parseBlock (Elem e) =
            modify $ \st -> st{ dbSectionLevel = n }
            b <- getBlocks e
            modify $ \st -> st{ dbSectionLevel = n - 1 }
-           return $ headerWith (elId, classes, maybeToList titleabbrevElAsAttr++attrs) n' headerText <> b
+           return $ headerWith (elId, classes, maybeToList titleabbrevElAsAttr++attrs++getRoleAttr e) n'
```
jgm, line 1107: Isn't it unnecessary to add the role here if we're adding it in l. 980?

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1122,9 +1128,8 @@ parseBlock (Elem e) =
            b <- p
            case mbt of
              Nothing -> return b
-             Just t -> return $ divWith (attrValue "id" e,[],[])
-                         (divWith ("", ["title"], []) (plain t) <> b)
-
+             Just t  -> return $ divWith (attrValue "id" e, [], getRoleAttr e)  -- Updated!
+                         (divWith ("", ["title"], getRoleAttr e) (plain t) <> b)
```
jgm, line 1132: ditto

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1134,7 +1139,7 @@ parseBlock (Elem e) =
            b <- getBlocks e
            let t = divWith ("", ["title"], []) (plain $ fromMaybe mempty mbt)
            -- we also attach the label as a class, so it can be styled properly
-           return $ divWith (attrValue "id" e,[label],[]) (t <> b)
+           return $ divWith (attrValue "id" e, [label], getRoleAttr e) (t <> b)
```
jgm, line 1142: ditto

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1329,6 +1334,9 @@ parseInline (Elem e) =
         -- <?asciidor-br?> to in handleInstructions, above.
         "pi-asciidoc-br" -> return linebreak
         _          -> skip >> innerInlines id
+  return $ case qName (elName e) of
+    "emphasis" -> parsedInline
```
jgm, line 1338: why this special case for "emphasis"?

src/Text/Pandoc/Readers/DocBook.hs
yanntrividic: You're right! Is is not needed, I thought it was since [we have a return statement before](https://github.com/jgm/pandoc/blob/57d61da25a519a5e209b1f3d10bae904be003560/src/Text/Pandoc/Readers/DocBook.hs#L980), but that is actually not related.

src/Text/Pandoc/Readers/DocBook.hs
yanntrividic: See https://github.com/jgm/pandoc/pull/10665#discussion_r1986351032.

src/Text/Pandoc/Readers/DocBook.hs
yanntrividic: See https://github.com/jgm/pandoc/pull/10665#discussion_r1986351032.

src/Text/Pandoc/Readers/DocBook.hs
yanntrividic: I tried something, which failed, now I need to figure out a way to handle this.

Currently, Pandoc supports `role` attributes for some `Inlines`. It helps specifying `Strong`, `Strikeout`, `Underline` and `Emph` elements [there](https://github.com/jgm/pandoc/blob/d73475cec8c7363055c482f5107a2759e1f6a277/src/Text/Pandoc/Readers/DocBook.hs#L1317-L323]. All those elements need a wrapper to add attributes around them. But if we try to apply this updated code to:

```xml
<emphasis role="strong">word</emphasis>
```

We get:

```
Span
            ( ""
            , []
            , [ ( "wrapper" , "1" ) , ( "role" , "strong" ) ]
            )
            [ Strong [ Str "word" ] ]
```

Which is not what we would want... So my previous attemps in f0827eef027f732c36091f2723fbaeb4dc2e9e44 was to circumvent this issue, but then no `emphasis` DocBook element would have attributes, which is not what we want either.

yanntrividic:
Hey @lifeunleaded, thanks for taking the time to dig into some tests. You're right with what you mention in https://github.com/jgm/pandoc/pull/10665#issuecomment-2704776158, it is not intended... And I'm not sure how this recursion happens.

With the updated code, here is what we get:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<article xmlns="http://docbook.org/ns/docbook"
         xmlns:xlink="http://www.w3.org/1999/xlink"
         xmlns:mml="http://www.w3.org/1998/Math/MathML" version="5.0">
    <title>Pandoc Test Suite</title>
    <section id="headers" role="sect1role">
      <title>title1</title>
      <para role="para1role">Test</para>
      <section role="sect2role"><title>title2</title>
      <para role="para2role">Test2</para>
      </section>
    </section>
</article>
```

```
[ Header
    1
    ( "headers" , [] , [ ( "role" , "sect1role" ) ] )
    [ Str "title1" ]
, Div
    ( ""
    , []
    , [ ( "role" , "sect1role" )
      , ( "wrapper" , "1" )
      , ( "role" , "para1role" )
      ]
    )
    [ Para [ Str "Test" ] ]
, Header
    2
    ( ""
    , []
    , [ ( "role" , "sect1role" ) , ( "role" , "sect2role" ) ]
    )
    [ Str "title2" ]
, Div
    ( ""
    , []
    , [ ( "role" , "sect1role" )
      , ( "role" , "sect2role" )
      , ( "wrapper" , "1" )
      , ( "role" , "para2role" )
      ]
    )
    [ Para [ Str "Test2" ] ]
]
```

I really don't get how the attribute propagates in the child elements, it just doesn't make sense to me. Any hints on your side? Maybe this is something obvious, and if so, I am sorry. I gotta tell you that I'm quite new to this, as those are my very first lines of Haskell 😅

src/Text/Pandoc/Readers/DocBook.hs
jgm: Oh, right.  Well, you're just preventing a role attribute from going on something parsed from an `emphasis` tag, and that's fine given that we already handle the roles in another way.

jgm:
Take a look at jgm/commonmark-hs , in particular commonmark-pandoc, which is where the `addAttributes` used in `addPandocAttributes` is defined.
You'll see
```hs
instance HasAttributes (Cm a B.Blocks) where
  addAttributes attrs b = fmap (addBlockAttrs attrs) <$> b

instance HasAttributes (Cm a B.Inlines) where
  addAttributes attrs il = fmap (addInlineAttrs attrs) <$> il
```

So `addAttributes`, applied to a  Blocks (which can be a sequence of Block elements), will set the attribute in every Block in the sequence. I think that's what is going on here, does it help?

src/Text/Pandoc/Readers/DocBook.hs
yanntrividic: But then, that would mean that we can't have any other `role` value than `bf`, `bold`, `strong`, `strikethrough`, or `underline`, on `emphasis` elements ? Maybe in that case the role attributes should be assigned to `phrase` elements... That's a compromise I'm willing to make yes, if you feel it makes sense.

yanntrividic:
So, regarding this propagation of `role` attributes to the c, you think it is fine like this? Wouldn't we want to have only the `role` attribute on the element that holds it?

jgm:
> So, regarding this propagation of role attributes to the children, you think it is fine like this? Wouldn't we want to have only the role attribute on the element that holds it?

No, it isn't okay.  It's important to figure out why that is happening and avoid it.

lifeunleaded:
Given that `addPandocAttributes` applies to Blocks (list of block), I'm not sure it can be applied so generally here (although I don't fully understand how it gets a Blocks in parseBlock.

If it were me, I would go through the list in `parsedBlock` and try to be a bit more granular. For the ones without attributes, there is `divWith` (like informalequation already uses), and for some there are other places to apply it. E.g. for sections, one could extend `sect n` to be:

```
sect n = sectWith(attrValue "id" e) [] (getRoleAttr e) n
```

This seems to pass the trivial beginning of test coverage I've tried here.

A more elegant solution would perhaps be polymorphic functions for adding the role attribute, defined separately for inlines and blocks with and without attributes, but that's not something I can whip up an example of here, if indeed it is even possible. That function name could then be used throughout.

lifeunleaded:
Worth noting that it looks like `phrase` and `indexterm` in `parseInline` already has handling for `role`. I wonder if that would also see strange effects from applying `addPandocAttributes` to the result.

jgm:
Here's why you're getting the role applied to all the children of section:
https://github.com/jgm/pandoc/blob/17d3ad28fe856ac015366b8a71e1038b49ad35ec/src/Text/Pandoc/Readers/DocBook.hs#L1107

The `sect` function returns a Blocks that is a sequence of Block elements starting with the Header and including all the children.

Perhaps it should put all of this in a Div, either always or at least in the case where there is a role attribute.

Then the role will just be attached to the Div.

Pandoc sometimes uses this structure for sections:

- Div with identifier and classes "section" and "level1" (or "level2" etc.)
  - Header
  - other contents
 
So that would be quite reasonable in this case.

lifeunleaded:
If I may be so bold: I think the DocBook _writer_ is based on the role being in the Header attributes, so the roundtrip becomes a bit awkward if the reader creates an enclosing Div instead. 

I would suggest moving the `getRoleAttr` into `sect` and use `addPandocAttributes` for the cases that do not use sect.

So `sect` becomes `sect n = sectWith(attrValue "id" e) [] (getRoleAttr e) n` and e.g. `para` uses this pattern in `parseBlock`:

```
parseBlock (Elem e) = do
  parsedBlock <- case qName (elName e) of
        "toc"   -> skip -- skip TOC, since in pandoc it's autogenerated                                                                                                                 
        "index" -> skip -- skip index, since page numbers meaningless                                                                                                                   
        "para"  -> addPandocAttributes (getRoleAttr e) <$> parseMixed para (elContent e)
```

This becomes more work, since it needs to be applied for each case in `parseBlock` not using `sect`, but the upside is that the behaviour can be more based on which element it is.

jgm:
The docbook writer can handle this structure just fine:
```
% pandoc -f native -t docbook
[ Div
    ( "" , [ "section" ] , [] )
    [ Header 1 ( "test" , [] , [] ) [ Str "Test" ]
    , Para [ Str "paragraph" ]
    ]
]
<section xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="test">
  <title>Test</title>
  <para>
    paragraph
  </para>
</section>
```

lifeunleaded:
Oh! Excellent, thank you, I wasn't aware and didn't manage to test before your reply. Then I have nothing to add, enclosing sect in Div with the role sounds like the best way forward.

yanntrividic:
> Worth noting that it looks like `phrase` and `indexterm` in `parseInline` already has handling for `role`. I wonder if that would also see strange effects from applying `addPandocAttributes` to the result.

With those elements, the `role` attributes are parsed as `classes`. For consistency, wouldn't it make more sense just to remove this so that every `role` is treated as an attribute ?

src/Text/Pandoc/Readers/DocBook.hs
yanntrividic: From what I understand from [the documentation](https://tdg.docbook.org/tdg/5.2/ref-elements), I think that DocBook documents expect only one `role` attribute per element. To specify more precise roles, it is recommended to parameterize a pattern to do so, but you still only have one `role` attribute.

So, knowing this, I think that it shouldn't be possible to have a `role` on a `Strong` element, because that would imply that there are necessarily two `role` attributes on the DocBook element, which shouldn't be possible.

And in that case, we would only need to change the last line from this part:

```haskell
"emphasis" -> case attrValue "role" e of
                             "bf"            -> innerInlines strong
                             "bold"          -> innerInlines strong
                             "strong"        -> innerInlines strong
                             "strikethrough" -> innerInlines strikeout
                             "underline"     -> innerInlines underline
                             _               -> innerInlines emph
```

Does it make sense like this? What do you think?

lifeunleaded: @yanntrividic Not strictly true, docbook would allow multiple tokens in a role attribute, so e.g. ` <emphasis role="bold special">` is entirely valid. However, I think that is an extreme edge case and I've never seen it used. I think it would be perfectly fine to only process the known role tokens for emphasis as you describe.

lifeunleaded:
> With those elements, the role attributes are parsed as classes. For consistency, wouldn't it make more sense just to remove this so that every role is treated as an attribute ?

Yes, I agree that would be better. Unless we want to be very careful about not breaking existing filters out there that expect it in `classes`. If we do, I suppose there is no great problem with leaving them in classes and still adding them to attributes. @jgm , any opinion?

yanntrividic:
> > With those elements, the role attributes are parsed as classes. For consistency, wouldn't it make more sense just to remove this so that every role is treated as an attribute ?
> 
> Yes, I agree that would be better. Unless we want to be very careful about not breaking existing filters out there that expect it in `classes`. If we do, I suppose there is no great problem with leaving them in classes and still adding them to attributes. @jgm , any opinion?

I have a working solution for this--I think--but I'll wait for the third opinion on the matter before committing anything!

src/Text/Pandoc/Readers/DocBook.hs
yanntrividic: I see, you are right... Well I guess it would not be so much work to support this feature, but in that case we would support only one pattern, that is to say, space-separated values for `role` attributes. I think at this point, I would prefer leaving this for a filter to handle.

yanntrividic: On another note, I've been trying to work on something to discriminate the last case from the others... Not sure how to do it properly though... Here is my attempt: https://github.com/jgm/pandoc/pull/10665/commits/370375728f298a6f00340bb36eb5405214be35bd.

I know there is a parsing error in the code, but as I said, I'm quite new to Haskell, and I am really not sure how I could make this kind of type assertion.

Basically, the thought process here is that if the element is an `emphasis` element, and that has been parsed as an `Emph` element, then we can add a `role` attribute. But I'm really not sure what would be the right way to check the second condition. Any ideas?

jgm:
Probably safest to keep them as classes and also add them as roles.  But I don't know if anyone is actually relying on this behavior. If the duplication is bad, we could mark it as a potentially breaking change in the changelog.

lifeunleaded:
@yanntrividic pulled recent changes to start looking at tests, but it doesn't build right now:
```
pandoc                           > /home/erik/git/personal/pandoc/src/Text/Pandoc/Readers/DocBook.hs:1342:25: error: [GHC-83865]
pandoc                           >     • Couldn't match expected type ‘Inline’ with actual type ‘Element’
pandoc                           >     • In the expression: e :: Inline
pandoc                           >       In an equation for ‘inlineElement’: inlineElement = e :: Inline
pandoc                           >       In the expression:
pandoc                           >         do let inlineElement = ...
pandoc                           >            if not (isEmphElement inlineElement) then
pandoc                           >                addPandocAttributes (getRoleAttr e) parsedInline
pandoc                           >            else
pandoc                           >                parsedInline
pandoc                           >      |
pandoc                           > 1342 |     let inlineElement = e :: Inline
pandoc                           >      |                         ^
```

yanntrividic:
> @yanntrividic pulled recent changes to start looking at tests, but it doesn't build right now:

Yes sorry! That's what I tried to explain here: https://github.com/jgm/pandoc/pull/10665#discussion_r1991569081, I think I'm close to a solution (that's why I committed this and made this comment).

lifeunleaded:
> > @yanntrividic pulled recent changes to start looking at tests, but it doesn't build right now:
> 
> Yes sorry! That's what I tried to explain here: [#10665 (comment)](https://github.com/jgm/pandoc/pull/10665#discussion_r1991569081), I think I'm close to a solution (that's why I committed this and made this comment).

No worries. I see the problem and don't have an immediate solution. If you get stuck, maybe it's easier to apply `addPandocAttributes` (or use `spanWith` directly) in the case matches in `parseInline` and just omit it on `"emphasis"`, if that was the idea. 

Good luck and thank you for the effort.

lifeunleaded:
Also, if you don't yet, I recommend using `stack test` before pushing to identify issues quicker than the pipeline. E.g. `stack test --test-arguments='-p docbook -j4 --hide-successes'` will run fairly quickly (if you use haskell-stack)

jgm:
The problem is with
```hs
    let inlineElement = e :: Inline
```
`e` here does not have type Inline, it has type Element.
Inline is a pandoc AST concept, Element is a type from the xml parser.

yanntrividic:
> No worries. I see the problem and don't have an immediate solution. If you get stuck, maybe it's easier to apply `addPandocAttributes` (or use `spanWith` directly) in the case matches in `parseInline` and just omit it on `"emphasis"`, if that was the idea.

Okay! So I tried just that, and it seems to do what you proposed. I added a class `emphasis` on the `Span` so that this information is not just gone, as this is done with other DocBook elements. What do you think? Should we try to make it into an `Emph` element or is it fine as such?

lifeunleaded:
> > No worries. I see the problem and don't have an immediate solution. If you get stuck, maybe it's easier to apply `addPandocAttributes` (or use `spanWith` directly) in the case matches in `parseInline` and just omit it on `"emphasis"`, if that was the idea.
> 
> Okay! So I tried just that, and it seems to do what you proposed. I added a class `emphasis` on the `Span` so that this information is not just gone, as this is done with other DocBook elements. What do you think? Should we try to make it into an `Emph` element or is it fine as such?

I think it's important that the existing mappings that create Strong, Emph, and so on are left intact, yes.

lifeunleaded:
A quick test and a look at the latest change looks like it works as I would expect, so I think we could start writing tests.

@yanntrividic Would you like me to write up a few and push to your fork, or do you feel comfortable starting them yourself?

yanntrividic:
While trying to write a few tests, I realized that what I did in a2321612dd4a7074b74fc424134bb5b58d24c5eb wasn't really working as expected, and `Headers` were getting the role attribute alongside the `Div` that wrapped them. It is now solved in fa815ce487d31de0faaa8ff978a6e6f81e8b58be.

And now that I am facing those tests for real, I am not sure were to start. Maybe @lifeunleaded you could show me the way and I will continue if needed?

lifeunleaded:
@yanntrividic I started looking at the failing tests, and it looks like `emphasis` without role just becomes the `Span` with `role`, but the inlines are just the strings, not an `Emph`. I would expect that `emphasis` without role becomes `Emph`, but potentially within a `Span` with the `role` attribute. Is this something that could be fixed?

src/Text/Pandoc/Readers/DocBook.hs
```diff
@@ -1329,7 +1325,8 @@ parseInline (Elem e) = do
                              "strong"        -> innerInlines strong
                              "strikethrough" -> innerInlines strikeout
                              "underline"     -> innerInlines underline
-                             _               -> innerInlines emph
+                             _               -> innerInlines $
```
lifeunleaded, line 1332: This seems to create a span with the role attribute but then just the string content. I think it needs to become an Emph to not break existing writers too much. If it's difficult to achieve, I think reverting the _ case to `innerInlines emph` (and not getting the Span for this particular case) is better than not creating an Emph

yanntrividic: Yes, this is what I thought you meant in your first comment. As I said, I am not sure how to achieve what we want there... So if nobody has an idea to pull that without too much effort, I think it is fine to admit that if you want a specific element on an `Emph`, it could be put on a `phrase` element and do the work. What do you think @jgm ?

jgm: I don't understand the context; you'd have to explain the issue to me more fully.

lifeunleaded: @jgm As far as I can discern from testing, `<emphasis>emphasis</emphasis>` (without a `role` attribute or with a role attribute that is not `bf`, `strong`, `bold`, `strikethrough`, or `underline`) becomes `Span ( "" , [ "emphasis" ] , [] ) [ Str "emphasis" ]`, I would prefer `Span ( "" , [ "emphasis" ] , [] ) [ Emph [ Str "emphasis" ] ]` or, failing that, revert back to the current behavior `Emph [ Str "emphasis" ]`. I fear not creating the `Emph` will be breaking for a lot of conversions.

@yanntrividic My apologies if I've made an unclear comment. I may have misunderstood the intentions of a particular change at some point. The overall change is a really useful one and I think we're close to a good implementation.

yanntrividic: No problem! Though yes, I tried again to work around those types for a bit, but I just can't figure out a nice way to do wrap the `Emph` around a `Span`. I'm fine with reverting the changes if nobody has a better proposition :)

jgm: I'm a bit lost here: surely, `<emphasis>emphasis</emphasis>` should be converted as simply `Emph [ Str "emphasis" ]`.  I'm not sure why one would even consider the Span conversion?  I may be missing more of the context, though?

yanntrividic: Ok, so let's start again from the beginning.

By default, `addPandocAttributes` is now applied to all the `Inline` elements to add a `role` attribute, if present. But for a few DocBook elements, and especially here for `emphasis`, the `role` attribute was already taken into account.

For the `emphasis` element, the `role` attribute is to discriminate `Strong`, `Strikeout`, `Underline` and `Emph`. So if we apply the `addPandocAttributes` function after doing that, we can get outputs such as:

```
Span
    ( ""
        , []
        , [ ( "wrapper" , "1" ) , ( "role" , "strong" ) ]
    )
    [ Strong [ Str "word" ] ]
```

Which is unnecessary.

On the other end, we would like to be able to get outputs such as:

```
Span
    ( ""
        , []
        , [ ( "wrapper" , "1" ) , ( "role" , "special" ) ]
    )
    [ Emph [ Str "word" ] ]
```

Or even:


```
Span
    ( ""
        , []
        , [ ( "wrapper" , "1" ) , ( "role" , "special" ) ]
    )
    [ Strong [ Str "word" ] ]
```

But not (I think?):

```
Span
    ( ""
        , []
        , [ ( "wrapper" , "1" ) , ( "role" , "special" ), ( "role" , "strong" ) ]
    )
    [ Strong [ Str "word" ] ]
```

But if we want to get this output, we have to modify this [bit of code](https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/Readers/DocBook.hs#L1317-L1323).

But I don't know how to modify those lines (or modify others?) in a good way to achieve this, because the `emph` function's types work well with the `innerInlines` function, but not with the `addPandocAttributes` function.

Is it a bit clearer now @jgm?

yanntrividic:
Hello! Some time has passed, but I am back on this PR.

For reminders, this contribution adds support for the `role` attribute to the Docbook reader, though it is not as straightforward as one may think, as attributes can not be handled the same way depending on the element we are dealing with.

Here is where we are now:

* `Header` elements, in Docbook, are actually sections. So, according to this [comment](https://github.com/jgm/pandoc/pull/10665#issuecomment-2711877186), we enclose the `Header` and the content of a section with a `role` in a `Div` with the `role` attribute.
* Block elements such as `para`, `simpara` (that use the `parseMixed` function), when with a `role` attribute, are enclosed in a `Div` with the `role` attribute.
* Block elements such as `blockquote`, `orderedlist`, `itemizedlist` (that use the `withOptionalTitle` function) with a `role` attribute, are enclosed in a `Div` with the `role` attribute.
* Inlines such as `phrase` elements (except for `emphasis` elements) with a `role` attribute, keep their `role` attribute attached (this behavior was already present).
* `emphasis` elements, when with a `role` attribute, can become either `bf`, `strong`, `bold`, `strikethrough`, or `underline`. When using an arbitrary value for the `role` attribute, the attribute is discarded (this behavior was already present). I tried extending the support here, but [I did not succeed](https://github.com/jgm/pandoc/pull/10665#discussion_r1998371948).

This contribution does not covers all the types of elements that are currently implemented. For example, `listitem` does not keep the `role` attribute, and I don't think that any of the element that go through the `filterChildren` function keeps the `role` attribute after being read. (If you have a proposition for this, you are welcome to share it!)

I have written tests for this contribution, and it can now pass the CI. The only breaking change to the previous tests are the ones regarding the first point, about `Header` elements. Other tests were written to verify the other statements are true.

Finally, I think that this PR is now ready to be merge! Or at least we are getting really close!

yanntrividic:
Hello again!

It's me again. Not sure what is blocking with the PR but I would love to help if I can.

Also, just a bit more context on why I needed this to work. It was in the context of working on an IDML Pandoc reader... For which I just finished working on the english docs. Here it is: https://outdesign.deborderbollore.fr/en.

I thought it might be worth mentioning here :)

Have a good day.

jgm:
Thanks for the reminder, It's rather complex, and I need to have a look at it.  It would help if you could rebase the commits into one commit, or a sequence of commits that make logically unrelated changes (not, as currently, reflecting the history of development).  The commit message should clearly describe the changes that are being made and their motivation.

yanntrividic:
At this point, I feel like it would be easier to open up a new pull request with only the right commits in the right order. I am not comfortable in rebasing so many commits, with the various fork syncing in the middle... It seems like I'll be tearing my hair out over this!

PR #10687, opened 2025-03-13
data/templates/default.beamer: only emit \date if set

josch:
This allows one, for example, to set a custom \date in the header-includes of the rmarkdown yaml frontmatter. Without this conditional, the custom \date in the frontmatter would be overridden by a (potentially empty) date from this template later.


jgm:
thanks!

jgm:
This is being reverted.  The commit reverting it will include an explanation.

jgm:
No worries, thanks anyway for the contribution!

PR #10688, opened 2025-03-14
Latex: allow phantomsection in captions

etclub:
https://tex.stackexchange.com/questions/228973/argument-of-captionydblarg-has-an-extra

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX/Util.hs
+++ b/src/Text/Pandoc/Writers/LaTeX/Util.hs
@@ -267,7 +267,7 @@ hypertarget ident = do
       return $ text "\\protect\\hypertarget" <> braces ref <> "{}"
      else do
       label <- labelFor ident
-      return $ text "\\phantomsection" <> label
+      return $ text "\\protect\\phantomsection" <> label
 
 labelFor :: PandocMonad m => Text -> LW m (Doc Text)
 labelFor ""    = return empty


jgm:
If you want this PR to be considered, please explain in words what it is doing and why.
A link to a huge discussion isn't enough.  The explanation should also go in the commit message.

etclub:
> If you want this PR to be considered, please explain in words what it is doing and why. A link to a huge discussion isn't enough. The explanation should also go in the commit message.

I find the following html
```
<table>
<caption>...<span id="test">test</span>...</caption>
...
```
would result in latex a `\phantomsection` in caption, which causes an error.
And the link above gives a solution.

jgm:
Complete example:
```html
<table>
<caption>...<span id="test">test</span>...</caption>
<tr><td>A</td><td>b</td></tr>
</table>
```
gets converted to
```tex
\begin{longtable}[]{@{}ll@{}}
\caption{...\phantomsection\label{test}{test}...}\tabularnewline
\toprule\noalign{}
\endfirsthead
\endhead
\bottomrule\noalign{}
\endlastfoot
A & b \\
\end{longtable}
```
and compiling this produces the LaTeX error
```
! Undefined control sequence.
\Hy@MakeCurrentHref ...def \HyperLocalCurrentHref 
                                                  {#1}\@onelevel@sanitize \H...
l.65 ...n{...\phantomsection\label{test}{test}...}
```
Adding `\protect` before `\phantomsection` makes this error go away.

What isn't clear to me is whether there would be any drawbacks if we add `\protect` before `\phantomsection` in every context (not just inside a caption).

jgm:
It seems to work in my tests with `\protect` in every context, so I think we can do this.

PR #10707, opened 2025-03-19
Skip at most one argument to LaTeX tabular newline

### LINKED ISSUE #7512: LaTeX reader sometimes omits expressions containing square brackets [
In LaTeX, I have an expression `[d,\delta]=0` that I've placed in an `equation*` environment. Upon converting to HTML, all I see is "=0" (see screenshot below), even if we were to get rid of MathJax. When placed in the `\[ \]` environment, however, the expression converts correctly.

The following is a working example:
```
\documentclass{article}
\begin{document}
\noindent With the arguments
\begin{verbatim}
pandoc test.tex -f latex -t html --mathjax -s -o out.html
\end{verbatim}
the code
\begin{verbatim}
\begin{equation*}
  [d,\delta]=0.
\end{equation*}
\end{verbatim}
results in
\begin{equation*}
  [d,\delta]=0.
\end{equation*}
but the code
\begin{verbatim}
\[ [d,\delta] = 0\]
\end{verbatim}
results in
\[ [d,\delta]=0.\]
\end{document}
```
resulting in
Screen Shot 2021-08-19 at 8 35 57 PM

Interestingly, throwing a character `a[d,\delta]=0` in front of the opening square bracket in the `equation*` environment solves the problem. Is this a bug or a feature that I should work around? Thanks!

**Version:** I'm running pandoc version 2.14.1 on macOS Big Sur 11.4 on a Macbook Air (M1 2020).

silby:
In LaTeX's tabular environment, the tabular newline takes an optional argument that we skip. But it only takes a single optional argument, and any further square-bracketed text that follows shouldn't be skipped.

Fixes #7512, specifically [the followup issue](https://github.com/jgm/pandoc/issues/7512#issuecomment-905245320) identified by @fedeinthemix, and also adds a test for the original problem raised in that issue which was already fixed at some point.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX/Table.hs
+++ b/src/Text/Pandoc/Readers/LaTeX/Table.hs
@@ -57,7 +57,7 @@ hline = try $ do
 
 lbreak :: PandocMonad m => LP m Tok
 lbreak = (controlSeq "\\" <|> controlSeq "tabularnewline")
-         <* skipopts <* spaces
+         <* optional (void rawopt) <* spaces
 
 amp :: PandocMonad m => LP m Tok
 amp = symbol '&'
--- a/test/command/7512.md
+++ b/test/command/7512.md
@@ -1,0 +1,65 @@
+```
+% pandoc -t native -f latex
+\begin{equation*}
+[d,\delta]=0.
+\end{equation*}
+^D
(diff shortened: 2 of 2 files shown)


jgm:
excellent, thanks

PR #10735, opened 2025-03-28
Typo: impage should be image

cchristiansen:
AVIF is an image file extension. Image was previously written incorrectly as impage. This commit corrects that error.


jgm:
Thanks!

PR #10736, opened 2025-03-28
Add AVIF to Org Reader image extensions

cchristiansen:
AVIF is an image extension. This commit enables the Org reader to recognise AVIF files as images to be displayed, and not simply as links.

I'm unsure whether this pull request should be merged though, as currently the default Emacs (and hence org-mode) image file name extensions are `("png" "jpeg" "jpg" "gif" "tiff" "tif" "xbm" "xpm" "pbm" "pgm" "ppm" "pnm" "svg" "webp")` and do not include `avif`. However, AVIF is already recognised by pandoc in `MIME.hs` so it seems remiss to not let the Org reader recognise AVIF as an image.


jgm:
@tarleb what do you think?
Generally we try to match Emacs behavior. On the other hand, users can modify the Emacs defaults to allow `avif`.

tarleb (approved):
LGTM. I find it hard to imagine a case where this would lead to unexpected behavior.

jgm:
thanks!

Issue #10761, opened 2025-04-06
MediaWiki reader disagrees with MediaWiki proper on emphasis of possessive trailing quotes


silby:
In the simplest case:

```
pandoc -t native -f mediawiki
''foo'''
^D
[ Para [ Emph [ Str "foo" ] , Str "'" ] ]
```

but MediaWiki itself would want `Emph [Str "foo'"]`, as can be seen in [their regression test](https://raw.githubusercontent.com/wikimedia/mediawiki-services-parsoid/refs/heads/master/tests/parser/quotes.txt).

I came across this while trying to port some of that test file to Pandoc tests, in service of fixing #3044.

Should this be fixed (in principle) in the MediaWiki reader or is this an accepted difference between Pandoc and MediaWiki?

jgm:
In principle it should be fixed.

PR #10766, opened 2025-04-07
mediawiki reader: improve strong/emph conformance

silby:
cf. #10761 and #3044.

I made some progress with this today without completely blowing up the existing strong and emph parsers but weird edge cases remain. E.g. consider `''foo''''bar''`. Pandoc today will give you `Emph [ Str "foo" , Str "bar" ]`, which has an obvious appeal. My work in progress gives `Emph [ Str "foo''" ] , Str "bar''"`, which is odder but defensible given other requirements for emphasized quote marks. The actual _correct_ answer, according to MediaWiki, is `Emph [ Str "foo'" , Strong [ Str "bar" ] ]`, i.e. _foo'**bar**_, which is basically a koan.

Parsoid has a lot of code just for [processing quotes](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/services/parsoid/+/refs/heads/master/src/Wt2Html/TT/QuoteTransformer.php), presumably aiming to maintain bug-for-bug compatibility with whatever MediaWiki's first parser did. So what a string of single-quotes means varies depending on what comes after it in the line, in a more context-sensitive way than I expected.

Would it be better to merge code that makes us more conformant with MediaWiki for some cases and "wrong in a different way" for others, or to try to reach perfection?

### DIFF
--- a/pandoc.cabal
+++ b/pandoc.cabal
@@ -845,6 +845,7 @@ test-suite test-pandoc
                   Tests.Readers.FB2
                   Tests.Readers.Pod
                   Tests.Readers.DokuWiki
+                  Tests.Readers.MediaWiki
                   Tests.Writers.Native
                   Tests.Writers.ConTeXt
                   Tests.Writers.DocBook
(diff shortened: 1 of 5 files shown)


jgm:
In general I'm not too concerned with divergences in edge cases.  Nobody is ever going to write `''foo''''bar''` and intend to get emph "foo'" + strong "bar".  Your original case `''foo'''`, by contrast, seems like something that would come up naturally.

tarleb:
This looks ready but is still marked as draft. @silby, can we merge this?

PR #10799, opened 2025-04-23
Use `int-native` subtagged image

benz0li:
- Related to https://github.com/benz0li/ghc-musl/commit/53efca3dc38c0534603d53aa585ac3d71aada152


benz0li:
Because IMHO the following still applies:

> ### GMP licensing restrictions
> 
> The regular <nobr>*GHC musl*</nobr> images produce binaries linked against the [GNU Multiple Precision Arithmetic Library (GMP)](https://gmplib.org/), which is used by default by the [`integer-gmp`](https://hackage.haskell.org/package/integer-gmp) library to provide a big-integer implementation for Haskell.
> 
> Unlike most Haskell code, which is licensed under the permissive BSD3 license, the GMP library is licensed under LGPL. This means resulting *statically linked* binaries [must be provided with source code or object files](http://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic).
> 
> If that is not acceptable for your situation, use images with the `int-native` subtag. These images provide a GHC that links against the Haskell-native big-integer backend and produces *statically linked* binaries that and are not
> subject to GMP's licensing restrictions.  
> :information_source: Available for versions 9.6.7, 9.8.4, 9.10.1, 9.12.2 and later.

– https://github.com/benz0li/ghc-musl#gmp-licensing-restrictions

benz0li:
Cross references:

* https://github.com/commercialhaskell/stack/pull/6722
* [FAQ - The Haskell Tool Stack](https://docs.haskellstack.org/en/stable/faq/#windows-related) > What is licensing restrictions on Windows?

jgm:
To be honest, I am not sure I understand the motivation for this.
Pandoc is already GPL licensed, so depending on an LGPL licensed library changes nothing, right?

benz0li:
> Does the LGPL have different requirements for statically vs dynamically linked modules with a covered work?
> For the purpose of complying with the LGPL (any extant version: v2, v2.1 or v3):
> 
> (1) If you statically link against an LGPLed library, you must also provide your application in an object (not necessarily source) format, so that a user has the opportunity to modify the library and relink the application.
> 
> (2) If you dynamically link against an LGPLed library already present on the user's computer, you need not convey the library's source. On the other hand, if you yourself convey the executable LGPLed library along with your application, whether linked with statically or dynamically, you must also convey the library's sources, in one of the ways for which the LGPL provides.

– https://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic

jgm:
> you must also provide your application in an object (not necessarily source) format

OK, but since pandoc is GPL licensed we're already obligated to provide the application in source format, aren't we?

benz0li:
But what about https://github.com/jgm/pandoc/releases/download/3.6.4/pandoc-3.6.4-1-amd64.deb and https://github.com/jgm/pandoc/releases/download/3.6.4/pandoc-3.6.4-linux-amd64.tar.gz?

Because of

> [...] if you yourself convey the executable LGPLed library along with your application, whether linked with statically or dynamically, you must also convey the library's sources, in one of the ways for which the LGPL provides.

The library's (GMP's) sources are not part of e.g. https://github.com/jgm/pandoc/releases/download/3.6.4/pandoc-3.6.4-1-amd64.deb or [pandoc-3.6.4-linux-amd64.tar.gz](https://github.com/jgm/pandoc/releases/download/3.6.4/pandoc-3.6.4-linux-amd64.tar.gz) (your application).

benz0li:
I interpret whatever is written [by 'legal experts'[^1]] to the best of my knowledge.

To be sure: Ask a 'legal expert'.

[^1]: mostly self proclaimed

jgm:
The way pandoc is distributed in these Debian packages is more or less how Debian and other linux distros distribute it.  They don't include the actual sources; it's considered sufficient to include something that tells you where to find the sources.
(If we don't do THAT, then we should, but my memory is that we do include COPYING and COPYRIGHT; the latter says (actually wrongly) that "pandoc's complete source code is available from https://pandoc.org).  (We should change that to link to the github page.)

benz0li:
@jgm Thank you for the clarification.

P.S.:

> [...]
> 
> EDIT: The above are now met, so I have turned back on statically-linked Stack executables linking against GMP.

_Originally posted by @mpilgrem in https://github.com/commercialhaskell/stack/issues/6722#issuecomment-2825555405_

benz0li:
> [...] my memory is that we do include COPYING and COPYRIGHT; the latter says (actually wrongly) that "pandoc's complete source code is available from <https://pandoc.org>. (We should change that to link to the github page.)

If you do that, you may revert this merge request.

jgm:
Perhaps we need to add something to COPYRIGHT saying that statically linked builds of pandoc include GMP, and stating its license and linking to its source code.

Issue #7172, opened 2021-03-22
Remove spaces after list marker


reagle:
When pandoc converts a list to markdown, the list marker is followed by three spaces. A few people seem to prefer this not be the case. @jgm [wrote](https://groups.google.com/g/pandoc-discuss/c/XIa4iQmbmag/m/8jgwWHVyCAAJ):

> The motivation is to make sure embedded lists and block-level elements line up properly, obeying the "four-space rule."  Note that pandoc's markdown reader no longer obeys the four-space rule, and one could imagine revising the writer to be sensitive to the `four_space_rule` extension flag. Nonetheless, obeying it creates the most portable markdown output.

uvtc:
When converting *to* markdown, I think that rather than formatting lists like:

~~~
Lorem ipsum dolor sit amet.

-   foo
-   bar
-   baz

Lorem ipsum dolor sit amet.
~~~

(no space before the list marker, and three spaces after) they should instead be formatted like this:

~~~
Lorem ipsum dolor sit amet.

  - foo
  - bar
  - baz

Lorem ipsum dolor sit amet.
~~~

with two spaces before the list marker and one space after.

In both cases the list content starts at 4-places in. Rationale:

  * indentation before the list marker is how its generally rendered in html and pdf output, and looks better. For reference, see attached screenshots of html and pdf output.
  * if you don't include any leading indentation before the list marker, then as you scan your eyes down the plain text document, everything looks jammed to the left. It just looks better with that indentation before the marker. And markdown should firstly be natural looking and easy to read.

***

![image](https://user-images.githubusercontent.com/212308/112148214-ffaf0880-8bb3-11eb-94a7-e695e16246dc.png)

![image](https://user-images.githubusercontent.com/212308/112148240-076ead00-8bb4-11eb-8d13-61fe5a8a9ba3.png)

uvtc:
Nested lists example. Here's the markdown:

~~~
Lorem ipsum dolor sit amet.

  - foo
      - foo a
      - foo b
  - bar
      - bar a
      - bar b
          - bar b 1
          - bar b 2
      - bar c
  - baz
  - moo

Lorem ipsum dolor sit amet.

-   foo
    -   foo a
    -   foo b
-   bar
    -   bar a
    -   bar b
        -   bar b 1
        -   bar b 2
    -   bar c
-   baz
-   moo

Lorem ipsum dolor sit amet.
~~~

Here's the pdf output (of course, both styles render identically):

[...]

PR #10813, opened 2025-04-28
Plain writer: Support the four_space_rule extension.

### LINKED ISSUE #10812: Plain writer lacks support for four_space_rule extension
The fix for #7172 (first included in pandoc 3.6.4) also affects the output of markdown -> plain transformation. Because `+four_space_rule` is not supported by the Plain writer, it is not possible to emulate the output of previous pandoc versions.

A markdown -> plain transformation is useful e.g. for pretty-formatting git messages in a pre-commit hook. In such uses, the new behaviour of pandoc introduces a change in the formatting of the messages which currently cannot be avoided.

Command:
```shell
pandoc -f markdown -t plain+four_space_rule
```

Input:
```console
This is the title

Here we fix:

- a
- b
- c
```

Output:
```console
This is the title

Here we fix:

- a
- b
- c
```

Expected output:
```console
This is the title

Here we fix:

-   a
-   b
-   c
```

m000:
Allows using `-t plain+four_space_rule` to emulate the output of pandoc before #7172. This is useful for users that use pandoc e.g. to autoformat commit messages.

Fixes #10812.

### DIFF
--- a/src/Text/Pandoc/Writers/Markdown.hs
+++ b/src/Text/Pandoc/Writers/Markdown.hs
@@ -794,6 +794,9 @@ bulletListItemToMarkdown opts bs = do
               Markdown
                 | isEnabled Ext_four_space_rule opts
                   -> "- " <> T.replicate (writerTabStop opts - 2) " "
+              PlainText
+                | isEnabled Ext_four_space_rule opts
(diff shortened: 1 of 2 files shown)


jgm:
Thanks!

m000:
> Thanks!

Wow, that was fast! *I* thank you! :-)

PR #10835, opened 2025-05-10
Powerpoint writer: use reference-doc font for captions.

### LINKED ISSUE #9896: Figure captions not inheriting master slide font settings
**Explain the problem.**

When pandoc renders a pptx from an md with a figure in it that has a caption, it does not use the font defined in the master slide template for the caption.

- In this example in the repo I set the master slide fonts to `Verdana` and the subtitle comes out in `Calibri`.
- This behavior is the same with tables and the tables captions
- This behavior was initially posted in a [quarto discussion](https://github.com/quarto-dev/quarto-cli/discussions/10052) and was diagnosed as a bug in pandoc and not quarto by the developers.

```
git clone git@github.com:yonicd/test_pandoc_pptx.git
pandoc test.md -o test.pptx --reference-doc=template.pptx
```

**Pandoc version**

```
pandoc --version
pandoc.exe 3.2
Features: +server +lua
Scripting engine: Lua 5.4
```

rnwst:
I know nothing about PresentationML, nor do I understand much about what is happening in the Haskell code (as it turns out a weekend is not enough to learn Haskell), but this change seems to fix #9896.


jgm:
Run `make TESTARGS=--accept`; that should update the golden tests.
Then you might want to open them up with Powerpoint and make sure they look good.

rnwst:
Thanks for the tip. I opened the updated files in Powerpoint to compare before/after, and didn't notice any breaking changes. That being said, quite a few of the PPTX files seem to have broken formatting to begin with (see below for an example -- that one's from `test/pptx/comparison/extra-image/templated.pptx`):
image

jgm:
Not sure why.  But as long as nothing that worked before breaks, we're good.

PR #10843, opened 2025-05-14
Lua: allow to fill the ersatz file system of the read sandbox

tarleb:
The `pandoc.read` function now allows to populate the ersatz file system of the sandbox via the (optional) fourth parameter. The given value must be a table that lists the files which will be read from the file system before entering the sandbox. It's also possible to use a mapping from filenames to contents in the table.


tarleb:
This comes with a new public function in T.P.Class, thus requiring a minor version bump.

Issue #10768, opened 2025-04-08
Add Attribute Support for Paragraphs: Enabling Para Elements to Carry HTML Classes, IDs, and Custom Attributes


Valgard:
**Describe your proposed improvement and the problem it solves.**

I propose to extend the `Para` constructor in Pandoc's AST to include attributes (ID, classes, key-value pairs), similar to how `Header`, `CodeBlock`, and `Div` elements already support attributes.

Current definition:
```haskell
data Block
  = ...
  | Para [Inline]
  | ...
```

Proposed definition:
```haskell
data Block
  = ...
  | Para Attr [Inline]
  | ...
```

This would solve the problem of preserving paragraph-level attributes (like IDs, classes, and other HTML attributes) during document conversion. Currently, when converting HTML with paragraphs that have CSS classes or IDs, this information is lost unless the paragraph is wrapped in a Div. This feature would make attribute handling more consistent across block elements.

This proposal addresses part of the broader goal described in issue #684 ("Permit adding attributes to all Markdown elements"), focusing specifically on paragraph elements as a first step. Paragraphs are one of the most common elements in documents, making this a high-value improvement that affects many conversion scenarios.

Implementation would include:
1. Updating the Block data type in Text.Pandoc.Definition
2. Extending the builder in Text.Pandoc.Builder by adding a `paraWith` function and modifying the existing `para` function
3. Modifying the HTML reader to capture paragraph attributes
4. Updating all writers to respect paragraph attributes
5. Adding appropriate tests

This would be consistent with the existing pattern used for other block elements that support attributes.

**Describe alternatives you've considered.**

1. **Wrapping paragraphs with attributes in Divs**: This is the current workaround, but it creates an additional nesting level that complicates the document structure and isn't semantically accurate.

2. **Using a `Maybe Attr` field**: I considered using `Maybe Attr` to maintain backward compatibility, but this would be inconsistent with other block types like `Header` and `CodeBlock` which use a direct `Attr` field with `nullAttr` for no attributes.

3. **Creating a new `ParaWith` constructor**: This would preserve the existing `Para` constructor unchanged, but would introduce redundancy and complicate pattern matching throughout the codebase.

4. **Custom filter for post-processing**: A filter could be used to convert specially-marked Divs to paragraphs with attributes, but this would be a workaround rather than a proper solution.

The direct addition of an `Attr` field to the `Para` constructor seems most consistent with the existing design of the Pandoc AST and would provide the cleanest interface for users of the library.

[...]

PR #10850, opened 2025-05-17
Preserve attributes on HTML paragraphs

Valgard:
This PR implements the preservation of attributes on HTML paragraphs, addressing issue #10768.

HTML reader now wraps attributed &lt;p&gt; tags in a Div with wrapper="1".
HTML writer unwraps these Divs back to attributed &lt;p&gt; tags.

This approach is similar to the Djot reader/writer as discussed in #10768, ensuring that semantic information in HTML attributes on paragraphs is preserved during conversion.


src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -623,13 +623,49 @@ pPlain = do
      then return mempty
      else return $ B.plain contents
 
-pPara :: PandocMonad m => TagParser m Blocks
-pPara = do
+-- Helper function for pPara when significant attributes are present
+pParaWithWrapper :: PandocMonad m => Attr -> TagParser m Blocks
+pParaWithWrapper (ident, classes, kvs) = do
+  guardEnabled Ext_native_divs -- Ensure native_divs is enabled for this behavior
+  pInhalt <- trimInlines <$> pInTags "p" inline
```
jgm, line 630: I usually use the naming convention of beginning parsers with `p`; so it would be better to use something like `inhalt` instead for this name.

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -623,13 +623,49 @@ pPlain = do
...
+  (do guardDisabled Ext_empty_paragraphs
+      guard (null pInhalt)
+      return mempty) <|> do
+    let wrapperAttr = ("wrapper", "1")
+    let alignValue = lookup "align" kvs
+    let otherKVs = filter (\(k,_) -> k /= "align") kvs
+    let validAlignKV = case alignValue of
+                         Just algn | algn `elem` ["left","right","center","justify"] -> [("align", algn)]
+                         _ -> []
+    let finalKVs = wrapperAttr : (validAlignKV ++ otherKVs)
```
jgm, line 640: What is the motivation for treating the "align" attribute specially in this way?

src/Text/Pandoc/Readers/HTML.hs
```diff
@@ -623,13 +623,49 @@ pPlain = do
...
   contents <- trimInlines <$> pInTags "p" inline
+  let paraBlock = B.para contents
   (do guardDisabled Ext_empty_paragraphs
       guard (null contents)
-      return mempty)
-    <|> return (B.para contents)
+      return mempty) <|>
+    return (case alignValue of
+              Just algn | algn `elem` ["left","right","center","justify"] ->
+                            B.divWith ("", [], [("align", algn)]) paraBlock
```
jgm, line 654: I don't understand the motivation for this.

jgm (commented):
Thanks for this.  As noted I didn't understand the special treatment of "align".

Another question I have is how common it is for paragraphs to have classes or other attributes in HTML in the wild.  If it is very common, then I suppose this change will lead to more cluttered HTML -> markdown conversions and we'd need to weight that.

Valgard:
> Thanks for this. As noted I didn't understand the special treatment of "align".
> 
> Another question I have is how common it is for paragraphs to have classes or other attributes in HTML in the wild. If it is very common, then I suppose this change will lead to more cluttered HTML -> markdown conversions and we'd need to weight that.

Thanks for catching that! You're right about the align logic - that was actually an idea I had initially discarded, but it seems to have somehow made its way into the pull request anyway. I'll remove that special handling.

Regarding your question about paragraph attributes in the wild - you're absolutely right to be concerned. Paragraphs with classes and other attributes are extremely common in modern HTML, especially with:

- CSS frameworks (Bootstrap's `text-center`, `lead`, `text-muted`)
- Utility-first frameworks (Tailwind's `text-lg`, `mb-4`, `text-gray-600`) 
- CMS-generated content (WordPress, Drupal automatically add classes)
- JavaScript hooks (`js-expandable`, `track-click`)
- Semantic styling (`introduction`, `disclaimer`, `highlight`)

A configurable approach would be ideal here. We could add a command-line option or extension setting that controls attribute preservation behavior:

- **Default mode**: Strip most attributes for clean, readable Markdown
- **Preserve mode**: Keep essential attributes (IDs for anchor links, alt text for images)
- **Full preservation**: Maintain all attributes for round-trip conversion

This would allow users to choose between clean output (which most expect from HTML→Markdown conversion) and technical preservation (needed for specific use cases like documentation sites requiring anchor links). The majority of conversions prioritize readability over technical fidelity, so defaulting to clean output while providing flexibility makes the most sense.

src/Text/Pandoc/Readers/HTML.hs
Valgard: Good point! I've renamed it to `contents` to follow your `p` prefix convention for parsers. Thanks for catching that!

Valgard:
Hi @jgm,

It's been about 3 weeks since my last message and I haven't heard back yet. I wanted to follow up to see how you'd like me to proceed with this pull request.

I've already removed the special handling for the align attribute as discussed. The main question remaining is how to handle the broader issue of attribute preservation, particularly given how common classes and other attributes are on paragraphs in real-world HTML.

Should I:
- Implement a configurable approach for attribute handling as we discussed?
- Keep the current behavior and accept that some HTML→Markdown conversions might be cluttered?
- Take a different approach entirely?

I'm happy to iterate on this or adjust the direction - just let me know what works best for the project.

Thanks!

tarleb:
Great work!

My two cents on the matter: a new extension would probably be best, and the default should be to remove the "clutter".

Valgard:
Hi @jgm,

I've completed the implementation of the `paragraph_attributes` extension as discussed. The solution follows the approach we agreed upon:

**Implementation Summary:**
- Extension only available for HTML output format (`-t html+paragraph_attributes`)
- HTML reader always creates wrapper Divs for paragraphs with attributes
- HTML writer unwraps wrapper Divs to `<p>` tags when extension is enabled, strips attributes when disabled
- All other writers unwrap wrapper Divs to plain paragraphs (attributes always discarded)
- Native writer preserves complete AST unchanged

**Key Features:**
- Clean default behavior: attributes are stripped without the extension
- Opt-in functionality: users enable extension only when needed
- No attribute filtering: all attributes preserved when extension is active
- Consistent wrapper Div handling across all writers
- Full test coverage with no regressions

The implementation is ready for review. All tests pass and the extension works as intended for the use cases we discussed (EPUB index processing, semantic markup preservation, etc.).

Thanks for your guidance on the architecture - the extension-based approach works really well!

jgm:
> - HTML writer unwraps wrapper Divs to <p> tags when extension is enabled, strips attributes when disabled

Is there a reason not to just always unwrap these?  The other writers are doing that, and I don't see what's gained by throwing away this information.

jgm:
Update: in response to #11014 I added (unconditional) unwrapping of wrapper Divs to the HTML writer.  In addition, thinking about this issue made me realize that `paragraph_attributes` is probably too narrowly focused.  What about e.g. attributes on a `ul` element?  I think a better approach would be an extension `wrapper_divs` that causes the HTML reader to create wrapper divs whenever an element contains attributes that would otherwise be ignored because there is no space for attributes in the corresponding AST constructor. Thoughts?

Valgard:
Hi @jgm!

Thanks for the update and for adding the unconditional unwrapping to the HTML writer in response to #11014!

You're absolutely right that `paragraph_attributes` is too narrowly focused. The `wrapper_divs` extension approach is much more elegant and would handle `<ul>`, `<ol>`, `<li>`, `<blockquote>`, and other elements that face the same attribute preservation issue.

Interestingly, I've actually already implemented unconditional unwrapping of wrapper Divs across all writers in this PR (including the HTML writer 😆) - so we're aligned on that approach! The current implementation unwraps wrapper Divs to plain elements regardless of any extension settings.

Given that the unconditional unwrapping is already in place, would it make sense to complete this PR as-is and then implement the broader `wrapper_divs` extension as a follow-up? My thinking:

1. **This PR demonstrates the pattern working** across all writers with unconditional unwrapping
2. **The broader scope** (all block elements + potential consolidation of existing attribute extensions) might be better suited for a separate, focused PR
3. **Incremental progress** - we get immediate value for paragraph use cases while planning the comprehensive solution

The current implementation actually provides a solid foundation for `wrapper_divs` since the unwrapping mechanism is already proven to work consistently across all writers.

What's your preference - finish this focused PR and follow up with the broader `wrapper_divs` extension, or expand this PR to cover the full scope?

jgm:
No, I think it's better not to introduce `paragraph_attributes` if it is just going to be obsoleted by `wrapper_divs`.  Sorry to introduce this wrinkle into your PR!

thom4parisot:
What about this `wrapper_divs` extensions?  I could not find anything about it in the documentation.

jgm:
`wrapper_divs` is suggested as a possible extension above, but not yet implemented or even decided on.

PR #10878, opened 2025-05-27
HTML writer: use the ID prefix in the ID for the footnotes section.

bdesham:
In commit 9190b19fc, we began to use the ID prefix for the `<aside>` holding the footnotes. In commit 41da8ad9e, we started to use a `<section>` for this instead and the ID-prefix functionality was lost. This commit resumes using the ID prefix in the ID of this element so that multiple documents can be combined into one without ID conflicts.


jgm:
great, thanks!

Issue #7022, opened 2021-01-12
Western fonts cannot be changed in the output docx


TomBener:
I use Pandoc 2.11.3.2 on macOS 11.1.

When the `lang` is set as `zh-CN` in yaml block:

```yaml
lang: zh-CN
```

Then run the command:

```sh
$ pandoc -C input.md -o main.docx 
```
The western fonts in the output `main.docx` is the same as the Chinese fonts `宋体` like the screen below. In addition, it seems the western fonts are frozen which can’t be changed, although the default fonts in the template of Pandoc for western fonts are `Calibri` and `Cambria`.

![CleanShot 2021-01-12 at 15 02 06@2x](https://user-images.githubusercontent.com/49151155/104281247-2f70ce80-54e8-11eb-93c0-e5f7f9424488.png)

If the `lang: zh-CN` is removed, the issue will disappear, but some other problems emerge. For example, the Chinese quotation marks `“”` and `‘’` are set as western fonts, which should be avoided.

So I'd like the western fonts are with different fonts from Chinese fonts `宋体`. Is it possible when `lang: zh-CN` is added or other solutions? Thank you!

jgm:
I don't know. To my knowledge, we don't do anything to change the font depending on the `lang`.
You might have a look at `data/docx/word/styles.xml` in this repository and let us know if anything there seems wrong.

TomBener:
After nearly 18 months, I’ve tried to look into this issue again. Sorry for the long time…

---

With the option `--metadata lang=zh-CN`, i.e. `pandoc main.md -o main.docx --metadata lang=zh-CN`, there will be two changes in the output document compared to the default command `pandoc main.md -o main.docx`.

1. The property `w:val` for [specifying Latin Languages](https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.languages?view=openxml-2.8.1#properties) were changed from `en-US` to `zh-CN` in `word/styles.xml`.
2. `<dc:language>zh-CN</dc:language>` was added in `docProps/core.xml`.

The first change is the cause for this issue, as the Chinese language is _not_ the Latin Language but the **East Asian Language**.

[...]

PR #10901, opened 2025-06-10
DOCX template: Improve default handling of East Asian languages

TomBener:
This pull request improves the default `reference.docx` to better support documents containing East Asian text (Chinese, Japanese, Korean) when **no explicit `lang` metadata** is provided.

**The Problem**

Currently, the default DOCX template that ships with Pandoc contains two settings in [`data/word/styles.xml`](https://github.com/jgm/pandoc/blob/main/data/docx/word/styles.xml) that lead to suboptimal rendering and proofing of East Asian characters:

1.  The default run properties map East Asian characters to the `minorHAnsi` theme font, which is intended for Latin scripts: `<w:rFonts ... w:eastAsiaTheme="minorHAnsi" ... />`.
2.  The default language settings explicitly tag East Asian text as US English: `<w:lang ... w:eastAsia="en-US" ... />`.

When a user converts a Markdown file containing CJK characters without specifying a `lang` variable (e.g., `lang: zh-CN`), Microsoft Word is instructed to treat the text as English. This can lead to incorrect font rendering (e.g., using a Japanese font for Chinese characters or vice-versa) and issues with spell-checking and other proofing tools.

**The Solution**

This PR makes two small but significant changes to the default `data/word/styles.xml` to create a more language-neutral starting point:

1.  **Use the correct theme font for East Asian text.** The `eastAsiaTheme` attribute is changed from `minorHAnsi` to `minorEastAsia`. This allows Word to use the theme's designated East Asian font for CJK characters, improving rendering.

2.  **Set a sensible default for the East Asian language.** The `w:eastAsia` attribute is changed from `en-US` to `zh-CN`. This approach parallels the existing behavior for bidirectional text, where the template already sets a specific language default (`w:bidi="ar-SA"`). Given that Chinese is the most widely spoken language in East Asia, `zh-CN` serves as a more useful and practical default than `en-US`. And it has been [noted](https://github.com/jgm/pandoc/issues/7022#issuecomment-1236589386) this is the default behavior in Microsoft Word. This ensures that in the absence of more specific metadata from the user, Microsoft Word will default to using proofing tools for Chinese rather than English for CJK characters, which is a significant improvement for a large number of users.

The screenshots below compares the language recognition in Microsoft Word before and after the current revision.

CleanShot 2025-06-10 at 16 21 00@2x

CleanShot 2025-06-10 at 16 21 31@2x

### DIFF
--- a/data/docx/word/styles.xml
+++ b/data/docx/word/styles.xml
@@ -3,7 +3,7 @@
   <w:docDefaults>
     <w:rPrDefault>
       <w:rPr>
-        <w:rFonts w:asciiTheme="minorHAnsi" w:eastAsiaTheme="minorHAnsi" w:hAnsiTheme="minorHAnsi" w:cstheme="minorBidi" />
+        <w:rFonts w:asciiTheme="minorHAnsi" w:eastAsiaTheme="minorEastAsia" w:hAnsiTheme="minorHAnsi" w:cstheme="minorBidi" />
         <w:sz w:val="24" />
         <w:szCs w:val="24" />
         <w:lang w:val="en-US" w:eastAsia="zh-CN" w:bidi="ar-SA" />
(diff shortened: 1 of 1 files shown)


jgm:
I'm sorry - I can't spot the difference between these pictures. What should I be looking at?

TomBener:
At the bottom left corner, the first picture shows that selected Chinese characters `中文` are recognized as `English (United States)`. After the revision, in the second picture, selected Chinese characters `中文` are recognized as `Chinese (Simplified, Mainland China)`, which is expected.

jgm:
I can fix the tests. I'll merge this now.

PR #10922, opened 2025-06-18
Typst writer: rename *numbering* variable to *section-numbering*.

tarleb:
This is the name expected by the default template. This should fix the `--number-sections`/`-N` option when used with Typst.


tarleb:
The change *might* break custom templates, but that seems extremely unlikely.

jgm:
I agree. Good catch.

PR #10929, opened 2025-06-23
Lua: add more UTF-8-aware file operations to `pandoc.system`.

tarleb:
Functions that expect UTF-8-encoded filenames should make it easier to write platform-independent scripts, as the encoding of the actual filename depends on the system.

Additionally, this also adds a generalized method to run commands, and functions to retrieve XDG directory names.

The new functions are `command`, `copy`, `read_file`, `remove`, `rename`, `times`, `write_file`, `xdg`.


tarleb:
Resolves #10877

Issue #6571, opened 2020-07-28
Lua filters doc: reintroduce example on link info table


tarleb:
The original example was:

```` markdown
## Extracting information about links

This filter prints a table of all the URLs linked to in the
document, together with the number of links to that URL.

``` lua
links = {}

function Link (el)
  if links[el.target] then
    links[el.target] = links[el.target] + 1
  else
    links[el.target] = 1
  end
  return el
end

function Doc (blocks, meta)
  function strCell(str)
    return {pandoc.Plain{pandoc.Str(str)}}
  end
  local caption = {pandoc.Str "Link", pandoc.Space(), pandoc.Str "count"}
  local aligns = {pandoc.AlignDefault, pandoc.AlignLeft}
  local widths = {0.8, 0.2}
  local headers = {strCell "Target", strCell "Count"}
  local rows = {}
  for link, count in pairs(links) do
    rows[#rows + 1] = {strCell(link), strCell(count)}
  end
  return pandoc.Doc(
    {pandoc.Table(caption, aligns, widths, headers, rows)},
    meta
  )
end
```
````

The code is outdated and doesn't work with pandoc 2.10.

tarleb:
Fixed in #10956.

Issue #6569, opened 2020-07-28
Table constructor not documented or usable in Lua filter


mjfs:
Lua table constructor in Pandoc 2.10 (as well as in 2.10.1) does not match released Lua filters documentation (https://pandoc.org/lua-filters.html).

The latest documentation in Pandoc repository does mention changes to Lua table, but the example on how to construct the table (e.g. in "Extracting information about links" section) is not updated (and consequently does not work).

All attempts to construct a table in Lua filter fail with a non descriptive error, containing no line number or element name e.g.: 

`Could not get Pandoc value: Could not read list: Could not get Block value: expected table, got 'nil' (nil)`

It therefore appears that it is not possible to construct a table in Lua filter with recent Pandoc releases.

kysko:
See #6516, _"more work"_ will be done as he said. I'm sure he just missed that particular example (the only one involving a `Table`).

Meanwhile, until more is done on that side, you might consider the [conversion functions I created in pandoc-discuss](https://groups.google.com/forum/#!msg/pandoc-discuss/E1k7Wf9VGwQ/GsiUHhG3BQAJ). You might take some hints from the `import_from` function for the new _structure_, if I've interpreted well their meaning. I've done this to learn how the new `Table`s worked, hoping it helps you. Adapt to your needs.

(Note that in releases after 2.10.1, the table constructor will have to be of the form:

```
 pandoc.Table(caption, colspecs, head, bodies, foot, attr)
```
)

tarleb:
Thanks for the report.

> I'm sure he just missed that particular example

I did indeed. The example has been removed for now (see #6571). Handling tables in Lua filters is quite unpleasant with the current tooling, and I'd like to use the example to drive the development of something better. Converter and convenience functions like those posted by @kysko seem like the way to go. Maybe we can expose `T.P.Writers.Shared.toLegacyTable` and `T.P.Builders.table` to Lua.

tarleb:
Fixed by #10956.

PR #10956, opened 2025-07-08
Add example in documentation on using pandoc.Table constructor.

SeanESCA:
This PR adds a code example of how to use the pandoc.Table constructor in a Lua filter and a few notes on the `colspec` and `bodies` parameters. These changes were made under the pandoc.Table constructor section of the documentation. I think that this will help provide enough information for someone to use this constructor without prior knowledge. This may be relevant to [#6569](https://github.com/jgm/pandoc/issues/6569) and [#6571](https://github.com/jgm/pandoc/issues/6571).

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -767,7 +767,7 @@ Note that:
   constructor.
 
 ```lua
-function Pandoc (blocks, meta)
+function Pandoc ()
   local caption = pandoc.Caption( "This is my table caption." )
   local colspecs = {
     { pandoc.AlignLeft },
@@ -808,10 +808,9 @@ function Pandoc (blocks, meta)
       pandoc.Cell( "This is my table footer.", pandoc.AlignDefault, 1, 4 )
     }
   }
-  return pandoc.Pandoc(
-    { pandoc.Table(caption, colspecs, head, bodies, foot) }, 
-    meta
-  )
+  return pandoc.Pandoc { 
+    pandoc.Table(caption, colspecs, head, bodies, foot) 
+  }
 end
 ```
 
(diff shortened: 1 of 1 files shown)


rnwst:
I think this part of the documentation is auto-generated (see comment at line 2501). @tarleb will know where to make this change.

tarleb:
Thanks for this, those are good improvements. The function is defined here and the docs are then generated from that. https://github.com/pandoc/pandoc-lua-marshal/blob/22c865629b67a78291b3c6eb568ecc79eca1e0ad/src/Text/Pandoc/Lua/Marshal/Block.hs#L463

We should probably add a TableBody constructor.

SeanESCA:
Since the Examples section in lua-filters.md is not auto-generated (right?), I could move my changes as they are to that section. If that works, I could also try to rewrite the link info table example from [#6571](https://github.com/jgm/pandoc/issues/6571) and include it back in.

tarleb:
That's a good idea. 👍

SeanESCA:
I've changed my original example into a filter, and tested both examples on some of my documents and they seem to be working. If the TableBody constructor is a good first issue, I could try taking it on.

tarleb:
Great, thank you! Excellent additionsj! The only suggestion I have is to make it a bit clearer that the argument to `Pandoc` in the second example is discarded, so maybe `function Pandoc (_)`?

A PR to add `TableBody` would be most welcome. Most work would probably happen in the `Text.Pandoc.Lua.Marshal.TableParts` module, or in a new module based on that one. It's a bit DSL-heavy, so if you run into issues, please let me know, and I'll try my best to help.

tarleb:
Thanks!

SeanESCA:
Thanks for the help @tarleb !

PR #10964, opened 2025-07-13
feat: Add track changes to LaTeX conversion

edithatogo:
This change introduces a new Lua filter that converts DOCX track changes (insertions and deletions) into LaTeX commands using the `changes` package.

The filter, `track-changes.lua`, intercepts the `insertion` and `deletion` spans generated by the DOCX reader's `AllChanges` mode and transforms them into `\added` and `\deleted` commands, respectively. The author and date information is preserved in the LaTeX output.

A new test case has been added to verify that the DOCX reader correctly parses the track changes information, and a test script has been created to validate the Lua filter's output.

The documentation has been updated to include a section on how to use the new filter.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -984,6 +984,18 @@ function RawBlock(el)
   end
 end
 
+## SmartArt and Charts to LaTeX
+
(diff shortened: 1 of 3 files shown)


jgm:
This looks like an AI-bot-generated PR.  I only want PRs from people who actually understand the code they are submitting.  In addition, I see the following problems: (a) we don't ship Lua filters with pandoc, (b) the PR contains modifications to the docx reader that aren't described or motivated at all in the description.

Issue #10918, opened 2025-06-17
Docbook reader: support rowspan and colspan


ryan-carpenter:
Row and column spans have been implemented in some writers, including ODT (https://github.com/jgm/pandoc/pull/6682) and DOCX (https://github.com/jgm/pandoc/issues/6315), but these features are missing from the docbook reader. Related: https://github.com/jgm/pandoc/issues/6701

The DocBook [entry tag](https://tdg.docbook.org/tdg/5.1/entry) has been implemented in the reader, but not the `morerows` attribute. 

Here is a DocBook XML example that contains row spans and a column span.

![Image](https://github.com/user-attachments/assets/5af1bc5c-a307-443a-9d94-5a1c80e8a36b)

```xml
<tbody>
<row>
<entry align="left" valign="top" namest="col_1" nameend="col_2"><simpara>A1 + B1</simpara></entry>
<entry align="left" valign="top"><simpara>C1</simpara></entry>
</row>
<row>
<entry align="left" valign="top" namest="col_1" nameend="col_2"><simpara>A2 + B2</simpara></entry>
<entry align="left" valign="top"><simpara>C2</simpara></entry>
</row>
<row>
<entry align="left" valign="top" morerows="1"><simpara>A3 + A4</simpara></entry>
<entry align="left" valign="top"><simpara>B3</simpara></entry>
<entry align="left" valign="top"><simpara>C3</simpara></entry>
</row>
<row>
<entry align="left" valign="top"><simpara>B3</simpara></entry>
<entry align="left" valign="top"><simpara>C3</simpara></entry>
</row>
</tbody>
```

Relevant source:
https://github.com/jgm/pandoc/blob/ccec1e0b65dad97765303e9640b40503961d32ca/src/Text/Pandoc/Readers/DocBook.hs#L1039

SeanESCA:
Hi, I've taken a look at the source and it looks like we just need to read the `morerow` attribute in `parseEntry` and pass it to the cell constructor. If no work has been done yet, I can open a PR and add a test for this.

ryan-carpenter:
I'm not a developer, so I only traced the problem down to what I thought was the root. Thanks for the PR!

jgm:
Closed by #10981

PR #10981, opened 2025-07-22
Add rowspan support to DocBook reader. 

SeanESCA:
This PR adds rowspan support to the DocBook reader by reading the `morerows` attribute in table cells, as mentioned in [#10918]. Previously, the rowspan was always set to 1 regardless of this attribute. I added a test based on the table in the issue as well. 

I also found that the reader does not support multiple rows in the TableHead, and does not consider the number of cells in the Tablehead when setting the number of entries in the `colspecs` parameter. I may open another PR to add these if these are worth looking into.


SeanESCA:
Thanks @tarleb! I've removed the trailing space and changed the unpack import.

tarleb:
The test failure is unrelated to these changes, so I'm merging this.

Thanks!

Issue #10163, opened 2024-09-08
Improve infrastructure for CLI options


jgm:
Adding a CLI option is now a somewhat complex process, involving changes in a number of separate places.

- `MANUAL.txt` -- documentation for new option, both in the list of options and in the section on defaults files.
- `Text.Pandoc.App.Opt` -- new constructor for Opt and default value
- `Text.Pandoc.App.CommandLineOptions` -- the option parser
- `Text.Pandoc.App` or `Text.Pandoc.App.OutputSettings` -- handle the new option
- possibly in pandoc-server: `Text.Pandoc.Server` -- handle the new option

Some day it would be nice to create a slick system so everything can be in one place.  I note that with `--link-images`, for example, not all of these things got done, and that might not be the only case.

jgm:
For example, there is code in T.P.App.OutputSettings to set WriterOptions from Opts, and much of it is duplicated in pandoc-server.

jgm:
Maybe pandoc-lua-engine could use generics to provide a compile-time check that it was marshalling all the fields of WriterOptions.

Issue #9384, opened 2024-01-29
[feature] Json schema support


Freed-Wu:
**Describe your proposed improvement and the problem it solves.**

pandoc has a usage as following:

`defaults.yaml`:

```yaml
---
to: beamer
pdf-engine: xelatex
```

```sh
pandoc -ddefaults.yaml -oXXX.pdf XXX.md
```

Can `defaults.yaml` have a [json schema](https://json-schema.org/) and been published in [schemastore](https://www.schemastore.org/json/)? Like [alacritty](https://github.com/alacritty/alacritty)?

![Screenshot from 2024-01-11 17-31-20](https://github.com/ClementTsang/bottom/assets/32936898/2d4d2226-2259-4e6d-b164-7db6b2ef34cd)

This is a reference about `latexindent.yaml` <https://github.com/cmhughes/latexindent.pl/issues/206>

The steps are following:

- Create a json schema
- Add its URL to schemastore.

perrotuerto:
Hi! As I commented in #11117, I would like to see a JSON Schema for Pandoc JSON AST. I read #10996 and I am under the impression that the ideal JSON Schema would be produced from the code instead of being done and maintained manually. Is that right?

If not, would it be a good idea to at least make a draft from [the module pandoc documentation](https://pandoc.org/lua-filters.html#module-pandoc)?

jgm:
This issue appears to be for the defaults.yaml format, rather than the AST. Could open a new issue for the AST.

PR #10996, opened 2025-07-27
Request for Comments: Add new package "pandoc-codecs"

tarleb:
This is exploratory work, aimed at improving the situation around JSON schemas (#9384) and command line options (#10163). It would include a good bit of busywork, so my intention is to gathering feedback with this PR before doing any of that.

The idea is to use [autodocodec](https://github.com/NorfairKing/autodocodec) to define schemas for pandoc data types. These could then be reused in various places, e.g., to define a JSON schema, thus allowing us to keep all important definitions in one place. Or at least in fewer places, because we'd still be duplicating part of the ToJSON/FromJSON instances from `Text.Pandoc.Definition`, unless we decide to go all-in on autodocodec.

I've put the code in a new package as to avoid an additional dependency in pandoc. Consistency checks for JSON encodings could happen via QuickCheck.

If this whole plan gets a tentative thumbs up, then my next step would be to write a `autodocodec-hslua` package that can use Codec definitions to do Lua marshaling.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,4 +1,5 @@
 packages: .
+          pandoc-codecs
           pandoc-lua-engine
           pandoc-server
           pandoc-cli
--- a/pandoc-codecs/CHANGELOG.md
+++ b/pandoc-codecs/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Revision history for pandoc-codecs
+
+## 0.1.0.0 -- YYYY-mm-dd
+
+* First version. Released on an unsuspecting world.
--- a/pandoc-codecs/LICENSE
+++ b/pandoc-codecs/LICENSE
@@ -1,0 +1,339 @@
+             GNU GENERAL PUBLIC LICENSE
+                Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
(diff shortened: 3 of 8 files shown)


tarleb:
The documentation for autodocodec is a bit sparse, but @sshine wrote a nice post that explains it a bit better: https://dev.to/sshine/json-schema-in-haskell-using-autodocodec-35cg

jgm:
Possibly related: https://github.com/jgm/pandoc/pull/10914#issuecomment-2997095502
though I don't think autodocodecs covers XML schemas!

I'm not sure what I think about this.  Duplicating pandoc-types stuff in a separate package seems a bad idea.  If autodocodecs can simultaneously generate the To/FromJSON instances and the schemas and docs, then maybe it could be integrated in pandoc-types.  (But, one worries about the dependency tree and stability.)

jgm:
I see, looking now at the changes, that this is not (yet anyway) about the Pandoc type, but about the types used for Options etc.

Can you say more about what this gets us? How would it simplify our lives?

It doesn't look like the dependency footprint is too bad.  If there is utility in this, maybe it would be better the schema documentation in the same modules as the types themselves, rather than in a separate package -- which you'd have to remember to update independently on any changes to the types

Issue #10995, opened 2025-07-26
Allow tables of filters to be used in Lua `walk` methods


rnwst:
I was surprised to find out that the [`Pandoc:walk`](https://pandoc.org/lua-filters.html#type-pandoc:walk) method does not support using a table of filters as its argument:
```lua
local doc = pandoc.Pandoc{}

-- This works:
doc = doc:walk { Span = filter_fn1 }:walk{ Span = filter_fn2 }

-- This doesn't work (the filters aren't run):
doc = doc:walk {
   { Span = filter_fn1 },
   { Span = filter_fn2 }
}
```
Given that one can return a table of filters from a Lua filter, I would have expected the same to work for the `walk` method. To be fair, the documentation states nowhere that a table of filters can be used as a function argument -- I just made that assumption.

I suppose the same applies to [`walk_inline`](https://pandoc.org/lua-filters.html#pandoc.walk_inline) and [`walk_block`](https://pandoc.org/lua-filters.html#pandoc.walk_block), as well as the `walk` methods on `Inlines` and `Blocks`, as well as the [`walk`](https://pandoc.org/lua-filters.html#common-methods) method on AST elements with contents (though I haven't tested this to verify).

**Version**: 3.7.0.2

kysko:
For different elements you can do, e.g.:
```
doc = doc:walk {
    Span = filter_fn1,
    Table = filter_fn2
}
```

For same element, you could do:
```
doc = doc:walk { Span = filter_fn1 }:walk { Span = filter_fn2 }
```

tarleb:
The current behavior is intentionally, and the "multiple filters" where originally a bit of a kludge: Initial versions of Lua filters didn't offer the option to do nested filtering with `walk`, or to change the traversal order to *topdown*. Running multiple filters was the only way to do certain modifications. Possibly the most common example was the configuration of the filter via metadata, where the first filter needed to store the metadata into a variable, and then a second filter could access it. Now we have `walk` and lazy marshalling, so filtering the `Pandoc` element and then traversing the document with `walk` has become both possible and efficient.

Long story short: I wanted to keep the interface as simple as possible, and since multiple calls to `walk` shouldn't put much of a burden on filter authors, I decided to allow only a single filter as argument.

rnwst:
Thanks for the detailed backstory @tarleb. I think keeping things as simple as possible is a worthwhile aim, so I'll close this issue.

[...]

PR #11000, opened 2025-07-28
docs: Don't encourage returning tables of filters from Lua filters

rnwst:
See #10995. Use the `Pandoc:walk` method instead.

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -92,13 +92,20 @@ applies all filters (including JSON filters specified via
 `--filter` and Lua filters specified via `--lua-filter`) in the
 order they appear on the command line.
 
-Pandoc expects each Lua file to return a list of filters. The
-filters in that list are called sequentially, each on the result
-of the previous filter. If there is no value returned by the
-filter script, then pandoc will try to generate a single filter
-by collecting all top-level functions whose names correspond to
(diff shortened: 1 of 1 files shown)


jgm:
Aren't there cases where you really need multiple filters?  E.g. suppose you want conversion of the top of a document to depend on something that happens later.

tarleb:
I think the solution with `walk` in a filter on Pandoc elements is sufficient. Maybe we should mention that returning multiple filters works, but that it's slightly discouraged now.

The "Replacing placeholders with their metadata values" example also relies on multiple filters. Maybe we should remove that example entirely, as it doesn't work if the placeholder isn't in it's own `Str` element? Or maybe change it to use something different than `Str` elements as placeholders?

rnwst:
@jgm -- Could you provide an example of a case that cannot be addressed by the `walk` method? I don't think there is one.

@tarleb -- Good spot, I missed that example. I'll update that tonight. Regarding that example, I think it should work fine provided the placeholder doesn't contain any spaces. I can add a note about that. I'll also mention in the [Lua filter structure](https://pandoc.org/lua-filters.html#lua-filter-structure) section that it is possible to return multiple filters, but that this is discouraged.

jgm:
Well, I guess you can do everything you need to do if you put the walks in the top level Pandoc function.

Issue #8819, opened 2023-05-02
Add a comment in the default HTML style template with some instruction on how to disable it and a link to the documentation


hftf:
I finally upgraded Pandoc and discovered that the standalone HTML output comes by default with an extremely opinionated block of CSS. My suggestion is to add a comment at the [very top of this block](https://github.com/jgm/pandoc/blob/52196617c819eb07eb1e1a531e2e2f1eb885127e/data/templates/styles.html#L1) that contains some sort of instruction for how to disable it along with a link to the relevant documentation, because it is not easy to find. For example:

```css
/*
To disable most styles, invoke pandoc with the option `-M document-css=false`.
See further documentation at https://pandoc.org/MANUAL.html#variables-for-html
*/
```

While discussion about the merits of the opinionated styles is out of scope for this particular issue and probably the Pandoc issue tracker as a whole, very briefly, I find the majority of the styles to be not an improvement, not backwards-compatible, and not trivial to audit over 100 lines of CSS. In my opinion, any styles that do not undeniably affect the semantic interpretation of the document, such as the existing style `span.smallcaps{font-variant: small-caps;}`, are out of scope for the default template, and possibly for Pandoc period. The attempt is well-intentioned, but it should use an [opt-in](https://en.wikipedia.org/wiki/CSS_Zen_Garden#:~:text=Aside%20from%20reference%20to%20an%20external%20CSS%20file%2C%20the%20HTML%20markup%20itself%20never%20changes.) design because there is no one clear better option; users should not need to wonder why on earth hyperlinks started showing up as `#1a1a1a`, the same non-black color as the text.

Returning to the issue topic, I think more code comments like this in the default template would be very helpful; unexplained CSS defaults in a document-conversion app is an easy-to-fix [code smell](https://en.wikipedia.org/wiki/Code_smell), and users deserve more than just two comments to not have to dig through a maze of commits, issues, and listserv emails to understand why a style was included.

PR #11002, opened 2025-07-29
HTML styles: prefix default styles with informative CSS comment

tarleb:
Closes: #8819


jgm:
presumably this just needs `make TESTARGS=--accept`

tarleb:
My bad. Fixed.

Issue #1603, opened 2014-09-04
Markdown footnotes are duplicated


lkiesow:
If you add multiple references to one footnote in Markdown like this:

```
This is a sentence [^fn] with footnote.

This is another sentence [^fn] with footnote.

[^fn]: This is the footnote.
```

pandoc will create multiple footnotes instead of referencing only one. For example, if converted into Markdown again, one would end up with:

```
This is a sentence[^1] with footnote.

This is another sentence [^2] with footnote.

[^1]: This is the footnote.

[^2]: This is the footnote.
```

The same happens when converted to HTML (I didn't try other formats). Obviously, this should not happen but instead it should be one footnote which is just linked to multiple points in the text.

I've tested this locally using version 1.12.3.3 as well as with 1.13.1 at [Try pandoc!](http://johnmacfarlane.net/pandoc/try/?text=This+is+a+sentence+%5B%5Efn%5D+with+footnote.%0A%0AThis+is+another+sentence+%5B%5Efn%5D+with+footnote.%0A%0A%5B%5Efn%5D%3A+This+is+the+footnote.&from=markdown&to=markdown)

van-de-bugger:
Why the issue is marked as "enhancement"? I would say it is a bug.

wilx:
What about using the solution described in <http://tex.stackexchange.com/a/262956/28495>?

jgm:
+++ Václav Haisman [Jan 04 16 07:16 ]:
>   What about using the solution described in
>   [1]http://tex.stackexchange.com/a/262956/28495?

If we did make a change allowing multiple references
to a single footnote, I'd probably implement it in
LaTeX without relying on an external package like fixfoot,
to keep down dependencies.

But one thing to keep in mind is that pandoc supports many
output formats, not just LaTeX.  Any changes in how
footnotes are handled require changes in all of them.

van-de-bugger:
It seems this issue has a long history and discussion thread I am not aware of. Moreover, I am not experienced in this area, but…

mpickering wrote:

> It would probably be better to present a uniform solution which works across all output formats.

I would not agree with this. Output formats differ, and uniform solution may not (or even will not) work for all. 

For example: HTML output format assumes one (probably long) page. All the footnotes are printed at the very bottom. And having a separate footnote for every link looks ugly. (Backlinks is not a problem, at least conceptually, look at Mediawiki/Wikipedia solution: they have multiple backlinks in a single footnote.) 

[...]

PR #11009, opened 2025-08-01
Extract citationSuffix, citationPrefix.

### LINKED ISSUE #10894: Provide a way to add a prefix or suffix to a whole citation group
For background see https://github.com/jgm/citeproc/issues/155.

Currently prefixes and suffixes go on individual citation-items (`Citation` in the pandoc types), not on whole citaitons (groups of items, or `Cite` in the pandoc types).

But conceptually, some prefixes and suffixes belong to the whole group rather than the individual items.

For example:  in `[e.g., @Zoo23; @Norway33, and others]` we want `e.g.` at the beginning of the whole group and `and others` at the end, *even if the citations are sorted*.  (So maybe it should render as `(e.g., Norway 1933, Zoo 2023, and others)`.

biblatex and org-mode citations have a way to distinguish between prefixes/suffixes on items and on groups.

To provide this in pandoc, we'd need to make a change to pandoc-types and a change to Markdown syntax.

* pandoc-types:  instead of `Cite [Citation] [Inline]`, we'd need something like `Cite{ citePrefix :: [Inline], citeSuffix :: [Inline], citeItems :: [Citation], citeRendered :: [Inline] }`.
* Markdown syntax: we'd need a way to indicate a prefix or suffix that goes with the whole group instead of individual items.  Perhaps they could be separated from the rest by pipes: `[e.g.,| @Zoo23; @Norway33|, and others]` though this is a bit ugly.
* citeproc: possibly we'd need changes in jgm/citeproc as well.

jgm:
In transforming pandoc Cite to citeproc Citation,
extract a `citationSuffix` and `citationPrefix` from the last item's suffix and first item's prefix, respectively, if they contain a `|` character which separates the item's suffix or prefix from the whole Citation's.

for example:

    [for example, see |@C1; @A3; @B4|, and others]

Here "for example, see" acts as a prefix for the whole group and will remain at the beginning even if the citation items are reordered by citeproc. Similarly, ", and others" will be a suffix for the whole group.

The terminology is confusing, because a pandoc Citation is a citeproc CitationItem, and a pandoc Cite (which can contain multiple Citations) is a citeproc Citation. A citeproc `citationPrefix` is a prefix on the whole group of items.

Closes #10894.

Notes:

1. The org reader now adds global prefixes and suffixes the same way as the Markdown reader: as affixes to the first item's prefix or the last item's suffix, separated by a pipe (`|`).

2. The org writer, however, has not been modified to convert the `|` to a `;`, as required by org-cite syntax.

3. This change doesn't currently do what one would expect, because of changes that were made to citeproc to prevent citation items with prefixes and suffixes from being sorted.  Hence in `test/command/10894.md`, we have test output

   ```
   (Doe, 2020; Smith, 2021)
   ```
   without affixes, but
   ```
   (see Smith, 2021; Doe, 2020, and others)
   ```

   with affixes.  To make this work well, we'd need to remove the citeproc code that prevented bad results before we had proper global prefixes and suffixes.  However, removing this code would mean that existing documents would render differently, unless the new pipe syntax for citation affixes were used.  That may be something we want to avoid.

4. The use of pipes to separate out global affixes from item-level affixes is a kludge that could be avoided if we added additional fields to Cite in the pandoc AST. However, AST changes are disruptive, so perhaps it's not worth doing that.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -5865,6 +5865,19 @@ the suffix as locator by prepending curly braces:
     [@smith, {pp. iv, vi-xi, (xv)-(xvii)} with suffix here]
     [@smith{}, 99 years later]
 
+A prefix or suffix can be marked as going with the whole citation
+rather than an individual item. A horizontal bar (`|`) is used
+to separate the global citation prefix or suffix from the individual
+items:
+
+    [for example, see |@C1; @A3; @B4, in part|, and others]
+
+The global prefix/suffix will always be at the beginning/end
+of the rendered citation, even if the style causes the
+items to be sorted in a different order. By contrast, the
+prefixes and suffixes on each item will move with the items when
+they are sorted.
+
 A minus sign (`-`) before the `@` will suppress mention of
 the author in the citation.  This can be useful when the
 author is already mentioned in the text:
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -6102,6 +6102,19 @@ the suffix as locator by prepending curly braces:
     [@smith, {pp. iv, vi-xi, (xv)-(xvii)} with suffix here]
     [@smith{}, 99 years later]
 
+A prefix or suffix can be marked as going with the whole citation
+rather than an individual item. A horizontal bar (`|`) is used
+to separate the global citation prefix or suffix from the individual
+items:
+
+    [for example, see |@C1; @A3; @B4, in part|, and others]
+
+The global prefix/suffix will always be at the beginning/end
+of the rendered citation, even if the style causes the
+items to be sorted in a different order. By contrast, the
+prefixes and suffixes on each item will move with the items when
+they are sorted.
+
 A minus sign (`-`) before the `@` will suppress mention of
 the author in the citation.  This can be useful when the
 author is already mentioned in the text:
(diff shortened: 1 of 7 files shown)


jgm:
I'm very unsure about whether to merge this one.  As noted above, it doesn't really help much unless we make another change to citeproc, which might break current documents.  And it's a bit hacked together without changes in the Pandoc Inline type.  But, comments welcome. @adunning @tarleb 

If we do end up making AST changes to Note (see #1603), that would be a good time to make changes to the Cite constructor too.  (I'd even be tempted to make the terminology line up with citeproc: Citation -> CitationItem, Cite -> Citation -- or even to put the citeproc types in a separate package that can be used by both pandoc and citeproc.)

tarleb:
I'm all for radical changes once in a while. Seems like both, a good investment and a good time to do it.

Aside: this might also be a good time to revisit endnotes, e.g. jgm/pandoc-types#34.

jgm:
Here's the code in citeproc that blocks sorting of citation items with prefixes/suffixes:
https://github.com/jgm/citeproc/blob/master/src/Citeproc/Eval.hs#L215-L239
We'd have to revert this for this PR to make much sense.  This would mean that existing documents would render differently; users would have to add the special pipe syntax to keep global prefixes and suffixes in place.

PR #11020, opened 2025-08-04
Update 11013.md

reagle:
Original test had a poor/misleading title.

### DIFF
--- a/test/command/11013.md
+++ b/test/command/11013.md
@@ -4,7 +4,7 @@
 references:
 - id: test10
   type: blog-post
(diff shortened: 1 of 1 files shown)


jgm:
Fails because output is wrapped.  I would suggest either changing the expected output accordingly, or adding `--wrap=none` to the command line.

reagle (commented):
I added `--wrap=none`.

jgm:
Now there's a capitalization problem because the style title cases titles...

reagle:
My apologies, had to figure out how to run the test harness. I think it is now correct.

PR #11022, opened 2025-08-04
reference.docx: don't left-align table header row

rnwst:
See #11019. Previously, centering tables in `reference.docx` would leave the header row left-aligned. Why the OOXML 'standard' would allow table elements to be aligned differently from the rest of the table in the first place is anyone's guess.


jgm:
Thanks!

PR #11027, opened 2025-08-06
LaTeX reader and writer: add support for soft hypens

tarleb:
The `\-` command is parsed as a soft hyphen character; likewise, the character is written as that command when outputting LaTeX.


tarleb:
Not sure if there was a reason for this to be left out, so I'm opening a PR.

jgm:
Good addition.

PR #11029, opened 2025-08-08
Add default WebTeX URL to Options.hs [API change]

SeanESCA:
When using WebTeX in a defaults file without giving a URL, it doesn't use the default URL <https://latex.codecogs.com/png.latex?> given in the documentation. I think it is because in Options.hs, the URL defaults to a blank string instead. This PR replaces the blank string with the default URL <https://latex.codecogs.com/png.latex?>.

https://github.com/jgm/pandoc/blob/7639e800c5af85e5ded862a6e218d54489d17bfc/src/Text/Pandoc/Options.hs#L117

https://github.com/jgm/pandoc/blob/7639e800c5af85e5ded862a6e218d54489d17bfc/src/Text/Pandoc/App/CommandLineOptions.hs#L1023-L1029

### DIFF
--- a/src/Text/Pandoc/App/CommandLineOptions.hs
+++ b/src/Text/Pandoc/App/CommandLineOptions.hs
@@ -1023,8 +1023,8 @@ options =
     , Option "" ["webtex"]
                  (OptArg
                   (\arg opt -> do
-                      let url' = fromMaybe defaultWebTeXURL arg
-                      return opt { optHTMLMathMethod = WebTeX $ T.pack url' })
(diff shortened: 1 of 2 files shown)


tarleb:
To elaborate a little: the new variable should probably be added to the export list of the Options module. Only then can it be used in CommandLineOptions the same way that `defaultMathJaxURL` is used there. Options is an exposed module, so this would make it an API change, which should be noted in the commit message, e.g. by appending `[API change]`.

SeanESCA:
Thanks @tarleb, I've made those changes.

tarleb:
Thanks!

PR #11031, opened 2025-08-09
ODT Reader: Add table-header-rows

### LINKED ISSUE #9327: ODT reader drops content of table header
**Explain the problem.**
Hi, when I try to convert a simple odt file that includes a table with header rows that repeat across pages, the content of header rows is missing from the output in any format, including native AST.
If I try the same with a similar docx file and everything seems fine.

`pandoc -t native in.odt -o AST.txt`

**Pandoc version?**
pandoc 3.1.11
Features: +server +lua
Scripting engine: Lua 5.4

[in.odt](https://github.com/jgm/pandoc/files/13887194/in.odt)
[AST.txt](https://github.com/jgm/pandoc/files/13887273/AST.txt)

TuongNM:
Added `table-header-rows` into the table parsing. Since there may be multiple header rows, I had to extend `simpleTable'` for that.
I also had to adjust the existing test files for tables as the previous version was adding empty rows into the header even though there was no header in the source ODT file.

Fixes #9327


tarleb:
Excellent, thanks!

Issue #6550, opened 2020-07-21
Support papersize variable when converting .md to .docx


ad-si:
```sh
$ pandoc --version
pandoc 2.9.2.1
Compiled with pandoc-types 1.20, texmath 0.12.0.1, skylighting 0.8.3.2
Default user data directory: /Users/adrian/.local/share/pandoc or /Users/adrian/.pandoc
Copyright (C) 2006-2020 John MacFarlane
Web:  https://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.

$ cat test.md
---
title: Test
papersize: a4
---

This is just a test!

$ pandoc --output test.docx test.md
```

The test.docx now has a size of 215.9 mm x 279.4 mm (US Letter) and not A4.

agusmba:
AFAIK, `papersize` is only supported on latex-like outputs.
If you want to change the default page size in docx, I would suggest using a custom reference doc.

Whether pandoc's internal default reference docx has sensible-sized pages is something I haven't looked at, since I always use my own reference doc for docx output .

agusmba:
BTW, I just repeated your test (with pandoc version 2.9.1.1 and 2.9.2.1 running on windows and using the official docker image) and both tests have standard "letter" size 215.9mm x 279.4mm, so there must be something else strange on your side.

ad-si:
Well, then this issue is a feature request that `papersize` should be supported for .docx files 😉.

Mh, seems to be a problem of Apple Pages. In LibreOffice it also has the standard letter size.

agusmba:
Fair enough (you might want to change the title of this feature request then).
In the meantime, I'd still suggest that you use a reference-doc with your desired dimensions. This will allow you to also tweak the styling to your hearts content.
BR.

ad-si:
> In the meantime, I'd still suggest that you use a reference-doc with your desired dimensions. This will allow you to also tweak the styling to your hearts content.

If I just want to convert to docx once in a while, that's a little overkill.

ant-222:
*d-si*

> If I just want to convert to docx once in a while, that's a little overkill.

I think it is all right to specify all *docx* settings in the reference document, especially because there are so many. Deciding for which settings to provide alternative means of specification is a fuzzy matter. Once `papersize' is supported, someone else will ask about margins, fonts, *&c*...

ad-si:
> Once papersize is supported, someone else will ask about margins, fonts, &c...

[...]

PR #11035, opened 2025-08-09
reference.docx: Set default paper size to A4

rnwst:
If no paper size is specified, MS Word defaults to US letter size. Since most of the world uses the A4 format, it makes sense to specify this format in `reference.docx`.

Close jgm/pandoc#6550.


alerque:
I don't think this should be changed. It will affect too many people with existing workflows that don't know how to change it.

I say that living and working in an A4 part of the world too. And yes this might have been a good default to have out of the gate. I just think it will be more disruptive than beneficial to change.

rnwst:
@alerque -- I take your point, but I would argue that people with existing workflows likely also have existing reference-docs, and therefore wont' be affected by this change. If not, changing paper size in Word is trivial and can also be done after the document has been generated. I don't have a strong opinion either way, but from #6550 I gathered that most people seem to want A4. Another option might be to have pandoc automatically choose a paper size if none is specified in the reference-doc, based on the [`lang` variable](https://pandoc.org/MANUAL.html#language-variables).

tarleb:
If we change this, then we should probably make sure that it's the same for all formats. Typst intentionally defaults to US Letter in pandoc.  LaTeX defaults to Letter, too, unless when installed through `tlmgr`, as that sets the default to A4. I think the other formats are A4, including weasyprint and prince, as `A4` is the default value for `@page { size: ... }`.

cagix:
this would make life easier for new workflows if based on the lang setting (if this is even possible). 

existing workflows depending on letter would probably still get letter due to the lang setting, and outside the us most workflows probably already set the papersize to a4, so no harm there. the fallback could still be letter if neither papersize nor lang is given. just make sure to be able to overwrite an automatic setting derived from lang.

rnwst:
Thank you all for your comments. Some further thoughts on this. If the user doesn't specify a page size, I think it makes sense to allow the output format tool to use its default page size. This philosophy would seem to me to adhere to the principle of least astonishment.
This seems to be what is being done for LaTeX output, where LaTeX chooses a default paper size if none is specified.
In the case of typst however, that would have to be changed, since pandoc currently specifies US Letter if the user specifies no page size, even though typst's default paper size seems to be A4 (based on cursory research -- please correct me if I'm wrong!). I'm happy to submit a PR for the typst case if we were to agree that we want to adopt this philosophy.

The DOCX case is a little bit trickier. When the reference-doc contains no page size, Word defaults to US Letter. However, when no page size is specified, Word automatically adds a page size to the document on the next save, and chooses US Letter, instead of the page size that is specified in `Normal.dotm` (which is A4 in my case). One might therefore argue that US Letter isn't a default, but rather a random value that Word chooses for a document setting that should never have been left unpopulated in the first place. I am therefore leaning towards using the `lang` variable to let pandoc determine a default page size when none is specified in the reference-doc, and I don't think this PR should be merged anymore. If you agree, I'll close this PR and start work on implementing the `lang` suggestion. If we go down that route, we should also add a note in the manual in the `--reference-doc` section reminding users to update the page size if they use a custom reference-doc.

jgm:
Agreed that this should be closed.  Note also that in various places we default to `en-US` for `lang`.  That fits well with defaulting to US letter for paper size.

The idea of simply not setting a default paper size in Word, if one is not required, is a good one to explore further.

The idea of setting the default paper size based on `lang` is interesting, and maybe makes sense if the correlation between `lang` and paper size is robust enough.  But we'd need to map every `lang onto a paper size.  (It could be that this is easy, if everyone but the US defaults to A4, but I only know about a handful of countries myself)

jgm:
One further reflection:  people who are writing in a language other than `en-US` but living in the US will still want to use letter paper.  So, `lang` will be an imperfect guide to the paper size. Maybe that's okay for a default, but it's also true that a deterministic, constant default is easier to remember and reason about than one that changes depending on something else.

It also means that if we did change the default paper size to depend on `lang`, some existing documents would break (those by authors who write in a language other than `en-US` but live in the US, or who write in `en-US` but live outside the US).

rnwst:
Good point, let's forget about the `lang` suggestion then.

@jgm -- note that our reference-doc doesn't currently include a paper size. This causes Word to set the paper size to US Letter, irrespective of system settings.

I have another suggestion. To get default values for language and paper size, in the absence of a specified reference-doc, pandoc could attempt to parse the user's `Normal.dotm`, just as Word would do when creating a new document.

jgm:
@rnwst - I don't know how we'd locate the user's `Normal.dotm`.  Where is it found?
I'm also hesitant to do this, since it may be confusing to the user where the default is coming from (as it comes from outside the pandoc ecosystem).

jgm:
Would it be better to specify a paper size in the `reference.docx`?
(Then at least it could be changed....or maybe it can already be changed there?)

rnwst:
@jgm -- the user's `Normal.dotm` can be found in `C:\Users\user name\AppData\Roaming\Microsoft\Templates` on Windows and `~/Library/Group Containers/*.Office/User Content/Templates` on Mac, or so [this SO answer](https://superuser.com/a/1075603) claims. Regarding this being confusing, I think a simple note in the manual should address this concern. Note also that I was suggesting only doing this if no `--reference-doc` is specified. If one is specified, its paper size would be used instead. If no reference-doc is given and `Normal.dotm` cannot be found, we could default to US Letter.

Currently, there is no paper size specified in `reference.docx`. As soon as the resulting document is opened and saved, Word sets the paper size to US Letter (irrespective of what's in `Normal.dotm`). The paper size can already be changed, it will just always default to US Letter unless the user chooses otherwise.

jgm:
Is Normal.dotm created automatically when people install Word?

rnwst:
> Is Normal.dotm created automatically when people install Word?

I believe it is, or at the very least when Word is launched for the first time. Apparently if one deletes it, Word recreates it automatically next time Word is launched.

PR #11043, opened 2025-08-12
Lua: add new module `pandoc.highlighting`.

tarleb:


tarleb:
I'm not entirely sure if this is really worth the additional code.

tarleb:
I admit that my primary motivation was to a give a one-line answer to a [StackOverflow question](https://stackoverflow.com/q/62774695/), because I thought it would be neat to get CSS definitions by running

``` lua
pandoc lua -e "print(pandoc.highlighting.definitions('espresso', 'css')"
```

After playing around with this some more, I also really liked the ability to implement a simple syntax highlighter for the console:

``` lua
local infile       = arg[1]
local _, fileext   = pandoc.path.split_extension(infile)
local style = pandoc.highlighting.style('kate')
local wopts = pandoc.WriterOptions{highlight_style = style}
local codeblock = pandoc.CodeBlock(
  pandoc.system.read_file(infile),
  {class = fileext:sub(2)} -- drop leading dot
)

print(pandoc.highlighting.highlight(codeblock, 'ansi', wopts))
```

So maybe it is worth the code?

Issue #11049, opened 2025-08-14
`pandoc.image.size` errors out on JPG


rnwst:
**Explain the problem.**
I stumbled across a JPG that causes `pandoc.image.size` to throw an error: `could not determine image type`. There must be something special about this JPG:
https://github.com/user-attachments/assets/30b45158-a692-4ebb-bd95-b7de81b3c0a4
```console
$ pandoc lua
Lua 5.4.7  Copyright (C) 1994-2024 Lua.org, PUC-Rio
Embedded in pandoc 3.7.0.2  Copyright (C) 2006-2024 John MacFarlane
> _, img = pandoc.mediabag.fetch('toyota-mirai-tank.jpg')
> pandoc.image.size(img)
could not determine image type
stack traceback:
	stdin:1: in main chunk
```
I can open it just fine in any image viewer or editor, and Imagemagick has no problem determining its size:
```console
> identify -ping -format '%w %h' toyota-mirai-tank.jpg
2400 1600
```

**Pandoc version?**
3.7.0.2

jgm:
The file starts with
```
00000000  ff d8 ff ee 00 0e 41 64  6f 62 65 00 64 00 00 00  |......Adobe.d...|
00000010  00 01 ff e1 32 60 45 78  69 66 00 00 4d 4d 00 2a  |....2`Exif..MM.*|
```
For exif, pandoc expects ff d8 ff e1 and that is what I see documented.  I am not sure why this one has ee instead of e1?

PR #11051, opened 2025-08-15
ImageSize: Fix JPEG file signature detection.

rnwst:
The previous implementation didn't include all possible JPEG file signatures. All signatures in the range of `ff d8 ff e0` and `ff d8 ff ef` should be valid, in addition to `ff d8 ff db` (no application segment). See <https://gist.github.com/leommoore/f9e57ba2aa4bf197ebc5?permalink_comment_id=3863054#gistcomment-3863054> and p. 32 of <https://www.w3.org/Graphics/JPEG/itu-t81.pdf>.

See also #11049.

### DIFF
--- a/src/Text/Pandoc/ImageSize.hs
+++ b/src/Text/Pandoc/ImageSize.hs
@@ -61,7 +61,6 @@ import Codec.Compression.Zlib (decompress)
 -- import Debug.Trace
 
 -- quick and dirty functions to get image sizes
--- algorithms borrowed from wwwis.pl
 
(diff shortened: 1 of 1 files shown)


rnwst:
I see the [commit message length check](https://github.com/jgm/pandoc/blob/main/.github/workflows/commit-validation-pr.yml) makes no allowance for URLs. Would you like me to remove the URLs from the commit message or are we going to ignore this check?

jgm:
We can ignore the check.

jgm:
Thanks!

PR #11057, opened 2025-08-16
RST Reader: Add multiple header rows

### LINKED ISSUE #10338: RST: Simple Tables: multiple header rows not supported
```rst
========== =========
Header A1  Header A2
Header B1  Header B2
========== =========
body a1    body a1
body b1    body b2
========== =========
```

The header rows are parsed as body and the header separator becomes a row with HorizontalRules.

Relevant [part of the spec](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#simple-tables):
> A simple table begins with a top border of equals signs with one or more spaces at each column boundary (two or more spaces recommended). Regardless of spans, the top border must fully describe all table columns. There must be at least two columns in the table (to differentiate it from section headers). **The top border may be followed by header rows, and the last of the optional header rows is underlined with '=', again with spaces at column boundaries.** There may not be a blank line below the header row separator; it would be interpreted as the bottom border of the table. 

Perhaps would be reasonable to implement jointly with #10127
Related: #6701

Pandoc 3.5

TuongNM:
This adds support for multiple header rows into tables.

In order to accomplish this, the shared `GridTable` module was to extended to also be able to handle tables with multiple header rows.
Also adjusted the `simpleTableRow` function to prevent headless tables from being interpreted as multiple header rows and fixed hlint warnings about redundant `$`.

Fixes #10338

### DIFF
--- a/src/Text/Pandoc/Parsing/GridTable.hs
+++ b/src/Text/Pandoc/Parsing/GridTable.hs
@@ -34,6 +34,7 @@ import Text.Pandoc.Parsing.General
 import Text.Pandoc.Sources
 import Text.Parsec (Stream (..), ParsecT, optional, sepEndBy1, try)
 
+import Data.Maybe (mapMaybe)
 import qualified Data.Text as T
 import qualified Text.GridTable as GT
 import qualified Text.Pandoc.Builder as B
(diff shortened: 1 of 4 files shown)


jgm:
Thank you! I'm okay with merging this as is, but I'd also be in favor of just changing `tableWith` so it takes a parser that yields a list of header rows (`[[Blocks]]` rather than `[Blocks]`).  After all, the Table type accepts this, and most of the formats that use `tableWith` can have multiple header rows.  The formats that can accept these should be updated to allow it.  @tarleb any thoughts on that?

TuongNM:
After looking at it, the Markdown reader is currently the only other reader to use these table functions. So if you would like to change `tableWith`, it should be fairly easy to adjust the uses inside the Markdown reader, provided I can just leave its current header row behavior as is.

tarleb:
I concur, changing `tableWith` seems like the best option.

TuongNM:
Alright, I changed `tableWith` and made the existing Markdown reader work with it.

Not sure what happened with the CI tests here. Looks like it tried to download a ghc version which somehow failed?

jgm:
I'll try re-running the actions.

PR #11075, opened 2025-08-23
common.latex: add RTL support for LuaTeX engine

rrthomas:


rrthomas:
I'm baffled by the test failure now, sorry: I see the fail message, but the diff between actual and expected output appears to be empty.

jgm:
I'm baffled too. I'll merge this and see if I can figure it out on my end.

jgm:
It was a difference involving a newline at end of file.
I modified your commit and force-pushed to main.

Issue #10816, opened 2025-05-02
Unicode characters in headers are not stripped from identifier when `+auto_identifiers`


cderv:
````
❯ pandoc -t native -f markdown
# Test①
^Z
[ Header 1 ( "test\9312" , [] , [] ) [ Str "Test\9312" ] ]
````

This header is using U+2460 character, which gets kept into the identifier created for the Header.

And it makes Typst rendering fails for example, as it leads to this typ content
````
=== Test①
<test①>
xxx
````

````
❯ pandoc -t typst -f markdown -s -o test.pdf index.md
error: unclosed label
    ┌─ \\?\C:\Users\chris\Documents\DEV_OTHER\00-TESTS\test-quarto\toPB3E1.typ:122:0
    │
122 │ <test①>
    │ ^^^^^

Error producing PDF.
````

For typst 
> A label's name can contain letters, numbers, _, -, :, and .

I would have expected the `+auto_identifiers` extension (https://pandoc.org/MANUAL.html#extension-auto_identifiers) to also support only letters as ANSI letters, when it says
> The default algorithm used to derive the identifier from the heading text is:
> (...)
> Remove all non-alphanumeric characters, except underscores, hyphens, and periods.

I also tried LaTeX and it will no prevent rendering as the `.tex` content is 
````latex
\section{Test①}\label{testux2460}
````

However, it will not render the `①` in document but only in the TOC link

![Image](https://github.com/user-attachments/assets/049a4957-a117-483c-a58d-c75bdf5c90c2)

Should Unicode chars be dropped for identifier creation?  Or at least for Typst where this is not supported like with LaTeX ? 

## Current workaround to this

Manually setting the id is the solution here. 

```
# Test① {#test-1}
```

Originally posted in https://github.com/quarto-dev/quarto-cli/issues/12660

jgm:
One workaround is to use `-f markdown+ascii_identifiers`.

cderv:
My search lead me to 
https://github.com/typst/typst/blob/14241ec1aae43ce3bff96411f62af76a01c7f709/crates/typst-syntax/src/lexer.rs#L1091-L1095

https://github.com/typst/typst/blob/14241ec1aae43ce3bff96411f62af76a01c7f709/crates/typst-syntax/src/lexer.rs#L1075-L1077

It seems they use 
````
use unicode_ident::{is_xid_continue, is_xid_start};
````

From this crate: https://docs.rs/unicode-ident/latest/unicode_ident

[...]

PR #11077, opened 2025-08-24
Typst Writer: Check XID_Continue in identifiers

TuongNM:
Check for `XID_Continue` for the identifiers in typst to match the rust implementation.

Added `unicode-data` as a dependency for this for access to the `isXIDContinue` function to be able to easily check this. I hope this is an acceptable dependency addition in this case.

Also applied the following types of hlint suggestions that came up:
- redundant $
- literal pattern
- moving brackets to avoid $

Fixes: #10816


jgm:
We already have a transitive dependency on unicode-data, so I think that's okay.

jgm:
thank you!

Issue #11076, opened 2025-08-23
PDF generation with typst fails


TuongNM:
The current main branch (at 93f8145f as of this writing) fails to generate PDFs using typst from a simple markdown file.

Compiling pandoc-cli and running `pandoc -t typst -o test.pdf test.md`
causes the following error:

```
error: expected string or array, found none
   ┌─ toPdfViaTempFile177255-0.typ:64:12
   │
64 │     author: authors.map(author => content-to-string(author.name)).join(", ", last: " & "),
   │             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

help: error occurred in this call of function `conf`
    ┌─ toPdfViaTempFile177255-0.typ:142:14
    │  
142 │   #show: doc => conf(
    │ ╭───────────────^
143 │ │   abstract-title: [Abstract],
144 │ │   pagenumbering: "1",
145 │ │   cols: 1,
146 │ │   doc,
147 │ │ )
    │ ╰─^
```

This seems to have been caused by the template changes in https://github.com/jgm/pandoc/pull/9970

[test.md](https://github.com/user-attachments/files/21952769/test.md)

jgm:
@christopherkenny can you have a look at this?

jgm:
I have reverted #9970 for now.

christopherkenny:
Sorry about that @jgm. I just made a small PR which writes the metadata more cautiously.

jgm:
Closed by 6070ad2

PR #11078, opened 2025-08-24
Write Typst data more cautiously

christopherkenny:
In response to #11076. Now makes multiple calls to write document metadata.


data/templates/template.typst
```diff
@@ -29,9 +29,13 @@
...
   set document(
     title: title,
+  ) if title != none
+  set document(
     author: authors.map(author => content-to-string(author.name)),
+  ) if authors != none
+  set document(
     keywords: keywords,
-  )
+  ) if keywords != none
```
jgm, line 38: If I'm not mistaken, `authors` is the only one that really needs this treatment.  For the others, we have checks for `!= none` down below where they are used.  Why move the `content-to-string` business into the clause below where `authors` are rendered -- it already has a check for `none` -- and then this part can be simple again?

christopherkenny: The error I see in the issue is:
`expected string or array, found none`.

This fixes that because `content-to-string` breaks on `none` since you can't access fields.

Is there another issue?

jgm: Oh, I understand now; I wasn't looking at the whole context.

I reverted your original change.  Did you want to integrate those back into this PR?  OR should I apply this now first?

christopherkenny: Conceptually, yes, `content-to-string` creates strings, but it's operating on the map, which is creating an array of strings. This block converts it from an array of `content` which is Typst's type for (broadly put) structured text to an array of `str`s. Then that array is joined element-wise with `,` or `&` if it's the last entry.

Let me rebuild this sometime today and verify.

PR #11082, opened 2025-08-25
Explain how to set descriptive alt text as well as caption

yaxu:
Ref discussion here: https://github.com/jgm/pandoc/discussions/11081

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -5634,7 +5634,8 @@ The link text will be used as the image's alt text:
 
 An image with nonempty alt text, occurring by itself in a
 paragraph, will be rendered as a figure with a caption.  The
-image's alt text will be used as the caption.
+image's alt text will be used as the caption. To set alt text 
+as well as a caption, use the 'alt' link_attribute (see below).
 
     ![This is the caption](/url/of/image.png)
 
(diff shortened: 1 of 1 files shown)


yaxu:
As an aside, as a fellow maintainer of a popular Haskell app, I am moving my projects away from the closed-source [microsoft](https://www.theverge.com/news/757461/microsoft-github-thomas-dohmke-resignation-coreai-team-transition) github for ethical reasons ([e.g.](https://www.bdsmovement.net/microsoft)), and recommend codeberg.org as a nice alternative.

MANUAL.txt
```diff
@@ -5634,7 +5634,8 @@ The link text will be used as the image's alt text:
 
 An image with nonempty alt text, occurring by itself in a
 paragraph, will be rendered as a figure with a caption.  The
-image's alt text will be used as the caption.
+image's alt text will be used as the caption. To set alt text 
+as well as a caption, use the 'alt' link_attribute (see below).
```
jgm, line 5638: I'd suggest using `` `alt` `` instead of `'alt'`.

MANUAL.txt
```diff
@@ -5657,7 +5658,7 @@ and the caption and figure tags will be omitted.
 
 Attributes can be set on links and images:
 
-    An inline ![image](foo.jpg){#id .class width=30 height=20px}
+    An inline ![image](foo.jpg){#id .class width=30 height=20px alt="Alt text"}
```
jgm, line 5661: This is not a good example, because in this inline image, the alt text will be set from the image description. Thus, you can just do `alt text`.  You would only have a reason to use `alt=` when the image is by itself in a paragraph, when the `implicit_figures` extension is set.

jgm:
I'm going to make a change addressing this.

yaxu:
Thanks!

PR #11084, opened 2025-08-25
Proposal to add new pandoc wrapper tool to pandoc.org/extras.html

castedo:
This is a proposal to add an entry to https://pandoc.org/extras.html.

As the author of the tool, obviously I'm not unbiased on whether this entry is worth adding. :sweat_smile:

If y'all think it is worth adding, here's a PR all ready to merge (assuming this is the right source file for the webpage).

Feel free to ask any critical questions or clarifications on this wrapper tool.


jgm:
Sure, seems a worthy addition!

PR #11085, opened 2025-08-26
Support header and footer in MAN reader

### LINKED ISSUE #11079: Man conversion to markdown looses headers
I was generating a couple of markdown files from some manual pages when I noticed that some headers were missing in the conversion.

For example the header of the `curl` man page:
```roff
.TH curl 1 "October 23 2022" "curl 7.86.0" "curl Manual"
```

Results in the following markdown, when running `pandoc -s test.1 --output test.md` with just the above in an empty file:

```md
---
date: October 23 2022
section: 1
title: curl
---

```

In my own testing I see that if you add `header: curl Manual` to the markdown file, the fifth position is correctly filled in; however I have no idea in how the fourth one is "translated". 

```
  .TH curl 1 "October 23 2022" "curl 7.86.0" "curl Manual"
                               _____________
```

I've tried adding it as `source: curl 7.86.0` but no such luck.

I'm quite new with using pandoc, please let me know if this is just a user error.

## Software versions

```
MacOS:     13.2.1
Bash:      5.3.0
pandoc:    3.7.0.2
Features:  +server +lua
Engine:    Lua 5.4
```

SeanESCA:
This PR resolves #11079 by reading the source and manual arguments of the [`TH`](https://manpages.ubuntu.com/manpages/jammy/man7/man-pages.7.html) command into the metadata as footer and header, respectively.


jgm:
thanks!

Issue #11073, opened 2025-08-22
Ods to markdown conversion runs infinitely : stuck


edi9999:
**Problem Description**

When executing the following command:

```
pandoc AntarcticaTemps.ods -o - -t markdown
```

using the file [AntarcticaTemps.ods](https://github.com/user-attachments/files/21944915/AntarcticaTemps.ods), the process consumes 100% CPU and runs indefinitely, or for an extended period of several minutes. The expected behavior is to receive either a result or an error message within 1-5 seconds.

**Pandoc Version**

$ pandoc --version
pandoc 3.7.0.2
Features: +server +lua
Scripting engine: Lua 5.4

jgm:
`.ods` (what is that?) is not a supported format, so pandoc tries interpreting it using the default format, which is markdown.  Not surprising that this causes problems.

edi9999:
Oh I understand.

Ods is the equivalent of xlsx for spreadsheets in libreoffice : https://www.libreoffice.org/discover/what-is-opendocument/

Maybe we could say that if the input file is a zipped file (magic mime of : 50 4B 03 04 → “PK\x03\x04”), then, if we did't recognize it as a handled file, we just throw an error saying we can't convert a zip file ?

jgm:
That's a good idea.

jgm:
Perhaps the best place for this would be in
https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/Class/PandocMonad.hs#L463-L483
We'd need a new constructor for PandocError, something like PandocInputNotText.

edi9999:
Thanks a lot for the contribution @reptee !

PR #11087, opened 2025-08-27
fix: recognize zip and fail early

reptee:
Fixes <https://github.com/jgm/pandoc/issues/11073> by implementing proposed changes <https://github.com/jgm/pandoc/issues/11073#issuecomment-3216812903>. I would like to implement test case or two, but not sure how to test whether converting something fails correctly. For now I verified that by hand. If someone can suggest where do I put the assets (one .zip and one .ods) and how to check that something fails with expected code (I skimmed through tests and see only comparison between produced and expected native output), I will implement the tests as well.

1. Specify .ods as input

   `cabal run ./pandoc-cli -- -i AntarcticaTemps.ods -o - -t native`

   ```
   [WARNING] Could not deduce format from file extension .ods
     Defaulting to markdown
   Expected text as an input, but received file AntarcticaTemps.ods of type zip.
   If you supplied something based on zip (eg. .ods) double check
   the format is supported (.odt/.docx are supported). To force specific
   input format, use -f FORMAT.
   ```

2. Read .ods from stdin

   `cabal run ./pandoc-cli -- -i - -o - -t native < AntarcticaTemps.ods`

   ```
   Expected text as an input, but received stdin input of type zip.
   If you supplied something based on zip (eg. .ods) double check
   the format is supported (.odt/.docx are supported). To force specific
   input format, use -f FORMAT.
   ```

3. Read .ods from stdin and specify reader

   `cabal run ./pandoc-cli -- -f ods -i - -o - -t native < AntarcticaTemps.ods`

   ```
   Unknown input format ods
   ```

4. Provide .docx as input

   `cabal run ./pandoc-cli -- -i test/docx/tabs.docx -o - -t native`

   ```
   [ Para
       [ Str "Some"
       , Space
       , Str "text"
       , Space
       , Str "separated"
       , Space
       , Str "by"
       , Space
       , Str "a"
       , Space
       , Str "tab."
       ]
   , Para [ Str "Tab-indented" , Space , Str "text." ]
   ]
   ```

5. Read .docx from stdin

   `cabal run ./pandoc-cli -- -i - -o - -t native < test/docx/tabs.docx`

   ```
   Expected text as an input, but received stdin input of type zip.
   If you supplied something based on zip (eg. .ods) double check
   the format is supported (.odt/.docx are supported). To force specific
   input format, use -f FORMAT.
   ```

6. Read .docx from stdin and specify reader

   `cabal run ./pandoc-cli -- -i - -f docx -o - -t native < test/docx/tabs.docx`

   ```
   [ Para
       [ Str "Some"
       , Space
       , Str "text"
       , Space
       , Str "separated"
       , Space
       , Str "by"
       , Space
       , Str "a"
       , Space
       , Str "tab."
       ]
   , Para [ Str "Tab-indented" , Space , Str "text." ]
   ]
   ```

Thank you for the invaluable tool.


jgm:
Thanks for this!  However, on thinking about it some more, I think I've come up with a better/simpler solution.
Stand by...

jgm:
See my TODO comment added in 73b5908825fe60de50848d67f8ca390be58207c2 for the proper place to put the zip check (changed my mind on this).

Also, in 1bf2564b5cd363c0b5c92b73a6912de98782a883 I added code that will give a more informative error if someone tries to read an .ods.  Still probably worth checking for the zip signature in the place indicated, but not as urgent now -- if you think it's a good idea, go ahead and modify your PR.

Note: I'd prefer a more concise error message without the extra supplement for "zip".

reptee:
I gave it some thought too and think it is too late to check for binary signatures, because if supported file is detected (ie. .docx), then it would make sense to propagate upwards and use appropriate reader, but again, if we detect binary signatures, we should detect text formats too and it should be detected in the `convertWithOpts'`

https://github.com/jgm/pandoc/blob/1bf2564b5cd363c0b5c92b73a6912de98782a883/src/Text/Pandoc/App.hs#L152-L162

and not after the decision on whether or not format is textual.

I may overthink things

reptee:
Just ran pandoc piping docx to it and it oomed, so the signature check may come in handy :sweat_smile: 

But handling the binary signature when already sent to `inputToText` still feels wrong

Wrapping up, I can't decide whether it is a good idea, but if you think this is appropriate place for checking signatures, I am happy to (try to) implement that

jgm:
My thinking is that `inputToText` is the only place we currently read any of the input; it's nice to have that localized.  We could "take a sip" of the input at the beginning of the whole process just to check for a zip signature, but that adds an unnecessary IO access to every use of pandoc, just to deal with a case that is going to be pretty infrequent.

If I were starting over at this point, I'd be tempted to remove the behavior that defaults to markdown as input format, and require this to be specified explicitly for files without a `.md` extension.  But that would be a big breaking change.

reptee:
I simplified the error message and added a few more formats that (I assume) would be of interest to convert from: DjVu, old MS/Office formats, PDF but are unsupported

I used <https://en.wikipedia.org/wiki/List_of_file_signatures> as a reference and verified they are recognized by hand with `cabal run ./pandoc-cli -- -i - < somefile`

jgm:
Great!

PR #11094, opened 2025-08-30
Add figure placement for MD to LaTeX

SeanESCA:
This PR resolves #10369 by using the `latex-pos` attribute of a figure to set its [positioning](https://www.overleaf.com/learn/latex/Positioning_images_and_tables#The_figure_environment) in LaTeX output. Additionally, the MD reader removes this attribute from the image in a figure and passes it to the figure itself. Originally, only the image's identifier is passed to the figure while all the other attributes are kept with the image.

I think that the LaTeX reader should store a figure's positioning as well, but I am not sure how to edit the reader to do this.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -5643,9 +5643,13 @@ image's alt text will be used as the caption.
     ![This is the caption](/url/of/image.png)
 
 How this is rendered depends on the output format. Some output
-formats (e.g. RTF) do not yet support figures.  In those
+formats (e.g. RTF) do not yet support figures. In those
 formats, you'll just get an image in a paragraph by itself, with
-no caption.
+no caption. For LaTeX output, you can specify a [figure's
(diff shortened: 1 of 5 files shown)


jgm:
Looks great. Only question for me is whether `latex-pos` is the best name for this.  `float-hint` could be another option -- is there something comparable for typst, I wonder?

jgm:
I will merge this now and we can think more about the name issue later.

SeanESCA:
> Looks great. Only question for me is whether `latex-pos` is the best name for this.  `float-hint` could be another option -- is there something comparable for typst, I wonder?
> 

I thought it would be a good middle ground based on the discussion. I couldn't find a name for the argument in LaTeX. Typst has a similar argument called `placement`, so we could do something similar for Typst. It is also making me think that naming the attributes something like `typst-placement` and `latex-placement`, like what you originally suggested in the discussion, is a better idea.

jgm:
I changed it to `latex-placement`. `typst-placement` may not be needed given the existing generic property support for typst.

Issue #10127, opened 2024-08-29
RST simple tables: support col spans


jgm:
```
=====  =====  ======
   Inputs     Output
------------  ------
  A      B    A or B
=====  =====  ======
False  False  False
True   False  True
False  True   True
True   True   True
=====  =====  ======
```
https://github.com/docutils/docutils/blob/master/docutils/docs/ref/rst/restructuredtext.rst#simple-tables

jgm:
Note also:

> Underlines of '-' may also be used to visually separate rows, even if there are no column spans. This is especially useful in long tables, where rows are many lines long.

PR #11115, opened 2025-09-06
RST Reader: Add col spans for simple tables

TuongNM:
This adds col span support for simple tables into the RST reader. This is only the reader part of #10127.

Added a couple of additional functions into `GridTable` that use `RowSpan` and `ColSpan` to support this.
Also adjusted `simpleDashedLines` to not count trailing whitespaces after the last dashes as this would otherwise mess with aligning the col spans with the table border.


src/Text/Pandoc/Parsing.hs
```diff
@@ -61,13 +61,19 @@ module Text.Pandoc.Parsing ( module Text.Pandoc.Sources,
...
                              gridTableWith,
                              gridTableWith',
                              TableComponents (..),
                              TableNormalization (..),
                              toTableComponents,
                              toTableComponents',
+                             toTableComponentsWithSpans,
+                             toTableComponentsWithSpans',
+                             singleRowSpans,
+                             singleColumnSpans,
```
jgm, line 76: This looks great! Just be sure to use the string `[API change]` in your commit message and indicate what has changed (this is an exported module).

jgm:
thanks!

PR #11118, opened 2025-09-07
pptx writer: Handle single column

### LINKED ISSUE #9923: Single column PowerPoint removes images and duplicate caption
A bit Unorthodox but if one uses a single column with an image in it to output to PowerPoint, the image gets removed and the caption is duplicated.

Tested with `pandoc 3.2`.

Originally reported in Quarto:

- https://github.com/quarto-dev/quarto-cli/issues/10101 

<table>
<tr><th>Input</th><th>Output</th></tr>
<tr><td>

````md
:::: {.columns}

::: {.column}
![An image](https://placehold.co/600x400.png)
:::

::::
````

</td><td>

image

</td></tr>
</table>

TuongNM:
Fixes #9923

The problem was that the existing guards only allowed for a columns Div with at least 2 column elements. Otherwise it would fall through. The columns Div would then end up in `blockToShape` where it would also fall all the way through to the point where only the paragraphs would be extracted. It would then find both the caption as well as the identical image alt text, resulting in the duplicated text.

I just added an additional guard for the single column element to be able to put into a `ContentSlide`.


jgm:
Excellent, thanks.

PR #11122, opened 2025-09-08
PandocMonad: Let fetching fail if the HTTP request is not successful.

tarleb:
HTTP requests that don't return a 200 error code are now treated as an
error. This ensures that a warning is triggered when using
`--embed-resources` or `--extract-media`.

### DIFF
--- a/src/Text/Pandoc/Class/IO.hs
+++ b/src/Text/Pandoc/Class/IO.hs
@@ -46,7 +46,7 @@ import qualified Network.TLS as TLS
 import qualified Network.TLS.Extra as TLS
 import Network.HTTP.Client
        (httpLbs, Manager, responseBody, responseHeaders,
-        Request(port, host, requestHeaders), parseRequest, newManager)
+        Request(port, host, requestHeaders), parseUrlThrow, newManager)
 import Network.HTTP.Client.Internal (addProxy)
 import Network.HTTP.Client.TLS (mkManagerSettings)
 import Network.HTTP.Types.Header ( hContentType )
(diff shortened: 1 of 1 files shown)


tarleb:
The string representation of PandocHttpErrors is rather ugly but used in the warning, so we might want to change that, too.

Cc: @cderv

jgm:
What happened before, when an error was not raised?

tarleb:
Most servers answer with a 404 page with content type `text/html`. That's then used as the image.

E.g.:

```
% echo '![test](https://pandoc.org/doesNotExist.jpg)' | pandoc --extract-media=.
<figure>
<embed src="./a1a4b7fad9765ecfdd72b534161247de9a6ea3cf.shtml" />
<figcaption aria-hidden="true">test</figcaption>
</figure>
```

where

``` sh
% head a1a4b7fad9765ecfdd72b534161247de9a6ea3cf.shtml 
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
  "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
<head>
<title>Object not found!</title>
<link rev="made" href="mailto:fiddlosopher@gmail.com" />
<style type="text/css"><!--/*--><![CDATA[/*><!--*/ 
    body { color: #000000; background-color: #FFFFFF; }
    a:link { color: #0000CC; }
```

jgm:
Let's also think about changing the warning message. Maybe we can decompose the type for HTTP error messages and produce a nicer string.

PR #11126, opened 2025-09-09
add support for reading typst pagebreak

### LINKED ISSUE #11101: Parse `#pagebreak()` in Typst
**Explain the problem.**

`pagebreak.typ`:
```typ
First Page
#pagebreak()
Second Page

```

Running: `pandoc pagebreak.typ -f typst -t pdf -s -o pagebreak.pdf`

The output is missing the page break in PDF. docx also doesn't show it but pandoc to docx [isn't supporting pagebreaks by default](https://github.com/pandoc-ext/pagebreak).

I would expect the content after the `#pagebreak()` to start on a new page.

Related typst docs: https://typst.app/docs/reference/layout/pagebreak/

Note: Typst supports two parameters:
```
pagebreak(
weak: bool, to: none | str,
) -> content 
```
However, for a first version it would probably be fine to ignore those options and give a warning if they're seen.

**Pandoc version?** MacOS and commit: `22ff5a8909dc8ce4f9d8a7dbdb97dcd657f9ebb0`

RayBB:
Closes #11101

### DIFF
--- a/src/Text/Pandoc/Readers/Typst.hs
+++ b/src/Text/Pandoc/Readers/Typst.hs
@@ -396,6 +396,7 @@ blockHandlers = M.fromList
       getField "body" fields >>= pWithContents pBlocks)
   ,("pad", \_ _ fields ->  -- ignore paddingy
       getField "body" fields >>= pWithContents pBlocks)
+  ,("pagebreak", \_ _ _ -> pure $ B.divWith ("", ["pagebreak"], []) B.horizontalRule)
   ]
 
 inlineHandlers :: PandocMonad m =>
--- a/src/Text/Pandoc/Readers/Typst.hs
+++ b/src/Text/Pandoc/Readers/Typst.hs
@@ -396,7 +396,7 @@ blockHandlers = M.fromList
       getField "body" fields >>= pWithContents pBlocks)
   ,("pad", \_ _ fields ->  -- ignore paddingy
       getField "body" fields >>= pWithContents pBlocks)
-  ,("pagebreak", \_ _ _ -> pure $ B.divWith ("", ["pagebreak"], []) B.horizontalRule)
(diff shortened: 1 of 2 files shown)


jgm:
If we are using a div here, we might consider adding `wrapper="1"`, so that it can be "unwrapped" in formats that allow putting an attribute directly on a horizontal rule.

@tarleb does that make sense?

For the class name, should it be `pagebreak` or `page-break`?

Or should we go whole hog and add PageBreak to the AST?

tarleb:
Ah right, `page-break` would be better. English orthography remains a challenge for me, even after all those years :D

I had forgotten about our `wrapper` convention. @RayBB, the LLM will probably be able to assist with that. If not, please ping me here.

tarleb:
> Or should we go whole hog and add PageBreak to the AST?

I'll probably have about a week of quiet time soon, in which I should be able to work on some of those bigger changes. I guess we'd want to bundle a couple of AST changes into a single release?

jgm:
Right, I guess in German you are used to jamming words together!

I'm not sure about the AST change.  I don't like to rely on string typing, though.  So if there are a significant number of formats that would support page breaks, and we're willing to put in the work to support them in readers and writers, it might be worth considering. AST changes are disruptive, though.

RayBB:
It seems that you both are still considering a different approach of adding this to the AST more broadly.

I'm not sure if that means you want to merge this PR in the mean time or just wait.

In any case, I renamed pagebreak to page-break and added wrapper=1 so I think your suggestions are addressed if you do want to merge this. @tarleb 

PS: If you want to squash and merge that's fine. Or I can do that locally and push it up if necessary.

tarleb:
Excellent, thanks! I think the plan is to merge this, we're just thinking out loud about possible alternatives and future changes.

Issue #9883, opened 2024-06-15
Obsidian Flavored Markdown reader/writer


kepano:
**Describe your proposed improvement and the problem it solves.**

Add a reader and writer for [Obsidian Flavored Markdown](https://help.obsidian.md/Editing+and+formatting/Obsidian+Flavored+Markdown). Obsidian Flavored Markdown combines features from CommonMark, GitHub Flavored Markdown and LaTeX, as well as several extensions inspired by wiki markup.

Syntax | Description
-- | --
`[[Link]]` | Internal links
`![[Link]]` | Embed files
`![[Link#^id]]` | Block references
`^id` | Defining a block
`%%Text%%` | Comments
`~~Text~~` | Strikethroughs
`==Text==` | Highlights
`` ``` ``  | Code blocks
`- [ ]` | Incomplete task
`- [x]` | Completed task
`> [!note]` | Callouts
(GFM syntax) | Tables

**Describe alternatives you've considered.**

Most of these extensions exist in one form or another within Pandoc, but it seems like it would be convenient to bundle them all together in a way that makes converting to/from Obsidian files easier. I'd like to explore implementing this myself, but wanted to first gauge interest.

jgm:
I think we already have all the needed extensions, except for `[^id]` for defining a block (and similarly "block references") and comments.  EDIT: and embed files.
So, if these things were implemented, one could easily package obsidian flavored markdown as a set of extensions on top of `commonmark`.

jgm:
Note that to make these extensions on top of `commonmark`, one would have to add modules to `commonmark-extensions`. This parsing work would not be done in pandoc itself, though support for the new features in the writer would have to be done in pandoc's markdown writer.

pidgeon777:
Hello, I would like to kindly inquire about any updates regarding the support for **Obsidian**'s syntax.

xieve:
I tried to use the existing extensions, but they cover even less than I had assumed they would:

```
markdown_strict+mark+wikilinks_title_after_pipe+space_in_atx_header+startnum+autolink_bare_uris+footnotes+pipe_tables+strikeout+task_lists+tex_math_dollars+yaml_metadata_block+inline_notes+backtick_code_blocks+fenced_code_attributes+all_symbols_escapable+intraword_underscores+markdown_in_html_blocks+auto_identifiers+lists_without_preceding_blankline+hard_line_breaks
```

is most notably missing `alerts` (not supported by `markdown_strict`), and

```
[...]

Issue #8853, opened 2023-05-15
Image wikilinks for export in Obsidian


anutator:
I use [Obsidian](https://obsidian.md/) and [Pandoc Plugin](https://github.com/OliverBalfour/obsidian-pandoc) to export to epub, html, docx.

There are new extensions in Pandoc 3.0 for wikilinks:
"Add new extensions **wikilinks_title_after_pipe** and **wikilinks_title_before_pipe** for commonmark and markdown". 

They can't export Obdisian images with **!** to markdown image links:
`![[image1.png]]` — image wikilink in Obsidian.

I get ! and an image URL, so images are not inserted.
Besides I have two problems:
- I keep images in "assets" subfolder. I can't use option `--resource-path=assets`.
- With `--from=gfm+wikilinks_title_after_pipe` text is exported from Obsidian without colors from Obsidian theme.

I'd like to automatically export wikilinks to markdown links not only for URLs but also for images, with title from image file name if it is not specified. Please add features to your extensions and give examples.

Current alternative:
At first I convert wikilinks to markdown links from every article I want to export with another plugin [obsidian-link-converter ](https://github.com/ozntel/obsidian-link-converter). Or I can convert all wikilinks from my vault.

jankap:
Good question, I don't know either. So maybe a filter to convert wikilinks
to MD links would be recommended? Or using only MD links?

jgm:
For an obsidian-specific solution, you just need a filter that finds the sequence
```
Str "!", Link {}
```
and converts this to an Image.  That's not the easiest kind of filter to write but it's possible.

jgm:
Wikilink images now supported.

FeralFlora:
> Wikilink images now supported.
> 

Great news! Has this already been released?

jgm:
No, it will be in the next release. Probably within the next couple of weeks?

FeralFlora:
> No, it will be in the next release. Probably within the next couple of weeks?

Okay, thanks for the answer. Looking forward to it!

Demon-tk:
I am running Pandoc 3.1.8, with Obsidian.md and Pandoc plugin 0.4.1

Simply running the pandoc MD to DOCX results in an error. The path of the image files are incorrect, regardless of `--resource-path=files`; it is ignored.

What settings need to be configured to properly export to a Docx?

FeralFlora:
[...]

PR #11135, opened 2025-09-12
Obsidian reader extensions for comments and wikilink transclusions

kepano:
Add [Obsidian-Flavored Markdown](https://help.obsidian.md/obsidian-flavored-markdown) #9883 using Markdown extensions.

## Extensions

- `comments`: Parses `%%text%%` as comments (removed from output)
- `wikilink_transclusions`: Parses `![[Title]]` for [full file transclusion](https://help.obsidian.md/embeds)
- `wikilink_heading_transclusions`: Parses `![[Title#Heading]]` for [heading transclusions](https://help.obsidian.md/links#Link+to+a+block+in+a+note)
- `wikilink_block_transclusions`: Parses `![[Title#^id]]` for [block transclusions](https://help.obsidian.md/links#Link+to+a+block+in+a+note)
- `block_ids`: Parses `^id` at paragraph end, wraps paragraph in `Div` with that ID

Uses `insertIncludedFile` for transclusion.

Includes tests and manual updates.


jgm:
Transclusion should be possible. See `insertIncludedFile` from Text.Pandoc.Parsing.
And look at the RST reader's treatment of include directives for an example of its use.

kepano:
@jgm I think this is in a good place for your review now.

Some notes:

1. I revised how wikilink images (e.g. `![[image.jpg]]`) are parsed because the implementation of #8853 via 09de84fff41d4a6e8d227baeea0037531fcef9e2 seemed to be originally driven by Obsidian's syntax. In Obsidian, the `![[link]]` wikilink syntax can be used to transclude any type of file (not just images), as well as headings and blocks within Markdown files. I consolidated all of these into a new `transclusion` parser (which could alternatively be called "include" or "embed" if you prefer).
2. I ended up building on top of the Markdown reader instead of the Commonmark reader. Hopefully this gives flexibility to other apps that support Obsidian's syntax but not other Commonmark extensions.

jgm:
Can you rebase this into one commit, or a few logically coherent commits? All API changes should be clearly described and marked as such `[API change]` in the commit messages, so we won't miss them in the changelog.

kepano:
> Can you rebase this into one commit, or a few logically coherent commits? All API changes should be clearly described and marked as such `[API change]` in the commit messages, so we won't miss them in the changelog.

Done!

jgm:
Re `block_transclusions`, `heading_transclusions`: are these really transclusions?  From the docs, it looks like they just produce links.

jgm:
Does `block_ids` also handle the case where the id is on a line by itself before a block? And the case where it is meant to apply to a a line inside a block?

pidgeon777:
Hello, thanks for working on it. What about Obsidian callouts? Do you think those are doable also?

kepano:
@jgm Thanks for the update. I saw a notification from you about other requested changes on #9883 but it seems you removed that comment. Is there something else you want me to work on?

> Re `block_transclusions`, `heading_transclusions`: are these really transclusions? From the docs, it looks like they just produce links.

They are indeed transclusions, not links. The content within the source block or heading is transcluded into the document. See also [this documentation page](https://help.obsidian.md/embeds#Embed+a+note+in+another+note) about embeds. In Obsidian-Flavored Markdown adding an exclamation mark `!` before a wikilink changes it to a transclusion (AKA embed), e.g. `[[link]]` or `![[transclusion]]`, the additional modifiers `#^` and `#` provide access to the block or heading.

> Does `block_ids` also handle the case where the id is on a line by itself before a block? And the case where it is meant to apply to a a line inside a block?

I'll check and add tests for this. 

> Hello, thanks for working on it. What about Obsidian callouts? Do you think those are doable also?

Callouts are already handled by the `alerts` extension — the GitHub syntax was inspired by Obsidian.

tobiasvl:
@kepano Should transclusions perhaps add a class to the transcluded content?

Obsidian detects transcluded content and adds classes (such as "markdown-embed-content" if the embed is a heading or Markdown block), and uses that to display it differently (like a specially styled blockquote, with a link icon that leads to the embedded content), but in this implementation there's no way to differentiate regular content from transcluded content - it's all just one flat Pandoc document.

So if one were to export an Obsidian note with a transclusion in it to PDF/TeX or HTML using this extension, the information about the fact that this is a transclusion is lost, and the result can't be styled to look like the Obsidian note.

PR #11157, opened 2025-09-20
OpenDocument writer: Add missing table elements

### LINKED ISSUE #10002: Row disappeared in table header for grid table converted to ODT
### Problem: 
Grid tables from markdown to ODT is not converted correctly.
If table header contain more than one row, the last row of header would disappear,
and first row of table body is truncated and would misplace in table header.

Command:
```
pandoc test.md -f markdown+grid_tables  -t odt -o test.odt
```
Input (test.md):
```
+---------------------+-----------------------+
| Location            | Temperature 1961-1990 |
|                     | in degree Celsius     |
|                     +-------+-------+-------+
|                     | min   | mean  | max   |
+=====================+=======+=======+=======+
| Antarctica          | -89.2 | N/A   | 19.8  |
+---------------------+-------+-------+-------+
| Earth               | -89.2 | 14    | 56.7  |
+---------------------+-------+-------+-------+
```
(Copied from https://pandoc.org/chunkedhtml-demo/8.9-tables.html)

Output:
image

## System and version
**Pandoc version?**:
pandoc 3.1.8
Features: +server +lua
Scripting engine: Lua 5.4

System: Arch Linux with kernel 6.9.2

TuongNM:
Fixes #10002

I noticed that the `TableFoot` as well as the header rows of the `TableBody` were also missing so I added them as well.

Also applied hlint suggestions that came up.

### DIFF
--- a/src/Text/Pandoc/Writers/OpenDocument.hs
+++ b/src/Text/Pandoc/Writers/OpenDocument.hs
@@ -21,7 +21,7 @@ import Data.Char (chr)
 import Data.Foldable (find)
 import Data.List (sortOn, sortBy, foldl')
 import qualified Data.Map as Map
-import Data.Ord (comparing)
+import Data.Ord (comparing, Down (Down))
 import qualified Data.Set as Set
 import Data.Text (Text)
 import qualified Data.Text as T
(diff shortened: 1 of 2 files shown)


jgm:
thanks!

PR #11173, opened 2025-09-28
RST writer: Add col spans for simple tables

### LINKED ISSUE #10127: RST simple tables: support col spans

TuongNM:
Together with the reader part from https://github.com/jgm/pandoc/pull/11115 this writer part fixes #10127

### DIFF
--- a/src/Text/Pandoc/Writers/RST.hs
+++ b/src/Text/Pandoc/Writers/RST.hs
@@ -343,7 +343,7 @@ blockToRST (Table _attrs blkCapt specs thead tbody tfoot) = do
       renderGrid = gridTable opts blocksToDoc specs' thead tbody tfoot
       isSimple = all (== 0) widths && length widths > 1
       renderSimple = do
-        tbl' <- simpleTable opts blocksToDoc headers rows
+        tbl' <- simpleTable opts blocksToDoc thead tbody tfoot
         if offset tbl' > writerColumns opts
           then renderGrid
           else return tbl'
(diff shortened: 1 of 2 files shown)


src/Text/Pandoc/Writers/RST.hs
```diff
@@ -918,25 +918,72 @@ imageDimsToRST attr = do
...
-  let numChars = maybe 0 maximum . NE.nonEmpty . map offset
-  let colWidths = map numChars $ transpose (headerDocs : rowDocs)
-  let toRow = mconcat . intersperse (lblock 1 " ") . zipWith lblock colWidths
+                   else fixEmpties <$> mapM rowToDoc headers
+  rowDocs <- fixEmpties <$> mapM rowToDoc ((concatMap tableBodyToRows tbody) ++ footers)
+  let numChars = maybe 0 maximum . NE.nonEmpty . map (offset . fst)
+  let colWidths = map numChars $ transpose (headerDocs ++ rowDocs)
   let hline = nowrap $ hsep (map (\n -> literal (T.replicate n "=")) colWidths)
-  let hdr = if all null headers
+  let hdr = if null headers
```
jgm, line 933: Why the change from `all null` to `null`?  This could make a difference.

src/Text/Pandoc/Writers/RST.hs
```diff
@@ -918,25 +918,72 @@ imageDimsToRST attr = do
...
+    tableBodyToRows (TableBody _ _ headerRows bodyRows) = headerRows ++ bodyRows
+
+    applyColSpan col@(ColSpan colSpan) doc
+      | colSpan > 1 =
+          -- Fill up columns for the col spans by adding empty docs without a ColSpan.
+          let emptyDoc = (literal "", Nothing)
+          in  (doc, Just col) : replicate (colSpan - 1) emptyDoc
+      | otherwise = [(doc, Just col)]
+
+    mapToRow colWidths docsWithColSpans = vcat $ concatMap (toRow colWidths) docsWithColSpans
```
jgm, line 961: Could be written point free: omit `docsWithColSpans` and change `$` to `.`.

src/Text/Pandoc/Writers/RST.hs
jgm: Still, maybe we need to also check for the case where the Rows aren't empty but they contain only empty cells?  That was the case we were catching before.

jgm: I'm not absolutely sure it's necessary, but I worry that it will be a regression.

TuongNM: Gotcha. I'm gonna try to adjust my code to ensure that this check will still hold.

jgm:
thanks!

PR #11233, opened 2025-10-23
fix(templates/default.typst): wrong syntax for bibliography inclusion

mcanouil:
`#bibliography(...)` should be an array when multiple bibliography files are provided (https://typst.app/docs/reference/model/bibliography/#parameters-sources).

Currently, `#bibliography("path/one.bib", "path/two.bib")` is produced instead of `#bibliography(("path/one.bib", "path/two.bib"))`

### DIFF
--- a/data/templates/default.typst
+++ b/data/templates/default.typst
@@ -149,7 +149,7 @@ $elseif(bibliographystyle)$
 $endif$
 $if(bibliography)$
 
-#bibliography($for(bibliography)$"$bibliography$"$sep$,$endfor$$if(full-bibliography)$, full: true$endif$)
+#bibliography(($for(bibliography)$"$bibliography$"$sep$,$endfor$)$if(full-bibliography)$, full: true$endif$)
 $endif$
 $endif$
 $for(include-after)$


data/templates/default.typst
```diff
@@ -149,7 +149,7 @@ $elseif(bibliographystyle)$
 $endif$
 $if(bibliography)$
 
-#bibliography($for(bibliography)$"$bibliography$"$sep$,$endfor$$if(full-bibliography)$, full: true$endif$)
+#bibliography(($for(bibliography)$"$bibliography$"$sep$,$endfor$)$if(full-bibliography)$, full: true$endif$)
```
jgm: Is this still going to work for the case where there is just one entry?
Don't you need `("mybib",)` rather than `("mybib")` for it to be treated as an array?

mcanouil: Double checking this.

mcanouil:
For reference; I've checked the syntax produced with:

```sh
echo "@book{article-1,\n  title = {Article A}\n}" > references.bib
echo "@book{article-2,\n  title = {Article B}\n}" > references-1.bib
echo "@article-1 @article-2" | pandoc -f markdown -t typst \
  --bibliography references.bib \
  --bibliography references-1.bib \
  --template default.typst
```

Issue #11261, opened 2025-10-30
Make --sandbox default behavior


eoinwm-cisa:
Due to the potential for SSRF and similar attacks (CVE-2022-35583, CVE-2025-51591), consider making `--sandbox` default behavior, and allow users to unset the option, e.g., `--no-sandbox` (or `sandbox: false`).
Another option could be for those repackaging pandoc to add a defaults file including `sandbox: true`.

As noted in the documentation:

> Anyone using pandoc on untrusted user input should use this option.

Related: #10682 #8874 #11200

amanion-cisa:
CC @dw4rren.

jgm:
No, we're not going to do that.

Pandoc is designed as a CLI tool, and we optimize for that usage.  If someone wants to have their web server shell out to this tool, they need to take special care not to shoot themselves in the foot. Our own clear recommendation for this use is to use `--sandbox`.

Issue #10682, opened 2025-03-12
SSRF When Generating PDFs from User-Controlled HTML


RealestName:
Hello, 

During a recent engagement, I came across an "Export to PDF" function which takes user controlled HTML and passes it to the pandoc PDF generator. While testing, I discovered that the PDF generation library is vulnerable to Server-Side Request Forgery (SSRF) when rendering PDFs from HTML containing <iframe> elements. An attacker can exploit this by embedding an iframe with a URL pointing to internal resources, potentially exposing sensitive data or interacting with internal services.

tarleb:
I don't see that as a security issue with pandoc, but we could consider adding an item for that to the "Security" section in the manual. Would you want to submit a PR?

tarleb:
Also note that you can prevent pandoc from resolving iframe elements by enabling the `raw_html` option.

RealestName:
> I don't see that as a security issue with pandoc, but we could consider adding an item for that to the "Security" section in the manual. Would you want to submit a PR?

I think it would be best to add an item to the security section, if possible.

jgm:
Could you describe the problem in more detail?  You have an HTML file, with an iframe whose `src` attribute points to something internal.  If you view this in a browser, the internal thing will show up in the iframe, now?  And so if you ask pandoc to produce a PDF, the PDF will also include the internal thing?

So, what is the risk you envision?  If you use pandoc to convert an HTML file containing private information, the private information will end up in the resulting file.  But that is true whether or not the information is included via an iframe.

tarleb:
I think the issue opens up when offering HTML-to-PDF-conversions as a service: A server in an internal company network might leak information that's only available from within the network.

RealestName:
> Could you describe the problem in more detail? You have an HTML file, with an iframe whose `src` attribute points to something internal. If you view this in a browser, the internal thing will show up in the iframe, now? And so if you ask pandoc to produce a PDF, the PDF will also include the internal thing?
> 
> So, what is the risk you envision? If you use pandoc to convert an HTML file containing private information, the private information will end up in the resulting file. But that is true whether or not the information is included via an iframe.

The key risk is that an attacker can exploit this behavior to force the server to make requests to internal resources, this is known as Server-Side Request Forgery (SSRF).

Let's consider the following attack scenario:

Suppose an attacker submits or uploads an HTML file with an <iframe> like this:

<iframe src="http://localhost/admin" />

[...]

Issue #8874, opened 2023-05-28
SSRF Vulnerability in PANDOC


0x1717:
- I found a critical ssrf vulnerability in pandoc software and I can hack any website on the internet that uses pandoc.
- Check out the article I wrote for more information.
- https://github.com/0x1717/ssrf-via-img

jgm:
This is #3 in "Note on security."  As noted there, we advise anyone using pandoc on untrusted input to use the `--sandbox` option, which will solve this problem.

PR #11262, opened 2025-10-30
Security notes update

eoinwm-cisa:
We did some analysis regarding SSRF vulnerabilities (CVE-2025-51591, CVE-2022-35583) and came up with suggested changes for the documentation.

Related: #11261 #10682 #8874 #11200


amanion-cisa:
CC @dw4rren.

jgm:
It would be good to know if any of the other via-HTML pdf-engines are subject to the same problem as wkhtmltopdf. That would allow more useful guidance.

yanntrividic:
> It would be good to know if any of the other via-HTML pdf-engines are subject to the same problem as wkhtmltopdf. That would allow more useful guidance.

WeasyPrint, when correctly configured, is not subject to the SSRF vulnerability (https://github.com/Kozea/WeasyPrint/issues/1288, ping @liZe).

I don't know if is the case for pagedjs-cli though, as I have not found anything on the repos about this. Maybe @fchasen or @julientaq know?

liZe:
WeasyPrint does not support iframes at all, so it should be safe with this one.

zmanion:
In our testing, `pagedjs-cli` is vulnerable similarly to `wkhtmltopdf`:
```
pandoc -o output.pdf input.html -f html+raw_html --pdf-engine pagedjs-cli
pandoc -o output.pdf input.md --pdf-engine pagedjs-cli
```

`prince` is not vulnerable by default but can be made vulnerable using the `--iframes` option:
```
pandoc -f html+raw_html --pdf-engine prince --pdf-engine-opt=--iframes -o out.pdf input.html
```

Edit: [CVE-2018-19858](https://www.cve.org/CVERecord?id=CVE-2018-19858) looks like the PrinceXML analog to [CVE-2022-35583](https://www.cve.org/CVERecord?id=CVE-2022-35583) for wkhtmltopdf.

input.html and input.md contain something like:
```
<html><iframe src=http://internal-webserver:8000/sensitive-data.html /></html>
```

MANUAL.txt
```diff
@@ -1,7 +1,7 @@
 ---
 title: Pandoc User's Guide
 author: John MacFarlane
-date: 2026-03-19
+date: 2025-10-30
```
jgm, line 4: do not change this; I will change before a release

MANUAL.txt
```diff
@@ -1763,9 +1746,10 @@ Nonzero exit codes have the following meanings:
 # Defaults files
 
 The `--defaults` option may be used to specify a package
-of options, in the form of a YAML or JSON file. Examples in
-this section will be given in YAML, but the equivalent forms in
-JSON will also work.
+of options, in the form of a YAML file.
+
+pandoc does not include a defaults file. The tables below are examples
+and do not document default configuration options.
```
jgm, line 1753: This is an extraneous change and shouldn't be part of this PR.

MANUAL.txt
jgm: No reason to change the heading

jgm:
Since this PR seems to have been abandoned, I've made my own update, taking into account your observations here.  Thank you.

eoinwm-cisa:
Thanks for the update, we appreciate it and agree with the changes in your new commit.

PR #11265, opened 2025-11-01
LaTeX reader: fix spurious paragraph breaks in math environments

### LINKED ISSUE #11257: Extra paragraph break before `\end{equation}` in LaTeX reader
**Explain the problem.**

In the LaTeX reader, if an `\end{equation}` (or `\end{align}`, etc.) is preceded by a space, then a spurious paragraph break `\n \n` is inserted in the display math environment, which completely breaks the output.

This is likely caused by a combination of https://github.com/jgm/pandoc/commit/b263f383040618ff991912512ccd396b83d43da5, and this line https://github.com/jgm/pandoc/blob/d353e1d11cb582706d9e2e00fa2216cc77d1f175/src/Text/Pandoc/Readers/LaTeX/Math.hs#L85 not actually stripping the final whitespace because the inner math content ends in a space and not a newline.

To reproduce:

Command: `pandoc -f latex -t latex`

Input:

```
\begin{equation}
  a
 \end{equation}
```

(Note the space before `\end{equation}`)

Output:

```
\begin{equation}
  a
 
\end{equation}
```

Expected output: no paragraph break.

**Pandoc version?**
I can reproduce it on https://pandoc.org/try/.

emmanuel-ferdman:
## PR Summary
When trailing spaces appeared before `\end{equation}` or `\end{align}`, pandoc incorrectly inserted blank lines in the output, breaking the math environment. This occurred because `mathEnv` used `stripTrailingNewlines` which only removed newlines but left spaces. The fix changes to `trimr` which removes all trailing whitespace (spaces, tabs, and newlines).

Example input:
```latex
\begin{equation}
 a
 \end{equation}
```
Outpt before (broken):
```latex
\begin{equation}
 a

\end{equation}
```
Spurious blank line creates paragraph break
Output after (fixed):
```latex
\begin{equation}
 a
\end{equation}
```

Note: The added tests may be a bit too specific, but I included them for completeness. Let me know if we should drop them.

Fixes #11257.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX/Math.hs
+++ b/src/Text/Pandoc/Readers/LaTeX/Math.hs
@@ -21,7 +21,7 @@ import qualified Data.Sequence as Seq
 import Text.Pandoc.Readers.LaTeX.Parsing
 import Text.Pandoc.TeX
 import Text.Pandoc.Class
-import Text.Pandoc.Shared (trimMath, stripTrailingNewlines)
+import Text.Pandoc.Shared (trimMath, trimr)
 import Text.Pandoc.Parsing hiding (blankline, mathDisplay, mathInline,
                             optional, space, spaces, withRaw, (<|>))
 import Control.Applicative ((<|>), optional)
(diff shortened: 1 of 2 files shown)


jgm:
Looks good, thanks!

Issue #8199, opened 2022-07-25
Add to asciidoc writer support for tables with merged cells


trianon1983:
When i convert from html to asciidoc page with tables with merged cells lose their merging.
It is necessary to add support for colspan and rovspan in assidoc writer.
Thanks

trianon1983:
![image](https://user-images.githubusercontent.com/109978316/180831138-a88c02e1-b4a2-45e0-b0e3-bc175384959c.png)

jgm:
Can you give the output you'd expect for this input?

trianon1983:
yes, sure. in this example i am expecting the following
![image](https://user-images.githubusercontent.com/109978316/180943012-e044c68a-0fc5-473c-9fc2-0d918fb1118d.png)

trianon1983:
full example:
[example.zip](https://github.com/jgm/pandoc/files/9187495/example.zip)

trianon1983:
note that the alignment in the cells also does not work correctly

Issue #7326, opened 2021-05-28
Table with span column and span row are not transform correctly from rst to asciidoc


kappa-wingman:
Greetings, here's the information.

I)
pandoc from GitHub:
pandoc 2.13
Compiled with pandoc-types 1.22, texmath 0.12.2, skylighting 0.10.5,
citeproc 0.3.0.9, ipynb 0.1.0.1

II) Source rst code:
```
+------------+------------+-----------+-----------------------------------------+
| Header 1   | Header 2   | Header 3  | Header 4                                |
+============+============+===========+=========================================+
| body row 1 | column 2   | column 3  | column 4                                |
+------------+------------+-----------+-----------------------------------------+
| body row 2 | Cells may span columns.| fff                                     |
+------------+------------+-----------+-----------------------------------------+
| body row 3 | Cells may  | - Cells   | - This is a very long line of text      |
|            | span rows. | - can     | - Python_                               |
+------------+            | - contain | - b                                     |
| body row 4 |            | - blocks. | - c                                     |
+------------+------------+-----------+-----------------------------------------+

.. _Python: http://www.python.org/
```
This is what it looks in HTML
![image](https://user-images.githubusercontent.com/64772920/120025423-2d636400-bfe0-11eb-965b-e8a2b995d4fb.png)

IV) conversion to AsciiDoc
$ pandoc table.rst -t asciidoctor -o table.adoc

```
[width="100%",cols="17%,16%,15%,52%",options="header",]
|===
|Header 1 |Header 2 |Header 3 |Header 4
|body row 1 |column 2 |column 3 |column 4
|body row 2 |Cells may sp |an columns. |fff
|body row 3 |Cells may span rows. a|
* Cells
* can

a|
* This is a very long line of text
* http://www.python.org/[Python]

|===

+------------+ | - contain | - b | | body row 4 | | - blocks. | - c |
+------------+------------+-----------+-----------------------------------------+
```

This is what it looks like in HTML
![image](https://user-images.githubusercontent.com/64772920/120025613-7adfd100-bfe0-11eb-82f9-6a70415cef2c.png)

tarleb:
For colspans and rowspans in rst, see also #6991.

jgm:
Partly addressed by #8202, which handles the RST reader side of this.
The asciidoc writer, however, still doesn't handle the more complex table features.

ryan-carpenter:
Corrected AsciiDoc for the example table.

```asciidoc
[...]

Issue #8665, opened 2023-03-03
Vertical bar (|) character is not properly escaped in AsciiDoc table when converting from DocBook


pOH7:
**Explain the problem.**

When converting a DocBook file containing a table with a vertical bar (|) character in its content to AsciiDoc using Pandoc, the vertical bar character is not properly escaped in the output AsciiDoc table.
**How to reproduce:**
Run the following command using Pandoc 3.1 on any operating system:
```
pandoc -f docbook -t asciidoc  demo.xml -o demo.adoc
```
Where demo.xml is the following sample DocBook file:
```
<?xml version="1.0" encoding="UTF-8"?>
<article>
    <section>
        <informaltable frame="all" rowsep="1" colsep="1">
            <tgroup cols="1">
                <thead>
                    <row>
                        <entry align="left" valign="top">h1</entry>
                        <entry align="left" valign="top">h2</entry>
                    </row>
                </thead>
                <tbody>
                    <row>
                        <entry align="left" valign="top">
                            <simpara>!@#$%^&amp;*(){}|~?+-',."&lt;&gt;[]\`</simpara>
                        </entry>
                        <entry align="left" valign="top">
                            <simpara>col 2</simpara>
                        </entry>
                    </row>
                </tbody>
            </tgroup>
        </informaltable>
    </section>
</article>
```
**Actual output:**
```
[cols=",",options="header",]
|===
|h1 |h2
|!@#$%^&*()\{}|~?+-',."<>[]\` |col 2
|===
```
**Expected output:**
```
[cols=",",options="header",]
|===
|h1 |h2
|!@#$%^&*()\{}\|~?+-',."<>[]\` |col 2
|===
```

**Pandoc version?**

pandoc 3.1 (https://pandoc.org/try/)

PR #11267, opened 2025-11-02
asciidoc writer: Add more table features

TuongNM:
This adds the following table features:
- row span and column span
- footer row
- individual horizontal cell alignment
    - had to adjust the test for #8665 this since the ascidoc writer is now writing out the alignment operators from the source table

In addition to handling row spans in headers and footer as described in https://github.com/jgm/pandoc/issues/7326#issuecomment-3402888147 I also had to handle row spans over empty rows. You can't really write an empty row in asciidoc but a row span that would go into an empty row still has to be handled. Otherwise the row span wouldn't add up.
I made those row spans basically consume empty rows. Whenever there are row spans that would go over an empty row, that empty row is removed and the row span is decremented. This corrects the row span count in the asciidoc output and handles empty rows.
Any empty rows that remain get filled up with empty columns.

T.P.Writers.Shared: Add functions `allRowsEmpty` and `tableBodiesToRows` from the RST writer for reuse in the asciidoc writer.

I got hlint warnings about `NonEmpty` and `unzip`: Currently `NonEmpty` exports a polymorphic `unzip` function but that will become monomorphic in base-4.22 (https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-List-NonEmpty.html#v:unzip). To avoid any hassles in the future I restricted the `NonEmpty` imports so that the Prelude `unzip` function is used.

Fixes: #7326
Fixes: #8199


jgm:
This is great!  Two suggestions:

- Mark the commit with `[API change]` on the new Shared exports.
- Could we use HTML instead of native for the source tables in the new command test, just to cut down its size?

TuongNM:
I added the missing `[API change]` into the commit message (one of these days I will remember to do it beforehand).

I replaced most of the source tables with HTML. I kept the test for cell aligments native due to the HTML reader inferring default column alignments that I don't think fit a test about individual cell alignments.

I also kept the test for empty rows native due to the HTML reader proactively adding empty cells rather than keeping the row empty. 

This HTML table:
```
<table>
  <tr>
    <td>Body 1</td>
  </tr>
  <tr></tr>
  <tr>
    <td>Body 2</td>
  </tr>
</table>
```

gets parsed as:
```
[ Table
    ( "" , [] , [] )
    (Caption Nothing [])
    [ ( AlignDefault , ColWidthDefault ) ]
    (TableHead ( "" , [] , [] ) [])
    [ TableBody
        ( "" , [] , [] )
        (RowHeadColumns 0)
        []
        [ Row
            ( "" , [] , [] )
            [ Cell
                ( "" , [] , [] )
                AlignDefault
                (RowSpan 1)
                (ColSpan 1)
                [ Plain [ Str "Body" , Space , Str "1" ] ]
            ]
        , Row
            ( "" , [] , [] )
            [ Cell
                ( "" , [] , [] ) AlignDefault (RowSpan 1) (ColSpan 1) []
            ]
        , Row
            ( "" , [] , [] )
            [ Cell
                ( "" , [] , [] )
                AlignDefault
                (RowSpan 1)
                (ColSpan 1)
                [ Plain [ Str "Body" , Space , Str "2" ] ]
            ]
        ]
    ]
    (TableFoot ( "" , [] , [] ) [])
]
```

Is this a bug? I don't think it's correct that an empty row is parsed as a row with an empty cell.

TuongNM:
I also noticed that the HTML writer is not preserving the empty row.

This native table:
```
[ Table
    ( "" , [] , [] )
    (Caption Nothing [])
    [ ( AlignDefault , ColWidthDefault )
    ]
    (TableHead ( "" , [] , [] ) [])
    [ TableBody
        ( "" , [] , [] )
        (RowHeadColumns 0)
        []
        [ Row
            ( "" , [] , [] )
            [ Cell
                ( "" , [] , [] )
                AlignDefault
                (RowSpan 1)
                (ColSpan 1)
                [ Plain [ Str "Body" ] ]
            ]
        , Row ( "" , [] , [] ) []
        ]
    ]
    (TableFoot ( "" , [] , [] ) [])
]
```

is converted to this HTML table:
```
<table>
  <tbody>
    <tr>
      <td>Body</td>
    </tr>
    <tr>
      <td></td>
    </tr>
  </tbody>
</table>
```

I also don't think that this is right either. The empty row should be preserved, right?

jgm:
I agree that these things seem like bugs, but there might be a reason for them that I've either forgotten or never knew (because someone else wrote that code).
Feel free to open new issues for them!

Issue #10743, opened 2025-03-31
TWiki reader dies with "unexpected '9' or '\n'"


chicks-net:
**Explain the problem.**

Base file is https://github.com/fini-net/www-fini-net/blob/main/content/CCexhibitA.txt

I run `pandoc -f twiki -t commonmark CCexhibitA.txt`

I get

```
Error at "CCexhibitA.txt_chunk_chunk" (line 3, column 1):
unexpected '9' or '\n'
expecting end of input or inline
```

All of the other files in this directory converted with this command.

* Folding the lines to be shorter and removing random 9's made no difference.  
* Removing the head matter does not matter.
* Removing the 9 heading does not help.  Removing those lines at the top does not effect the line noted in the error message.
* `--verbose` did not change the output
* `--log` did not produce a log (separate bug?)

**Pandoc version?**

```
% pandoc --version
pandoc 3.6.4
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: /Users/chicks/.local/share/pandoc
Copyright (C) 2006-2024 John MacFarlane. Web: https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.
```

Installed on MacOS with homebrew.

jgm:
By bisection I determined that the line in question is `96 Mowat Avenue` in
```
<blockquote><strong>TUCOWS Inc.

96 Mowat Avenue

Toronto, Ontario

M6K 3M1

Attention:&#65533; Legal Affairs</strong></blockquote>
```
I have nothing to do with the twiki reader, but I suspect the issue is that the parser for `<strong>` elements is expecting inline content without paragraph breaks.  Removing these blank lines in the input should solve it.

`--log` only logs warnings and info messages. In this case you didn't get any of these, just an error which stops execution.

chicks-net:
Removing the `<strong>` tags fixed it.  https://github.com/fini-net/www-fini-net/pull/3

Thanks for finding the 9 that was causing the problem.

I'm going to leave this open in case this is actually a bug you want to fix, but my problem is 💯 % solved.

PR #11282, opened 2025-11-08
TWiki reader: Fix handling of blank lines in TWiki HTML inline spans

InvalidPathException:
Fixes https://github.com/jgm/pandoc/issues/10743

Issue: TWiki reader treats `<strong>/<em>/<b>` elements as plain inline streams, so embedded blank lines triggers parse errors.

Fix: added `parseHtmlInlineContentWithAttrs` that converts blank lines to spaces during inline parsing and switched the bold/emphasis HTML handlers to use it

Testing: included test cases for different elements with multiple lines

### DIFF
--- a/src/Text/Pandoc/Readers/TWiki.hs
+++ b/src/Text/Pandoc/Readers/TWiki.hs
@@ -79,6 +79,20 @@ parseHtmlContentWithAttrs tag parser = do
     parseContent = parseFromString' $ manyTill parser endOfContent
     endOfContent = try $ skipMany blankline >> skipSpaces >> eof
 
+parseHtmlInlineContentWithAttrs :: PandocMonad m
+                                => Text -> TWParser m B.Inlines
+                                -> TWParser m (Attr, [B.Inlines])
+parseHtmlInlineContentWithAttrs tag inlineParser = do
+  (attr, content) <- htmlElement tag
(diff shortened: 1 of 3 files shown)


jgm:
> Fix: added parseHtmlInlineContentWithAttrs that converts blank lines to spaces during inline parsing and switched the bold/emphasis HTML handlers to use it

I'm not sure this is the right fix, as it will change the semantics; in TWiki, if I understand correctly, you'd get multiple paragraphs here, right?  Not just text separated by spaces?

InvalidPathException:
> I'm not sure this is the right fix, as it will change the semantics; in TWiki, if I understand correctly, you'd get multiple paragraphs here, right? Not just text separated by spaces?

https://twiki.org/cgi-bin/view/TWiki/%20TextFormattingRules?skin=plain#Using_HTML_CSS_and_JavaScript
You are right about this, the original example would lead to invalid HTML for TWiki anyways, so it should also be invalid here:
> TWiki inserts `<p />` paragraph tags on empty lines, which causes problems if done between HTML tags that do not allow paragraph tags … Do not span a tag over more than one line

I encountered this issue in a previous internship when I had to convert many things to Markdown, now I'm not so sure what led to that invalid input in the first place. I will close this for now.

PR #11287, opened 2025-11-09
added xlsx reader

aantich:
- Reads multiple spreadsheets
- Converts tables to tables
- Basic tests added

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -282,6 +282,7 @@ header when requesting a document from a URL:
     - `twiki` ([TWiki markup])
     - `typst` ([typst])
     - `vimwiki` ([Vimwiki])
+    - `xlsx` ([Excel spreadsheet][XLSX])
     - `xml` (XML version of native AST)
     - the path of a custom Lua reader, see [Custom readers and writers] below
     :::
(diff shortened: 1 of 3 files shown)


jgm:
I think the test files may need to be included in extra-source-files in pandoc.cabal.

jgm:
The new format should be added to MANUAL.txt ; see the list under `--readers`.

jgm:
One thing I noticed testing this locally:  I got a table with hundreds of empty rows.
Generally when one starts a spreadsheet in Excel, it already comes with lots of rows, only some of which get used.  It would probably make sense to strip all trailing empty rows before parsing.

aantich:
> One thing I noticed testing this locally: I got a table with hundreds of empty rows. Generally when one starts a spreadsheet in Excel, it already comes with lots of rows, only some of which get used. It would probably make sense to strip all trailing empty rows before parsing.

We'll fix the trailing empty rows; for empty rows in the middle of the sheet I believe it makes sense to keep them, since I may want to reference the exact cell coordinates eventually, e.g. a table like this: 
image

is converted to markdown like this:

## Main {#sheet-1}

  **Person**           **Age**   **Location**
  -------------------- --------- --------------
  Anton Antich         23.0      Switzerland
  James Bond           35.0      Moscow
                                 
                                 
  Just a random cell   
  
  I believe this is the correct behavior.

aantich:
Believe we addressed all comments. Didnt test empty rows rigorously, but on a couple of quick files it works.

jgm:
I'm making these small changes and merging this manually. Thanks!

Issue #11281, opened 2025-11-08
H1-H3 Headers Inside Blockquotes Generate Invalid LaTeX (Incomplete Fix from 2014)


manwithacat:
# Description

Headers at levels 1-3 (H1, H2, H3) inside blockquotes generate invalid LaTeX that fails to compile. This appears to be an incomplete fix from issue #1221 (2014), which only addressed H4-H6 headers.

**Note:** Headers in blockquotes are valid CommonMark syntax (see [CommonMark Spec v0.31.2, Example 228](https://spec.commonmark.org/0.31.2/#example-228)). Pandoc correctly parses this syntax but generates invalid LaTeX output.

## Environment

- **Pandoc version:** 3.8.2.1
- **LaTeX engine:** XeTeX 3.141592653-2.6-0.999997 (TeX Live 2025)
- **Operating system:** macOS (Darwin 25.2.0)

### Test Case 1: H1 in Blockquote (FAILS)

**Input Markdown:**
```markdown
# Test Document

> # H1 Header in Blockquote
>
> This should work but fails to compile.
```

**Command:**
```bash
pandoc test-h1.md -o test-h1.pdf
```

**Generated LaTeX:**
```latex
\begin{quote}
\section{H1 Header in Blockquote}\label{h1-header-in-blockquote}

This should work but fails to compile.
\end{quote}
```

**Error:**
```
Error producing PDF.
! LaTeX Error: Something's wrong--perhaps a missing \item.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...

l.60 \end{quote}
```

### Test Case 2: H4 in Blockquote (WORKS)

**Input Markdown:**
```markdown
# Test Document

> #### H4 Header in Blockquote
>
> This works correctly.
```

**Generated LaTeX:**
```latex
\begin{quote}
\mbox{}%
\paragraph{H4 Header in Blockquote}\label{h4-header-in-blockquote}

This works correctly.
\end{quote}
```

**Result:** ✅ Compiles successfully

## Systematic Testing Results

I tested all header levels (H1-H6) inside blockquotes:

| Header Level | LaTeX Command | Inside `\begin{quote}` | Result |
|--------------|---------------|------------------------|--------|
| H1 (#)       | `\section`    | ❌ Invalid LaTeX       | FAILS  |
| H2 (##)      | `\subsection` | ❌ Invalid LaTeX       | FAILS  |
[...]

PR #11288, opened 2025-11-10
Modify prefix logic for quoted LaTeX elements

manwithacat:
Adjust prefix for quoted paragraphs and subparagraphs in LaTeX.

Ref [Issue 11281: H1-H3 Headers Inside Blockquotes Generate Invalid LaTeX](https://github.com/jgm/pandoc/issues/11281)

> Have you tried testing LaTeX with the mbox before a \section command inside a block quote? If this works properly, I see no reason not to implement this simple change. It should just be a simple matter of changing https://github.com/jgm/pandoc/blob/main/src/Text/Pandoc/Writers/LaTeX.hs#L835-L839

### DIFF
--- a/src/Text/Pandoc/Writers/LaTeX.hs
+++ b/src/Text/Pandoc/Writers/LaTeX.hs
@@ -832,7 +832,7 @@ sectionHeader classes ident level lst = do
                           5  -> "subparagraph"
                           _  -> ""
   inQuote <- gets stInQuote
-  let prefix = if inQuote && level' >= 4
+  let prefix = if inQuote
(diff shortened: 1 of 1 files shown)


jgm:
Just to confirm, did you confirm that LaTeX behaves as expected with pandoc's output after this change?

manwithacat:
The pandoc output I get is valid and it can then be ingested upstream by the next stage in my process. So “it works on my machine.” This is an interesting problem in terms of validating it. I could probably come up with some ideas on another CI step: pull in the commonmark spec and turn it  into a set of test cases.

Issue #10149, opened 2024-09-03
ANSI writer ignores RowSpan in table cells


silby:
As of #9565 the table renderer in the ANSI writer doesn't account for table cells with a row span. For example:

`a.html`:

```html
<table>
	<tbody>
		<tr><td colspan="2">a</td><td rowspan="2">c</td><td>d</td></tr>
		<tr><td>a</td><td>b</td><td>d</td></tr>
		<tr><td>a</td><td>b</td><td>c</td><td>d</td></tr>
	</tbody>
</table>
```

Command and output:

```
♪ pandoc$ cabal exec -- pandoc --columns 12 -w ansi a.html
── ── ── ──
a     c  d
a  b  d
a  b  c  d
── ── ── ──
Table 1
```

Should be:

```
── ── ── ──
a     c  d
a  b     d
a  b  c  d
── ── ── ──
Table 1
```

PR #11294, opened 2025-11-17
ANSI writer: Apply row spans in tables

TuongNM:
I reused code from the asciidoc writer which has been extracted into `T.P.Writers.Shared` as the functions `insertCurrentSpansAtColumn`, `takePreviousSpansAtColumn` and `decrementTrailingRowSpans`. With these the ANSI writer is now able to keep track of row spans and apply them in rows.

Fixes: #10149


silby:
Nicely done!

PR #11296, opened 2025-11-17
Docx reader: Handle REF link instruction

Ezwal:
This PR aims to handle a common run field instruction (`fieldInstr`) from docx format :  `REF`  [microsoft documentation](https://support.microsoft.com/en-us/office/field-codes-ref-field-b2531c23-05d6-4e3b-b54f-aee24447ceb2), specifically those with the "link" switch `\h`.

In word software, you can create `REF` field instruction with the *Cross-reference* button. You can create cross-reference to many things such as Equation, Table, Title... 

I mostly re-use existing code structure so it may not be very DRY.

Thank you !


jgm:
thanks!

Issue #10052, opened 2024-08-03
Add scroll: view option to revealjs template


memeplex:
**Describe your proposed improvement and the problem it solves.**

https://revealjs.com/scroll-view/

It provides a vertical document view for the entire slideshow.

Besides providing an alternative nice-looking navigation style, it may help with some previewers that reset the position of the slideshow after each refresh.

**Describe alternatives you've considered.**

Creating a custom template.

memeplex:
A nice demo: https://revealjs.com/demo/?view=scroll

jgm:
Seems reasonable. Want to make a PR?

memeplex:
Yes, no problem, it will take me a few weeks though because I'm quite busy
right now. It's just modifying the templates, isn't it?

tarleb:
@memeplex Kind reminder about this issue. Do you think you'll have time to work on this within the next weeks?

abdivasiyev:
Hello.

I am going to get this issue. Can I get?

abdivasiyev:
Hello. I saw this issue was created last year and till now has not been resolved. So, I decided to get and add this feature.

rschmehl:
Hello, thanks for implementing this feature! Since release 5.0.0 Reveal.js automatically enables scroll view for viewports below 435px wide. A user can disable this behavior by setting the scrollActivationWidth config value to null or 0 (See [breaking change](https://github.com/hakimel/reveal.js/releases/tag/5.0.0). 
The way you have now implemented the feature in the template, scrollActivationWidth can only be set when the config value view is set at the same time. This coupling is not optimal, in my opinion, as it does not allow users to independently disable scroll view in situation where it was automatically enabled, without explicitly setting view=scroll. My solution
```
$if(view)$
      ...
$endif$
$if(scrollActivationWidth)$
        // see https://revealjs.com/scroll-view/#url-activation
        scrollActivationWidth: '$scrollActivationWidth$',
$endif$
```

jgm:
@rschmehl want to submit a PR?
data/templates/default.revealjs and perhaps some test outputs need to be changed

PR #11305, opened 2025-11-28
feat: add reveal.js scroll view option to HTML writer

abdivasiyev:
Added `view` option with its dependent options to control scroll view.

To enable scroll view, user must set `view: scroll`.

Here is available options:

- scrollProgress (defaults `true`)
- scrollActivationWidth (defaults `'0'`)
- scrollLayout (defaults `'full'`)

Fixes https://github.com/jgm/pandoc/issues/10052


jgm:
Thanks!

PR #11306, opened 2025-11-28
Update output option description for pandoc usage

jidanni:
Just like below this you say -s or --standalone.

By the way, sometimes the page says flags, sometimes it says options. Maybe pick one name and stick with it.


jgm:
For consistency with the rest, should be
```
`-o`/`--output`
```

jgm:
I went ahead and made this change.

Issue #7536, opened 2021-09-01
pptx: Reference doc with no slides causes no slides in output


undergroundquizscene:
# Summary

If you provide a reference doc that contains no slides, the resulting pptx displays as containing no slides, regardless of the input markdown. The slides' xml files are present in the output archive, but PowerPoint is not aware of them.

This is because [modifySldIdLst](https://github.com/jgm/pandoc/blob/0617d3a88bb1f84622b8a94bd3366c7c8c9e8f8a/src/Text/Pandoc/Writers/Powerpoint/Output.hs#L1847) searches for an existing `sldIdLst` element in `presentation.xml` to replace it, but if you delete all slides from the reference doc the `sldIdLst` element is also removed.

Pandoc version 2.14.2, macOS

# To Reproduce

- Open a copy of the default reference in PowerPoint
- Delete all slides (but not the layouts)
- Use the resulting doc as reference doc in a call to pandoc, e.g. `pandoc --reference-doc reference-empty.pptx -o slides.pptx input.md`

jgm:
Do you want to create a PR for this too?

undergroundquizscene:
I haven't got to it yet, but I think I should be able to get to it today

undergroundquizscene:
I made the change I suggested, but unfortunately it doesn't seem to have completely fixed the problem: though the slides are now listed in the `presentation.xml`, PowerPoint still says the pptx file needs fixing. I'll take another look when I can, but I will have less time to work on pandoc from today on.

PR #11310, opened 2025-11-30
pptx writer: Handle reference doc without slides

TuongNM:
An empty `sldIdLst` is now added if the reference doc is missing one so that `modifySldIdLst` can replace it. To ensure PowerPoint doesn't say that the file will need fixing, the `sldIdLst` has to be placed after the `sldMasterIdLst`. I also added a test to ensure that if there are notes, they will be placed between the `sldMasterIdLst` and `sldIdLst`. Otherwise PowerPoint wouldn't show the slide of a note when viewing Notes Pages.

Since I'm reusing existing logic, I combined some of it into shared functions to avoid code duplication.

Fixes: #7536


jgm:
Thanks!

PR #11317, opened 2025-12-02
Typst template: Fix keywords usage.

har7an:
Previously the template turned any `keywords` defined in Pandoc metadata into empty objects due to a misnomer in a loop variable. This in turn was rejected by the Typst compiler as invalid syntax, making document generation with keywords impossible.

### DIFF
--- a/data/templates/default.typst
+++ b/data/templates/default.typst
@@ -65,7 +65,7 @@ $endfor$
     ),
 $endif$
 $if(keywords)$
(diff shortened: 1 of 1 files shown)


har7an:
I know the contributing guide asks to write tests, but I'm not familiar with this repository and I don't see any tests for other language templates. Do you think this change requires a test and if so, where should I add it?

jgm:
Thanks! I don't think we need a test for this, which is clearly correct.

PR #11328, opened 2025-12-07
Update MANUAL.txt with Markdown extensions info

jidanni:
Without my change, users will not realize they have been reading the default extensions section until they finally encounter the non-default section!

Please also hyperlink my see also target if possible.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -3595,8 +3595,8 @@ Note that Markdown extensions added to the `ipynb` format
 affect Markdown cells in Jupyter notebooks (as do command-line
 options like `--markdown-headings`).
 
-The following extensions are enabled by defaut. (See also 
-[Non-default extensions] below.)
+The following extensions are enabled by default. (See also 
+[Non-default extensions].)
 
 ## Typography
 


MANUAL.txt
```diff
@@ -3595,6 +3595,9 @@ Note that Markdown extensions added to the `ipynb` format
 affect Markdown cells in Jupyter notebooks (as do command-line
 options like `--markdown-headings`).
 
+The following extensions are enabled by defaut. (See also 
+"Non-default extensions" below.)
```
jgm: `[Non-default extensions]` should give you the link you seek.

jidanni:
Thanks for the hyperlink advice.
Alas, last time I tried to edit a commit it crashed the entire github website for everybody. So perhaps you could fix it for me and merge it into "main". Thanks.

jgm:
Actually this is more complex than I thought. "the following" is too vague since even the non-default ones follow.  We probably ought to have a section "Default Markdown extensions," but this would require other changes that are too involved for me right now.  So, closing.

PR #11335, opened 2025-12-10
HTML writer: support parenthesis list delimeters

### LINKED ISSUE #11326: No single parenthesis
Pandoc man page says:
> Thus, the following yields a list with numbers **followed by a single parenthesis**,
>>               9)  Ninth
>>              10)  Tenth
>>              11)  Eleventh

Sorry, there is no single parenthesis:
```
$ for i in 1 2 3; do echo $i\) $i; done
1) 1
2) 2
3) 3
$ for i in 1 2 3; do echo $i\) $i; done | pandoc
<ol type="1">
<li>1</li>
<li>2</li>
<li>3</li>
</ol>
```
No I didn't forget to turn on an extension. Same problem as `1)` for `(1)`.

Anyway you'll need pretty fancy HTML5 to get those parenthesis rendered, so maybe on the man page mention the parenthesis don't work in HTML output mode.
```
$ pandoc --version
pandoc 3.1.11.1
```

silby:
Current browsers support defining custom @counter-styles to be used as list-style-type values for lists. This change adds rules to the default HTML styles, as needed, to support lists using the OneParen or TwoParens style. The type="1" (etc.) attribute is retained, which means the browser should do the next-best thing if the @counter-style rule goes missing in somebody's workflow. The custom @counter-styles will not be used in markup written without a template. ("Is there a template at all" doesn't actualy answer the question "will the styles.html partial emit the @counter-styles like we expect" but we resort to this sort of thing in other places.)

This change unconditionally, and arguably redundantly, adds `style="list-style-type: decimal"` (or "lower-alpha", etc.) to `<ol>` elements where it didn't appear before, including in non-standalone output.

That last bit is somewhat unaesthetic (cf. the changed tests) but the code to eliminate the problem would've been conditional-heavy.

Fixes #11326.

### DIFF
--- a/test/command/11335.md
+++ b/test/command/11335.md
@@ -1,0 +1,109 @@
+Without fancy_lists, we get the Default counter styles
+
+```
+% pandoc -t native -f markdown_strict
+1. A
+2. B
+
+(a) A
+(b) B
+^D
+[ OrderedList
+    ( 1 , DefaultStyle , DefaultDelim )
+    [ [ Plain [ Str "A" ] ] , [ Plain [ Str "B" ] ] ]
+, Para
+    [ Str "(a)"
+    , Space
+    , Str "A"
+    , SoftBreak
+    , Str "(b)"
+    , Space
+    , Str "B"
+    ]
+]
+```
+
+and thus as HTML we don't have `style` or `type` in the `<ol>`:
+
+```
+% pandoc -t html -f markdown_strict
+1. A
+2. B
+
+(a) A
+(b) B
+^D
+<ol>
+<li>A</li>
+<li>B</li>
+</ol>
+<p>(a) A (b) B</p>
+```
+
+With fancy_lists, we get Decimal and Period instead of Default.
+
+```
+% pandoc -t native -f markdown_strict+fancy_lists
+1. A
+2. B
+
+(a) A
+(b) B
+^D
+[ OrderedList
+    ( 1 , Decimal , Period )
+    [ [ Plain [ Str "A" ] ] , [ Plain [ Str "B" ] ] ]
+, OrderedList
+    ( 1 , LowerAlpha , TwoParens )
+    [ [ Plain [ Str "A" ] ] , [ Plain [ Str "B" ] ] ]
+]
+```
+
+With commonmark, we get Decimal and Period even without fancy_lists.
+
+```
+% pandoc -t native -f commonmark
+1. A
+2. B
+
+(a) A
+(b) B
+^D
+[ OrderedList
+    ( 1 , Decimal , Period )
+    [ [ Plain [ Str "A" ] ] , [ Plain [ Str "B" ] ] ]
+, Para
+    [ Str "(a)"
+    , Space
+    , Str "A"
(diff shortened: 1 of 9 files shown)


silby:
Ooh I've learned from the test matrix that older GHC can't tell when previous case analysis eliminated certain possibilities, neat. I'll make it compatible.

jgm:
Very slick!  Here's the one thought I have. We have a distinction between DefaultStyle and Decimal, which gets effaced by your code.  Would it be better to make counterStyle a Maybe, so that the explicit indication of a "decimal" style is left out of the ol tag when we have DefaultStyle?

silby:
That's a good point on handling Default more appropriately, I'll work on that.

silby:
The phenomenon with the Default vs. Decimal counter styles is preexisting. I added a [command test](https://github.com/jgm/pandoc/blob/cfa2c8b092146788bb858e9e6ec2b545cd285cdc/test/command/11335.md) with some demonstrations of how things stand.

silby:
Now passing CI on GHC 8.10. Somewhat distressingly I have to conditionally-compile the catchall branch in the offending case statement because it warns for redundancy (and thus would fail CI due to -Werror) in later GHCs.

You dropped GHC 8.6 and 8.8 at 6182769307a5dc99cfd0e95fadd2a131d7cc8a02 2.5 years ago and the last release of GHC 8.10 was a couple months after that with 8.10.7. No idea what your usual approach for dropping major versions is but this is a data point.

jgm:
CONTRIBUTING.md says that we aim to support at least the last 3 ghc versions, and often more (currently back to 8.10).

So, I think it would be safe to drop ghc 8.10 support at this point. Indeed, we might even consider dropping everything less than 9.6.  (9.6 is the version in Debian stable, which I always try to support; it's usually the oldest version in any Linux distro.)  @tarleb what are your thoughts?

tarleb:
Even Debian oldstable has GHC >= 9, so dropping 8.10 seems like a safe choice. And I don't mind making GHC 9.6 the minimum version.

jgm:
OK, I've dropped support for ghc < 9.6.  Hopefully this allows us to streamline some things!

silby:
Ok cool. I’m out of town til Tuesday so you could cherry-pick the main commit from this PR without the compatibility fix or if you have further feedback let me know and I’ll address it next when I’m back.

silby:
rebased and pushed with the Haskell 8.10 support dropped. Any further feedback on this?

test/command/11335.md
```diff
@@ -0,0 +1,109 @@
...
+    , Str "(b)"
+    , Space
+    , Str "B"
+    ]
+]
+```
+
+With fancy_lists and parens in the markdown source, we don't try to use a
+custom counter-style as the list-style-type in HTML output when no template is
+used.
```
jgm: This is unusual -- the usual expectation is that the output you get without `-s` is the same you'll get with `-s` (save for the template).  Do we need this feature ?  What happens if we specify the custom list style type and it isn't defined in styles ?

Issue #11300, opened 2025-11-23
Docbook chapter title overwrites book title


nicorikken:
# Explain the problem.

When creating a file from a DocBook file, Pandoc doesn't distinguish between the title of a chapter (or similar like appendix, colophon, etc.) and the title of the main book.

## Failing

Docbook modeled as is typical, with the `title` in the main `info` element on the top level. This `info` element is at the top of the file.

### failing.dbk

```xml
<?xml version="1.0"?>
<book xmlns="http://docbook.org/ns/docbook" version="5.0" dir="ltr">
  <info>
    <title>Book title</title>
    <subtitle>Book subtitle</subtitle>
  </info>
  <chapter>
    <info>
      <title>Chapter title</title>
    </info>
    <para>My sentence</para>
  </chapter>
</book>
```

### Command

```
pandoc -f docbook -o failing.odt failing.dbk
```

### Output

<img width="886" height="365" alt="Image" src="https://github.com/user-attachments/assets/e8ac3cd2-ced4-4ef8-ad81-e6e5a132b2e8" />

## Succesful

Note the book info is now below the chapter.

### succeeding.dbk

```xml
<?xml version="1.0"?>
<book xmlns="http://docbook.org/ns/docbook" version="5.0" dir="ltr">
  <chapter>
    <info>
      <title>Chapter title</title>
    </info>
    <para>My sentence</para>
  </chapter>
  <info>
    <title>Book title</title>
    <subtitle>Book subtitle</subtitle>
  </info>
</book>
```

### Command

```
pandoc -f docbook -o succeeding.odt succeeding.dbk
```

### Output

<img width="886" height="365" alt="Image" src="https://github.com/user-attachments/assets/1d46af93-d9f8-486e-8479-4d0b42db87ef" />

# Pandoc version?

Verified this occurs with the lastest version of Pandoc 3.8.2.1, installed from the .deb file from GitHub.

PR #11347, opened 2025-12-13
DocBook reader: Fix adding wrong metadata

TuongNM:
Now keep track of the current element stack to only add metadata if inside an appropriate parent element.

Fixes: #11300


jgm:
Great! Thanks.

PR #11358, opened 2025-12-25
fix typo about --chunk-template

AlbertLei:
The manual about `--chunk-template=PATHTEMPLATE` says:

> ... `%section-%s-%i.html` might  be resolved to `section-1.1-introduction.html`

I believe the beginning `%` in `%section-%s-%i.html` should be removed.


jgm:
Thanks!

Issue #8783, opened 2023-04-15
Conversion of bullet list from RTF to Markdown or Commonmark does not work properly


jamesemoody:
Note this problem is very similar in nature to #8767 but involves numbered & unnumbered lists, as opposed to tables.

**Explain the problem:**
I have attached BulletLists.rtf, which contains some regular text, a numbered list, more text, an unnumbered list, and more text again. It was produced by exporting from MacOS Pages to RTF.

I execute this command line:
`pandoc --to commonmark ./BulletLists.rtf -o ./BulletLists.md`

Expected result:

The .md file should contain some lines of regular text, and then a Markdown-formatted list, more regular text, another Markdown-formatted list, and more regular text.

Actual result:

The .md file does correctly contain Markdown-formatted lists (both numbered and unnumbered). BUT the final bullet seems to additionally contain what should be the regular text following the table. I'll include the .md file inline below to demonstrate, since it's small.

I get a similar result if I use `--to markdown`, so I'm suspecting this is a problem parsing/processing the RTF input, as opposed to a problem in the Commonmark/MD generators.

**Pandoc version**
```
$ pandoc --version
pandoc 3.1.2
Features: +server +lua
Scripting engine: Lua 5.4
```
Running on Ubuntu 22.04.2 LTS.

The resulting markdown looks like this:

```
This is preamble text.  
This is another line of preamble text.

1.  The first numbered bullet point

2.  The second numbered bullet point  
    This is text after the numbered list.  
    This is more text after the numbered list but before the unnumbered
    list.

- This is the first unnumbered bullet point

- This is the second unnumbered bullet point.  
  And finally some more text outside of the bullets.
```

[BulletLists.zip](https://github.com/jgm/pandoc/files/11240001/BulletLists.zip)

PR #11360, opened 2025-12-26
RTF reader: Ensure new paragraph on \pard

TuongNM:
New paragraphs may start with \pard alone without an explicit paragraph break with \par first.

Fixes: #8783


TuongNM:
Further testing showed that this attempted fix causes other issues where other paragraphs aren't properly parsed anymore.

PR #11361, opened 2025-12-27
RTF reader: Ensure new paragraph on \pard

TuongNM:
Fixed from previous PR with better test case.

New paragraphs may start with \pard alone without an explicit paragraph break with \par preceding it.

Fixes: #8783


jgm:
thanks!

Issue #10942, opened 2025-06-30
Conversion of table from RTF to Docx or html doesn work when there is an embedded complex  hyperlink with  bookmark  in the field 


danieltomasz:
The RTF is based on the content of rich text field  queried from the database. I am using latest pandoc 3.7.

I cannot attach RTF  which  cause me trouble but I created a new repository wih rtf file and output html
https://github.com/danieltomasz/PandocTest

The table in rtf looks normal ( But it  have an  embedded hyperlinks for some of  the cells)

![Image](https://github.com/user-attachments/assets/8a2f5f21-857f-4af4-a29c-9ee3943a9c3e)
But the output  from `pandoc problem.rtf -s  -f rtf -t HTML --wrap=none -o problem.html` is lacking one of the fields where hyperlink has bookmark

![Image](https://github.com/user-attachments/assets/f531e7aa-a6e2-420d-8722-54ac0e573279)

Workaround:

I can use LibreOffice `soffice --headless --convert-to docx  problem.rtf`  to correctly convert it.

Expected behaviour, 

So learned that the pandoc implementantion of "RTF" is  not complete, I understand limitation, but may in case of such problem would be possible to put the text "above" the hiperlinks, instead of providing empty cells? In comment below, more minimal example with hypelinks and bookmark markers which cause pandoc to choke

Below an MVE (problematic RTF) 
```
{\rtf1\ansi\deff0
{\fonttbl{\f0\froman Times New Roman;}}
{\colortbl;\red0\green0\blue0;}
\deftab720

\trowd\trleft0\cellx2000\cellx4000
\pard\intbl
{{\field{\*\fldinst HYPERLINK "https://example.com" }{\fldrslt {\hich\af0\loch\hich\af0\loch\cf0\f0\cf0\f0\loch
{\*\bkmkstart _dx_frag_StartFragment}{\*\bkmkend _dx_frag_StartFragment}Problem Text}}}}\cell
\pard\intbl Normal Text\cell
\row

\trowd\trleft0\cellx2000\cellx4000
\pard\intbl Simple Text\cell
\pard\intbl More Text\cell
\row
}
```
And resulting HTML (with doesnt include problem cell)
```

<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" lang="" xml:lang="">
<head>
  <meta charset="utf-8" />
  <meta name="generator" content="pandoc" />
[...]

PR #11370, opened 2026-01-02
RTF reader: Improve hyperlink parsing more

TuongNM:
Both the field instruction and its result may be ungrouped.

Fixes: #10942


jgm:
Excellent!

Issue #10859, opened 2025-05-21
Add the ability to modify the resource path from lua


CodeSandwich:
I need to dynamically load markdown files referencing resources relatively. E.g. `docs/03/doc.md` references `docs/03/MEDIA/image.png` as `![](MEDIA/image.png)`. If I run `pandoc --resource-path=docs/03 docs/03/doc.md ...`, they are found just fine, but if I load the markdown file dynamically, with `pandoc.read` there doesn't seem to be any way to perform `pandoc.mediabag.fetch` that would include `docs/03` as the search path unless I declare this path with `--resource-path` when starting pandoc. There's the `PANDOC_STATE.resource_path` list, but it's read-only, modifying it has no effect.

**Describe your proposed improvement and the problem it solves.**

Being able to modify `PANDOC_STATE.resource_path` inside the filter could be the right solution.

**Describe alternatives you've considered.**

I'm not aware of any alternatives, this use case is simply unsupported.

tarleb:
My first attempt at this can be found in the [pandoc.init](https://github.com/jgm/pandoc/compare/main...tarleb:pandoc:pandoc.init) branch. The approach here is to allow initializing the common state in Lua, but only if it hasn't been initialized before, as to prevent too much confusion.  The approach mostly works, but I stopped working on it because it's a very fragile solution and not very flexible.

However, I still like the idea of keeping the common state read-only after it's been initialized

My preferred solution would therefore be to support the creation of new Lua-threads, which could be initialized with any `PANDOC_STATE`. That would allow to run certain parts of the code with an arbitrary state, and it could open up a path towards multi-threading, which some heavy Lua users have been asking about.

Ideally we'd use Lua coroutines, as to get an idiomatic interface, but those are a bit tricky to work with.

CodeSandwich:
Now that I'm thinking about it, my use case could be solved by just passing the extra resource paths to `pandoc.mediabag.fetch`. It would only affect the specific call to `fetch` and it wouldn't need to alter the common state. I'm not aware of uses of `PANDOC_STATE.resource_path` other than in the mediabag filling, are there any?

tarleb:
The resource path currently used during citeproc processing and in the Roff and LaTeX parsers, plus every time that the Haskell function `fetchItem` is called, which happens in a lot of writers and in the RST reader.

CodeSandwich:
[...]

PR #11371, opened 2026-01-02
Lua: add function `pandoc.with_state`

tarleb:
The function allows to run a callback with a modified pandoc state. This
provides the ability to temporarily modify the resource path, the user
data directory, and the HTTP request headers.

Closes: #10859

### DIFF
--- a/doc/lua-filters.md
+++ b/doc/lua-filters.md
@@ -3890,6 +3890,33 @@ Parameters:
 
 Returns: the transformed inline element
 
+### `with_state (options, callback)` {#pandoc.with_state}
+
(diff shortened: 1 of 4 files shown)


kainctl:
Hey, this is great! Thanks! 
One question though, why silently ignore errors?
I tried this feature and it didn't seem to work at first. Only later did I discover that I typed `resource_paths` instead of `resource_path`.

tarleb:
Honest answer: because checking for invalid options is more work and didn't strike me as necessary at first. We can add it though, please raise an issue so we won't forget.

Issue #11378, opened 2026-01-04
Pandoc DOCX: adds invalid /word/media/ Override in [Content_Types].xml when using reference doc (Word reports corruption)


Harryoung:
## Summary
  When converting Markdown to DOCX with a reference document, pandoc adds an invalid `<Override>` entry for a directory (`/word/media/`) in `[Content_Types].xml`. This violates OPC rules and makes Microsoft Word report corruption.

  ## Environment
  - Pandoc: 3.7.0.2
  - OS: macOS (Darwin 25.1.0)

  ## Steps to Reproduce
  1. Use the attached reference doc: `pandoc_template.docx`
  2. Use the provided markdown: `test.md`
  3. Run:
     ```bash
     pandoc test.md --reference-doc=pandoc_template.docx -o test.docx

  4. Open test.docx in Microsoft Word.

  ## Expected

  Word opens the generated DOCX without errors.

  ## Actual

  Word shows: “Word found unreadable content in test.docx. Do you want to recover the contents of this document?”

  ## Investigation

  Unzipping test.docx and inspecting [Content_Types].xml shows:

  <Override PartName="/word/media/" ContentType="application/octet-stream" />

  This is invalid because part names must not end with / (OPC/ECMA-376).

  The reference template’s [Content_Types].xml does not contain that entry.

  ## Suspected Cause

  When building overrides from the reference doc, pandoc seems to include the word/media/ directory entry itself.

  ## Workaround

  Remove the invalid Override from [Content_Types].xml after generation.

  ## Attachments

  - pandoc_template.docx
  - test.md
  - test.docx

[pandoc_template.docx](https://github.com/user-attachments/files/24419659/pandoc_template.docx)
[test.docx](https://github.com/user-attachments/files/24419660/test.docx)
[test.md](https://github.com/user-attachments/files/24419658/test.md)

PR #11379, opened 2026-01-04
Fix docx writer: skip directory entries when building media overrides

Harryoung:
## Problem
  Pandoc docx writer adds an `<Override>` for `/word/media/` in `[Content_Types].xml` when the reference doc contains media, which violates OPC rules and causes Word to report corruption.

  ## Fix
  Filter out directory entries when building media overrides so only real files produce `Override` nodes.

  ## Tests
  - Added a minimal regression test to ensure `[Content_Types].xml` does not contain `/word/media/` override when using a reference doc with media.

  ## Related Issue
  - https://github.com/jgm/pandoc/issues/11378


jgm:
Thanks!

PR #11394, opened 2026-01-10
feat(typst.hs): Add alt text support for images in Typst writer (accessibility)

mcanouil:
Introduce support for alt text in images within the Typst writer, enhancing accessibility.
Behaviour is by design consistent with LaTeX and HTML writers.

~~TO-DO: Check if the following files need to account for alt as well~~

- [x] ~~`test/command/9945.md` this might now get an alternative-text not expected at the time.~~

---

```bash
echo '![Caption "some title"](image.png){alt="Alt text describing the image"}' | pandoc -f markdown -t typst
echo '![](image.png){alt="Alt text describing the image"}' | pandoc -f markdown -t typst
```

---

This was identified in Quarto CLI that the alt-text was not forwarded.

Resolves https://github.com/quarto-dev/quarto-cli/issues/13868

PR #11395, opened 2026-01-10
Typst writer: add image alt text support

gordonwoodhull:
Simple fix implemented by Claude.

Since there is no corresponding issue, here are the details:

- Add `alt` parameter to Typst `image()` calls for accessibility (PDF/UA compliance)
- Follows same pattern as LaTeX writer: checks for explicit `alt` attribute first, falls back to image caption
- If neither is available, omits `alt`

## Background

Typst supports PDF/UA-1 accessibility standards via the `alt` parameter on `image()` calls. Without this, accessible PDFs generated through Typst fail validation with "missing alt text" errors.

**Before:**
```typst
#figure(image("cat.png"),
  caption: [
    A cat sleeping
  ]
)
```

**After:**
```typst
#figure(image("cat.png", alt: "A cat sleeping"),
  caption: [
    A cat sleeping
  ]
)
```

The implementation mirrors the existing LaTeX writer behavior where:
1. Explicit `{alt="..."}` attribute takes priority
2. Falls back to stringified image caption
3. Omits `alt` parameter when no alt text is available

### DIFF
--- a/src/Text/Pandoc/Writers/Typst.hs
+++ b/src/Text/Pandoc/Writers/Typst.hs
@@ -33,7 +33,7 @@ import Control.Monad.State ( StateT, evalStateT, gets, modify )
 import Text.Pandoc.Writers.Shared ( lookupMetaInlines, lookupMetaString,
                                     metaToContext, defField, resetField,
                                     setupTranslations )
-import Text.Pandoc.Shared (isTightList, orderedListMarkers, tshow)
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Oops, this is almost the same as #11394 - the only difference in output is that @mcanouil puts `alt` first after the filename, which I agree with. I'll close this and suggest a couple more tests there.

Issue #8024, opened 2022-04-17
[bug]:  `defaults` in default YAML does not interpolate `${.}`


gabyx:
**Explain the problem.**

The ``defaults:`` section does not interpolate `${.}`, which it should according to documentation.
Probably the interpolation was forgotten in these paths.

`general.yaml`:

```
defaults: ${.}/general2.yaml
```

`general2.yaml`:

```
eol: lf
```

```shell
echo "asd" | pandoc --data-dir=. --defaults=general.yaml -f markdown -t markdown
```

```
pandoc: ${.}/general2.yaml: openBinaryFile: does not exist (No such file or directory)
```

**Pandoc version?**
2.18

wlupton:
Unless it's intentional that the `${}` syntax is not supported in this context, this does indeed seem to be a bug?

svenboulanger:
This is also an issue for the `to`-field, which can be used to specify the writer.

The following defaults yml file fails to resolve the `${.}`:

```
from: markdown
to: ${.}\writer.docx.lua
```

jgm:
The way the architecture works currently, it is not trivial to add interpolation in `defaults`.  I'll need to restructure things somehow to make that work.  (Technical note: the issue is that the JSON/YAML parser doesn't have access to state.)

Adding interpolation to `to` and `from` should be straightforward; I'll do that.

PR #11432, opened 2026-01-30
Enhance variable expansion in defaults files

jkub6:
I have been using these changes successfully with my own document pipeline without issues. I believe these changes should resolve [#8024](https://github.com/jgm/pandoc/issues/8024). Putting this here in case it is found useful enough to merge.

- Factor out expansion logic into expandVars and expandEnv.
- Ensure the '.' variable correctly resolves to the directory of the current defaults file
- Expand environment variables in 'data-dir' and use the expanded path when resolving child defaults.
- Allow environment variable expansion within 'defaults' file paths.


jgm:
Thank you!  This looks great. I integrated it off the platform and pushed your commit, together with some tests I added.  So I'll close this PR.

Issue #11441, opened 2026-02-04
Describe the Lua `Source` type in the docs


tarleb:
The `Source` type is mentioned as part of, e.g., the `pandoc.read` function signature, but there is no documentation for that type.

PR #11448, opened 2026-02-06
Lua: add new function `pandoc.types.Sources`.

tarleb:
This provides access to the `Sources` abstraction that pandoc uses for input files.

Closes: #11441


tarleb:
The module comes with two functions, `pandoc.sources.new` and `pandoc.sources.files`. The former allows to create a list of sources from raw strings, while the latter takes filenames.

Feedback on the interface and naming is most welcome.

Pinging @fiapps and @igorgubin, who pointed out that something like this was missing.

jgm:
I think I've fixed the issue with the wasm ci.

fiapps:
I gather that the content of sources created with `pandoc.sources.new()` is the string(s) passed to the function, but  understand that from the PR comments and the contrast with `pandoc.sources.files()` rather than the description of `source_def`. I think the following would be better (this is the generated documentation, since I didn't look at the Haskell.

```md
`source_def`
:   String or list of strings, each becoming the content of a separate Source. (string|{string,...})
```

Otherwise it looks good from the documentation. I didn't look at the Haskell because I don't know the language.

tarleb:
Thanks a lot for the feedback!

However, I changed my mind about this, and instead there is now only a new function `pandoc.types.Sources`. This is more in line with the current module structure.

I've also given up on the idea of providing a constructor that creates sources from filenames. I/O operations should be limited to as few modules as possible, and a constructor like that would go against that principle. Something like this can be used instead:

``` lua
local function source_files(paths)
  return pandoc.types.Sources(
    pandoc.List.map(paths, function (fp)
      return {name = fp, text = pandoc.system.read_file(fp) }
    end)
  )
end
```

Issue #11452, opened 2026-02-08
Support Beamer logo graphic options


sidneymau:
**Describe your proposed improvement and the problem it solves.**

It would be convenient to be able to specify options for the included logo graphic in beamer (similar to what was done for the title graphic with #9207). Because the logo is passed directly to `\includegraphics` in the same way as the titlegraphic, the same solution could apply here (cf. https://github.com/jgm/pandoc/blob/e9ec0c509325668e3718403550d4e87676f1101b/data/templates/default.beamer#L121-L129).

**Describe alternatives you've considered.**

The only alternative I think would be to manually change the logo image (e.g., making a smaller PDF) or to use `header-includes`

sidneymau:
I think the change would be
```
$if(logo)$
\logo{\includegraphics$if(logooptions)$[$logooptions$]$endif${$logo$}}
$endif$
```
since the logo is singular, there would be no iteration. Happy to open a PR if this seems reasonable

jgm:
Sure, go ahead. Don't forget to update the portion of the manual where beamer-relevant variables are described.

Issue #10906, opened 2025-06-12
titlegraphicoptions is insufficiently documented


schanzen:
#9207 introduces titlegraphicsoptions, but the documented does not describe what kind of options can be set. It took me to open #9207 to find what syntax is expected.

jgm:
For these you need to consult the beamer manual.

sidneymau:
Just adding a clarifying point for anyone who finds their way here in the future: the options for `titlegraphicoptions` (and `logooptions` with #11456) are the same as for `\includegraphics` -- https://mirrors.mit.edu/CTAN/macros/latex/required/graphics/grfguide.pdf

PR #11456, opened 2026-02-09
Beamer template: add `logooption` variable

sidneymau:
Addresses #11452. One question that came up was whether or not to include more details in the documentation for both `titlegraphicoptions` and the new `logooptions` (in light of #10906). Specifically, these options are both being passed to an underyling `\includegraphics` call rather than the `\titlegraphic` or `\logo` commands, which can be somewhat confusing


jgm:
Perhaps the documentation issue could be addressed by adding `(e.g., width, height)`.

PR #11464, opened 2026-02-13
readers/docx: Support w:gridBefore

jtojnar:
Table rows in Docx files can use `w:gridBefore` property to specify a number of extra empty table cells to be inserted at the beginning of a row. Without this, table columns in tables using the element
will become misaligned.

https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.gridbefore?view=openxml-3.0.1

In the worst case, this can lead to cells at the end of some rows to be dropped by the parser.

The test file was produced as follows based on the “3GPP TS 24.282 V19.5.0 (2025-12)” document obtained from
<https://www.3gpp.org/ftp/Specs/archive/24_series/24.282/24282-j50.zip> where I encountered the issue:

1. Extracted the `document.xml`.
2. Formatted it with `xmllint --format`.
3. Found the raw XML for “Table 15.2.13-2: Payload content type”.
4. Removed `w:rsid*` and `w14:*` attributes, and `w:pStyle` elements.
5. Replaced the context of `test/docx/raw-blocks.native`.
6. Used pandoc to generate the docx.


jgm:
Can you describe more plainly what this is for?

jtojnar:
Table rows in Docx files can use `w:gridBefore` to specify a number of extra empty table cells to be inserted at the beginning of a row. Without this, table columns might become misaligned.

I am trying to fix Docx reader ignoring some table cells in the aforementioned document, presumably because of lack of support for `w:gridBefore`.

I will continue investigation later, and include that in the commit message once I am sure.

By the way, what am I supposed to do about link in the commit message, when the URL is longer that 78 characters? Could `check-commit-msg-length` modified to ignore overlong lines starting with `https://`?

jgm:
> Table rows in Docx files can use w:gridBefore to specify a number of extra empty table cells to be inserted at the beginning of a row. Without this, table columns might become misaligned.

I got that from the link, but I'm struggling to understand the practical application.  Can you give an example?

Don't worry about the commit message check; I'll ignore it if it goes over for this reason. Good idea about the modification.

jgm:
Btw, I've updated the commit length check to ignore lines containing only URLs.

jtojnar:
I noticed that pandoc was dropping the last column in a table using `w:gridBefore` in the only row that did not have `w:gridBefore`. You can see the affected table in the test file.

I now confirmed that adding support for `w:gridBefore` keeps all columns in the affected table.

Issue #3390, opened 2017-01-28
Inline Verbatim text isn't given a specific style in ODT output


gareththered:
If I write \`Pandoc\` inline within a markdown input file, it displays with a different font as `Pandoc` which is fine.  However, that seems to be done by simply changing the font to _Courier New_ for an ODT output file.  
The same input used to create a DOCX file attaches the _Verbatim Char_ character style to the verbatim inline text, which I can control by adjusting this style in my `custom-reference.docx` file I add to the command line options.  
Is there a way to get the ODT output to behave the same.  That is, add a character style to inline verbatim text instead of simply changing the font?  I could then control this style in a `custom-reference.odt` file.

Warmduscher:
Also the font-size of verbatim is fixed to 10pt. This gives arkward results when used inside a header or if used inline inside text blocks using a different font-size.

jgm:
Yes, the opendocument writer is currently fairly primitive.  I didn't write it, and I don't understand the details of the format well enough to change it.  But it would be nice for someone to take this on, replacing all the auto-generated styles with real styles.

jtojnar:
Verified this still happens with `pandoc 3.7.0.2`. If I create a [document](https://github.com/user-attachments/files/25318447/source-test-pandoc.odt) as follows:

```ShellSession
$ pandoc -f markdown -o source-test-pandoc.odt
# Test `foo`

Hello `world`
```

it will contain:

### `content.xml`

```xml
<text:h text:style-name="Heading_20_1" text:outline-level="1">
  <text:bookmark-start text:name="test-foo" />Test
  <text:span text:style-name="Source_Text">foo</text:span>
  <text:bookmark-end text:name="test-foo" />
</text:h>
<text:p text:style-name="First_20_paragraph">Hello
  <text:span text:style-name="Source_Text">world</text:span>
</text:p>
```

### `styles.xml`

```xml
<style:style style:family="text" style:name="Source_Text">
      <style:text-properties style:font-name="Courier New" style:font-name-asian="Courier New" style:font-name-complex="Courier New" style:font-size-asian="10pt" style:font-size-complex="10pt" fo:font-size="10pt" fo:language="zxx" />
</style:style>
```

---

Creating a [document in LibreOffice Writer](https://github.com/user-attachments/files/25318427/source-text.odt) and using the built-in _Source Text_ character style will produce:


### `content.xml`

```xml
<text:h text:style-name="Heading_20_1" text:outline-level="1">Test
[...]

Issue #9136, opened 2023-10-14
ODT writer makes one hardcoded style set for each list/entry-level found in the MD source


chrisaga:
**Pandoc dynamically generates redundant list styles which is a problem when you need some customization**
I used `pandoc -o test.odt test-odt-list.md` to generate an ODT document from a MD source containing 2 nested bullet lists and 2 nested ordered lists.

Pandoc's ODT writers generated 6 different `text:list-style` named `L1` to `L6` and 6 diferent `style:style` named `P1` to `P6`: 

* One of each for every bullet list level
  * The first bullet list has 2 level and uses the following styles: 
    * `<text:list text:style-name="L1">` for the whole list
    * `<text:p text:style-name="P1">` for the first level
    * `<text:p text:style-name="P2">` for the second level
  * The second list is a duplicate of the firs one and uses styles "L3", "P3", and "P4".
* One of each for every ordered list
  * The first ordered list uses the following styles:
    * `<text:list text:style-name="L5">` for the whole list
    * `<text:p text:style-name="P5">` for every entry, whatever the level
  * The second ordered list uses styles "L6" and "P6"

N.B. **no** custom reference.odt

**Why is it a problem ?**

1) It makes list customization very difficult to perform even with post ODT file post-processing.
2) Contrary to the ordered list, the bullet list doesn't use correctly the list hierarchy witch makes the problem worse.
3) It generate extra code in the ODT file

**Remark**

I don't see the point of genarating dynamic hardcoded list styles since Pandoc has an embeded `reference.odt`. Whouldn't it be easier to define the list styles in `reference.odt` and use them in the writer ?

I confess, I had a look to ODT.hs and OpenDocument.hs but I definitely cannot read Haskell code. So maybe there is a good reason for hardcoded styles.

**Pandoc version 3.1.8 on Ubuntu 22.04.3 LTS**
pandoc 3.1.8
Features: +server +lua
Scripting engine: Lua 5.4
User data directory: /home/chris/.local/share/pandoc
Copyright (C) 2006-2023 John MacFarlane. Web: https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.

**MD Source code**
```

# Un Titre

## Des listes

Une liste :

* bla
	- lala
	- lili
	- lolo
* blabla
* blablabla

## Un Sous-titre

Et encore des puces ?

* bla
	- lala
	- lili
	- lolo
* blabla
* blablabla

Numéros :

[...]

PR #11468, opened 2026-02-14
Odt: Improve code style handling

jtojnar:
Implements reading `CodeBlock` from ODT files and adds support for LibreOffice-standard style name for `Code`. Also synchronises the styles with current LibreOffice, including the removal of font size for `Code`, allowing to inherit the font size from paragraph/heading.

See individual commits for detailed descriptions.

Fixes: https://github.com/jgm/pandoc/issues/3390
Fixes: #4841


jgm:
Great! This looks good.  

> Fixes: https://github.com/jgm/pandoc/issues/3390
> Fixes: https://github.com/jgm/pandoc/issues/4841

Can you add `Closes #3390` and `Closes #4841` to the appropriate commits? Otherwise this information won't make it into the changelog.  Then it will be ready to go.

One minor comment: you use `_x` in a type annotation. That's legal but unidiomatic; normally we just use a lowercase letter like `a` when types are parametric.  (We'd use `_x` for a function parameter that isn't used.)  It's fine to leave this, though, if there's a reason I haven't seen.

On the use of temporary styles by pandoc (e.g. P123), see #9136.  Would be helpful to improve this some day!

jtojnar:
I used `_x` as it was used in most of the file. I added another patch that normalises it throughout the reader.

Added `Closes` tags to the commit messages.

jgm:
thanks!

Issue #11481, opened 2026-02-22
EPUB 3: update allowed values for metadata`identifier.scheme` (ONIX codelist issue 72)


Wandmalfarbe:
The EPUB 3 metadata for identifiers uses the ONIX codelist 5 scheme to indicate the product identifier type (e.g., ISBN or DOI etc.). [The pandoc supported list of `scheme` values](https://github.com/jgm/pandoc/blob/94dc2de439e864c12ccb613dc54db82034be16a6/src/Text/Pandoc/Writers/EPUB.hs#L1133) could be updated to support newer values from [ONIX codelist Issue 72 (published January 2026)](https://www.editeur.org/files/ONIX%20for%20books%20-%20code%20lists/ONIX_BookProduct_Codelists_Issue_72_Changes.pdf). The last supported value `OLCC` in the list has been added in [ONIX codelist Issue 18 (published August 2012)](https://www.editeur.org/files/ONIX%20for%20books%20-%20code%20lists/ONIX_BookProduct_Codelists_Issue_18_Changes.pdf).

``` hs
        schemeToOnix :: Text -> Text
        schemeToOnix "ISBN-10"              = "02"
        schemeToOnix "GTIN-13"              = "03"
        schemeToOnix "UPC"                  = "04"
        schemeToOnix "ISMN-10"              = "05"
        schemeToOnix "DOI"                  = "06"
        schemeToOnix "LCCN"                 = "13"
        schemeToOnix "GTIN-14"              = "14"
        schemeToOnix "ISBN-13"              = "15"
        schemeToOnix "Legal deposit number" = "17"
        schemeToOnix "URN"                  = "22"
        schemeToOnix "OCLC"                 = "23"
        schemeToOnix "ISMN-13"              = "25"
        schemeToOnix "ISBN-A"               = "26"
        schemeToOnix "JP"                   = "27"
        schemeToOnix "OLCC"                 = "28"
        schemeToOnix _                      = "01"
```

There is an HTML version available of the latest ONIX codelist Issue 72 at https://ns.editeur.org/onix/en/5. New additions to pandoc could be:

| Code  | Heading |
| ------ | -------- |
| `29` | JP Magazine ID |
| `30` | UPC-12+5 |
| `31` | BNF Control number |
| `34` | ISSN-13 |
| `35` | ARK |
| `36` | Digital file internal version number |

I also noticed that code `24` *Co-publisher’s ISBN-13* is missing from pandoc.

jgm:
[...]

PR #11485, opened 2026-02-23
EPUB 3 update allowed values for metadata `identifier.scheme`

Wandmalfarbe:
See #11481

I will format the code and add documentation as separate commits to this PR later if everything is alright.


jgm:
This looks fine. Don't worry about formatting the code to line up `=`. This just makes the diff less informative.
Adding to documentation would be good, though, if there is already a place where these values are documented.

Wandmalfarbe:
I added/changed the entries in the manual. The PR should be ready now. Unfortunately I made the commit message of the first commits too long, so the pipeline check fails ...

jgm:
thanks!

Issue #11510, opened 2026-03-09
docx -> typst silently failing to extrct media when shape present


RayBB:
**Explain the problem.**
Command: `pandoc --extract-media=. image_bug.docx -o image_bug.typ`
Input file: [image_bug.docx](https://github.com/user-attachments/files/25838841/image_bug.docx)

I would expect the output to be a `image_bug.typ` containing some text with an image  and a `media` folder with one image in it.
Unfortunately, it doesn't output the media folder or have a link to the image. It just shows some text.

I'm not sure what's going on with this Word document. I isolated the issue down the to smallest reproducible bug which was just the text and the image. If the text is removed the image seems to work fine. Also, this is an extract from a document I did not create so I cannot explain why the text is inside the shape like that. However, it's probably a case that should cause an error or be handled appropriately.

As always, thanks for taking the type to help with this!

**Pandoc version?**
pandoc 3.9 on macOS

RayBB:
Ah it even seems to be causing some text to disappear. That's not good.

[image_bug_3.docx](https://github.com/user-attachments/files/25839439/image_bug_3.docx)

jgm:
Closing as a dup of #11053.

RayBB:
@jgm in this case are you also sure it's because of the shape element that text is disappearing?

RayBB:
@jgm I opened [a PR](https://github.com/jgm/pandoc/pull/11515) to address the issue you linked above.

That said, it doesn’t seem to fix the issue I’m reporting here.

From discussing it with AI, my understanding is that there’s a fundamental difference between the two cases: the original issue involves `images embedded inside textboxes`, whereas in the case I’m reporting here, the `image is not in a textbox` — it’s a standard `<w:drawing>` element.

jgm:
OK, in this doc you have an mc:AlternateContent with two children, one mc:Choice Requires="wps", one with mc:Fallback. The current pandoc code just looks at the fallback. The problem is that the jpg image is used in the other branch only; the fallback seems to have a v:shape element with some binary data, and pandoc doesn't know what to do with that.

That's as far as I got so far.  I'm not actually sure what mc:Choice Requires="wps" means.  Perhaps pandoc should be looking at this instead of the fallback, in general.

RayBB:
That's good info.

I tried to dig in a bit further and found some relevant links:
https://wiki.openoffice.org/wiki/OOXML/Markup_Compatibility_and_Extensibility
[...]

PR #11515, opened 2026-03-09
fix: unwrap direct docx textbox images for media extraction

### LINKED ISSUE #11053: Docx --> LaTeX: --extract-media omits certain figures
**Explain the problem.**

[Example.docx](https://github.com/user-attachments/files/21800489/Example.docx)

```
pandoc Example.docx --extract-media=. --standalone -o Example.tex
```

The LaTeX file contains no reference to the figure. The media folder is not created. If other figures would generate  a media folder, the submitted figure is still missing.

I expect the figure to be extracted.

Image

When analyzing the docx file, we can find the image in the `word/media` folder. It's also listed in `word/_rels/document.xml.rels` (rId11) and referenced in `word/document.xml`.

The figure's XML structure (containing `<w:pict>` and `<v:shape>` tags) is different compared to newly inserted figures.

**Pandoc version?**

Windows 11 Pro
Pandoc 3.7.0.2

RayBB:
Closes #11053

I believe this fixes the issue stated above. At least, it appears to when I build pandoc locally.

Disclaimer: This code is AI generated and I am not a haskell expert so my understanding of it is minimal.
However, it does seem to successfully fix the example provided by the user.

If anything needs to be changed please commit directly to this branch or whatever it takes to get it merged :)


RayBB:
If this PR is correct, I think #11510 should be reopened @jgm

RayBB:
@jgm good catch. It should be fixed now.

PR #11555, opened 2026-03-27
Update bash completion options for --pdf-engine

lodwicz:
Removed 'pdfroff' from the list of options for --pdf-engine, to avoid lookup for `pdfroff.exe` win executable on cygwin, and enable native cygwin script...

The alternative `groff -Tpdf` groff engine is not enumerated, and runs perfectly well on cygwin without searching for a Win executable.

### DIFF
--- a/data/bash_completion.tpl
+++ b/data/bash_completion.tpl
@@ -34,7 +34,7 @@ _pandoc()
              return 0
              ;;
          --pdf-engine)
-             COMPREPLY=( $(compgen -W "pdflatex lualatex xelatex latexmk tectonic wkhtmltopdf weasyprint prince context pdfroff" -- ${cur}) )
+             COMPREPLY=( $(compgen -W "pdflatex lualatex xelatex latexmk tectonic wkhtmltopdf weasyprint prince context" -- ${cur}) )
              return 0
              ;;
          --print-default-data-file)


lodwicz:
it seems to me that this bash completion is redundant, since appending the `.exe` extension affects only windows executables, and windows does not support bash (for pdfroff) completion: being a shell script, there is no direct executable required on *ix, or it should be provided separately for windows, the same as `gropdf`, the script for the alternative `groff` pdf-engine, which is a perl script. Again: either (both) should be provided with bat or exe stubs to run on windows.

alerque:
What are you talking about? The code you removed isn't adding or using an exe extension in the first place.

lodwicz:
my apologies: the issue is that `pdfroff` is no longer distributed with groff 1.24.x ([discussed 2024-11-15](https://cgit.git.savannah.gnu.org/cgit/groff.git/commit/?h=1.24.x&id=b58576c18b8f696e5a7fb62722caad37c92a6752); [announced 28 feb 2026](https://lwn.net/Articles/1060786/)), so it could be removed; the executable extension appended came from older code. 

the engine supported now by default by groff is the option `-Tpdf` as currently documented.

jgm:
If we only had to support Windows, this change might make sense. But since there are other platforms that have pdfroff, it doesn't...

lodwicz:
> my apologies: the issue is that `pdfroff` is no longer distributed with groff 1.24.x ([discussed 2024-11-15](https://cgit.git.savannah.gnu.org/cgit/groff.git/commit/?h=1.24.x&id=b58576c18b8f696e5a7fb62722caad37c92a6752); [announced 28 feb 2026](https://lwn.net/Articles/1060786/)), so it could be removed; the executable extension appended came from older code.
> 
> the engine supported now by default by groff is the option `-Tpdf` as currently documented.

perhaps the change should be to _add_ the `groff` engine, which is now the supported engine by default by groff itself...

jgm:
Good idea, I'll do that.

PR #11600, opened 2026-04-26
Update instructions for modifying reference.docx

jidanni:
Mention open source alternatives too.


jgm:
Maybe we should leave it as "Word or LibreOffice" -- since we know those will work.  I really don't know whether there are other open-source word processors that can handle docx and handle the styles in a way that works with pandoc.

jidanni:
All I care about is like when we need to update "if a man" to "if a person" to not be exclusive.

Anyway please edit it to your taste. If I tried to edit an edit I would botch it.

PR #11606, opened 2026-05-01
remove css warning on weasyprint

### LINKED ISSUE #11524: Default inline HTML CSS emits avoidable WeasyPrint warnings (while syntax-highlighted PDF output is otherwise fine)
### Description

When generating PDF via HTML with `--pdf-engine=weasyprint`, pandoc emits CSS declarations in its built-in inline styles that trigger avoidable WeasyPrint warnings.

The PDF still renders, and syntax-highlighted code blocks are useful and should be preserved. So this is not a request to drop pandoc's inline CSS entirely. The issue seems to be a small subset of browser-oriented declarations in pandoc's default HTML CSS / syntax-highlighting CSS.

Observed warnings:

```text
WARNING: Ignored `gap: min(4vw, 1.5em)` at 7:32, invalid value.
WARNING: Ignored `overflow-x: auto` at 8:28, unknown property.
WARNING: Ignored `user-select: none` at 46:32, unknown property.
```

### Repro

A minimal repro is:

```bash
pandoc input.md \
  --from gfm \
  --standalone \
  --toc \
  --number-sections \
  --section-divs \
  --pdf-engine=weasyprint \
  -o out.pdf
```

I reproduced this locally with:

- pandoc 3.9
- WeasyPrint 68.1

### Notes

The warnings appear to come from pandoc's generated inline CSS, not from my document stylesheet.

The relevant declarations seem to be:

- default HTML CSS:
  - `div.columns { display: flex; gap: min(4vw, 1.5em); }`
  - `div.column { flex: auto; overflow-x: auto; }`
  - `table { ... overflow-x: auto; display: block; ... }`
- syntax-highlighting / numbered source line CSS:
  - `user-select: none` on the line-number pseudo-element

### Expected behavior

When targeting `weasyprint`, pandoc should ideally emit default CSS that is compatible with WeasyPrint, while preserving syntax highlighting and numbered source lines.

### Suggested direction

A targeted fix would be better than removing the whole inline style block:

- keep syntax-highlighting CSS
- keep numbered source lines
- omit or rewrite only the WeasyPrint-incompatible declarations when `--pdf-engine=weasyprint`

For example:

- avoid `gap: min(...)` in the default columns CSS for WeasyPrint output
- avoid `overflow-x: auto` in the default table / column CSS for WeasyPrint output
- omit `user-select` vendor rules in the syntax-highlighting CSS for WeasyPrint output

This seems especially relevant now that WeasyPrint is getting more attention in pandoc's PDF workflow.

### References

- Pandoc default HTML CSS template: <https://github.com/jgm/pandoc-templates/blob/master/styles.html>
- Related issue: <https://github.com/jgm/pandoc/issues/10142>

kbrock:
Issue
-----
Styles unconditionally emits css that uses screen-only properties. Paged-media engines (weasyprint, prince, pagedjs) have no viewport and complain. Fix it to hide these properties from the engines.

Before
------
Building any document with `pandoc --pdf-engine=weasyprint` prints:

    WARNING: Ignored `gap: min(4vw, 1.5em)` at 6:32, invalid value.
    WARNING: Ignored `overflow-x: auto` at 7:28, unknown property.

After
-----
Same two rules, wrapped in `@media screen { ... }`. HTML browsers render identically. Print engines skip the block, no warnings.

Fixes: jgm/pandoc#11524

### DIFF
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -185,8 +185,12 @@ header {
 }
 $endif$
 span.smallcaps{font-variant: small-caps;}
-div.columns{display: flex; gap: min(4vw, 1.5em);}
-div.column{flex: auto; overflow-x: auto;}
+div.columns{display: flex; gap: 1.5em;}
+div.column{flex: auto;}
+@media screen {
+div.columns{gap: min(4vw, 1.5em);}
+div.column{overflow-x: auto;}
+}
 div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
 /* The extra [class] is a hack that increases specificity enough to
    override a similar rule in reveal.js */
--- a/test/s5-basic.html
+++ b/test/s5-basic.html
@@ -15,8 +15,12 @@
     ** See https://pandoc.org/MANUAL.html#variables-for-html for config info.
     */
     span.smallcaps{font-variant: small-caps;}
-    div.columns{display: flex; gap: min(4vw, 1.5em);}
-    div.column{flex: auto; overflow-x: auto;}
+    div.columns{display: flex; gap: 1.5em;}
+    div.column{flex: auto;}
+    @media screen {
+    div.columns{gap: min(4vw, 1.5em);}
+    div.column{overflow-x: auto;}
+    }
     div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
     /* The extra [class] is a hack that increases specificity enough to
        override a similar rule in reveal.js */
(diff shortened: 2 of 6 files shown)


jgm:
`make TESTARGS=--accept` will regenerate the tests; you should carefully check any changes.

jgm:
If I understand correctly, it's the use of the `vw` dimension that makes this screen-only.
So maybe with non-screen viewers we should just specify 1.5em unconditionally, rather than `min(4vw, 1.5em)`. Otherwise we lose the gap between columns, which we should keep.

jgm:
thanks!

kbrock:
Thank you @jgm 
Really appreciate pandoc. Also appreciate your attention to detail.

have a great weekend

PR #11628, opened 2026-05-07
Enhance EPUB writer to support multiple EPUB versions and improve HTML format detection.

vreoo:
Added checks for EPUB format variants and updated transformation functions to accommodate versioning. Introduced tests for raw EPUB attributes rendering.

### DIFF
--- a/test/command/8880.md
+++ b/test/command/8880.md
@@ -15,12 +15,16 @@ Raw EPUB attributes are rendered in EPUB output.
 Raw EPUB2 attributes are omitted from EPUB3 output.
 
 ```
-% pandoc -f markdown -t epub3 --metadata title=Raw -o - | pandoc -f epub -t html | grep ok
+% pandoc -f markdown -t epub3 --metadata title=Raw -o - | pandoc -f epub -t html
 ~~~ {=epub2}
 <p>ok</p>
 ~~~
 ^D
-=> 1
+<p><span id="title_page.xhtml"></span></p>
+<p><span id="ch001.xhtml"></span></p>
+<section id="ch001.xhtml_raw" class="unnumbered level1">
+<h1 class="unnumbered">Raw</h1>
+</section>
 ```
 
 ```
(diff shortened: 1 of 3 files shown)


test/command/8880.md
```diff
@@ -0,0 +1,41 @@
...
+% pandoc -f native -t epub --metadata title=Raw -o - | pandoc -f epub -t html
+[ RawBlock (Format "epub") "<p>ok</p>" ]
+^D
+<p><span id="title_page.xhtml"></span></p>
+<p><span id="ch001.xhtml"></span></p>
+<section id="ch001.xhtml_raw" class="unnumbered level1">
+<h1 class="unnumbered">Raw</h1>
+<p>ok</p>
+</section>
+```
```
jgm: I don't think this test adds anything to the yreceding one.  They both test the same thing on the writer side. Maybe just use this one (which doesn't also test the markdown reader) and omit the first one?
However, it would be good to ADD a test that shows that content marked epub2 is omitted for epub3 output (for example).

vreoo: Addressed in 329ad27948e0897289493069b90eb12e77f22eee

I removed the redundant first positive test and kept the `native -> epub` writer-side check, and added a negative case showing that content marked `epub2` is omitted from `epub3` output, while keeping `epub3` case.

test/command/8880.md
```diff
@@ -0,0 +1,38 @@
...
+
+Raw EPUB2 attributes are omitted from EPUB3 output.
+
+```
+% pandoc -f markdown -t epub3 --metadata title=Raw -o - | pandoc -f epub -t html | grep ok
```
jgm: The `| grep ok` is not needed; just make sure the output is right. We don't want to assume that `grep` is installed.

vreoo: Gotcha, it's done in f8a61dde994edd9fe9c8e2f6d78a1550b3f7c865

jgm:
thanks!

PR #11632, opened 2026-05-08
Update footnote references in MANUAL.txt

jidanni:
Avoid user possibly thinking they are responsible for numbering footnotes.


jgm:
This is already clear from the example (`longnote`).

PR #11666, opened 2026-05-24
Docx writer: fix empty keywords in core document properties

### LINKED ISSUE #11665: --to=docx keywords all there, but each one empty
pandoc 3.9.0.2
```
$ man pandoc
   Metadata variables
       keywords
              list of keywords to be included in HTML, PDF, ODT, pptx, docx and
              AsciiDoc metadata; repeat as for author, above

$ for f in docx html
  do echo test|pandoc --standalone -M title:bla \
  -M author:Ralph `for i in a b c d e; \
  do echo -M keywords:$i; done` --output=test.$f; done
$ unzip -c test.docx|perl -nwle 'print for /<dc:.*keywords/g'
<dc:title>bla</dc:title><dc:creator>Ralph</dc:creator><cp:keywords>, , , , </cp:keywords
$ grep keywords test.html
  <meta name="keywords" content="a, b, c, d, e" />
```

As you see above,
```
<cp:keywords>, , , , </cp:keywords
```
pandoc has almost succeeded.
It has made the five places.
It just needs to do the final step
of sticking each value into those places.

SAY-5:
Fixes #11665.

When keywords are supplied as a metadata list, every entry in `<cp:keywords>` of `docProps/core.xml` was rendered empty:

```
$ echo test | pandoc -s -M keywords:a -M keywords:b -M keywords:c -o test.docx
$ unzip -p test.docx docProps/core.xml | grep -o '<cp:keywords>.*</cp:keywords>'
<cp:keywords>, , </cp:keywords>
```

The list elements are `MetaString` values, and `stringify` returns the empty string for a `MetaString` (it only collects `Inline`s). The fix converts each element the same way `lookupMetaString` does, so `MetaString`/`MetaInlines`/`MetaBlocks` are all handled.

Verified the repro and the fix output against pandoc 3.9. I was not able to run the docx golden-test harness locally (no GHC toolchain available); happy to add a core.xml properties test if you'd like one.

### DIFF
--- a/src/Text/Pandoc/Writers/Docx.hs
+++ b/src/Text/Pandoc/Writers/Docx.hs
@@ -738,8 +738,13 @@ mkStylesEntry epochtime styledoc styleMaps st opts =
 -- | Create core document properties entry
 mkCorePropsEntry :: Integer -> UTCTime -> Meta -> Entry
 mkCorePropsEntry epochtime utctime meta =
-  let keywords = case lookupMeta "keywords" meta of
-                       Just (MetaList xs) -> map stringify xs
+  let metaValueToText (MetaString s)    = s
+      metaValueToText (MetaInlines ils) = stringify ils
+      metaValueToText (MetaBlocks bs)   = stringify bs
+      metaValueToText (MetaBool b)      = T.pack (show b)
+      metaValueToText _                 = ""
+      keywords = case lookupMeta "keywords" meta of
+                       Just (MetaList xs) -> map metaValueToText xs
                        _                  -> []
       docPropsPath = "docProps/core.xml"
       extraCoreProps = ["subject","lang","category","description"]
--- a/src/Text/Pandoc/Writers/Docx.hs
+++ b/src/Text/Pandoc/Writers/Docx.hs
@@ -738,13 +738,9 @@ mkStylesEntry epochtime styledoc styleMaps st opts =
 -- | Create core document properties entry
 mkCorePropsEntry :: Integer -> UTCTime -> Meta -> Entry
 mkCorePropsEntry epochtime utctime meta =
-  let metaValueToText (MetaString s)    = s
-      metaValueToText (MetaInlines ils) = stringify ils
-      metaValueToText (MetaBlocks bs)   = stringify bs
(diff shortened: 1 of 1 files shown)


src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -738,8 +738,13 @@ mkStylesEntry epochtime styledoc styleMaps st opts =
 -- | Create core document properties entry
 mkCorePropsEntry :: Integer -> UTCTime -> Meta -> Entry
 mkCorePropsEntry epochtime utctime meta =
-  let keywords = case lookupMeta "keywords" meta of
-                       Just (MetaList xs) -> map stringify xs
+  let metaValueToText (MetaString s)    = s
+      metaValueToText (MetaInlines ils) = stringify ils
+      metaValueToText (MetaBlocks bs)   = stringify bs
+      metaValueToText (MetaBool b)      = T.pack (show b)
+      metaValueToText _                 = ""
```
jgm: Why not just use `lookupMetaString` from T.P.Writers.Shared, instead of reproducing the code?

src/Text/Pandoc/Writers/Docx.hs
```diff
@@ -738,13 +738,9 @@ mkStylesEntry epochtime styledoc styleMaps st opts =
...
 mkCorePropsEntry epochtime utctime meta =
-  let metaValueToText (MetaString s)    = s
-      metaValueToText (MetaInlines ils) = stringify ils
-      metaValueToText (MetaBlocks bs)   = stringify bs
-      metaValueToText (MetaBool b)      = T.pack (show b)
-      metaValueToText _                 = ""
+  let toText v = lookupMetaString "x" (Meta $ M.singleton "x" v)
       keywords = case lookupMeta "keywords" meta of
-                       Just (MetaList xs) -> map metaValueToText xs
+                       Just (MetaList xs) -> map toText xs
```
jgm: Sorry, when I made my suggestion I hadn't taken into account the need to create these singleton Meta values, which is going to introduce unnecessary allocations.  I now think it's cleaner to use the original approach, or we could provide `metaValueToText` as an exported function in T.P.Writers.Shared and use it both here and in the definition of `lookupMetaString`.

PR #11671, opened 2026-05-27
Markdown reader: allow grid tables to be indented

jolars:
Like the other table syntaxes (pipe, simple, and multiline tables) and block-level constructs generally, a grid table may now be indented by up to three spaces and still be recognized as a table.  Previously the grid-table parser required the table to begin at the left margin, so an indented grid table was parsed as a paragraph.

The leading indentation is stripped uniformly from each line before the table is parsed, so an indented grid table produces the same AST as its non-indented equivalent.

Adds a command test.


jgm:
Thanks!  I don't love `parseFromString`, but there seems to be no good alternative since the parsing is done in an external library.

PR #11695, opened 2026-06-10
LaTeX reader and ODT writer fixes for image handling

caniko:
*(Updated: the tikzpicture extraction commits have been dropped per the discussion below — whole-diagram images via a Lua filter are the right approach for that problem. What remains are five self-contained fixes.)*

Working with LaTeX documents that build figures from generated panel layouts surfaced several independent bugs in the LaTeX reader and the ODT/OpenDocument writers:

* **LaTeX reader: expand `\csname...\endcsname`** — dynamically built command names (`\expandafter\def\csname ...\endcsname`, data-driven `\input` paths) now expand. Control sequences built by `\csname` are re-expanded, so csname-built user macros work too.
* **LaTeX reader: resolve `\includesvg` sources, support `\svgpath`** — the svg package's `\includesvg` names an `.svg` file, so resolving it against the raster extension list used for `\includegraphics` found nothing; candidate files are also searched on the resource path.
* **LaTeX reader: expand macros in `\input`/`\include`/`\subfile` filenames** — filenames built from macros were used verbatim and never found.
* **OpenDocument writer: anchor image frames `as-char`** — `Image` is an `Inline`, but frames were emitted without `text:anchor-type`, so consumers applied paragraph anchoring and floated the image beside the surrounding text. (Golden test updates included.)
* **ODT writer: resolve percent image dimensions to valid lengths** — when both width and height are given and the width is a percentage, resolve it against the text width from the reference document's page layout — the same approach the docx writer takes with its reference docx section properties (`envPrintWidth`) — and derive the height from the image's aspect ratio (docx parity: a relative height has no well-defined base).

All tests pass (`cabal test`). Happy to split further if preferred.

### DIFF
--- a/src/Text/Pandoc/Readers/LaTeX.hs
+++ b/src/Text/Pandoc/Readers/LaTeX.hs
@@ -35,7 +35,7 @@ import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Either (partitionEithers)
 import Skylighting (defaultSyntaxMap)
-import System.FilePath (addExtension, replaceExtension, takeExtension)
+import System.FilePath ((</>), addExtension, replaceExtension, takeExtension)
 import Text.Collate.Lang (renderLang)
 import Text.Pandoc.Builder as B
 import Text.Pandoc.Class (PandocPure, PandocMonad (..), getResourcePath,
@@ -231,7 +231,12 @@ doLHSverb =
     <$> manyTill (satisfyTok (not . isNewlineTok)) (symbol '|')
 
 mkImage :: PandocMonad m => [(Text, Text)] -> Text -> LP m Inlines
-mkImage options (T.unpack -> src) = do
+mkImage options = mkImageWithExts options
+  [".pdf", ".png", ".jpg", ".mps", ".jpeg", ".jbig2", ".jb2"]
+
+mkImageWithExts :: PandocMonad m
+                => [(Text, Text)] -> [String] -> Text -> LP m Inlines
+mkImageWithExts options exts' (T.unpack -> src) = do
    let replaceRelative (k,v) =
          case numUnit v of
               Just (num, "\\textwidth") -> (k, showFl (num * 100) <> "%")
@@ -243,19 +248,30 @@ mkImage options (T.unpack -> src) = do
    let attr = ("",[], kvs)
    let alt = maybe (str "image") str $ lookup "alt" options
    defaultExt <- getOption readerDefaultImageExtension
-   let exts' = [".pdf", ".png", ".jpg", ".mps", ".jpeg", ".jbig2", ".jb2"]
    let exts  = exts' ++ map (map toUpper) exts'
-   let findFile s [] = return s
+   resourcePath <- getResourcePath
+   let candidates s =
+         s : [ dir </> s | dir <- resourcePath, not (null dir), dir /= "." ]
+       findExisting [] = return Nothing
+       findExisting (s:ss) = do
+         exists <- fileExists s
+         if exists
+            then return $ Just s
+            else findExisting ss
+       findFile s [] = do
+         found <- findExisting $ candidates s
+         return $ fromMaybe s found
        findFile s (e:es) = do
          let s' = addExtension s e
-         exists <- fileExists s'
-         if exists
-            then return s'
-            else findFile s es
+         found <- findExisting $ candidates s'
+         case found of
+           Just resolved -> return resolved
+           Nothing       -> findFile s es
    src' <- case takeExtension src of
-               "" | not (T.null defaultExt) -> return $ addExtension src $ T.unpack defaultExt
+               "" | not (T.null defaultExt) ->
+                      findFile src [T.unpack defaultExt]
                   | otherwise -> findFile src exts
-               _  -> return src
+               _  -> findFile src []
    return $ imageWith attr (T.pack src') "" alt
 
 removeDoubleQuotes :: Text -> Text
@@ -432,7 +448,8 @@ inlineCommands = M.unions
     -- svg
     , ("includesvg",      do options <- option [] keyvals
                              src <- bracedFilename
-                             mkImage options . unescapeURL $ src)
+                             mkImageWithExts options [".svg", ".png"]
+                               . unescapeURL $ src)
     -- hyperref
     , ("url", (\url -> linkWith ("",["uri"],[]) url "" (str url))
                         . unescapeURL . untokenize <$> bracedUrl)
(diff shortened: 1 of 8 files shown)


jgm:
Regarding the tikz picture extraction: it's a bit hard to see the use of this, as usually the images in a tikz will be parts of a larger diagram; when they're extracted, the context will be lost.  Or is it common to have tikz diagrams that are basically just shells around images?

caniko:
Thanks, that concern makes sense. I pushed an update that gates this behind a new LaTeX reader extension, `tikz_images`:

* default behavior is unchanged: without `raw_tex` the `tikzpicture` is skipped as before;
* `raw_tex` takes precedence and preserves the raw `tikzpicture`, even if `tikz_images` is also enabled;
* extraction only happens with `latex+tikz_images`, and the manual now explicitly documents it as a lossy fallback that does not preserve TikZ coordinates, labels, drawing commands, arrows, or node placement.

I also added reader tests for the default skip behavior, raw preservation, the opt-in extraction path, no-image fallback, and a mixed drawing+image case showing that only the image is extracted.

jgm:
I am reluctant to add a new extension; it's a lot of extra complexity. It's a question of how common this need is. If it's something few people are going to need, handling it with a Lua filter might make more sense.

caniko:
I work in the scientific community, and I think it is very useful for large-scale publications where accuracy matters. In our case, it will be used in our Nature Communications manuscript submission.

Even fewer people would use it if you put it behind Lua IMO; I can make the change — just give me a signal

jgm:
If accuracy matters, then do we want a feature that will convert a diagram in which images serve as parts into a mere list of these images, without the surrounding diagrammatic context?  This isn't obvious to me.  It might be better to have an obvious omission (which can be worked around using a Lua filter that uses LaTeX to generate the tikz image and then includes an image of the whole diagram) than to appear to have translated the figure, but wrongly...

caniko:
The goal was an incremental improvement plan, but I'd be willing to work on a solution for converting native `Table` from TikZ node geometry instead. It is a bit more work.

caniko:
With the current patch, `latex+tikz_images` only converts recognizable TikZ image-composition layouts where the node geometry gives rectangular panels. In that case it emits a native Pandoc `Table`, preserving the inferred grid, gaps, row/column spans, panel labels, and image dimensions. If the picture contains mixed drawing commands or unsupported node syntax, it falls back to the existing skipped-content behavior instead of listing the images out of context. `raw_tex` still takes precedence and preserves the raw `tikzpicture`.

So the extension no longer tries to translate arbitrary TikZ, and it no longer produces a mere list of images for diagrams whose surrounding context matters.

jgm:
I still don't really see the point.  A table with images might capture *some* of the structure of a tikz diagram, depending on what kind of diagram it is. But it isn't rich enough to capture everything; this will be a degraded experience.  Why do this when there are already good solutions that will get the whole diagram as an image?

Here's a filter I've used for a book. Save this as tikzpicture.lua and use `-L tikzpicture.lua` on the command line when converting to ODT.  (This presupposes that you have lualatex and pdftocairo installed, and also TeX Gyre Schola and TeX Gyre Schola Math fonts.  If you don't want to use these fonts then you can simply delete the lines that call for them.  But this allowed me to match the font of the rest of the book.)

```lua
-- This filter converts tikzpicture environments into svg
-- images and adds links to them.

local system = require 'pandoc.system'

local doc_template = [[
\documentclass[10pt,preview=true,class=memoir]{standalone}
\usepackage{unicode-math}
\defaultfontfeatures[\rmfamily]{Ligatures=TeX,Scale=1}
\setmainfont[Numbers=Proportional,Scale=MatchUppercase]{TeX Gyre Schola}
\setmathfont[Scale=1]{TeX Gyre Schola Math}
\usepackage{bookgraphics}
\begin{document}
%s
\end{document}
]]

local function tex2image(src)
  local fbasename = pandoc.sha1(src) .. '.svg'
  if pandoc.mediabag.lookup(fbasename) then
    return fbasename
  end
  local tmp = 'tmp.tikz'
  local outfile = tmp .. '/' .. fbasename
  local texfile = outfile:gsub('%.svg$', '.tex')
  local pdffile = outfile:gsub('%.svg$', '.pdf')
  -- don't run tex again if the file is already created
  local file = io.open(outfile, "r")
  if file then
    local contents = file:read("*a")
    pandoc.mediabag.insert(fbasename, 'image/svg+xml', contents)
    file:close()
    return fbasename
  end
  local f = io.open(texfile, 'w')
  f:write(doc_template:format(src))
  f:close()
  local _, _, status = os.execute('lualatex -pdflua -auxdir=' .. tmp .. ' -interaction=nonstopmode -halt-on-error -shell-escape --output-directory=' .. tmp .. ' ' .. texfile)
  if status ~= 0 then
      os.exit(1)
  end
  local _, _, status = os.execute('pdftocairo -r 300 -origpagesizes -nocenter -svg ' .. pdffile .. ' ' .. outfile)
  if status ~= 0 then
      os.exit(1)
  end
  file = io.open(outfile, "r")
  local contents = file:read("*a")
  pandoc.mediabag.insert(fbasename, 'image/svg+xml', contents)
  return fbasename
end

local function file_exists(name)
  local f = io.open(name, 'r')
  if f ~= nil then
    io.close(f)
    return true
  else
    return false
  end
end

function RawBlock(el)
  if el.format == 'latex' and
    (el.text:match("^\\begin%{tikz") or el.text:match("^%$")) then
    local txt = el.text:gsub("scale=0.7", "scale=1.0")
    local fbasename = tex2image(txt)
    local alt = {}
    for alttext in string.gmatch(el.text, '%% alt="([^"]*)"') do
      alt = alttext
    end
    return pandoc.Para({pandoc.Image(alt, fbasename)})
  else
   return el
  end
end
```

caniko:
That settles it — you're right, and your argument applies to our own output more than I'd like to admit: the panel extraction was silently dropping the panel letter labels (they're TikZ nodes, not images), which is exactly the "appears translated, but wrongly" failure mode you describe. A faithful whole-diagram image is the correct representation.

Thanks for the filter. Our pipeline already renders each composite figure to a standalone 300 dpi image, so we'll inject those with a small Lua filter along the same lines — no reader involvement needed.

I've dropped the tikzpicture commits from this PR (title and description updated). What remains are five self-contained fixes that stand on their own, and which the whole-figure-image approach still relies on for sizing: `\csname` expansion, `\includesvg`/`\svgpath` resolution, macro expansion in include filenames, `as-char` anchoring for inline image frames, and percent dimensions resolved against the reference document's page layout (docx parity). Full test suite passes.

caniko:
Following up on the discussion with @jgm — the tikz-picture extraction approach has been dropped in favor of the Lua filter pattern he recommended. The remaining ODT/OpenDocument writer fixes (image `as-char` anchoring, percent dimension resolution against reference document text width) have been split into a focused PR at https://github.com/jgm/pandoc/pull/NEWPRNUMBER.

Closing this PR. The ODT fixes are available for review in the new PR once I open it.

caniko:
Following up on the discussion with @jgm — the tikz-picture extraction approach has been dropped in favor of the Lua filter pattern he recommended. The remaining ODT/OpenDocument writer fixes (image `as-char` anchoring, percent dimension resolution against reference document text width) have been split into a focused PR: #11746

PR #11699, opened 2026-06-11
Add Txt2Tags writer

luginf:
Adds `t2t` as an output format, producing Txt2Tags markup
(<https://txt2tags.org>). Covers headers, lists (bullet, ordered,
definition), tables, links, images, code blocks, inline formatting,
and raw txt2tags/HTML passthrough.

Supersedes #6551, rebased onto current `main` as a single commit
(the history there had gotten tangled across forks, which was
tripping the `check-commit-msg-length` check).


data/templates/default.t2t
```diff
@@ -0,0 +1,19 @@
+$pagetitle$
+$author-meta$
+$date-meta$
+
+
+
+$for(include-before)$
```
jgm, line 5: Should this all be inside a conditional `$if(pagetitle)$..$endif$`?
otherwise you'll get a big blank space at the beginning if the doc doesn't have a title.

src/Text/Pandoc/Writers/Txt2Tags.hs
```diff
@@ -0,0 +1,396 @@
+{-# LANGUAGE OverloadedStrings #-}
+{- |
+   Module      : Text.Pandoc.Writers.Txt2Tags
+   Copyright   : Copyright (C) 2008-2024 Eric Forgeot, based on John MacFarlane DokuWiki writer
+   License     : GNU GPL, version 2 or above
+
+   Maintainer  : Clare Macrae <clare.macrae@googlemail.com>
```
jgm: Is this correct? I note that the author and maintainer don't match.

src/Text/Pandoc/Writers/Txt2Tags.hs
```diff
@@ -0,0 +1,396 @@
...
+import Text.Pandoc.Options (WrapOption (..), WriterOptions (writerTableOfContents,
+                            writerTemplate, writerWrapText), isEnabled)
+import Text.Pandoc.Shared (figureDiv, linesToPara, removeFormatting, trimr)
+import Text.Pandoc.URI (escapeURI, isURI)
+import Text.Pandoc.Templates (renderTemplate)
+import Text.DocLayout (render, literal)
+import Text.Pandoc.Writers.Shared (defField, metaToContext, toLegacyTable)
+
+data WriterState = WriterState {
+  }
```
jgm: Is there is no need for state, is there a reason to keep the StateT part?

test/writer.txt2tags
```diff
@@ -0,0 +1,653 @@
+
```
jgm: I assume this file, test/writer.txt2tags, is obsolete and should be deleted (in favor of writer.t2t)?

MANUAL.txt
```diff
@@ -363,6 +363,7 @@ header when requesting a document from a URL:
     - `rtf` ([Rich Text Format])
     - `texinfo` ([GNU Texinfo])
     - `textile` ([Textile])
+    - `t2t` ([txt2tags])
```
jgm, line 366: In addition to this, you may want to add something to the "Math" section where it describes how math is rendered in the various output formats.

src/Text/Pandoc/Writers/Txt2Tags.hs
```diff
@@ -0,0 +1,396 @@
...
+  return $ contents <> if T.null indent then "\n" else ""
+
+blockToTxt2Tags opts (LineBlock lns) =
+  blockToTxt2Tags opts $ linesToPara lns
+
+blockToTxt2Tags opts b@(RawBlock f str)
+  | f == Format "txt2tags" = return str
+  -- Use the Txt2Tags raw area syntax (""") for block-level HTML pass-through:
+  | f == Format "html"
+  , isEnabled Ext_raw_html opts = return $ "\"\"\"\n" <> str <> "\n\"\"\"\n"
```
jgm: If your writer is sensitive to an extension, you will need to add something to Text.Pandoc.Extensions to make sure that `raw_html` is a possible extension, and to activate it by default if desired.

However: do you really need to gate this behind an extension?  If the standard t2t language allows raw HTML in this way, then it should just happen by default.  Extensions are for optional language features.  You can simplify by getting rid of extension-sensitivity.

src/Text/Pandoc/Writers/Txt2Tags.hs
```diff
@@ -0,0 +1,396 @@
...
+  return $ "--" <> contents <> "--"
+
+-- Txt2Tags has no superscript/subscript syntax; fall back to HTML tags.
+inlineToTxt2Tags opts (Superscript lst) = do
+  contents <- inlineListToTxt2Tags opts lst
+  return $ "<sup>" <> contents <> "</sup>"
+
+inlineToTxt2Tags opts (Subscript lst) = do
+  contents <- inlineListToTxt2Tags opts lst
+  return $ "<sub>" <> contents <> "</sub>"
```
jgm: Tempting to simplify this recurring pattern with a helper function, e.g.
```
= surroundInlines "<sub> "</sub>" lst
```

jgm (changes requested):

data/templates/default.t2t
luginf: Fixed differently: $ pagetitle $/$author-meta$/$date-meta$ were never actually populated for t2t (those are HTML/LaTeX-only vars), so the header was always 6 blank lines. Switched to $ title $/$author$/$date$ (the real metadata, t2t requires exactly these 3 header lines per its spec) and trimmed the extra spacer to 1 line. 

So if there is a big blank space in the case the doc doesn't have a title, it's the expected behavior.

No-title docs now get 4 blank lines instead of 6.

src/Text/Pandoc/Writers/Txt2Tags.hs
luginf: Fixed. Maintainer now matches Copyright (Eric Forgeot)

src/Text/Pandoc/Writers/Txt2Tags.hs
luginf: No, removed. WriterState was empty. Now just ReaderT WriterEnvironment m.

test/writer.txt2tags
luginf: Yes, it is indeed obsolete. test/writer.t2t is the one used by the test suite

MANUAL.txt
luginf: It has been added. Txt2Tags now listed alongside Markdown/Org/ConTeXt/ZimWiki for verbatim $...$/$$...$$ math.

src/Text/Pandoc/Writers/Txt2Tags.hs
luginf: ok, it has been removed. Ext_raw_html gating : t2t's \"\"\"..\"\"\"/\"\"..\"\" raw-pass-through is standard
    syntax, not an optional feature. Also updated MANUAL.txt's raw-HTML format list.

src/Text/Pandoc/Writers/Txt2Tags.hs
luginf: Done. Added and used for Emph/Underline/Strong/Strikeout/Superscript/Subscript/Quoted.

jgm:
Thanks!

PR #11703, opened 2026-06-12
Docx writer: use namespace URI instead of prefix in extractPageLayout

bitkojine:
follow up to PR with same title to fix tests failing


jgm:
sorry, I didn't notice the test failures before merging.

PR #11706, opened 2026-06-14
Note mystery behaviour in both spots in MANUAL.txt

jidanni:
This mysterious quirk needs to also be documented here as well as the other spot already in the man page.

Else half of users will still scratch their heads as to why output goes to STDOUT.

Consider also triggering an error if the user uses an --output setting that will be ignored.

Or consider making it work as users expect, and removing mention in the man pages of this quirk.

Wait, it turns out
- --print-highlight-style
- --print-default-data-file
- --print-default-template

all need to be mentioned. No, the user doesn't necessarily check each one of them, some users only check --output.

Anyway, currently if the order is not followed, output mysteriously goes to STDOUT. Just like a bug.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -394,7 +394,8 @@ header when requesting a document from a URL:
     then instead of producing a `.zip` file pandoc will create
     a directory *FILE* and unpack the zip archive there
     (unless *FILE* already exists, in which case an error
-    will be raised).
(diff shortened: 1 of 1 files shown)


jgm:
The commit message should be made clearer and more descriptive of what has changed.  (We use these to construct the changelog.)
Also, you might as well include those other options.

PR #9952, opened 2024-07-04
WIP: Create a working `flake.nix`

srid:
- [x] Bootstrap `flake.nix` based on haskell-flake, overriding dependencies
- [ ] `nix build` to work
- [ ] devShell (`nix develop`) to work
- [ ] HLS to work
- [ ] Upstream `skylighting` Nix
- [ ] How should Nix package deal with cabal flags?

PR #11707, opened 2026-06-15
Revamp flake.nix to use haskell-flake (nix build / nix run)

srid:
## Summary

Revamps `flake.nix` to use [haskell-flake](https://haskell.nixos.asia/) so that:

- `nix build` builds the `pandoc` executable (from `pandoc-cli`)
- `nix run` runs it
- `nix develop` gives a dev shell (cabal, HLS, the wasm toolchain, the various profiling/validation tools — unchanged from before)

This is a fresh take on the stale #9952.

Verified end-to-end from a clean checkout on both platforms (`nix build` + `nix run -- --version` + a markdown→HTML conversion):

| | Linux (x86_64) | macOS (aarch64) |
|---|---|---|
| `nix build` | ✅ | ✅ |
| `nix run -- --version` → `pandoc 3.10`, `+server +lua` | ✅ | ✅ |

A `Nix` GitHub Actions workflow builds the flake on `ubuntu-latest` and `macos-latest`.

## Notable points

- haskell-flake's `cabal.project` parser can't handle the `if arch(wasm32)` conditional blocks, so the packages are listed explicitly in the flake instead of auto-discovered.
- nixpkgs's pinned `all-cabal-hashes` is too old for pandoc 3.10's dependency versions, so the flake points the Haskell package set at a recent snapshot.
- The dependencies that nixpkgs ships too old (or, for `asciidoc`, not at all) for pandoc 3.10's `build-depends` are pinned to the highest in-range Hackage versions (collected in one `bumpedDeps` map): `typst`, `citeproc`, `commonmark*`, `emojis`, `djot`, `pandoc-types`, `asciidoc`, and the `hslua` 2.5 family. `texmath` uses the `source-repository-package` git pin via a flake input.
- The sub-packages' `COPYING.md` is a `../COPYING.md` symlink that dangles once a sub-directory is isolated as a source; the flake builds them from source (not sdist) and copies the real license file in.

Opening as a **draft** for discussion (e.g. whether to track the fast-moving pandoc-ecosystem deps as flake inputs instead of Hackage version pins).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,75 +1,167 @@
 {
   description = "pandoc";
 
+  # Uses haskell-flake (https://haskell.nixos.asia) to build the local Haskell
+  # packages (pandoc, pandoc-cli, pandoc-lua-engine, pandoc-server) defined in
+  # cabal.project.
+  #
+  #   nix build   # builds the `pandoc` executable (from pandoc-cli)
+  #   nix run     # runs it
+  #   nix develop # dev shell (cabal, HLS, the wasm toolchain, etc.)
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs";
-    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    haskell-flake.url = "github:srid/haskell-flake";
+
+    # WASM toolchain, used (in the dev shell) to build pandoc for wasm32.
     ghc-wasm-meta.url = "gitlab:haskell-wasm/ghc-wasm-meta?host=gitlab.haskell.org";
+
+    # Git dependency pinned in cabal.project's source-repository-package.
+    texmath = {
+      url = "github:jgm/texmath/170899673ee31de9096e178605e8da31a36e4185";
+      flake = false;
+    };
   };
 
-  outputs = { self, nixpkgs, flake-utils, ghc-wasm-meta }:
-    flake-utils.lib.eachDefaultSystem (system:
-      let
-        pkgs = nixpkgs.legacyPackages.${system};
+  outputs = inputs@{ flake-parts, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      systems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" "x86_64-darwin" ];
+      imports = [ inputs.haskell-flake.flakeModule ];
 
-        haskellPackages = pkgs.haskellPackages;
-        # Or, override some dependencies as follows:
-        # haskellPackages = (pkgs.haskellPackages.override {
-        #   all-cabal-hashes = pkgs.fetchurl {
-        #    # The hash in the URL is just a git commit hash
-        #    url = "https://github.com/commercialhaskell/all-cabal-hashes/archive/2e3f153549871ada6ecbf36719339a5da051bc76.tar.gz";
-        #    sha256 = "ymHZb6wCZlrtKb+T+iOL17jyN8IzA7s52uiamUIkWNI=";
-        #   };
-        # }).extend(self: super: {
-        #   citeproc = pkgs.haskell.lib.dontCheck (self.callHackage "citeproc" "0.9.0.1" {});
-        #   texmath = self.callHackage "texmath" "0.12.10.2" {};
-        # });
+      perSystem = { self', pkgs, lib, system, ... }:
+        let
+          # Hackage versions that nixpkgs ships too old (or, for asciidoc, not
+          # at all) for pandoc 3.10's `build-depends` bounds. These are the
+          # highest Hackage versions in range; pandoc 3.10 itself proves they
+          # are mutually compatible.
+          bumpedDeps = {
+            typst = "0.10";
+            typst-symbols = "0.2";
+            citeproc = "0.13.0.1";
+            commonmark = "0.3";
+            commonmark-extensions = "0.2.7";
+            commonmark-pandoc = "0.3";
+            emojis = "0.1.5";
+            djot = "0.1.4";
+            pandoc-types = "1.23.1.2";
+            asciidoc = "0.1.0.3";
 
-        jailbreakUnbreak = pkg:
-          pkgs.haskell.lib.doJailbreak (pkg.overrideAttrs (_: { meta = { }; }));
-
-        # PUT YOUR PACKAGE NAME HERE:
-        packageName = "pandoc";
-
-        wasmToolchain = ghc-wasm-meta.packages.${system}.default;
-        # Alternatively, if you want a specific "bundle":
-        # wasmToolchain = ghc-wasm-meta.packages.${system}.all_9_14;
-      in {
-        packages.${packageName} =
-          haskellPackages.callCabal2nix packageName self rec {
-            # Dependency overrides go here
+            # The hslua 2.5 ecosystem (used by pandoc-lua-engine). The
+            # sub-libraries version independently of the `hslua` umbrella.
+            hslua = "2.5.0";
+            hslua-core = "2.3.2.1";
+            hslua-marshalling = "2.3.2";
+            hslua-classes = "2.3.2";
+            hslua-objectorientation = "2.5.0";
+            hslua-packaging = "2.4.1";
+            hslua-aeson = "2.3.2";
+            hslua-typing = "0.2.0";
+            hslua-module-path = "1.2.0";
+            hslua-module-system = "1.3.0";
+            hslua-module-text = "1.2.0";
+            hslua-module-version = "1.2.0.1";
+            hslua-module-zip = "1.2.1";
+            hslua-module-doclayout = "1.2.1.1";
           };
 
-        packages.default = self.packages.${system}.${packageName};
-        defaultPackage = self.packages.${system}.default;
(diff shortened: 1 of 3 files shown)


srid:
@jgm I'll undraft this PR once it is ready for review. In the meanwhile, could you please approve the Github workflows?

EDIT: Not critical though, as I got Claude to do it on my fork:

image

srid:
Okay this is ready for a review!

image

srid:
One design question before taking this further: how would you like the flake to handle dependencies that nixpkgs ships **too old (or not at all)** for pandoc 3.10?

nixpkgs' Haskell set currently lags pandoc's `build-depends`, so the flake has to override ~25 packages — e.g. `citeproc` (0.9 → 0.13), `typst` (0.8 → 0.10), `commonmark*` (0.2 → 0.3), `pandoc-types` (1.23.1 → 1.23.1.2), `emojis`, `djot`, `asciidoc` (not in nixpkgs at all), and the whole `hslua` 2.5 family. I resolve these by (a) pointing the package set at a recent `all-cabal-hashes` snapshot and (b) pinning exact in-range Hackage versions (collected in one `bumpedDeps` map).

Three ways to maintain this going forward — which do you prefer?

1. **Exact Hackage version pins** (what's here now) — precise and always matches your `build-depends`; downside is the `all-cabal-hashes` pin and the version map need bumping whenever you raise a dependency.
2. **Flake `inputs` (git) for the fast-moving deps you maintain** (`texmath`, `citeproc`, `typst`, the `hslua` family, …) — `nix flake update` pulls them automatically; downside is a git ref can drift from the actual released version on Hackage.
3. **Rely on nixpkgs** — drop most overrides and only support building pandoc once nixpkgs' Haskell set has caught up to its dependencies.

This mostly determines who carries the update burden and how closely the flake tracks your releases, so your preference would steer the rest of the PR.

jgm:
Thanks for doing this and for the detailed writeup!

The thing I don't like about this (not necessarily a deal-breaker) is the need to manually track versions.  We already have to do something like this in stack.yaml, and it's a pain; I've considered removing stack.yaml because of it.  I need to think about the costs and benefits.

I didn't understand what you said here:

> haskell-flake's cabal.project parser can't handle the if arch(wasm32) conditional blocks, so the packages are listed explicitly in the flake instead of auto-discovered.

I didn't see the special versions needed for the wasm32 build listed in the flake?

srid:
> I've considered removing stack.yaml because of it. I need to think about the costs and benefits.

What would you use for pandoc development without stack.yaml? Direct cabal, or would you switch to Nix primarily for development?

In any case, I'm happy to assist with the Nix related stuff.

> I didn't see the special versions needed for the wasm32 build listed in the flake?

Do we need the flake to build the wasm version as well? I've explicitly ignored it for now.

srid:
In c9e7c50 the flake now derives its dependency overrides (and the texmath git pin) from stack.yaml's extra-deps, parsed at eval time — single source of truth, so bumping stack.yaml on a release updates the flake automatically. Only a 3-package gap (hslua-core/classes/aeson) stays pinned flake-side, since your LTS already ships those new enough but nixpkgs doesn't.

jgm:
> What would you use for pandoc development without stack.yaml? Direct cabal, or would you switch to Nix primarily for development?

I'd just use cabal. It has worked well for a long time and I haven't used stack in ages.

I will check out your changes here. If it just requires manual intervention on stack.yaml, which I already do, then I think we might have a winner!

flake.nix
```diff
@@ -1,75 +1,164 @@
...
+          # name -> version, derived from stack.yaml's plain `pkg-version` entries.
+          bumpedDeps =
+            lib.listToAttrs
+              (map parseDep (builtins.filter builtins.isString stackExtraDeps))
+            // {
+              # nixpkgs ships these older than stack's LTS resolver does, so they
+              # are not pinned in stack.yaml but still need overriding here.
+              hslua-core = "2.3.2.1";
+              hslua-classes = "2.3.2";
+              hslua-aeson = "2.3.2";
```
jgm, line 54: I wonder if it would be less fragile to just add these to stack.yaml with a comment to check the nixpkgs version (with a URL for where to look) before removing the lines?

flake.nix
```diff
@@ -1,75 +1,164 @@
...
 
-        # PUT YOUR PACKAGE NAME HERE:
-        packageName = "pandoc";
+          # git `source-repository-package`s from stack.yaml (currently texmath).
+          # The package name is taken from the repo basename.
+          gitSources = lib.listToAttrs (map
+            (d: lib.nameValuePair
+              (lib.removeSuffix ".git" (baseNameOf d.git))
+              { source = builtins.fetchGit { url = d.git; rev = d.commit; allRefs = true; }; })
+            (builtins.filter builtins.isAttrs stackExtraDeps));
```
jgm, line 63: What if the repo basename is different from package name? E.g. my typst-hs, where the package name is typst. This is not uncommon for the pandoc deps I pin to git commits.  Another example is the skylighting dependencies. They all live in one repository, with several packages in different subdirectories.  This latter problem could be avoided if the stack `subdirs` is parsed too.

flake.nix
```diff
@@ -1,75 +1,164 @@
...
-        # Alternatively, if you want a specific "bundle":
-        # wasmToolchain = ghc-wasm-meta.packages.${system}.all_9_14;
-      in {
-        packages.${packageName} =
-          haskellPackages.callCabal2nix packageName self rec {
-            # Dependency overrides go here
+          # The sub-packages' `COPYING.md` is a `../COPYING.md` symlink that
+          # dangles once the sub-directory is isolated as a source, breaking
+          # the `license-file` copy during install. Replace it with the real
+          # file from the repo root.
```
jgm, line 68: I'm open to replacing these symlinks with regular files if they are troublesome.  That would simplify this.

flake.nix
```diff
@@ -1,75 +1,164 @@
...
+              rm -f COPYING.md
+              cp ${inputs.self}/COPYING.md COPYING.md
+            '';
+          });
+        in
+        {
+        haskellProjects.default = {
+          # pandoc 3.10's dependencies (citeproc 0.13, typst 0.10, commonmark
+          # 0.3, ...) are newer than the all-cabal-hashes snapshot pinned in
+          # nixpkgs, so `callHackage` can't find them. Point the package set at
```
jgm, line 80: Probably shouldn't hard-code versions in this comment, as it will get stale.

jgm: Hm, but I guess this URL will need to be manually updated every time we rely on a new package from Hackage?

flake.nix
```diff
@@ -1,75 +1,164 @@
...
+            };
           };
 
-        packages.default = self.packages.${system}.${packageName};
-        defaultPackage = self.packages.${system}.default;
+          # haskell-flake's cabal.project parser can't handle pandoc's
+          # `if arch(wasm32)` conditional blocks, so disable auto-discovery
+          # and list the packages explicitly. Sub-package sources are derived
+          # from the flake root (`self`) so they are recognized as local
+          # (i.e. under `projectRoot`).
```
jgm, line 93: Would this be a lot simpler if we didn't use `if arch()`?  We could maybe switch to using the `--project-file` cabal option with a separate wasm-specific cabal project file for the wasm build.

flake.nix
```diff
@@ -1,75 +1,164 @@
...
+              # `+embed_data_files`), so the binary works without a data dir.
+              cabalFlags.embed_data_files = true;
+              # Skip the (slow) test suite and haddock in `nix build`.
+              check = false;
+              haddock = false;
+            };
+            # The sub-package `.cabal` files reference `license-file: COPYING.md`
+            # via a `../COPYING.md` symlink that dangles once the sub-directory
+            # is isolated as a source; building from sdist would fail on it. The
+            # license file isn't needed to compile, so build from source directly.
```
jgm, line 118: There's this comment and the above comment about COPYING.md. Is one of them obsolete?  I was a bit confused about the workaround. Again, the easiest solution might be getting rid of the symlink in our sources.

flake.nix
```diff
@@ -1,75 +1,164 @@
...
+            pandoc-cli = { check = false; haddock = false; buildFromSdist = false; custom = fixLicense; };
+            pandoc-lua-engine = { check = false; haddock = false; buildFromSdist = false; custom = fixLicense; };
+            pandoc-server = { check = false; haddock = false; buildFromSdist = false; custom = fixLicense; };
+
+            # The git sources (texmath) and the bumped Hackage deps: skip
+            # tests/haddock and relax stale bounds against this GHC's boot libs.
```
jgm, line 124: Let's remove the reference to texmath, as that may change and we'll forget to change the comment.

flake.nix
```diff
@@ -1,75 +1,164 @@
...
+          }
+          // gitSources
+          // lib.mapAttrs (_: version: { source = version; }) bumpedDeps;
+
+          settings = {
+            pandoc = {
+              # Build the self-contained executable (mirrors cabal.project's
+              # `+embed_data_files`), so the binary works without a data dir.
+              cabalFlags.embed_data_files = true;
+              # Skip the (slow) test suite and haddock in `nix build`.
```
jgm, line 111: Is there a need to set this, given that it is set to True by default in the cabal.project file ?

flake.nix
jgm: I did this...no more symlinks.

flake.nix
jgm: I added these to stack.yaml so that this can be simplified.

PR #11708, opened 2026-06-16
Update MANUAL.txt clarifying fontsize variable works for screen only

jidanni:
Added note about using --css for printed media font size, else users expect this variable will control it.


jgm:
I think instead I'm going to take out the hard-coded font-size override for print css.-

PR #11709, opened 2026-06-16
Update MANUAL.txt with output argument note

jidanni:
Add note about command line argument order for output.

Users looking at the paragraphs for
--print-highlight-style,
--print-default-data-file, or
--print-default-template
will see an --output caveat.

This pull request adds the corresponding caveat
to the --output paragraph. Else users only looking at the --output paragraph will miss the caveat.

(No need to specify what happens if the user doesn't do as instructed.)

Replaces #11706.


jgm:
I had second thoughts and reverted this. I don't think it's needed because the caveat is in the other clauses.  Best to keep the manual as compact as possible.

PR #11730, opened 2026-06-23
Typst template: allow multiword keywords

tarleb:
Typst expects single words as keywords, so keywords consisting of multiple words must be quoted (and should not be wrapped).


tarleb:
This change allows to compile documents like this, which currently fails with an error.

``` md
---
title: Typst template test
keywords:
  - template testing
  - document conversion
  - unreasonably long keyword that's more like a sentence
---

Nothing to see here.
```

Issue #11712, opened 2026-06-16
Empty task list is converted to a ballot box, ☐ or ☒


lollipopman:
**Explain the problem.**

I looked for a similar issue, but couldn't find one, so I'm not sure if this behavior is expected. Task list items which are empty, i.e. they lack a description, are converted to ballot boxes. My expectation was that they would remain as markdown checkboxes. This occurs both with checked and unchecked tasks.

input.md

```
- [ ] 
- [ ] bubbles
```

command:

```
$ pandoc -f gfm -t gfm input.md -o output.md
```

output.md

```
- ☐
- [ ] bubbles
```

**Pandoc version?**

This was tested on Debian with pandoc 3.9.0.2. The results are similar with the 3.10 wasm version, though indentation is also added on the second task item.

Diamagnetic:
Hi, I would like to work on this issue

jgm:
@Diamagnetic Go for it!

jgm:
Looks like the problem is really in commonmark-extensions (jgm/commonmark-hs).
```
% cabal run commonmark-cli -- -xtask_lists
- [ ] a
- [ ] b
<ul class="task-list">
<li><input type="checkbox" disabled="" />a
</li>
<li><input type="checkbox" disabled="" />b
</li>
</ul>
% cabal run commonmark-cli -- -xtask_lists
- [ ]
- [ ] b
<ul class="task-list">
<li><input type="checkbox" disabled="" /><ul class="task-list">
<li><input type="checkbox" disabled="" />b
</li>
</ul>
</li>
</ul>
jgm@Johns-MacBook-Air:~/src/commonmark-hs %
```
In the second case we get a nested list for some reason.

jgm:
There may also be an issue in pandoc's Markdown writer, which can be illustrated thus:
```
% pandoc -f gfm -t gfm                    
- [ ]
- ☐
```
We should get `- [ ]` here.

lollipopman:
Just built pandoc and it passed my local tests, thanks for fixing! Now even my empty todo lists will render properly 🫠

PR #11743, opened 2026-07-02
Fix GFM empty task list conversion to ballot, and nesting regression

Diamagnetic:
This PR preserves empty task lists on GFM input, and prevents malformed nested lists fixing #11712.

What was done:
- parse GFM input with `Ext_task_lists` disabled so the upstream `taskListSpec` no longer produces malformed nested lists.
- post-process parsed blocks with `convertTaskListItems` to restore checkbox glyphs only when the extension is enabled.
- extended `taskListItemFromAscii` and `taskListItemToAscii` to handle bare markers and add command tests covering empty and nested task-list scenarios.

### DIFF
--- a/src/Text/Pandoc/Readers/CommonMark.hs
+++ b/src/Text/Pandoc/Readers/CommonMark.hs
@@ -22,10 +22,11 @@ import Commonmark.Extensions
 import Commonmark.Pandoc
 import Data.Text (Text)
 import Text.Pandoc.Class.PandocMonad (PandocMonad)
-import Text.Pandoc.Definition
+import Text.Pandoc.Definition as D
 import Text.Pandoc.Builder as B
 import Text.Pandoc.Options
 import Text.Pandoc.Readers.Metadata (yamlMetaBlock)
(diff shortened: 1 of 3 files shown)


jgm:
I see, you're just bypassing commonmark-extensions's task list extension and reimplementing it here.

I would rather see the underlying issue fixed in commonmark-extensions.

If you don't want to take that on, you could just fix the markdown writer issue I noted in my last comment on the linked issue, and I could look at commonmark-extensions.

Diamagnetic:
Thanks for the clarification.

 I’ll rework this as:
  1. a fix in [commonmark-hs/commonmark-extensions](https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions) for the nested - [ ] behavior, and
  2. a separate fix for the markdown writer

Diamagnetic:
@jgm I have made a fix for the Markdown writer issue. Would you prefer that I open it as a separate PR, or should I include it in this one?

jgm:
Cleaner to close this and open a new one so that the message can clearly describe what is being done.

PR #11748, opened 2026-07-06
Add support for tight terms in Typst writer

adql:
By making the item concatenation function depend on `isTightList`.

### DIFF
--- a/src/Text/Pandoc/Writers/Typst.hs
+++ b/src/Text/Pandoc/Writers/Typst.hs
@@ -264,8 +264,11 @@ blockToTypst block =
       return $ (if isTightList items
                    then vcat items'
                    else vsep items') $$ blankline
-    DefinitionList items ->
-      ($$ blankline) . vsep <$> mapM defListItemToTypst items
+    DefinitionList items -> do
+      let concat' = if all (isTightList . snd) items
+                    then vcat
+                    else vsep
+      ($$ blankline) . concat' <$> mapM defListItemToTypst items
     Table (ident,tabclasses,tabkvs) (Caption _ caption) colspecs thead tbodies tfoot -> do
       let lab = toLabel FreestandingLabel ident
       capt' <- if null caption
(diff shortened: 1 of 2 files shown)


jgm:
Is this a merely cosmetic change in the typst output, or does the lack of blank lines make a difference in the PDF produced?

adql:
This is a part of the Typst [terms markup](https://typst.app/docs/reference/model/terms/#parameters-tight):

> **tight**
>
> Defines the default [spacing](https://typst.app/docs/reference/model/terms/#parameters-spacing) of the term list. If it is `false`, the items are spaced apart with [paragraph spacing](https://typst.app/docs/reference/model/par/#parameters-spacing). If it is `true`, they use [paragraph leading](https://typst.app/docs/reference/model/par/#parameters-leading) instead. This makes the list more compact, which can look better if the items are short.
>
> In markup mode, the value of this parameter is determined based on whether items are separated with a blank line. If items directly follow each other, this is set to `true`; if items are separated by a blank line, this is set to `false`. The markup-defined tightness cannot be overridden with set rules.

However, come to think of it, Pandoc's default template uses a `#show terms.item: it => ...`, where the `tight` parameter of `terms` is not used and so there is no effect in the pdf. Still, as this is a part of Typst's markup, it's generally more than a cosmetic change, as it is useful when using a custom template or not using `--standalone` at all.

I use it myself with a custom template which makes the `terms` tabulated, and use the `tight` parameter in the `show` rule. Some definition lists in the document (a CV) have longer definitions and some are one-liners, which calls for different gaps between items for compactness/clarity.

We could also add usage of `tight` in the template to have an effect in the default pdf output, but probably we'll want to consider the aesthetics of it to make it look good. Typst vanilla styling of a tight definition list is pretty ugly since the gap between items is smaller than the gap between terms and definitions.

jgm:
I see.  You can use `make TESTARGS=--accept` to update the test results -- just be sure to check any changes there carefully for correctness.

jgm:
thanks!

PR #11753, opened 2026-07-10
RST reader: Fix nested placeholder resolution for inline elements

haxtibal:
pandoc produces `##REF##` and `##SUBST##` placeholders during parsing, then resolves them with actual values in `resolveReferences` with `walkM`. However, a single `walkM` run is not generally enough to fully resolve placeholders. For example, this RST

```
.. _target:

See |sub|.

.. |sub| replace:: `text <target_>`_
```

processed with `pandoc -f rst -t html` has an unresolved placeholder in the result

```
<div id="target">
<p>See <a href="##REF##target">text</a>.</p>
</div>
```

We can fix it by applying `resolveReferences` recursively for Link nodes until the resolved node has no more placeholders. See individual commit messages for details.

Note: There's a similar issue for `##NOTES##` at block level, but I considered it out of scope in the current PR.

### DIFF
--- a/src/Text/Pandoc/Readers/RST.hs
+++ b/src/Text/Pandoc/Readers/RST.hs
@@ -25,6 +25,7 @@ import Data.List (deleteFirstsBy, elemIndex, nub, partition, sort, transpose)
 import qualified Data.Map as M
 import Data.Maybe (fromMaybe, maybeToList, isJust, isNothing, catMaybes)
 import Data.Sequence (ViewR (..), viewr)
+import qualified Data.Set as Set
 import Data.Text (Text)
 import qualified Data.Text as T
 import Text.Printf (printf)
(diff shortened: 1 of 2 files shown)


haxtibal:
My Haskell is bad and consulting AI doesn't necessarily compensate, sorry in advance. If the code doesn't meet your standards feel free to close and consider the PR as pure bug report.

src/Text/Pandoc/Readers/RST.hs
```diff
@@ -186,7 +186,10 @@ resolveBlockSubstitutions (Para [Link _attr ils (s,_)])
 resolveBlockSubstitutions x = return x
 
 resolveReferences :: PandocMonad m => Inline -> RSTParser m Inline
-resolveReferences x@(Link _ ils (s,_))
+resolveReferences = resolveReferences' []
+
+resolveReferences' :: PandocMonad m => [Key] -> Inline -> RSTParser m Inline
```
jgm: I'd suggest using `Set Key` instead of `[Key]` for performance reasons. It probably doesn't matter much with realistic sizes, but it's an easy change.

haxtibal: Good idea, references are meant to be unique inside `seen`. Don't know about performance impact, but semantically it makes more sense to think of `seen` as a set.

jgm:
thanks!

PR #11784, opened 2026-07-30
Typst template: rename horizontalrule to horizontalRule

iandol:
The polyfill calls a camelCase `horizontalRule` yet the function is not camelCase. It should either be camelCase or not, I chose to use camelCase for the function name, though as long as they match it doesn't matter...


jgm:
Thanks! You just need to update the expected test outputs to make the failures go away.

PR #11790, opened 2026-08-03
RTF Reader: nested table support

ahornyai:
I implemented support for nested table parsing in RTF Reader. The parser can process `nestrow` and `nestcell` now.


jgm:
thanks!

PR #11805, opened 2026-08-14
Update MANUAL.txt with missing cross reference

jidanni:
Add missing cross reference from --include-in-header to header-includes.

### DIFF
--- a/MANUAL.txt
+++ b/MANUAL.txt
@@ -1029,6 +1029,7 @@ header when requesting a document from a URL:
     CSS or JavaScript in HTML documents.  This option can be used
     repeatedly to include multiple files in the header.  They will be
     included in the order specified.  Implies `--standalone`.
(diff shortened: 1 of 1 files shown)


jidanni:
My motivation is this is very useful to know. For instance if there is no write permission on the file system, the only way to add a style would be via a variable. 

I was using temporary files, when all along I could have been using a variable. 

I think it would be useful for other users to know that one doesn't necessarily need to use a temporary file. 

Thanks.

jgm:
Maybe so, but if so, same should be done for the similar include-before and include-after.
And it would be good for these to be links to the subsection in which they are introduced.
(Automatic links would not be introduced in this case; that is just for command options.)

jidanni:
Thanks. Maybe someone who is more familiar with the manual can look into that.

PR #11818, opened 2026-08-18
Add --completion={bash,zsh,fish}

### LINKED ISSUE #8542: [feature] more shell completions
**Describe your proposed improvement and the problem it solves.**

Now `pandoc --bash-completion` can generate bash completion script. Why not provide more `--XXsh-completion`?
For example, zsh has a [completion script](https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_pandoc)
for pandoc, however, it is written manually:

```zsh
_arguments -s \
  {-f+,-r+,--from=,--read=}'[specify input format]: :_pandoc_format -T input' \
  {-t+,-w+,--to=,--write=}'[specify output format]: :_pandoc_format -T output' \
  {-o+,--output=}'[write output to FILE instead of stdout]:file:_files' \
  '--data-dir=[specify the user data directory to search for pandoc data files]:data directory:_files -/' \
  {-d+,--defaults=}'[read default from YAML file]: :_pandoc_defaults_file' \
  '--shift-heading-level-by=[shift heading levels by specified number]:positive or negative integer: ' \
  '!--base-header-level=:number [1]:(1 2 3 4 5)' \
  '!--strip-empty-paragraphs[deprecated. Use the +empty_paragraphs extension instead]' \
  '--indented-code-classes=[classes to use for indented code blocks]:class list (comma-separated)' \
  '--default-image-extension=[specify a default extension to use when image paths/URLs have no extension]:extension: ' \
  '--file-scope[parse each file individually before combining for multifile documents]' \
  ...
```

When the option of pandoc changed, the maintainers of zsh have to update it manually. If pandoc can provide `--zsh-completion` which use this file as a template but update automatically, it will be greater.

TIA!

Same issues:
- https://github.com/sharkdp/vivid/issues/86
- https://github.com/sharkdp/hexyl/issues/155
- https://github.com/rhysd/actionlint/issues/300
- https://github.com/mvdan/sh/issues/914

Freed-Wu:
Make --bash-completion as an alias of --bash-completion
Add short description for every command line option for zsh, fish
Add some data CompletionKind, CompletionShell for completion
Add a test for completion

Fix #8542


src/Text/Pandoc/App/Opt.hs
```diff
@@ -20,15 +20,23 @@ Options for pandoc when used as an app.
...
+          , CompletionKind(..)
+          , Description
+          , OptionSpec(..)
+          , toOptDescr
+          , option
           ) where
+import System.Console.GetOpt (OptDescr (..), ArgDescr (..))
 import Control.Monad.Except (throwError)
 import Control.Monad.Trans (MonadIO, liftIO, lift)
+import Control.Monad.Trans.Except (ExceptT)
```
jgm: `ExceptT` can be imported from `Control.Monad.Except` (above).
And then the `transformers` dependency can be removed from pandoc.cabal.

src/Text/Pandoc/App/Opt.hs
```diff
@@ -85,9 +93,63 @@ data IpynbOutput =
...
+  | Engines          -- ^ a PDF engine program
+  | Files            -- ^ a file path
+  | Fixed [String]   -- ^ one of a fixed set of values
+  deriving (Show)
+
+-- | A short, human-readable description for an option.  These are used
+-- for the zsh and fish completions; they are intentionally kept free of
+-- the characters that would need escaping in those shells
+-- (colons, square brackets and single quotes).
+type Description = Text
```
jgm: `type` is pretty useless, because it doesn't make a difference to the compiler ; I'd recommend either using a `newtype` here or just using `Text`.

jgm (commented):
This looks good at first look! Please fix compiler warnings and remove the transformer dependency as noted.  Also, the commit message should clearly describe any API changes, making them with `[API change]`.  This will include the new definition exported from Text.Pandoc.App, as well as any changes in the types from Text.Pandoc.Opt that are re-exported from Text.Pandoc.App.

src/Text/Pandoc/App.hs
```diff
@@ -25,6 +25,7 @@ module Text.Pandoc.App (
           , parseOptions
           , parseOptionsFromArgs
           , options
+          , toOptDescr
```
jgm: Does it make sense to export this `toOptDescr` in the public API?
It is only useful with the `OptionSpec` type, which isn't exported.
It seems to me that we don't need to export it -- and so probably shouldn't.

jgm (commented):
This is almost there. I made one comment about whether we can avoid exporting a new function from T.P.App.  Also, I think `--bash-completions` should be deprecated. See the examples of deprecated options in the CommandLineOptions module and the MANUAL.txt to see how this is done.

MANUAL.txt
```diff
@@ -431,12 +431,21 @@ header when requesting a document from a URL:
...
 
-:   Generate a bash completion script.  To enable bash completion
-    with pandoc, add this to your `.bashrc`:
+    The generated scripts are produced from pandoc's list of options
+    and do not require pandoc to be invoked while completing.
+
+`--bash-completion`
 
-        eval "$(pandoc --bash-completion)"
+:   *Deprecated.  Use `--completion`=bash instead.
```
jgm: Need a `*` at the end of the line to close the emphasis.
Also, should be `` `--completion=bash` ``.

src/Text/Pandoc/App/CommandLineOptions.hs
```diff
@@ -996,188 +1068,225 @@ options =
...
                   (\arg opt -> return opt{ optLogFile = Just $
                                             normalizePath arg })
                 "FILE")
-                "" -- "Log messages in JSON format to this file."
+                Files
+                (T.pack "Log messages in JSON format to this file")
 
-    , Option "" ["bash-completion"]
-                 (NoArg (\_ -> optInfo BashCompletion))
-                 "" -- "Print bash completion script"
```
jgm, line 1129: Leave the "bash-completion" clause in there, but use `deprecatedOption`. See example `--self-contained`.

jgm:
Thank you!

PR #11831, opened 2026-08-26
Dark mode support for HTML output

clarfonthey:
Originally jgm/djot#409, before I found out this was part of pandoc.

Essentially, the goal here is to add a "just works" dark mode support to pandoc's HTML generation while being minimally intrusive and not breaking custom schemes. It consists of the following changes to the styles.html template:

* `quotecolor` and `quotebordercolor` are added as these are the only two unique colors in the stylesheet that weren't already configurable. They affect the styling of `blockquote` tags.
* Each of the various `*color` variables for the HTML generation now has a `*colordark` friend, which controls the color in dark mode.
* Padding is emitted for `monobackgroundcolor` if `monobackgroundcolor` *or* `monobackgroundcolordark` is provided, although the provided color will only work in the respective mode. (it uses `inherit` otherwise)
* `color-scheme` is emitted to either allow both dark and light mode by default, or restrict to only light or dark mode depending on what variables the user defines. (see below)

Part of being minimally intrusive is only enabling light and dark mode in the following cases:

1. No colors have been overridden
2. The user has explicitly overridden colors for both schemes

However, because pandoc's template conditions can only check one variable at a time, only the `fontcolor` and `backgroundcolor` (and their `*dark` versions) variables is checked to determine whether the user has overrided the colors, and thus, which schemes they would like to be enabled. This means that, in general, if a user has changed `backgroundcolor` or `fontcolor` only, that means they will default to only light mode, and if a user has changed `backgroundcolordark` or `fontcolordark` only, it will default to only dark mode.

There are a few other caveats, but these have been added to `MANUAL.txt` and hopefully are clear.

----

Since the `light-dark` coloring is a "new" standard ([around 2024, 88% coverage](caniuse.com/wf-light-dark)), it includes explicit fallbacks for light mode. However, as part of implementing this change, `currentColor` was added to a few color definitions as opposed to repeating the font color, since this has been supported [since 2010 at 98% availability](https://caniuse.com/currentcolor).

That said, I have no idea what pandoc wishes to support browser-wise, and am only taking a conservative approach out of caution. I don't mind removing the fallbacks if you'd prefer.


MANUAL.txt
```diff
@@ -2937,17 +2937,24 @@ ODT or pptx.
...
+:   sets the CSS `color` property for `blockquote` elements.
+
 `monofont`
-:   sets the CSS `font-family` property on `code` elements.
+:   sets the CSS `font-family` property on `code` and `pre` elements.
 
-`monobackgroundcolor`
-:   sets the CSS `background-color` property on `code` elements
+`monobackgroundcolor`, `monobackgroundcolordark`
+:   sets the CSS `background-color` property on `code` and `pre` elements
```
clarfonthey, line 2957: Note: affecting `pre` also is not a new change, just documenting existing behavior.

data/templates/styles.html
```diff
@@ -12,8 +12,59 @@
...
 $endif$
-  color: $if(fontcolor)$$fontcolor$$else$#1a1a1a$endif$;
-  background-color: $if(backgroundcolor)$$backgroundcolor$$else$#fdfdfd$endif$;
+$if(fontcolor)$
+  color: $fontcolor$;
+  $if(fontcolordark)$
+    color: light-dark($fontcolor$, $fontcolordark$);
+  $else$
+    color: light-dark($fontcolor$, #fdfdfd);
+  $endif$
```
jgm: Could be
```
  color: light-dark($fontcolor$, $if(fontcolordark)$$fontcolordark$$else$#fdfdfd$endif$)
```

jgm: also it's best to keep the same indentation level inside the if, or the resulting CSS will have uneven indentation.

data/templates/styles.html
```diff
@@ -12,8 +12,59 @@
...
+$elseif(fontcolordark)$
+  color: #1a1a1a;
+  color: light-dark(#1a1a1a, $fontcolordark$);
+$else$
```
jgm: Instead of using all these complex nested ifs and elseifs, why not make the earlier clause something like
```
color: $if(fontcolor)$$fontcolor$$else$#1a1a1a;$endif$;
color: light-dark($if(fontcolor)$$fontcolor$$else$#1a1a1a$endif$, $if(fontcolordark)$$fontcolordark$$else$#fdfdfd$endif$);
```

data/templates/styles.html
```diff
@@ -12,8 +12,59 @@
...
+  $else$
+    background-color: light-dark($backgroundcolor$, #1a1a1a);
+  $endif$
+$elseif(backgroundcolordark)$
+  background-color: #fdfdfd;
+  background-color: light-dark(#fdfdfd, $backgroundcolordark$);
+$else$
+  background-color: #fdfdfd;
+  background-color: light-dark(#fdfdfd, #1a1a1a);
+$endif$
```
jgm: similar comments to above for color

data/templates/styles.html
```diff
@@ -12,8 +12,59 @@
...
+    color-scheme: light dark;
+  $else$
+    color-scheme: dark;
+  $endif$
+$elseif(backgroundcolor)$
+  color-scheme: light;
+$else$
+  color-scheme: light dark;
+$endif$
 }
```
jgm: Would be great if this could be simplified in a similar way to the one suggested above.

data/templates/styles.html
```diff
@@ -55,11 +107,20 @@
...
+    color: light-dark($linkcolor$, $linkcolordark$);
+  $else$
+    color: light-dark($linkcolor$, inherit);
+  $endif$
+$elseif(linkcolordark)$
+  color: inherit;
+  color: light-dark(inherit, $linkcolordark$);
+$else$
+  color: inherit;
+$endif$
```
jgm, line 72: similar

pandoc-cli/man/pandoc.1
```diff
@@ -531,17 +531,39 @@ See the section Defaults files for more information on the file format.
...
-\f[CR]\-\-bash\-completion\f[R]
-Generate a bash completion script.
-To enable bash completion with pandoc, add this to your
-\f[CR].bashrc\f[R]:
+\f[CR]\-\-completion=\f[R]\f[I]SHELL\f[R]
+Generate a shell completion script for the given shell, one of
+\f[CR]bash\f[R], \f[CR]zsh\f[R], or \f[CR]fish\f[R].
+To enable completion with pandoc, evaluate the output of this command in
+your shell\(cqs startup file.
+For example, for bash, add this to your \f[CR].bashrc\f[R]:
```
jgm: Don't regenerate man pages at this point. I'll do that right before release.

jgm (commented):
I made some suggestions on simplifying the CSS template.  (Not exhaustive but you should get the idea.) As noted, leave the man pages unchanged for now. As for browser support, if the CSS can be simpler I'd accept 98% support for post-2010 browsers, I think.

clarfonthey:
Okay, removed the man page changes and updated the template per your suggestions. In some cases this is definitely clearer, in others it looks a bit messier to me, but it at least matches your suggestions. The only weird case is with the `monobackgroundcolor` and `linkcolor` cases, which I think is the style you'd prefer, but let me know if that's not the case.

data/templates/styles.html
```diff
@@ -13,7 +13,10 @@
   line-height: $linestretch$;
 $endif$
   color: $if(fontcolor)$$fontcolor$$else$#1a1a1a$endif$;
-  background-color: $if(backgroundcolor)$$backgroundcolor$$else$#fdfdfd$endif$;
+  color: light-dark($if(fontcolor)$$fontcolor$$else$#1a1a1a$endif$, $if(fontcolordark)$$fontcolordark$$else$#fdfdfd$endif$);
+  background-color: $if(backgroundcolor)$$backgroundcolor$$else$#1a1a1a$endif$;
+  background-color: light-dark($if(backgroundcolor)$$backgroundcolor$$else$#1a1a1a$endif$, $if(backgroundcolordark)$$backgroundcolordark$$else$#fdfdfd$endif$);
+  color-scheme:$if(fontcolor)$ light$elseif(backgroundcolor)$ light$elseif(fontcolordark)$$elseif(backgroundcolordark)$$else$ light$endif$$if(fontcolordark)$ dark$elseif(backgroundcolordark)$ dark$elseif(fontcolor)$$elseif(backgroundcolor)$$else$ dark$endif$;
```
clarfonthey, line 19: The way the logic works out, this should always output at least one of the two.

data/templates/styles.html
```diff
@@ -55,11 +59,13 @@
...
-}
-a:visited {
-  color: $if(linkcolor)$$linkcolor$$else$#1a1a1a$endif$;
+a, a:visited {
+  color: $if(linkcolor)$$linkcolor$$else$inherit$endif$;
+$if(linkcolor)$
+  color: light-dark($linkcolor$, $if(linkcolordark)$$linkcolordark$$else$inherit$endif$);
+$elseif(linkcolordark)$
+  color: light-dark(inherit, $linkcolordark$);
+$endif$
```
clarfonthey, line 72: These avoid outputting a redundant `light-dark(inherit, inherit)`.

MANUAL.txt
```diff
@@ -2985,6 +2992,40 @@ To override or extend some [CSS] for just one document, include for example:
...
+
+And if only one "mode" is provided (light or dark), `color-scheme` is set so
+that this chosen mode is used always, regardless of the system's preference for
+light or dark mode.
+
+Note that there are a few caveats to this, namely that:
+
+* If the browser does not support `light-dark` (older than ~2024), then it will
+  always choose light mode, regardless of this override.
+* When printing, light mode is always chosen.
```
jgm: Maybe this is misleading.  We have in print css:
```
  html {
    background-color: $if(backgroundcolor)$$backgroundcolor$$else$white$endif$;
  }
  body {
    background-color: transparent;
    color: black;
  }
```
which means that `black` on transparent is always used regardless of `color`.  So, it's not that light mode is chosen; rather, black-on-transparent is chosen.

I'm actually not sure why `background-color` is still defined in HTML, since it's overridden in `body`: is there a point to this?

clarfonthey: Yeah, I definitely think that we shouldn't be overriding `background-color` for print mode at all, but technically, it was, so, I kept it. In this case, `html`'s background color still shows through *because* of the choice of `transparent` for `body`.

MANUAL.txt
```diff
@@ -2985,6 +2992,40 @@ To override or extend some [CSS] for just one document, include for example:
...
+
+In general, if you wish to override the theming for a single scheme, you should
+choose to override either the normal `*color` variables (light mode) or the
+`*colordark` variables (dark mode) to ensure that the styles do not conflict
+with other styles added by new versions of Pandoc or the browser itself.
+
```
jgm, line 3029: Probably we need a comment here about how pandoc's syntax highlighting will not react to light/dark mode. (That would require changes in the CSS produced by jgm/skylighting; I think there may be an issue there.)

clarfonthey: (This one is easy to mention, so, fixed.)

MANUAL.txt
jgm: The rationale for setting the print background-color to transparent was that, in the normal case where you print a web page, you generally just want a white paper background; you don't want the printer to make the whole background slightly darker just because the screen is that way.  But if the html background color is showing through, then it's not working as intended.  Why do we set background color on html at all?  Shouldn't all visible elements be in body?

clarfonthey: Decided to take a stronger approach: the `@media print` CSS now explicitly overrides all colors to black and sets all background colors to transparent where they might be overridden. It also resets padding on `code` and `pre` since those are dependent on color variables.

It currently ignores syntax highlighting, which as discussed is probably a separate issue.

jgm:
thanks!

PR #11836, opened 2026-08-29
feat: Case-insensitive alert keywords, basic styling

nathanlesage:
## Description

This commit changes the emitted classes from the alert extension of the basic Markdown reader to add the class `alert` to the produced DIVs to allow for namespacing. In addition, it adds some basic styling to make the alerts visible in HTML exports. This PR also adds the correct selectors to enable this highlighting for alerts produced by the CommonMark alerts extension.

## Rationale

While writing an [Admonition parser for Zettlr](https://github.com/Zettlr/Zettlr/pull/6497), I realized that GitHub allows the alert keywords to be lowercase, too, but the Markdown alerts extension of Pandoc did not. This PR aims to fix that so that the alerts extension aligns with the behavior observable on GitHub.

In addition, when exporting a Markdown document containing alert syntax to HTML, those alerts were essentially indistinguishable from surrounding blocks due to a lack of CSS. I know that unlike with blockquotes, adding some styling is an opinionated choice. However, since I added the styles to the standard HTML stylesheet, users can easily override everything. If you'd like to have the layout more aligned with how GitHub does it, I'll happily change the CSS accordingly such that the alerts render exactly like on GitHub, but I personally (!) prefer the way they look right now. (Disclaimer: Since I'm bad at design, and it took me a while to figure out good colors for the Zettlr implementation, I have taken the liberty to use these colors, but again, I'm all in for changes.)

## Backwards compatibility

This PR should be fully backwards compatible, meaning that all assumptions users may already have baked into their own pipelines should still work out. This merely loosens up the strictness of the parser to allow for additional styles (read: lowercase). Since I also refrained from changing the existing class names, existing CSS overrides should still work well.

## Miscellaneous

I have also read and acknowledge John's comment elsewhere that additional syntactical elements such as custom titles aren't in scope since GitHub doesn't support them. However, given that a lot of other software support this syntax (to name a few important ones: Zettlr, VuePress and VitePress, Obsidian), I'm wondering if this shouldn't be considered in the future…? (I agree that this can be an issue, but I didn't want to open one given your relatively fresh comment that you do not consider this.)

**Question**

I have seen that the corresponding CommonMark alert extension uses a slightly different class structure, producing `alert alert-note` as well as `alert-title` instead of `note` and `title`. Would it be welcome to change the CommonMark classes to align with this extension (which, from a CSS perspective, makes more sense), or vice versa? In any case, I'd assume it makes sense to harmonize the emitted structure from both readers, since changing `markdown+alerts` to `commonmark+alerts` should ultimately produce the same HTML (or other export syntax, for that matter).

### DIFF
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -91,41 +91,6 @@ blockquote {
   border-left: 2px solid #e6e6e6;
   color: #606060;
 }
-/* Alert styling (when +alerts is active) */
-div.alert {
-  padding: 0.5em 1em;
-  border-radius: 0.25em;
-  margin: 1em 0;
-}
-
-div.alert div.title p, div.alert div.alert-title p {
-  font-weight: bold;
-}
-
-div.alert.note, div.alert.alert-note {
-  color: rgb(125, 125, 125);
-  background-color: rgb(220, 220, 220);
-}
-
-div.alert.tip, div.alert.alert-tip {
-  color: rgb(75, 134, 75);
-  background-color: rgb(200, 250, 200);
-}
-
-div.alert.important, div.alert.alert-important {
-  color: rgb(133, 87, 133);
-  background-color: rgb(250, 180, 250);
-}
-
-div.alert.warning, div.alert.alert-warning {
-  color: rgb(130, 130, 48);
-  background-color: rgb(250, 250, 180);
-}
-
-div.alert.caution, div.alert.alert-caution {
-  color: rgb(148, 76, 76);
-  background-color: rgb(250, 210, 210);
-}
 $if(abstract)$
 div.abstract {
   margin: 2em 2em 2em 2em;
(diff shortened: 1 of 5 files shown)


jgm:
Would it be possible to split this into two PRs?  Or at least two commits?  One for the parser change (accepting lowercase in admonitions), and one for the CSS changes?

nathanlesage:
Good point, I just removed the style changes and pushed them to a separate branch, to which I'll open a second PR momentarily. Sorry!

Also, because I'm a doofus, the other PR now includes the parser changes. I'd assume we can first merge this PR (assuming the code review works out) before tending to the other one since I can then just update that branch accordingly… today is not my lucky git day, I assume.

jgm:
Once this is merged, you should be able to rebase the other one and force push it to get a clean commit.

PR #11837, opened 2026-08-29
feat: Add basic styling to Alerts in the HTML exporter

nathanlesage:
This PR contains the style changes I pulled out of #11836. See that PR for context.

EDIT: I just saw that this thing also includes the two changes from the other branch because I drafted this branch from the other one. I suspect we can first merge #11836 and then I can update this branch from main and get rid of the changes to the Markdown parser.

### DIFF
--- a/data/templates/styles.html
+++ b/data/templates/styles.html
@@ -91,6 +91,41 @@ blockquote {
   border-left: 2px solid #e6e6e6;
   color: #606060;
 }
+/* Alert styling (when +alerts is active) */
+div.alert {
+  padding: 0.5em 1em;
+  border-radius: 0.25em;
+  margin: 1em 0;
+}
+
+div.alert div.title p, div.alert div.alert-title p {
+  font-weight: bold;
+}
+
+div.alert.note, div.alert.alert-note {
+  color: rgb(125, 125, 125);
+  background-color: rgb(220, 220, 220);
+}
+
+div.alert.tip, div.alert.alert-tip {
+  color: rgb(75, 134, 75);
+  background-color: rgb(200, 250, 200);
+}
+
+div.alert.important, div.alert.alert-important {
+  color: rgb(133, 87, 133);
+  background-color: rgb(250, 180, 250);
+}
+
+div.alert.warning, div.alert.alert-warning {
+  color: rgb(130, 130, 48);
+  background-color: rgb(250, 250, 180);
+}
+
+div.alert.caution, div.alert.alert-caution {
+  color: rgb(148, 76, 76);
+  background-color: rgb(250, 210, 210);
+}
 $if(abstract)$
 div.abstract {
   margin: 2em 2em 2em 2em;
--- a/test/writer.html4
+++ b/test/writer.html4
@@ -92,6 +92,41 @@
       border-left: 2px solid #e6e6e6;
       color: #606060;
     }
+    /* Alert styling (when +alerts is active) */
+    div.alert {
+      padding: 0.5em 1em;
+      border-radius: 0.25em;
+      margin: 1em 0;
+    }
+
+    div.alert div.title p, div.alert div.alert-title p {
+      font-weight: bold;
+    }
+
+    div.alert.note, div.alert.alert-note {
+      color: rgb(125, 125, 125);
+      background-color: rgb(220, 220, 220);
+    }
+
+    div.alert.tip, div.alert.alert-tip {
+      color: rgb(75, 134, 75);
+      background-color: rgb(200, 250, 200);
+    }
+
+    div.alert.important, div.alert.alert-important {
+      color: rgb(133, 87, 133);
+      background-color: rgb(250, 180, 250);
+    }
+
+    div.alert.warning, div.alert.alert-warning {
+      color: rgb(130, 130, 48);
+      background-color: rgb(250, 250, 180);
+    }
+
+    div.alert.caution, div.alert.alert-caution {
+      color: rgb(148, 76, 76);
+      background-color: rgb(250, 210, 210);
+    }
     code {
       white-space: pre-wrap;
       font-family: Menlo, Monaco, Consolas, 'Lucida Console', monospace;
--- a/test/writer.html5
+++ b/test/writer.html5
@@ -92,6 +92,41 @@
       border-left: 2px solid #e6e6e6;
       color: #606060;
     }
+    /* Alert styling (when +alerts is active) */
+    div.alert {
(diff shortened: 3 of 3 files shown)


nathanlesage:
It seems to have worked — the only changes currently showing in the diff are the three HTML files. Sorry again, and thanks for your help @jgm.

jgm:
If you rebase against main and force push, you'll get a clean commit with just the style changes.

nathanlesage:
So, that took a while, which is quite embarassing, but at least now I know what to do in such cases in the future… the branch is now clean and ready for review. Sorry about the hiccups.

jgm:
Can you paste a screenshot that shows how these look with this CSS?

Also, if we do this, it might make sense to gate the inclusion of the CSS for alerts under a template variable which would be set automatically by the writer if there are actually alerts present. This keeps the size down for other cases.

My main reservation so far is that we don't do automatic styling of these in other formats; our thought was that there are too many ways to do it, so we wanted to leave it to users to handle with filters etc. However, in HTML everything can be done with CSS, so perhaps this is an exception.

nathanlesage:
> Can you paste a screenshot that shows how these look with this CSS?

Sure, here you go:

image

> Also, if we do this, it might make sense to gate the inclusion of the CSS for alerts under a template variable which would be set automatically by the writer if there are actually alerts present. This keeps the size down for other cases.

I had the same thought, but couldn't find a variable that is being set in the metadata when there are alerts in the file. Is there one? If not, should I see if I can add one in the Markdown reader?

> My main reservation so far is that we don't do automatic styling of these in other formats; our thought was that there are too many ways to do it, so we wanted to leave it to users to handle with filters etc. However, in HTML everything can be done with CSS, so perhaps this is an exception.

I absolutely agree that styling, especially in this case, is very opiniated and people may disagree on the styles. However, I decided to propose this styling for two reasons: (a) if someone actually uses alerts, they would currently not actually see *any* alerts because they are not at all styled. I would argue that it is more user-friendly to just provide some default styles but make it simple to change them. I figure that many Pandoc users want just some working default styles, especially now that Pandoc is established in a lot of tools where people don't necessarily want to or know how to customize exports. And (b) that's why it's very easy with Pandoc HTML templates to override these styles if people prefer, so I fathom it wouldn't be detrimental to anyone who dislikes these defaults.

***

Regarding other export profiles: I was actually considering adding something like that for at least PDF and Word exports, too, but at least for PDFs that would necessitate a change to the writer, since it would require custom LaTeX code to be emitted. Again, I do think that if some syntax is detected and properly parsed, it seems odd if the resulting code is not at all styled, even if it can be overwritten.

jgm:
> I had the same thought, but couldn't find a variable that is being set in the metadata when there are alerts in the file. Is there one? If not, should I see if I can add one in the Markdown reader?

We would have to add code to the HTML writer that sets this variable if the document contains admonitions.  Probably we'd need to add a field in the writer state that tracks this, and set it when we hit admonitions.  Then in the main function we'd check this state variable and set the template variable accordingly. There are probably already some examples of things like this in the HTML writer.

jgm:
I wonder if it would make sense to style them more like the GitHub default:

image

I believe the icons could even be inserted via CSS.

nathanlesage:
I can certainly take a look; I found that a bit too close to the blockquote styling, but I can absolutely see the reasoning. Regarding the icons, I can do some research, but as long as the obvious candidates didn't change in the past years, I think it may be tricky without having to import a font library (which I'd argue might be a very large dependency for a small effect).

I'll get to it as soon as possible, albeit I'm currently very maxed out in terms of availability due to switching affiliations.

jgm:
You may be right, and I'm open to being convinced that the icon is too much.
GitHub does this:
```
<svg data-component="Octicon" class="octicon octicon-light-bulb mr-2" viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true"><path d="M8 1.5c-2.363 0-4 1.69-4 3.75 0 .984.424 1.625.984 2.304l.214.253c.223.264.47.556.673.848.284.411.537.896.621 1.49a.75.75 0 0 1-1.484.211c-.04-.282-.163-.547-.37-.847a8.456 8.456 0 0 0-.542-.68c-.084-.1-.173-.205-.268-.32C3.201 7.75 2.5 6.766 2.5 5.25 2.5 2.31 4.863 0 8 0s5.5 2.31 5.5 5.25c0 1.516-.701 2.5-1.328 3.259-.095.115-.184.22-.268.319-.207.245-.383.453-.541.681-.208.3-.33.565-.37.847a.751.751 0 0 1-1.485-.212c.084-.593.337-1.078.621-1.489.203-.292.45-.584.673-.848.075-.088.147-.173.213-.253.561-.679.985-1.32.985-2.304 0-2.06-1.637-3.75-4-3.75ZM5.75 12h4.5a.75.75 0 0 1 0 1.5h-4.5a.75.75 0 0 1 0-1.5ZM6 15.25a.75.75 0 0 1 .75-.75h2.5a.75.75 0 0 1 0 1.5h-2.5a.75.75 0 0 1-.75-.75Z"></path></svg>
```
So, that's not doable in pure CSS.

Maybe we could get a similar look but without the icons.

nathanlesage:
There is another route: I've seen that the common mark alerts extension actually inserts SVG icons into the DOM when exporting to HTML. Is there a reason that the common mark alerts extension differs from the markdown alerts extension? If not, it might make sense to just port the code from the common mark extension into the markdown reader. This would also align the output of these two readers and, coincidentally, make styling easier.

Furthermore, it's trivial to hide DOM nodes, so if someone doesn't want the icon, it's easy to hide it (and since the icons are fairly simple, they don't take up too much space in any export).

I had already contemplated just trying out what would happen, but I'm still not versatile enough in Haskell to do that.

jgm:
? I don't think pandoc's commonmark extension does that.
```
% pandoc -f commonmark+alerts
> [!TIP]
> a tip
<div class="tip">
<div class="title">
<p>Tip</p>
</div>
<p>a tip</p>
</div>
```

jgm:
However, there was an early version of the commonmark extension that did add svgs. In the end, I added the more minimal version. Here's the early version:
https://github.com/jgm/commonmark-hs/commit/1a20e75c84a39230839d415b20f3ce3a33ca3226#diff-6271cb6585eddcbe1857b2896e74bd2825dc5e0741f431b36eb18af9553a105fR101
It could be worth reconsidering.

jgm:
I think my thinking was that people would have different opinions about how these things should be styled, including which icons to use.

PR #11846, opened 2026-09-05
feat(typst): map `form: "prose"` citations to AuthorInText.

sghng:
Typst's #cite supports a form parameter: "normal", "prose", "full", "author", and "year". The reader previously only special-cased
 "year" (→ SuppressAuthor); everything else became NormalCitation.

 Minimal repro:

 ```typst
   #cite(<doe2020>, form: "prose")
   #bibliography("refs.bib")
 ```

 With --citeproc, this rendered (Doe and Smith 2020); Typst renders narrative form Doe and Smith (2020), which corresponds to pandoc's
 AuthorInText.

 This PR maps form: "prose" → AuthorInText. The author and full forms have no pandoc citation-mode equivalent, so they intentionally
 still fall back to NormalCitation (noted in a code comment); proper support for those would need a design discussion (e.g. author-only
 citations aren't expressible in pandoc's AST today).

 Adds golden test coverage for all four forms in test/typst-reader.typ / .native.

### DIFF
--- a/src/Text/Pandoc/Readers/Typst.hs
+++ b/src/Text/Pandoc/Readers/Typst.hs
@@ -468,7 +468,10 @@ inlineHandlers = M.fromList
                 B.citationPrefix = mempty,
                 B.citationSuffix = mempty,
                 B.citationMode = case form of
+                                    "prose" -> B.AuthorInText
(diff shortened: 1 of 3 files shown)


jgm:
thanks!

PR #11850, opened 2026-09-08
Typst: preserve labels on no-figure tables

### LINKED ISSUE #11849: Typst writer drops table labels when typst:no-figure is set
The `typst:no-figure` class requests a table without a Typst `figure`
wrapper. Adding this class also causes the table's label to disappear,
even though its identifier is unchanged. An explicitly supplied
`typst-label` is lost as well.

## Normal table output

Start with this Markdown document, `table.md`:

```markdown
| A | B |
|---|---|
| 1 | 2 |

: Example {#table-id}
```

The `table_attributes` extension attaches `#table-id` to the table.
Convert the document with:

```sh
pandoc -f markdown+table_attributes -t typst table.md
```

Pandoc emits a figure-wrapped table followed by the label `<table-id>`:

```typst
#figure(
  align(center)[#table(
    columns: 2,
    align: (auto,auto,),
    table.header([A], [B],),
    table.hline(),
    [1], [2],
  )]
  , caption: [Example]
  , kind: table
  )
<table-id>
```

## Adding typst:no-figure

Now change only the caption's attributes, keeping the same identifier:

```markdown
: Example {#table-id .typst:no-figure}
```

Running the same command produces:

```typst
#table(
  columns: 2,
  align: (auto,auto,),
  table.header([A], [B],),
  table.hline(),
  [1], [2],
)
```

The figure wrapper has been removed, but `<table-id>` has also
disappeared. I would expect the table to retain its label:

```typst
#table(
  columns: 2,
  align: (auto,auto,),
  table.header([A], [B],),
  table.hline(),
  [1], [2],
)
<table-id>
```

## Why this appears to be a bug

The table still has the identifier `table-id` in both inputs. Running
`pandoc -f markdown+table_attributes -t native table.md` after adding the
class shows `( "table-id" , [ "typst:no-figure" ] , [] )` on the `Table`.
The identifier survives parsing; it is omitted from the Typst output
when the writer takes the no-figure path.

The class controls whether the table has a figure wrapper. The existing
identifier-to-label behavior should therefore continue to apply to the
table emitted without that wrapper.

An explicit `typst-label` makes the inconsistency clearer. Change the
caption to:

```markdown
: Example {#table-id .typst:no-figure typst-label="table-label"}
```

The [Typst output documentation] states:

> The attribute `typst-label` will produce a typst label (overriding
> the identifier attribute if present).

This explicitly requests `<table-label>`, but the same bare table is
emitted with no label at all. Removing `.typst:no-figure` makes
`<table-label>` appear again. Thus an existing table attribute is being
ignored depending on whether the figure wrapper is enabled.

In `src/Text/Pandoc/Writers/Typst.hs`, the table writer computes `lab`
before checking `typst:no-figure`. The figure branch emits it, while the
no-figure branch returns the table without the already computed label.

[Typst output documentation]: https://github.com/jgm/pandoc/blob/8813d0f984d6b6984b8e099222f5dc837d1cf679/doc/typst-property-output.md#typst-labels

## Version

Reproduced with a source build of `pandoc 3.11` at revision
`8813d0f984d6b6984b8e099222f5dc837d1cf679` (upstream `main`).
OS: Ubuntu 22.04.1 LTS, x86_64.

zenor0:
Fixes #11849.

Tables with `typst:no-figure` currently lose the label derived from
their identifier or explicit `typst-label` attribute. Preserve that
label when emitting a table without a figure wrapper.

## Change

The table branch in `src/Text/Pandoc/Writers/Typst.hs` already computes
`lab` from `typst-label`, or from the table identifier when no explicit
label is supplied. The figure-wrapped branch appends `lab`, but the
`typst:no-figure` branch returns only the table output and discards it.

Append that existing label to the result of `toTypstBracesSetText`:

```haskell
then toTypstBracesSetText typstTextAttrs table $$ lab
```

This places the label after `#table(...)`, or after the complete
`#{set text(...);table(...)}` block when text properties require a
scope. In the latter case, the label attaches to the scoped content
containing the table, and the text settings remain local to that block.
The existing `toLabel` function continues to handle label syntax,
including labels with spaces that require `#label(...)`.

## Regression coverage

Add three command tests covering:

1. A table identifier preserved with `typst:no-figure`.
2. An explicit `typst-label` taking precedence over the identifier.
3. A label containing spaces emitted after scoped text properties,
   followed by another paragraph.

All three fail before the fix solely because their label lines are
missing, and pass after it. The patch changes one writer line and adds
one 50-line command-test file.

## Validation

After the fix, all 43 Typst tests and the full suites (4163 Pandoc tests
and 48 Lua engine tests) pass. Checks of the generated output with
Typst 0.14.0 confirm the labels and preserved text scope.
HLint reports no hints for the modified writer.

The source build succeeds with `-Werror`. Local warnings about
dynamic-symbol metadata also occur with the unmodified writer using
the system C toolchain (GCC 11.4.0 and GNU ld 2.38).

### DIFF
--- a/src/Text/Pandoc/Writers/Typst.hs
+++ b/src/Text/Pandoc/Writers/Typst.hs
@@ -364,7 +364,7 @@ blockToTypst block =
             )
             $$ ")"
       return $ if "typst:no-figure" `elem` tabclasses
-        then toTypstBracesSetText typstTextAttrs table
+        then toTypstBracesSetText typstTextAttrs table $$ lab
         else "#figure("
             $$
             nest 2
--- a/test/command/typst-table-labels.md
+++ b/test/command/typst-table-labels.md
@@ -1,0 +1,50 @@
+Identifiers are preserved on tables without a figure wrapper.
+
+```
+% pandoc -f html -t typst
+<table id="table-id" class="typst:no-figure">
+  <tr><td>A</td><td>B</td></tr>
+</table>
+^D
+#table(
+  columns: 2,
+  align: (auto,auto,),
+  [A], [B],
+)
+<table-id>
+```
+
+An explicit Typst label takes precedence over the identifier.
+
+```
+% pandoc -f html -t typst
+<table id="table-id" class="typst:no-figure" typst-label="table-label">
+  <tr><td>A</td><td>B</td></tr>
+</table>
+^D
+#table(
+  columns: 2,
+  align: (auto,auto,),
+  [A], [B],
+)
+<table-label>
+```
+
+Labels follow scoped text properties.
+
+```
+% pandoc -f html -t typst
+<table class="typst:no-figure" typst-label="table label"
+       typst:text:size="3em">
+  <tr><td>A</td><td>B</td></tr>
+</table>
+<p>Paragraph after.</p>
+^D
+#{set text(size: 3em);table(
+  columns: 2,
+  align: (auto,auto,),
+  [A], [B],
+)}
+#label("table label")
+Paragraph after.
+```


jgm:
I took a slightly different approach, so closing this.

Issue #11864, opened 2026-09-11
Docx writer: Para blocks in table cells get "Body Text" spacing (9pt before/after), bloating table rows


sghng:
Pandoc version: 3.x (current main)

Steps to reproduce:

```
echo '#table(columns: 2, [a], [b], [c], [d])' > t.typ
pandoc t.typ -o t.docx
```

A bloated table looks like this:

<img width="386" height="516" alt="Image" src="https://github.com/user-attachments/assets/581d01c5-e3ad-4d52-9998-3169fce00b99" />

Also reproducible with any input where table cells contain `Para` rather than Plain blocks (e.g. multi-paragraph grid table cells).

Expected behavior: paragraphs in table cells render with compact spacing, matching cells that contain Plain blocks.

Actual behavior: every paragraph in a cell is styled "Body Text", whose spacing of 180 twips before and 180 twips after (9pt each) adds about 18pt of vertical padding per row. Table rows appear roughly double-spaced and bloated.

Cause: in `Text.Pandoc/Writers/Docx/OpenXML.hs`, the `Para` branch of `blockToOpenXML` unconditionally applies the "Body Text" style with no check of `stInTable`. The Plain branch just above it checks `isInTable`/`isInList` and applies
 "Compact", and empty cells also use "Compact" (see changelog: "Use Compact style for empty table cells (#1353)"). The Para-in-table case was never given the same treatment. The Typst reader always produces Para in table cells
 (`parseTable` uses `pWithContents` `pBlocks`, and `pBlock` yields paragraphs), so every Typst-to-docx table conversion triggers this.

Suggested fix: either wrap the `blocksToOpenXML` contents call for non-empty cells in `withParaPropM` (`pStyleM` "Compact") in `Text.Pandoc/Writers/Docx/Table.hs` (`ooxmlCellToOpenXML`), or extend the `Para` branch in `OpenXML.hs` to apply
 "Compact" when `stInTable` is set, mirroring the existing Plain handling.

Workaround: none without side effects. Setting "Body Text" spacing to 0 in a reference.docx fixes the tables but also removes spacing from body paragraphs.

jgm:
We don't always want compact style in tables, I think.  (Especially for cells with multiple paragraphs or other block-level content.)

If you want this, a solution is to use a Lua filter that changes Para to Plain inside tables.

sghng:
@jgm Fair point on multi-paragraph cells. Compact everywhere would flatten those,
[...]

PR #11866, opened 2026-09-11
Typst reader: emit Plain for single-paragraph table cells

sghng:
## Summary

Table cells consisting of a single paragraph are now emitted as `Plain` instead of `Para` by the Typst reader, per cell. This matches the convention of every other table-capable reader and fixes the bloated docx spacing reported in #11864.

Partially addresses #11864.

## Problem

The Typst reader always wrapped cell content in `Para`, no matter how simple the cell. Downstream, the docx writer maps `Plain` in table cells to the Compact paragraph style (1.8pt before/after) but `Para` to Body Text (9pt before/after). As a result every table converted from Typst rendered with bloated rows in docx, and a cell's styling depended on an encoding artifact rather than its content.

Every other reader normalizes single-paragraph cells to `Plain`, per cell:

| Reader | mechanism |
|---|---|
| Markdown pipe tables | `B.plain` in `pipeTableCell` |
| Markdown / RST grid tables | `plainify` in `Text.Pandoc.Parsing.GridTable` |
| LaTeX | `plainify` in `LaTeX/Table.hs` |
| Org | `B.plain` per cell |
| HTML | reflects markup: bare `<td>` content is `Plain`, `<p>` is `Para` |
| docx | `singleParaToPlain` per cell |

The Typst reader was the only outlier.

## Fix

New `plainifyCell` helper applied in `addCell` in `parseTable`, converting `[Para ils]` to `[Plain ils]`. Covers header, body, and footer cells. Cells with more than one block are untouched and keep `Para`, preserving Body Text separation between paragraphs in docx output.

The criterion is per cell rather than per table, following the established convention: mixed `Plain`/`Para` tables already arise from grid tables, LaTeX, HTML, and docx round trips, so writers must handle them regardless.

## Tests

- Regenerated `test/typst-reader.native`; the diffs are exactly the expected `Para` to `Plain` conversions inside table cells.
- All 40 Typst reader/writer tests pass.
- End to end: the table from #11864 now produces Compact-styled cells in docx output.

## Notes

- This supersedes the earlier writer-side version of this PR, which applied Compact to `Para` cells in the docx writer.
- #11867 (docx writer drops `pStyle` when cell alignment is set) is a separate writer-side issue and remains open.
- The branch name predates the pivot; the fix is now entirely in the Typst reader.

### DIFF
--- a/src/Text/Pandoc/Readers/Typst.hs
+++ b/src/Text/Pandoc/Readers/Typst.hs
@@ -732,7 +732,7 @@ parseTable mbident fields = do
   let toCell tableSection tableData contents = do
         case contents of
           [Elt (Identifier "grid.cell") _pos fs] -> do
-            bs <- B.toList <$> (getField "body" fs >>= pWithContents pBlocks)
+            bs <- B.toList . plainify <$> (getField "body" fs >>= pWithContents pBlocks)
             rowspan <- getField "rowspan" fs <|> pure 1
             colspan <- getField "colspan" fs <|> pure 1
             align' <- (toAlign <$> getField "align" fs) <|> pure B.AlignDefault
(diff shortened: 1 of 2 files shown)


jgm:
I'm handling this in a different way, so closing.

PR #11877, opened 2026-09-18
ANSI writer: fix missing bar on blockquote's first line

### LINKED ISSUE #11804: 'ansi' can omit vertical bar for blockquote
I have a Markdown file named `fruit.md`:

```md
1.  > Rhubarb rhubarb rhubarb
    > rhubarb rhubarb rhubarb
    > rhubarb rhubarb rhubarb
    > rhubarb rhubarb rhubarb
    > rhubarb rhubarb rhubarb
    > rhubarb rhubarb rhubarb.
    > Pineapple pineapple pineapple
    > pineapple pineapple pineapple
    > pineapple pineapple pineapple
    > pineapple pineapple pineapple
    > pineapple pineapple pineapple
    > pineapple pineapple pineapple
    > pineapple pineapple pineapple
    > pineapple pineapple pineapple
    > pineapple pineapple pineapple.

    Banana banana banana.
```

Using pandoc 3.10.2 on Linux, I ran this command:

```bash
pandoc -t ansi fruit.md
```

Every line of the blockquote should begin with a vertical bar, but pandoc rendered the first line without the bar:

Image

The terminal was 145 columns wide. I mention that because of [this recent comment](https://github.com/jgm/pandoc/issues/11775#issuecomment-5096182715) on another bug report.

jadhavgaurav:
Fixes #11804

`pandoc -t ansi` dropped the leading vertical bar on the first line
of a blockquote whenever that line shared a row with preceding
content, most commonly when the blockquote is the first block of a
list item (the item's marker and the quote's content are joined onto
the same physical line). Later, wrapped lines of the same blockquote
rendered the bar correctly.

Root cause: `Text.DocLayout.prefixed` (used by the ANSI writer's
`BlockQuote` handler via `D.prefixed "| "`) only emits its prefix on
a line that begins at column 0. When the quote starts mid-line, its
first physical line never reaches that check, so the prefix is
silently skipped for that one line - this matches doclayout's own
documented behavior for `prefixed`, it's just visibly wrong for a
blockquote's bar.

Fix: force the blockquote onto its own line with `D.cr` before
applying the prefix, in `src/Text/Pandoc/Writers/ANSI.hs`. `D.cr` is
a no-op when rendering is already at the start of a line, so
top-level and already-nested blockquotes are unaffected.

Added a regression case to `test/ansi-test.txt` /
`test/ansi-test.ansi` (a numbered list item whose first block is a
blockquote) and confirmed it fails against the unfixed source and
passes with the fix.

## Testing

- `cabal test test-pandoc`: 4188/4189 tests pass. The one failure
  (`test/command/7099.md`) is `pandoc was compiled without HTTP
  support`, an environment/build-flag issue unrelated to this
  change - reproduces identically on the unmodified source.
- `hlint src/Text/Pandoc/Writers/ANSI.hs --hint .hlint.yaml`: no
  hints.
- Manually reproduced the issue's exact repro from #11804 with
  `pandoc -t ansi --columns=145`, confirmed the bar is present on
  every line of the blockquote after the fix and missing on the
  first line without it (verified by reverting only the source
  change and rebuilding).


jgm:
thanks!

Issue #10747, opened 2025-04-01
Support highlighting via `.mark` for Typst


cwickham:
It would be nice if for Typst output the markdown:
```markdown
[Mark]{.mark}
```

was translated to 
```typst
#highlight[Mark]
```

Via the Typst [`highlight()` function ](https://typst.app/docs/reference/text/highlight/)

Issue #11017, opened 2025-08-03
Typst input with heading show rules that apply text processing functions directly fails


smheidrich:
**Explain the problem.**
Minimal reproducible example (boiled down from the ["smallcaps headings" example](https://typst.app/docs/reference/text/smallcaps/#smallcaps-headings) in Typst's official docs) showing that `heading` [show rules](https://typst.app/docs/reference/styling/#show-rules) given text processing functions directly don't work in Pandoc:

**`bug.typ`**:

```typst
#show heading: smallcaps

= Introduction
```

Running e.g. `pandoc -o bug.md bug.typ` on this yields:

```
unexpected Elt {eltName = Identifier "heading", eltPos = Just "bug.typ" (line 1, column 2), eltFields = fromList [(Identifier "body",VContent (fromList [Elt {eltName = Identifier "text", eltPos = Just "bug.typ" (line 1, column 2), eltFields = fromList [(Identifier "body",VContent (fromList [Txt "Introduction"]))]}])),(Identifier "level",VInteger 1)]} or Elt {eltName = Identifier "smallcaps", eltPos = Just "bug.typ" (line 1, column 2), eltFields = fromList [(Identifier "body",VContent (fromList [Elt {eltName = Identifier "heading", eltPos = Just "bug.typ" (line 1, column 2), eltFields = fromList [(Identifier "body",VContent (fromList [Elt {eltName = Identifier "text", eltPos = Just "bug.typ" (line 1, column 2), eltFields = fromList [(Identifier "body",VContent (fromList [Txt "Introduction"]))]}])),(Identifier "level",VInteger 1)]}]))]}
expecting end of input
```

But it compiles fine with `typst compile bug.typ` (using Typst 0.13.1), producing a heading in small capitals as we wanted:

<img width="408" height="180" alt="Cropped Typst output for the above bug.typ file, showing a single heading reading INTRODUCTION in small capital letters" src="https://github.com/user-attachments/assets/11335964-7a96-4497-8e5b-c8bf4b86c120" />

It's not limited to usage of `smallcaps` specifically but happens with most [functions that operate on text](https://typst.app/docs/reference/text/). For some of them, there is no error and Pandoc appears to run fine, but then produces no actual heading in the output. For only one of them it works completely fine. Here are all the text functions grouped by into which of these categories they fall (all of them work fine when compiling with Typst itself, of course):

<table>
<tr>
<th>❌ Error</th>
<th>0️⃣ No heading in output</th>
[...]

PR #11881, opened 2026-09-20
Typst reader/writer: handle multi-paragraph highlight bodies

sghng:
Follow-up to #11879 (which in turn addressed #10747).

`highlight` only had an inline handler, so a body containing
paragraph breaks:

```typst
    #highlight[
      Para one.

      Para two.
    ]
```

failed with `unexpected Elt {text …} expecting end of input`:
`pWithContents pInlines` stops at the parbreak and its trailing `eof`
then fails. (Before #11879 such content was silently dropped instead.)

This mirrors how `box` is handled:

- Reader: new block handler produces `Div ("", ["mark"], [])`;
  single-paragraph bodies still parse inline as `mark` spans through
  the existing handler.
- Writer: `Div ("", ["mark"], [])` renders as `#highlight[...]`,
  symmetric with the existing `Span` `.mark` handling, so round-trips
  are faithful.
- Tests: extend `test/command/typst-highlight.md` with the
  multi-paragraph case (reader → native, typst → typst round-trip,
  markdown `::: mark` div → typst).


jgm:
I feel like this is a more general issue and needs a more general fix. For example, in typst you can do
```
#emph[hi

there]
```
and pandoc won't handle this either.

sghng:
@jgm You're right, it's more general than highlight. Testing the whole family: `emph`, `strong`, `underline`, `strike`, `sub`, `super`, `smallcaps`, `highlight`, `link`, `lower`, and `upper` all fail with a parse error when their bodies contain a paragraph break, in block or inline position, also inside list items and table cells. Elements with block handlers (text, box, quote, align, place) happened to work, and footnote was always fine since its body is parsed as blocks.

I've rewritten the PR with a general fix (about which I'm not very confident...) `fixNesting` (#11017) already hoisted a lone block child out of an inline element; it now also splits any inline element at the paragraph breaks in its content, re-wrapping each paragraph in the element and keeping real parbreaks between the pieces:

```typst
#emph[hi

there]
```

reads as two paragraphs, each with `Emph`, which is also how Typst lays it out. A mid-paragraph occurrence such as `hello #emph[a … blank line … b] world` splits the surrounding paragraph instead of failing. Since a pandoc inline can't span paragraphs, applying the styling per paragraph is the closest structural rendering.

PR #11884, opened 2026-09-21
Typst reader: capture block and box fill as background-color attribute

sghng:
The block handler reads only the body, so a filled block loses both its boundary and its color: the region silently disappears from the AST, leaving filters and writers no way to recover it. A filled `box` keeps its span but drops the color the same way.

When a block or box has a fill, emit a `div` or `span` carrying the color as a `background-color` attribute, mirroring the LaTeX reader's `\colorbox` handling. Unfilled blocks and boxes stay transparent, as do explicit `fill: none` and `fill: auto`.

Writers ignore the attribute. Lua filters can read it and do whatever they like with the region: style it, reclassify it by color, or pass it through to a format that supports backgrounds natively.

Tests: extend the typst reader golden fixture with unfilled, none, and filled blocks and boxes.


changelog.md
```diff
@@ -41,6 +41,12 @@
 
   * Typst reader:
 
+    + Capture a `block`'s or `box`'s `fill` as a `background-color`
+      attribute on the resulting `div` or `span`, mirroring the LaTeX
+      reader's `\colorbox` handling. Unfilled blocks and boxes and
+      explicit `fill: none`/`auto` stay transparent. Writers ignore the
+      attribute; it keeps the region's boundary and color available to
+      filters.
```
jgm: Certainly you shouldn't be adding to the changelog for an already released version!
If this is AI generated, please review carefully.
Do not add anything to the changelog.  The changelog will be created manually on release; all relevant info needs to go in the commit message.

jgm:
I'm ambivalent about this. A very typst-specific color specification is getting into the attribute.

changelog.md
sghng: Sorry about that, I mistook it for the unreleased section (as in `typst-hs`)

sghng:
> I'm ambivalent about this. A very typst-specific color specification is getting into the attribute.

Here's my rationale for why this is useful

1. In an ideal world, everyone would use `#highlight` for highlighting, since it is semantic markup rather than style. But Typst applies highlight to text only: math inside a highlighted region is never painted, neither inline nor display. See typst/typst#5872.
2. So until that's fixed in Typst, users who need a visibly marked region that contains math work around the limitation with block and box plus a fill, which is the only construct in Typst that covers math at all.
3. I agree that the fill color is style rather than markup, and pandoc should not try to convert it into anything. The gap this PR addresses is that the information currently disappears entirely: a filled block loses both its boundary and its color in the AST, and a filled box keeps the span but drops the color. Without the fill in the tree, a downstream Lua filter cannot even tell that the region was marked.

On the specific concern about a typst-specific color specification in the attribute: happy to normalize the value to plain hex or CSS rgb() form instead of Typst's repr, if that makes the attribute feel less Typst-flavored.

In my own use case, I don't care about the specific color at all. I just need a way to pass the information to Pandoc like "hey, the user painted this with a color", and my Lua filter will capture that and emit actual highlighting.

sghng:
I opened #11885, since the OMML covered by `mark` doesn't get highlight is a standalone Docx writer bug.

jgm:
OK, looks good.

PR #11885, opened 2026-09-21
Docx writer: highlight math runs in marked content

sghng:
A marked region containing math loses its highlight on the math runs, while the HTML and LaTeX writers highlight the same content, math included, through `<mark>` and `\hl`. Reproduced from markdown:

```
% pandoc -f markdown+mark -t docx
Marked: ==alpha $x^2 + y$ beta==.
```

image

The text runs get `w:highlight`; the OMML runs do not. The same AST comes from the Typst reader's `#highlight` (#10747) and from the LaTeX reader's `\hl`.

Cause: the mark span handler wraps its content in `withTextProp (w:highlight ...)`, which accumulates into the writer's text-property environment. Text runs consult that environment when emitting `w:rPr`, but the `Math` case hands the expression to `convertMath`, which builds its OMML directly and never looks at it.

Fix: after conversion, if the ambient text properties contain a highlight, inject it into every `m:r` of the converted expression as a `w:rPr` child, placed after an optional `m:rPr`. That is where OMML takes Word run properties, and it is the shape Word itself emits when parts of an equation are highlighted. Unmarked math is unchanged.

Tests: a docx writer golden with mark spans containing inline and display math beside an unmarked control.

Split out from #11881, since this is a standalone writer bug independent of the Typst reader changes.


jgm:
For the first example in the golden test I am seeing
image

That doesn't look great.  Is there a way to make it look more like the display example, with a constant width highlighter?

sghng:
> For the first example in the golden test I am seeing image
> 
> That doesn't look great. Is there a way to make it look more like the display example, with a constant width highlighter?

@jgm Yes I noticed that as well. I have already came up with a fix locally. I alos discovered a few more bugs in Pandoc (can't believe I need so many patches to make my Typst -> Word work...), I am patching all of them on a local HEAD and will split them into PRs later...

